PR #13, opened 2023-02-03
chore: add logs for share extension

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we add logs to the share extension

----
##### References


wire-ios-system/Source/WireLogger.swift
```diff
@@ -21,11 +21,13 @@ import Foundation
 public struct WireLogger: LoggerProtocol {
 
   public static var provider: LoggerProtocol?
+  public let payloadEncoder: JSONEncoder
```
lewandowskit93: quick note:
this should be private

wire-ios/Wire-iOS Share Extension/ShareExtensionViewController.swift
```diff
@@ -281,27 +306,51 @@ final class ShareExtensionViewController: SLComposeServiceViewController {
...
+                logger.info(
+                    logger.prepareMessage(
+                        title: "SHARING",
+                        payload: SharingLogPayload(status: .error)
+                    )
+                )
                 if let errorDescription = (error as? UnsentSendableError )?.errorDescription {
                     let alert = UIAlertController.alertWithOKButton(title: nil, message: errorDescription)
-
+                    logger.info("we hit an error: \(errorDescription)")
```
lewandowskit93: quick note:
to be removed

wire-ios/Wire-iOS Share Extension/ShareExtensionViewController.swift
```diff
@@ -281,27 +306,51 @@ final class ShareExtensionViewController: SLComposeServiceViewController {
...
+                    )
+                )
                 self.popConfigurationViewController()
 
                 let alert = UIAlertController.alertWithOKButton(title: "share_extension.timeout.title".localized, message: "share_extension.timeout.message".localized)
 
                 self.present(alert, animated: true)
 
             case .error(let error):
+                logger.info(
```
lewandowskit93: quick note:
this goes down and includes error description

lewandowskit93 (commented):
I've done a self review here as well

wire-ios-system/Source/WireLogger.swift
```diff
@@ -136,4 +148,5 @@ public protocol LoggerProtocol {
   func error(_ message: String, attributes: LogAttributes?)
   func critical(_ message: String, attributes: LogAttributes?)
 
+  func prepareMessage<Payload: LoggablePayload>(title: String, payload: Payload) -> String
```
johnxnguyen: Why is this a protocol requirement when the `WireLogger` implements it?

johnxnguyen (commented):
I like the direction this is going, but I think we can already make a few minor improvements. I'd like to be able to use an api that allows me to pass in either a simple string message (as we currently have), or also a structure that prints to a json payload. Something like:

```swift
logger.info("this is a string")
logger.info(StructuredLog(status: .done))
```

Suppose we have a protocol to define a log:

```swift
protocol LogConvertible {
    
    var logDescription: String { get }

}
```

Suppose we have these log methods: 

```swift
func info<Message: LogConvertible>(message: Message, attributes: LogAttributes?) {
  self.log(level: .info, message: message.logDescription, attributes: attributes)
}
```

Then we defer all of the work to converting a message to a string to the actual message itself:

```swift
extension String: LogConvertible {
  
    var logDescription: String { self }

}

struct SharingLog: LogConvertible {

    let status: SharingStatus

    var logDescription: String {
        let payload = ...
        return "SHARING: \(payload)"
    }

}
```

What do you think?

wire-ios-system/Source/WireLogger.swift
lewandowskit93: I've thought about aby other payloads that may be shared here. So there is no more need to implement that.

lewandowskit93:
> ```swift
> logDescription
> ```

I think it looks good, could you push it?

wire-ios-system/Source/WireLogger.swift
johnxnguyen: I don't quite understand what you mean.

wire-ios/Wire-iOS/Sources/Managers/DatadogWrapper.swift
```diff
@@ -164,6 +167,14 @@ extension RemoteMonitoring.Level {
 
 extension DatadogWrapper: WireSystem.LoggerProtocol {
 
+    public func prepareMessage<Payload>(title: String, payload: Payload) -> String where Payload : Encodable {
+        guard let payloadData = try? payloadEncoder.encode(payload),
+              let payloadString = String(data: payloadData, encoding: .utf8) else {
+          return "\(title): PAYLOAD ENCODING ERROR"
+        }
+        return "\(title): \(payloadString)"
+    }
```
johnxnguyen: This is a duplication of the method in the `WireLogger`, and this method is never actually called. I would argue that we don't need this protocol requirement.

lewandowskit93: I would suggest to move it to some other protocol, or do the encoding in the message itself, or remove it from protocol and leave only on WireLogger if we won't use it on DatadogWrapper.
This method is called only on WireLogger currently.

wire-ios-system/Source/WireLogger.swift
lewandowskit93: It is `LoggablePayload` not `LoggableProtocol` :) So that any payload could type could be passed here and transformed to string. However in the upgrade you've suggested we could move encoding to `logDescription` of `SharingLog` that would implement `LogConvertible`.

agisilaos:
Closing this in favor of #95

PR #22, opened 2023-02-09
chore: update core crypto

David-Henner:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR does two things:
1. Move core crypto to data model
2. Update core crypto to `0.6.0-rc.7`

### Moving core crypto to data model

Up until now, core crypto was distributed as a local swift package and located in the root of the dependency tree (wire-ios).
Thanks to the monorepo, we can relocate this package in the data model layer.

This allows us to:
- get rid of all the core crypto types defined in data model, because we can access core crypto directly 
- get rid of the convoluted setup process involving passing a setup closure upstream in order to instantiate core crypto

### Updating core crypto

We're updating core crypto to `0.6.0-rc.7`, which contains APIs for proteus that were missing in our previously used version.

----

----

----
##### References


wire-ios-notification-engine/WireNotificationEngine.xcodeproj/project.pbxproj
```diff
@@ -611,7 +611,7 @@
 					"@executable_path/Frameworks",
 					"@loader_path/Frameworks",
 				);
-				ONLY_ACTIVE_ARCH = NO;
+				ONLY_ACTIVE_ARCH = YES;
```
johnxnguyen, line 487: Why did this change?

wire-ios-request-strategy/WireRequestStrategy.xcodeproj/project.pbxproj
```diff
@@ -2200,7 +2200,7 @@
 			baseConfigurationReference = 166901961D7075D7000FE4AF /* WireRequestStrategy.xcconfig */;
 			buildSettings = {
 				INFOPLIST_FILE = Resources/Info.plist;
-				ONLY_ACTIVE_ARCH = NO;
+				ONLY_ACTIVE_ARCH = YES;
```
johnxnguyen, line 2211: Why did this change?

wire-ios-share-engine/WireShareEngine.xcodeproj/project.pbxproj
```diff
@@ -521,7 +521,7 @@
 			buildSettings = {
 				CODE_SIGN_IDENTITY = "";
 				INFOPLIST_FILE = Resources/Info.plist;
-				ONLY_ACTIVE_ARCH = NO;
+				ONLY_ACTIVE_ARCH = YES;
```
johnxnguyen, line 524: Why did this change?

wire-ios-sync-engine/WireSyncEngine.xcodeproj/project.pbxproj
```diff
@@ -3723,7 +3717,7 @@
 				GCC_PREFIX_HEADER = "$(SRCROOT)/Source/WireSyncEngine-iOS.pch";
 				INFOPLIST_FILE = "$(SRCROOT)/Resources/WireSyncEngine-ios-Info.plist";
 				MOMC_NO_INVERSE_RELATIONSHIP_WARNINGS = YES;
-				ONLY_ACTIVE_ARCH = NO;
+				ONLY_ACTIVE_ARCH = YES;
```
johnxnguyen, line 3720: Why did this change?

wire-ios-notification-engine/WireNotificationEngine.xcodeproj/project.pbxproj
David-Henner: This is because the core crypto library doesn't support `arm64`. In `wire-ios` this setting was already the default, allowing us to build with core crypto on intel Macs (because `arm64` is not an active architecture there).
Now, since core crypto was moved to data model, we need to replicate this setting from the leaf to the root node of the dependency tree.

Unfortunately, this means it won't build for M1 Macs when building for simulator (which is already the case?). 
This is also what was blocking us from using remote swift package earlier. Because core crypto team couldn't build an xcframework from the library. It's possible they have a fix already, we'll need to ask them

wire-ios-data-model/CoreCrypto/Package.swift
```diff
@@ -28,6 +28,6 @@ let package = Package(
         .target(
             name: "CoreCryptoSwift",
             dependencies: ["LibCoreCrypto"]
-        )
+        ),
```
netbe, line 31: extra comma ?

netbe:
@David-Henner is there a reason to use corecrypto locally instead of using the package remotely?

wire-ios-data-model/CoreCrypto/Package.swift
David-Henner: I just copy pasted from the package, I don't think that's an issue

David-Henner:
> @David-Henner is there a reason to use corecrypto locally instead of using the package remotely?

I think it boils down to the core crypto team not being able to produce an xcframework due to the static library not supporting `arm64`.  There's more details on this page https://wearezeta.atlassian.net/wiki/spaces/IOS/pages/677839071/How+to+setup+core+crypto+swift+package+dependency

wire-ios-notification-engine/WireNotificationEngine.xcodeproj/project.pbxproj
johnxnguyen: >  this means it won't build for M1 Macs when building for simulator

Right... this is the problem I was facing. So CoreCrypto is blocking us all from building for simulator in Apple silicon. This will be a problem for CI machines running Apple silicon. For now we can accept it but we absolutely need to fix this.

David-Henner: Completely agree. I talked to cc team and they can build for m1 simulator now so they'll do a release soon, that way we'll be able to use the remote package

johnxnguyen (approved):

wire-ios-data-model/Source/ManagedObjectContext/NSManagedObjectContext+MLSController.swift
```diff
@@ -17,6 +17,7 @@
 //
 
 import Foundation
+import CoreCryptoSwift
```
netbe, line 20: needed?

johnxnguyen: Yes because the `CoreCryptoProtocol` used to be defined in data model but has been removed, and we want to refer to the one defined in `CoreCryptoSwift`.

netbe (approved):

PR #44, opened 2023-02-23
feat: migrate proteus sessions from Cryptobox to Core Crypto FS-894

KaterinaWire:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR introduce the Cryptobox migration to Core Crypto. 
Migration process:
1. When useSession becomes active, we check if we need to start the migration. The conditions of the migration: the `proteusViaCoreCrypto` developer flag isOn and cryptobox directory exists.

2. If migration is needed, we transit the app to the "migration" state.

3. We use` proteusCryptoboxMigrate(path: String)` method form CoreCrypto to execute the migration.

4. If we don't get any errors, we delete the Cryptobox directory.
5. We transit the app to the "authenticated" state.

----

----

----
##### References


wire-ios-data-model/Source/Proteus/ProteusService.swift
```diff
@@ -314,4 +314,21 @@ public final class ProteusService: ProteusServiceInterface {
         }
     }
 
+    // MARK: - cryptoboxMigration
+
+        enum MigrationError: Error {
+            case failedToMigrateData
+        }
+
+        public func proteusCryptoboxMigrate(path: String) throws {
```
johnxnguyen: Name suggestion: `migrateCryptoboxSessions(at url: URL)`

wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack+Migration.swift
```diff
@@ -149,4 +149,9 @@ extension CoreDataStack {
         }
     }
 
+    public var cryptoboxDirectoryExists: Bool {
```
johnxnguyen: Why is this in the Core Data Stack? Could we put it somewhere more appropriate?

KaterinaWire: I was trying to find a class that can take this responsibility and can provide a public `cryptoboxDirectoryExists` property. `CoreDataStack` currently contains the info about `account`,`accountContainer` and `applicationContainer`.
Another idea was to use `ZMUserSession`.
I have not found any reason why ProteusService should be responsible for providing such info.

wire-ios-data-model/Source/Proteus/ProteusService.swift
KaterinaWire: Done

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -811,9 +811,20 @@ public final class SessionManager: NSObject, SessionManagerType {
...
-                                                                         with: coreDataStack)
-                        completion(userSession)
+                        let userSession = self.startBackgroundSession(for: account, with: coreDataStack)
+
+                        /// If `proteusViaCoreCrypto` developer flag isOn and cryptobox directory exists,
+                        /// we need to migrate the existing proteus sessions, prekeys, and identity key to CoreCrypto keystore.
+                        let cryptoboxNeedsMigration = DeveloperFlag.proteusViaCoreCrypto.isOn && coreDataStack.cryptoboxDirectoryExists
+
+                        if cryptoboxNeedsMigration {
+                            self.delegate?.sessionManagerWillMigrateAccount {
```
netbe: I think we can spare a part of this comment as it is literally what you wrote in code 
```suggestion
                        let cryptoboxNeedsMigration = DeveloperFlag.proteusViaCoreCrypto.isOn && coreDataStack.cryptoboxDirectoryExists

                        if cryptoboxNeedsMigration {
                            /// we need to migrate the existing proteus sessions, prekeys, and identity key to CoreCrypto keystore.

                            self.delegate?.sessionManagerWillMigrateAccount {
```

wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack+Migration.swift
netbe: what about a MigrationHelper struct with static methods?

David-Henner (commented):
LGTM 👌 Just waiting for the tests

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
KaterinaWire: Done

wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack+Migration.swift
KaterinaWire: I added `CryptoboxMigrationManager` and updated this PR

wire-ios-sync-engine/Tests/Source/SessionManager/CryptoboxMigrationManagerTests.swift
```diff
@@ -0,0 +1,123 @@
...
+        proteusViaCoreCryptoFlag.isOn = false
+        super.tearDown()
+    }
+
+    // MARK: - Verifying if migration is needed
+
+    func test_itReturnsTrue_WhenMigrationIsNeeded() {
+        // Given
+        let cryptoboxDirectory = FileManager.keyStoreURL(accountDirectory: accountDirectory, createParentIfNeeded: false)
+        XCTAssertTrue(FileManager.default.fileExists(atPath: cryptoboxDirectory.path))
```
johnxnguyen, line 45: Where is the cryptobox directory being created in these tests?

wire-ios-sync-engine/Tests/Source/SessionManager/CryptoboxMigrationManagerTests.swift
```diff
@@ -0,0 +1,123 @@
...
+
+    func test_itReturnsFalse_WhenMigrationIsNotNeeded_AccountDirectoryDoesNotExist() {
+        // Given
+        let fakeAccountDirectory: URL = URL(fileURLWithPath: "something")
+
+        // When
+        proteusViaCoreCryptoFlag.isOn = true
+
+        // Then
+        XCTAssertFalse(sut!.isNeeded(in: fakeAccountDirectory))
```
johnxnguyen: It would be a more accurate test it the actual cryptobox directory was deleted in the "given" and the real account url is passed in to the migration manager.

wire-ios-sync-engine/Tests/Source/SessionManager/CryptoboxMigrationManagerTests.swift
```diff
@@ -0,0 +1,123 @@
...
+        let cryptoboxDirectory = FileManager.keyStoreURL(accountDirectory: accountDirectory, createParentIfNeeded: false)
+        XCTAssertTrue(FileManager.default.fileExists(atPath: cryptoboxDirectory.path))
+
+        // When
+        proteusViaCoreCryptoFlag.isOn = true
+        syncMOC.proteusService = mockProteusService(with: nil)
+        try? sut?.perform(in: accountDirectory, syncContext: syncMOC)
+
+        // Then
+        XCTAssertFalse(FileManager.default.fileExists(atPath: cryptoboxDirectory.path))
```
johnxnguyen: Two things:
- If you mark the test as `throws` then the test will fail when you `try`.
- I think it'd be good to also make assertions on the proteus service, checking that certain methods are being called.

wire-ios-sync-engine/Source/SessionManager/CryptoboxMigration/CryptoboxMigrationManager.swift
```diff
@@ -0,0 +1,82 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+enum CryptoboxMigrationError: Error {
+    case failedToMigrateData
+    case failedToRemoveDirectory
```
johnxnguyen: Suggestion: 'failedToDeleteLegacyData'

wire-ios-sync-engine/Source/SessionManager/CryptoboxMigration/CryptoboxMigrationManager.swift
```diff
@@ -0,0 +1,82 @@
...
+
+class CryptoboxMigrationManager: CryptoboxMigration {
+
+    private let logger = Logging.cryptoboxMigration
+
+    func isNeeded(in accountDirectory: URL) -> Bool {
+        let cryptoboxDirectory = getDirectory(in: accountDirectory)
+        let cryptoboxDirectoryExists = FileManager.default.fileExists(atPath: cryptoboxDirectory.path)
+
+        return DeveloperFlag.proteusViaCoreCrypto.isOn && cryptoboxDirectoryExists
```
johnxnguyen: Suggestion:

```swift
        guard DeveloperFlag.proteusViaCoreCrypto.isOn else { return false }
        let cryptoboxDirectory = getDirectory(in: accountDirectory)
        let cryptoboxDirectoryExists = FileManager.default.fileExists(atPath: cryptoboxDirectory.path)
        return cryptoboxDirectoryExists
```

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -811,9 +812,20 @@ public final class SessionManager: NSObject, SessionManagerType {
...
-                        let userSession = self.startBackgroundSession(for: account,
-                                                                         with: coreDataStack)
-                        completion(userSession)
+                        let userSession = self.startBackgroundSession(for: account, with: coreDataStack)
+
+                        if self.cryptoboxMigrationManager.isNeeded(in: coreDataStack.accountContainer) {
+                            /// We need to migrate the existing proteus sessions, prekeys, and identity key to CoreCrypto keystore.
+                            self.delegate?.sessionManagerWillMigrateAccount {
+                                userSession.syncContext.performAndWait {
+                                    try? self.cryptoboxMigrationManager.perform(in: coreDataStack.accountContainer,
```
johnxnguyen: I suggest as a first implementation to fail hard in case of an error, making sure that we can easily see why it failed. I suggest to do this with a `do catch` and an assertion, possibly some logs too.

wire-ios-sync-engine/Tests/Source/SessionManager/CryptoboxMigrationManagerTests.swift
KaterinaWire: In the `setupKeyStore` method of the `MessagingTest` class.

wire-ios-sync-engine/Source/SessionManager/CryptoboxMigration/CryptoboxMigrationManager.swift
KaterinaWire: Doe

wire-ios-sync-engine/Tests/Source/SessionManager/CryptoboxMigrationManagerTests.swift
KaterinaWire: Done

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
KaterinaWire: Done

wire-ios-sync-engine/Source/SessionManager/CryptoboxMigration/CryptoboxMigrationManager.swift
KaterinaWire: Done

wire-ios-sync-engine/Tests/Source/SessionManager/CryptoboxMigrationManagerTests.swift
KaterinaWire: Done

wire-ios-sync-engine/Source/SessionManager/CryptoboxMigration/CryptoboxMigrationManager.swift
KaterinaWire: Done

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -811,9 +812,24 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+
+                        if self.cryptoboxMigrationManager.isNeeded(in: coreDataStack.accountContainer) {
+                            /// We need to migrate the existing proteus sessions, prekeys, and identity key to CoreCrypto keystore.
+                            self.delegate?.sessionManagerWillMigrateAccount {
+                                userSession.syncContext.performAndWait {
+                                    do {
+                                    try self.cryptoboxMigrationManager.perform(in: coreDataStack.accountContainer,
+                                                                                syncContext: userSession.syncContext)
+                                    } catch {
+                                        log.error("Failed to migrate data from CryptoBox CoreCrypto keystore, error : \(error.localizedDescription)")
```
johnxnguyen: What do you think about intentionally crashing here?

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -811,9 +812,24 @@ public final class SessionManager: NSObject, SessionManagerType {
                     if error != nil {
                         self.delegate?.sessionManagerDidFailToLoadDatabase()
                     } else {
-                        let userSession = self.startBackgroundSession(for: account,
-                                                                         with: coreDataStack)
-                        completion(userSession)
+                        let userSession = self.startBackgroundSession(for: account, with: coreDataStack)
+
+                        if self.cryptoboxMigrationManager.isNeeded(in: coreDataStack.accountContainer) {
```
David-Henner: I just have a concern related to where we check and start the migration. There's an `if else` statement starting at line 797 which looks like this

```
if let session = self.backgroundUserSessions[account.userIdentifier] {
    // session already loaded. call completion handler with session
} else {
    // create session
}
```

The migration happens in the `else` block. So I wonder what happens when there are sessions already loaded. Do we not migrate them ? If we don't, should we? Or do we only support proteus via core crypto when creating new sessions?

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
KaterinaWire: done

johnxnguyen (approved):

David-Henner (approved):

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnxnguyen: As discussed offline, we only need to migrate once when the session is first loaded. Since sessions are kept in memory, each session must be loaded at least once during the lifetime of the app. Since the app will upgrade, it will need to be launched.

PR #51, opened 2023-02-27
fix: Implement `swiftlint --fix` in mono repo

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we add the ability to fix SwiftLint warnings for the entire mono-repo project. We use a [pre-commit](https://pre-commit.com) hook which allows us to run SwiftLint on modified files before we commit the changes. Documentation will be added on Confluence as soon as this is merged.

A tool needs to be installed for this to work and the process is this:

- `brew install pre-commit`
Then go into the mono-repo folder and in your terminal type:
- `pre-commit install`

You're ready to go. now `pre-commit` will run automatically on `git commit`

----
##### References


netbe:
@agisilaos There is no issue with swiftlint modifying files during build time?
I would rather have it run before pushing commits `pre-push` hook, or as a `post-commit` hook

agisilaos:
> @agisilaos There is no issue with swiftlint modifying files during build time?
> I would rather have it run before pushing commits `pre-push` hook, or as a `post-commit` hook

@netbe I didn't face any issues no. When you build the project, SwiftLint will take care of the linting issues.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
+  - repo: https://github.com/realm/SwiftLint
+    rev: 0.50.3
+    hooks:
+      - id: swiftlint
+      - entry: swiftlint --fix
```
johnxnguyen: How does this work?

agisilaos: After I finish with the changes, I will create a document explaining the steps so everyone can be aware of how it works and what needs to be installed.

johnxnguyen: But can you explain it in the pull request too? Pull requests are also a form of documentation and it's helpful to be able to review it.

agisilaos: Just updated the PR with more information

johnxnguyen: So you have to install an additional tool on the local machine for this to work?

johnxnguyen (commented):
What's the advantage of doing this via a pre-commit hook over a build phase script?

.pre-commit-config.yaml
agisilaos: Yes I updated the PR description with the steps!

agisilaos:
@johnxnguyen I think @netbe's concern was modifying files during build time. Which I understand. Pre-hook can be a better alternative and also gives us the opportunity to have a more central system for improving the codebase not only via SwiftLint but using other tools in the future too.

johnxnguyen:
> @johnxnguyen I think @netbe's concern was modifying files during build time.

@agisilaos ok makes sense. Another question though, how does this work when there is currently no `.swiftlint.yml` file in the root directory, but in each of the project directories? Does it matter?

agisilaos:
> > @johnxnguyen I think @netbe's concern was modifying files during build time.
> 
> @agisilaos ok makes sense. Another question though, how does this work when there is currently no `.swiftlint.yml` file in the root directory, but in each of the project directories? Does it matter?

@johnxnguyen For now, it uses the .swiftlint file from each project. In the future though, I would like to centralize and have a single SwiftLint file. That it will take some time to implement though that's why I didn't go on that path with this PR.

netbe:
I didn't know there was a tool to setup git hooks... I think we can spare it and add to setup.sh a step to setup the pre-commit.

This is more a general comment on how we have things setup but depending on `homebrew` does not enforce the version of tools, so it can happen that one dev has one version of SwiftGen or Sourcery and others have another one which results in different outputs.

agisilaos:
> This is more a general comment on how we have things setup but depending on `homebrew` does not enforce the version of tools, so it can happen that one dev has one version of SwiftGen or Sourcery and others have another one which results in different outputs.

@netbe I think that it would be best to do on a separate PR? But yeah I agree with the comment. Definitely makes sense.

agisilaos:
Closing this in favor of #65

PR #52, opened 2023-02-27
feat: establish session from prekey - FS-1552 

netbe:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Refactor the `UserClient.establishSessionWithClient` method to use the `ProteusService` if needed.

### Solutions

Wherever we use the cryptobox methods (via keystore): 
* check the DeveloperFlag.proteusViaCoreCrypto 
* if enabled use the ProteusService to replicate the logic
* if disabled, preserve the existing logic with the keystore

### Dependencies (Optional)

This is based on https://github.com/wireapp/wire-ios-mono/pull/47 in order to use ProteusService

Needs releases with:

- https://github.com/wireapp/wire-ios-mono/pull/47 

#### Test Coverage (Optional)

----

----

----
##### References


wire-ios-data-model/Source/Model/UserClient/UserClient.swift
```diff
@@ -597,21 +598,76 @@ public extension UserClient {
...
+        return establishSessionWithClient(client, usingPreKey: preKey, proteusProviding: proteusProvider)
+    }
+
+    private func establishSession(through proteusService: ProteusServiceInterface,
+                                  client: UserClient,
+                                  sessionId: EncryptionSessionIdentifier,
+                                  preKey: String
+    ) -> Bool {
+        do {
+            // TODO: check if we should delete session if it exists before creating new one
```
johnxnguyen, line 644: From previous discussions we arrive at the conclusion that there is a valid case where you want to delete a session, such as resetting the session. It seems like this method serves two purposes depending on how it is called:

1. Establishing a new session for the first time
2. Resetting a session by deleting the old one and creating a new one, when called any time after the initial setup.

If this is true, then it would be better to split the responsibility over two separate methods. But as a first step we can simply preserve the existing logic, then refactor afterwards. What do you think?

wire-ios-data-model/Tests/Source/Model/UserClient/UserClientTests.swift
```diff
@@ -1076,3 +1076,79 @@ extension UserClientTests {
...
+            var preKeys: [(id: UInt16, prekey: String)] = []
+            sut.keysStore.encryptionContext.perform { sessionsDirectory in
+                preKeys = try! sessionsDirectory.generatePrekeys(0 ..< 2)
+            }
+
+            // WHEN
+            resultOfMethod = sut.establishSessionWithClient(clientB, usingPreKey: preKeys.first!.prekey, proteusProviding: mock)
+
+            // THEN
+            XCTAssertEqual(spy.accessEncryptionContextCount, 2)
```
johnxnguyen, line 1155: If you want some extra assertions, you could assert that the mock proteus service was never invoked.

wire-ios-data-model/Source/Model/UserClient/UserClient.swift
netbe: preserve and do it in a second time I guess. I don't know it would affect keystore for exmple

wire-ios-data-model/Tests/Source/Model/UserClient/UserClientTests.swift
netbe: yes I'll add that

wire-ios-data-model/Source/Model/UserClient/UserClient.swift
johnxnguyen: > I don't know it would affect keystore for exmple

I'm not sure what you mean. It would affect the keystore? Or you're asking if it would affect the keystore?

netbe: yeah that's my question, does this affect the keystore method? I guess no

KaterinaWire (approved):

wire-ios-data-model/Source/Model/UserClient/UserClient.swift
johnxnguyen: Right, yes we'd need to refactor it so that the same behavior is present whether we use cryptobox or core crypto.

johnxnguyen (approved):

PR #53, opened 2023-02-27
refactor: crypto stack setup FS-1577 FS-1578

David-Henner:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR is a refactoring of the crypto stack setup flow

#### In the main app

Split the crypto stack setup in the main app into two steps: `proteus` and `mls`
The `proteus` step happens before request strategies are created so that the `ProteusService` can be available to generate prekeys upon client registration. It creates an instance of core crypto that can serve proteus APIs.
The `mls` step is called in either one of two places. The first one being in `ZMUserSession` init if the self client is already registered. The second, if the client is not already registered on init, will be after client registration. This step will call `mlsInit` on core crypto and provide the client ID.

#### In the extensions

In the extensions, the client is already registered, so we setup the whole stack at once.

### Testing

----

----

 - [x] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References


wire-ios-data-model/Source/MLS/SafeCoreCrypto.swift
```diff
@@ -24,33 +24,57 @@ import CoreCryptoSwift
...
 
-public struct SafeCoreCrypto: SafeCoreCryptoProtocol {
+public class SafeCoreCrypto: SafeCoreCryptoProtocol {
     public enum CoreCryptoSetupFailure: Error, Equatable {
         case failedToGetClientIDBytes
     }
 
     private let coreCrypto: CoreCryptoProtocol
     private let safeContext: SafeFileContext
+    private var isReadyForMLS: Bool = false
```
johnxnguyen:
```suggestion
    private var isReadyForMLS = false
```

wire-ios-data-model/Source/MLS/SafeCoreCrypto.swift
```diff
@@ -24,33 +24,57 @@ import CoreCryptoSwift
...
+            path: path,
+            key: key,
+            entropySeed: nil
+        )
+
+        self.init(coreCrypto: coreCrypto, path: path)
+    }
+
+    public func mlsInit(clientID: String) throws {
+        guard !isReadyForMLS else { return }
```
johnxnguyen: Would it be better to throw if not ready for MLS? Otherwise how would we know it succeeded?

David-Henner: This is just to make sure we don't try to call `coreCrypto.mlsInit` more than once.
We call `setupCryptoStack(stage: .mls)` twice. Once on session init, and once after client registration.

However, now that I think about it, we either:
1. Have a client id on session init and call `mlsInit`. Then don't go through client registration
2. Don't have client id on session init, so we'd only call `mlsInit` after registration

So with that in mind we wouldn't need a safeguard here.

johnxnguyen: Ah I misread and misinterpreted the guard. Perhaps if we could rename it to `didInitializeMLS` to more clearly express that we don't init MLS more than once.

wire-ios-data-model/Source/MLS/SafeCoreCrypto.swift
```diff
@@ -24,33 +24,58 @@ import CoreCryptoSwift
...
+        guard let clientIdBytes = ClientId(from: clientID) else {
+            throw CoreCryptoSetupFailure.failedToGetClientIDBytes
+        }
+
+        try coreCrypto.mlsInit(clientId: clientIdBytes)
+        didInitializeMLS = true
     }
 
-    init(coreCrypto: CoreCryptoProtocol, coreCryptoConfiguration: CoreCryptoConfiguration) {
+    init(coreCrypto: CoreCryptoProtocol, path: String) {
```
netbe: maybe we need to be more  specific about the path here since we modify the path afterwards

wire-ios-data-model/Tests/MLS/CoreCryptoConfigProviderTests.swift
```diff
@@ -55,7 +55,31 @@ class CoreCryptoFactoryTests: ZMConversationTestsBase {
...
+                selfUser: selfUser
+            )
+
+            // THEN
+            XCTAssertEqual(configuration.key, key.base64EncodedString())
+            XCTAssertEqual(configuration.path, self.expectedPath(selfUser))
+        }
+    }
+
+    func test_itReturnsFullCoreCryptoConfiguration() throws {
```
netbe, line 82: method name does not match the test

netbe: test_createFullConfigurationThrowsErrorsIfAlreadyCreated ?

wire-ios-data-model/Tests/MLS/SafeCoreCryptoTests.swift
```diff
@@ -52,4 +54,53 @@ class SafeCoreCryptoTests: ZMBaseManagedObjectTest {
...
+
+        // WHEN
+        try sut.mlsInit(clientID: "id")
+
+        XCTAssertEqual(mlsInitCalls, 1)
+    }
+
+    // MARK: - Helper
+
+    private func createTempFolder() -> URL {
```
netbe: this has been moved to WireTesting XCTestCase+Helpers

johnxnguyen (commented):
Looks good to me, but I'll wait for the open conversations to be resolved before approving.

wire-ios-data-model/Tests/MLS/CoreCryptoConfigProviderTests.swift
David-Henner: the test name is correct, it's just that github auto-collapsed two test methods together

wire-ios-data-model/Source/MLS/SafeCoreCrypto.swift
David-Henner: I'm not sure what else would make sense, do you have a suggestion?

wire-ios-data-model/Tests/MLS/SafeCoreCryptoTests.swift
David-Henner: removed

wire-ios-data-model/Source/MLS/SafeCoreCrypto.swift
netbe: `databasePath`?
or coredataStackDirectory and move the `let directoryPathUrl = URL(fileURLWithPath: path).deletingLastPathComponent()` outside of init

David-Henner: changed to `databasePath`

netbe (approved):

johnxnguyen (approved):

PR #55, opened 2023-02-28
chore: get file lock once when decrypting a batch of events FS-1580

johnxnguyen:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

In this PR we address a TODO from a previous PR for decrypting update events with core crypto. The existing flow with decrypting proteus events via Cryptobox was to first get safe access to cryptobox, then iterate through a batch of update events and decrypt them one by one. Finally, once the decrypted update events were stored in the database, the file lock on cryptobox was released.

While this exact behavior was not yet implemented when decrypting events via Core Crypto, we still had safe access to Core Crypto at the level we decrypt a single event. However, since it is possible that we may decrypt a large batch of events at one time, it may be inefficient to get the lock, encrypt one event, release the lock, get the lock, decrypt one more event, release the lock.... 100 or so times.

Instead, we should get the lock once, decrypt a batch of events, then release the lock. This PR introduces a way to achieve this by adding a new `ProteusService` method to ensure safe access to Core Crypto across the life time of a closure.

#### Test Coverage

- updated `EventDecoder` tests to assert that `performBatchedOperations` is called
- added unit test to `ProteusServiceTests` asserting safe access to core crypto when invoking the `performBatchedOperations` method

----

----

----
##### References

### DIFF
--- a/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder.swift
+++ b/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder.swift
@@ -135,32 +135,32 @@ extension EventDecoder {
         _ events: [ZMUpdateEvent],
         startingAtIndex startIndex: Int64
     ) -> [ZMUpdateEvent] {
-        var decryptedEvents: [ZMUpdateEvent] = []
+        var decryptedEvents = [ZMUpdateEvent]()
+        let proteusService = syncMOC.proteusService!
 
-        // TODO: get core crypto lock
-        decryptedEvents = events.compactMap { event -> ZMUpdateEvent? in
-            switch event.type {
-            case .conversationOtrMessageAdd, .conversationOtrAssetAdd:
-                let proteusService = syncMOC.proteusService!
+        proteusService.performBatchedOperations {
+            decryptedEvents = events.compactMap { event -> ZMUpdateEvent? in
+                switch event.type {
+                case .conversationOtrMessageAdd, .conversationOtrAssetAdd:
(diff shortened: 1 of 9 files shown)


KaterinaWire (approved):

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -135,32 +135,32 @@ extension EventDecoder {
...
-                }
+        proteusService.performBatchedOperations {
+            decryptedEvents = events.compactMap { event -> ZMUpdateEvent? in
+                switch event.type {
+                case .conversationOtrMessageAdd, .conversationOtrAssetAdd:
+                    return self.decryptProteusEventAndAddClient(event, in: self.syncMOC) { sessionID, encryptedData in
+                        try proteusService.decrypt(
+                            data: encryptedData,
+                            forSession: sessionID
+                        )
```
David-Henner, line 149: isn't this attempting to acquire the lock too?

netbe (approved):

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder.swift
johnxnguyen: Yes, but we discussed that we will test it first to see if it is a problem. It may not be given that it is being called in the same process and the same thread.

David-Henner (approved):

PR #57, opened 2023-03-01
chore: Get rid of applyColorScheme method from various cell classes

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we get rid of the applyColorScheme method from those classes:

- `ConversationCreateOptionsCell`
- `ConversationCreateReceiptsCell`
- `ConversationCreateServicesCell`
- `ConversationCreateGuestsCell`

Cleaned up the appropriate tests files and added some documentation in `IconToggleCell`

----
##### References


netbe (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/Cells/ConversationCreateOptionsCell.swift
```diff
@@ -56,26 +56,32 @@ final class ConversationCreateOptionsCell: RightIconDetailsCell {
         }
     }
 
+    // MARK: - Overrides
+    // MARK: - Override setUp
```
johnxnguyen: Is this second pragma necessary?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/Cells/ConversationCreateOptionsCell.swift
```diff
@@ -56,26 +56,32 @@ final class ConversationCreateOptionsCell: RightIconDetailsCell {
...
-
+        setupIconForCell()
         setupAccessibility()
     }
 
-    override func applyColorScheme(_ colorSchemeVariant: ColorSchemeVariant) {
-        super.applyColorScheme(colorSchemeVariant)
-
+    // MARK: Methods
+    /// Set up the Icon its color for the cell
```
johnxnguyen:
```suggestion
    /// Sets up the icon and its color.
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/Cells/ConversationCreateOptionsCell.swift
```diff
@@ -56,26 +56,32 @@ final class ConversationCreateOptionsCell: RightIconDetailsCell {
...
+    private func setupIconForCell() {
```
johnxnguyen:
```suggestion
    private func setUpIcon() {
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/Cells/ConversationCreateOptionsCell.swift
agisilaos: Fixed it!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/Cells/ConversationCreateOptionsCell.swift
agisilaos: Done

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/Cells/ConversationCreateOptionsCell.swift
agisilaos: Done!

agisilaos:
@johnxnguyen Ready for another round of review!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/Cells/ConversationCreateOptionsCell.swift
```diff
@@ -89,17 +94,18 @@ final class ConversationCreateOptionsCell: RightIconDetailsCell {
     }
 }
 
+// MARK: - ConversationCreateOptionsCell Extension
 extension ConversationCreateOptionsCell: ConversationCreationValuesConfigurable {
     func configure(with values: ConversationCreationValues) {
-        let guests = values.allowGuests.localized.localizedUppercase
-        let services = values.allowServices.localized.localizedUppercase
-        let receipts = values.enableReceipts.localized.localizedUppercase
+        let guests = values.allowGuests.localized.capitalized
```
KaterinaWire, line 100: Do we need use `capitalized`? In general we should not use `capitalized` , because it uppercases the first character in each word, and we don't need it. We still have many places in the app where we need to fix it.

agisilaos: I think in this case it's just fine. We want to get rid of `.Uppercase` anyways. 

Screenshot 2023-03-08 at 17 29 17

KaterinaWire: Yes, in this case it is not a problem, because "on" or "off" is just one word. But in general we shouldn't capitalize because by design we don't have use cases where we need to capitalize every word. I think the usage in this particular case can make the code confusing.
What if just` values.allowGuests.localized`?

agisilaos: If we don't use capitalized, then we will have to modify the Localizable.strings files to make the appropriate changes and I think we already dismissed that idea.

KaterinaWire: Yes, we don't want to change Localizable.strings files.

KaterinaWire: But do we need to change them
"general.off" = "Off";
"general.on" = "On";
?

agisilaos: Then I think it is appropriate to keep it as it is (with the applied changes I made). Also, I checked where we use .capitalized so far and I didn't find any wrong use cases.

KaterinaWire (approved):

PR #58, opened 2023-03-02
feat: Fetch fingerprints FS-1556

KaterinaWire:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Refactor fetch fingerprints methods to use the ProteusService if needed.

### Solutions
`UserClient`
`SelfLegalHoldSubject`

----

----

----
##### References


wire-ios-data-model/Source/Model/User/ZMUser+LegalHoldRequest.swift
```diff
@@ -32,6 +33,9 @@ public protocol SelfLegalHoldSubject {
     /// Whether the user needs to acknowledge the current legal hold status.
     var needsToAcknowledgeLegalHoldStatus: Bool { get }
 
+    /// The user's fingerprint.
+    var fingerprint: String? { get }
```
johnxnguyen: This is the fingerprint of the legal hold device, not the user, right?

wire-ios-data-model/Source/Model/UserClient/UserClient.swift
```diff
@@ -503,37 +500,98 @@ public extension UserClient {
...
         })
     }
+
+}
+
+// MARK: - Fetch fingerprint
+
+extension UserClient {
+
+    private var remoteFingerprint: Data? {
```
johnxnguyen: I think this should be a method rather than a computed property because it involves a bit of work which is not apparent if it's exposed as a property.

wire-ios-data-model/Source/Model/UserClient/UserClient.swift
```diff
@@ -503,37 +500,98 @@ public extension UserClient {
...
+                    fingerprintData = sessionsDirectory.fingerprint(for: sessionIdentifier)
+                    print(fingerprintData)
+                })
+            }
+        )
+
+        return fingerprintData
+    }
+
+    private var localFingerprint: Data? {
```
johnxnguyen: I think this should be a method rather than a computed property because it involves a bit of work which is not apparent if it's exposed as a property.

johnxnguyen (commented):
Should we add a test for fetching the remote fingerprint too?

wire-ios-data-model/Source/Model/User/ZMUser+LegalHoldRequest.swift
```diff
@@ -272,4 +276,42 @@ extension ZMUser: SelfLegalHoldSubject {
...
     }
 
+    // MARK: - Fingerprint
+
+    var proteusProvider: ProteusProviding? {
+        if let context = self.managedObjectContext?.zm_sync {
+            return ProteusProvider(context: context)
+        }
+        return nil
+    }
```
netbe: maybe move this to managedObjectContext extension as we have the exact code on UserClient

johnxnguyen: I'm also questioning whether we should be traversing the to the sync context via `zm_sync`. `ProteusService` should be only on the sync context, and any code that wants to use it should be on the sync context.

With this code, we may be on the view context and inadvertently move to the sync context, which may or may not cause problems.

wire-ios-data-model/Source/Model/User/ZMUser+LegalHoldRequest.swift
KaterinaWire: Yes, fixed

wire-ios-data-model/Source/Model/UserClient/UserClient.swift
KaterinaWire: Done

wire-ios-data-model/Source/Model/UserClient/UserClient.swift
KaterinaWire: Done

wire-ios-data-model/Source/Model/User/ZMUser+LegalHoldRequest.swift
```diff
@@ -272,4 +276,42 @@ extension ZMUser: SelfLegalHoldSubject {
...
+            log.error("Could not fetch fingerprint for \(self)")
+            return nil
+        }
+    }
+
+    private func fetchFingerprint(through keystore: UserClientKeysStore) -> String? {
+        guard let preKey = legalHoldRequest?.lastPrekey,
+              let fingerprintData = EncryptionSessionsDirectory.fingerprint(fromPrekey: preKey.key) else {
+                  return nil
+              }
```
David-Henner:
```suggestion
        guard 
            let preKey = legalHoldRequest?.lastPrekey,
            let fingerprintData = EncryptionSessionsDirectory.fingerprint(fromPrekey: preKey.key) 
        else { return nil }
```

wire-ios-data-model/Source/Model/UserClient/UserClient.swift
```diff
@@ -503,37 +500,97 @@ public extension UserClient {
...
+
+// MARK: - Fetch fingerprint
+
+extension UserClient {
+
+    private func remoteFingerprint() -> Data? {
+        guard let proteusProvider = self.proteusProvider,
+              let sessionIdentifier = self.sessionIdentifier else {
+                  return nil
+              }
```
David-Henner:
```suggestion
        guard let proteusProvider = self.proteusProvider,
              let sessionIdentifier = self.sessionIdentifier 
        else { return nil }
```

David-Henner (approved):
nice work 👍 
I just left some suggestions for `guard` statements indentation, feel free to keep it or leave it

johnxnguyen (dismissed):

KaterinaWire:
@netbe @johnxnguyen I moved `proteusProvider` to the managedObjectContext. Please check it

wire-ios-data-model/Source/ManagedObjectContext/NSManagedObjectContext+ProteusProvider.swift
```diff
@@ -0,0 +1,29 @@
...
+//
+
+import Foundation
+import CoreCryptoSwift
+
+extension NSManagedObjectContext {
+
+    public var proteusProvider: ProteusProviding? {
+        precondition(zm_isSyncContext, "ProteusProvider should only be accessed on the sync context")
+        return ProteusProvider(context: zm_sync)
```
johnxnguyen:
```suggestion
        return ProteusProvider(context: self)
```

wire-ios-data-model/Source/Model/User/ZMUser+LegalHoldRequest.swift
```diff
@@ -272,4 +276,41 @@ extension ZMUser: SelfLegalHoldSubject {
...
     }
 
+    // MARK: - Fingerprint
+
+    public var fingerprint: String? {
+        guard let syncContext = managedObjectContext?.zm_sync,
+              let proteusProvider = syncContext.proteusProvider else {
+                  log.error("Could not access proteusProvider")
+                  return nil
+              }
```
johnxnguyen: Is this safe though? It's a public method so it may be called from the UI. If `managedObjectContext` is the view context, then we must be on the main thread. I don't see anywhere we switch to the sync context. This may only show up as a runtime crash.

You may need to switch to the sync context thread then do the work you need to.

wire-ios-data-model/Source/Model/UserClient/UserClient.swift
```diff
@@ -657,10 +655,11 @@ public extension UserClient {
     }
 
     var proteusProvider: ProteusProviding? {
-        if let context = self.managedObjectContext?.zm_sync {
-            return ProteusProvider(context: context)
+        guard let context = self.managedObjectContext?.zm_sync else {
```
netbe: can't we just remove this since you added the extension on NSManagedObjectContext?

johnxnguyen (approved):

wire-ios-data-model/Source/Model/UserClient/UserClient.swift
KaterinaWire: Yes, we can.
Done

wire-ios-data-model/Source/Model/User/ZMUser+LegalHoldRequest.swift
```diff
@@ -272,4 +276,46 @@ extension ZMUser: SelfLegalHoldSubject {
         needsToAcknowledgeLegalHoldStatus = false
     }
 
+    // MARK: - Fingerprint
+
+    public var fingerprint: String? {
```
netbe, line 281: as for localFingerprint() method should make this a method

PR #72, opened 2023-03-07
feat: migrate to v3 session ids FS-1601

johnxnguyen:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Before migrating proteus session from cryptobox to core crypto, we should migrate the proteus session ids from v2 to v3, since we won't have the opportunity to do so after the move to core crypto.

A proteus session id is simply an identifier used to refer to a proteus session. The session is a file with this id as its name, and is stored in the otr directory. The scope of the id is local to the device, so we can change it freely to anything, so long as we are consistent. In other words, we can rename the proteus session, but then we need to make sure we can derive the new identifier from the associated `UserClient` instance.

There are three versions of proteus session ids:
- v1 uses only the client id, e.g `<clientID>`
- v2 uses also the user id, e.g: '<userID>_<clientID>'
- v3 adds the domain, e.g: '<domain>_<userID>_<clientID>'

Migration from v1 to v2 occurred years ago, so we can be sure that almost all sessions use v2 identifiers. v3 identifiers are only used when a domain is available, which only occurs when federation is enabled for the backend.

Production backends have been proving their domains to the clients for some time now, so even if federation is not enabled, we should still be able to migrate from v2 identifiers to v3 identifiers, using the local domain.

The identifier migration code already exists, but it was never applied. This PR applies it.

#### Test Coverage

- Added unit test for the migration

### Notes

There are a couple things to note here:
- When we are migrating from Cryptobox to Core Crypto, we won't have an instance of the keystore available since we create instead a `ProteusService`. 
- The actual migration of session id is the responsibility of Cryptobox code, for this we need the keystore.
- In case we are migrating to Core Crypto, we create a temporary keystore to perform the migration.

Additionally, this is the last time we are using the legacy persisted data patch system. This system will no longer be viable to use in the monorepo because it relies on framework versions which are no longer a thing in the monorepo. I initially tried to use the newer `PatchApplicator` to do this migration but discovered that it's not quite ready yet. Fortunately, we can still add one last legacy patch because the current data model version is higher than what is in production.

Despite this I did make some minor changes to the patch system which I decided not to remove from this PR.

----

----

----
##### References


wire-ios-data-model/Source/Model/UserClient/UserClient+Patches.swift
```diff
@@ -43,23 +70,63 @@ extension UserClient {
...
+            WireLogger.proteus.info("migrating all session ids to v3: using existing keystore")
+            keyStore = existingKeyStore
+        } else {
+            WireLogger.proteus.info("migrating all session ids to v3: creating temp keystore")
+
+            guard
+                let accountDirectory = moc.accountDirectoryURL,
+                let applicationContainer = moc.applicationContainerURL
+            else {
+                fatalError("Can not migration proteus session ids")
```
johnxnguyen: Is this a suitable way to handle this?

johnxnguyen:
Some questions that I have:

- are we sure that every client out there can rely on getting a local domain?
- what happens if they can't?

wire-ios-data-model/Patches/Legacy/LegacyPersistedDataPatches+Directory.swift
```diff
@@ -42,6 +46,7 @@ extension LegacyPersistedDataPatch {
         LegacyPersistedDataPatch(version: "279.0.4", block: InvalidFeatureRemoval.restoreDefaultConferenceCallingConfig),
         LegacyPersistedDataPatch(version: "285.0.0", block: ZMConversation.introduceAccessRoleV2),
         LegacyPersistedDataPatch(version: "290.0.1", block: ZMUser.refetchSelfUserDomain),
+        LegacyPersistedDataPatch(version: "297.0.2", block: UserClient.migrateAllSessionsClientIdentifiersV3)
```
netbe, line 49: where is this version number coming from? Read from db?

wire-ios-data-model/Source/Model/UserClient/UserClient+Patches.swift
```diff
@@ -17,24 +17,56 @@
...
+        let keyStore: UserClientKeysStore
+
+        if let existingKeyStore = moc.zm_cryptKeyStore {
+            WireLogger.proteus.info("migrating all session ids to v2: using existing keystore")
+            keyStore = existingKeyStore
+        } else {
+            WireLogger.proteus.info("migrating all session ids to v2: creating temp keystore")
+
+            guard
+                let accountDirectory = moc.accountDirectoryURL,
```
netbe: how come the moc has this setup and not a UserClientKeysStore as `setupUserKeyStore` is called just after that

netbe: is this else case just a safe case but can't really happen?

wire-ios-data-model/Patches/Legacy/LegacyPersistedDataPatches+Directory.swift
johnxnguyen: It's stored in the bundle, which is read from the `version.xcconfig` file.

wire-ios-data-model/Source/Model/UserClient/UserClient+Patches.swift
johnxnguyen: > how come the moc has this setup and not a UserClientKeysStore as setupUserKeyStore is called just after that
I don't quite understand the first question.

> is this else case just a safe case but can't really happen?
I understood that when `DeveloperFlag.proteusViaCoreCrypto` is enabled we wouldn't create the keystore, to avoid having both the keystore and the proteus service. I don't think we made that change yet but it is another PR.

So this else case is both being safe, but also anticipating that we won't have the keystore when moving to core crypto.

wire-ios-data-model/Source/Model/UserClient/UserClient+Patches.swift
```diff
@@ -17,24 +17,56 @@
...
+                let applicationContainer = moc.applicationContainerURL
+            else {
+                fatalError("Can not migration proteus session ids")
```
David-Henner:
```suggestion
                fatalError("Can not migrate proteus session ids")
```

wire-ios-data-model/Source/Model/UserClient/UserClient+Patches.swift
```diff
@@ -43,23 +75,62 @@ extension UserClient {
...
+            }
+
+            keyStore = UserClientKeysStore(
+                accountDirectory: accountDirectory,
+                applicationContainer: applicationContainer
+            )
+        }
+
```
David-Henner: This code looks like it duplicates the same code defined around line 50, is this intentional?

johnxnguyen: I will de-duplicate it

johnxnguyen: done

wire-ios-data-model/Source/Model/UserClient/UserClient+Patches.swift
johnxnguyen: fixed

MarcinR (approved):

netbe (approved):

johnxnguyen:
> Some questions that I have:
> 
> * are we sure that every client out there can rely on getting a local domain?
> * what happens if they can't?

Answering my own questions: the backend will tell the client its domain when the client tries to resolve the api version, so it will always be able to get the domain.

PR #80, opened 2023-03-08
chore: Get rid of deprecated system message types: `ZMSystemMessageTypeUsingNewDevice`, `ZMSystemMessageTypeReactivatedDevice`

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we get rid of some deprecated system message types. In addition to that we clean up the tests and delete unneeded reference snapshot images.

----
##### References


netbe:
why these messages are deprecated? Is there a related ticket?

agisilaos:
> why these messages are deprecated? Is there a related ticket?

I don't think that there is a related ticket and I'm not sure why they're deprecated. Maybe @johnxnguyen has more information on that. 

It was on my list to take care of those deprecations for a while now and since we moved to mono repo that became much easier.

David-Henner (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCell.swift
```diff
@@ -948,12 +944,6 @@ final class ConversationNewDeviceSystemMessageCellDescription: ConversationMessa
         return StyleKitIcon.exclamationMark.makeImage(size: 16, color: SemanticColors.LegacyColors.vividRed)
     }
 
-    private static func configureForReactivatedSelfClient(_ selfUser: UserType, link: URL) -> View.Configuration {
-        let string = "content.system.reactivated_device".localized(args: link.absoluteString)
```
johnxnguyen, line 995: We should remove the string too.

wire-ios-sync-engine/Source/Synchronization/ZMHotFixDirectory.m
```diff
@@ -74,11 +74,6 @@ - (NSArray *)patches
                      patchCode:^(NSManagedObjectContext *context) {
                          [ZMHotFixDirectory insertNewConversationSystemMessage:context];
                      }],
-                    [ZMHotFixPatch
-                     patchWithVersion:@"45.1"
-                     patchCode:^(NSManagedObjectContext *context) {
-                         [ZMHotFixDirectory updateSystemMessages:context];
-                     }],
```
johnxnguyen, line 81: We shouldn't be removing hot fixes like this unless we're sure that this version isn't existing in the wild anymore. It's such an old version that it's unlikely to still exist, but it would be good to find out which wire version contains this version.

Also, if we can remove this hot fix, then we can remove all previous hot fixes.

agisilaos: >  but it would be good to find out which wire version contains this version.

How we can find that information? 

> Also, if we can remove this hot fix, then we can remove all previous hot fixes.

After I verify what I mentioned above I will also remove the previous hot fixes

johnxnguyen: - figure out which version of the app contained SE 45.1, check the git tags
- figure out what the earliest non blacklisted client version is, you can check the blacklist for this.

PR #95, opened 2023-03-15
fix: can not send assets from share extension SQSERVICES-1788

johnxnguyen:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Sometimes it is not possible to send assets from the share extension, as it gets stuck in the sending phase and then times out. It seems kind of random, but there may be some patterns to reliably reproduce it, such as sending a photo with a text, then a file, then a photo again.

### Causes

There are two issues. First, sending a photo with a text image was always failing due to changes in Core Data and iOS 16. This is the same problem that was fixed in https://github.com/wireapp/wire-ios-sync-engine/pull/1782, which also has the same fix here.

The second issue is that we seem to get stuck at the point where we want to upload the asset. We observed that we make the request to upload the asset, but we don't get a response, halting the flow until the extension times out. In actuality, we do get the response almost immediately (as observed with using network tools like Proxyman), and this response does sometimes get delivered to the url session delegate, albeit about 1 minute late. Why this is? We don't know. As far as we can tell, this issue is affecting only the share extension (and not the main app), and possibly from iOS 15.7 (earliest observed version with the problem). My suspicion is that it may be due to background tasks related to using the background url session. However I can not tell at all whether this is a fault on our side or Apple's.

### Solutions

This issue has drained our resources for several weeks so here I present a workaround solution instead of (the ultimately satisfying) explanation of what is actually going wrong (bummer!).

The issue is likely related to using the background url session, because if we switch to using the standard foreground url session then the issue no longer appears. So, I've added a flag to conditionally use the foreground session for asset uploads coming from the share extension. Note that the share extension doesn't dismiss until all assets are sent, so in theory we shouldn't need a background session anyway. The only possible side effect could be a timeout due to a extremely slow or long upload.

#### How to Test
Share all types of assets using the share extension, combining them with accompanying text messages or not.

----

----

----
##### References


agisilaos (approved):

netbe (approved):
let's merge it and see on thursday playtest, will have to backport this frida

wire-ios-share-engine/Sources/OperationLoop.swift
```diff
@@ -158,9 +158,9 @@ final class OperationLoop: NSObject, RequestAvailableObserver {
 
     func setupObserver(for context: NSManagedObjectContext, onSave: @escaping SaveClosure) -> NSObjectProtocol {
         return NotificationCenter.default.addObserver(forName: .NSManagedObjectContextDidSave, object: context, queue: callBackQueue) { note in
-            if let insertedObjects = note.userInfo?[NSInsertedObjectsKey] as? Set<NSManagedObject>, let updatedObjects = note.userInfo?[NSUpdatedObjectsKey] as? Set<NSManagedObject> {
-                onSave(note, insertedObjects, updatedObjects)
-            }
+          let insertedObjects = (note.userInfo?[NSInsertedObjectsKey] as? Set<NSManagedObject>) ?? Set<NSManagedObject>()
```
lewandowskit93, line 161: I've noticed that we have exactly the same case for NotificationEngine's OperationLoop that may cause similar issues in the future for notifications.
What do you think about applying this change there as well?

agisilaos: We already discussed that with @johnxnguyen! I'm planning to open a PR with those changes as well. @lewandowskit93

lewandowskit93 (approved):

PR #100, opened 2023-03-15
fix: core crypto key errors in NSE FS-1613

David-Henner:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Decrypting events in the NSE while the phone is locked with a passcode is causing a crash due to the proteus service being nil.

### Causes 

The reason the proteus service is nil is because we're not able to fetch the core crypto key from the keychain in order to create an instance of core crypto.

Fetching the core crypto key doesn't work because when the key is created we do not specify a value for the `kSecAttrAccessible` attribute. Since none is specified, it defaults to `kSecAttrAccessibleWhenUnlocked`. That means once we try to fetch the key from the keychain while the device is locked, we will fail, since it's only accessible when the device is unlocked.

### Solutions

Set `kSecAttrAccessibleAfterFirstUnlock` as the value for the `kSecAttrAccessible` attribute when creating / fetching the key.

**Note:** keychain items are not cleared on app uninstallation. So we need to add some logic to delete the existing keychain item and recreate a new one with the correct accessibility attribute. This only affects us internally, we would remove this migration logic before releasing.

----

----

 - [x] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References

### DIFF
--- a/wire-ios-data-model/Source/MLS/CoreCryptoKeyProvider.swift
+++ b/wire-ios-data-model/Source/MLS/CoreCryptoKeyProvider.swift
@@ -26,7 +26,13 @@ public class CoreCryptoKeyProvider {
     }
 
     public func coreCryptoKey() throws -> Data {
+        removeLegacyKeyIfNeeded()
+        return try fetchOrCreateCoreCryptoKey()
+    }
+
+    private func fetchOrCreateCoreCryptoKey() throws -> Data {
         let item = CoreCryptoKeychainItem()
+
         if let key: Data = try? KeychainManager.fetchItem(item) {
             WireLogger.coreCrypto.info("Core crypto key exists: \(key.base64String()). Returning...")
             return key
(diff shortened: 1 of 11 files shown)


wire-ios-data-model/Source/MLS/CoreCryptoKeyProvider.swift
```diff
@@ -26,7 +26,13 @@ public class CoreCryptoKeyProvider {
     }
 
     public func coreCryptoKey() throws -> Data {
+        removeLegacyKeyIfNeeded()
+        return try fetchOrCreateCoreCryptoKey()
+    }
+
+    private func fetchOrCreateCoreCryptoKey() throws -> Data {
```
johnxnguyen: Perhaps this should be:

```swift
public func coreCryptoKey(createIfNeeded: Bool = false) throws -> Data
```

so we can make sure we only try to create it from the main app and not the extensions.

wire-ios-data-model/Source/MLS/CoreCryptoKeyProvider.swift
```diff
@@ -39,25 +45,66 @@ public class CoreCryptoKeyProvider {
...
+            WireLogger.coreCrypto.error("Failed to delete legacy core crypto key: \(String(describing: error))")
+        } catch {
+            // key was not found. no action needed
+        }
+    }
+}
+
+class CoreCryptoKeychainItem: KeychainItemProtocol {
+
+    static let tag = "com.wire.mls.key".data(using: .utf8)!
```
johnxnguyen: So we don't have to change the tag?

David-Henner: It's not necessary because the legacy vs the new key can be differentiated with the `kSecAttrAccessible` attribute

netbe:
@David-Henner `kSecAttrAccessibleAfterFirstUnlock` what happens if you never unlocked?
1. so install the app
2. lock phone
3. receive a notification

David-Henner:
> @David-Henner `kSecAttrAccessibleAfterFirstUnlock` what happens if you never unlocked?
> 
> 1. so install the app
> 2. lock phone
> 3. receive a notification

this would work fine as "first unlock" refers to the first unlock after the device has been restarted 

> The data in the keychain item cannot be accessed after a restart until the device has been unlocked once by the user.

netbe (approved):

wire-ios-data-model/Source/MLS/CoreCryptoKeyProvider.swift
David-Henner: addressed in [bbac05a](https://github.com/wireapp/wire-ios-mono/pull/100/commits/bbac05a0e4022ba7d4bc7bdc80020925b1bb842c)

johnxnguyen (approved):

PR #106, opened 2023-03-17
fix: Share Extension - Show an alert when an attachment exceeds the size limit of 25MB - SQSERVICES - 1788

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we fix an issue where we would try and send an attachment over 25 MB but it would eventually fail since we have a size limit. So instead of trying to send it, we show an alert that we can't send an attachment over 25 MB. 

In addition to that we dismiss the Share Extension when user taps OK on the alert. That was also something that it was missing and fixed with this PR.

----
##### References

### DIFF
--- a/wire-ios/Wire-iOS Share Extension/UnsentSendable.swift
+++ b/wire-ios/Wire-iOS Share Extension/UnsentSendable.swift
@@ -234,10 +234,6 @@ class UnsentFileSendable: UnsentSendableBase, UnsentSendable {
                     return completion()
                 }
 
-                if url?.fileSize > AccountManager.fileSizeLimitInBytes {
-                    weakSelf.error = .fileSizeTooBig
-                    return completion()
-                }
                 weakSelf.prepareAsFileData(name: url?.lastPathComponent, completion: completion)
             }
         } else if typePass {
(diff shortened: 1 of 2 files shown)


netbe (approved):

wire-ios/Wire-iOS Share Extension/UnsentSendable.swift
```diff
@@ -291,11 +287,22 @@ class UnsentFileSendable: UnsentSendableBase, UnsentSendable {
...
+                    return completion()
+                }
+
                 self?.prepareForSending(withUTI: UTIString,
                                         name: name,
                                         data: data,
                                         completion: prepareColsure)
             } else if let dataURL = data as? URL {
+
+                guard dataURL.fileSize ?? .max <= AccountManager.fileSizeLimitInBytes else {
```
KaterinaWire, line 301: @johnxnguyen Could we do like 
```
guard let fileSize = dataURL.fileSize, 
           fileSize  <= AccountManager.fileSizeLimitInBytes else {
                     self?.error = .fileSizeTooBig
                     return completion()
}
```
?
If `fileSize` is nil we send the `.fileSizeTooBig` error.

KaterinaWire (approved):

netbe:
@agisilaos Please target release/3.109 or reapply to release/3.109 for the RC

wire-ios/Wire-iOS Share Extension/UnsentSendable.swift
johnxnguyen: If we don't know the file size then we can't assume that it's smaller than the limit, so I think it's fine to show the error.

PR #109, opened 2023-03-20
fix: crash when session expires FS-1666

johnxnguyen:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

If the session expires (e.g the self client is deleted remotely), then the app will crash.

### Causes

When the self client is deleted, we try to delete the keystore, even if proteus via core crypto is enabled. The keystore doesn't exist, resulting in a crash.

### Solutions

Use `ProteusProvider` to access the keystore, and also tear down the crypto stack.

----

----

----
##### References

### DIFF
--- a/wire-ios-sync-engine/Source/UserSession/NSManagedObject+EncryptionContext.swift
+++ b/wire-ios-sync-engine/Source/UserSession/NSManagedObject+EncryptionContext.swift
@@ -21,10 +21,12 @@ import WireDataModel
 
 extension NSManagedObjectContext {
 
-    // TODO: [John] use flag here
-
     @objc
     public func deleteAndCreateNewEncryptionContext() {
-        zm_cryptKeyStore.deleteAndCreateNewBox()
+        proteusProvider.perform(
+            withProteusService: { _ in },
+            withKeyStore: { keyStore in keyStore.deleteAndCreateNewBox() }
+        )
     }
+
 }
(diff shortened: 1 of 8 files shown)


wire-ios-sync-engine/Source/UserSession/NSManagedObject+EncryptionContext.swift
```diff
@@ -21,10 +21,12 @@ import WireDataModel
...
 
-    // TODO: [John] use flag here
-
     @objc
     public func deleteAndCreateNewEncryptionContext() {
-        zm_cryptKeyStore.deleteAndCreateNewBox()
+        proteusProvider.perform(
+            withProteusService: { _ in },
+            withKeyStore: { keyStore in keyStore.deleteAndCreateNewBox() }
+        )
```
johnxnguyen, line 29: @typfel there's no core crypto api to clear the proteus data, should there be?

typfel: There's no such API today, so you should would need delete underlaying sqlite store.

typfel: Would it be useful to have such an API?

johnxnguyen: I guess it's just to do some clean up, wouldn't it make sense from a security perspective to keep doing this?

typfel: Yes, I think it makes sense to delete a defunct keystore, which anyway needs to do done if the user decide to login again.

But why do you need a CC API for this? You decide where CC stores the sqlite database. If it's stored in its own directory you can delete the whole directory and any auxiliary files will also be deleted.

johnxnguyen: You're right, I was getting confused and thought that I needed to delete only the proteus keystore but keep the mls keystore intact, but we actually want to delete all of it.

David-Henner (approved):

agisilaos (approved):

PR #110, opened 2023-03-20
feat: security banner redesign SQSERVICES-1933

lewandowskit93:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Security level banner as of now is not sufficiently discernible. Because of that the banner may be easily overlooked.

### Solutions

Applied design changes to the view that improves contrasts for all types of banners in both light and dark modes.
There were some missing and/or unfitting colours that needed to be added.
The new exclamation mark has changed, but applied only to the banner, so it has no effect on other usages.
Most of the current icons are code-generated by external tool that we do not use. This method also ignores light/dark colours variations which caused the checkmark to be rendered incorrectly in tests for dark mode.
So the icons added are exclamation mark & checkmark in the format of SVG file that we apply tint colour onto.
The logic for deprecated & new calling UI got quite conditional, so I've split the implementation into separate methods to improve readability.

#### Test Coverage

- [x] Added snapshot tests of classification banner for light mode
- [x] Added snapshot tests of classification banner for dark mode 

#### How to Test

Since it was a redesign of elements and no logic change, I've forced the banner to render by applying desired classification mode in `configure` method of `SecurityLevelView`.
On the built app it could be tested with classified & unclassified domains and federated/non-federated backends.

----

----

----
##### References


wire-ios/Wire-iOS Tests/ReferenceImages/SecurityLevelViewTests/testThatItRendersWithClassified.DarkTheme.png
lewandowskit93, line 1: The checkmark should be of the same color as text, however it's not.
The colors are set the same for both text and icon: .green500Light; .green500Dark

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Views/SecurityLevelView.swift
```diff
@@ -122,9 +109,52 @@ final class SecurityLevelView: UIView {
...
+
+        case .notClassified:
+            securityLevelLabel.textColor = SemanticColors.Label.textDefaultWhite
+            backgroundColor = SemanticColors.View.backgroundSecurityDisabled
+            iconImageView.image = Asset.Images.attention.image.withTintColor(SemanticColors.Icon.foregroundCheckMarkSelected)
+            topBorder.backgroundColor = .clear
+
+        default:
+            isHidden = true
+            assertionFailure("should not reach this point")
```
johnxnguyen, line 137: What is "this point" exactly? That we have no classification but we expect there to be one?

lewandowskit93: Yes. It's actually moved from the old implementation.
It should never arrive at this place as the guard in the method that calls these ones should filter-out `.none` classification prior to calling them.

wire-ios/Wire-iOS Tests/ReferenceImages/SecurityLevelViewTests/testThatItRendersWithClassified.DarkTheme.png
lewandowskit93: I've found that rendering an icon always takes color from light theme in tests regardless that I've set dark theme to the view (I've changed colors  to .green500Dark; .green500Dark and checkmark has correct color on dark mode).
Its caused probably by of the fact that the dark mode is set on a single view that I'm testing, not in a whole app.
The icon-setting method is auto-generated in the root so we cannot modify painting there.

lewandowskit93: I'm setting svg checkmark file with tint as an icon instead of the paint methods, as I've done with exclamation mark.
This way it works correctly :)

wire-ios/Wire-iOS Tests/SecurityLevel/SecurityLevelViewTests.swift
```diff
@@ -0,0 +1,61 @@
...
+
+    private var sut: SecurityLevelView!
+
+    override func setUp() {
+        super.setUp()
+        callingUIFlag = DeveloperFlag.deprecatedCallingUI
+        deprecatedUIFlagStateBackup = callingUIFlag.isOn
+        callingUIFlag.isOn = false
+        sut = SecurityLevelView()
+        sut.backgroundColor = .white
```
agisilaos: It would be good to set the value for background color the same as we do on the main app. `SemanticColors.View.xxx` and then for dark mode specifically we can set the `overrideInterfaceStyle` to .dark within the test.

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Views/SecurityLevelView.swift
johnxnguyen: Perhaps we can avoid this by defining two methods `updateForClassified`, `updateForNotClassified`, then use a switch case earlier.

johnxnguyen (approved):

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Views/SecurityLevelView.swift
```diff
@@ -122,9 +109,52 @@ final class SecurityLevelView: UIView {
...
+        }
+    }
+
+    private func configureLegacyCallingUI(with classification: SecurityClassification) {
+        switch classification {
+
+        case .classified:
+            securityLevelLabel.textColor = SemanticColors.Label.textDefault
+            backgroundColor = SemanticColors.View.backgroundSecurityLevel
+            iconImageView.setIcon(.checkmark, size: .tiny, color: SemanticColors.Icon.backgroundCheckMarkSelected)
```
agisilaos: I don't think that would work when switching from light to dark mode. Instead of `setIcon`. We need to use `setTemplateIcon`

```suggestion
            iconImageView.setTemplateIcon(.checkmark, size: .tiny)
            iconImageView.tintColor = SemanticColors.Icon.backgroundCheckMarkSelected
```

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Views/SecurityLevelView.swift
```diff
@@ -122,9 +109,52 @@ final class SecurityLevelView: UIView {
...
+            topBorder.backgroundColor = SemanticColors.View.backgroundSeparatorCell
+
+        case .notClassified:
+            securityLevelLabel.textColor = SemanticColors.Label.textDefault
+            backgroundColor = SemanticColors.View.backgroundSecurityLevel
+            iconImageView.setIcon(.exclamationMarkCircle, size: .tiny, color: SemanticColors.Icon.foregroundCheckMarkSelected)
```
agisilaos: I don't think that would work when switching from light to dark mode. Instead of `setIcon`. We need to use `setTemplateIcon`

```suggestion
           iconImageView.setTemplateIcon(.exclamationMarkCircle, size: .tiny)
            iconImageView.tintColor = SemanticColors.Icon.foregroundCheckMarkSelected
```

agisilaos (changes requested):
Looking good. I left a few suggestions that need to be implemented before merging.

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Views/SecurityLevelView.swift
lewandowskit93: Then we would need to split with legacy/updated style again in these 2/3 functions. Or have 4/6 funcs for that.
I guess checking it with single if inside each of the 2/3 functions is not that big deal, but these amounts of `?:` operators got quite messy for sure :D

lewandowskit93: But you mean removing assertion only? Because the case of the switch would have to stay there anyway.
I would think about replacing default with `.none` though in case of modifying the enum.

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Views/SecurityLevelView.swift
lewandowskit93: Should we be changing this though? It's a legacy/deprecated UI. So my thinking was that I did not want to change logic on that legacy UI part.

wire-ios/Wire-iOS Tests/SecurityLevel/SecurityLevelViewTests.swift
lewandowskit93: Changing this one 👍 
In case of dark & light modes it's no need to change it directly on `overrideInterfaceStyle`, as it all logic for that happens in `verifyInAllColorSchemes`.

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Views/SecurityLevelView.swift
agisilaos: We should change it before actually, the previous implementation was not right. Somehow it slipped the cracks. That's the correct approach.

lewandowskit93: 👍

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Views/SecurityLevelView.swift
lewandowskit93: Modified to `case .none`. It should alert as about this place when change is needed.

agisilaos (approved):

PR #123, opened 2023-03-23
fix: Update copy for confirmation dialog when initiating a large call - SQSERVICES - 1955

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We have two issues. The first one is the copy of the alert. And the second issue is that we don't show that alert when trying to initiate a large video call. We only show the alert when trying to initiate an audio call.

### Solutions

With this PR we fix the two issues I mentioned above. In addition to that we make the code in `UIAlertController+OngoingCall.swift` more readable as well as replacing `.localized` with `L10n`.

----
##### References


wire-ios/Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -290,8 +290,7 @@
 "conversation.input_bar.audio_message.keyboard.record_tip" = "Tap to record\nYou can  %@  it after that";
 "conversation.input_bar.audio_message.keyboard.filter_tip" = "Choose a filter above";
 
-"conversation.call.many_participants_confirmation.title" = "Start a call";
-"conversation.call.many_participants_confirmation.message" = "This will call %d people";
+"conversation.call.many_participants_confirmation.title" = "Are you sure you want to call %d people?";
```
johnxnguyen, line 293: Just double checking that this copy came from the product / copy team and not from developers?

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationCallController.swift
```diff
@@ -52,10 +52,24 @@ final class ConversationCallController: NSObject {
...
+    func startVideoCall(started: Completion?) {
+        let startVideoCall = { [weak self] in
+            guard let `self` = self else { return }
+            self.conversation.confirmJoiningCallIfNeeded(alertPresenter: self.target) {
+                started?()
+                self.conversation.startVideoCall()
+            }
+        }
+
+        if conversation.localParticipants.count <= 4 {
```
netbe: use a constant, does this number come from some config too?

wire-ios/Wire-iOS/Resources/Base.lproj/Localizable.strings
agisilaos: Yep, that came directly from Astrid. She left a comment on the ticket.

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationCallController.swift
agisilaos: That number came from the product team as well. What would be a good name for that constant?

netbe: what about ConfirmGroupCallParticipantsLimit ?

wire-ios/Wire-iOS/Sources/UserInterface/Calling/Helper/UIAlertController+OngoingCall.swift
```diff
@@ -57,17 +66,17 @@ extension UIAlertController {
...
 
         let controller = UIAlertController(
-            title: effectiveStyle == .alert ? titleKey.localized : messageKey.localized,
-            message: effectiveStyle == .alert ? messageKey.localized : nil,
+            title: effectiveStyle == .alert ? titleKey : messageKey,
+            message: effectiveStyle == .alert ? messageKey : nil,
             preferredStyle: effectiveStyle
         )
-        controller.addAction(.init(title: buttonTitleKey.localized, style: .default) { _ in completion(true) })
+        controller.addAction(.init(title: buttonTitleKey, style: .default) { _ in completion(true) })
```
johnxnguyen: `titleKey`, `messageKey`, and `buttonTitleKey` aren't keys anymore are they? If not, then I suggest to update the parameter names to reflect that.

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationCallController.swift
```diff
@@ -52,10 +52,24 @@ final class ConversationCallController: NSObject {
...
     }
 
-    func startVideoCall(started: (() -> Void)?) {
-        conversation.confirmJoiningCallIfNeeded(alertPresenter: target) { [conversation] in
-            started?()
-            conversation.startVideoCall()
+    func startVideoCall(started: Completion?) {
+        let startVideoCall = { [weak self] in
+            guard let `self` = self else { return }
+            self.conversation.confirmJoiningCallIfNeeded(alertPresenter: self.target) {
```
johnxnguyen, line 59: What does `confirmJioningCallIfNeeded` do?

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationCallController.swift
agisilaos: Sounds good. I'm going to use that name for the property.

agisilaos: Done!

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationCallController.swift
agisilaos: Here's what it does:

```swift
func confirmJoiningCallIfNeeded(alertPresenter: UIViewController, forceAlertModal: Bool = false, completion: @escaping () -> Void) {
        guard ZMUserSession.shared()?.isCallOngoing == true else {
            return completion()
        }

        let controller = UIAlertController.ongoingCallJoinCallConfirmation(forceAlertModal: forceAlertModal) { confirmed in
            guard confirmed else { return }
            self.endAllCallsExceptIncoming(completion: completion)
        }

        alertPresenter.present(controller, animated: true)
    }
```

wire-ios/Wire-iOS/Sources/UserInterface/Calling/Helper/UIAlertController+OngoingCall.swift
agisilaos: Done!

David-Henner (commented):
should there be new snapshots?

agisilaos:
> should there be new snapshots?

@David-Henner I don't think we have snapshots with that alert. 🤔

netbe (approved):

johnxnguyen (approved):

PR #135, opened 2023-03-28
fix: setupCoreCrypto checks proper developer flag on each stage

netbe:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When setting up CoreCrypto, the flags are generally check but not individually. This results in times where mls and proteus stack are setup but their respective flag is not. 

### Solutions

Add individual check on stage setup

#### Test Coverage (Optional)

- [ ] I have added automated test to this contribution

#### How to Test

_Briefly describe how this change was tested and if applicable the exact steps taken to verify that it works as expected._

### Notes (Optional)

_Specify here any other facts that you think are important for this issue._

### Attachments (Optional)

_Attachments like images, videos, etc. (drag and drop in the text box)_

----

----

----
##### References


wire-ios-sync-engine/Source/UserSession/ZMUserSession+CoreCrypto.swift
```diff
@@ -35,10 +35,12 @@ extension ZMUserSession {
         }
 
         switch stage {
-        case .proteus(let userID):
+        case .proteus(let userID) where shouldSetupProteusService:
```
johnxnguyen: I think we should rename these to `shouldSetUpProteus` and `shouldSetUpMLS` since these flags were previous used to conditionally create the controllers, but now they are more general.

johnxnguyen (commented):
Perhaps also a unit test?

David-Henner (commented):
I just realised that this might not work in the case where proteus via core crypto is disabled but MLS is enabled. Because we'd never go to `setupProteus` which creates the instance of core crypto.

David-Henner:
> Perhaps also a unit test?

The current unit tests already cover quite well the different combinations of having proteus and mls flags on/off.
The only test we could add is one that specifically checks that core crypto won't be instantiated by the `setupProteus` method if proteus is disabled. But I'd argue that's not necessary since we already verify that the outcome of each combination is correct.

David-Henner (approved):

johnxnguyen (approved):

PR #145, opened 2023-03-30
fix: migration cryptobox data to core crypto FS-894

johnxnguyen:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

After migrating crypt box data to core crypto, there are a lot of decryption errors for any new messages received.

### Causes

Although the migration code didn't fail, core crypto didn't do any migration because `proteusInit` was already called (when setting up the core crypto stack. As a result, there were no sessions in core crypto, so new ones were created, and this was causing proteus errors.

### Solutions

- Call `proteusInit` only after a completing the cryptobox migration, or call it after we know we don't need to make the migration.
- Don't create `NotificationSession` or `SharingSession` if a migration is pending (only the main app should do this)
- To achieve this, more the `CryptoboxMigrationManager` to data model and reuse it.

#### Test Coverage

- Unit test for new `completeInitialization` in `ProteusService
- Unit test for `CryptoboxMigrationManager`
- Unit test to assert `NotificationSession` doesn't init when migration is pending
- Unit test to assert `SharingSession` doesn't init when migration is pending

----

----

----
##### References


wire-ios-notification-engine/WireNotificationEngineTests/MockCoreCrypto.swift
```diff
@@ -49,6 +49,11 @@ class MockSafeCoreCrypto: SafeCoreCryptoProtocol {
 
         try mock(clientID)
     }
+
+    var tearDownCount = 0
+    func tearDown() throws {
+        tearDownCount += 1
+    }
```
johnxnguyen, line 56: This is to fix a conformance error.

wire-ios-share-engine/WireShareEngineTests/SharingSessionTests+CryptoStack.swift
```diff
@@ -392,8 +392,12 @@ class MockCoreCrypto: CoreCryptoProtocol {
...
     }
 
-    func proteusNewPrekeyAuto() throws -> [UInt8] {
-        return []
+    func markConversationAsChildOf(childId: ConversationId, parentId: ConversationId) throws {
+
+    }
+
+    func proteusNewPrekeyAuto() throws -> ProteusAutoPrekeyBundle {
+        fatalError("not implemented")
```
johnxnguyen, line 400: This is to fix conformance errors

wire-ios-data-model/Source/Proteus/CryptoboxMigrationManager.swift
```diff
@@ -0,0 +1,135 @@
...
+                WireLogger.proteus.info("migrating cryptobox data...")
+                let cryptoboxDirectory = fileManager.cryptoboxDirectory(in: accountDirectory)
+                try proteusService.migrateCryptoboxSessions(at: cryptoboxDirectory)
+                WireLogger.proteus.info("migrating cryptobox data... success")
+            } catch {
+                throw Failure.failedToMigrateData
+            }
+
+            do {
+                WireLogger.proteus.info("removing legacy cryptobox data...")
```
netbe, line 89: is it safe to remove the old data before the proteusInit()? I mean if proteusInit() fails do we restart the migration?

wire-ios-data-model/Tests/Proteus/CryptoboxMigrationManagerTests.swift
```diff
@@ -0,0 +1,179 @@
...
+        // When
+        try sut.completeMigration(syncContext: syncMOC)
+
+        // Then
+        XCTAssertEqual(mockProteusService.completeInitialization_Invocations.count, 0)
+    }
+
+}
+
+class FileManagerMock: FileManagerInterface {
```
netbe: make private or move to own file

+ is there a reason for not using sourcery?

wire-ios-notification-engine/WireNotificationEngineTests/BaseNotificationSessionTests.swift
```diff
@@ -148,7 +152,25 @@ class BaseTest: ZMTBaseTest {
...
             operationLoop: operationLoop,
             accountIdentifier: accountIdentifier,
-            pushNotificationStrategy: pushNotificationStrategy
+            pushNotificationStrategy: pushNotificationStrategy,
+            cryptoboxMigrationManager: mockCryptoboxMigrationManager
         )
     }
 }
+
+class MockCryptoboxMigrationManager: CryptoboxMigrationManagerInterface {
```
netbe: move to own file?

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -785,63 +785,103 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+        completion: @escaping () -> Void
+    ) {
+        guard cryptoboxMigrationManager.isMigrationNeeded(accountDirectory: accountDirectory) else {
+            WireLogger.proteus.info("cryptobox migration is not needed")
+
+            syncContext.performAndWait {
+                do {
+                    try cryptoboxMigrationManager.completeMigration(syncContext: syncContext)
+                } catch {
+                    fatalError("failed to complete proteus initialization")
```
netbe, line 856: so if it fails, we don't have a way to restore the previous state, right?
it will keep on crashing, do we have a recovery plan in this case? What do the other platform do?

netbe, line 856: add log of the error

wire-ios-data-model/Source/Proteus/CryptoboxMigrationManager.swift
johnxnguyen: I don't think proteus init is part of the migration, so in theory it's safe to delete old data if the migration call doesn't throw.

wire-ios-data-model/Tests/Proteus/CryptoboxMigrationManagerTests.swift
johnxnguyen: I think I wasn't initially in data model so did it manually, but I could use it now

wire-ios-share-engine/WireShareEngineTests/SharingSessionTests.swift
```diff
@@ -64,4 +65,22 @@ class SharingSessionTests: BaseSharingSessionTests {
...
+            // Given
+            mockCryptoboxMigrationManager.isMigrationNeeded = true
+
+            // When
+            _ = try createSharingSession()
+        } catch SharingSession.InitializationError.pendingCryptoboxMigration {
+            // Then
+            return
+        } catch {
+            XCTFail("unexpected error: \(error.localizedDescription)")
```
David-Henner, line 82: this only fails if another error is thrown, should we also fail if no error is being thrown?

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
David-Henner: I suppose this would fail if we fail to create the proteus service, which I would consider a developer error. An acceptable way to handle it could be to log the error and find what in the proteus service setup code can go wrong

David-Henner (approved):

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnxnguyen: I think Android crashes if any part of the migration isn't successful.

wire-ios-share-engine/WireShareEngineTests/SharingSessionTests.swift
johnxnguyen: done

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnxnguyen: done

wire-ios-notification-engine/WireNotificationEngineTests/BaseNotificationSessionTests.swift
johnxnguyen: auto mocked this and moved to a new file

wire-ios-data-model/Tests/Proteus/CryptoboxMigrationManagerTests.swift
johnxnguyen: auto mocked this and moved to a new file

netbe (approved):

PR #154, opened 2023-03-31
fix: conversation type for connection requests FS-1764

David-Henner:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Opening an outgoing federated connection request opens the conversation view, although the remote user didn't accept the connection request yet.

### Causes

When fetching the conversation linked to the connection request, the backend returns an erroneous conversation type. The `type` is `oneOnOne` instead of `connection`.

We see the "cancel request" button and archive icon at the bottom because its visibility depends on the status of the connection request, while the conversation input view and call buttons visibility depend on the conversation type.

### Solutions

When processing the backend response, verify the connection status of the conversation and change the conversation type to `connection` is the connection status is `sent`.

Note that this is a workaround and that the real solution is for the backend to return the right conversation type.

### DIFF
--- a/wire-ios-request-strategy/Sources/Payloads/Processing/PayloadProcessing+Conversation.swift
+++ b/wire-ios-request-strategy/Sources/Payloads/Processing/PayloadProcessing+Conversation.swift
@@ -86,21 +86,29 @@ extension Payload.Conversation {
         }
     }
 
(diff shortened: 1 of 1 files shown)


johnxnguyen (approved):

netbe:
@David-Henner is this how Android and other clients handle it? Backend is not going to fix it now ?

netbe (approved):

PR #155, opened 2023-04-03
fix: Transparent View Controller Navigation Bars - IC - 58

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we fix a couple of navigation bars being transparent in certain scenarios where use would tap on `started using a new device` as shown in the screenshot below. 

testy

Screenshot 2023-04-03 at 11 19 08

----
##### References

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
@@ -347,7 +347,7 @@ final class ZClientViewController: UIViewController {
     /// - Parameter conversation: conversation to open
     func openDetailScreen(for conversation: ZMConversation) {
         let controller = GroupDetailsViewController(conversation: conversation)
-        let navController = controller.wrapInNavigationController()
+        let navController = controller.wrapInNavigationController(setBackgroundColor: true)
         navController.modalPresentationStyle = .formSheet
 
         present(navController, animated: true)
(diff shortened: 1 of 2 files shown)


KaterinaWire (approved):

wire-ios/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
```diff
@@ -347,7 +347,7 @@ final class ZClientViewController: UIViewController {
     /// - Parameter conversation: conversation to open
     func openDetailScreen(for conversation: ZMConversation) {
         let controller = GroupDetailsViewController(conversation: conversation)
-        let navController = controller.wrapInNavigationController()
+        let navController = controller.wrapInNavigationController(setBackgroundColor: true)
```
johnxnguyen, line 350: Is there any case that we would set background color false?

agisilaos: @johnxnguyen Good question. I was thinking about that while I was making those changes. I couldn't think of any. 

How do you think we should proceed?

johnxnguyen: Maybe it solves this transparent issue one and for all, but maybe it introduces some other bugs. If I remember correctly, it had something to do with iOS versions. It doesn't hurt to try and run the app on different iOS versions to see if everything is fine.

agisilaos: I can definitely do that. Can we merge this PR? And I will create a ticket for further investigating this.

MarcinR (approved):

wire-ios/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
agisilaos: @johnxnguyen I will create a ticket and further investigate!

PR #163, opened 2023-04-04
fix: Wipe unencrypted user data in the tmp directory- WPB-1151

johnranj:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-1151" title="WPB-1151" target="_blank">TaskWPB-1151</a>  Unencrypted Data Persists in Supposedly Temporary Directory
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
 
Unencrypted data, that is stored in a temporary directory  is
not deleted after logout.

### Causes (Optional)

An attacker with access to the file system could extract sensitive information.
The app container of the Wire app holds a temporary directory “tmp” where files
being used by the app are stored.

### Solutions

Temporary directory is cleared after logout action is performed.
----

----

----
##### References


netbe:
@johnranj can you add tests please?

johnranj:
> @johnranj can you add tests please?

tests added

wire-ios-system/Source/ZMSLog.swift
```diff
@@ -278,6 +280,20 @@ extension ZMSLog {
...
     }
 
+
+    //clears tmp directory
+    @objc public static func clearTmpDirectory() {
+        guard let tmpDirectoryPath = tmpDirectory else { return }
+        closeHandle()
+        let manager = FileManager.default
+        try? manager
+            .contentsOfDirectory(at: tmpDirectoryPath, includingPropertiesForKeys: nil, options: .skipsSubdirectoryDescendants)
```
johnxnguyen: Why do we need the option `skipsSubdirectoryDescendants` ?

wire-ios-system/Source/ZMSLog.swift
```diff
@@ -293,6 +309,12 @@ extension ZMSLog {
         return manager.urls(for: .cachesDirectory, in: .userDomainMask).first
     }
 
+
+    static var tmpDirectory: URL? {
+        let tempUrl = URL(string: NSTemporaryDirectory())
+        return tempUrl
+    }
```
johnxnguyen: Where was the temp directory previous written to? It seems to be a new addition to this file so I'm a little confused how logs are written to the temp directory.

wire-ios-system/Source/ZMSLog.swift
johnranj: skipsSubdirectoryDescendants reduces iteration into subdirectories. As it is for deleting only top-level directories are enough to fetch.

wire-ios-system/Source/ZMSLog.swift
johnranj: yes confused with log cache and temp cache. Moved from zmlogs

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -730,6 +730,15 @@ public final class SessionManager: NSObject, SessionManagerType {
             if deleteAccount {
                 self?.deleteAccountData(for: account)
             }
+
+            // Clear tmp directory when the user logout from the session.
+            guard let tmpDirectoryPath = URL(string: NSTemporaryDirectory()) else { return }
```
johnxnguyen: The temp directory is not scoped to the current user right? It's just a general temp directory for the whole application?

johnranj: yes it is the temp directory for entire application.

wire-ios-sync-engine/Tests/Source/SessionManager/SessionManagerTests.swift
```diff
@@ -396,6 +396,27 @@ final class SessionManagerTests: IntegrationTest {
...
+    func tmpDirectoryPath() -> URL {
+        return URL(fileURLWithPath: NSTemporaryDirectory())
+    }
+
+    func testThatItDestroyedTmpDirectoryAfterLoggedOut() {
+
+        // GIVEN
+        XCTAssertTrue(login())
+        let observer = SessionManagerObserverMock()
+        let token = sessionManager?.addSessionManagerDestroyedSessionObserver(observer)
```
johnxnguyen, line 412: I suggest to add something to the temp directory and assert it's not empty, then we can be sure that after logging out the directory was emptied.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -730,6 +730,15 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+            let manager = FileManager.default
+            try? manager
+                .contentsOfDirectory(at: tmpDirectoryPath, includingPropertiesForKeys: nil, options: .skipsSubdirectoryDescendants)
+                .forEach { file in
+                    try? manager.removeItem(atPath: file.path)
+                }
```
johnxnguyen: I suggest to move this into a new method `deleteTemporaryData()`.

johnxnguyen (changes requested):

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnranj: updated.

wire-ios-sync-engine/Tests/Source/SessionManager/SessionManagerTests.swift
johnranj: yes, added test file to double check deletion

wire-ios-sync-engine/Tests/Source/SessionManager/SessionManagerTests.swift
```diff
@@ -400,6 +400,31 @@ final class SessionManagerTests: IntegrationTest {
...
+
+        var tempUrl = self.tmpDirectoryPath().appendingPathComponent("testFile.txt")
+        let testData = "Test Message"
+        try? testData.write(to: tempUrl, atomically: true, encoding: .utf8)
```
johnxnguyen: Don't forget to assert the temp directory is not empty before logging out.

johnranj: As suggested, asserted temp directory when it Is not empty.

johnxnguyen: I don't see the assert here. I see the assert at the end that the temp directory is empty. But I was suggesting to assert the temp director is not empty:

```swift
// Given
let tempDirectory = ...
let file = ...
tempDirectory.addFile(file)
XCTAssertFalse(tempDirectory.isEmpty)

// When
...

// Then
...

```

The reason is that if we failed to set up the test and not add any file to the temp directory, then it wouldn't matter what happens in the WHEN part because the assertion that the temp directory is empty would always be true. By adding this assertion in GIVEN we can see if we failed to set up the test.

johnxnguyen (changes requested):

wire-ios-sync-engine/Tests/Source/SessionManager/SessionManagerTests.swift
johnranj: As suggested, assert is added

netbe (approved):

wire-ios-sync-engine/Tests/Source/SessionManager/SessionManagerTests.swift
johnxnguyen: `XCTAssertFalse(tempUrl.path.isEmpty)` looks like it's asserting that the path (a string maybe?) is not empty. Instead we should assert like you do below:

```
XCTAssertFalse(FileManager.default.fileExists(atPath: tempUrl.path))
```

johnxnguyen (changes requested):

wire-ios-sync-engine/Tests/Source/SessionManager/SessionManagerTests.swift
johnranj: It is added already, if you notice the entire method, this assert already exists.

johnxnguyen: But not on line 417. We want to assert that there is some data in the temp url, not asser that the path to the temp url is non empty.

johnxnguyen (approved):

PR #167, opened 2023-04-11
chore: add contextMenu and copy to clipboard action for DeveloperMenu cells

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Currently, if we want to copy the datadog id (as an example) and send it to someone, we can't. We either have to send a screenshot or just write down the first couple of characters of the id. That's time-consuming and kind of inefficient.

With this PR we fix this problem by providing a context menu to those cells and the ability for us to copy the contents of the cell.

----
##### References


wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DeveloperToolsView.swift
```diff
@@ -97,9 +97,14 @@ private struct TextItemCell: View {
                 .lineLimit(1)
                 .truncationMode(.middle)
                 .foregroundColor(.secondary)
+                .contextMenu {
+                    SwiftUI.Button(
+                        action: { UIPasteboard.general.string = value },
+                        label: { Text("Copy") }
+                    )
```
johnxnguyen: As discussed in another channel, this duplicates logic we have to copy on tapping the cell, however this provides better UI hints. However, this view shouldn't have any business logic in it, i.e I don't think we should be modifying the pasteboard here in the view.

wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DeveloperToolsView.swift
```diff
@@ -60,8 +60,12 @@ struct DeveloperToolsView: View {
 
         case let .text(textItem):
             TextItemCell(title: textItem.title, value: textItem.value)
-                .onTapGesture {
-                    viewModel.handleEvent(.itemTapped(item))
+                .contextMenu {
+                    SwiftUI.Button {
+                        viewModel.handleEvent(.itemTapped(item))
```
johnxnguyen: I suggest to rename the case `itemCopyRequested`

agisilaos: Done!

johnxnguyen: I mean, `viewModel.handleEvent(.itemCopyRequested(item))`

agisilaos: Done! 🎉

MarcinR (approved):

netbe:
can't we keep both behaviour like tapping and the context menu?
@agisilaos can you add haptic feedback or/and notification banner to inform user the copy was successful?

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
```diff
@@ -0,0 +1,50 @@
...
+}
+
+extension SwiftUI.Button {
+    init(
+        _ style: UINotificationFeedbackGenerator.FeedbackType,
+        action: @escaping () -> Void,
+        @ViewBuilder label: () -> Label
+    ) {
+        self.init(action: withFeedback(style, action), label: label)
+    }
```
johnxnguyen: Why create an initializer instead of using the modifier?

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
```diff
@@ -0,0 +1,50 @@
...
+}
+
+struct HapticTapGestureViewModifier: ViewModifier {
+    var style: UINotificationFeedbackGenerator.FeedbackType
+    var action: () -> Void
+
+    func body(content: Content) -> some View {
+        content.onTapGesture(perform: withFeedback(self.style, self.action))
+    }
+}
```
johnxnguyen: Where is this modifier being used?

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
```diff
@@ -0,0 +1,50 @@
...
+    _ style: UINotificationFeedbackGenerator.FeedbackType,
+    _ action: @escaping () -> Void
+) -> () -> Void {
+    { () in
+        let feedback = UINotificationFeedbackGenerator()
+        feedback.prepare()
+        feedback.notificationOccurred(style)
+        action()
+    }
+}
```
johnxnguyen: Why is this a free function?

wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DeveloperToolsView.swift
```diff
@@ -60,8 +60,12 @@ struct DeveloperToolsView: View {
...
         case let .text(textItem):
             TextItemCell(title: textItem.title, value: textItem.value)
-                .onTapGesture {
-                    viewModel.handleEvent(.itemTapped(item))
+                .contextMenu {
+                    SwiftUI.Button(.success) {
+                        viewModel.handleEvent(.itemCopyRequested(item))
+                    } label: {
+                        Label("Copy", systemImage: "doc.on.doc")
+                    }
```
johnxnguyen: `Button(.success)` is not does not reveal that the button will not only cause haptic feedback. A named argument would help here, but perhaps a view modifier would be more expressive?

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
agisilaos: Good question. We don't use it anywhere. I initially created that modifier to check if can be used on the button. But for some reason, I couldn't use it on the button. It wouldn't work.

Not sure if I tried to use it the wrong way but I doubt so.

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
agisilaos: "Free" as it's not part of any extension?

wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DeveloperToolsView.swift
agisilaos: I agree, I think a named argument would be much better here.

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
agisilaos: Same thing applies here: https://github.com/wireapp/wire-ios-mono/pull/167#discussion_r1170047560

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
```diff
@@ -0,0 +1,50 @@
...
+//
+
+import SwiftUI
+
+func withFeedback(
+    _ style: UINotificationFeedbackGenerator.FeedbackType,
+    _ action: @escaping () -> Void
+) -> () -> Void {
+    { () in
+        let feedback = UINotificationFeedbackGenerator()
```
netbe: you should hold the feedback inside the struct ViewModifier

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
netbe: did you get an error?

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
johnxnguyen: yes, it's define at the top level and isn't attached to any type.

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
```diff
@@ -0,0 +1,56 @@
...
+
+import SwiftUI
+
+struct HapticTapGestureViewModifier: ViewModifier {
+    var style: UINotificationFeedbackGenerator.FeedbackType
+
+    func body(content: Content) -> some View {
+        content.onTapGesture {
+            let feedback = UINotificationFeedbackGenerator()
+            feedback.prepare()
```
netbe:
```suggestion
struct HapticTapGestureViewModifier: ViewModifier {
    let feedback = UINotificationFeedbackGenerator()
    var style: UINotificationFeedbackGenerator.FeedbackType

    init(style: UINotificationFeedbackGenerator.FeedbackType) {
            self.style = style
            feedback.prepare()
    }

    func body(content: Content) -> some View {
        content.onTapGesture {
```

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
```diff
@@ -0,0 +1,56 @@
...
+
+    init(
+        hapticFeedbackStyle: UINotificationFeedbackGenerator.FeedbackType,
+        action: @escaping () -> Void,
+        @ViewBuilder label: () -> Label
+    ) {
+        self.init(
+            action: {
+                let feedback = UINotificationFeedbackGenerator()
+                feedback.prepare()
```
netbe: isn't it a problem to call it right after?

```
Think about when and where you can best prepare your generators. If you call [prepare()](https://developer.apple.com/documentation/uikit/uifeedbackgenerator/2369818-prepare) and then immediately trigger feedback, the system won’t have enough time to get the Taptic Engine into the prepared state, and you may not see a reduction in latency. On the other hand, if you call [prepare()](https://developer.apple.com/documentation/uikit/uifeedbackgenerator/2369818-prepare) too early, the Taptic Engine may become idle again before you trigger feedback.
```
https://developer.apple.com/documentation/uikit/uifeedbackgenerator#2555404

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
netbe: + at the end we don't use it actually

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
```diff
@@ -0,0 +1,56 @@
...
+            feedback.notificationOccurred(style)
+        }
+    }
+}
+
+extension View {
+    func hapticFeedback(_ feedbackType: UINotificationFeedbackGenerator.FeedbackType) -> some View {
+        self.modifier(HapticTapGestureViewModifier(style: feedbackType))
+    }
+}
```
johnxnguyen: This isn't being used anymore, perhaps we should delete it.

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
agisilaos: Yep! I will delete it

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
agisilaos: Hmm, interesting. As I was testing it I didn't experience any issues so not sure how to follow up on that

wire-ios/Wire-iOS/Sources/SwiftUI Helpers/HapticFeedbackModifier.swift
```diff
@@ -0,0 +1,38 @@
...
+        hapticFeedbackStyle: UINotificationFeedbackGenerator.FeedbackType,
+        action: @escaping () -> Void,
+        @ViewBuilder label: () -> Label
+    ) {
+        self.init(
+            action: {
+                let feedback = UINotificationFeedbackGenerator()
+                feedback.prepare()
+                feedback.notificationOccurred(hapticFeedbackStyle)
+                action()
```
netbe: maybe better to have the feedback after the action, no?

agisilaos: Done!

netbe (approved):

PR #170, opened 2023-04-11
fix: proteus messages not working app extensions FS-1791 FS-1829 FS-1830

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1791" title="FS-1791" target="_blank">StoryFS-1791</a>  [iOS] Crash on Bund RC col3
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR fixes 3 issues...

### Push notifications not shown for new messages 
This is because we failed to invoke `proteusInit` on Core Crypto. After fixing the proteus migration, we removed this invocation when setting up the `ProteusService`, and instead call it when completing the proteus migration. Note that completing proteus migration should be called even when no migration was performed. The solution is to call this method after setting up core crypto in the notification extension.

### Can't send messages from the share extension
This is the exact same problem as with the notification extension. The solution is also the same.

### Can't initialize Core Crypto when the device is locked
This problem manifests itself in different ways: 
- if the notification extension is awoken while the device is locked, we fail to set up core crypto and silently fail, not showing any notification. 
- if the main app is woken in the background while the device is locked to refresh itself, we'll actually force crash the app because we expect core crypto to be set up and can't proceed without it.

The cause comes from within Core Crypto and appears to originate from the keychain. Core Crypto tries to fetch a keychain item that it stores, but is unable to because the device is locked. It turns out that Core Crypto doesn't specify an access level for the item, so the default `kSecAttrAccessibleWhenDeviceIsUnlocked` is used.  

Core Crypto should set the correct access level `kSecAttrAccessibleAfterFirstUnlock` but it can't do so easily with the library it currently uses. It's easier at the moment for us to update the access level on our side, until Core Crypto can figure it out on their side.

The workaround solution is to query the keychain for all the items that Core Crypto creates, then update those items with the correct access level.

### Testing

I've tested this manually, not sure how to test this programatically yet, I'll try to do so in a separate PR.

----

----

----
##### References

### DIFF
--- a/wire-ios-sync-engine/Source/UserSession/ZMUserSession+CoreCrypto.swift
+++ b/wire-ios-sync-engine/Source/UserSession/ZMUserSession+CoreCrypto.swift
@@ -60,6 +60,7 @@ extension ZMUserSession {
                     key: configuration.key
                 )
 
+                updateKeychainItemAccess()
                 syncContext.coreCrypto = coreCrypto
                 createProteusServiceIfNeeded(coreCrypto: coreCrypto)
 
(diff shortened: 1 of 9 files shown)


KaterinaWire (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession+CoreCrypto.swift
```diff
@@ -70,6 +71,59 @@ extension ZMUserSession {
...
+    //
+    // Workaround: set the access level for the keychain item on our side.
+
+    private func updateKeychainItemAccess() {
+        WireLogger.coreCrypto.info("updating keychain item access")
+
+        for account in accountsForAllItemsNeedingUpdates() {
+            let query = [
+                kSecClass: kSecClassGenericPassword,
+                kSecAttrService: "wire.com",
```
MarcinR, line 92: you can put hardcoded constats as class let

netbe (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession+CoreCrypto.swift
```diff
@@ -70,6 +71,59 @@ extension ZMUserSession {
...
+            return []
+        }
+
+        let items = result as? [[String: Any]] ?? []
+
+        return items.compactMap { item in
+            item[kSecAttrAccount as String] as? String
+        }.filter { account in
+            // Core Crypto says that the items are all prefixed with this.
+            account.hasPrefix("keystore_salt")
```
MarcinR, line 123: as above

PR #198, opened 2023-05-03
fix: call not ringing when already in a call on lock screen FS-1920

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1920" title="FS-1920" target="_blank">BugFS-1920</a>  [iOS] Incoming group call doesn't ring while already in a 1:1 call while the screen is locked
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When the user is in a call while on the Lock Screen, and a second call is incoming, the call doesn't ring / show up on the screen

### Causes 

Since we're already on a call, Wire is active despite being in the background. As a result, it syncs the events and processes other incoming call events.
However, the `CallKitManager` filters out call state updates if Wire is in the background, because we want to avoid interference with the NSE eventually reporting call state updates.

### Solutions

We don't need to filter out call states when in the background, as long as we avoid reporting call events more than once.

#### Test Coverage 

- [x] Added unit tests to ensure we don't process the same call events more than once

#### How to Test

1. Be on the Lock Screen (with Wire in the background)
2. Accept a 1:1 call (while on lock screen)
3. Receive a group call (while on lock screen)

Expected: the second call rings and appears on the screen as incoming

### DIFF
--- a/wire-ios-sync-engine/Source/Calling/CallKitManager.swift
+++ b/wire-ios-sync-engine/Source/Calling/CallKitManager.swift
@@ -807,18 +807,6 @@ extension CallKitManager: CXProviderDelegate {
 
 extension CallKitManager: WireCallCenterCallStateObserver, WireCallCenterMissedCallObserver {
 
-    private var shouldHandleCallStates: Bool {
-        switch (requirePushTokenType, application.applicationState) {
-        case (.standard, .background):
-            // Processing call states in the background wiLl interfere with
-            // CallKit calls triggered from the NSE.
-            return false
-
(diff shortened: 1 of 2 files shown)


agisilaos (approved):

agisilaos:
@David-Henner **FYI**: In **Sync Engine**, under **UserTests**, **testThatAllUsersAreDownloaded** is failing

KaterinaWire (approved):

PR #199, opened 2023-05-04
fix: after rejecting a call via CallKit the call rings again when opening the app FS-1918

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1918" title="FS-1918" target="_blank">BugFS-1918</a>  [iOS] After rejecting a call via CallKit the call rings again when opening the app
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

After rejecting a call via CallKit (with the app in the background), the call rings again once the app is brought to the foreground 

### Causes 

When coming to the foreground, the update event corresponding to the call (that was rejected while the app was in the background) is processed and reported as an incoming call. 

### Solutions

Sync and process update events before rejecting the call. That way we process the incoming call event while it is still present in the call register, meaning that it won't be reported twice as incoming. Then we can reject the call and remove it from the call register.

#### Test Coverage (Optional)

- [x] Added some unit tests

#### How to Test

1. Enable CallKit and microphone permissions
2. Put app in the background.
3. Receive incoming call (1-1 or group).
4. Reject the call
5. Bring app to foreground

Expected result: no incoming call

### DIFF
--- a/wire-ios-sync-engine/Tests/Source/Calling/CallKitManagerTests.swift
+++ b/wire-ios-sync-engine/Tests/Source/Calling/CallKitManagerTests.swift
@@ -131,6 +131,21 @@ class MockStartCallAction: CXStartCallAction {
 
 }
 
+class MockEndCallAction: CXEndCallAction {
+
+    var isFulfilled: Bool = false
+    var hasFailed: Bool = false
+
+    override func fulfill() {
+        isFulfilled = true
+    }
+
+    override func fail() {
+        hasFailed = true
+    }
+
+}
+
 class MockProvider: CXProvider {
 
     var connectingCalls: Set<UUID> = Set()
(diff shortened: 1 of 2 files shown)


wire-ios-sync-engine/Tests/Source/Calling/CallKitManagerTests.swift
```diff
@@ -945,4 +962,75 @@ class CallKitManagerTest: DatabaseTest {
...
+        let otherUser = otherUser(moc: uiMOC)
+
+        sut.reportIncomingCall(from: otherUser, in: conversation, hasVideo: false)
+
+        guard let callKitCall = sut.callRegister.lookupCall(by: conversation) else {
+            XCTFail("no call handle")
+            return
+        }
+
+        mockCallKitManagerDelegate.mockConversations = [:]
```
johnxnguyen, line 1025: Have no mock conversation is what causes it to fail?

agisilaos (approved):

wire-ios-sync-engine/Tests/Source/Calling/CallKitManagerTests.swift
David-Henner: Yes

johnxnguyen (approved):

PR #203, opened 2023-05-08
refactor: process call events in background FS-1596

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1596" title="FS-1596" target="_blank">TaskFS-1596</a>  [iOS] Support CallKit when EAR is enabled
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
When trying to answer a CallKit call from the Lock Screen, we're unable to successfully connect the call.

### Causes
In order to connect the call, we need to fetch the stored call event from the event database and process it. Processing call events means to pass them to AVS. Once AVS has processed the event, then we are able to use AVS to connect the call.

The problem in this flow is that Wire is active in the background while the device is locked. Since the device is locked, the database is locked, which is preventing us from being able to process events. This is an artificial condition, but one that usually makes sense, since we don't want to process events that could require interacting with a locked database. However, this is not the case with call events, since we only forward the events to AVS.

### Solutions
- Introduce new operation state `backgroundPendingCall`. 
- We transition to this state on demand when processing CallKit actions.
- When the app is in this state, it is allowed to process events in the background.
- Otherwise, use existing conditions to determine whether event processing is allowed.

#### Test Coverage

- Unit tests for operation state
- Unit tests for event processor "is ready to process events"

----

----

----
##### References


MarcinR (approved):

wire-ios-sync-engine/Source/UserSession/OperationStatus.swift
```diff
@@ -86,8 +114,15 @@ public class OperationStatus: NSObject {
...
 
+    var hasPendingCall = false {
+        didSet {
+            updateOperationState()
+        }
+    }
+
     public var hasOngoingCall = false {
         didSet {
+            hasPendingCall = false
```
David-Henner: Is this safe to set this to false each time `hasOngoingCall` is set ?

johnxnguyen: I think so. My reasoning was that if we say a call is ongoing, or it is no longer ongoing, then we no longer have a pending call.

David-Henner: I can think of one scenario where this doesn't work. 
- User A has 1:1 call with user B
- User C calls users A & B in a group call
- User A answers the incoming call

-> The call between user A & B gets terminated > `hasOngoingCall` is set to false > `hasPendingCall` is also set to false 

User B still hasn't answered the call but cannot because there is no pending call according to the operation status

David-Henner (approved):

PR #207, opened 2023-05-10
refactor: call event processing

David-Henner:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

**Processing call events** is a **necessary** step in the process of answering / rejecting calls we receive in the background. 

We trigger event processing by requesting a quick sync. To process the events, we also need to be _"ready"_ for processing. Which means we're **not currently syncing** and the **database is unlocked**.

In order to process call events while the database is locked, we changed the conditions for being _ready_ to process events. If the current state of the app was **in the background** and **pending a call** we would allow event processing. (we'd assume the only events to process were call events, which we're able to process while the app is locked thanks to secondary keys)

However, maintaining an accurate operation state is tricky and ended up not being reliable, which caused issues answering calls. 

To solve this, we stop requesting the quick sync to process all events. Instead we directly process call events that were stored by the NSE.

### DIFF
--- a/wire-ios-sync-engine/Tests/Source/Calling/CallKitManagerTests.swift
+++ b/wire-ios-sync-engine/Tests/Source/Calling/CallKitManagerTests.swift
@@ -1,6 +1,6 @@
 //
 // Wire
-// Copyright (C) 2016 Wire Swiss GmbH
+// Copyright (C) 2023 Wire Swiss GmbH
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
--- a/wire-ios-sync-engine/Tests/Source/Calling/CallKitManagerTests.swift
+++ b/wire-ios-sync-engine/Tests/Source/Calling/CallKitManagerTests.swift
@@ -500,22 +500,22 @@ class CallKitManagerTest: DatabaseTest {
     }
 
     /* Disabled for now, pending furter investigation
-    func testThatCallAnswerActionFailWhenCallCantBeJoined() {
-        // given
-        let otherUser = self.otherUser(moc: self.uiMOC)
-        let provider = MockProvider(foo: true)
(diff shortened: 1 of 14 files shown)


johnxnguyen (approved):
Looks good 👍🏻

wire-ios-sync-engine/Tests/Source/Calling/CallKitManagerTests.swift
```diff
@@ -500,22 +500,22 @@ class CallKitManagerTest: DatabaseTest {
...
-        let action = MockCallAnswerAction(call: sut.callUUID(for: conversation)!)
-        self.sut.provider(provider, perform: action)
-
-        // when
-        mockWireCallCenterV3.update(callState: .terminating(reason: .lostMedia), conversationId: conversation.remoteIdentifier!, callerId: otherUser.remoteIdentifier, isVideo: false)
-        XCTAssertTrue(waitForAllGroupsToBeEmpty(withTimeout: 0.5))
-        // then
-        XCTAssertTrue(action.hasFailed)
-    }
+     func testThatCallAnswerActionFailWhenCallCantBeJoined() {
```
netbe, line 503: just passing through this, is there a ticket somewhere about this? if so put a comment with the Jira ticket, else create one

netbe: or is it still relevant? to the current code

netbe (approved):
just a comment on some commented code

wire-ios-sync-engine/Tests/Source/Calling/CallKitManagerTests.swift
David-Henner: I have no idea, it only made it into this PR because the indentation changed somehow

PR #233, opened 2023-05-23
feat: Refetch missing metadata FS-1475 

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1475" title="FS-1475" target="_blank">TaskFS-1475</a>  [iOS]  Implemented client protocol for missing conversation/users metadata
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

1. Whenever the user opens a conversation details screen and any of the participants of the conversation is missing metadata, try again to fetch those users.

2. Refetch conversations and users without metadata on the Search UI.

3. Whenever the app comes to the foreground, if at least 3 hours have passed since the last time this check was performed, then try again to fetch those conversations and users without metadata.

----

----

----
##### References


wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
```diff
@@ -0,0 +1,116 @@
...
+//
+
+import Foundation
+
+@objc
+public protocol TimerActionsManagerType: AnyObject {
+
+    func applyTimerActionsIfNeeded(_ checkTime: Date)
+
+}
```
johnxnguyen: Some naming suggestions:

```swift
@objc
public protocol RecurringActionServiceInterface: AnyObject {

    func performActionsIfNeeded()

}
```

Is there a need to pass in the date if it's always going to be the current date?

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
```diff
@@ -0,0 +1,116 @@
...
+
+public final class TimerActionsManager: NSObject, TimerActionsManagerType {
```
johnxnguyen: In general, this class from the outside would appear to be not related to metadata refreshing, but internally its implementation is specific to metadata refreshing. I suggest to make this class agnostic to the type of actions it performs and the frequency they should be performed.

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
```diff
@@ -0,0 +1,116 @@
...
+
+    // MARK: - Models
+
+    struct Event {
+        let action: (() -> Void)?
+    }
+
+    /// Performing actions every 3 hours.
+    private let interval: TimeInterval = 3 * 60 * 60
```
johnxnguyen: I suggest to remove replace the hardcoded interval and provide a method to add an action for a specific time interval:

```swift
func registerRecurringAction(forInterval interval: TimeInterval, action: () -> Void)
```

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
```diff
@@ -0,0 +1,116 @@
...
+
+    weak var managedObjectContext: NSManagedObjectContext?
```
johnxnguyen: Why does this class need to know about a context? Doing so tightly couples it to Core Data. If you need to switch to a context and perform saves, do it within the action instead.

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
```diff
@@ -0,0 +1,116 @@
...
+
+}
+
+public final class TimerActionsManager: NSObject, TimerActionsManagerType {
+
+    // MARK: - Models
+
+    struct Event {
+        let action: (() -> Void)?
+    }
```
johnxnguyen: A type alias would be more suitable here: `typealias Action = () -> Void`

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
```diff
@@ -0,0 +1,116 @@
...
+
+    // MARK: - Initialization
+
+    public init(managedObjectContext: NSManagedObjectContext) {
+        self.managedObjectContext = managedObjectContext
+
+        super.init()
+    }
+
+    public func applyTimerActionsIfNeeded(_ currentCheckTime: Date) {
```
johnxnguyen: There's probably no need to pass in the current check time, just take the current time from inside the method.

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
```diff
@@ -0,0 +1,116 @@
...
+        managedObjectContext?.performGroupedBlock {
+            guard let lastCheckDate = UserDefaults.standard.lastDataRefreshDate else {
```
johnxnguyen: I suggest to inject the user defaults instance, this will help prevent unstable tests since we can inject a new dummy user defaults for each test run.

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    private func setLastCheckTime(_ lastCheckDate: Date?) {
+        UserDefaults.standard.lastDataRefreshDate = lastCheckDate
+    }
+
+    private func makeEventsList() -> [Event] {
+        return [refreshUsersWithMissingMetadata(),
+                refreshConversationsWithMissingMetadata()]
+    }
```
johnxnguyen: This tightly couples the class with metadata refreshing. In another comment I suggested to register actions using a new method. Doing so will allow us to reuse this class for any use case.

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
```diff
@@ -0,0 +1,116 @@
...
+    var lastDataRefreshDate: Date? {
+
+        get {
+            return object(forKey: lastDataRefreshDateKey) as? Date
+        }
+
+        set {
+            set(newValue, forKey: lastDataRefreshDateKey)
+        }
+    }
```
johnxnguyen: This doesn't need to be public, I would make it private.

wire-ios-sync-engine/Source/UserSession/Search/SearchDirectory.swift
```diff
@@ -52,6 +52,25 @@ import Foundation
         return task
     }
 
+    public func refetchIncompleteUserMetadata() {
+        let managedObjectContext = contextProvider.viewContext
```
johnxnguyen: Is there a specific reason to do this on the view context?

wire-ios-sync-engine/Source/UserSession/Search/SearchDirectory.swift
```diff
@@ -52,6 +52,25 @@ import Foundation
...
+        let fetchRequest = ZMUser.sortedFetchRequest(with: ZMUser.predicateForUsersWithIncompleteMetadata())
+        guard let users = managedObjectContext.fetchOrAssert(request: fetchRequest) as? [ZMUser] else {
+            return
+        }
+        users.forEach { $0.refreshData() }
+    }
+
+    public func refetchIncompleteConversationMetadata() {
+        let managedObjectContext = contextProvider.viewContext
```
johnxnguyen: Is there a specific reason to do this on the view context?

wire-ios-sync-engine/Source/UserSession/Search/SearchDirectory.swift
johnxnguyen: This appears to duplicate code in the timer manager class.

wire-ios-sync-engine/Source/UserSession/Search/SearchDirectory.swift
johnxnguyen: This appears to duplicate code in the timer manager class.

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsViewController.swift
```diff
@@ -340,3 +342,18 @@ extension GroupDetailsViewController: GroupDetailsSectionControllerDelegate, Gro
     }
 
 }
+
+extension ZMConversation {
+
+    func refetchParticipantsIfNeeded() {
+        guard let session = ZMUserSession.shared() else {
```
johnxnguyen: I don't think the `ZMConversation` should really know about the `ZMUserSession`, I would suggest to do something like: 

```swift
session.perform {
    conversation.refetchParticipantsIfNeeded()
}
```

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsViewController.swift
```diff
@@ -340,3 +342,18 @@ extension GroupDetailsViewController: GroupDetailsSectionControllerDelegate, Gro
...
+            return
+        }
+        sortedOtherParticipants.filter { $0.hasIncompleteMetadata }.forEach { user in
+            session.perform {
+                user.refreshData()
+            }
+        }
```
johnxnguyen: Suggestion:

```swift
for user in sortedOtherParticipants where user.hasIncompleteMetadata {
    user.refreshData()
}
```

wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests+Timers.swift
```diff
@@ -28,4 +28,28 @@ class ZMUserSessionTimersTests: ZMUserSessionTestsBase {
...
+        // when
+        XCTAssertFalse(timerActionsManager.applyTimerActionsIsCalled)
+        sut.didFinishQuickSync()
+
+        // then
+        XCTAssertTrue(timerActionsManager.applyTimerActionsIsCalled)
+    }
+}
+
+private class MockTimerActionsManager: TimerActionsManagerType {
```
netbe: use separate file for class

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
KaterinaWire: Done

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
KaterinaWire: Updated

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
KaterinaWire: Changed

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
KaterinaWire: removed

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
KaterinaWire: changed

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
KaterinaWire: done

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
KaterinaWire: changed

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
KaterinaWire: done

wire-ios-sync-engine/Source/UserSession/Search/SearchDirectory.swift
KaterinaWire: because we are on the main thread. Updated this code

wire-ios-sync-engine/Source/UserSession/Search/SearchDirectory.swift
KaterinaWire: because we are on the main thread. Updated this code

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsViewController.swift
KaterinaWire: done

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsViewController.swift
KaterinaWire: done

wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests+Timers.swift
KaterinaWire: done

wire-ios-sync-engine/Source/Synchronization/TimerActionsManager.swift
KaterinaWire: Done

wire-ios-sync-engine/Source/Synchronization/RecurringActionService.swift
```diff
@@ -0,0 +1,95 @@
...
+    private var actions = [RecurringAction]()
+
+    public func performActionsIfNeeded() {
+        let currentDate = Date()
+
+        actions.forEach { action in
+
+            guard let lastActionDate = lastActionDate(for: action.id) else {
+                persistLastActionDate(for: action.id, newDate: currentDate)
+                return
```
johnxnguyen, line 50: Returning will break the loop. Did you mean to use `continue`?

wire-ios-sync-engine/Source/Synchronization/RecurringActionService.swift
```diff
@@ -0,0 +1,95 @@
...
+
+    // MARK: - Helpers
+
+    func lastActionDate(for actionID: String) -> Date? {
+        return userDefaults(for: actionID)?.lastRecurringActionDate
+    }
+
+    func persistLastActionDate(for actionID: String, newDate: Date) {
+        userDefaults(for: actionID)?.lastRecurringActionDate = newDate
+    }
```
johnxnguyen: Do you need to pass in the new date if it is always `Date()`?

wire-ios-sync-engine/Source/Synchronization/RecurringActionService.swift
```diff
@@ -0,0 +1,95 @@
...
+
+    private func userDefaults(for actionID: String) -> UserDefaults? {
+        return UserDefaults(suiteName: "com.wire.recurringAction.\(actionID)")
```
johnxnguyen: I suggest to create this once and store it in a property. Also I think we just allow injecting the defaults so we can set a different one for testing.

wire-ios-sync-engine/Source/UserSession/ZMUserSession+RecurringAction.swift
```diff
@@ -0,0 +1,56 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension ZMUserSession {
+
+    func refreshUsersMissingMetadata() -> RecurringAction {
+        /// Performing actions every 3 hours.
+        let interval: TimeInterval = 3 * 60 * 60
```
johnxnguyen: Suggestion:

```swift
let interval: TimeInterval = 3 * .hour
```

wire-ios-sync-engine/Source/UserSession/ZMUserSession+RecurringAction.swift
```diff
@@ -0,0 +1,56 @@
...
+
+        return RecurringAction(id: UUID().uuidString, interval: interval) {
```
johnxnguyen: Every time you create a new action it will have a different ID. This ID is persisted in the user defaults which will persist data between launches. This means you'll sometimes have no last check date because you're querying with a new id.

I suggest to hardcode an id name, like `refreshUserMetadata`

wire-ios-sync-engine/Source/Synchronization/RecurringActionService.swift
KaterinaWire: it's useful for tests

KaterinaWire: and it may take some time for the action to run, and I think it's good to persist the same date that we use to execute the actions.

wire-ios-sync-engine/Source/Synchronization/RecurringActionService.swift
KaterinaWire: No, 'continue' is only allowed inside a loop. I will return only from guard.

wire-ios-sync-engine/Source/UserSession/ZMUserSession+RecurringAction.swift
KaterinaWire: done

wire-ios-sync-engine/Source/UserSession/ZMUserSession+RecurringAction.swift
KaterinaWire: done

wire-ios-sync-engine/Source/Synchronization/RecurringActionService.swift
KaterinaWire: But each action should have its own userDefault, because the interval can be different. I created a property for a UserDefaults name, but keep the same method. 
I also use a different defaults in tests.

wire-ios-sync-engine/Source/Synchronization/RecurringActionService.swift
KaterinaWire: I removed the date from the method parameters.

wire-ios-sync-engine/Source/UserSession/Search/SearchDirectory.swift
```diff
@@ -25,12 +25,18 @@ import Foundation
...
+    public var refetchIncompleteConversationMetadata: (() -> Void)?
+
     deinit {
         assert(isTornDown, "`tearDown` must be called before SearchDirectory is deinitialized")
     }
 
     public convenience init(userSession: ZMUserSession) {
         self.init(searchContext: userSession.searchManagedObjectContext, contextProvider: userSession, transportSession: userSession.transportSession)
+
+        self.refetchIncompleteUserMetadata = userSession.refreshUsersMissingMetadata().perform
```
netbe: what's perform method?

wire-ios-sync-engine/Source/UserSession/ZMUserSession+RecurringAction.swift
```diff
@@ -0,0 +1,62 @@
...
+                let fetchRequest = ZMUser.sortedFetchRequest(with: ZMUser.predicateForUsersArePendingToRefreshMetadata())
+                guard let users = self?.managedObjectContext.fetchOrAssert(request: fetchRequest) as? [ZMUser] else {
+                    return
+                }
+                users.forEach { $0.refreshData() }
+            }
+        }
+    }
+
+    func refreshConversationsMissingMetadata() -> RecurringAction {
```
netbe: maybe add the time as default param in morder to test this method

wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests+RecurringActions.swift
```diff
@@ -0,0 +1,120 @@
...
+        // when
+        XCTAssertEqual(mockRecurringActionService.actions.count, 0)
+        sut.recurringActionService.registerAction(action)
+
+        // then
+        XCTAssertEqual(mockRecurringActionService.actions.count, 1)
+    }
+}
+
+class RecurringActionServiceTests: ZMTBaseTest {
```
netbe: move to own file

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileActionsFactory.swift
```diff
@@ -200,7 +200,8 @@ final class ProfileActionsFactory {
             }
 
             // If the user is not from the same team as the other user, allow blocking
-            if user.isConnected && !isOnSameTeam && !user.isWirelessUser && !user.hasIncompleteMetadata {
+
+            if user.isConnected && !isOnSameTeam && !user.isWirelessUser && !user.hasEmptyName {
```
netbe: is hasEmptyName a stored property or a computed one?

KaterinaWire: `hasEmptyName` is computed property. It's a helper to reduce repeated code
```
 public var hasEmptyName: Bool {
        guard let name = name else {
            return true
        }
        return name.isEmpty
    }
```

wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests+RecurringActions.swift
KaterinaWire: moved

wire-ios-sync-engine/Source/UserSession/ZMUserSession+RecurringAction.swift
KaterinaWire: I moved timestamp to the method parameters with a default value and added tests:
`testThatItUpdatesUsersMissingMetadata()`
`testThatItUpdatesConversationsMissingMetadata()`

wire-ios-sync-engine/Source/UserSession/Search/SearchDirectory.swift
KaterinaWire: `perform` is actually an action of the `RecurringAction`. I changed the code a bit and added a public `updateIncompleteMetadataIfNeeded() `method that will update missing metadata when the user opens the search UI.

MarcinR (approved):

wire-ios-sync-engine/Source/Synchronization/RecurringActionService.swift
johnxnguyen: But one user defaults instance can store several pieces of data. Why not have a single user defaults instances for the recurring action service, and just have a separate key (using the action id) for each action that needs to be persisted? 

Having a single user defaults instance for a single action is like having a whole database to store a single piece of data.

wire-ios-sync-engine/Source/UserSession/ZMUserSession+RecurringAction.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    }
+
+}
+
+extension TimeInterval {
+
+    static let hour: Self = 60 * 60
+
+}
```
johnxnguyen: There already exists these helpers in DataModel and elsewhere, perhaps we should collect them and move them to WireUtilities so we avoid duplicating this code.

wire-ios-sync-engine/Source/Synchronization/RecurringActionService.swift
```diff
@@ -0,0 +1,96 @@
...
+protocol RecurringActionServiceInterface {
+
+    func performActionsIfNeeded()
+    func registerAction(_ action: RecurringAction)
+
+}
+
+class RecurringActionService: RecurringActionServiceInterface {
+
+    var userDefaultsName = "com.wire.recurringAction."
```
johnxnguyen: I think it would be better to inject in the user defaults instance for test purposes rather than modifying the name used to create the user defaults.

wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests+RecurringActions.swift
```diff
@@ -0,0 +1,126 @@
...
+        let action = RecurringAction(id: "11", interval: 5, perform: {})
+        sut.recurringActionService = mockRecurringActionService
+
+        // when
+        XCTAssertEqual(mockRecurringActionService.actions.count, 0)
+        sut.recurringActionService.registerAction(action)
+
+        // then
+        XCTAssertEqual(mockRecurringActionService.actions.count, 1)
+    }
```
johnxnguyen: This is testing the mock recurring action service rather than the user session.

wire-ios-sync-engine/Source/UserSession/ZMUserSession+RecurringAction.swift
KaterinaWire: Yes. Moved it form UI project and DataModel to WireUtilities

wire-ios-sync-engine/Source/Synchronization/RecurringActionService.swift
KaterinaWire: How do you suggest injecting the user defaults instance, since we should have those instances per action, and not per `RecurringActionService`? Each action has its own time interval. `RecurringActionService` will contain several actions. 
If you think we should inject user defaults instance per action, it looks like we should extend `RecurringAction`:
```
struct RecurringAction {

    typealias Action = () -> Void

    let id: String
    let interval: TimeInterval
    let perform: Action

    /// New
   let storage: UserDefaults
}
```

wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests+RecurringActions.swift
KaterinaWire: Why not test this logic with a mock?

wire-ios-sync-engine/Source/Synchronization/RecurringActionService.swift
johnxnguyen: I think you should have one user defaults instance for the whole service, and have an individual key for each action. The key could be for example: "lastCheckDate_/(action.id)"

wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests+RecurringActions.swift
johnxnguyen: I mean to say, the "when" part is not calling a method on sut, but on the recurring action service, which appears to be a mock instance.

wire-ios-sync-engine/Source/Synchronization/RecurringActionService.swift
KaterinaWire: I applied your suggestion. Please check it out.

wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests+RecurringActions.swift
KaterinaWire: Yes, you are right. I removed the part with a `sut`

wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests+RecurringActions.swift
```diff
@@ -0,0 +1,125 @@
...
+        XCTAssertTrue(mockRecurringActionService.performActionsIsCalled)
+    }
+
+    func testThatItAddsActions() {
+        // given
+        let action = RecurringAction(id: "11", interval: 5, perform: {})
+
+        // when
+        XCTAssertEqual(mockRecurringActionService.actions.count, 0)
+        mockRecurringActionService.registerAction(action)
```
johnxnguyen: Why are we testing the behavior of a mock. I think if we want to test registering actions then we should do it in RecurringActionServiceTests.

KaterinaWire: Moved to RecurringActionServiceTests

johnxnguyen (approved):

PR #287, opened 2023-06-20
chore: Use CirrusCI API instead of cirrus action for GH Actions  WPB-2276

SzymonMrozek:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-2276" title="WPB-2276" target="_blank">Sub-taskWPB-2276</a>  PoC experiment with implementing GithubActions that would trigger CirrusCI builds
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Changes how CirrusCI builds are triggered from Github Actions 

### Solutions

The new solution uses triggering by CirrusCI API 

#### How to Test

Trigger all possible variants after merge 

----

----

----
##### References


fastlane/Fastfile
```diff
@@ -127,6 +129,31 @@ platform :ios do
         build_and_upload_app(options)
     end
 
+    lane :trigger_api_build_cirrus_ci do |options|
+        authorization_header = { "Authorization" => "Bearer #{ENV['CIRRUS_CI_API_TOKEN']}" }
+        uri = 'https://api.cirrus-ci.com/graphql'
+        create_build = HTTParty.post(
+            uri,
+            body: "{ \"query\": \"mutation CreateBuildMutation($input: RepositoryCreateBuildInput!) { createBuild(input: $input) { build { id } } }\", \"variables\": { \"input\": {\"repositoryId\": \"4507403593449472\", \"branch\": \"#{ENV['GITHUB_REF_NAME']}\", \"clientMutationId\": \"trigger-build-on-cirrus\"}} }",
```
netbe: can we use multiline string and avoid escaping the `"`? like
```suggestion
            body: <<-BODY
"{ \"query\": \"mutation CreateBuildMutation($input: RepositoryCreateBuildInput!) { createBuild(input: $input) { build { id } } }\", \"variables\": { \"input\": {\"repositoryId\": \"4507403593449472\", \"branch\": \"#{ENV['GITHUB_REF_NAME']}\", \"clientMutationId\": \"trigger-build-on-cirrus\"}} }"
BODY
```

fastlane/Fastfile
```diff
@@ -127,6 +129,31 @@ platform :ios do
...
+            headers: authorization_header
+        )
+        created_build_id =  JSON.parse(create_build.body, symbolize_names: true)[:data][:createBuild][:build][:id]
+        `sleep 7`
```
netbe: feels a bit hacky

netbe:
@SzymonMrozek I would prefer we stick to the github action, did you try with the task argument https://github.com/cirruslabs/cirrus-action/blob/ac5191d18a9bff05ff1a38f280f2b6d768e32153/action.yml#L11 ?

SzymonMrozek:
> @SzymonMrozek I would prefer we stick to the github action, did you try with the task argument https://github.com/cirruslabs/cirrus-action/blob/ac5191d18a9bff05ff1a38f280f2b6d768e32153/action.yml#L11 ?

The cirrus-ci action is not designed to be used for our case. What it does is to run the cirrus-ci task locally on github action environment (it uses tart to build up local enviroment identical to the one from cirrus-ci machines) 
The use case we have here is that we just want to trigger the build on the cirrus-ci side, not replicate it locally

fastlane/Fastfile
SzymonMrozek: Yeah, but it seems there is no other option rather than creating a websocket and listening for changes which for me seems to be to complicated for that case

netbe: sorry I still don't get why we have a sleep of 7?

SzymonMrozek: Seems that scheduling jobs for the newly created pipeline is an async event (which seems reasonable actually to process .cirrus.yml for all possible jobs that may be triggered by the API event)
So I preferred to wait a bit longer than 1 second (that would be probably enough for our repo) since the whole GH action would take less than a minute

netbe: can you add a comment then to it so we don't forget in 6 months?

fastlane/Fastfile
SzymonMrozek: Please recheck

netbe (approved):

agisilaos (approved):

PR #292, opened 2023-06-26
feat: leave subconversation when leaving MLS conference WPB-636

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-636" title="WPB-636" target="_blank">StoryWPB-636</a>  [iOS] Leave subgroup after leaving conference call
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?
When leaving an MLS conference, we must also leave the subconversation hosting the conference. In this PR:

- add action to leave a subconversation
- add mls service method to leave the subconversation (calling the action and wiping the conversation)
- when leaving the call, invoke the mls service method to leave the subconversation
- to handle stale conferences (due to a crash during a call), also leave the subconversation when we detect an incoming call, terminated call, or a missed call.

#### Test Coverage

- unit tests asserting leave subconversation action
- unit tests asserting leaving subconversation
- unit tests asserting leaving subconversation if needed
- unit tests asserting WireCallCenter invokes methods to leave the subconversation

----

----

----
##### References


wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
```diff
@@ -664,6 +664,40 @@ extension WireCallCenterV3 {
...
+            }
+
+            Task {
+                try await mlsService.leaveSubconversation(
+                    parentQualifiedID: QualifiedID(uuid: id.identifier, domain: domain),
+                    parentGroupID: parentGroupID,
+                    subconversationType: .conference
+                )
+            }
+        }
```
David-Henner: This implementation seems very similar to `leaveStaleConferenceIfNeeded` in `WireCallCenterV3+MLS`. I see that the main difference is that we don't call the same method from `mlsService` (`leaveSubconversation` vs `leaveSubconversationIfNeeded`). Could you help me understand the difference between the two and why one is used over the other?

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1255,6 +1270,91 @@ public final class MLSService: MLSServiceInterface {
...
+                parentGroupID: parentGroupID,
+                subconversationType: subconversationType
+            )
+        }
+
+        if
+            let subConversationGroupID = subconverationGroupIDRepository.fetchSubconversationGroupID(
+                forType: subconversationType,
+                parentGroupID: parentGroupID
+            ),
```
David-Henner, line 1292: should the subconversation be removed from the repository after being left?

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
johnxnguyen: When we close a call we know we should leave the sub conversation, so we invoke `mlsService.leaveSubconversation` directly. This will leave the subconversation on the backend and then wipe the subgroup in CC. This is the happy path, where the self client is present when the call ends (by either side).

Suppose the self client crashes while in a call. When it relaunches, the call may be ongoing still, but the self client isn't part of the call anymore. It is however still part of the subconversation (backend) and subgroup (CC). To correct this state, we need to detect whether the subconversation has gone stale. This is done for instance by observing an incoming call state, which the self client will receive from the ongoing call. In this case we can detect that we're part of the subconversation but not the call, so we correct the state by leaving the subconversation.

`mlsService.leaveSubconversationIfNeeded` contains the logic for checking if we're part of a subconversation, and assuming that we shouldn't be, will leave the subconversation if we are. It does so by invoking `mlsService.leaveSubconversation`.

Perhaps the naming between the mls service methods and the call center methods is a bit confusing.

wire-ios-data-model/Source/MLS/MLSService.swift
johnxnguyen: yes, fixed

wire-ios-data-model/Tests/MLS/MLSServiceTests.swift
```diff
@@ -1872,6 +1872,172 @@ class MLSServiceTests: ZMConversationTestsBase, MLSServiceDelegate {
         XCTAssertEqual(subconversationGroupID.parentGroupID, parentID)
     }
 
+    func test_LeaveSubconversation() async throws {
```
netbe, line 1875: this can be taken in another PR, but I think we should split up this tests class > 1800 lines

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
David-Henner: Yes perhaps renaming the methods or adding comments could help bring more clarity

wire-ios-request-strategy/Sources/Request Strategies/MLS/Actions/LeaveSubconversationActionHandler.swift
```diff
@@ -0,0 +1,99 @@
...
+
+final class LeaveSubconversationActionHandler: ActionHandler<LeaveSubconversationAction> {
+
+    // MARK: - Request
+
+    override func request(
+        for action: Action,
+        apiVersion: APIVersion
+    ) -> ZMTransportRequest? {
+        var action = action
```
netbe, line 30: why do we need to redefine as var here ?

are the .fail methods mutating?

wire-ios-sync-engine/Tests/Source/Calling/WireCallCenterV3Tests.swift
```diff
@@ -68,14 +68,16 @@ class WireCallCenterV3Tests: MessagingTest {
     override func setUp() {
         super.setUp()
 
+        BackendInfo.domain = "wire.com"
```
netbe, line 72: we should avoid changing this directly/alone. change the storage before so it does not affect other tests
```
BackendInfo.storage = UserDefaults(suiteName: UUID().uuidString)!
```

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
```diff
@@ -664,6 +664,40 @@ extension WireCallCenterV3 {
...
+        guard
+            let context = uiMOC,
+            let syncContext = context.zm_sync,
+            let domain = id.domain ?? BackendInfo.domain,
+            let conversation = ZMConversation.fetch(
+                with: id.identifier,
+                domain: domain,
+                in: context
+            ),
+            conversation.messageProtocol == .mls,
```
netbe: should this be in the +MLS extension?

wire-ios-request-strategy/Sources/Request Strategies/MLS/Actions/LeaveSubconversationActionHandler.swift
johnxnguyen: Correct, reporting the results are mutating.

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
johnxnguyen: I refactored the methods a bit and changed the names, also added comments.

wire-ios-sync-engine/Tests/Source/Calling/WireCallCenterV3Tests.swift
johnxnguyen: fixed

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
johnxnguyen: moved

netbe (approved):

David-Henner (approved):

PR #311, opened 2023-06-30
fix: auto-deleting pictures sent from other device of current user WPB-1238

lewandowskit93:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-1238" title="WPB-1238" target="_blank">BugWPB-1238</a>  [iOS] Self-deleting pictures sent from another device are not deleted
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Counter for self-deleting pictures does not start for pictures sent from current user's another device (web/android).
The timer on the pictures was flickering on the UI as well when the cells were repainted.

### Causes

There is a split of logic implemented. Pictures sent by other users undergo other behaviour than pictures sent by current user. When the V3 asset is downloaded, the AssetClientMessage is updated on uiContext which goes the route of deleting the picture. For the current user it should go the route of obfuscation instead of deletion. This behaviour can be triggered only on sync context which for that case is never fired as the `hasDownloadedFile` variable is not synced via CoreData. So since the update is being done on uiContext, all pictures are treated as sent by another user, which is not the case therefore deletion nor obfuscation is never triggered.

### Solutions

The provided fix verifies if the AssetClientMessage was sent by current user or another sender while preparing for firing auto-deletion. This additional check allows for checking `hasDownloadedFile` only for another users and skipping this requirement for current user and triggering auto-obfuscation immediately after receiving a message.
The UI timer flickering was fixed by providing another flag for animation & visibility control.

Needs releases with:

- [ ] GitHub link to other pull request

#### Test Coverage (Optional)

#### How to Test

Send self-deleting picture from web and iOS app. Images sent from current user should obfuscate. Images from other users should auto-delete. 

----

----

----
##### References

### DIFF
--- a/wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
+++ b/wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
@@ -66,11 +66,11 @@ extension ZMAssetClientMessage {
 
     @discardableResult @objc public override func startDestructionIfNeeded() -> Bool {
         let imageNotDownloaded = imageMessageData != nil && !hasDownloadedFile
-        let fileNotUploadedNorUploaded = fileMessageData != nil
+        let fileHasNoUploadState = fileMessageData != nil
             && underlyingMessage?.assetData?.hasUploaded == false
             && underlyingMessage?.assetData?.hasNotUploaded == false
-        let otherSenderCheck = sender == nil || (sender?.isSelfUser == false && managedObjectContext?.zm_isUserInterfaceContext == true)
-        let destructionStartCondition = !(fileNotUploadedNorUploaded || (otherSenderCheck && imageNotDownloaded))
+        let hasOtherUserSender = sender == nil || (sender?.isSelfUser == false && managedObjectContext?.zm_isUserInterfaceContext == true)
+        let destructionStartCondition = !(fileHasNoUploadState || (hasOtherUserSender && imageNotDownloaded))
         if destructionStartCondition {
             return super.startDestructionIfNeeded()
         }
--- a/wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
+++ b/wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
@@ -67,11 +67,11 @@ extension ZMAssetClientMessage {
     @discardableResult @objc public override func startDestructionIfNeeded() -> Bool {
         if let isSelfUser = sender?.isSelfUser {
             // check for download state only for images that were sent by another user
-            if !isSelfUser && managedObjectContext?.zm_userInterface != nil
+            if !isSelfUser && managedObjectContext?.zm_isUserInterfaceContext == true
                 && imageMessageData != nil
                 && !hasDownloadedFile {
                 return false
-            } else if isSelfUser && managedObjectContext?.zm_isSyncContext != nil
+            } else if isSelfUser && managedObjectContext?.zm_isSyncContext == true
                         && fileMessageData != nil
                         && underlyingMessage?.assetData?.hasUploaded == false
                         && underlyingMessage?.assetData?.hasNotUploaded == false {
--- a/wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
+++ b/wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
@@ -69,14 +69,16 @@ extension ZMAssetClientMessage {
         let fileHasNoUploadState = fileMessageData != nil
             && underlyingMessage?.assetData?.hasUploaded == false
             && underlyingMessage?.assetData?.hasNotUploaded == false
-        let hasOtherUserSender = sender == nil || (sender?.isSelfUser == false && managedObjectContext?.zm_isUserInterfaceContext == true)
-        let destructionStartCondition = !(fileHasNoUploadState || (hasOtherUserSender && imageNotDownloaded))
-        if destructionStartCondition {
-            return super.startDestructionIfNeeded()
-        }
-        else {
+        // in super.startDestructionIfNeeded() there is a logic split that triggers destruction or obfuscation depending on sender and context
+        let isOtherSender = sender?.isSelfUser == false && managedObjectContext?.zm_isUserInterfaceContext == true
+        let isNotSelfSender = sender == nil || isOtherSender
+        if fileHasNoUploadState {
+            return false
+        } else if isNotSelfSender && imageNotDownloaded {
+            // we do not destroy images sent by other user that are not downloaded yet, it is synced after asset is downloaded
             return false
         }
+        return super.startDestructionIfNeeded()
     }
 
     /// Extends the destruction timer to the given date, which must be later
--- a/wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
+++ b/wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
@@ -65,15 +65,27 @@ extension ZMAssetClientMessage {
     }
 
     @discardableResult @objc public override func startDestructionIfNeeded() -> Bool {
-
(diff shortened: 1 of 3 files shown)


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/EphemeralCountdownView.swift
```diff
@@ -25,6 +25,7 @@ class EphemeralCountdownView: UIView {
     fileprivate let destructionCountdownView: DestructionCountdownView = DestructionCountdownView()
     fileprivate let containerView =  UIView()
     fileprivate var timer: Timer?
+    fileprivate var everAnimated = false
```
netbe: I don't get the naming, do you mean alwaysAnimated ?

wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
```diff
@@ -65,15 +65,26 @@ extension ZMAssetClientMessage {
...
+                    return false
+            }
+        } else {
+            if imageMessageData != nil && !hasDownloadedFile {
+                return false
+            } else if fileMessageData != nil
+                && underlyingMessage?.assetData?.hasUploaded == false
+                && underlyingMessage?.assetData?.hasNotUploaded == false {
+                return false
+            }
```
netbe: I wonder if we can improve this if else return false...

Like store the conditions to more meaningful constant name and if/else at the end on them

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/EphemeralCountdownView.swift
lewandowskit93: It holds true if the animation was already fired once for this view.

wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
```diff
@@ -65,27 +65,18 @@ extension ZMAssetClientMessage {
...
-        } else {
-            if imageMessageData != nil && !hasDownloadedFile {
-                return false
-            } else if fileMessageData != nil
-                && underlyingMessage?.assetData?.hasUploaded == false
-                && underlyingMessage?.assetData?.hasNotUploaded == false {
-                return false
-            }
+        let imageNotDownloaded = imageMessageData != nil && !hasDownloadedFile
+        let fileNotUploadedNorUploaded = fileMessageData != nil
```
netbe: fileNotUploadedNorUploaded 🤔 so fileNotUploaded period or is there more to it?

lewandowskit93: Thats a weird condition because it checks whether the file has been uploaded, but it also checks if file was has not been uploaded.

netbe: what other state is there then ?

lewandowskit93: It's related to `status` field behind. It can also be `nil`, so perhaps it refers to this state.

netbe: what about `fileHasNoUploadedState` ?

lewandowskit93: It's a better name I think, good idea :)

wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
```diff
@@ -65,16 +65,18 @@ extension ZMAssetClientMessage {
...
-
-        if imageMessageData != nil && !hasDownloadedFile {
-            return false
-        } else if fileMessageData != nil
+        let imageNotDownloaded = imageMessageData != nil && !hasDownloadedFile
+        let fileNotUploadedNorUploaded = fileMessageData != nil
             && underlyingMessage?.assetData?.hasUploaded == false
-            && underlyingMessage?.assetData?.hasNotUploaded == false {
+            && underlyingMessage?.assetData?.hasNotUploaded == false
+        let otherSenderCheck = sender == nil || (sender?.isSelfUser == false && managedObjectContext?.zm_isUserInterfaceContext == true)
```
netbe: while we're at it, can we improve the name of this one?

wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
```diff
@@ -65,16 +65,18 @@ extension ZMAssetClientMessage {
...
-
-        if imageMessageData != nil && !hasDownloadedFile {
-            return false
-        } else if fileMessageData != nil
+        let imageNotDownloaded = imageMessageData != nil && !hasDownloadedFile
+        let fileHasNoUploadState = fileMessageData != nil
             && underlyingMessage?.assetData?.hasUploaded == false
-            && underlyingMessage?.assetData?.hasNotUploaded == false {
+            && underlyingMessage?.assetData?.hasNotUploaded == false
+        let hasOtherUserSender = sender == nil || (sender?.isSelfUser == false && managedObjectContext?.zm_isUserInterfaceContext == true)
```
johnxnguyen: What's the significance of the view context here? Maybe a comment is needed.

wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
lewandowskit93: Changed :)

netbe (approved):

wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
```diff
@@ -65,16 +65,18 @@ extension ZMAssetClientMessage {
...
+        let destructionStartCondition = !(fileHasNoUploadState || (hasOtherUserSender && imageNotDownloaded))
+        if destructionStartCondition {
+            return super.startDestructionIfNeeded()
+        }
```
johnxnguyen: This is quite tricky to read, perhaps we can break it up a bit more:

```swift
if selfIsSender && !isAssetUploaded {
  return false
}

if !selfIsSender && !isAssetDownloaded {
  return false
}

return super.startDestructionIfNeede()

```

wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
lewandowskit93: As described in PR description, the context splits the logic into obfuscation or deletion depending on the context.
But this split happens in the super call. Some logic can be triggered only on given contexts and not the other ones.

wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
lewandowskit93: It's not the same logic. Its checking the other sender instead not the self, Self sender logic was broken. Also the file upload state is checked regardless of the sender.

johnxnguyen: Sure, but could it still be rewritten in a clearer way? I'm still struggling to understand condition.

wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
johnxnguyen: It was still a bit confusing but after checking the source it makes more sense. Still I suggest to make it clearer using names or a comment.

johnxnguyen (approved):

PR #326, opened 2023-07-11
chore: end call when failing to setup mls conference WPB-3027

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-3027" title="WPB-3027" target="_blank">BugWPB-3027</a>  [iOS] MLS conference call: Initiating a video mls conference call doesn't enable the video 
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?
Setting up an MLS conference can fail at many different points. If it does fail, then the call will appear to be connected but will not work properly. For instance, the video may be enabled but the receiver side doesn't see it. 

To avoid misleading the user to think the call connected successfully, it would be better to end the call if the mls conference couldn't be successfully setup. This way, if there is a problem with the call, then have already narrowed down the problem space significantly.

----

----

----
##### References

### DIFF
--- a/wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
+++ b/wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
@@ -591,12 +591,16 @@ extension WireCallCenterV3 {
     }
 
     private func setUpMLSConference(in conversation: ZMConversation) throws {
+        guard let conversationID = conversation.avsIdentifier else {
+            throw Failure.failedToSetupMLSConference
+        }
+
         guard
-            let conversationID = conversation.avsIdentifier,
             let parentQualifiedID = conversation.qualifiedID,
             let parentGroupID = conversation.mlsGroupID,
             let syncContext = conversation.managedObjectContext?.zm_sync
         else {
+            onMLSConferenceFailure(id: conversationID)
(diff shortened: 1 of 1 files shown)


agisilaos (approved):

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
```diff
@@ -591,12 +591,16 @@ extension WireCallCenterV3 {
     }
 
     private func setUpMLSConference(in conversation: ZMConversation) throws {
+        guard let conversationID = conversation.avsIdentifier else {
```
netbe, line 594: just to double check: Is it possible to end up here, and the conversation has no avsIdentifier ?
Could it be we have an ongoing call and the avsIdentifier is not saved properly in the conversation?

johnxnguyen: The existence of the avs identifier is dependent on the existence of the `remoteIdentifier`... which I think is expected to be there, so in theory we should always be able to get the avs id. Kind of annoying that it appears optional, we could try to change this in another PR.

netbe (approved):

PR #338, opened 2023-07-14
fix: Tapping on the deleted message makes the display name animate to the side and back - WPB-1955

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-1955" title="WPB-1955" target="_blank">BugWPB-1955</a>  [iOS] Tapping on the deleted message makes the display name animate to the side and back
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR I fix the issue where tapping on the deleted message would make the display namer animate to the side and back. 

By calling layoutIfNeeded() in the configure method we fix the problem. I also added a bit of documentation why we need to call the method there. 

----
##### References


netbe (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/CellComponents/SenderCellComponent.swift
```diff
@@ -80,6 +80,11 @@ final class SenderCellComponent: UIView {
            let userSession = ZMUserSession.shared() {
             observerToken = UserChangeInfo.add(observer: self, for: user, in: userSession)
         }
+
+        // We need to call that method here to restraint the authorLabel moving
+        // outside of the view and then back to its position. For more information
+        // check the ticket: https://wearezeta.atlassian.net/browse/WPB-1955
+        self.layoutIfNeeded()
```
lewandowskit93: I would think about moving this call into `configureViews` method, I think it fits better there as an immediate action related to views. What do you think?

agisilaos: Done!

lewandowskit93 (approved):

PR #349, opened 2023-07-20
fix: can't send messages WPB-3106

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-3106" title="WPB-3106" target="_blank">BugWPB-3106</a>  [iOS] Can't send messages
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Given
-  My app is in the background
- I'm in a conversation with another user "Bob"

When
- Bob logs in with a new client and sends a message in our conversation from the new client

Then
- When I open the app, I see Bob's new message, but if I try to send a message then it fails

### Causes
I think what is happening is that because the self user receives a new message from a new client while the app is in the background, the notification service extension will fetch the update event and create a new `UserClient` instance for the newly discovered client. New clients have `needsToBeUpdatedFromBackend` set to `true` by default.

When the app is woken and the self user wishes to send a message, we first detect that the message is dependent on fetching the new client from the backend. However, the request to sync the client is never generated, therefore the message is blocked by the dependent object and never gets sent.

The `FetchingClientRequestStrategy` is responsible for syncing user clients. It tracks changes to `UserClient` objects, particularly those that have `needsToBeUpdatedFromBackend` set to `true`. But this only happens when the object **changes** and those changes have been saved. Since the new client was created in the NSE (another process), the main app doesn't get notified of this (also because it's in the background) and so the `FetchingClientRequestStrategy` doesn't attempt to sync the client.

### Solutions

When `FetchingClientRequestStrategy` is asked for a request, check if there are any user clients that need to be updated from the backend, and sync those clients in order to generate a request.

#### Test Coverage

- Unit test asserting that clients are synced (not via changed object tracking)

#### How to Test

- User A logs in with Client A1 on iOS
- User B logs in with Client B1 on any platform
- User A and B exchange messages in a conversation C
- User A locks phone (i.e puts app in BG)
- User B logs in with new Client B2
- User B sends a message from B2 into conversation C
- User A observes that they do not see a push notification.
- User A opens the app, sees the new message, tries to send a message

The message should send successfully.

### Notes

There were some failing tests which I have fixed here as well.

----

----

----
##### References

### DIFF
--- a/wire-ios-request-strategy/Sources/Request Strategies/User Clients/FetchingClientRequestStrategy.swift
+++ b/wire-ios-request-strategy/Sources/Request Strategies/User Clients/FetchingClientRequestStrategy.swift
@@ -110,11 +110,20 @@ public final class FetchingClientRequestStrategy: AbstractRequestStrategy {
     }
 
     public override func nextRequestIfAllowed(for apiVersion: APIVersion) -> ZMTransportRequest? {
+        // There may exist some clients that need an update, so try to sync any before asking
+        // for requests.
+        syncClientsNeedingUpdateIfNeeded()
+
         return
             userClientsByUserClientID.nextRequest(for: apiVersion) ??
             userClientsByUserID.nextRequest(for: apiVersion) ??
             userClientsByQualifiedUserID.nextRequest(for: apiVersion) ??
             entitySync.nextRequest(for: apiVersion)
+    }
+
+    private func syncClientsNeedingUpdateIfNeeded() {
+        let clients = UserClient.fetchClientsNeedingUpdateFromBackend(in: managedObjectContext)
+        fetch(userClients: clients)
     }
 
 }
--- a/setup.sh
+++ b/setup.sh
@@ -69,10 +69,10 @@ else
 fi 
 echo ""
 
-echo "ℹ️  Installing bundler and Ruby dependencies..."
-which bundle || gem install bundler
-bundle check || bundle install
-echo ""
+#echo "ℹ️  Installing bundler and Ruby dependencies..."
+#which bundle || gem install bundler
+#bundle check || bundle install
+#echo ""
 
 echo "ℹ️  Downloading additional assets..."
 scripts/download-assets.sh "$@"
--- a/setup.sh
+++ b/setup.sh
@@ -69,10 +69,10 @@ else
 fi 
 echo ""
 
-#echo "ℹ️  Installing bundler and Ruby dependencies..."
-#which bundle || gem install bundler
-#bundle check || bundle install
-#echo ""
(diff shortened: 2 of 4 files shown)


agisilaos (approved):

wire-ios-request-strategy/Sources/Request Strategies/User Clients/FetchingClientRequestStrategy.swift
```diff
@@ -110,13 +110,22 @@ public final class FetchingClientRequestStrategy: AbstractRequestStrategy {
...
         return
             userClientsByUserClientID.nextRequest(for: apiVersion) ??
             userClientsByUserID.nextRequest(for: apiVersion) ??
             userClientsByQualifiedUserID.nextRequest(for: apiVersion) ??
             entitySync.nextRequest(for: apiVersion)
     }
 
+    private func syncClientsNeedingUpdateIfNeeded() {
+        let clients = UserClient.fetchClientsNeedingUpdateFromBackend(in: managedObjectContext)
+        fetch(userClients: clients)
```
netbe, line 126: I assume this does the syncing but this is async, right, we don't wait for it to be finished?
Is that an issue?

johnxnguyen: It extracts the client ids and enqueues them the appropriate "sync" object, which is what generates a request. For instance, if I add client A to the `useClientsByClientID` sync object, then if we ask it for a next request it will produce one.

johnxnguyen: The flow is something like:

- generate request for new message
- new message is dependent on new client
- generate request for new client
- process response for new client
- generate request for new message (succeeds this time because new message has no dependencies anymore)
- process response for new message

netbe: Do we get a response or an error code from the first generated request in your flow?

johnxnguyen: No, actually no request is generated. When we ask to generate a request (from the dependency sync) it will not return one because the dependency needs to be dealt with first.

netbe: ok cool, so the order is respected, thks

netbe (approved):

PR #361, opened 2023-07-25
fix: There is no pen icon on the preview of an edited message someone is replying to - WPB-3494

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-3494" title="WPB-3494" target="_blank">BugWPB-3494</a>  [iOS] There is no pen icon on the preview of an edited message someone is replying to
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we show a the edit icon when someone is replying to a message which has been edited. The problem was that instead of modifying the insets we're adding an empty string + our text attachment. For some reason our icon was still hidden and that solution wasn't working as expected. 

So now, instead of an empty string we add the proper left inset and the issue has been fixed. 

| BEFORE | AFTER |
|---|---|
| Icon-Edited | Screenshot 2023-07-25 at 17 28 46  |

----
##### References

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Components/MessagePreviewView.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Components/MessagePreviewView.swift
@@ -62,6 +62,7 @@ extension UITextView {
 
 final class MessageThumbnailPreviewView: UIView {
     private let senderLabel = UILabel()
+    private var leftEditIconInset: CGFloat = 10
     private let contentTextView = UITextView.previewTextView()
     private let imagePreview = ImageResourceView()
     private var observerToken: Any?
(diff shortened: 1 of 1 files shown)


KaterinaWire (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Components/MessagePreviewView.swift
```diff
@@ -144,7 +145,15 @@ final class MessageThumbnailPreviewView: UIView {
 
     private func editIcon() -> NSAttributedString {
         if message.updatedAt != nil {
-            return "  " + NSAttributedString(attachment: NSTextAttachment.textAttachment(for: .pencil, with: iconColor, iconSize: 8))
+            return NSAttributedString(
+                attachment: NSTextAttachment.textAttachment(for: .pencil,
+                with: iconColor,
```
SzymonMrozek: is that intended formatting?

SzymonMrozek (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Components/MessagePreviewView.swift
agisilaos: I think so, otherwise, the line would be too long. @SzymonMrozek Any specific suggestions on how it should be?

SzymonMrozek: I've been used to keep indentation of the same level arguments, in your case it's mixed

I would suggest something like this for this initializer: 
```swift
return NSAttributedString(
  attachment: NSTextAttachment.textAttachment(
    for: .pencil,
    with: iconColor,
    iconSize: 8,
    insets: UIEdgeInsets(
      top: 0,
      left: leftEditIconInset,
      bottom: 0,
      right: 0
    )
  )
)
```

agisilaos: Done!

SzymonMrozek (approved):

PR #372, opened 2023-07-28
fix: Transition when presenting a `ProfileViewController`

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

In a case where we wanted to tap on someone's profile (to see the profile) and to connect, there was that weird transition that was happening because of `TransitionDelegate`. `TransitionDelegate` was the source of the problem. And it is not needed anymore, so it's being removed. 

In addition to that `ZoomTransition` is getting removed as well since we were using it only on `TransitionDelegate`. 

Last but not least we replace `viewDidAppear` with `viewWillApear` so we can remove the delay in which the x icon is shown when we present the view

| BEFORE | AFTER |
|:-:|:-:|
| https://github.com/wireapp/wire-ios-mono/assets/10944108/1ba6e79b-5804-4d80-a301-c3401764ec81 | https://github.com/wireapp/wire-ios-mono/assets/10944108/63513949-c63f-4d9a-ae8f-cdd88e7c5f4a |

----
##### References


SzymonMrozek (approved):

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController.swift
```diff
@@ -179,8 +177,8 @@ final class ProfileViewController: UIViewController {
         updateShowVerifiedShield()
     }
 
-    override func viewDidAppear(_ animated: Bool) {
-        super.viewDidAppear(animated)
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
```
johnxnguyen, line 181: > Last but not least we replace viewDidAppear with viewWillApear so we can be the delay in which the x icon is shown when we present the view

I don't understand what you mean here, could you elaborate?

agisilaos: Yes, so before that change there was a delay between presenting the view and showing the navigation bar buttons (take a look at the after video) But after making that change, there's no delay and the icons show when they should in the first place.

lewandowskit93 (approved):

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController.swift
agisilaos: @johnxnguyen Any more questions on this?

johnxnguyen: Is it fine to setup the navigation items repeatedly?

agisilaos: I think that's a different discussion and indeed I don't like that approach. I kid you not, I was just thinking that we need a plan to improve how we handle navigation, navigations controllers and in extent items. I think we had a brief discussion on this during the accessibility sprint. 

Happy to further discuss it and come up with a plan.

johnxnguyen (approved):

PR #374, opened 2023-07-31
fix: Update the color of the mention in Conversation View to match the design specs

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

After merging #366 I realized that the colors for the mentions in conversation view aren't according to design specs and in sync with android. After a quick lookup on the android files and a sync with an android dev I updated the colors so the platforms to be in sync and according to design specs. I updated the tests as well. 

Design specs can be found [here](https://www.figma.com/file/UgKehvwIbiIiQ2CCEdsVuf/Android?type=design&node-id=2446-155165&mode=design&t=CpfFdGI3vvPioPFn-0)

----
##### References


wire-ios/Wire-iOS Tests/ConversationCell/TextMessageMentionsTests.swift
```diff
@@ -16,27 +16,35 @@
...
         super.setUp()
-
         otherUser = MockUserType.createUser(name: "Bruno")
         selfUser = MockUserType.createDefaultSelfUser()
-
+        FontScheme.configure(with: .large)
         UIColor.setAccentOverride(.vividRed)
     }
 
+    // MARK: - tearDown
```
SzymonMrozek, line 46: Side question - does anyone use these markers?

wire-ios/Wire-iOS Tests/ConversationCell/TextMessageMentionsTests.swift
```diff
@@ -16,27 +16,35 @@
...
 
+    // MARK: - setUp
+
     override func setUp() {
         super.setUp()
-
         otherUser = MockUserType.createUser(name: "Bruno")
         selfUser = MockUserType.createDefaultSelfUser()
-
+        FontScheme.configure(with: .large)
```
SzymonMrozek, line 42: Hm... Should this change be here? I don't see any font change in snapshots diff

SzymonMrozek (approved):

wire-ios/Wire-iOS Tests/ConversationCell/TextMessageMentionsTests.swift
agisilaos: I do, usually when searching through a test file or a class.

wire-ios/Wire-iOS Tests/ConversationCell/TextMessageMentionsTests.swift
agisilaos: The change needs to be here because if you try to run the test individually it will crash. The reason why is that it would search for a font that it wouldn't find (aka nil). This plans to change in the future as I'm going to change how we're handling fonts within the app.

johnxnguyen (commented):
I don't see in the design spec where it says the mention highlight should be the accent color. Could you help me find it?

agisilaos:
> I don't see in the design spec where it says the mention highlight should be the accent color. Could you help me find it?

You can find it in the design specs (Android) [here](https://www.figma.com/file/uuHT6JYRCLpq7JUbcFUFqt/Mobile-UI-(Android)?type=design&node-id=611-3513&mode=design&t=Bl8LzGtyOJeXKsTB-0) @johnxnguyen 

Web does it according to the accentColor and Android will implement it as well. The blue example on the design specs is just one of the accent colors.

johnxnguyen (approved):

agisilaos:
closed in favor of #382 which is already merged.

PR #375, opened 2023-07-31
feat: indicate participant bad connection WPB-530

MarcinR:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-530" title="WPB-530" target="_blank">StoryWPB-530</a>  [iOS]  “Indicate users with bad / lost connection during the call“
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?
App will indicate when during call participant is connecting or have bad connection 
image


johnxnguyen (approved):

wire-ios/Wire-iOS/Sources/Helpers/SemanticColors.swift
```diff
@@ -55,6 +55,8 @@ public enum SemanticColors {
         static let textTabBar = UIColor(light: Asset.Colors.gray70, dark: Asset.Colors.gray60)
         static let textFieldFloatingLabel = UIColor(light: Asset.Colors.gray80, dark: Asset.Colors.gray50)
         static let textSecurityEnabled = UIColor(light: Asset.Colors.green500Light, dark: Asset.Colors.green500Dark)
+        static let textInactive = UIColor(light: Asset.Colors.gray60, dark: Asset.Colors.gray70)
```
agisilaos, line 58: Can we rename this property to something more descriptive? textInactive isn't clear enough to where it's being used.

agisilaos (approved):
Approved, but I also left a comment on renaming a color since it's not clear where it's being used. 

Can we rename that property before merging?

wire-ios/Wire-iOS/Sources/Helpers/SemanticColors.swift
johnxnguyen: Can you suggest a name @agisilaos ?

agisilaos: @johnxnguyen How about `textParticipantConnecting`?

johnxnguyen: I would argue that `textInactive` is better. It suggests that it will be used for labels that are inactive. Maybe 
`textDimmed` could also be good. 

Being more descriptive also makes it more restrictive because now this color could only be used for a single purpose (for connecting participants). In my opinion that isn't semantic coloring (well defined semantics that are reusable), it's just naming colors (single purpose colors).

agisilaos: Sounds good! I think it's fair enough reasoning to keep it as it is for now. Good to go from my side. 👍

PR #395, opened 2023-08-10
feat: remove stale participants from subconversation WPB-637

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-637" title="WPB-637" target="_blank">StoryWPB-637</a>  [iOS] Remove stale clients from subgroup
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we ensure stale participants are removed from an MLS conference's corresponding subconversation if they do not reconnect to the call after a given timeout.

### Changes

I've created a `MLSConferenceStaleParticipantsRemover` class that handles changes in participants states. Whenever a participant becomes `unconnected` to the call: a timer starts. When the timer expires: the participant is removed from the MLS subconversation. If the stale participant manages to reconnect to the call, the timer is canceled. 

I'm leveraging Combine to subscribe to changes in the participants list. I've also created a `TimerManager` class to keep 
track of multiple timers associated to stale participants.

There is also some logic taking care of some edges cases where the self user leaves the call while being the last connected participant (either by user interaction, or as a result of the call terminating). In such cases, the self user would remove any participants scheduled for removal without waiting for the removal timeout to expire. This needs to be done because the self user leaves the subconversation once the call ends, thus they wouldn't be able to remove members if the removal timers expire after they left.

#### Test Coverage (Optional)


wire-ios-sync-engine/Source/Calling/MLSConferenceStaleParticipantsRemover.swift
```diff
@@ -0,0 +1,218 @@
...
+    private func subconversationMembers(for groupID: MLSGroupID) -> [MLSClientID]? {
+        do {
+            return try mlsService.subconversationMembers(for: groupID)
+        } catch {
+            logger.warn("failed to fetch subconversation members: \(String(describing: error))")
+            return nil
+        }
+    }
+
+    private func remove(
```
johnxnguyen: Suggestion: `enqueueRemove` to make it even more clearer that the remove will happen later.

wire-ios-sync-engine/Source/Calling/MLSConferenceStaleParticipantsRemover.swift
```diff
@@ -0,0 +1,218 @@
...
+                        client: clientID,
+                        from: groupID
+                    )
+                }
+            )
+
+            logger.info("started timer for removal of stale participant (clientdID: \(clientID), groupID: \(groupID))")
+        } catch {
+            // timer already exists, do nothing
+        }
```
johnxnguyen: Is there a better way to check that we already have a timer?

wire-ios-sync-engine/Tests/Source/Calling/MLSConferenceStaleParticipantsRemoverTests.swift
```diff
@@ -0,0 +1,270 @@
...
+            )
+        )
+
+        participants[0].updateState(connectedState)
+        participants[1].updateState(connectedState)
+
+        let expectations = expectations(from: participants)
+        mlsService.mockRemoveMembersFromConversation = { clientIDs, _ in
+            guard let id = clientIDs.first else { return }
+            expectations[id]?.fulfill()
```
johnxnguyen, line 118: Why do we expect the members to be removed when the test name suggests otherwise?

wire-ios-sync-engine/Tests/Source/Calling/MLSConferenceStaleParticipantsRemoverTests.swift
```diff
@@ -0,0 +1,270 @@
...
+        ]
+
+        // mock subconversation members
+        mlsService.mockSubconversationMembers = { _ in
+            return []
+        }
+
+        // set expectations
+        let expectation = XCTestExpectation()
+        expectation.isInverted = true
```
johnxnguyen, line 148: Interesting, I didn't know about this. So inverting it means we expect it not to be fulfilled within a certain timeout?

wire-ios-sync-engine/Tests/Source/Calling/MLSConferenceStaleParticipantsRemoverTests.swift
johnxnguyen: I see now it's configured in the helper method

wire-ios-sync-engine/Source/Calling/MLSConferenceStaleParticipantsRemover.swift
johnxnguyen: I see that the error thrown here is `timerAlreadyExists`, I suggest we catch it explicitly.

wire-ios-sync-engine/Tests/Source/Calling/MLSConferenceStaleParticipantsRemoverTests.swift
David-Henner: Yep!

wire-ios-sync-engine/Tests/Source/Calling/MLSConferenceStaleParticipantsRemoverTests.swift
David-Henner: We expect the participants that reconnected to not be removed. The helper method `expectations(from: participants)` creates expectations. For the participants that should not be removed, the expectations will be inverted. If `removeMembersFromConversation` is called for participants that shouldn't be removed, the expectations will be fulfilled and it will result in the test failing.

wire-ios-sync-engine/Source/Calling/MLSConferenceStaleParticipantsRemover.swift
David-Henner: done ✅

wire-ios-sync-engine/Source/Calling/MLSConferenceStaleParticipantsRemover.swift
David-Henner: done ✅

johnxnguyen (approved):

wire-ios-sync-engine/Source/Calling/MLSConferenceStaleParticipantsRemover.swift
```diff
@@ -0,0 +1,224 @@
...
+    private let removalTimeout: TimeInterval
+    private let mlsService: MLSServiceInterface
+    private let context: NSManagedObjectContext
+    private var previousInput: MLSConferenceParticipantsInfo?
+
+    private typealias TimerError = TimerManager<MLSClientID>.TimerError
+
+    // MARK: - Life cycle
+
+    init(mlsService: MLSServiceInterface, context: NSManagedObjectContext, removalTimeout: TimeInterval = 190) {
```
netbe: is the removalTimeout agreed across clients? suprised about the value, did you mean 180?

wire-ios-sync-engine/Source/Calling/MLSConferenceStaleParticipantsRemover.swift
```diff
@@ -0,0 +1,224 @@
...
+        }
+    }
+}
+
+private extension Array where Element == CallParticipant {
+
+    func excludingSelf(in context: NSManagedObjectContext) -> Self {
+        var selfUserID: AVSIdentifier!
+
+        context.performAndWait {
```
netbe, line 206: on which thread is this executed?

wire-ios-sync-engine/Tests/Source/Calling/MLSConferenceStaleParticipantsRemoverTests.swift
```diff
@@ -0,0 +1,270 @@
...
+        super.setUp()
+        mlsService = MockMLSService()
+        sut = MLSConferenceStaleParticipantsRemover(
+            mlsService: mlsService,
+            context: uiMOC,
+            removalTimeout: 0.4
+        )
+
+        let selfUser = ZMUser.selfUser(in: uiMOC)
+        selfUser.remoteIdentifier = UUID()
```
netbe, line 45: why do we do this?

does this need to be reverted at teardown?

wire-ios-sync-engine/Source/Calling/MLSConferenceStaleParticipantsRemover.swift
David-Henner: Yes, it's an agreed value of 190 seconds

wire-ios-sync-engine/Source/Calling/MLSConferenceStaleParticipantsRemover.swift
David-Henner: `excludingSelf` is called from the main thread and `context` is the sync context, so it will be executed there (on the sync context)

netbe: so blocking the mainThread with performAndWait

wire-ios-sync-engine/Source/Calling/MLSConferenceStaleParticipantsRemover.swift
netbe: ok can you put it as a constant then?

wire-ios-sync-engine/Tests/Source/Calling/MLSConferenceStaleParticipantsRemoverTests.swift
David-Henner: > why do we do this?

When the sut calls `excludingSelf`, it needs the self user to be set 

> does this need to be reverted at teardown?

We already set the self user in the same way in a myriad of other tests without tearing it down, so it's safe to leave it as it is here. Although ideally we probably shouldn't

wire-ios-sync-engine/Source/Calling/MLSConferenceStaleParticipantsRemover.swift
David-Henner: good catch, I missed it. I'll update soon

wire-ios-sync-engine/Tests/Source/Calling/MLSConferenceStaleParticipantsRemoverTests.swift
netbe: right I am sure we do it elsewhere, just wondering if it can be a cause of flakyness.

the selfUser has no remoteIdentifier by default ? does that normally come from sync ?

wire-ios-sync-engine/Source/Calling/MLSConferenceStaleParticipantsRemover.swift
David-Henner: I could but having it as a parameter is useful for injecting the value for tests

netbe: you can have both, just declare the constant and use it as default param

David-Henner: right, will do

wire-ios-sync-engine/Tests/Source/Calling/MLSConferenceStaleParticipantsRemoverTests.swift
David-Henner: I don't know

wire-ios-sync-engine/Source/Calling/MLSConferenceStaleParticipantsRemover.swift
David-Henner: done [6a14183](https://github.com/wireapp/wire-ios-mono/pull/395/commits/6a1418373b73eb0b753ec7b683b640adde9ee17e)

wire-ios-sync-engine/Source/Calling/MLSConferenceStaleParticipantsRemover.swift
David-Henner: done [6a14183](https://github.com/wireapp/wire-ios-mono/pull/395/commits/6a1418373b73eb0b753ec7b683b640adde9ee17e)

netbe (approved):

wire-ios-sync-engine/Source/UserSession/URLActionProcessors/DeepLinkURLActionProcessor.swift
```diff
@@ -58,31 +61,38 @@ class DeepLinkURLActionProcessor: URLActionProcessor {
...
+                                transportSession: strongSelf.transportSession,
+                                eventProcessor: strongSelf.eventProcessor,
+                                contextProvider: strongSelf.contextProvider
+                            ) { [weak self] (response) in
+
+                                guard let strongSelf = self else { return }
+
                                 switch response {
                                 case .success(let conversation):
                                     strongSelf.synchronise(conversation) { result in
```
David-Henner, line 82: @netbe this were self was retained

PR #415, opened 2023-08-18
fix: Sorting of reactions based on count and then name - WPB-3937

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-3937" title="WPB-3937" target="_blank">BugWPB-3937</a>  [iOS] [Reactions] Reactions under the message they're not sorted based on the name if two reactions have the same count
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we fix an issue where reactions were only sorted by the count. But when the count between reactions is the same they need to be sorted by name. 

Fixes this [issue](https://wearezeta.atlassian.net/jira/software/c/projects/WPB/boards/149?assignee=6170266af6da6a006a990919&selectedIssue=WPB-3945) as well

----
##### References


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/Reactions/MessageReactionsCellDescription.swift
```diff
@@ -44,7 +44,11 @@ final class MessageReactionsCellDescription: ConversationMessageCellDescription
                 performReaction: nil
             )
         }.sorted {
-            return $0.count > $1.count
+            if $0.count == $1.count {
+                return $0.emoji.name < $1.emoji.name
```
johnxnguyen: I think it would be better to use lexicographical ordering rather than strictly alphabetical ordering, because otherwise all words starting with a capital would precede words without capitals.

johnxnguyen: Is there a test covering the sorting?

agisilaos: > I think it would be better to use lexicographical ordering rather than strictly alphabetical ordering, because otherwise all words starting with a capital would precede words without capitals.

Here we're talking about the reactions **below the message**. We don't capitalize any letter so I don't think that would be a problem. 

Regarding the reactions in **Message Details** we capitalize the first letter of all reaction names in every section so that's by default. I don't think we will have a problem there either. Please let me know if you think otherwise.

agisilaos: > Is there a test covering the sorting?

I don't think we have a test for that. @johnxnguyen but I can add one.

KaterinaWire: can you use `.sortedByCountThenName()` here?

agisilaos: @KaterinaWire no!

johnxnguyen: I think you can, the whole point was that we could define the sort once and reuse it, it will just take a bit of type massaging.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Message Details/MessageDetailsDataSource.swift
```diff
@@ -165,16 +165,15 @@ final class MessageDetailsDataSource: NSObject, ZMMessageObserver, ZMUserObserve
...
+        }.filter { _ , count in
+            return !count.isEmpty
         }
-
-        self.reactions.sort(by: { $0.items.count > $1.items.count })
+        .sortedByCountThenName()
+        .map({ (emoji, users) in
+            MessageDetailsSectionDescription(headerText: "\(emoji.value) \(emoji.name?.capitalizingFirstCharacterOnly ?? "") (\(users.count))",
+                                             items: MessageDetailsCellDescription.makeReactionCells(users))
+        })
```
johnxnguyen: This is a lot of sequence transformations which can be expensive if there are a lot of reactions. Here are some ideas to make it more efficient.

By using `lazy` we can merge two transformations (filter and map) into one:

```swift
func sortedReactions() -> [(Emoji, [UserType])] {
    return usersReaction.lazy
        .filter { _, users in !users.isEmpty }
        .map { key, users in (Emoji(value: key), users) }
        .sortedByCountThenName()
}
```

Then you can do additional maps after than, such as to make the section description:

```swift
message.sortedReactions().map { emoji, users in
    ...
}
```

You can also use `sortedReactions` in other places wherever you have the message.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/Reactions/MessageReactionsCellDescription.swift
johnxnguyen: With code from my other comment you could do something like:

```swift
let reactions = [MessageReactionMetadata] = message.sortedReactions().map { emoji, users in
    MessageReactionMetadata(
        emoji: emoji,
        count: Int(users.count),
        ...
    )
}
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Message Details/MessageDetailsDataSource.swift
agisilaos: Done! Thank you for the suggestions @johnxnguyen!

KaterinaWire (approved):

johnxnguyen (changes requested):
Can you add a test for the `MessageReactionCellDescription`  too?

johnxnguyen:
@agisilaos can we close this since the fix is in https://github.com/wireapp/wire-ios-mono/pull/464 ?

agisilaos:
> @agisilaos can we close this since the fix is in https://github.com/wireapp/wire-ios-mono/pull/464 ?

We can close this.

PR #429, opened 2023-08-26
feat: recover from quick sync 404 WBP-2143

David-Henner:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

If a client has been inactive for a long time, it may miss messages in a group, which could result in the group’s epoch advancing and diverging from the clients local epoch. In this state no messages can be decrypted or encrypted.

This PR handles the case where the user's event queue has been deleted by the backend, which is detected during quick sync (by receiving a 404 response when fetching missed events). 

In this scenario, a slow sync will be triggered, during which the user's conversations will be refetched from the backend. When the sync finishes, we can go through the conversations and compare their remote epoch with their local epoch. If they differ, we'll rejoin the conversation.

### Changes

- Added logic in `MLSService` to repair local conversations if needed
- Triggered conversation repair after the end of the slow/quick sync
- Added developer action to force a 404 response during quick sync

### Testing


wire-ios-data-model/Tests/MLS/MLSServiceTests.swift
```diff
@@ -1086,6 +1086,70 @@ class MLSServiceTests: ZMConversationTestsBase, MLSServiceDelegate {
         XCTAssertEqual(groupInfoInvocations.count, 0)
     }
 
+    // MARK: - Handling out of sync conversations
+
+    func test_RepairOutOfSyncConversations_RejoinsOutOfSyncConversation() {
```
David-Henner:
```suggestion
    func test_RepairOutOfSyncConversations_RejoinsOutOfSyncConversations() {
```

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -815,6 +817,60 @@ public final class MLSService: MLSServiceInterface {
...
+
+    private func outOfSyncConversations(in context: NSManagedObjectContext) -> [ZMConversation] {
+        return coreCrypto.perform { coreCrypto in
+            return ZMConversation.fetchMLSConversations(in: context).filter {
+                isOutOfSync(conversation: $0, coreCrypto: coreCrypto)
+            }
+        }
+    }
+
+    private func isOutOfSync(
```
johnxnguyen: Suggestion 

```swift
extension CoreCryptoProtocol {

    func isConversationOutOfSync(_ conversation: ZMConversation) -> Bool {
        ....
    }

}
```

wire-ios-data-model/Tests/MLS/MLSServiceTests.swift
```diff
@@ -1086,6 +1086,70 @@ class MLSServiceTests: ZMConversationTestsBase, MLSServiceDelegate {
...
+            createConversation(outOfSync: false)
+        ]
+
+        // mock conversation epoch
+        mockCoreCrypto.mockConversationEpoch = { groupID in
+            guard let tuple = conversationAndOutOfSyncTuples.first(
+                where: { $0.conversation.mlsGroupID?.bytes == groupID }
+            ) else {
+                return 1
+            }
```
johnxnguyen, line 1105: I think it would be easier to read if the conversations were in a dictionary mapping the group id to the conversation.

David-Henner: I agree but I think it would make the setup too complex

David-Henner:
```swift
        // ...

        let conversationAndOutOfSyncTuples = [
            createConversation(outOfSync: true),
            createConversation(outOfSync: true),
            createConversation(outOfSync: false)
        ]

        let conversationAndOutOfSyncTuplesByGroupID = try conversationAndOutOfSyncTuples.reduce(
            into: [MLSGroupID: ConversationAndOutOfSyncTuple]()
        ) { result, tuple in
            let groupID = try XCTUnwrap(tuple.conversation.mlsGroupID)
            result[groupID] = tuple
        }

        // mock conversation epoch
        mockCoreCrypto.mockConversationEpoch = { groupID in
            guard let tuple = conversationAndOutOfSyncTuplesByGroupID[MLSGroupID(groupID)] else {
                return 1
            }

            let epoch = tuple.conversation.epoch
            return tuple.isOutOfSync ? epoch + 1 : epoch
        }

        // ...
```

wire-ios-data-model/Source/MLS/MLSService.swift
David-Henner: applied [177c2a0](https://github.com/wireapp/wire-ios-mono/pull/429/commits/177c2a064663cedfe702e2759c866252c1476e5d)

wire-ios-data-model/Tests/MLS/MLSServiceTests.swift
johnxnguyen: yeah you're right

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -815,6 +817,60 @@ public final class MLSService: MLSServiceInterface {
...
+    private func isConversationOutOfSync(
+        _ conversation: ZMConversation,
+        coreCrypto: CoreCryptoProtocol
+    ) -> Bool {
+        guard let groupID = conversation.mlsGroupID else {
+            return false
+        }
+
+        guard let epoch = try? coreCrypto.conversationEpoch(conversationId: groupID.bytes) else {
+            // TODO: Decide if failing to fetch the epoch is considered being out of sync
```
netbe: any decision on this?

wire-ios-data-model/Source/Model/Conversation/ZMConversation+MLS.swift
```diff
@@ -173,24 +173,21 @@ public extension ZMConversation {
...
-            argumentArray: [
-                Self.messageProtocolKey,
-                MessageProtocol.mls.rawValue,
-                Self.mlsGroupIdKey
-            ]
-        )
-
         request.predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [
-            isSelfConversation, isMLSConveration
+            isSelfConversation, .isMLSConversation
```
netbe: can we do the same for isSelfConversation ?

netbe (approved):
just some comments

wire-ios-data-model/Source/Model/Conversation/ZMConversation+MLS.swift
David-Henner: done ✅

wire-ios-data-model/Source/MLS/MLSService.swift
David-Henner: not really, I guess it depends on why we fail to fetch the conversation epoch. It may be that rejoining the conversation is not the solution.

David-Henner: according to the [documentation](https://wireapp.github.io/core-crypto/core_crypto/struct.CoreCrypto.html#method.conversation_epoch), if there's an error it's related to the conversation not being found, which doesn't mean that it's out of sync. So I updated the logic here [3886e0f](https://github.com/wireapp/wire-ios-mono/pull/429/commits/3886e0f959a9f01a7e39fab6e5e1f539b79192a2)

johnxnguyen (approved):

PR #448, opened 2023-08-31
fix: call participant not removed from subconversation WPB-3061

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-3061" title="WPB-3061" target="_blank">BugWPB-3061</a>  [iOS] Participant still in MLS conference after leaving
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Given I (user A) am in an MLS conference with another user B, when B hangs up to leave the call, I still see B in my call participant list.

### Causes

When B leaves the call, the backend will send a remove proposal on their behalf to all other participants. A receives the proposal, but when they try to commit pending proposals Core Crypto returns an error stating there exist no pending proposals. Therefore B remains in the subconversation and the conference.

The reason why Core Crypto thinks there are no pending proposals is because we actually to try commit pending proposals in the parent group when the proposal is actually for the subgroup.

### Solutions

When committing pending proposals, check if a conversation has any sub conversations. If it doesn't, include those as part of the list of groups containing possible pending proposals.

#### Test Coverage 

- Added a unit test asserting we commit pending proposals for subgroups if they exist

#### How to Test

- Be in an an MLS conference with user B.
- User B hangs up.
- Check for user B they left the call.
- Check for user A they left the call.

----

----

----
##### References


wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1046,15 +1046,24 @@ public final class MLSService: MLSServiceInterface {
...
-                    return nil
+                    continue
                 }
 
-                return (groupID, timestamp)
+                result.append((groupID, timestamp))
+
+                // The pending proposal might be for the subconversation,
+                // so include it just in case.
+                if let subgroupID = subconverationGroupIDRepository.fetchSubconversationGroupID(
```
netbe, line 1061: so how can we be sure the pending proposal is for the subconversation?

from what I understand if it is for subconversation you also commit for the parent, no?

David-Henner, line 1061: So we have no way of knowing if the proposal is for the subgroup or the parent group?

johnxnguyen: We currently don't persist enough state to be able to determine at this point in time whether there exists a pending proposal for the parent, the sub conversation, or both. We could add this functionality, but I'm not sure that it is necessary.

We already don't exactly know which groups have pending proposals since only CC knows. We simply mark `ZMConversation`s as possibly having a pending proposal and eagerly try to commit them. If there are pending proposals, they will be committed, if there are none, then that's ok and nothing bad will happen.

> from what I understand if it is for subconversation you also commit for the parent, no?

The parent and subconversation have different group ids and I've found that we need to use the right id, otherwise CC returns some puzzling errors. In this case, I was trying to commit the pending proposals for the parent group, but CC had the pending proposal associated with the subgroup, so it was throwing "no pending proposals" which was confusing because I could clearly see we were getting it and it was valid.

johnxnguyen: We do, at the time we receive the message containing the pending proposal. If we really wanted to we could also persist this information such that at this later point in time we could know which of the two groups has proposals. But I couldn't see the benefit that would warrant the necessary overhead. Do you have something in mind?

David-Henner: Thanks for the explanation, I wasn't quite sure what the whole flow looked like. I agree the benefit isn't worth the overhead

David-Henner (approved):

netbe (approved):

PR #475, opened 2023-09-13
feat: recover from "not able to decrypt a message" WPB-534

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-534" title="WPB-534" target="_blank">StoryWPB-534</a>  [iOS] MLS Client recover from "not able to decrypt a message"
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

In this PR, we handle the scenario stated below

> A client can detect that it has missed handshake messages when decrypting an incoming message.  The user needs to be informed about potentially missed messages and the broken MLS group needs to be repaired.

When message decryption fails, we check the error. If the error is `WrongEpoch`, we fetch the conversation, verify the epoch is out of sync, then rejoin the conversation and insert a gap message.

We also handle the case of subconversations being out of sync. 

### Changes

- Moved `SyncConversationAction` to DM
- Added methods to fetch and repair a conversation / a subgroup to `MLSService`
- Added handling of the `WrongEpoch` error to `MLSDecryptionService`
- Handling wrong epoch scenario by repairing conversation in `MLSService`
- Conformed `MLSGroupdID` to `SafeForLoggingStringConvertible` for safe and readable logging
- Added condition to join self group only if it doesn't already exist (as a fix for a request loop)

### Testing

### DIFF
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -832,58 +832,14 @@ public final class MLSService: MLSServiceInterface {
             launchConversationRepairTaskIfNotInProgress(for: groupID) {
                 do {
                     try await self.joinGroup(with: groupID)
-                    self.logger.info("repaired out of sync conversation (\(groupID))")
+                    self.logger.info("repaired out of sync conversation (\(groupID.safeForLoggingDescription))")
                 } catch {
-                    self.logger.warn("failed to repair out of sync conversation (\(groupID)). error: \(String(describing: error))")
+                    self.logger.warn("failed to repair out of sync conversation (\(groupID.safeForLoggingDescription)). error: \(String(describing: error))")
                 }
             }
         }
     }
 
-    func fetchAndRepairConversation(with groupID: MLSGroupID) {
-        Task {
-            await fetchAndRepairConversation(with: groupID)
-        }
-    }
-
-    func fetchAndRepairConversation(with groupID: MLSGroupID) async {
-        guard let context = context else { return }
-
-        do {
-            logger.info("repairing conversation (\(groupID))")
-
-            guard
-                let conversation = ZMConversation.fetch(with: groupID, in: context),
-                let qualifiedID = conversation.qualifiedID
-            else {
-                logger.warn("failed to fetch conversation's qualified id")
-                return
-            }
-
-            try await actionsProvider.syncConversation(
-                qualifiedID: qualifiedID,
-                context: context.notificationContext
-            )
-
-            guard isConversationOutOfSync(conversation) else {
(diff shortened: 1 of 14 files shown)


netbe (approved):

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -832,14 +832,124 @@ public final class MLSService: MLSServiceInterface {
...
+            )
+
+            let localEpoch = try coreCrypto.perform {
+                return try $0.conversationEpoch(conversationId: subgroup.groupID.bytes)
+            }
+
+            guard localEpoch != subgroup.epoch else {
+                logger.info("subgroup is not out of sync (parent: \(parentGroupID.safeForLoggingDescription), subgroup: \(subgroup.groupID.safeForLoggingDescription))")
+                return
+            }
```
johnxnguyen: Would it be a good idea to refactor 

```swift
private func isConversationOutOfSync(
        _ conversation: ZMConversation,
        coreCrypto: CoreCryptoProtocol,
        context: NSManagedObjectContext
    )
```

so that it can work for both any mls group?

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1068,11 +1192,21 @@ public final class MLSService: MLSServiceInterface {
...
+            )
+        } catch {
+            if case MLSDecryptionService.MLSMessageDecryptionError.wrongEpoch = error {
+                fetchAndRepairConversation(
+                    with: groupID,
+                    subconversationType: subconversationType
+                )
+            }
+            throw error
+        }
```
johnxnguyen: What do you think about:

```swift
catch MLSDecryptionService.MLSMessageDecryptionError.wrongEpoch {
  ...
} catch {
  ...
}
```

wire-ios-data-model/Source/MLS/MLSService.swift
David-Henner: Yep, good idea 👍 
Done [0f08bbb](https://github.com/wireapp/wire-ios-mono/pull/475/commits/0f08bbb40381582ba71258bafb4762c1bcfa1637)

wire-ios-data-model/Source/MLS/MLSService.swift
David-Henner: Better indeed
Done [0f08bbb](https://github.com/wireapp/wire-ios-mono/pull/475/commits/0f08bbb40381582ba71258bafb4762c1bcfa1637)

johnxnguyen (approved):

PR #500, opened 2023-09-19
feat: Add progress bar to the AddParticipantsViewController WPB-4451

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4451" title="WPB-4451" target="_blank">BugWPB-4451</a>  [iOS] Group creation does nothing when backend offline
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Creating a conversation sometimes takes a few seconds if there are users from different domains or one of the domains is offline. From the user's point of view, it looks like the "Done" button does not work and can be clicked many times. As a result, several conversations are created in a few seconds.

### Solutions

The solution is to add a loading view and prevent users from creating many conversations while the backend is responding.

----

----

----
##### References

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/AddContacts/AddParticipantsViewController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/AddContacts/AddParticipantsViewController.swift
@@ -87,7 +87,7 @@ extension AddParticipantsViewController.Context {
     }
 }
 
-final class AddParticipantsViewController: UIViewController {
+final class AddParticipantsViewController: UIViewController, SpinnerCapable {
 
     enum CreateAction {
         case updatedUsers(UserSet)
@@ -120,6 +120,8 @@ final class AddParticipantsViewController: UIViewController {
             updateValues()
         }
     }
+
+    var dismissSpinner: SpinnerCompletion?
 
     deinit {
         userSelection.remove(observer: self)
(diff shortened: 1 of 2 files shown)


David-Henner (approved):

wire-ios/Wire-iOS/Sources/UserInterface/AddContacts/AddParticipantsViewController.swift
```diff
@@ -345,10 +347,18 @@ final class AddParticipantsViewController: UIViewController {
     @objc private func rightNavigationItemTapped(_ sender: Any!) {
         switch viewModel.context {
         case .add: navigationController?.dismiss(animated: true, completion: nil)
-        case .create: conversationCreationDelegate?.addParticipantsViewController(self, didPerform: .create)
+
+        case .create:
+            setLoadingView(isVisible: true)
```
netbe: That would be easier to maintain if putting it visible and not visible are at the same place

so either do this in delegate, or setting it off after delegate. your pick?

KaterinaWire: Yes, I moved it to the same place

netbe (approved):

PR #539, opened 2023-09-22
fix: add recovery step for database migration failure WPB-3313

lewandowskit93:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-3313" title="WPB-3313" target="_blank">BugWPB-3313</a>  Missing recovery step when database migration error occurs
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When database migration fails the user is presented with alert to open settings or delete database files. When user selects option of "open settings" and get back to the app, he is stuck on the shield page.

### Causes (Optional)

The alert is not presented again when getting to the app from background.

### Solutions

1. Show migration failure alert again when app returns from background
2. Add "Retry" action to try migrating database once more

#### How to test

1. Log in into app
2. Use Shake Gesture
3. Open `Configure flags` menu
4. Turn `forceDatabaseLoadingFailure` switch on
5. Restart the app
6. The database migration failure alert should be presented

#### Test Coverage (Optional)

----

----

----
##### References


wire-ios/Wire-iOS/Sources/UserInterface/BlockerViewController.swift
```diff
@@ -100,6 +106,16 @@ final class BlockerViewController: LaunchImageViewController {
 
         databaseFailureAlert.addAction(settingsAction)
 
+        let retryAction = UIAlertAction(
+            title: L10n.Localizable.Databaseloadingfailure.Alert.retry,
+            style: .default,
+            handler: { [weak self] _ in
+                self?.sessionManager?.recoverDatabase()
```
netbe: what does this do? show again the popup?

wire-ios/Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -367,6 +372,14 @@ extension AppRootRouter {
                                completion: completion)
     }
 
+    private func showRecovery(completion: @escaping () -> Void) {
+        guard let launchOptions = lastLaunchOptions else { return }
+        completion()
+        enqueueTransition(to: .headless) { [weak self] in
+            self?.sessionManager.start(launchOptions: launchOptions)
+        }
+    }
```
netbe: what does this do exactly?

wire-ios/Wire-iOS/Sources/UserInterface/BlockerViewController.swift
lewandowskit93: It is there just for allowing to access AppStateCalculator indirectly. It is triggered by sessionManager via delegate. So the function just runs the delegate which would then generate a new state as a result.
I've done it that way to not store & pass around completion handlers via enums as then it would be difficult to compare states in tests. It is also the cleaner option.

wire-ios/Wire-iOS/Sources/AppRootRouter.swift
lewandowskit93: Basically it re-triggers the steps that are performed while starting a session (trying to migrate database) just as the first time.
It turns out that retrying was not possible at all as it was dependent on launchOptions and launching process.

wire-ios/Wire-iOS/Sources/UserInterface/BlockerViewController.swift
netbe: ok

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -853,7 +854,12 @@ public final class SessionManager: NSObject, SessionManagerType {
                 }
 
                 coreDataStack.loadStores { error in
-                    if error != nil {
+                    if DeveloperFlag.forceDatabaseLoadingFailure.isOn {
+                        var flag = DeveloperFlag.forceDatabaseLoadingFailure
```
netbe, line 859:
```suggestion
                        // flip off the flag in order not to be stuck in failure
                        var flag = DeveloperFlag.forceDatabaseLoadingFailure
```

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -877,6 +883,10 @@ public final class SessionManager: NSObject, SessionManagerType {
         }
     }
 
+    public func enqueueRetryStart() {
```
netbe: does it really enqueue here ?

```suggestion
    public func retryStart() {
```

wire-ios/Wire-iOS/Sources/UserInterface/BlockerViewController.swift
```diff
@@ -43,6 +44,11 @@ final class BlockerViewController: LaunchImageViewController {
         fatalError("init?(coder aDecoder: NSCoder) is not implemented")
     }
 
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        setupApplicationNotifications()
+    }
+
     override func viewDidAppear(_ animated: Bool) {
         showAlert()
```
netbe, line 54: is it intended we don't call super here?

lewandowskit93: fixed

netbe (approved):

agisilaos (approved):

PR #558, opened 2023-09-27
fix: crash - WPB-4895

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4895" title="WPB-4895" target="_blank">BugWPB-4895</a>  crash fetching multiple Users
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

On a federated environment, we got the following crash (see attachment)

### Causes (Optional)

It crashes because we end up fetching ZMUser with same remoteIdentifier without specifying the domain (and we assert only one occurrence should exist).

The call is made by `UserProfileByIDTranscoder` which should not be used on api version 4 as we deal with QualifiedId only.  

The `UserProfileRequestStrategy`'s `nextRequest(for:)` method will call `UserProfileByIDTranscoder` if the `UserProfileByQualifiedIDTranscoder` does not return a request, which will perform the mentioned fetch causing the crash.

### Solutions

Use `UserProfileByIDTranscoder` only on api version 0.

#### Test Coverage (Optional)

### Attachments (Optional)

Screenshot 2023-09-27 at 21 27 06

----

----

----
##### References

### DIFF
--- a/wire-ios-request-strategy/Sources/Request Strategies/User/UserProfileRequestStrategyTests.swift
+++ b/wire-ios-request-strategy/Sources/Request Strategies/User/UserProfileRequestStrategyTests.swift
@@ -82,6 +82,25 @@ class UserProfileRequestStrategyTests: MessagingTestBase {
         }
     }
 
+    func testThatRequestInV4_DoesNotUseLegacyEndpointWhenNoRequestFromCurrentEndpoint() {
+        syncMOC.performGroupedBlockAndWait {
+            // given
+            self.apiVersion = .v0
+            self.otherUser.domain = "example.com"
+            self.otherUser.needsToBeUpdatedFromBackend = true
+            // set object to sync to legacy endpoint transcoder
+            self.sut.objectsDidChange(Set([self.otherUser]))
+
+            // when
+            // use v4 endpoint
+            self.apiVersion = .v4
+            let request = self.sut.nextRequest(for: self.apiVersion)
+
+            // then
+            // .v0 endpoint should not be used
+            XCTAssertNil(request)
+        }
+    }
     // MARK: - Slow Sync
 
     func testThatRequestToFetchConnectedUsersIsGenerated_DuringFetchingUsersSyncPhase() {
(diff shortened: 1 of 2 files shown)


wire-ios-request-strategy/Sources/Request Strategies/User/UserProfileRequestStrategyTests.swift
```diff
@@ -201,6 +220,7 @@ class UserProfileRequestStrategyTests: MessagingTestBase {
     // MARK: - Response processing
 
     func testThatUsesLegacyEndpoint_WhenFederatedEndpointIsDisabled() {
+        // TODO: check if this test is right?
```
netbe: this passes because we're on v0 but I am questioning the veracity of the description of this test.

wire-ios-request-strategy/Sources/Request Strategies/User/UserProfileRequestStrategyTests.swift
```diff
@@ -82,6 +82,25 @@ class UserProfileRequestStrategyTests: MessagingTestBase {
...
+
+            // when
+            // use v4 endpoint
+            self.apiVersion = .v4
+            let request = self.sut.nextRequest(for: self.apiVersion)
+
+            // then
+            // .v0 endpoint should not be used
+            XCTAssertNil(request)
+        }
```
johnxnguyen: I was a bit confused by this test but I think what it's doing is:
- By reporting the `otherUser` did change while on `v0`, `sut` will case the legacy transcoder to get in a state where it would produce a next request
- By switching to `v4` and asking for a next request, we get `nil` because we would only ask the non legacy transcoder for a request, but it's not in a state to do that

Is that correct?

wire-ios-request-strategy/Sources/Request Strategies/User/UserProfileRequestStrategyTests.swift
johnxnguyen: `v0` and `v1` are virtually the same APIs. just that in v0 the federated endpoints (with qualified ids) don't work and shouldn't be used, and in v1 they do work and should be used (rather than their non federated versions).

But that's not really common knowledge and hard to extract from this test name. Is this the concern you had?

We could name it... "use legacy endpoint on v0".

johnxnguyen (approved):

wire-ios-request-strategy/Sources/Request Strategies/User/UserProfileRequestStrategyTests.swift
netbe: Exactly i'll add your comment to the test so it's less confusing

KaterinaWire (approved):

PR #568, opened 2023-09-28
fix: popup about users started using new device WPB-4906

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4906" title="WPB-4906" target="_blank">BugWPB-4906</a>  [iOS] Pop-up about users started using new devices
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When sending a message to a conversation they've been added to, users see the **security degraded** ("... started using new devices") popup and must confirm they want to send a message, even though the conversation was never verified 

### Causes

When added to a conversation, the user receives a `conversation.member-join` event. Processing this events results in insert a new conversation object in the context, adding the self client, and setting the security level. The newly inserted conversation is then synced with the backend, which returns the members of the conversation. As a result, the conversation's security level is reevaluated to account for the new members.

When first setting the security level, there is only the self client in the conversation. So the conversation is considered as secure. 
When reevaluating the security level for the new members, the security level is degraded, since they're not verified.

Then when the user sends their first message in the conversation, they see the security degraded alert.

### Solutions

Only upgrade the conversation's security level to `secure` once there isn't only the self client in the conversation and the other clients are verified

### DIFF
--- a/wire-ios-data-model/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
+++ b/wire-ios-data-model/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
@@ -209,6 +209,7 @@ extension ZMConversation {
             securityLevel != .secure &&
             allUsersTrusted &&
             allParticipantsHaveClients &&
+            hasMoreClientsThanSelfClient &&
             conversationType.isOne(of: .group, .oneOnOne, .invalid)
         else {
             return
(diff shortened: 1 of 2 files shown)


agisilaos (approved):

wire-ios-data-model/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
```diff
@@ -676,6 +677,23 @@ extension ZMConversation {
...
+            return false
+        }
+
+        let clients = localParticipants.flatMap(\.clients)
+
+        if clients.contains(selfClient) && clients.count == 1 {
+            return false
+        }
+
+        return true
```
netbe, line 694:
```suggestion
        return !(clients.contains(selfClient) && clients.count == 1)
```
shorter, though might be less easier to read, your call

netbe (approved):

PR #569, opened 2023-09-29
fix: request loop at /users/list-clients WPB-4894

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4894" title="WPB-4894" target="_blank">BugWPB-4894</a>  iOS request loop for /users/list-clients
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

If making a request for `users/list-clients` is for a user from an unreachable backend then it enters a request loop.

### Causes

The backend returns a 200 success response but contains no client payload data for the unreachable user. As a result, we can't update the local `UserClient` objects in the database, specifically marking them as updated. The operation loop continues and since the local `UserClient` objects are still marked as needing a backend update, we generate the same request again and have the same response.

### Solutions

Regardless if we have updated the local `UserClient` from response payloads, mark all clients as updated so the request isn't made again. 

#### Test Coverage

- Unit test asserting we mark all clients as updated even for an empty response.

#### How to Test

- Log in with user A on backend A.
- Be in a conversation with user B from backend B.
- User A puts app in background.
- User B logs in with a new device, sends a message in conversation with A.
- Backend B goes offline.
- User A puts app to foreground.
- Assert no request loop.

----

----

----
##### References

### DIFF
--- a/wire-ios-request-strategy/Sources/Request Strategies/User Clients/FetchingClientRequestStrategy.swift
+++ b/wire-ios-request-strategy/Sources/Request Strategies/User Clients/FetchingClientRequestStrategy.swift
@@ -363,6 +363,12 @@ final class UserClientByQualifiedUserIDTranscoder: IdentifierObjectSyncTranscode
     }
 
     private func commonResponseHandling(response: ZMTransportResponse, for identifiers: Set<QualifiedID>) {
+        defer {
+            // We mark all clients as synced, even if they did not appear in
+            // the reponse payload, in order to avoid a possible request loop.
+            markAllClientsAsUpdated(identifiers: identifiers)
+        }
+
         guard
             let rawData = response.rawData,
             let payload = ResponsePayload(rawData, decoder: decoder),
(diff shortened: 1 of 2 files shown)


wire-ios-request-strategy/Sources/Request Strategies/User Clients/FetchingClientRequestStrategy.swift
```diff
@@ -389,6 +395,22 @@ final class UserClientByQualifiedUserIDTranscoder: IdentifierObjectSyncTranscode
...
 
+    private func markAllClientsAsUpdated(identifiers: Set<QualifiedID>) {
+        for identifier in identifiers {
+            if let user = ZMUser.fetch(
+                with: identifier.uuid,
+                domain: identifier.domain,
+                in: managedObjectContext
+            ) {
+                for client in user.clients {
+                    client.needsToBeUpdatedFromBackend = false
```
netbe, line 406: to fully understand, I guess this property has been set to true before making the request, right?

netbe (approved):

KaterinaWire (approved):

wire-ios-request-strategy/Sources/Request Strategies/User Clients/FetchingClientRequestStrategy.swift
johnxnguyen: yes that's correct

PR #577, opened 2023-09-29
fix: request loop at /users/list-clients (take 2) WPB-4849

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4849" title="WPB-4849" target="_blank">BugWPB-4849</a>  Missing pending icon on outgoing connection request
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Follow up from https://github.com/wireapp/wire-ios/pull/569. 

### Issues

That fix didn't work when push notifications was enabled. 

### Causes

When a message from a new client is received while the app is in the background, then the new client is created in the notification extension. I suspect that when the main app is opened, although we find the client created in the extension, it isn't yet visible in the `ZMUser.clients` relationship. However, the opposite relationship `UserClient.user` is set.

As a result, we were not always marking the clients as updated when after processing the response, and these unmarked clients would then trigger the request loop again.

### Solutions

Refactor how we fetch the user clients needing to be marked as updated, obtaining directly from a fetch request rather than via the user and `clients` relationship.

#### How to Test

- Log in to backend A with user A, enable push notifications.
- Log in to backend B with user B, share a conversation with user A.
- User A puts app in background.
- User B logs in to a new device, sends messages to A.
- A receives push notifications for the new messages.
- Backend B goes offline.
- User A opens the app.
- Assert there is no request loop.

----

----

----
##### References

### DIFF
--- a/wire-ios-request-strategy/Sources/Request Strategies/User Clients/FetchingClientRequestStrategy.swift
+++ b/wire-ios-request-strategy/Sources/Request Strategies/User Clients/FetchingClientRequestStrategy.swift
@@ -396,19 +396,17 @@ final class UserClientByQualifiedUserIDTranscoder: IdentifierObjectSyncTranscode
     }
 
     private func markAllClientsAsUpdated(identifiers: Set<QualifiedID>) {
-        for identifier in identifiers {
-            if let user = ZMUser.fetch(
-                with: identifier.uuid,
-                domain: identifier.domain,
-                in: managedObjectContext
-            ) {
-                for client in user.clients {
+        let clients = UserClient.fetchClientsNeedingUpdateFromBackend(in: managedObjectContext)
+
+        for client in clients {
+            if let qualifiedID = client.user?.qualifiedID {
+                if identifiers.contains(qualifiedID) {
                     client.needsToBeUpdatedFromBackend = false
                 }
             }
         }
 
-        managedObjectContext.enqueueDelayedSave()
+        managedObjectContext.saveOrRollback()
     }
 
 }


wire-ios-request-strategy/Sources/Request Strategies/User Clients/FetchingClientRequestStrategy.swift
```diff
@@ -396,19 +396,17 @@ final class UserClientByQualifiedUserIDTranscoder: IdentifierObjectSyncTranscode
...
 
     private func markAllClientsAsUpdated(identifiers: Set<QualifiedID>) {
-        for identifier in identifiers {
-            if let user = ZMUser.fetch(
-                with: identifier.uuid,
-                domain: identifier.domain,
-                in: managedObjectContext
-            ) {
-                for client in user.clients {
+        let clients = UserClient.fetchClientsNeedingUpdateFromBackend(in: managedObjectContext)
```
netbe, line 399: I guess you don't need the identifiers now, or do you want to log clients ids and the identifiers too so we can check on DD?

johnxnguyen: We do use the identifiers, see a couple lines below

netbe: my bad I missread

netbe (approved):

KaterinaWire (approved):

PR #593, opened 2023-10-10
fix: can't send reactions in MLS conversations WPB-2215

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-2215" title="WPB-2215" target="_blank">BugWPB-2215</a>  [iOS] Can't like or unlike a message from iOS in an MLS conversation
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
If I send a reaction in an MLS group, then I don't receive it on any other client in that group.

### Causes
When we want to send a reaction message in the group, we "sync" the message via the `MLSMessageSync`. However we erroneously call the completion handler with a success result after syncing the message, instead of when the message is actually encrypted and sent and the backend responds with a success. When the completion handler is invoked it will delete the reaction message, which is expected because they don't need to stay in the database. This premature deletion is preventing the reactions from being sent.

### Solutions
Remove the call to the completion handler (it is passed along in the right place already).

#### How to Test

- Send a reaction in an MLS conversation
- Assert that it can be received by other users in the conversation.

----

----

----
##### References

### DIFF
--- a/wire-ios-request-strategy/Sources/Object Syncs/Helpers/MLSMessageSync.swift
+++ b/wire-ios-request-strategy/Sources/Object Syncs/Helpers/MLSMessageSync.swift
@@ -83,7 +83,6 @@ class MLSMessageSync<Message: MLSMessage>: NSObject, ZMContextChangeTrackerSourc
             context.performGroupedBlock { [dependencySync] in
                 dependencySync.synchronize(entity: message, completion: completion)
                 RequestAvailableNotification.notifyNewRequestsAvailable(nil)
-                completion(.success(()), .init())
             }
         }
     }
--- a/wire-ios-request-strategy/Tests/Sources/Mocks/MockMLSService.swift
+++ b/wire-ios-request-strategy/Tests/Sources/Mocks/MockMLSService.swift
@@ -164,10 +164,6 @@ class MockMLSService: MLSServiceInterface {
         fatalError("not implemented")
     }
 
-    func subconversationMembers(for subconversationGroupID: WireDataModel.MLSGroupID) throws -> [WireDataModel.MLSClientID] {
-        fatalError("not implemented")
-    }
-
     // MARK: - New epoch
 
     func generateNewEpoch(groupID: MLSGroupID) async throws {


agisilaos (approved):

wire-ios-request-strategy/Sources/Object Syncs/Helpers/MLSMessageSync.swift
```diff
@@ -83,7 +83,6 @@ class MLSMessageSync<Message: MLSMessage>: NSObject, ZMContextChangeTrackerSourc
             context.performGroupedBlock { [dependencySync] in
                 dependencySync.synchronize(entity: message, completion: completion)
                 RequestAvailableNotification.notifyNewRequestsAvailable(nil)
```
netbe, line 85: should this (RequestAvailableNotification) be called within the passed completion? looks weird too

johnxnguyen: no, its needed to generate a request. The completion should be called in case of a failure or a success response.

netbe: I wanted to know what's the reason for generating a request here

johnxnguyen: To send a message you need to:
- Insert a new `ZMClientMessage` in the database
- "Sync" the message via `MessageSync`. This will in turn forward it to either `ProteusMessageSync` or `MLSMessageSync` depending on the conversation.
- This will forward it to a "dependency sync" which will generate one or more requests depending on the dependencies of the message (eg. maybe we need to fetch conversation or user metadata first before we can send the message)
- Start the operation loop polling by posting the request available notification: this will ask the `ClientMessageRequestStrategy` if it needs to make any request, which will ask message sync if it needs to make requests, which will ask the dependency sync if it needs to make requests, which will eventually generate the request to send the message.
- When generating the request to send the message, we encrypt the message, put it in a payload, and send it to the backend.
- We handle the response and update local state if needed (e.g mark the message as being sent).

netbe: thks I got it now - also looking at DependencyEntitySync help :)

netbe (approved):

PR #600, opened 2023-10-11
fix: Xcode and SwiftLint warnings across the entire project

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

From **151 warnings** down to **110**

With this PR I fixed a variety of SwiftLint and Xcode warnings across the entire project. Here's the list of the warnings that are fixed on this PR:

- Call to method 'finishWithoutShowingNotification' in closure requires explicit use of 'self' to make capture semantics explicit; this is an error in Swift 6
- Unneeded Overridden Functions Violation: Remove overridden functions that don't do anything except call their super `(unneeded_override)`
- Variable 'rewindedStep' was never mutated; consider changing to 'let' constant
- Trailing Whitespace Violation: Lines should not have trailing whitespace `(trailing_whitespace)`
- String interpolation produces a debug description for an optional value; did you mean to make this explicit?
- Control Statement Violation: `if`, `for`, `guard`, `switch`, `while`, and `catch` statements shouldn't unnecessarily wrap their conditionals or arguments in parentheses `(control_statement)`
- Initialization of immutable value 'recordButtonFrame' was never used; consider replacing with assignment to '_' or removing it
- Left side of nil coalescing operator '??' has non-optional type 'String', so the right side is never used
- Variable 'user' was never mutated; consider changing to 'let' constant
- Switch and Case Statement Alignment Violation: Case statements should vertically aligned with their closing brace `(switch_case_alignment)`
- Opening Brace Spacing Violation: Opening braces should be preceded by a single space and on the same line as the declaration `(opening_brace)`

----
##### References


wire-ios/Wire-iOS/Sources/UserInterface/Calling/BottomSheetContainerViewController/EstablishingCallStatusView.swift
```diff
@@ -40,7 +40,7 @@ extension CallStatusViewState {
         switch self {
         case .none: return ""
         case .connecting: return CallStatus.connecting
-        case .ringingIncoming(name: let name?): return CallStatus.Incoming.user("\(name ?? "")")
+        case .ringingIncoming(name: let name?): return CallStatus.Incoming.user("\(name)")
```
johnxnguyen, line 43: Is "name" not optional?

wire-ios/Wire-iOS/Sources/UserInterface/Components/Views/ConversationAvatarView.swift
```diff
@@ -166,7 +166,7 @@ final class ConversationAvatarView: UIView {
                 return
             }
 
-            accessibilityLabel = "Avatar for \(conversation.displayName)"
+            accessibilityLabel = "Avatar for \(conversation.displayNameWithFallback)"
```
johnxnguyen, line 169: Which warning was this?

agisilaos: String interpolation produces a debug description for an optional value; did you mean to make this explicit?

wire-ios/Wire-iOS/Sources/UserInterface/Calling/BottomSheetContainerViewController/EstablishingCallStatusView.swift
agisilaos: No! In that case I fixed this error:

Left side of nil coalescing operator '??' has non-optional type 'String', so the right side is never used

wire-ios/Wire-iOS/Sources/UserInterface/Components/Views/ConversationAvatarView.swift
johnxnguyen: What's the fallback?

johnxnguyen: If we want to keep the original behavior, we can use `String(describing: conversation.displayName)`

agisilaos: @johnxnguyen 

```swift
extension ConversationLike {

    var displayNameWithFallback: String {
        return displayName ?? L10n.Localizable.Profile.Details.Title.unavailable
    }

}
```

The fallback is `Name not Available` as recently introduced by Katerina! I guess we can use that. and that's why I'm using it.

wire-ios-data-model/Source/Model/Message/ConversationMessage+Deletion.swift
```diff
@@ -67,7 +67,7 @@ extension ZMMessage {
     }
 
     @discardableResult @objc func deleteForEveryone() -> ZMClientMessage? {
```
theswiftway, line 69: [Suggestion] Should this return a `Result` instead ?

theswiftway: Not in the scope of this PR though

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationQuoteCell.swift
```diff
@@ -37,12 +37,12 @@ final class ConversationReplyContentView: UIView {
 
         var showDetails: Bool {
             guard let message = quotedMessage,
-                  (message.isText
+                  message.isText
```
theswiftway, line 40: Looking at all the checks, I think `isFile` check is good enough for this logic

wire-ios-data-model/Source/Model/Message/ConversationMessage+Deletion.swift
agisilaos: I agree let's explore that on a separate PR.

wire-ios-data-model/Source/Model/Message/ConversationMessage.swift
```diff
@@ -293,7 +293,7 @@ extension ZMMessage: ZMConversationMessage {
 
     public var isRestricted: Bool {
         guard
-            (self.isFile || self.isImage),
+            self.isFile || self.isImage,
```
theswiftway, line 296: `self.isFile` should be good enough. We could skip `self.isImage`

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationQuoteCell.swift
agisilaos: let's explore that on a separate PR. On this one, I would like to focus on fixing the warnings.

wire-ios-data-model/Source/Model/Message/ConversationMessage.swift
agisilaos: let's explore that on a separate PR. On this one, I would like to focus on fixing the warnings.

theswiftway (commented):
LGTM

theswiftway (approved):

markusfassbender (approved):
Great styling cleanup!

PR #601, opened 2023-10-12
fix: crash when context data is unavailable WPB-4920

johnxnguyen:
<table>
<td>
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
If we are unable to get context data to encrypt / decrypt a message in an EAR protected database then we will crash the app.

### Causes
The context data we use is derived from the self user and self client ids. This data is always expected to exist, so there was a fatal error when it didn't.

### Solutions
While unlikely, it is possible that some code may be executed after the assumptions that certain objects exist are no longer valid. For example, when a notification is fired and then the user session is expired or torn down. Therefore we shouldn't fatal error, but rather make an assertion failure and fail softly. In this particular case, the worst thing the user will see is an error message in the conversation that the messages are not available. 

#### How to Test
There isn't really a way to test this.

----

----

----
##### References

### DIFF
--- a/wire-ios-data-model/Source/ManagedObjectContext/NSManagedObjectContext+EncryptionAtRest.swift
+++ b/wire-ios-data-model/Source/ManagedObjectContext/NSManagedObjectContext+EncryptionAtRest.swift
@@ -197,12 +197,17 @@ extension NSManagedObjectContext {
     enum EncryptionError: LocalizedError {
 
         case missingDatabaseKey
+        case missingContextData
         case cryptobox(error: ChaCha20Poly1305.AEADEncryption.EncryptionError)
 
         var errorDescription: String? {
             switch self {
             case .missingDatabaseKey:
                 return "Database key not found. Perhaps the database is locked."
+
+            case .missingContextData:
+                return "Couldn't obtain context data."
+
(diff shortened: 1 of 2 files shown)


KaterinaWire (approved):

wire-ios-data-model/Source/ManagedObjectContext/NSManagedObjectContext+EncryptionAtRest.swift
```diff
@@ -273,7 +282,8 @@ extension NSManagedObjectContext {
             let selfClientId = selfClient.remoteIdentifier,
             let context = (selfUserId + selfClientId).data(using: .utf8)
         else {
-            fatalError("Could not obtain self user id and self client id")
+            assertionFailure("Could not obtain self user id and self client id")
```
netbe, line 285: should we log this too to datadog?

johnxnguyen: done

netbe (approved):

PR #37, opened 2023-02-21
fix: problems with internet connection detection SQSERVICES-1917

MarcinR:
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [x] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [x] contains a reference JIRA issue number like `SQPIT-764`
  - [x] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [x] is free of optional paragraphs and you have filled the relevant parts to the best of your ability

PR #606, opened 2023-10-13
feat: User input for their own password - WPB-4638

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4638" title="WPB-4638" target="_blank">TaskWPB-4638</a>  [iOS] User input for their own password
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR I add new text fields to the screen and the ability of their contents to be validated. For that reason, I introduced a new passwordRuleSet for the guestLink feature and I made the appropriate modifications for that ruleSet to be used when it comes to validating the password on that view. In addition to that I've added some more tests. 

P.s I would like to refactor/improve the whole ValidatedTextField and password validation in general but it's going to happen in a separate PR.

----
##### References


wire-ios/Wire-iOS/Sources/Authentication/Interface/Views/ValidatedTextField.swift
```diff
@@ -315,6 +315,10 @@ final class ValidatedTextField: AccessoryTextField, TextContainer, Themeable {
         return enableConfirmButton?() ?? !input.isEmpty
     }
 
+    var isValid: Bool {
```
agisilaos, line 318: I know the name of that property is not that good. Any suggestions are more than appreciated!

wire-ios/Wire-iOS/Sources/Authentication/Helpers/PasscodeRules+Shared.swift
```diff
@@ -29,6 +29,14 @@ extension PasswordRuleSet {
         return try! decoder.decode(PasswordRuleSet.self, from: fileData)
     }()
 
+    /// The guestLinkWithPassword rule set.
+    static let guestLinkWithPasswordRuleSet: PasswordRuleSet = {
+        let fileURL = Bundle.main.url(forResource: "guestLinkWithPassword_rules", withExtension: "json")!
```
KaterinaWire: can we try to avoid forced unwrapping?
same in #35 and #37 lines

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreatePasswordSecuredLink/CreateSecureGuestLinkViewModel.swift
```diff
@@ -46,6 +47,25 @@ final class CreateSecureGuestLinkViewModel {
...
+
+        if !textField.isValid {
+            return false
+        }
+
+        if confirmPasswordField.text != updatedString {
+            return false
+        }
+
+        return true
```
KaterinaWire: can we do it in one guard ?
```
guard let updatedString = textField.text, 
           !updatedString.isEmpty,
           textField.isValid, 
           confirmPasswordField.text == updatedString else {
            return false
        }
        return true
```

wire-ios/Wire-iOS/Sources/Authentication/Helpers/PasscodeRules+Shared.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreatePasswordSecuredLink/CreateSecureGuestLinkViewModel.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreatePasswordSecuredLink/CreateSecureGuestLinkViewModel.swift
```diff
@@ -46,6 +47,21 @@ final class CreateSecureGuestLinkViewModel {
         delegate?.viewModel(self, didGeneratePassword: randomPassword)
     }
 
+    func validatePassword(
+        for textField: ValidatedTextField,
+        against confirmPasswordField: ValidatedTextField
+    ) -> Bool {
+
+        guard let updatedString = textField.text,
```
theswiftway: [Opinion] would this be clearer?
```suggestion
        return textField.text != nil &&
    confirmPasswordField.text != nil &&
    !textField.text!.isEmpty &&
    textField.text == confirmPasswordField.text &&
    textField.isValid
```

wire-ios/Wire-iOS/Sources/Authentication/Interface/Views/ValidatedTextField.swift
```diff
@@ -335,9 +339,12 @@ final class ValidatedTextField: AccessoryTextField, TextContainer, Themeable {
         validateInput()
     }
 
-    func validateInput() {
-        let error = textFieldValidator.validate(text: text, kind: kind)
+    @discardableResult
+    func validateInput(useGuestLinkRuleset: Bool = false) -> Bool {
+        let error = textFieldValidator.validate(text: text, kind: kind, useGuestLinkRuleset: useGuestLinkRuleset)
```
KaterinaWire: you can check if the error is nil, you can return the result and not continue the calculation.

wire-ios/Wire-iOS/Sources/Authentication/Interface/Views/ValidatedTextField.swift
```diff
@@ -335,9 +339,12 @@ final class ValidatedTextField: AccessoryTextField, TextContainer, Themeable {
         validateInput()
     }
 
-    func validateInput() {
-        let error = textFieldValidator.validate(text: text, kind: kind)
+    @discardableResult
+    func validateInput(useGuestLinkRuleset: Bool = false) -> Bool {
```
KaterinaWire: The name of the method sounds like an action and it is not clear whether it returns any value.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreatePasswordSecuredLink/CreateSecureGuestLinkViewModel.swift
theswiftway: Or
```suggestion
        guard let enteredPassword = textField.text, 
            let confimationText = confirmPasswordField.text,
            !enteredPassword.isEmpty,
            textField.isValid,
            enteredPassword == confimationText else {
        return false
    }
    return true
```

wire-ios/Wire-iOS/Sources/Authentication/Interface/Helpers/TextFieldValidator.swift
```diff
@@ -51,7 +57,7 @@ final class TextFieldValidator {
         }
     }
 
-    func validate(text: String?, kind: ValidatedTextField.Kind) -> TextFieldValidator.ValidationError? {
+    func validate(text: String?, kind: ValidatedTextField.Kind, useGuestLinkRuleset: Bool = false) -> TextFieldValidator.ValidationError? {
```
KaterinaWire: I would create an enum with a ruleset type instead of a property. It will be more expandable in the future.

wire-ios/Wire-iOS/Sources/Authentication/Interface/Views/ValidatedTextField.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/Authentication/Interface/Views/ValidatedTextField.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreatePasswordSecuredLink/CreateSecureGuestLinkViewModel.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/Authentication/Interface/Helpers/TextFieldValidator.swift
agisilaos: Done!

wire-ios/Wire-iOS Tests/CreateSecureGuestLinkViewModelTests.swift
```diff
@@ -19,12 +19,14 @@
 import XCTest
 @testable import Wire
 
-class CreatePasswordSecuredLinkViewModelTests: XCTestCase {
+class CreateSecureGuestLinkViewModelTests: BaseSnapshotTestCase {
```
johnxnguyen: Why is it a snapshot test when we're testing the view model?

wire-ios/Wire-iOS Tests/CreateSecureGuestLinkViewModelTests.swift
```diff
@@ -33,18 +35,62 @@ class CreatePasswordSecuredLinkViewModelTests: XCTestCase {
...
+    // MARK: - Validation
+
+    func testPasswordValidationWithEmptyTextField() {
+        // GIVEN
+        textField.text = ""
+        confirmPasswordField.text = "somePassword"
+
+        // WHEN && THEN
+        XCTAssertFalse(viewModel.validatePassword(for: textField, against: confirmPasswordField))
+    }
```
johnxnguyen: This should be in the test case for `ValidatedTextField`, not the view model.

wire-ios/Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -989,8 +989,13 @@
 
 // Secured Guest link with password
 "secured_guest_link_with_password.header.title" = "Create password secured link";
-"secured_guest_link_with_password.generate_password_button.title" = "Generate password";
 "secured_guest_link_with_password.warning_label.title" = "People who want to join the conversation via the guest link need to enter this password first.\n**You can’t change the password later. Make sure to copy and store it.**";
+"secured_guest_link_with_password.generate_password_button.title" = "Generate password";
+"secured_guest_link_with_password.textfield.header" = "Set password";
+"secured_guest_link_with_password.textfield.footer" = "Use at least 15 characters, with one lowercase letter, one capital letter, a number, and a special character.";
```
johnxnguyen, line 995: The rule says between 15 and 20 characters, but the 20 character limit isn't mentioned here.

wire-ios/Configuration/guestLinkWithPassword_rules.json
johnxnguyen, line 1: Suggestion: `guest_link_password_rules`

wire-ios/Wire-iOS/Sources/Authentication/Helpers/PasscodeRules+Shared.swift
```diff
@@ -29,6 +29,24 @@ extension PasswordRuleSet {
         return try! decoder.decode(PasswordRuleSet.self, from: fileData)
     }()
 
+    /// The guestLinkWithPassword rule set.
+    static let guestLinkWithPasswordRuleSet: PasswordRuleSet? = {
```
johnxnguyen: Suggestion: `guestLinkPassword`

wire-ios-system/Source/Logging/WireLogger.swift
```diff
@@ -174,5 +174,6 @@ public extension WireLogger {
     static let localStorage = WireLogger(tag: "local-storage")
     static let conversation = WireLogger(tag: "conversation")
     static let session = WireLogger(tag: "session")
+    static let passwordRuleSet = WireLogger(tag: "password-rule-set")
```
johnxnguyen: I don't think we need to add this here, it's such a specific logger that you could just create it in the file that you need it.

wire-ios/Wire-iOS/Sources/Authentication/Interface/Helpers/TextFieldValidator.swift
```diff
@@ -51,7 +72,11 @@ final class TextFieldValidator {
         }
     }
 
-    func validate(text: String?, kind: ValidatedTextField.Kind) -> TextFieldValidator.ValidationError? {
+    func validate(
+        text: String?,
+        kind: ValidatedTextField.Kind,
+        ruleSet: PasswordValidationRuleSet = .defaultRuleSet
```
johnxnguyen: This method is to validate some text given a `kind`, meaning the type of validation depends on the kind. But adding `ruleSet` as an argument makes the method appear to validate passwords, but it's not always true. 

I suggest to remove this argument and instead use `kind`.

wire-ios/Wire-iOS/Sources/Authentication/Interface/Helpers/TextFieldValidator.swift
```diff
@@ -35,10 +40,26 @@ final class TextFieldValidator {
...
+                                  isNew: Bool,
+                                  ruleSet: PasswordValidationRuleSet = .defaultRuleSet) -> TextFieldValidator.ValidationError? {
+
+        let currentRuleSet: PasswordRuleSet?
+        switch ruleSet {
+        case .guestLinkRuleSet:
+            currentRuleSet = PasswordRuleSet.guestLinkWithPasswordRuleSet
+        case .defaultRuleSet:
+            currentRuleSet = PasswordRuleSet.shared
+            // add other cases as required
```
johnxnguyen: By adding these, we're making `TextFieldValidator` specific to guest links. Instead, if we inject the rule set, the text field validator doesn't need to know anything about what kind of rule set it is... it just uses it.

wire-ios/Wire-iOS Tests/CreateSecureGuestLinkViewModelTests.swift
agisilaos: Why though? The validatePassword methods belong to the viewModel.

wire-ios/Wire-iOS/Resources/Base.lproj/Localizable.strings
agisilaos: 🤷 It's not mentioned in the design specs and as I talked with Wolfgang he said to not mention it. There is already too much explanation.

wire-ios/Wire-iOS Tests/CreateSecureGuestLinkViewModelTests.swift
johnxnguyen: Ah I misread the test

wire-ios/Wire-iOS/Sources/Authentication/Helpers/PasscodeRules+Shared.swift
agisilaos: Done!

wire-ios/Wire-iOS Tests/CreateSecureGuestLinkViewModelTests.swift
agisilaos: Fixed instead I'm using `FontScheme.configure(with: .large)` directly in the setup method.

wire-ios/Wire-iOS/Sources/Authentication/Helpers/PasscodeRules+Shared.swift
agisilaos: Done!

wire-ios-system/Source/Logging/WireLogger.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/Authentication/Interface/Helpers/TextFieldValidator.swift
agisilaos: So in this case you mean to not use an enum and work with the PasswordRuleSet directly?

wire-ios/Wire-iOS/Resources/Base.lproj/Localizable.strings
johnxnguyen: So what happens when the user tries to enter a 21 character password and it doesn't work? Will they know why? Also... the guest link password rule set is configurable, but this copy is not. Wouldn't the copy be wrong if the guest link password rules change?

wire-ios/Wire-iOS/Sources/Authentication/Interface/Helpers/TextFieldValidator.swift
johnxnguyen: Yep. What benefit does this enum actually give us?

agisilaos: As @KaterinaWire suggested to be more expandable in the future:

https://github.com/wireapp/wire-ios/pull/606#discussion_r1362076407

wire-ios/Wire-iOS/Sources/Authentication/Interface/Helpers/TextFieldValidator.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/Authentication/Interface/Helpers/TextFieldValidator.swift
agisilaos: Done!

wire-ios/Configuration/guestLinkWithPassword_rules.json
agisilaos: I removed it entirely as we discussed.

wire-ios/Wire-iOS/Sources/Authentication/Interface/Helpers/TextFieldValidator.swift
KaterinaWire: > As @KaterinaWire suggested to be more expandable in the future:
> 
> [#606 (comment)](https://github.com/wireapp/wire-ios/pull/606#discussion_r1362076407)

This suggestion was based on a previous implementation.

wire-ios-system/Source/Logging/WireLogger.swift
```diff
@@ -174,5 +174,4 @@ public extension WireLogger {
     static let localStorage = WireLogger(tag: "local-storage")
     static let conversation = WireLogger(tag: "conversation")
     static let session = WireLogger(tag: "session")
-
```
KaterinaWire: I would keep this empty line

wire-ios/Wire-iOS/Sources/Authentication/Interface/Descriptions/ViewDescriptions/EmailPasswordFieldDescription.swift
```diff
@@ -38,7 +38,8 @@ final class EmailPasswordFieldDescription: ValueSubmission {
 
 extension EmailPasswordFieldDescription: ViewDescriptor {
     func create() -> UIView {
-        textField.passwordField.kind = .password(isNew: forRegistration)
+        let rules = forRegistration ? PasswordRuleSet.accountRegistration : .nonEmpty
```
KaterinaWire:
```suggestion
        let rules: PasswordRuleSet = forRegistration ? .accountRegistration : .nonEmpty
```

wire-ios/Wire-iOS/Sources/Authentication/Interface/Descriptions/ViewDescriptions/EmailPasswordFieldDescription.swift
```diff
@@ -38,7 +38,8 @@ final class EmailPasswordFieldDescription: ValueSubmission {
...
+        textField.passwordField.kind = ValidatedTextField.Kind.password(rules, isNew: forRegistration)
```
KaterinaWire:
```suggestion
        textField.passwordField.kind = .password(rules, isNew: forRegistration)
```

wire-ios/Wire-iOS/Sources/Authentication/Interface/Views/ValidatedTextField.swift
KaterinaWire: do you still use it?

wire-ios-system/Source/Logging/WireLogger.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/Authentication/Interface/Descriptions/ViewDescriptions/EmailPasswordFieldDescription.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/Authentication/Interface/Descriptions/ViewDescriptions/EmailPasswordFieldDescription.swift
agisilaos: Done!

wire-ios-request-strategy/Sources/Request Strategies/User/UserProfileRequestStrategy.swift
```diff
@@ -80,9 +80,8 @@ public class UserProfileRequestStrategy: AbstractRequestStrategy, IdentifierObje
             syncProgress.finishCurrentSyncPhase(phase: .fetchingUsers)
         } else {
             fetch(users: allConnectedUsers, for: apiVersion)
+            isFetchingAllConnectedUsers = true
```
KaterinaWire: Are these changes part of this PR?

agisilaos: Nope that's fixed.

wire-ios/Wire-iOS/Sources/Authentication/Interface/Views/ValidatedTextField.swift
agisilaos: Yes!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreatePasswordSecuredLink/CreateSecureGuestLinkViewController.swift
```diff
@@ -105,16 +206,71 @@ class CreateSecureGuestLinkViewController: UIViewController, CreatePasswordSecur
...
+        }
+
+        for label in labels {
+            label.textColor = SemanticColors.Label.textFieldFloatingLabel
+        }
+
+    }
+
+    @objc
+    func handlePasswordValidation(for textField: ValidatedTextField) -> Bool {
```
KaterinaWire: Do we need to return a value? This method is for textField, right?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreatePasswordSecuredLink/CreateSecureGuestLinkViewController.swift
```diff
@@ -105,16 +206,71 @@ class CreateSecureGuestLinkViewController: UIViewController, CreatePasswordSecur
...
+        let labels: [UILabel] = textField == securedGuestLinkPasswordTextfield ? [passwordRequirementsLabel, setPasswordLabel] : [confirmPasswordLabel]
+
+        let isValid = viewModel.validatePassword(for: textField, against: securedGuestLinkPasswordTextfield)
```
KaterinaWire: here we check if `securedGuestLinkPasswordTextfield` is valid, but the method applies to `securedGuestLinkPasswordValidatedTextField` too. Should be a different validation for `securedGuestLinkPasswordValidatedTextField` ?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreatePasswordSecuredLink/CreateSecureGuestLinkViewController.swift
agisilaos: I don't think we need to return a value. I think the initial implementation was like that and then I forgot to remove it. Done!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreatePasswordSecuredLink/CreateSecureGuestLinkViewController.swift
agisilaos: It works both ways. There is no need for different validation for securedGuestLinkPasswordValidatedTextField

We do all checks in:

```swift
func validatePassword(
        for textField: ValidatedTextField,
        against confirmPasswordField: ValidatedTextField
    ) -> Bool {

        guard let enteredPassword = textField.text,
              !enteredPassword.isEmpty,
              textField.isValid,
              confirmPasswordField.text == enteredPassword else {
            return false
        }

        return true
    }
```

agisilaos: The textField (which is an instance of ValidatedTextField) is set to call the handlePasswordValidation(for:) method on the target self whenever its text changes (due to the .editingChanged event). When this happens, the textField itself (the sender) gets passed as the argument to the handlePasswordValidation(for:) method.

So in this case:

```swift
private lazy var securedGuestLinkPasswordValidatedTextField: ValidatedTextField = {
        let textField = ValidatedTextField(
            kind: .password(.guestLinkPassword, isNew: true),
            leftInset: 8,
            accessoryTrailingInset: 0,
            cornerRadius: 12,
            setNewColors: true,
            style: .default
        )
        textField.showConfirmButton = false
        textField.addTarget(self, action: #selector(handlePasswordValidation(for:)), for: .editingChanged)
        textField.addRevealButton(delegate: self)
        textField.placeholder = SecuredGuestLinkWithPasswordLocale.VerifyPasswordTextField.placeholder
        textField.addDoneButtonOnKeyboard()
        textField.returnKeyType = .done

        return textField
    }()
 ```
    
`securedGuestLinkPasswordValidatedTextField` is being passed as the argument to the handlePasswordValidation method. 


And in the case of the `securedGuestLinkPasswordTextField`, `securedGuestLinkPasswordTextField` is being passed as the argument to the handlePasswordValidation method.

File: wire-ios/Wire-iOS Tests/AccessoryTextField/AccessoryTextFieldValidationTests.swift
Comment (johnxnguyen) [inline wire-ios/Wire-iOS Tests/AccessoryTextField/AccessoryTextFieldValidationTests.swift:1]: These tests are testing password validation, but now that validation is injected (via password rules), I don't think they make sense to be here anymore.

In other words, tests in this file should be focused on the `ValidatedTextField` logic, and not the password rules validation. E.g, test that if

File: wire-ios/Wire-iOS Tests/CreateSecureGuestLinkViewModelTests.swift [hunk 21]
```diff
@@ -17,34 +17,82 @@
...
     var viewModel: CreateSecureGuestLinkViewModel!
     var mockDelegate: MockCreatePasswordSecuredLinkViewModelDelegate!
+    var textField: ValidatedTextField!
+    var confirmPasswordField: ValidatedTextField!
 
     // MARK: - setUp
 
     override func setUp() {
         super.setUp()
+        FontScheme.configure(with: .large)
```
johnxnguyen: As I've suggested in the past, this setup could be done in a base test class which all tests inherit.

wire-ios/Wire-iOS/Resources/Base.lproj/Localizable.strings
johnxnguyen: Is there any update on what design wants to do regarding the copy?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreatePasswordSecuredLink/CreateSecureGuestLinkViewController.swift
```diff
@@ -41,24 +55,79 @@ class CreateSecureGuestLinkViewController: UIViewController, CreatePasswordSecur
...
+        let textField = ValidatedTextField(
+            kind: .password(.guestLinkPassword, isNew: true),
+            leftInset: 8,
+            accessoryTrailingInset: 0,
+            cornerRadius: 12,
+            setNewColors: true,
+            style: .default
+        )
+        textField.addRevealButton(delegate: self)
+        textField.addTarget(self, action: #selector(handlePasswordValidation(for:)), for: .editingChanged)
```
johnxnguyen: This is diverging from the intended use of `ValidatedTextField`. If you want to be informed of validation when the text changes, you should set a `TextFieldValidationDelegate`. You could register the view model as the delegate.

wire-ios/Wire-iOS/Resources/Base.lproj/Localizable.strings
agisilaos: Nope!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreatePasswordSecuredLink/CreateSecureGuestLinkViewController.swift
agisilaos: You pushed the initial changes after our call and I followed up with a little bit of refactoring. Done!

wire-ios/Wire-iOS Tests/CreateSecureGuestLinkViewModelTests.swift
agisilaos: Done!

johnxnguyen (approved):

KaterinaWire (approved):

PR #615, opened 2023-10-24
feat:  Extract getting client fingerprint into a use case - WPB-5134

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5134" title="WPB-5134" target="_blank">TaskWPB-5134</a>  [iOS] Extract getting client fingerprint into a use case
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

In order to prepare CoreCrypto update to 1.0, we refactor part of the code. This PR moves logic from UserClient regarding fingerprints.

### Solutions

* remove fingerprintData property
* add UseCase for:
   *  fetching remote fingerprint asynchronously
   *  retrieving local fingerprint asynchronously

### Notes

* Added WireSyncEngineSupport to expose mocks of SyncEnince's protocols to Wire-iOS test targets.

#### Test Coverage (Optional)

----

----

----
##### References


wire-ios-data-model/Tests/Source/Utils/InvalidClientsRemovalTests.swift
```diff
@@ -117,6 +117,7 @@ class InvalidClientsRemovalTests: DiskDatabaseTest {
             syncMOC.zm_cryptKeyStore.encryptionContext.perform {
                 XCTAssertTrue($0.hasSession(for: clientId))
             }
+            // TODO: Ask why we check these - how is it linked to fingerpint?
             XCTAssertTrue(otherClient.hasSessionWithSelfClient)
             XCTAssertFalse(otherClient.isZombieObject)
```
netbe: these two assertions are failing but I don't get how it is related to my changes (fingerprint)?

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Devices/ProfileClientViewModel.swift
```diff
@@ -0,0 +1,51 @@
...
+
+final class ProfileClientViewModel {
+    let userClient: UserClient
+    private let fingerprintUseCase: FingerprintUseCase
+    private (set) var fingerprintData: Data?
+
+    var fingerprintDataClosure: ((Data?) -> Void)?
+
+    init(userClient: UserClient, fingerprintUseCase: FingerprintUseCase? = nil) {
+        guard let useCase = fingerprintUseCase ?? ZMUserSession.shared()?.fingerprintUseCase else {
```
typfel: Why not inject an non-optional `FingerprintUseCase`?

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Devices/ProfileClientViewModel.swift
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    func loadData() {
+        let isSelfClient = userClient.isSelfClient()
+        Task {
+            if isSelfClient {
+                self.fingerprintData = await fingerprintUseCase.localFingerprint()
+            } else {
+                self.fingerprintData = await fingerprintUseCase.fetchRemoteFingerprint(for: userClient)
+            }
```
typfel: This logic would preferably be inside the use case

wire-ios-data-model/Source/Public/FingerprintUseCase.swift
```diff
@@ -0,0 +1,112 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public class FingerprintUseCase {
```
typfel: I think it's preferred to name the use case as an action: `GetUserClientFingerprintUseCase`

wire-ios-data-model/Source/Public/FingerprintUseCase.swift
```diff
@@ -0,0 +1,112 @@
...
+
+        return FingerprintUseCase(proteusProvider: proteusProvider, managedObjectContext: managedObjectContext)
+    }
+
+    init(proteusProvider: ProteusProviding, managedObjectContext: NSManagedObjectContext) {
+        self.proteusProvider = proteusProvider
+        self.managedObjectContext = managedObjectContext
+    }
+
+    public func localFingerprint() async -> Data? {
```
typfel: The use case should only do one thing so there should only be one function exposed: `invoke() async -? Data?`

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
```diff
@@ -233,6 +233,8 @@ public class ZMUserSession: NSObject {
         tornDown = true
     }
 
+    public var fingerprintUseCase: FingerprintUseCase!
```
typfel: suggestion: By convention Kalium drops the `UseCase` suffix on the `var` since it reads better.

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Devices/ProfileClientViewController.swift
```diff
@@ -502,10 +506,6 @@ extension ProfileClientViewController: UserClientObserver {
 
     func userClientDidChange(_ changeInfo: UserClientChangeInfo) {
 
-        if changeInfo.fingerprintChanged {
-            updateFingerprintLabel()
-        }
```
typfel, line 513: Ok, so we no longer have a way of observing when the fingerprint changes, on the other hand if it did change that would be bug IMO since that would be mean that the remote user has replaced his existing proteus client with a new one.

So I think the observing was more for when it went from `null` to `something`. Do we handle scenario now as well?

typfel (changes requested):

wire-ios-data-model/Tests/Source/Utils/InvalidClientsRemovalTests.swift
typfel: they shouldn't be but if `XCTAssertFalse(otherClient.isZombieObject)` is failing then I would also expect `otherClient.hasSessionWithSelfClient` to fail.

Now `otherClient` is deleted and before it wasn't?

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Devices/ProfileClientViewController.swift
netbe: right now, no. Does it translate as we get a new UserClient?

wire-ios-data-model/Source/Public/FingerprintUseCase.swift
netbe: ok I'll move the logic from the viewModel and have it public.

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
netbe: so:
```
public var getUserClientFingerprint: GetUserClientFingerprintUseCase!
```
?

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Devices/ProfileClientViewModel.swift
netbe: done

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Devices/ProfileClientViewModel.swift
netbe: yes, but currently we don't have a good way guarantee this

typfel: How come? You aren't guaranteed to have a user session here?

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
typfel: Yes, exactly like that

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Devices/ProfileClientViewController.swift
typfel: I think it would be this scenario:

1. User A adds a new client (A2) but no message is sent so no session exists
2. You navigate the the device list of user A and look at A2

I suppose we establish a session with A2 at upon navigating to the client details screen but if this is async I suppose the fingerprint will not be visible initially.

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Devices/ProfileClientViewModel.swift
netbe: in theory, yes
in practice, we have a couple of issues where UserSession is not here actually, see https://github.com/wireapp/wire-ios/pull/619 for example.

Do you mean to force unwrap the usersession at the init call site ?

wire-ios-data-model/Tests/Source/Utils/InvalidClientsRemovalTests.swift
netbe: removed as it does not make sense since we have the duplicate clients fix

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Devices/ProfileClientViewModel.swift
johnxnguyen: Soon we will have a guaranteed user session available when https://github.com/wireapp/wire-ios/pull/616 is merged and cherry picked over to `develop`.

wire-ios-data-model/Source/Model/UserClient/UserClient.swift
johnxnguyen, line 1: It's very satisfying to read these changes 😆

wire-ios-data-model/Source/Public/GetUserClientFingerprintUseCase.swift
```diff
@@ -0,0 +1,129 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public class GetUserClientFingerprintUseCase {
+
+    let proteusProvider: ProteusProviding
+    let managedObjectContext: NSManagedObjectContext
```
johnxnguyen: Just a suggestion: to reduce the amount of reading and line lengths, I think simply `context` is enough.

wire-ios-data-model/Source/Public/GetUserClientFingerprintUseCase.swift
```diff
@@ -0,0 +1,129 @@
...
+
+    public static func create(for managedObjectContext: NSManagedObjectContext) -> GetUserClientFingerprintUseCase {
```
johnxnguyen: Why a static func rather than a convenience init?

wire-ios-data-model/Source/Public/GetUserClientFingerprintUseCase.swift
```diff
@@ -0,0 +1,129 @@
...
+    init(proteusProvider: ProteusProviding, managedObjectContext: NSManagedObjectContext) {
+        self.proteusProvider = proteusProvider
+        self.managedObjectContext = managedObjectContext
+    }
+
+    public func invoke(userClient: UserClient) async -> Data? {
+        let objectId = userClient.objectID
+
+        let isSelfClient: Bool = managedObjectContext.performAndWait {
+            guard let existingUserId = try? managedObjectContext.existingObject(with: objectId) as? UserClient else {
```
johnxnguyen: Suggestion: `existingClient`

wire-ios-data-model/Source/Public/GetUserClientFingerprintUseCase.swift
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    func localFingerprint() async -> Data? {
+        var canPerform: Bool = false
+
+        managedObjectContext.performAndWait {
+            canPerform = proteusProvider.canPerform
+        }
+        guard canPerform else {
+            return nil
```
johnxnguyen: Would it be helpful to throw an error or we don't really care what went wrong?

wire-ios-data-model/Source/Public/GetUserClientFingerprintUseCase.swift
```diff
@@ -0,0 +1,129 @@
...
+            userClientToUpdate = managedObjectContext.object(with: userClientObjectId) as? UserClient
+            if proteusProvider.canPerform {
+                sessionId = userClientToUpdate?.proteusSessionID
+            }
+        }
+
+        guard let sessionID = sessionId
+        else {
+            return nil
+        }
```
johnxnguyen:
```suggestion
        guard let sessionID = sessionId else {
            return nil
        }
```

wire-ios-data-model/Tests/Source/ManagedObjectContext/DatabaseMigrationTests.swift
```diff
@@ -428,7 +428,7 @@ final class DatabaseMigrationTests: DatabaseBaseTest {
 
             let store = NSManagedObjectModel(contentsOf: source.appendingPathComponent(modelFileName))!
             // then
-            XCTAssertTrue(store.versionIdentifiers.contains(version))
+            XCTAssertTrue(store.versionIdentifiers.contains(version), "\(version) should be contained")
```
johnxnguyen, line 405: What should one do to correct this? I think stating that in the message would be helpful.

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
```diff
@@ -233,6 +233,8 @@ public class ZMUserSession: NSObject {
         tornDown = true
     }
 
+    public var getUserClientFingerprint: GetUserClientFingerprintUseCase!
+
```
johnxnguyen: Could this be lazy instead of implicitly unwrapped?

wire-ios-data-model/Tests/Source/ManagedObjectContext/DatabaseMigrationTests.swift
netbe: added more details

wire-ios-data-model/Source/Public/GetUserClientFingerprintUseCase.swift
netbe: added a log and kept it non-throwing for now

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
netbe: I guess it could yes :)

wire-ios-data-model/Source/Model/UserClient/UserClient.swift
```diff
@@ -527,125 +507,6 @@ public extension UserClient {
...
-                return
-            }
-
-            if syncSelfClient == syncClient {
-                syncClient.fingerprint = syncClient.localFingerprint()
-                syncMOC.saveOrRollback()
-            } else {
-                if !syncClient.hasSessionWithSelfClient {
-                    syncSelfClient.missesClient(syncClient)
-                    syncSelfClient.setLocallyModifiedKeys(Set(arrayLiteral: ZMUserClientMissingKey))
```
netbe, line 538: I assume it is what the SessionEstablisher replaces in establishSession. @typfel ?

typfel: Exactly!

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
```diff
@@ -233,6 +233,12 @@ public class ZMUserSession: NSObject {
         tornDown = true
     }
 
+    public lazy var getUserClientFingerprint: GetUserClientFingerprintUseCase = {
+        // Note: this is safe if coredataStack and proteus are ready
```
johnxnguyen: Is this just a warning to not call this too early?

wire-ios-sync-engine/Source/Use cases/GetUserClientFingerprintUseCase.swift
```diff
@@ -0,0 +1,158 @@
...
+            }
+        }
+
+        guard let existingUser else { return nil }
+
+        let isSelfClient: Bool = await context.perform { existingUser.isSelfClient() }
+
+        let canPerform: Bool = await context.perform {
+            self.proteusProvider.canPerform
+        }
```
johnxnguyen:
```suggestion
        let isSelfClient = await context.perform { 
            existingUser.isSelfClient() 
        }

        let canPerform  = await context.perform {
            self.proteusProvider.canPerform
        }
```

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
netbe: yes, I'll move it above so we can see it in the docs. But I guess we'll make it safer once CC setup is in place

wire-ios-data-model/Source/Public/GetUserClientFingerprintUseCase.swift
netbe: replaced by convenience init

wire-ios/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/SettingsClientViewController.swift
```diff
@@ -162,7 +172,7 @@ final class SettingsClientViewController: UIViewController,
     }
 
     @objc func onVerifiedChanged(_ sender: UISwitch!) {
-        guard let userSession = ZMUserSession.shared() else { return }
+        guard let userSession = userSession as? ZMUserSession else { return }
```
johnxnguyen: Instead of casting back, it'd be better to add `selfUserClient` to `UserSession`.

netbe: Done, though should selfUserClient be of type UserType?

wire-ios-sync-engine/Source/Use cases/GetUserClientFingerprintUseCase.swift
```diff
@@ -0,0 +1,165 @@
...
+        return fingerprintData
+    }
+
+    // MARK: - Helpers
+
+    private func proteusPerform<T>(
+        withProteusService proteusServiceBlock: ProteusServicePerformBlock<T>,
+        withKeyStore keyStoreBlock: KeyStorePerformBlock<T>
+    ) rethrows -> T {
+        return try context.performAndWait {
```
typfel: This runs in an async context so I think we should replace i with `await perform()`

wire-ios-sync-engine/sourcery/config.yml
```diff
@@ -5,6 +5,6 @@ templates:
 output:
   ./generated
 args:
-  autoMockableTestableImports: ["WireSyncEngine"]
-  autoMockableImports: []
+  autoMockableTestableImports: []
+  autoMockableImports: ["WireSyncEngine"]
```
typfel, line 9: Why do we not want to import WireSyncEngine as @testable anymore?

typfel (changes requested):

wire-ios-sync-engine/sourcery/config.yml
netbe: the mocks generated are inside `WireSyncEngineSupport` so it just need and `import "WireSyncEngine"`

wire-ios-sync-engine/Source/Use cases/GetUserClientFingerprintUseCase.swift
netbe: updated the function

johnxnguyen (approved):

typfel (approved):

PR #616, opened 2023-10-24
fix: shared user session crashes WPB-4920

johnxnguyen:
<table>
<td>
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This is a feature branch for a collection of refactorings needed to prevent crashes when trying to access the shared user session.

### Issues

### Causes

The UI code depends heavily on the active user session, which is accessed via a singleton `ZMUserSession.shared()`. This is often safely unwrapped, but there are places where it is not, leading to a a crash. The most common point of crash is when the the static function `ZMUser.selfUser()` is accessed, which depends on the shared user session and implicitly unwraps its value.

### Solutions

### Note

#### Test Coverage

- [ ] Updated tests.

#### How to Test

- Be logged in.
- Remove your current device from another device, causing the current device to be logged out
- Assert you don't crash

You may need to repeat this a few times.

----

----

----
##### References


wire-ios-data-model/Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -116,7 +116,11 @@ public final class AppLockController: AppLockType {
 
     // MARK: - Life cycle
 
-    public init(userId: UUID, selfUser: ZMUser, legacyConfig: LegacyConfig? = nil) {
+    public init(
+        userId: UUID,
+        selfUser: ZMUser,
+        legacyConfig: LegacyConfig?
```
caldrian, line 122: I like that you removed the default value, I think default values can lead to unsafe code.

netbe (approved):

wire-ios-sync-engine/Source/UserSession/UserSession.swift
```diff
@@ -0,0 +1,505 @@
...
+
+    func enqueue(
+        _ changes: @escaping () -> Void,
+        completionHandler: (() -> Void)?
+    )
+
+    // TODO: rename to "shouldHideNotificationContent"
+    var isNotificationContentHidden: Bool { get set }
+
+    // TODO: rename to "isEncryptionAtRestEnabled"
```
caldrian, line 135: Should we add a ticket, a date or some other reference for when these `TODOs` should be solved?

johnxnguyen (commented):
@agisilaos do you still need to revert the fix for Xcode 15 errors?

agisilaos:
> @agisilaos do you still need to revert the fix for Xcode 15 errors?

@johnxnguyen Yep! Those changes have been reverted.

wire-ios-sync-engine/Source/UserSession/UserSession.swift
agisilaos: I will take care of it today, while I create the tickets for the rest of the work related to the userSession.

wire-ios/Wire-iOS Tests/ConversationMessageCell/ConversationMessageSnapshotTestCase.swift
```diff
@@ -32,7 +33,26 @@ extension ConversationMessageContext {
                                                                        spacing: 0)
 }
 
-extension XCTestCase {
+func XCTAssertArrayEqual(_ descriptions: [Any], _ expectedDescriptions: [Any], file: StaticString = #file, line: UInt = #line) {
```
caldrian, line 36: I suggest to remove this helper. I think it is already being used in a wrong way. I will create a PR, if you agree? @agisilaos

agisilaos: I can take a look on how it's being used but sure, please do open a separate PR. If not please do create a ticket so we can tackle it later.

agisilaos (approved):

wire-ios/Wire-iOS/Sources/Components/Settings/ZMUserSessionInterface.swift
```diff
@@ -20,6 +20,7 @@ import WireSyncEngine
 
 typealias UserSessionInterface = UserSessionSwiftInterface & UserSessionAppLockInterface
 
+// TODO: delete
```
caldrian, line 23: Same here, could we specify when and/or why this needs to be done?

PR #621, opened 2023-10-25
refactor: remove user session singleton (part 3) - WPB-4920

agisilaos:
<table>
<td>
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Often the app crashes shortly after logging out or the session expires.

## Causes

## Solutions

## Note

----

----

----
##### References


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+DragAndDrop.swift
```diff
@@ -47,7 +47,10 @@ extension ConversationInputBarViewController: UIDropInteractionDelegate {
                                                                          onConfirm: { [unowned self] _ in
                                                                             self.dismiss(animated: true) {
                                                                                 if let draggedImageData = draggedImage.pngData() {
-                                                                                    self.sendController.sendMessage(withImageData: draggedImageData)
+                                                                                    self.sendController.sendMessage(
+                                                                                        withImageData: draggedImageData,
+                                                                                        userSession: self.userSession
+                                                                                    )
```
johnxnguyen: Indentation seems wrong here

johnxnguyen (approved):

netbe (approved):

PR #625, opened 2023-10-27
refactor: remove user session singleton (part 4) - WPB-4920

agisilaos:
<table>
<td>
  </td></table>
  <br />
 

----

- The **PR Title**
  - [x] conforms to the style of semantic commits messages supported in Wire's Github Workflow²

- The **PR Description**
----

# What's new in this PR?

## Issues

## Causes

## Solutions

## Note

----
##### References


wire-ios-sync-engine/Source/UserSession/ZMUserSession+SecurityClassification.swift
johnxnguyen, line 1: You can leave this file as is, just make the first method public.

wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallController/ActiveCallRouter.swift
```diff
@@ -100,9 +100,9 @@ extension ActiveCallRouter: ActiveCallRouterProtocol {
         // NOTE: We resign first reponder for the input bar since it will attempt to restore
         // first responder when the call overlay is interactively dismissed but canceled.
         UIResponder.currentFirst?.resignFirstResponder()
-
+        guard let zclient = zClientViewController else { return }
```
johnxnguyen: Can we inject the user session instead of getting the zclientviewcontroller?

wire-ios-sync-engine/Source/UserSession/ZMUserSession+SecurityClassification.swift
agisilaos: I moved that method to our UserSession protocol. Should I move it back here?

wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallController/ActiveCallRouter.swift
agisilaos: Sure! Just fixed that!

wire-ios-sync-engine/Source/UserSession/UserSession.swift
```diff
@@ -0,0 +1,318 @@
+//
```
netbe: isn't it from last PR?

wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallViewController.swift
```diff
@@ -387,13 +390,12 @@ extension CallViewController: ZMConversationObserver {
...
         guard
             changeInfo.participantsChanged,
-            let userSession = ZMUserSession.shared(),
             let participants = conversation?.participants
         else {
             return
         }
 
-        classification = userSession.classification(with: participants)
+        classification = userSession.classification(with: participants, conversationDomain: nil)
```
netbe, line 398: supposed to be nil?

agisilaos: Nil was its initial value, yes!

netbe (approved):

johnxnguyen (approved):

PR #628, opened 2023-10-27
feat: periodically check if migration can start WPB-2895

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-2895" title="WPB-2895" target="_blank">StoryWPB-2895</a>  [iOS] Periodically check if migration can be initialised
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

To ensure the MLS migration starts and ends at specific times, we need to periodically check (every 24h) if the migration can start. Once it's started, we also need to periodically check if it should be finalised.

For this purpose I've introduced the `ProteusToMLSMigrationCoordinator`. Its responsibility is to keep track of the migration status and to initiate migration steps when necessary. In the scope of this PR, we're only implementing the coordinator's logic for starting the migration. Finalising the migration will come at a later time.

To satisfy the periodic check requirement, the coordinator is asked to update the migration status every 24h through a recurring action.

### Changes

- Introduced `ProteusToMLSMigrationCoordinator`
- Introduced `ProteusToMLSMigrationStorage` to handle storing the migration status per account
- Added the `Feature.MLSMigration` feature
- Updated `Feature.MLS`
- Added interface to start migration to `MLSService`
- Registered new `RecurringAction` to periodically update the migration status
- Handled `mls-not-enabled` error for `GET /mls/public-keys`
- Updated some utilities

### Testing

- Added tests for `ProteusToMLSMigrationCoordinator`
- Updated tests related to Feature management

### DIFF
--- a/wire-ios-data-model/Source/MLS/Migration/ProteusToMLSMigrationCoordinator.swift
+++ b/wire-ios-data-model/Source/MLS/Migration/ProteusToMLSMigrationCoordinator.swift
@@ -21,7 +21,7 @@ import WireUtilities
 import WireTransport
 
 public protocol ProteusToMLSMigrationCoordinating {
-    func updateMigrationStatus()
+    func updateMigrationStatus() async
 }
 
 public class ProteusToMLSMigrationCoordinator: ProteusToMLSMigrationCoordinating {
@@ -86,12 +86,10 @@ public class ProteusToMLSMigrationCoordinator: ProteusToMLSMigrationCoordinating
 
     // MARK: - Public Interface
 
-    public func updateMigrationStatus() {
+    public func updateMigrationStatus() async {
         switch storage.migrationStatus {
         case .notStarted:
-            Task {
-                await startMigrationIfNeeded()
-            }
+            await startMigrationIfNeeded()
         case .started:
             // check if it should be finalised
             break
--- a/wire-ios-data-model/Source/MLS/Migration/ProteusToMLSMigrationCoordinator.swift
+++ b/wire-ios-data-model/Source/MLS/Migration/ProteusToMLSMigrationCoordinator.swift
@@ -35,7 +35,7 @@ public class ProteusToMLSMigrationCoordinator: ProteusToMLSMigrationCoordinating
         case finalised
     }
 
-    enum MigrationStartStatus {
+    enum MigrationStartStatus: Equatable {
         case canStart
         case cannotStart(reason: CannotStartMigrationReason)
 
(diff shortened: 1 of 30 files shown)


wire-ios-data-model/Source/MLS/Migration/ProteusToMLSMigrationCoordinator.swift
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    // MARK: - Public Interface
+
+    public func updateMigrationStatus() {
+        switch storage.migrationStatus {
+        case .notStarted:
+            Task {
+                await startMigrationIfNeeded()
+            }
```
netbe: what about making the func async instead of using Task

wire-ios-data-model/Source/Model/FeatureConfig/FeatureRepository.swift
```diff
@@ -69,12 +73,12 @@ public class FeatureRepository: FeatureRepositoryInterface {
               let featureConfig = feature.config else {
                   return .init()
               }
-        let config = try! JSONDecoder().decode(Feature.AppLock.Config.self, from: featureConfig)
+        let config = try! decoder.decode(Feature.AppLock.Config.self, from: featureConfig)
```
netbe, line 76: Does it mean we crash if we fail to decode the config from backend with the force unwrapped ?
It does not align with how we handle the guard else above. Maybe we do the same returning a default value

wire-ios-data-model/Source/Model/FeatureConfig/FeatureRepository.swift
```diff
@@ -239,19 +243,42 @@ public class FeatureRepository: FeatureRepositoryInterface {
...
+        else {
+            return .init()
+        }
+
+        let config = try! decoder.decode(Feature.MLSMigration.Config.self, from: featureConfig)
+        return .init(status: feature.status, config: config)
+    }
+
+    public func storeMLSMigration(_ mlsMigration: Feature.MLSMigration) {
+        let config = try! encoder.encode(mlsMigration.config)
```
netbe, line 274: same goes here, be better to throw  or catch the error

wire-ios-data-model/Source/Model/FeatureConfig/FeatureRepository.swift
David-Henner: Not exactly. Yes, we would crash if we fail to decode. But the config is not decoded from the backend. It's decoded from the local database. 

The flow for fetching a feature is:
- Fetch feature on the backend
- Try to decode it, throw if it fails (see `FeatureConfigRequestStrategy`)
- Encode and store it into the database

Then when we need it we fetch it from the database. Unless we use different types for encoding / decoding, it's unlikely that decoding will fail.

I don't know what the reason was for force unwrapping, maybe convenience or making sure we catch programming errors? But I guess we could return a default value

wire-ios-data-model/Source/MLS/Migration/ProteusToMLSMigrationCoordinator.swift
David-Henner: done [4958715](https://github.com/wireapp/wire-ios/pull/628/commits/49587158acd457ec067da0f4d6df2cb4312fab79)

netbe (approved):

johnxnguyen (approved):
Very nice 👍🏻

PR #631, opened 2023-10-27
refactor: remove user session singleton (part 5) - WPB-4920

agisilaos:
<table>
<td>
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

## Issues

## Causes

## Solutions

## Note

----

----

----
##### References


wire-ios/Wire-iOS/Sources/UserInterface/CustomAppLock/Unlock/UnlockViewController.swift
```diff
@@ -32,7 +32,7 @@ protocol UnlockViewControllerDelegate: AnyObject {
 /// This VC should be wrapped in KeyboardAvoidingViewController as the "unlock" button would be covered on 4 inch iPhone
 final class UnlockViewController: UIViewController {
```
johnxnguyen, line 34: It'd be better if we only use `UserSession` and not the other protocols.

wire-ios/Wire-iOS/Sources/UserInterface/StartUI/StartUI/StartUIViewController.swift
```diff
@@ -207,7 +207,7 @@ final class StartUIViewController: UIViewController, SpinnerCapable {
 
     func showKeyboardIfNeeded() {
         // TODO: unwrap
```
netbe: remove todo

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Devices/UserClientListViewController.swift
```diff
@@ -128,7 +130,11 @@ final class UserClientListViewController: UIViewController, UICollectionViewDele
 
     func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
         let client = clients[indexPath.row]
-        let profileClientViewController = ProfileClientViewController(client: client as! UserClient, fromConversation: true) // TODO jacob don't force unwrap
+        let profileClientViewController = ProfileClientViewController(
+            client: client as! UserClient,
+            fromConversation: true, 
+            userSession: userSession
+        ) // TODO jacob don't force unwrap
```
netbe:
``` suggestion
        guard let client = clients[indexPath.row] as? UserClient else { return }
        let profileClientViewController = ProfileClientViewController(
            client: client,
            fromConversation: true, 
            userSession: userSession
        )
```

netbe (approved):
just some cleanup todos

wire-ios/Wire-iOS/Sources/UserInterface/CustomAppLock/Unlock/UnlockViewController.swift
agisilaos: I've done that in part 6. So for now we're good! @johnxnguyen

wire-ios/Wire-iOS/Sources/UserInterface/StartUI/StartUI/StartUIViewController.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Devices/UserClientListViewController.swift
agisilaos: Done!

johnxnguyen (approved):

PR #652, opened 2023-11-08
fix: App crashes when searching for the last message in a conversation - WPB-5398

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5398" title="WPB-5398" target="_blank">BugWPB-5398</a>  [iOS] App crashes when searching for the last message in a conversation
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

When it comes to cells we should avoid passing the userSession into the initializer or anything else. Instead, we should use a configure method such as we do in the case of `TextSearchResultCell`. That was the source of the crash. With the changes in this PR, the crash has been fixed. 

----
##### References

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
@@ -70,7 +70,7 @@ final class CollectionsViewController: UIViewController {
         return sections == .all
     }
 
-    private lazy var textSearchController: TextSearchViewController =  TextSearchViewController(conversation: collection.conversation)
+    private lazy var textSearchController =  TextSearchViewController(conversation: collection.conversation, userSession: userSession)
 
     convenience init(conversation: ZMConversation, userSession: UserSession) {
         let matchImages = CategoryMatch(including: .image, excluding: .GIF)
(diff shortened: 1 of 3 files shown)


netbe (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
```diff
@@ -70,7 +70,7 @@ final class CollectionsViewController: UIViewController {
         return sections == .all
     }
 
-    private lazy var textSearchController: TextSearchViewController =  TextSearchViewController(conversation: collection.conversation)
+    private lazy var textSearchController =  TextSearchViewController(conversation: collection.conversation, userSession: userSession)
```
KaterinaWire:
```suggestion
    private lazy var textSearchController = TextSearchViewController(conversation: collection.conversation, userSession: userSession)
```

markusfassbender (approved):
Probably as a next step think we should avoid all kinds of `let userSession: UserSession` and retaining references to it.. rather than using a singleton it might be worth to setup a dependency injection with a shared access via service.

agisilaos:
> Probably as a next step think we should avoid all kinds of `let userSession: UserSession` and retaining references to it.. rather than using a singleton it might be worth to setup a dependency injection with a shared access via service.

On a later stage we can explore such an idea, I agree.

KaterinaWire (approved):

PR #662, opened 2023-11-09
fix: Method should not be called on the main thread as it may lead to UI Unresponsiveness

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

As I was working on a different issue I stumbled upon:

Screenshot 2023-11-08 at 10 10 59

With this PR I got rid of that warning by refactoring that method and updating the tests. 

----
##### References

### DIFF
--- a/wire-ios/Wire-iOS Tests/MessagePresenterTests.swift
+++ b/wire-ios/Wire-iOS Tests/MessagePresenterTests.swift
@@ -76,14 +76,14 @@ final class MessagePresenterTests: XCTestCase {
 
     // MARK: - Pass
 
-    func testThatCreateAddPassesViewControllerThrowsAnErrorForFileMessage() async throws {
+    func testThatMakePassesViewControllerThrowsErrorForInvalidFileURL() async throws {
         // GIVEN
-        let message = MockMessageFactory.fileTransferMessage()
-        let fileMessageData = try XCTUnwrap(message.fileMessageData)
+        let message = MockMessageFactory.passFileTransferMessage()
+        let fileURL = try XCTUnwrap(URL(string: "https://apple.com"))
 
         // WHEN && THEN
         do {
-            _ = try await sut.makePassesViewController(fileMessageData: fileMessageData)
+            _ = try await sut.makePassesViewController(fileURL: fileURL)
             XCTFail("expected to throw an error!")
         } catch {
             // success
@@ -93,13 +93,13 @@ final class MessagePresenterTests: XCTestCase {
     func testThatCreateAddPassesViewControllerReturnsAViewControllerForPassFileMessage() async throws {
         // GIVEN
         let message = MockMessageFactory.passFileTransferMessage()
-        let fileMessageData = try XCTUnwrap(message.fileMessageData)
+        let fileURL = try XCTUnwrap(message.fileMessageData?.fileURL)
 
-        // WHEN
-        let viewController = try await sut.makePassesViewController(fileMessageData: fileMessageData)
-
-        // THEN
-        XCTAssertNotNil(viewController)
+        // WHEN && THEN
+        do {
+            _ = try await sut.makePassesViewController(fileURL: fileURL)
+        } catch {
+            XCTFail("expected not to throw an error!")
+        }
     }
-
 }
--- a/wire-ios/Wire-iOS Tests/MessagePresenterTests.swift
+++ b/wire-ios/Wire-iOS Tests/MessagePresenterTests.swift
@@ -76,7 +76,7 @@ final class MessagePresenterTests: XCTestCase {
 
     // MARK: - Pass
 
-    func testThatCreateAddPassesViewControllerReturnsNilForFileMessage() async throws {
+    func testThatCreateAddPassesViewControllerThrowsAnErrorForFileMessage() async throws {
         // GIVEN
         let message = MockMessageFactory.fileTransferMessage()
         let fileMessageData = try XCTUnwrap(message.fileMessageData)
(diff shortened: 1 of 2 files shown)


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/MessagePresenter.swift
```diff
@@ -204,27 +207,44 @@ final class MessagePresenter: NSObject {
...
-            return nil
+    @MainActor
+    func makePassesViewController(fileMessageData: ZMFileMessageData) async throws -> PKAddPassesViewController? {
+        let pass = try await loadPass(fileMessageData: fileMessageData)
+        return PKAddPassesViewController(pass: pass)
+    }
+
+    private func loadPass(fileMessageData: ZMFileMessageData) async throws -> PKPass {
+        try await withCheckedThrowingContinuation { continuation in
+            Task.detached(priority: .userInitiated) {
```
caldrian:
```suggestion
            Task {
```

I'm not fully ware of the consequences but I have read from [documentation](https://developer.apple.com/documentation/swift/task/detached(priority:operation:)-3lvix) advices like "Don’t use a detached task if it’s possible".

wire-ios/Wire-iOS Tests/MessagePresenterTests.swift
```diff
@@ -76,25 +76,30 @@ final class MessagePresenterTests: XCTestCase {
 
     // MARK: - Pass
 
-    func testThatCreateAddPassesViewControllerReturnsNilForFileMessage() {
+    func testThatCreateAddPassesViewControllerReturnsNilForFileMessage() async throws {
```
caldrian:
```suggestion
    func testThatCreateAddPassesViewControllerThrowsAnErrorForFileMessage() async throws {
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/MessagePresenter.swift
```diff
@@ -204,28 +207,46 @@ final class MessagePresenter: NSObject {
...
-        guard let pass = try? PKPass.init(data: passData) else { return nil }
+        targetViewController?.present(viewController, animated: true)
+    }
 
-        if PKAddPassesViewController.canAddPasses() {
-            return PKAddPassesViewController(pass: pass)
-        } else {
-            return nil
+    @MainActor
+    func makePassesViewController(fileMessageData: ZMFileMessageData) async throws -> PKAddPassesViewController? {
```
caldrian: I understand, that `PKAddPassesViewController.init` returns an optional if  ["there was a problem initializing the object"](https://developer.apple.com/documentation/passkit/pkaddpassesviewcontroller/1619251-init).
However, since we return an error for `missingFileURL` we should add an error case for when the initialisation of the `PKAddPassesViewController` failed.

markusfassbender: Could be a good idea, so you mean we should change the method return type to `PKAddPassesViewController` without the optional then?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/MessagePresenter.swift
```diff
@@ -204,28 +207,46 @@ final class MessagePresenter: NSObject {
...
+    private func loadPass(fileURL: URL) async throws -> PKPass {
+        // Without MainActor attribute, this runs on a background queue.
+        let task = Task.detached(
+            priority: .userInitiated,
+            operation: {
+                let passData = try Data(contentsOf: fileURL)
+                return try PKPass(data: passData)
+            }
+        )
+        return try await task.value
```
caldrian, line 249:
```suggestion
@MainActor
private func loadPass(fileURL: URL) async throws -> PKPass {
    let data = try await Task {
        try Data(contentsOf: fileURL)
    }.value
    return try PKPass(data: data)
}
```

I'm not sure if this runs the IO-part on a background queue and the `PKPass.init` on the main queue, but you could give it a try.

caldrian: I would try it, but I don't want to change my branch now.

caldrian (approved):
Looks good. It would be nice if you considered my comments, but it's up to you.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/MessagePresenter.swift
markusfassbender: Why do you want to change creating `PKPass` on a background queue to main queue? In our tests it worked this way and it's better to have less work on main

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/MessagePresenter.swift
caldrian: Exactly, sorry for being unprecise.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/MessagePresenter.swift
caldrian: `PKPass.init` must be run on the main queue, see the Xcode warning in the PR description.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/MessagePresenter.swift
markusfassbender: Done

markusfassbender (approved):

PR #667, opened 2023-11-10
feat: Implement Certificate Issuing flow (part 1) WPB-5434

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5434" title="WPB-5434" target="_blank">Sub-taskWPB-5434</a>  [iOS] Get an acme directory and a nonce for creating an account
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

The purpose of this PR is to implement the concept and the first part of the certificate enrollment process. The process is shown in this image:
figure 6

In this PR I introduced:
- `E2EIService` with coreCrypto methods for E2EI;
- `AcmeClient` with acme endpoints;
- `E2EIRepository` implements all the steps for the certificate enrollment. E2EIRepository uses E2EIService and AcmeClient;
- and `EnrollE2EICertificateUseCase` is the public interface to start the process.

`HttpClient`  has a temp implementation.

----

----

----
##### References


wire-ios-data-model/E2EIdentity/E2EIService.swift
```diff
@@ -0,0 +1,111 @@
...
+    public func setupNewEnrollment() {
+        guard
+            let context = context,
+            let selfUserHandle = ZMUser.selfUser(in: context).handle,
+            let selfUserName = ZMUser.selfUser(in: context).name,
+            let selfClient = ZMUser.selfUser(in: context).selfClient(),
+            let clientId = MLSClientID(userClient: selfClient)
+        else {
+            return
+        }
```
David-Henner:
```suggestion
        let selfUser = ZMUser.selfUser(in: context)
        
        guard
            let context = context,
            let selfUserHandle = selfUser.handle,
            let selfUserName = selfUser.name,
            let selfClient = selfUser.selfClient(),
            let clientId = MLSClientID(userClient: selfClient)
        else {
            return
        }
```

wire-ios-data-model/E2EIdentity/E2EIService.swift
```diff
@@ -0,0 +1,111 @@
...
+        // TODO: we should use the new CoreCrypto version: `e2eiNewRotateEnrollment` and `e2eiNewActivationEnrollment`
+        wireE2eIdentity = try? coreCrypto.perform { try $0.e2eiNewEnrollment(clientId: clientId.rawValue,
```
David-Henner: I'd suggest going to next line after `coreCrypto.perform {`

wire-ios-data-model/E2EIdentity/E2EIService.swift
```diff
@@ -0,0 +1,111 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireCoreCrypto
+
+public protocol E2EIServiceInterface {
+
+    func directoryResponse(directoryData: Data) throws -> WireCoreCrypto.AcmeDirectory
```
David-Henner: Would this work?
```suggestion
    func directoryResponse(directoryData: Data) throws -> AcmeDirectory
```

wire-ios-data-model/E2EIdentity/E2EIService.swift
```diff
@@ -0,0 +1,111 @@
...
+    public func isE2EIEnabled() -> Bool {
+        // TODO: we should use the new CoreCrypto version
+        // return coreCrypto.e2eiIsEnabled(defaultCipherSuite.rawValue)
+        return true
+    }
+
+    // MARK: - WireE2EIdentity methods
+
+    public func directoryResponse(directoryData: Data) throws -> WireCoreCrypto.AcmeDirectory {
+        let buffer = [UInt8](directoryData)
```
David-Henner: there's a `Data+Bytes` extension that does this, you can use it as below
```suggestion
        let buffer = directoryData.bytes
```

wire-ios-data-model/E2EIdentity/E2EIService.swift
```diff
@@ -0,0 +1,111 @@
...
+            throw Failure.failedToGetAccountRequest
+        }
+
+        return accountRequest.data
+    }
+
+    public func setAccountResponse(accountData: Data) async throws {
+        let buffer = [UInt8](accountData)
+        try? wireE2eIdentity?.newAccountResponse(account: buffer)
+    }
```
David-Henner: this method isn't throwing

wire-ios-data-model/E2EIdentity/E2EIService.swift
```diff
@@ -0,0 +1,111 @@
...
+
+    public func getNewAccountRequest(previousNonce: String) async throws -> Data {
+        guard let accountRequest = try? wireE2eIdentity?.newAccountRequest(previousNonce: previousNonce) else {
+            WireLogger.e2ei.warn("Failed to get new account request")
+
+            throw Failure.failedToGetAccountRequest
+        }
+
+        return accountRequest.data
```
David-Henner: I'd suggest catching and logging errors coming from `wireE2eIdentity`. Then you can throw your own error. I'd advise doing something similar for the other methods (`directoryResponse` and `setAccountResponse`)

```swift
do {
    let accountRequest = try wireE2eIdentity.newAccountRequest(previousNonce: previousNonce)
    return accountRequest.data
} catch {
    WireLogger.e2ei.warn("Failed to get new account request: \(String(describing: error))")
    throw Failure.failedToGetAccountRequest
}
```

It also would be nice to have the `wireE2eIdentity` property be non-optional if possible. If not maybe you could do something like 

```swift
func wireE2eIdentity() throws -> WireE2eIdentity {
   guard let identity = wireE2eIdentity else {
      throw Failure.missingE2eIdentity
   }
   return identity
}
```

wire-ios-request-strategy/Sources/E2EIdentity/ACME/AcmeAPI.swift
```diff
@@ -0,0 +1,115 @@
...
+
+                return nil
+            }
+
+            var request = URLRequest(url: url)
+            request.httpMethod = HTTPMethod.get.rawValue
+
+            let responseData = try await executeAsync(request: request)
+
+            return try JSONDecoder().decode(AcmeDirectoriesResponse.self, from: responseData)
```
David-Henner: I'd suggest making `JSONDecoder` a property of the class so it doesn't need to be initialised each time this is called

wire-ios-request-strategy/Sources/E2EIdentity/ACME/AcmeAPI.swift
```diff
@@ -0,0 +1,115 @@
...
+    public func getACMEDirectory() async -> AcmeDirectoriesResponse? {
+
+        do {
+            let path = "https://acme.\(BackendInfo.domain ?? "")/acme/defaultteams/directory"
+            guard
+                let url = URL(string: path) else {
+                WireLogger.e2ei.warn("Invalid get acme directory url")
+
+                return nil
+            }
```
David-Henner:
```suggestion
            guard let url = URL(string: path) else {
                WireLogger.e2ei.warn("Invalid get acme directory url")
                return nil
            }
```

wire-ios-request-strategy/Sources/E2EIdentity/ACME/AcmeAPI.swift
```diff
@@ -0,0 +1,115 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireCoreCrypto
+
+public protocol AcmeAPI {
```
David-Henner: would it make sense to rename `AcmeAPIClient` ?

wire-ios-request-strategy/Sources/E2EIdentity/CertificateEnrollment.swift
```diff
@@ -0,0 +1,72 @@
...
+
+        /// update prevNonce?
+        guard let newNonce = await e2eiManager.createNewAccount(prevNonce: prevNonce,
+                                                                createAccountEndpoint: acmeDirectory.newAccount) else {
+            WireLogger.e2ei.warn("Fail to create acme new account")
+
+            return .failure(Failure.failedToCreateAcmeNewAccount)
+        }
+
+        return .success("")
```
David-Henner: why is the result an empty string?

wire-ios-request-strategy/Sources/E2EIdentity/E2EIManager.swift
```diff
@@ -0,0 +1,107 @@
...
+
+    /// Create a new account.
+    func createNewAccount(prevNonce: String, createAccountEndpoint: String) async -> String?
+
+}
+
+public final class E2EIManager: E2EIManagerInterface {
+
+    private var apiProvider: APIProvider
+    private var context: NSManagedObjectContext?
```
David-Henner: looks like this doesn't have to be optional

wire-ios-request-strategy/Sources/E2EIdentity/E2EIManager.swift
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        context?.performAndWait {
+            guard let e2eiService = context?.e2eiService else {
+                WireLogger.e2ei.warn("E2EIService is missing")
+
+                return
+            }
+
+            guard let acmeDirectoryData = try? JSONEncoder().encode(acmeDirectory),
```
David-Henner: `JSONEncoder` can be declared as property to avoid creating it on the fly

wire-ios-request-strategy/Sources/E2EIdentity/E2EIManager.swift
```diff
@@ -0,0 +1,107 @@
...
+        var directory: WireCoreCrypto.AcmeDirectory?
+
+        guard let acmeAPI = apiProvider.acmeAPI(apiVersion: .v5),
+              let acmeDirectory = await acmeAPI.getACMEDirectory() else {
+            WireLogger.e2ei.warn("Failed to get acme directory from acme server")
+
+            return nil
+        }
+
+        context?.performAndWait {
```
David-Henner: you can do something like

```swift
do {

   let directory = try await context.perform {
      // ...
      let acmeDirectoryData = try encoder.encode(acmeDirectory)
      return try e2eiService.directoryResponse(directoryData: acmeDirectoryData)
   }

   return AcmeDirectory(
      newNonce: directory.newNonce,
      newAccount: directory.newAccount,
      newOrder: directory.newOrder
   )

} catch {
   // log error
   return nil
}
```

wire-ios-request-strategy/Sources/E2EIdentity/ACME/AcmeAPI.swift
```diff
@@ -0,0 +1,115 @@
...
+    func sendChallengeRequest(url: String, body: Data?) async -> ChallengeResponse?
+
+}
+
+public class AcmeAPIV5: NSObject, AcmeAPI {
+
+    // MARK: - Properties
+
+    private let httpClient: HttpClient
+    private let apiVersion: APIVersion = .v5
```
David-Henner: is this used anywhere?

wire-ios-request-strategy/Sources/E2EIdentity/ACME/AcmeAPI.swift
```diff
@@ -0,0 +1,115 @@
...
+
+import Foundation
+import WireCoreCrypto
+
+public protocol AcmeAPI {
+
+    func getACMEDirectory() async -> AcmeDirectoriesResponse?
+    func getACMENonce(url: String) async -> String?
+    func sendACMERequest(url: String, body: ZMTransportData) async -> ACMEResponse?
+    func sendChallengeRequest(url: String, body: Data?) async -> ChallengeResponse?
```
David-Henner: maybe we could throw errors instead of having optional return values

wire-ios-request-strategy/Sources/E2EIdentity/E2EIManager.swift
```diff
@@ -0,0 +1,107 @@
...
+public protocol E2EIManagerInterface {
+
+    /// Fetch acme directory for hyperlinks.
+    func loadACMEDirectory() async -> AcmeDirectory?
+
+    /// Get a nonce for creating an account.
+    func getACMENonce(endpoint: String) async -> String?
+
+    /// Create a new account.
+    func createNewAccount(prevNonce: String, createAccountEndpoint: String) async -> String?
```
David-Henner: same here, might be able to throw errors

wire-ios-request-strategy/Sources/E2EIdentity/E2EIManager.swift
```diff
@@ -0,0 +1,107 @@
...
+            return nil
+        }
+
+        do {
+            try await e2eiService.setAccountResponse(accountData: apiResponse.response)
+            return apiResponse.nonce
+        } catch {
+
+            return nil
+        }
```
David-Henner: might be better to have all of this in a `do catch` block

wire-ios-data-model/E2EIdentity/E2EIService.swift
caldrian: I think we should add `throws` to the method declaration and let the caller decide, if the the error should be ignored.

wire-ios-data-model/E2EIdentity/E2EIService.swift
caldrian: @David-Henner I would actually question this. For me it hardly makes sense to define extensions/methods with a single line of code.

wire-ios-data-model/E2EIdentity/E2EIService.swift
caldrian: ... but it should! Let the caller decide if the error should be ignored or not.
```suggestion
    public func setAccountResponse(accountData: Data) throws {
        let buffer = [UInt8](accountData)
        try wireE2eIdentity?.newAccountResponse(account: buffer)
    }
```

wire-ios-data-model/E2EIdentity/E2EIService.swift
David-Henner: there's a bit more to it than a simple line https://github.com/wireapp/wire-ios/blob/develop/wire-ios-utilities/Source/Public/Data%2BBytes.swift, but I agree

caldrian: I was referring to `Data.bytes`.
```
public extension Data {
    var bytes: [Byte] {
        return [Byte](self)
    }
```
Many projects have little helpers like this, but they don't add much value, in my opinion.

wire-ios-data-model/E2EIdentity/E2EIService.swift
```diff
@@ -0,0 +1,111 @@
...
+
+    public init(context: NSManagedObjectContext,
+                coreCrypto: SafeCoreCryptoProtocol) {
+        self.context = context
+        self.coreCrypto = coreCrypto
+    }
+
+    // MARK: - Setup enrollment
+
+    public func setupNewEnrollment() {
```
johnxnguyen: Should this be part of the interface?

wire-ios-data-model/E2EIdentity/E2EIService.swift
```diff
@@ -0,0 +1,111 @@
...
+        guard
+            let context = context,
+            let selfUserHandle = ZMUser.selfUser(in: context).handle,
+            let selfUserName = ZMUser.selfUser(in: context).name,
+            let selfClient = ZMUser.selfUser(in: context).selfClient(),
+            let clientId = MLSClientID(userClient: selfClient)
+        else {
+            return
```
johnxnguyen: Are you intending to handle the early return somehow?

wire-ios-data-model/E2EIdentity/E2EIService.swift
```diff
@@ -0,0 +1,111 @@
...
+                                                                             displayName: selfUserName,
+                                                                             handle: selfUserHandle,
+                                                                             expiryDays: UInt32(90),
+                                                                             ciphersuite: defaultCipherSuite.rawValue)
+
+        }
+
+    }
+
+    public func isE2EIEnabled() -> Bool {
```
johnxnguyen: Should this be part of the interface?

wire-ios-data-model/E2EIdentity/E2EIService.swift
johnxnguyen: I guess it's a personal thing, but I find reading the word "byte" more pleasant than `UInt8` and it slightly lessens the cognitive load.

wire-ios-data-model/Source/ManagedObjectContext/NSManagedObjectContext+E2EIService.swift
```diff
@@ -0,0 +1,38 @@
...
+            return userInfo[Self.e2eiServiceUserInfoKey] as? E2EIServiceInterface
+        }
+
+        set {
+            precondition(zm_isSyncContext, "E2EIService should only be accessed on the sync context")
+            userInfo[Self.e2eiServiceUserInfoKey] = newValue
+        }
+    }
+
+}
```
johnxnguyen: As discussed previously, prefer to create and inject the service where it's needed rather than tightly coupling it a managed object context.

wire-ios-request-strategy/Sources/E2EIdentity/ACME/AcmeAPI.swift
johnxnguyen: What goes into the bodies? In my opinion the `AcmeAPI` should be clear what inputs it needs and hide the detail of packaging those inputs into request body data.

wire-ios-request-strategy/Sources/E2EIdentity/ACME/AcmeAPI.swift
```diff
@@ -0,0 +1,115 @@
...
+    // MARK: - Life cycle
+
+    public init(httpClient: HttpClient) {
+        self.httpClient = httpClient
+    }
+
+    public func getACMEDirectory() async -> AcmeDirectoriesResponse? {
+
+        do {
+            let path = "https://acme.\(BackendInfo.domain ?? "")/acme/defaultteams/directory"
```
johnxnguyen: Wouldn't an empty domain lead to a broken path? Perhaps it should be unwrapped.

wire-ios-request-strategy/Sources/E2EIdentity/ACME/AcmeAPI.swift
```diff
@@ -0,0 +1,115 @@
...
+
+    public func sendChallengeRequest(url: String, body: Data?) async -> ChallengeResponse? {
+        return nil
+    }
+
+    private func executeAsync(request: URLRequest) async throws -> Data {
+        let session = URLSession(configuration: URLSessionConfiguration.default, delegate: self, delegateQueue: nil)
+        let (data, _) = try await session.data(for: request)
+
+        return data
```
johnxnguyen: As discussed, while this is fine as a temporary solution, we should aim to make `HTTPClient` flexible enough to suit your needs rather than bypassing it.

wire-ios-request-strategy/Sources/E2EIdentity/CertificateEnrollment.swift
```diff
@@ -0,0 +1,72 @@
...
+    case missingNonce
+    case failedToCreateAcmeNewAccount
+
+}
+
+public protocol CertificateEnrollmentInterface {
+
+    func invoke(idToken: String) async -> CertificateEnrollmentResult
+
+}
```
johnxnguyen: This looks like a use case, so I suggest to use a more descriptive name such as `EnrollE2EICertificateUseCase`.

wire-ios-request-strategy/Sources/E2EIdentity/CertificateEnrollment.swift
```diff
@@ -0,0 +1,72 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+// MARK: - Types
+
+public typealias CertificateEnrollmentResult = Swift.Result<String, Error>
```
johnxnguyen: `Error` could be any error, but you can contain it to be of a specific error like `Result<String, MyError>`. This makes it possible to deal with all errors.

johnxnguyen: What does the `String` represent?

wire-ios-request-strategy/Sources/E2EIdentity/CertificateEnrollment.swift
```diff
@@ -0,0 +1,72 @@
...
+
+enum Failure: Error, Equatable {
+
+    case failedToLoadACMEDirectory
+    case missingNonce
+    case failedToCreateAcmeNewAccount
+
+}
```
johnxnguyen: I suggest a more descriptive name like `EnrollE2EICertificateFailure`

wire-ios-request-strategy/Sources/E2EIdentity/E2EIManager.swift
```diff
@@ -0,0 +1,107 @@
...
+
+    public init(apiProvider: APIProvider, context: NSManagedObjectContext) {
+        self.apiProvider = apiProvider
+        self.context = context
+    }
+
+    public func loadACMEDirectory() async -> AcmeDirectory? {
+        var directory: WireCoreCrypto.AcmeDirectory?
+
+        guard let acmeAPI = apiProvider.acmeAPI(apiVersion: .v5),
```
johnxnguyen: Is the ACME api actually versioned? It's not part of the Wire REST api

wire-ios-request-strategy/Sources/E2EIdentity/E2EIManager.swift
```diff
@@ -0,0 +1,107 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireCoreCrypto
+
+public protocol E2EIManagerInterface {
+
+    /// Fetch acme directory for hyperlinks.
+    func loadACMEDirectory() async -> AcmeDirectory?
```
johnxnguyen: Does this need to be part of the interface? Couldn't we make it internal to the implementation since it seems that the result of this method is only used to invoke the other methods in the interface.

wire-ios-data-model/E2EIdentity/E2EIService.swift
```diff
@@ -0,0 +1,111 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireCoreCrypto
+
+public protocol E2EIServiceInterface {
```
johnxnguyen: It's not clear to me what the responsibility of `E2EIService` is and how it differs from `E2EIManager`, could you elaborate on this?

wire-ios-data-model/E2EIdentity/E2EIService.swift
typfel: throw an error?

wire-ios-request-strategy/Sources/E2EIdentity/ACME/AcmeAPI.swift
typfel: throw an error

wire-ios-request-strategy/Sources/E2EIdentity/ACME/AcmeAPI.swift
typfel: Don't return nil throw an error instead

wire-ios-request-strategy/Sources/E2EIdentity/ACME/AcmeAPI.swift
```diff
@@ -0,0 +1,115 @@
...
+            var request = URLRequest(url: url)
+            request.httpMethod = HTTPMethod.get.rawValue
+
+            let responseData = try await executeAsync(request: request)
+
+            return try JSONDecoder().decode(AcmeDirectoriesResponse.self, from: responseData)
+        } catch {
+            WireLogger.e2ei.info("Get acme directory request failed with error: \(error)")
+
+            return nil
```
typfel: Don't return nil on an error: consider adopting something like this https://github.com/wireapp/wire-ios/blob/93830c6b20d768a506347cd47ea717ac8c58763c/wire-ios-request-strategy/Sources/APIs/MessageAPI.swift#L110

wire-ios-request-strategy/Sources/E2EIdentity/ACME/AcmeAPI.swift
```diff
@@ -0,0 +1,115 @@
...
+            }
+
+            var request = URLRequest(url: url)
+            request.httpMethod = HTTPMethod.get.rawValue
+
+            let responseData = try await executeAsync(request: request)
+
+            return try JSONDecoder().decode(AcmeDirectoriesResponse.self, from: responseData)
+        } catch {
+            WireLogger.e2ei.info("Get acme directory request failed with error: \(error)")
```
typfel: If you always throw an error there's no need to log an error at every step. Logging can happen at the top level.

wire-ios-request-strategy/Sources/E2EIdentity/CertificateEnrollment.swift
typfel: If you throw errors instead returning nil and rely on bubbling up errors this becomes much simpler:

```suggestion
        let acmeDirectory = await e2eiManager.loadACMEDirectory()
        let prevNonce = await e2eiManager.getACMENonce(endpoint: acmeDirectory.newNonce)
        let newNonce = await e2eiManager.createNewAccount(prevNonce: prevNonce, createAccountEndpoint: acmeDirectory.newAccount)
```

wire-ios-request-strategy/Sources/E2EIdentity/E2EIManager.swift
typfel: It's not versioned

typfel (changes requested):

wire-ios-data-model/E2EIdentity/E2EIService.swift
```diff
@@ -0,0 +1,103 @@
...
+    private let mlsClientId: MLSClientID
+    private let userName: String
+    private let handle: String
+
+    // MARK: - Life cycle
+
+    public init(coreCrypto: SafeCoreCryptoProtocol,
+                mlsClientId: MLSClientID,
+                userName: String,
+                handle: String) {
```
typfel: Do we really store these properties? Can't they be an argument to `setupEnrollment`?

wire-ios-data-model/E2EIdentity/E2EIService.swift
```diff
@@ -0,0 +1,103 @@
...
+            e2eIdentity = try coreCrypto.perform {
+                try $0.e2eiNewEnrollment(clientId: mlsClientId.rawValue,
+                                         displayName: userName,
+                                         handle: handle,
+                                         expiryDays: UInt32(90),
+                                         ciphersuite: defaultCipherSuite.rawValue)
+            }
+
+        } catch {
+            throw Failure.failedToSetupE2eiEnrollment
```
typfel: When catching an re-throwing I think we should include the original error: 
```
throw Failure.failedToSetupE2eiEnrollment(error)
```

wire-ios-request-strategy/Sources/E2EIdentity/ACME/AcmeClient.swift
```diff
@@ -0,0 +1,58 @@
...
+            throw NetworkError.invalidRequestURL
+        }
+
+        let path = "https://acme.\(domain)/acme/defaultteams/directory"
+
+        let request = ZMTransportRequest(getFromPath: path, apiVersion: 0)
+        let result = try await httpClient.send(request)
+
+        guard let data = result.rawData else {
+            throw NetworkError.invalidRequestURL
```
typfel:
```suggestion
            throw NetworkError.invalidResponse
```

wire-ios-request-strategy/Sources/E2EIdentity/E2EIRepository.swift
```diff
@@ -0,0 +1,72 @@
...
+        logger.info("load ACME directory")
+
+        do {
+            try await e2eiService.setupEnrollment()
+            let acmeDirectoryData = try await acmeClient.getACMEDirectory()
+            return try await e2eiService.directoryResponse(directoryData: acmeDirectoryData)
+        } catch {
+            logger.error("failed to load ACME directory: \(error.localizedDescription)")
+
+            throw E2EIRepositoryFailure.failedToLoadACMEDirectory
```
typfel: Include the underlaying error

wire-ios-data-model/E2EIdentity/E2EIService.swift
```diff
@@ -0,0 +1,103 @@
...
+                handle: String) {
+        self.coreCrypto = coreCrypto
+        self.mlsClientId = mlsClientId
+        self.userName = userName
+        self.handle = handle
+    }
+
+    // MARK: - Setup enrollment
+
+    public func setupEnrollment() async throws {
```
typfel: This could return `WireE2eIdentityProtocol` then we  `Failure.missingE2eIdentity` can't happen.

typfel (changes requested):

typfel (approved):
👍

johnxnguyen (approved):

PR #679, opened 2023-11-14
chore: split unit test plan

markusfassbender:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The project has a `Wire-iOS Test` folder with a mixture of unit and snapshot tests. As developer this makes it difficult and time consuming just to write and run a few unit tests. It also has downsides, that you can execute the tests only on certain simulator devices.

### Solutions

Thus I've created a new `Wire-iOS UnitTest` target, included it in the existing test plans and added a new `UnitTests.xctestplan` to run only unit tests within a few seconds.

To make it work I've created a new folder "Tests", that contains shared code for mocking and fixtures like the massive `CoreDataFixture`. Also the `xctestplan` files moved there.

Screenshot 2023-11-14 at 08 01 20

### Known issue

- Not all unit tests have been moved to the new target yet. Why? Because many unit tests have shared resources with snapshot tests and extensions on Apple classes etc. and it is hard to split without duplicating everything and time consuming. To make a first step I've only migrated a few tests.

#### How to Test

- In Xcode select your testplan and run it with CMD+U 🥳
- On CI nothing should have changed, because the `xctestplan` files have been updated and contain the same tests as before.

### Notes

- `testThatSharedInstanceReturnQualityTypeWifi` seems to be dependent on your internet connection and doesn't work when you are offline.

----

----

----
##### References


wire-ios/Wire-iOS Tests/InputBar/InputBarRightMenuButtonPresenterTests.m
```diff
@@ -1,115 +0,0 @@
...
-// 
-// You should have received a copy of the GNU General Public License
-// along with this program. If not, see http://www.gnu.org/licenses/.
-// 
-
-
-#import <UIKit/UIKit.h>
-#import <XCTest/XCTest.h>
-
-#import "InputBarRightMenuButtonPresenter.h"
```
markusfassbender, line 23: This header file did not exist and the test was not in the target to build. So it was dead and never executed.

wire-ios/Tests/TestPlans/SecurityTests.xctestplan
```diff
@@ -137,6 +137,32 @@
...
+        "PhoneNumberTests",
+        "StringCapitalizationTests",
+        "StringWholeRangeTests",
+        "String_PhoneNumberTests",
+        "TemporaryFileServiceTests",
+        "TransformTextViewTests",
+        "UITextView_ReplaceTests",
+        "UITraitEnvironmentTests",
+        "UserRightTests",
+        "ZMUserMentionsTest"
```
netbe: why are these skipped? the SecurityTests plan should not have the option to automatically add new tests. So we explicitly add tests

wire-ios/Tests/TestPlans/UnitTests.xctestplan
```diff
@@ -0,0 +1,25 @@
...
+      "options" : {
+
+      }
+    }
+  ],
+  "defaultOptions" : {
+    "language" : "en",
+    "region" : "US"
+  },
+  "testTargets" : [
```
netbe, line 15: is it the equivalent of the GermanLocaleStrings plan. In this plan we have a couple of tests that need to be run on German locale, probably GermanLocaleStrings needs the testTarget Wire-iOS UnitTests"

netbe (changes requested):
You need to look also to the github workflows and update the tests runs in the two reusable workflows to make that work.

wire-ios/Tests/TestPlans/UnitTests.xctestplan
markusfassbender: I have copied this from the "AllTests.xctestplan", so I assume that's correct. There is another `GermanLocaleTest.xctestplan`, that is configured for it.
Screenshot 2023-11-14 at 19 01 42

wire-ios/Tests/TestPlans/SecurityTests.xctestplan
markusfassbender: Sorry when I added the new bundle, it was enabled automatically. Now I have disabled the "automatically include new tests" option.

markusfassbender:
@netbe I looked at the CI yaml, but they execute the `AllTests.xctestplan`, which includes the unit test target. So I think we don't need to update the CI workflows, is that right?

netbe:
> @netbe I looked at the CI yaml, but they execute the `AllTests.xctestplan`, which includes the unit test target. So I think we don't need to update the CI workflows, is that right?

@markusfassbender yes sounds good

netbe (approved):

agisilaos (approved):

PR #684, opened 2023-11-15
feat: implement action action handler for protocol update [WPB-5266]

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5266" title="WPB-5266" target="_blank">Sub-taskWPB-5266</a>  [iOS] Implement action + action handler for protocol update
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Implementation and tests for `UpdateConversationProtocolAction` and `UpdateConversationProtocolActionHandler`.

----

----

----
##### References


wire-ios-data-model/Source/MLS/Actions/UpdateConversationProtocolAction.swift
```diff
@@ -0,0 +1,55 @@
...
+    public var domain: String
+    public var conversationID: UUID
+    public var messageProtocol: MessageProtocol
+    public var resultHandler: ResultHandler?
+
+    // MARK: - Life cycle
+
+    public init(
+        domain: String,
+        conversationID: UUID,
```
caldrian: Should these two arguments be combined into one `qualifiedID: QualifiedID` argument?

typfel: I'd say yes

wire-ios-request-strategy/Sources/Request Strategies/MLS/Actions/UpdateConversationProtocolActionHandler.swift
```diff
@@ -0,0 +1,68 @@
...
+
+        case 200, 204:
+            action.succeed()
+
+        default:
+            let label = response.payloadLabel()
+            let message = response.payload?.asDictionary()?["message"] as? String
+            let error = Action.Failure.api(statusCode: response.httpStatus, label: label ?? "", message: message ?? "")
+            action.fail(with: error)
+        }
```
typfel: Decode using `ResponseFailure` struct instead, see https://github.com/wireapp/wire-ios/blob/eff95a8097bd8f1ada0e1443f1a4d3dccca5d1df/wire-ios-request-strategy/Sources/Request%20Strategies/Connection/Actions/UpdateConnectionActionHandler.swift#L100

typfel (changes requested):

caldrian (commented):
I just discovered the `MLSActionsProviderProtocol`. Do I have to add my action to it?

typfel:
> I just discovered the `MLSActionsProviderProtocol`. Do I have to add my action to it?

Don't think so.

wire-ios-data-model/Source/MLS/Actions/UpdateConversationProtocolAction.swift
```diff
@@ -0,0 +1,82 @@
...
+        qualifiedID: QualifiedID,
+        messageProtocol: MessageProtocol
+    ) {
+        self.qualifiedID = qualifiedID
+        self.messageProtocol = messageProtocol
+    }
+
+}
+
+extension UpdateConversationProtocolAction {
```
David-Henner, line 41: Is there a good reason for extending the struct rather than doing all the implementation in the initial declaration?

wire-ios-request-strategy/Sources/Request Strategies/MLS/Actions/UpdateConversationProtocolActionHandler.swift
```diff
@@ -0,0 +1,72 @@
...
+        var action = action
+
+        if [200, 204].contains(response.httpStatus) {
+            return action.succeed()
+        }
+
+        if
+            let label = response.payloadLabel(),
+            let apiFailure = Action.Failure.APIFailure(rawValue: label),
+            apiFailure.statusCode == response.httpStatus {
```
David-Henner: This only works under the condition that there's a label in the payload. Unfortunately this is not always the case, there could be a status but no label. e.g: status `404` without a label means the conversation or the domain wasn't found. (sadly this is not obvious in swagger)

I'd suggest to do something like this 

```swift
switch (response.httpStatus, response.payloadLabel()) {
   case (200, _), (204, _):
      action.succeed()
   case (400, "mls-migration-criteria-not-satisfied"):
      action.fail(with: .api(.mlsMigrationCriteriaNotSatisfied))
   case (400, _):
      action.fail(with: .api(.invalidBody))
   case (404, "no-team"):
      action.fail(with: .api(.noTeam))
   case (404, _):
      action.fail(with: .api(.conversationIDOrDomainNotFound))
   // ...
   default:
      action.fail(with: .unknown)
}
```

Doing the mapping like this in the handler also ensure that responsibilities are kept separate between the action and the handler. The action only concerns itself with the expected input and output, and the handler can take care of creating the API request and mapping the response

caldrian: > This only works under the condition that there's a label in the payload. Unfortunately this is not always the case

I see.

@typfel Then the code which you mentioned is probably also not correct, since it doesn't consider the status code at all, right?
https://github.com/wireapp/wire-ios/blob/eff95a8097bd8f1ada0e1443f1a4d3dccca5d1df/wire-ios-request-strategy/Sources/Request%20Strategies/Connection/Actions/UpdateConnectionActionHandler.swift#L100-L102
Maybe in other places as well.

wire-ios-data-model/Source/MLS/Actions/UpdateConversationProtocolAction.swift
caldrian: I like to structure it a bit using extensions, but if you prefer not to do it, I can merge the parts.

One minor advantage might be that all the stored properties are visible at one sight, within a few lines of code.  Extensions cannot declare stored properties.

wire-ios-request-strategy/Sources/Request Strategies/MLS/Actions/UpdateConversationProtocolActionHandlerTests.swift
```diff
@@ -0,0 +1,124 @@
...
+
+import XCTest
+@testable import WireRequestStrategy
+
+final class UpdateConversationProtocolActionHandlerTests: ActionHandlerTestBase<UpdateConversationProtocolAction, UpdateConversationProtocolActionHandler> {
+
+    override func setUp() {
+        super.setUp()
+
+        let uuid = UUID(uuidString: "b906579d-60dd-4510-a3ca-14b2ec225f4a")!
```
netbe: use constant that can be reuse in the assertion of the path

wire-ios-request-strategy/Sources/Request Strategies/MLS/Actions/UpdateConversationProtocolActionHandler.swift
typfel: > > This only works under the condition that there's a label in the payload. Unfortunately this is not always the case
> 
> I see.
> 
> @typfel Then the code which you mentioned is probably also not correct, since it doesn't consider the status code at all, right?
> 
> https://github.com/wireapp/wire-ios/blob/eff95a8097bd8f1ada0e1443f1a4d3dccca5d1df/wire-ios-request-strategy/Sources/Request%20Strategies/Connection/Actions/UpdateConnectionActionHandler.swift#L100-L102
> 
> 
> Maybe in other places as well.

No, it doesn't but the status code is also encoded inside the `Payload.ResponseFailure`, but never mind let go with what David suggested since that's the system in place now for action handlers.

typfel (approved):

wire-ios-request-strategy/Sources/Request Strategies/MLS/Actions/UpdateConversationProtocolActionHandlerTests.swift
caldrian: done: d362da01065a1abfa71436d2bb754088a680e40c

PR #687, opened 2023-11-16
feat: Scroll to the bottom of a conversation - WPB-5518

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5518" title="WPB-5518" target="_blank">StoryWPB-5518</a>  [iOS] Scroll to Bottom of a conversation
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This pull request introduces a new feature in our conversation view: a dynamic scrollToBottomButton. This feature enhances user interaction by providing a convenient and intuitive way to navigate to the latest messages. 

### Key Features

1. Dynamic Visibility:
The button becomes visible when the user scrolls up past a certain number of messages and automatically hides when the user is at the bottom of the conversation.

2. Smooth Animations for Visibility Changes:
The button's appearance and disappearance are accompanied by subtle fade-in and fade-out animations

3. Accessibility Support:
The feature respects the iOS "Reduce Motion" accessibility setting. When this setting is enabled, animations are replaced with immediate visibility changes

4. Clean and Modular Code:
The logic for animating and updating the button's visibility has been encapsulated in a new method, updateScrollToBottomButtonVisibility, enhancing code readability and maintainability.

### Rationale

This feature addresses the need for an efficient way to navigate long conversations. By adding this button, users can quickly jump to the most recent messages without manually scrolling, significantly improving the chat interface's usability.

--

Designs can be found [here](https://www.figma.com/file/UgKehvwIbiIiQ2CCEdsVuf/Android?type=design&node-id=38856-289959&mode=design&t=GN4vG9lPSn7yWRon-0)

### Demo

https://github.com/wireapp/wire-ios/assets/10944108/feac7d2f-2f9f-4085-93ea-edf296b43391

----
##### References


wire-ios/Wire-iOS/Sources/Components/Buttons/LegacyButton.swift
```diff
@@ -37,6 +37,14 @@ class Button: LegacyButton {
         applyStyle(style)
     }
 
+    convenience init(style: ButtonStyle, cornerRadius: CGFloat = 0) {
```
markusfassbender: If I run `Button(style: someStyle)`, which init will be used?

From line 27:
```swift
convenience init(style: ButtonStyle,
                     cornerRadius: CGFloat = 0,
                     fontSpec: FontSpec = .smallLightFont)
``` 

My suggestion is to remove the default parameters if we have multiple convince inits, so that it's clear.

wire-ios/Wire-iOS/Sources/Components/Styles/ButtonStyle.swift
```diff
@@ -101,4 +101,13 @@ public struct ButtonStyle {
...
                                                          border: ButtonColors.borderInputBarItemHighlighted))
+
+    static let scrollToBottomButtonStyle = ButtonStyle(normalStateColors: (
+                                                        background: ButtonColors.backgroundScrollToBottonEnabled,
+                                                        title: SemanticColors.Icon.foregroundDefaultWhite,
+                                                         border: nil),
+                                                        highlightedStateColors: (
+                                                        background: UIColor.accent(),
+                                                         title: SemanticColors.Icon.foregroundDefaultWhite,
+                                                         border: nil))
```
markusfassbender: 💅 The indentation looks strange

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+ScrollViewDelegate.swift
```diff
@@ -27,6 +27,37 @@ extension ConversationContentViewController: UIScrollViewDelegate {
...
+    /// - Note: This method should be called whenever the scroll position changes, typically in response to a
+    ///   `scrollViewDidScroll` event, to ensure that the button's visibility accurately reflects whether the
+    ///   user is scrolled to the bottom of the content.
+    func updateScrollToBottomButtonVisibility() {
+        let shouldHideButton = isScrolledToBottom
+
+        if scrollToBottomButton.isHidden != shouldHideButton {
+            if UIAccessibility.isReduceMotionEnabled {
+                self.scrollToBottomButton.alpha = shouldHideButton ? 0 : 1
+                self.scrollToBottomButton.isHidden = shouldHideButton
```
markusfassbender:
```suggestion
                scrollToBottomButton.alpha = shouldHideButton ? 0 : 1
                scrollToBottomButton.isHidden = shouldHideButton
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.swift
```diff
@@ -37,6 +37,31 @@ final class ConversationContentViewController: UIViewController, PopoverPresente
...
+        let image = Asset.Images.downArrow.image.withTintColor(SemanticColors.Icon.foregroundDefaultWhite)
+
+        button.setImage(image, for: .normal)
+        button.setImage(image, for: .highlighted)
+        button.translatesAutoresizingMaskIntoConstraints = false
+
+        button.accessibilityLabel = L10n.Accessibility.Conversation.ScrollToBottomButton.description
+        button.accessibilityHint = L10n.Accessibility.Conversation.ScrollToBottomButton.hint
+
+        button.addTarget(self, action: #selector(handleScrollToBottomTapped), for: .touchUpInside)
```
markusfassbender: Can we do it without Obj-C? 

```
let action = UIAction { [weak self] _ in
     self?.buttonTapped ()
}
button.addAction(action, for: .touchUpInside)
```

agisilaos: Done!

wire-ios/Wire-iOS/Sources/Components/Buttons/LegacyButton.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/Components/Styles/ButtonStyle.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
```diff
@@ -52,11 +52,28 @@ extension ConversationContentViewController {
...
 
+    /// Scrolls the tableView to the bottom-most row.
+    ///
+    /// This method checks if the tableView is not already scrolled to the bottom.
+    /// If not, it loads new messages from the dataSource and scrolls to the bottom row.
+    /// The scroll to the bottom is animated based on the user's accessibility settings
+    /// and the number of messages. If reduce motion is enabled or the number of messages
+    /// exceeds 20, the scroll animation is set to `.top`; otherwise, it's set to `.bottom`.
+    /// After scrolling, the tableView's header view is updated.
     func scrollToBottom() {
```
netbe: it's a bit paraphrasing the source code, I don't think we need that much explanation here

netbe:
```suggestion
    /// This method checks if the tableView is not already scrolled to the bottom.
    func scrollToBottomIfNeeded() {
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
```diff
@@ -52,11 +52,28 @@ extension ConversationContentViewController {
...
-        guard !isScrolledToBottom else { return }
+        guard !isScrolledToBottom else {
+            print("scrollToBottom was called, but we're already at the bottom. No action taken.")
```
netbe: remove print

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
```diff
@@ -52,11 +52,28 @@ extension ConversationContentViewController {
...
+            return
+        }
 
         dataSource.loadMessages()
-        tableView.scroll(toIndex: 0)
+
+        let lastRowIndexPath = IndexPath(row: 0, section: 0)
+        let shouldAnimate = !UIAccessibility.isReduceMotionEnabled && dataSource.messages.count <= 20
```
netbe: use constant, did you check on smaller devices, or iPad? How do you choose this number ?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
agisilaos: I find it useful to read such documentation when I option click on a method. Helps me wrap up my head around the code much faster than just lines of code. If you feel strongly about it, I can remove it of course.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
agisilaos: use constant for what? Ah you meant the number of messages. Well for performance reasons mostly. If it's above 20. We don't animate. Otherwise we do animate.

markusfassbender (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
agisilaos: On iPad it works fine as well as other devices.

agisilaos: Created a property for the number 20 @netbe

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
agisilaos: Fixed it!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+ScrollViewDelegate.swift
```diff
@@ -27,6 +27,37 @@ extension ConversationContentViewController: UIScrollViewDelegate {
...
+                scrollToBottomButton.alpha = shouldHideButton ? 0 : 1
+                scrollToBottomButton.isHidden = shouldHideButton
+            } else {
+                UIView.animate(withDuration: 0.25, animations: {
+                    self.scrollToBottomButton.alpha = shouldHideButton ? 0 : 1
+                }) { _ in
+                    self.scrollToBottomButton.isHidden = shouldHideButton
+                }
+            }
+        }
```
johnxnguyen: Would it also work to set duration 0 if reduced motion is enabled?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
```diff
@@ -52,11 +52,24 @@ extension ConversationContentViewController {
...
+            return
+        }
 
         dataSource.loadMessages()
-        tableView.scroll(toIndex: 0)
+
+        let lastRowIndexPath = IndexPath(row: 0, section: 0)
+        let shouldAnimate = !UIAccessibility.isReduceMotionEnabled && dataSource.messages.count <= maxMessageCountForAnimation
+
+        let scrollAnimation: UITableView.ScrollPosition = shouldAnimate ? .bottom : .top
```
johnxnguyen: Why does the scroll position change depending whether we should animate?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+ScrollViewDelegate.swift
agisilaos: Didn't try but in terms of responsibilities I wanted to keep those two things separate. Only use the animation block when `UIAccessibility.isReduceMotionEnabled == false`

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
agisilaos: It was a leftover as I was trying different solutions. It's not needed indeed. Good catch @johnxnguyen

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
```diff
@@ -52,11 +52,22 @@ extension ConversationContentViewController {
         updateTableViewHeaderView()
     }
 
-    func scrollToBottom() {
-        guard !isScrolledToBottom else { return }
+    /// Scrolls the tableView to the bottom-most row.
+    ///
+    /// This method checks if the tableView is not already scrolled to the bottom.
+    func scrollToBottomIfNeeded() {
+        let maxMessageCountForAnimation = 20
```
KaterinaWire: Why do you think we should have `maxMessageCountForAnimation` ?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
```diff
@@ -52,11 +52,22 @@ extension ConversationContentViewController {
...
+            return
+        }
 
         dataSource.loadMessages()
-        tableView.scroll(toIndex: 0)
+
+        let lastRowIndexPath = IndexPath(row: 0, section: 0)
+        let shouldAnimate = !UIAccessibility.isReduceMotionEnabled && dataSource.messages.count <= maxMessageCountForAnimation
+
+        tableView.scrollToRow(at: lastRowIndexPath, at: .bottom, animated: shouldAnimate)
```
KaterinaWire: Why don't you want to use same `tableView.scroll(toIndex: 0)` but with an additional `animated` parameter?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.swift
```diff
@@ -38,6 +38,35 @@ final class ConversationContentViewController: UIViewController, PopoverPresente
         }
     }
 
+    let scrollToBottomButtonTrailingMargin: CGFloat = -10
+    let scrollToBottomButtonBottomMargin: CGFloat = -10
```
KaterinaWire: I would keep the value as 10.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
agisilaos: The reason behind this is performance. What if we have to scroll through hundreds or more messages. It would degrade the performance. So that's why we have a limit of 20. Over 20 -> we don't animate. Less than 20 we animate.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
agisilaos: It is a custom method and we don't have the ability to animate using tableView.scroll(toIndex: 0). There is no additional parameter: animated.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.swift
agisilaos: for bottom and trailing needs to be -

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+ScrollViewDelegate.swift
```diff
@@ -27,6 +27,37 @@ extension ConversationContentViewController: UIScrollViewDelegate {
...
+    /// If the button's current visibility state is different from what it should be, it either directly sets the visibility
+    /// or performs a fade in/out animation based on the user's accessibility settings for reduced motion.
+    ///
+    /// When "Reduce Motion" is enabled in accessibility settings, the button's visibility is changed without animation
+    /// to respect the user's preference for reduced motion. Otherwise, a 0.25 second fade animation is used for
+    /// a smoother transition between visible and hidden states.
+    ///
+    /// - Note: This method should be called whenever the scroll position changes, typically in response to a
+    ///   `scrollViewDidScroll` event, to ensure that the button's visibility accurately reflects whether the
+    ///   user is scrolled to the bottom of the content.
```
johnxnguyen: In my opinion, I think you could summarize this to: 

```
Show or hide the scroll to bottom button for the current scroll position."
```

I find a doc comment most useful when it concisely tells me what the outcome of the invoking the method and not how it works. If I want to know how it works I'll read the method definition.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
johnxnguyen: Doesn't iOS optimize this though? What happens when you tap on the top status bar while at the beginning of a long conversation? It should scroll to the bottom with an animation.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
```diff
@@ -52,11 +52,22 @@ extension ConversationContentViewController {
...
+        let maxMessageCountForAnimation = 20
+
+        guard !isScrolledToBottom else {
+            return
+        }
 
         dataSource.loadMessages()
-        tableView.scroll(toIndex: 0)
+
+        let lastRowIndexPath = IndexPath(row: 0, section: 0)
```
johnxnguyen: I think you should check first that there is a row at this index path.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
KaterinaWire: You can add this parameter and it has a check that John mentioned above.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.swift
KaterinaWire: we usually add - when adding constraints

KaterinaWire: what about having only one value: `scrollToBottomButtonMargin`?  since they are the same.

agisilaos: > we usually add - when adding constraints

@KaterinaWire Ahhh, that's what you meant! I can make the change I guess. Good point! 👍

agisilaos: > what about having only one value: `scrollToBottomButtonMargin`? since they are the same.

Do you think it's going to be clear enough, since it's used on two different constraints?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+ScrollViewDelegate.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.swift
agisilaos: > > we usually add - when adding constraints
> 
> @KaterinaWire Ahhh, that's what you meant! I can make the change I guess. Good point! 👍

Done!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
agisilaos: We check if we should continue with this method or not by checking:

If it's false then we proceed. If it's true then we return. 

```swift
var isScrolledToBottom: Bool {
        return !dataSource.hasNewerMessagesToLoad &&
        tableView.contentOffset.y + tableView.correctedContentInset.bottom <= 0
    }
``` 

Check line 61:

```swift
guard !isScrolledToBottom else {
     return
}
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
agisilaos: I don't know how they optimize though and if they're using any private APIs. I was checking the rest of the messenger apps and they're doing what I'm trying to do here regarding animation based on the messages.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/f1ff30b49e7c016d6056822b8d4376f3522cfbc6

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
agisilaos: @johnxnguyen Took care of it! Please check: https://github.com/wireapp/wire-ios/commit/f1ff30b49e7c016d6056822b8d4376f3522cfbc6

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
```diff
@@ -64,11 +64,9 @@ extension ConversationContentViewController {
 
         dataSource.loadMessages()
 
-        let lastRowIndexPath = IndexPath(row: 0, section: 0)
         let shouldAnimate = !UIAccessibility.isReduceMotionEnabled && dataSource.messages.count <= maxMessageCountForAnimation
```
johnxnguyen: So does this mean it will only animate if the conversation has less than 20 messages?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.swift
```diff
@@ -125,25 +154,40 @@ final class ConversationContentViewController: UIViewController, PopoverPresente
...
+            ]
+        )
         let heightCollapsingConstraint = bottomContainer.heightAnchor.constraint(equalToConstant: 0)
         heightCollapsingConstraint.priority = .defaultHigh
         heightCollapsingConstraint.isActive = true
     }
 
     override func viewDidLoad() {
         super.viewDidLoad()
+        tableView.delegate = self
```
KaterinaWire: Why do we need this changes now?

KaterinaWire (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
agisilaos: That's correct and I just realized the point you're trying to make. 🤔 Maybe this isn't the best way to decide if we should animate or not. I will remove it for now and we can check if we have any performance degrades or not.

johnxnguyen (approved):

PR #692, opened 2023-11-17
feat: [MLS Migration] remove dead system message [WPB-5268]

markusfassbender:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5268" title="WPB-5268" target="_blank">Sub-taskWPB-5268</a>  [iOS] Add support for new system messages
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Remove handling of deprecated system messages.

### Dependencies (Optional)

Needs releases with:

- [ ] GitHub link to other pull request

#### Test Coverage (Optional)

- [x] I have ~~added~~ removed automated test to this contribution

#### How to Test

I think this does not need to be tested explicitly, because the feature was removed.

### Attachments (Optional)

----

----

----
##### References


wire-ios-data-model/Source/Public/ZMMessage.h
```diff
@@ -77,8 +77,6 @@ typedef NS_CLOSED_ENUM(int16_t, ZMSystemMessageType) {
     ZMSystemMessageTypeDecryptionFailed,
     ZMSystemMessageTypeDecryptionFailed_RemoteIdentityChanged,
     ZMSystemMessageTypeNewConversation,
-    ZMSystemMessageTypeReactivatedDevice __deprecated_enum_msg("Devices can't be reactivated any longer"),
-    ZMSystemMessageTypeUsingNewDevice __deprecated_enum_msg("We don't need inform users about new devices any longer"),
```
markusfassbender: The deprecation commit is from February 2021. I just asked myself if we can remove the cases without side effects. For example if the `int16` is stored and loaded, could it have a wrong value then?

typfel: No we can't remove system message types like this it would change the meaning any persisted system message with a higher value than 14

typfel: Either we keep the deprecated enum values or we start assigning explicit values to the enum cases.

markusfassbender: Thanks, I would tend to set explicit values then. Do you see any arguments against it?

wire-ios-data-model/Source/Public/ZMMessage.h
```diff
@@ -67,8 +67,8 @@ typedef NS_CLOSED_ENUM(int16_t, ZMSystemMessageType) {
     ZMSystemMessageTypeFailedToAddParticipants,
     ZMSystemMessageTypeParticipantsRemoved,
     ZMSystemMessageTypeConversationNameChanged,
-    ZMSystemMessageTypeConnectionRequest,
-    ZMSystemMessageTypeConnectionUpdate,
+    ZMSystemMessageTypeConnectionRequest __deprecated_enum_msg("deprecated"),
+    ZMSystemMessageTypeConnectionUpdate __deprecated_enum_msg("deprecated"),
```
markusfassbender, line 81: I took this cases as deprecate because of `ConversationSystemMessageCell`, that says the cases are deprecated.

markusfassbender (commented):
Comments

wire-ios-data-model/Source/Public/ZMMessage.h
markusfassbender: @typfel I've reverted the case removal, please take a look.

wire-ios-data-model/Source/Public/ZMMessage.h
netbe: as far as I understand we can't remove them, right?, so maybe add more info on the deprecated message to keep track of why we can't remove them

netbe (commented):
this is just one part of WPB-5268 right?

wire-ios-data-model/Source/Public/ZMMessage.h
markusfassbender: I have added a longer comment on top. For the two cases I just added the compiler attribute, but they were deprecated before (see `ConversationSystemMessageCell`), so I don't know the exact details-

PR #696, opened 2023-11-21
feat(mls-migration): handle mixed protocol for conversation participants operations WPB-529

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-529" title="WPB-529" target="_blank">StoryWPB-529</a>  [iOS] Maintaining MLS group's clients list during migration
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR covers adding and removing participants for conversations in `mixed` protocol. 
For adding participants, we need to support both Proteus and MLS. 
Whereas for participants removal, only the Proteus way needs to be supported

### Changes

As there was a lot of logic related to adding and removing participants that was implemented in an extension of `ZMConversation`, I decided to do a small refactoring and move that logic to the `ConversationService` and separated concerns into several classes.

- Created a `MLSConversationParticipantsService` responsible for adding and removing participants from MLS groups
- Created a `ProteusConversationParticipantsService` responsible for adding and removing participants from the Proteus part of the conversation
- Created a `ConversationParticipantsService` which decides which of the MLS or Proteus participants service to use based on the `messageProtocol` of the conversation
- Added `addParticipants` and `removeParticipant` interface to `ConversationService`. It uses the `ConversationParticipantsService` and also handles errors and retries related to federation.

#### Test Coverage (Optional)

- [x] I have added unit tests for each class


wire-ios-request-strategy/Sources/Request Strategies/Conversation/ProteusConversationParticipantsService.swift
```diff
@@ -0,0 +1,52 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireDataModel
+
+// sourcery: AutoMockable
+protocol ProteusConversationParticipantsServiceInterface: ConversationParticipantsServiceInterface {}
+
+class ProteusConversationParticipantsService: ProteusConversationParticipantsServiceInterface {
```
caldrian: This should be a struct.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsService.swift
```diff
@@ -0,0 +1,141 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireDataModel
+
+// sourcery: AutoMockable
+protocol MLSConversationParticipantsServiceInterface: ConversationParticipantsServiceInterface {}
+
+class MLSConversationParticipantsService: MLSConversationParticipantsServiceInterface {
```
caldrian: This should also be a struct, in my opinion. No need to allocate memory on the heap for it.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsService.swift
```diff
@@ -0,0 +1,141 @@
...
+
+    func addParticipants(
+        _ users: [ZMUser],
+        to conversation: ZMConversation,
+        completion: @escaping AddParticipantAction.ResultHandler
+    ) {
+        Logging.mls.info("adding \(users.count) participants to conversation (\(String(describing: conversation.qualifiedID)))")
+
+        guard
+            let mlsService = getMLSService(fromSyncContext: context.zm_sync),
```
caldrian: I find it a bit weird, that contexts have weak references to each other. I wonder why this approach was chosen.
Wouldn't it be better to ensure in `init`, that the context "is a sync context" instead of getting another context through the injected context?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsService.swift
```diff
@@ -0,0 +1,141 @@
...
+        // If we don't copy the id here (contexts thread), then the app will
+        // crash if we try to use it in the task (not on the contexts thread).
+        let qualifiedID = conversation.qualifiedID
+        let mlsUsers = users.compactMap(MLSUser.init(from:))
+
+        Task {
+            do {
+                try await mlsService.addMembersToConversation(with: mlsUsers, for: groupID)
+
+                await context.perform {
```
caldrian, line 69:
```suggestion
                await context.perform(schedule: .enqueued) {
```

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSClientIDsProvider.swift
```diff
@@ -1,5 +1,6 @@
+////
```
caldrian, line 1: Are the four slashes actually intended? I have seen that in some files and was about to replace them with two slashes.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsService.swift
```diff
@@ -0,0 +1,141 @@
...
+// MARK: - Helpers
+
+private extension MLSConversationParticipantsService {
+    func getMLSService(
+        fromSyncContext context: NSManagedObjectContext
+    ) -> MLSServiceInterface? {
+
+        var mlsService: MLSServiceInterface?
+
+        context.performAndWait {
```
caldrian: Can't the `userInfo` of a context be accessed without `performAndWait`?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationParticipantsService.swift
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        switch (conversation.messageProtocol, user.isSelfUser) {
+
+        case (.proteus, _), (.mixed, _), (.mls, true):
+            proteusParticipantsService.removeParticipant(
+                user,
+                from: conversation,
+                completion: completion
+            )
```
johnxnguyen: I don't understand why we remove the self user from an mls group via the proteus service.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSClientIDsProvider.swift
johnxnguyen: No I think it's something wrong with the header generation.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationParticipantsService.swift
David-Henner: yes me neither, maybe @netbe can comment on that?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsService.swift
```diff
@@ -0,0 +1,142 @@
...
+
+    init?(context: NSManagedObjectContext) {
+
+        guard let syncContext = context.performAndWait({ context.zm_sync }) else {
+            return nil
+        }
+
+        guard let mlsService = syncContext.performAndWait({ syncContext.mlsService }) else {
+            return nil
+        }
```
David-Henner, line 60: To avoid doing this we could enforce the context to be the sync context

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsService.swift
David-Henner: I don't think so

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ProteusConversationParticipantsService.swift
David-Henner: done

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsService.swift
David-Henner: done

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationParticipantsService.swift
David-Henner: @netbe bump ⬆️

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsService.swift
David-Henner: I agree it would be better to be on the correct context. I've decided not to change this yet because of time constraints.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsService.swift
```diff
@@ -0,0 +1,141 @@
...
+            Logging.mls.warn("failed to add participants to conversation (\(String(describing: qualifiedID))): missing group ID")
+            throw MLSConversationParticipantsError.invalidOperation
+        }
+
+        let mlsUsers = await context.perform { users.compactMap(MLSUser.init(from: )) }
+
+        do {
+            try await mlsService.addMembersToConversation(with: mlsUsers, for: groupID)
+        } catch MLSService.MLSAddMembersError.failedToClaimKeyPackages {
+            // TODO: Refactor key package claim in `MLSService` to throw an error with the users that couldn't be added
```
caldrian, line 102: How can we keep track of this TODO?

caldrian (approved):
As always the PR is very well structured! I also appreciate that you added comments to complicated sections.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationParticipantsService.swift
netbe: did I write that?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationParticipantsService.swift
```diff
@@ -0,0 +1,243 @@
...
+
+        case .mixed:
+
+            guard let mlsParticipantsService else {
+                throw ConversationParticipantsError.missingMLSParticipantsService
+            }
+
+            try await proteusParticipantsService.addParticipants(users, to: conversation)
+
+            // For mixed protocol we only try once and don't handle errors
```
netbe, line 128: great, thanks for the comment

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsService.swift
netbe: add ticket in jira and mention in code

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsServiceTests.swift
```diff
@@ -0,0 +1,189 @@
+////
```
netbe:
```suggestion
//
```

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsServiceTests.swift
```diff
@@ -0,0 +1,189 @@
...
+        await assertItThrows(error: MLSConversationParticipantsError.invalidOperation) {
+            // WHEN
+            try await sut.addParticipants([user], to: conversation)
+        }
+    }
+
+    func test_AddParticipants_RethrowsErrors() async {
+        // GIVEN
+        mockMLSService.addMembersToConversationMock = { _, _ in
+            throw ParticipantsError.failedToAddParticipants
```
netbe, line 107: is the real MLSService throw this error too? if so consider making ParticipantsError available

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ProteusConversationParticipantsService.swift
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    func addParticipants(
+        _ users: [ZMUser],
+        to conversation: ZMConversation
+    ) async throws {
+        var action = AddParticipantAction(users: users, conversation: conversation)
+
+        do {
+            try await action.perform(in: context.notificationContext)
```
netbe, line 52: I still don't get what is actually notificationContext under?

netbe: persistentStoreCoordinator ?

wire-ios/Wire-iOS/Sources/Helpers/syncengine/Conversation+Participants.swift
```diff
@@ -38,66 +39,110 @@ extension ZMConversation {
     }
 
     func addOrShowError(participants: [UserType]) {
-        guard let session = ZMUserSession.shared(),
-                session.networkState != .offline else {
-            self.showAlertForAdding(for: NetworkError.offline)
-            return
+        guard
```
netbe, line 42: Where is this called in the UI? Can we modify it to pass the userSession?
I would suggest to move all this to a useCase. UI calls the UseCase.

wire-ios/Wire-iOS/Sources/Helpers/syncengine/Conversation+Participants.swift
```diff
@@ -38,66 +39,110 @@ extension ZMConversation {
...
+        }
+
+        guard
+            let session = ZMUserSession.shared(),
+            session.networkState != .offline
+        else {
+            return fail(with: ConversationError.offline)
+        }
+
+        guard let user = user.materialize(in: session.viewContext) else {
```
netbe, line 88: what does this method do?

netbe (changes requested):
good rework!

1) I think we should get rid of also some logic updated and burried in model - UseCase

2) Maybe found a better naming than `service` for `MLSConversationParticipantsService` and `ProteusMLSConversationParticipantsService` in order to have a better understanding of hierarchy.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsService.swift
David-Henner: I will do this in another PR for the same ticket as the one for this PR https://wearezeta.atlassian.net/browse/WPB-529

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsServiceTests.swift
David-Henner: This is just an error I created within the tests in order to verify that the `addParticipants` method is rethrowing errors it receives

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ProteusConversationParticipantsService.swift
David-Henner: It **is** the `persistentStoreCoordinator`

wire-ios/Wire-iOS/Sources/Helpers/syncengine/Conversation+Participants.swift
David-Henner: it materialises a list of objects conforming to `UserType` into `ZMUser` either by fetching them from the store if they exist, or by creating them.

wire-ios/Wire-iOS/Sources/Helpers/syncengine/Conversation+Participants.swift
David-Henner: Is this what you had in mind ? https://github.com/wireapp/wire-ios/pull/696/commits/237afb77d602d4f9abfec340f21df4498aa5b2ba

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
```diff
@@ -244,7 +244,14 @@ final class ConversationCreationController: UIViewController {
                 values.participants = parts
             }
 
-            let participantsController = AddParticipantsViewController(context: .create(values), selfUser: selfUser)
+            guard let session = ZMUserSession.shared() else { return }
```
David-Henner: will update in follow up commit

wire-ios/Wire-iOS/Sources/Helpers/syncengine/Conversation+Participants.swift
David-Henner: If we agree the implementation is right then I'll do the same for removing participants and I'll add tests

David-Henner:
> Maybe found a better naming than service for MLSConversationParticipantsService and ProteusMLSConversationParticipantsService in order to have a better understanding of hierarchy.

Any suggestions for this?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationParticipantsService.swift
```diff
@@ -0,0 +1,243 @@
...
+        users: [ZMUser],
+        conversation: ZMConversation
+    ) async throws {
+        try await context.perform { [self] in
+            guard
+                conversation.conversationType == .group,
+                !users.isEmpty,
+                !users.contains(ZMUser.selfUser(in: context))
+            else {
+                throw ConversationParticipantsError.invalidOperation
```
johnxnguyen, line 143: Do you think it might be helpful to include a message describing what is invalid? E.g "conversation is not a group" etc

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationParticipantsService.swift
```diff
@@ -0,0 +1,243 @@
...
+        switch error {
+        case .unreachableDomains(let domains):
+            let unreachableUsers = await context.perform { users.belongingTo(domains: domains) }
+
+            if unreachableUsers.isEmpty {
+
+                /// If there are no users from unreachable domains, this means that the backend tried and failed to check for non-fully connected graphs
+                /// because some of the existing participants are currently unreachable.
+                /// As a result, we should inform that users can't be added and not retry the request.
+                /// https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/822149401/Non-fully+connected+federation+graphs
```
johnxnguyen: I think we could simplify this comment with something like:

```
// Backend is not able to determine which users are unreachable.

File: wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsService.swift (hunk 8 above)
Comment (johnxnguyen) [reply wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsService.swift:60]: We could also just inject the mls service and a context, and make it the responsibility of the caller to ensure it's the sync context.

File: wire-ios/Wire-iOS/Sources/Helpers/syncengine/Conversation+Participants.swift [hunk 19]
```diff
@@ -37,67 +34,63 @@ extension ZMConversation {
...
-    func addOrShowError(participants: [UserType]) {
-        guard let session = ZMUserSession.shared(),
-                session.networkState != .offline else {
-            self.showAlertForAdding(for: NetworkError.offline)
-            return
+    func removeOrShowError(participant user: UserType, completion: ((VoidResult) -> Void)? = nil) {
+
+        @Sendable func fail(with error: Error) {
+            showAlertForRemoval(for: error)
+            completion?(.failure(error))
```
johnxnguyen: It's probably out of scope of this PR, but I think we should factor away methods like this from managed objects. I see there's a use case for adding participants, why not one for removing participants?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationParticipantsService.swift
David-Henner: could @KaterinaWire comment on this? I think she wrote the original comment

wire-ios/Wire-iOS/Sources/Helpers/syncengine/Conversation+Participants.swift
David-Henner: There's one for removing participants but I haven't pushed it yet. I was waiting for feedback from @netbe before completing the implementation and starting to write tests

David-Henner: I also agree it's a bit out of the scope and I was considering making this in another PR

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationParticipantsService.swift
David-Henner: As of now, the caller doesn't need that level of detail, since it shows a generic error message to the user. But it could be improved later on

wire-ios/Wire-iOS/Sources/Helpers/syncengine/Conversation+Participants.swift
David-Henner: Will be done here https://wearezeta.atlassian.net/browse/WPB-5137

wire-ios/Wire-iOS/Sources/Helpers/syncengine/Conversation+Participants.swift
David-Henner: will be done here https://wearezeta.atlassian.net/browse/WPB-5137

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationParticipantsService.swift
KaterinaWire: I agree with John's suggestion. I would also add that in this case we shouldn't retry the request.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsService.swift
David-Henner: @netbe here

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsService.swift
David-Henner: I agree. But it would just move the problem to `ConversationParticipantsService` and I don't see much value in that. I'd rather wait until `mlsService` is decoupled from the context

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationParticipantsService.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/696/commits/b95a5350ed59908e8d9578ebf12399ddb197650f

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationParticipantsService.swift
David-Henner: actually it was me 😅

David-Henner: https://github.com/wireapp/wire-ios-data-model/pull/1409

> We need to make a distinction between removing other participants from the group and removing the self user from the group:
>
> - When removing others, we ask core crypto to remove members, which returns a commit that we then send to the group.
> - When removing the self user, we should use the RemoveParticipantAction which will ask the backend to remove the user. The backend then notifies the user's clients that they've been removed and this is where we'll ask core crypto to wipe the conversation (in RS).

netbe (approved):

PR #697, opened 2023-11-21
feat: establish mls one to one group WPB-2182

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-2182" title="WPB-2182" target="_blank">StoryWPB-2182</a>  [iOS] [Clients] Select protocol for 1-1 conversation
  </td></table>
  <br />
 

----

## What's in this PR?

This PR establishes mls one to one conversations when a connection is accepted.

Given that A sent a connection to request to B, when B accepts:
- the common supported protocols between the two users is determined.
- If there are no common protocols, the conversation is marked read only.
- If only proteus is supported, then nothing more is done because the proteus one to one conversation already exists.
- If mls is supported, then B will establish the mls group via core crypto and add A to the group.
- A will receive an mls welcome message and process it as normal, then resolve to use the mls conversation for the one on one.
- The old proteus one to one conversation is no longer used for the connection and it is hidden from the list (since it's now considered invalid)

Changes:
- `OneOnOneResolver` resolves which one on one conversation to use for a specific user. Internally it will find the common protocol, update the local state. This is currently called after accepting the connection and receiving a welcome message. 
- `OneOnOneProtocolSelector` resolves which message protocol to use with a specific user.
- `OneOnOneMigrator` migrates a one on one conversation to be used for mls. It will establish the mls group if needed and add the other user to the group, as well as switch the mls conversation to be used for the connection.

## Tests
 
- [x] Unit tests for `OneOnOneResolver`
- [x] Unit tests for `OneOnOneProtocolSelector`
- [x] Unit tests for `OneOnOneMigrator`
- [x] Unit test asserting one on one is resolved after accepting connection
- [x] Unit test asserting one on one is resolved after processing a welcome message
- [x] Unit test asserting invalid one on ones aren't shown in list

## How to test
- User A logs in with a single client to an mls enabled backend and team.
- User B logs in with a single client to the same backend but from a different mls enabled team.
- A sends a connection request to B.
- B accepts the request.
- Assert that both sides see a single mls one to one conversation and can send messages.


wire-ios-data-model/Source/Model/User/ZMUser.swift
```diff
@@ -406,15 +406,111 @@ extension ZMUser: UserConnections {
...
+                    // The connection now uses the mls 1-1.
+                    connection.conversation = conversation
+                }
+
+                await MainActor.run {
+                    completion(nil)
+                }
+            }
+        }
     }
```
markusfassbender: When reading this I feel like it's not a good idea to start running tasks from within a `context.perform`. This way we really mix a lot the different concurrency patterns.

We could write the code slightly different to make the whole function `async` and move the completion handle outside:
```swift
    private func establishMLSOneToOne(
        with otherUserID: QualifiedID,
        in context: NSManagedObjectContext
    ) async throws {
        let mlsService = try await context.perform {
            guard let mlsService = context.mlsService else {
                throw AcceptConnectionError.missingMLSService
            }
            return mlsService
        }

        let mlsGroupID = try await mlsService.establishOneToOneGroupIfNeeded(
            with: otherUserID,
            in: context
        )

        try await context.perform {
            guard
                let otherUser = ZMUser.fetch(
                    with: otherUserID,
                    in: context
                ),
                let connection = otherUser.connection,
                let conversation = ZMConversation.fetch(
                    with: mlsGroupID,
                    in: context
                )
            else {
                throw AcceptConnectionError.unableToSwitchToMLS
            }

            // The existing proteus 1-1 is no longer valid.
            if connection.conversation.messageProtocol == .proteus {
                connection.conversation.conversationType = .invalid
            }

            // The connection now uses the mls 1-1.
            connection.conversation = conversation
        }
    }
```

markusfassbender: Calling it would look like this (maybe we need to make sure the `MainActor` is used, but not 100% clear to me right now:
```swift
                Task {
                    do {
                        try await self.establishMLSOneToOne(
                            with: qualifiedID,
                            in: syncContext
                        )
                        completion(nil)
                    } catch {
                        completion(error)
                    }
                }
```

markusfassbender: Or maybe it's even better to refactor the calling function `async` too, since it executes async code.

wire-ios-request-strategy/Sources/Payloads/Processing/ConversationEventPayloadProcessor.swift
```diff
@@ -543,8 +574,14 @@ final class ConversationEventPayloadProcessor {
     ) {
         conversation.domain = BackendInfo.isFederationEnabled ? payload.qualifiedID?.domain : nil
         conversation.needsToBeUpdatedFromBackend = false
-        conversation.epoch = UInt64(payload.epoch ?? 0)
-        conversation.mlsGroupID = payload.mlsGroupID.flatMap { MLSGroupID(base64Encoded: $0) }
+
+        if let epoch = payload.epoch.flatMap(UInt64.init) {
+            conversation.epoch = epoch
+        }
```
markusfassbender: I find this `flatMap` usage harder to read, than normal casting, but I know it's personal preference.
```swift
        if let epoch = payload.epoch  {
            conversation.epoch = UInt64(epoch)
        }
```

wire-ios-request-strategy/Sources/Request Strategies/MLS/Actions/SyncMLSOneToOneConversationActionHandler.swift
```diff
@@ -0,0 +1,103 @@
...
+
+        case (400, "mls-not-enabled"):
+            action.fail(with: .mlsNotEnabled)
+
+        case (403, "not-connected"):
+            action.fail(with: .usersNotConnected)
+
+        default:
+            let errorInfo = response.errorInfo
+            action.fail(with: .unknownn(status: errorInfo.status, label: errorInfo.label, message: errorInfo.message))
```
markusfassbender: Typo `unknownn` change to `unknown`

wire-ios-sync-engine/Source/Services/SupportedProtocolsService.swift
```diff
@@ -170,6 +170,10 @@ private extension UserClient {
     }
 
     var isRecentlyActive: Bool {
+        guard !isSelfClient() else {
+            return true
+        }
```
markusfassbender: Maybe easier to read?
```swift
if isSelfClient() {
    return true
}
```

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -406,6 +411,46 @@ public final class MLSService: MLSServiceInterface {
...
+    ) async throws -> MLSGroupID {
+        WireLogger.mls.debug("syncing mls one to one")
+
+        var action = SyncMLSOneToOneConversationAction(
+            userID: userID.uuid,
+            domain: userID.domain
+        )
+
+        return try await action.perform(in: context.notificationContext)
+    }
```
David-Henner: Any reason why this didn't get added to `MLSActionsProvider` ?

wire-ios-data-model/Source/Model/User/ZMUser.swift
```diff
@@ -406,15 +406,111 @@ extension ZMUser: UserConnections {
...
+            guard let mlsService = context.mlsService else {
+                completion(AcceptConnectionError.missingMLSService)
+                return
+            }
+
+            Task {
+                let mlsGroupID = try await mlsService.establishOneToOneGroupIfNeeded(
+                    with: otherUserID,
+                    in: context
+                )
```
David-Henner: I don't think the completion would be called if this throws

wire-ios-request-strategy/Sources/Payloads/Processing/ConversationEventPayloadProcessor.swift
```diff
@@ -484,54 +496,73 @@ final class ConversationEventPayloadProcessor {
...
+                existingConversation.messageProtocol == .proteus,
+                conversation.messageProtocol == .mls
+            {
+                // Invalidate old proteus conversation, make mls conversation active.
+                existingConversation.conversationType = .invalid
+                connection.conversation = conversation
+            }
+
+            conversation.isPendingMetadataRefresh = otherUser.isPendingMetadataRefresh
+        }
```
typfel: Confirming for myself what this code does: 

If receive an event about a new MLS 1-1 conversation then we look up the current "active" 1-1 conversation with the same user. If this conversation a proteus conversation we invalidate and replace it with the MLS 1-1 conversation.

Shouldn't we also check what common supported protocols are before we switch to use the MLS 1-1 conversation?

typfel: Also is this supposed to resolve to the correct 1-1 conversation being set on `connection.conversation` after a slow sync? But wouldn't that only work if the MLS conversations are fetched after the proteus conversations during the slow sync, but the order is undefined I think.

typfel (changes requested):

wire-ios-data-model/Source/Model/User/ZMUser.swift
johnxnguyen: Actually I think what I wanted to do was switch to the context to get the mls service, then pass that into a task after the perform block, but that's not what I wrote 😅

I wanted to make it async all the way up but it got too difficult to manage so I avoided it for now. But I'll try your suggestion.

wire-ios-data-model/Source/MLS/MLSService.swift
johnxnguyen: I guess I didn't need to. The action provider is a way to be able to mock the actions, but I found this is also achievable by using a mock action handler.

wire-ios-request-strategy/Sources/Payloads/Processing/ConversationEventPayloadProcessor.swift
johnxnguyen: > Shouldn't we also check what common supported protocols are before we switch to use the MLS 1-1 conversation?

@typfel this is assuming that if we become aware of the mls one on one then it's because we've been added to the group and the supported protocol check was already done. But indeed I see that it may be a flaky assumption to make. I guess this goes back to our previous discussion about leaving the payload processor for only updating the local objects from remote objects and finding a better place to move the logic of switching the mls one on one.

I'll see if I can move it to where we receive the welcome message.

typfel (commented):
Looks good!

wire-ios-data-model/Source/Model/User/ZMUser.swift
johnxnguyen: Fixed

wire-ios-request-strategy/Sources/Payloads/Processing/ConversationEventPayloadProcessor.swift
johnxnguyen: Fixed

wire-ios-request-strategy/Sources/Request Strategies/MLS/Actions/SyncMLSOneToOneConversationActionHandler.swift
johnxnguyen: Fixed

wire-ios-sync-engine/Source/Services/SupportedProtocolsService.swift
johnxnguyen: Fixed

wire-ios-data-model/Source/MLS/MLSService.swift
johnxnguyen: This is no longer relevant anymore

wire-ios-data-model/Source/Model/User/ZMUser.swift
johnxnguyen: This is no longer relevant

wire-ios-request-strategy/Sources/Payloads/Processing/ConversationEventPayloadProcessor.swift
johnxnguyen: Refactored

markusfassbender (approved):

wire-ios-request-strategy/Sources/Payloads/Processing/Helpers/MLSEventProcessor.swift
```diff
@@ -144,31 +144,96 @@ public class MLSEventProcessor: MLSEventProcessing {
...
+            oneOnOneResolver: oneOnOneResolver
+        )
+    }
+
+    func process(
+        welcomeMessage: String,
+        conversationID: QualifiedID,
+        in context: NSManagedObjectContext,
+        mlsService: MLSServiceInterface,
+        oneOnOneResolver: OneOnOneResolverInterface
```
typfel, line 167: Why are injecting these in the method? Can't they be properties of the `MLSEventProcessor`?

wire-ios-request-strategy/Sources/Payloads/Processing/Helpers/MLSEventProcessor.swift
```diff
@@ -144,31 +144,96 @@ public class MLSEventProcessor: MLSEventProcessing {
...
+        } catch {
+            WireLogger.mls.warn("MLS event processor aborting processing welcome message: \(String(describing: error))")
+            return
+        }
+    }
+
+    private func resolveOneOnOneConversationIfNeeded(
+        conversation: ZMConversation,
+        mlsService: MLSServiceInterface,
+        oneOneOneResolver: OneOnOneResolverInterface,
```
typfel, line 206: Also there could be properties of the `MLSEventProcessor`

typfel (changes requested):
Looking mostly good with some nitpicks

wire-ios-request-strategy/Sources/Payloads/Processing/Helpers/MLSEventProcessor.swift
David-Henner: In order to have these as properties of the `MLSEventProcessor`, we would need to either:
- Inject the `MLSService` from the `ZMUserSession` down into all the classes that depend on payload processors
- Make them both optional properties

David-Henner: And I believe that the injection of the `MLSService` is covered by this ticket https://wearezeta.atlassian.net/browse/WPB-5827

typfel: Ok, all good then!

typfel (approved):

PR #699, opened 2023-11-21
feat(mls-migration): refactor adding members to MLS conversation WPB-529

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-529" title="WPB-529" target="_blank">StoryWPB-529</a>  [iOS] Maintaining MLS group's clients list during migration
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR updates the logic for adding members to an MLS conversation.

The current logic is to throw an error and abort the operation as soon as we fail to fetch key packages for one of the users we're trying to add to the conversation. This would result in not adding **any** participants if we only failed to get key packages for one or few of them. The updated logic would log an error for the participants we failed to add, but still add the participants for which we successfully fetched key packages.

This is needed in the context of the MLS Migration, because when adding participants in `mixed` mode, we need to add them to the Proteus part of the conversation, and attempt to add them to the MLS part of conversation.

This is also the expected behaviour for `mls` conversations

### Testing


wire-ios-data-model/Tests/MLS/MLSServiceTests.swift
```diff
@@ -468,6 +468,48 @@ class MLSServiceTests: ZMConversationTestsBase, MLSServiceDelegate {
...
+        // Mock claiming a key package. Works for user1, throws for user2
+        mockActionsProvider.claimKeyPackagesUserIDDomainExcludedSelfClientIDIn_MockMethod = { userID, _, _, _ in
+            if userID == userID1 {
+                return [keyPackage]
+            } else {
+                throw ClaimMLSKeyPackageAction.Failure.userOrDomainNotFound
+            }
+        }
+
+        // Mock adding members to the conversation
```
agisilaos: Love these comments. ⚡

agisilaos (approved):
Nice! 👏

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -515,42 +514,29 @@ public final class MLSService: MLSServiceInterface {
...
+            do {
+                let keyPackages = try await actionsProvider.claimKeyPackages(
+                    userID: user.id,
+                    domain: user.domain,
+                    excludedSelfClientID: user.selfClientID,
+                    in: context.notificationContext
+                )
                 result.append(contentsOf: keyPackages)
+            } catch {
+                logger.warn("failed to claim key packages for user (\(user.id)): \(String(describing: error))")
```
typfel, line 539: If i'm thinking ahead about how will want to insert a system message informing about which users's did not get added, I don't think we can swallow the error. 

Rather I think will need a more detailed error which includes which users we couldn't claim key packages for:
`failedToClaimKeyPackages(users: [MLSUser])`

The caller could then exclude these users and an retry the operation and then inserted a system mesasge with the excluded users.

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -515,42 +514,29 @@ public final class MLSService: MLSServiceInterface {
...
+
+        guard let context = context else { return [] }
 
-            var result = [KeyPackage]()
+        var result = [KeyPackage]()
 
-            for try await keyPackages in claimKeyPackages(for: users, in: context) {
+        for user in users {
+            do {
+                let keyPackages = try await actionsProvider.claimKeyPackages(
```
typfel, line 530: I noticed that we don't throw an error if a user a has run out out of key packages, the backend will then return an empty array of key packages. I think we should so that we can correctly inform why a user couldn't be added.

typfel (changes requested):

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationParticipantsService.swift
```diff
@@ -0,0 +1,257 @@
...
+            }
+
+            try await proteusParticipantsService.addParticipants(users, to: conversation)
+
+            // For mixed protocol we only try once and don't handle errors
+            try? await addMLSParticipants(users, to: conversation, showErrors: false)
+        }
+    }
+
+    private func addMLSParticipants(_ users: [ZMUser],
```
netbe: @David-Henner (test) added the function here

wire-ios-request-strategy/Sources/Request Strategies/Conversation/MLSConversationParticipantsServiceTests.swift
```diff
@@ -38,7 +38,7 @@ class MLSConversationParticipantsServiceTests: MessagingTestBase {
 
         mockClientIDsProvider = MockMLSClientIDsProviding()
         mockMLSService = MockMLSService()
-
+
```
David-Henner, line 41:
```suggestion
```

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationParticipantsService.swift
David-Henner: added tests

wire-ios-data-model/Source/MLS/MLSService.swift
David-Henner: Logic has been updated

wire-ios-data-model/Source/MLS/MLSService.swift
David-Henner: Empty key package list is now handled as an error

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationParticipantsService.swift
```diff
@@ -126,7 +117,45 @@ public class ConversationParticipantsService: ConversationParticipantsServiceInt
...
+                // Operation was aborted because some users didn't have key packages
+                // We filter them out and retry once
+
+                try await internalAddParticipants(
+                    Array(users.subtracting(failedUsers)),
+                    to: conversation
+                )
+            }
+
+            if showErrors {
```
typfel: I wouldn't catch the error here but rather bubble up handle it at the same level as `handleFederationError` and re-use the logic for inserting system message.

I understand that isn't in the scope of this PR, but how about letting the error bubble up here and catch and swallow it when adding participants in mixed mode and leave the error handling for WPB-2228

typfel (changes requested):

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationParticipantsService.swift
David-Henner: like that https://github.com/wireapp/wire-ios/pull/699/commits/327a79b95064e073b2204031f8a1242b1d7704ed ?

typfel: Yes

typfel (approved):

PR #702, opened 2023-11-22
fix: core data add gradual migration [WPB-5591]

markusfassbender:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5591" title="WPB-5591" target="_blank">BugWPB-5591</a>  [iOS] Not enough disk space in app
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

> The automatic migrations are performed as one-step migrations; directly from the source to destination model. So if we support 4 model versions, mapping models would exist for 1 to 4, 2 to 4 and 3 to 4. While this is the most efficient migration approach from a device performance point-of-view, it can actually be quite wasteful from a development point-of-view. For example if we added a new model version (5) we would need to create 4 new mapping models from 1 to 5, 2 to 5, 3 to 5 and 4 to 5 which as you can see doesn't reuse any of the mapping models for migrating to version 4. With a one-step migration approach, each newly added model version requires n-1 mapping models (where n is the number of supported model versions) to be created.
> https://williamboles.com/progressive-core-data-migration/

### Solutions

1. Implement here a gradual migration strategy for core data, that executes the migration one step after another.
2. Important from a backup file needs to add the data to the original Core Data model, then do the migration step by step to the latest model.

### Testing

Changes
- Remove test database for version 25, because the core data model does not exist anymore
- Update test database for 105 and 108, because they not matching the core data models

#### Test Coverage (Optional)

#### How to Test

**TBD**: We need old app builds with older core data models.

----

----

----
##### References


wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMessagingMigrator.swift
```diff
@@ -0,0 +1,228 @@
...
+        return try migrationSteps(fromSourceVersion: sourceVersion, toDestinationVersion: destinationVersion)
+    }
+
+    private func migrationSteps(
+        fromSourceVersion sourceVersion: CoreDataMessagingMigrationVersion,
+        toDestinationVersion destinationVersion: CoreDataMessagingMigrationVersion
+    ) throws -> [CoreDataMessagingMigrationStep] {
+        var sourceVersion = sourceVersion
+        var migrationSteps: [CoreDataMessagingMigrationStep] = []
+        // TODO: [F] check possible flaw in nextVersion
```
markusfassbender: @netbe Still open `TODO`?

wire-ios/Wire-iOS Tests/Utils/MockImageResource.swift
```diff
@@ -19,7 +19,7 @@
 import Foundation
 @testable import Wire
 
-class MockImageResource: NSObject, ImageResource {
+class MockImageResource: NSObject, WireImageResource {
```
markusfassbender: We need to revert the Xcode 15 commit 5f63b5ff1f0c314c8bd1761892585e7ea27f4e23

wire-ios-data-model/Tests/Source/ManagedObjectContext/DatabaseMigrationTests+UserClientUniqueness.swift
```diff
@@ -29,34 +29,114 @@ final class DatabaseMigrationTests_UserClientUniqueness: DatabaseBaseTest {
...
+        /*
+         Failed stores aka incompatible: 2.108, 2.105
+         Stores compatible: 2.109, 2.107
+         */
+        try migrateStoreToCurrentVersion(
+            sourceVersion: initialVersion,
+            preMigrationAction: insertDuplicates,
+            postMigrationAction: assertDuplicatesResolved
+        )
+    }
```
markusfassbender: Isn't the version already covered with test `testThatItPerformsMigrationFromOldVersionsBefore107_ToCurrentModelVersion()`?

wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack.swift
```diff
@@ -190,6 +193,42 @@ public class CoreDataStack: NSObject, ContextProvider {
...
+        DispatchQueue.global(qos: .userInitiated).async {
+            if self.needsMessagingStoreMigration() {
+                WireLogger.localStorage.info("start migration of core data messaging store!")
+
+                do {
+                    try self.migrateMessagingStore()
+                    WireLogger.localStorage.info("finished migration of core data messaging store!")
+                } catch {
+                    WireLogger.localStorage.error("failed migration of core data messaging store!")
+                    onFailure(error)
```
David-Henner: should this be called from the main thread?

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMessagingMigrationVersion.swift
```diff
@@ -0,0 +1,171 @@
...
+        case .version2_108:
+            return .version2_109
+        case .version2_107:
+            return .version2_108
+        case .version2_106:
+            return .version2_107
+        default:
+            // all older versions can be inferred to version 106
+            return .version2_106
+        }
```
David-Henner: so we'll need to manually add the next version each time we bump the data model ?

markusfassbender: Yes I think so

wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack.swift
netbe: fixed

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMessagingMigrationVersion.swift
netbe: I am adding a test to make sure we don't forget

netbe (approved):

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMessagingMigrationVersion.swift
David-Henner: Just wondering if there's a way to do this automatically. Perhaps something similar to 

```swift
if self < .version2_106 {
   return .version2_106
} else if self == Self.allCases.first {
   return nil
} else {
   return Self.allCases(self.index - 1)
}
```

David-Henner: Also I'm not sure I understand why all versions older than 2.106 can be inferred to 2.106. I guess we don't need to perform lightweight migrations for them? They're done automatically? Then starting from to 2.106 we have to do them all incrementally because otherwise we miss the standard migration from 2.106 to 2.107 ?

David-Henner (approved):
Approving since my previous comments shouldn't be blockers

PR #749, opened 2023-12-04
feat: handle conversation-leave event with reason user-deleted [WPB-4544]

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4544" title="WPB-4544" target="_blank">StoryWPB-4544</a>  [iOS] Stop processing team events
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The iOS client must be prepared to not receiving team events (join, leave) any more and ensure, the state is synced through other events.

### Causes (Optional)

Team events are deprecated.

### Solutions

Use `conversation.member-leave` with its new `reason` field in order to know, when a user has been deleted from the team.

----

----

----
##### References


wire-ios-request-strategy/Sources/Payloads/Processing/ConversationEventPayloadProcessor.swift
```diff
@@ -126,18 +126,41 @@ final class ConversationEventPayloadProcessor {
             )
         }
 
-        let sender = fetchOrCreateSender(
+        let initiatingUser = fetchOrCreateSender(
```
typfel: The SystemMessage inserted above at line 123 doesn't take reason field into account when deciding which system message type to insert.

If the reason is `userDeleted` I would expect the system message to be of type `ZMSystemMessageTypeTeamMemberLeave`

typfel (changes requested):
Nice work, I think it's just missing the to consider reason field when inserting the system message.

wire-ios-request-strategy/Sources/Payloads/Processing/ConversationEventPayloadProcessor.swift
caldrian: Okay, I will try to fix this.
What about the reason values `removed` vs `left`?
Do they both need to be mapped to `ZMSystemMessageTypeParticipantsRemoved`?
@typfel

typfel: Yes, the are both mapped  to `ZMSystemMessageTypeParticipantsRemoved` the two cases are are separated in the UI by checking if the sender is the same as the user who left.

https://github.com/wireapp/wire-ios/blob/08f927af367ca4eec14a02e7db07d2e8452fdaaa/wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ParticipantsCellViewModel.swift#L60

Now, this is how it used to work before we had the reason field. We could also refactor and introduce a new `ZMSystemMessageTypeParticipantLeft` system message.

caldrian: Thank you.
I agree to refactoring, however, I prefer to wait until the backend team is clear about the `reason` value `"legalhold-policy-conflict"` or similar, see https://wearezeta.atlassian.net/browse/WPB-5849.

caldrian: > If the reason is userDeleted I would expect the system message to be of type ZMSystemMessageTypeTeamMemberLeave

Done, please check commit: cd8eb092ea4f2be5f0501881893f045456869443

typfel (approved):

wire-ios-request-strategy/Sources/Payloads/Processing/Helpers/MLSEventProcessor.swift
```diff
@@ -27,7 +28,7 @@ protocol MLSEventProcessing {
 
 }
 
-final class MLSEventProcessor: MLSEventProcessing {
+struct MLSEventProcessor: MLSEventProcessing {
```
agisilaos, line 32: Why do we change this to a struct? Is it related with the changes of this PR?

agisilaos (approved):

wire-ios-request-strategy/Sources/Payloads/Processing/Helpers/MLSEventProcessor.swift
caldrian: It's neither related nor required here.
I usually try to fix issues when I come across them.
Let me know, if you prefer this change to be reverted.

agisilaos: Let's keep it for now!

I think it would be beneficial for us to adopt a more structured approach in the future. When alterations like these are made, it would be helpful to have them included in a separate pull request, accompanied by a detailed explanation. This way, it becomes easier for the team to review and understand the context and rationale behind each change. Additionally, this method would improve our ability to track when and why specific modifications are implemented, enhancing overall transparency and efficiency in our workflow.

PR #778, opened 2023-12-13
test: fix flaky date formatter test

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The test was assuming a 24 hour clock when the locale is actually fixed to us_US in the test plan. The test would fail if run betwen 02:00 - 03:00 because the `hour` variable would then be 0.

### Solutions

Simplify test and and assume us_US locale.

----

----

----
##### References

### DIFF
--- a/wire-ios/Wire-iOS Tests/DateFormatterTests.swift
+++ b/wire-ios/Wire-iOS Tests/DateFormatterTests.swift
@@ -129,25 +129,18 @@ final class DateFormatterTests: XCTestCase {
         // GIVEN
         let twoHourBefore = Calendar.current.date(byAdding: .hour, value: -2, to: Date())!
         var hour = Calendar.current.component(.hour, from: twoHourBefore)
-        var meridiem = "AM"
+
         // to fit US 12hr format
-        if hour > 12 {
-            hour -= 12
-            meridiem = "PM"
-        } else if hour == 12 {
-            meridiem = "PM"
+        if hour == 0 {
+            hour = 12
         }
 
         // WHEN
         let dateString = twoHourBefore.formattedDate
 
         // THEN
-        XCTAssertFalse(dateString.contains("Just now"), "dateString is \(dateString)")
         // If two hours before is yesterday, dateString looks like "Yesterday, 11:17 PM"
-        XCTAssert(dateString.hasPrefix(String(hour)) ||
-            (dateString.replacingOccurrences(of: "Yesterday, ", with: "").hasPrefix(String(hour))
-            ), "hour is \(hour), dateString is \(dateString)")
-        XCTAssert(dateString.hasSuffix(meridiem), "meridiem is \(meridiem), dateString is \(dateString)")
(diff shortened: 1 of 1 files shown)


wire-ios/Wire-iOS Tests/DateFormatterTests.swift
```diff
@@ -129,25 +129,18 @@ final class DateFormatterTests: XCTestCase {
         // GIVEN
         let twoHourBefore = Calendar.current.date(byAdding: .hour, value: -2, to: Date())!
```
caldrian, line 130: I think we should set a fixed time.
We can also test several values, like shortly before/after midnight or noon.

wire-ios/Wire-iOS Tests/DateFormatterTests.swift
```diff
@@ -129,25 +129,18 @@ final class DateFormatterTests: XCTestCase {
...
         let dateString = twoHourBefore.formattedDate
 
         // THEN
-        XCTAssertFalse(dateString.contains("Just now"), "dateString is \(dateString)")
         // If two hours before is yesterday, dateString looks like "Yesterday, 11:17 PM"
-        XCTAssert(dateString.hasPrefix(String(hour)) ||
-            (dateString.replacingOccurrences(of: "Yesterday, ", with: "").hasPrefix(String(hour))
-            ), "hour is \(hour), dateString is \(dateString)")
-        XCTAssert(dateString.hasSuffix(meridiem), "meridiem is \(meridiem), dateString is \(dateString)")
+        XCTAssertTrue(dateString.contains(String(hour)), "hour is \(hour), dateString is \(dateString)")
```
caldrian, line 143: I don't like the assertion "contains string".
Can't we have `XCTAssertEqual` with the exact string?

wire-ios/Wire-iOS Tests/DateFormatterTests.swift
typfel: Mm I thought about it, but how do we when then deal with the fact that we format the time differently depending on if the date/time is from yesterday, 5 minutes ago or two hours ago.

I suppose we would then also need to set the **now** time to some fixed value.

netbe (approved):

wire-ios/Wire-iOS Tests/DateFormatterTests.swift
caldrian: I see.

caldrian (approved):

wire-ios/Wire-iOS Tests/DateFormatterTests.swift
typfel: We could if we switch to fixed dates because then it would be deterministic. Right now it the string might contain a "Yesterday, " prefix if you run it just after midnight.

If you look at the other tests they also work like this. You have valid points but my aim with this PR was just make the test non-flaky.

PR #784, opened 2023-12-14
feat: state machine for MLS conversations WPB-3873

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-3873" title="WPB-3873" target="_blank">StoryWPB-3873</a>  [iOS] Create separate unverified/verified/degraded state machine for MLS/E2EI
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

In this PR, we introduce a mechanism to set the unverified/verified/degraded state for mls conversations when the E2EI feature is enabled.
There are 3 states:
- unverified
- verified
- degraded

**What should we ask CC:**
The CC API can inform client whether a conversation is unverified or verified. 
The `degraded` state should be calculated on the client side. This state should be set if the conversation has changed from verified to unverified.

**When we should ask CC:**
- on epoch change;
- when opening a conversation (maybe an edge case on the backup for example);
- when we get the revocation list (will be implemented later);
- when a certificate of a member gets expired (not a part of this PR)

TODO:
1. Add a property for the conversation entity to the db.
2. Add tests.

----

----

----
##### References


wire-ios-data-model/E2EIdentity/MLSConversationVerificationStatusProvider.swift
```diff
@@ -0,0 +1,91 @@
...
+        syncContext: NSManagedObjectContext) {
+            self.e2eIVerificationStatusService = e2eIVerificationStatusService
+            self.syncContext = syncContext
+        }
+
+    // MARK: - Public interface
+
+    public func updateStatus(_ groupID: MLSGroupID) {
+        syncContext.performAndWait {
+            let coreCryptoStatus = e2eIVerificationStatusService.getConversationStatus(groupID: groupID)
```
typfel: Note that this functional call will probably turn async when we upgrade CC

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
```diff
@@ -550,6 +550,17 @@ public class ZMUserSession: NSObject {
...
+    private func observeOnEpochChangeIfNeeded() {
+        guard e2eiFeature.isEnabled,
+              let mlsService = self.syncContext.mlsService else {
+            return
+        }
+        let epochChangeToken = EpochChangeObserver(
+            mlsService: mlsService,
+            block: self.mlsConversationVerificationStatusProvider?.updateStatus(_:))
+        self.tokens.append(epochChangeToken)
+    }
```
typfel: We should aim to not put logic inside the `ZMUserSession` this code could live in a `MLSConversationVerificationManager` class

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
```diff
@@ -550,6 +550,17 @@ public class ZMUserSession: NSObject {
         }
     }
 
+    private func observeOnEpochChangeIfNeeded() {
+        guard e2eiFeature.isEnabled,
+              let mlsService = self.syncContext.mlsService else {
```
typfel: If you are update to with develop `mlsService is always available now so no need to fetch from the context.

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
```diff
@@ -550,6 +550,17 @@ public class ZMUserSession: NSObject {
...
+            return
+        }
+        let epochChangeToken = EpochChangeObserver(
+            mlsService: mlsService,
+            block: self.mlsConversationVerificationStatusProvider?.updateStatus(_:))
```
typfel: When `mlsConversationVerificationStatusProvider?.updateStatus(_:)` turn async this won't work. I propose to expose an `AsynsSequence` of epoch changes on the `MLSService`. See `onConferenceInfoChange`, it would then look something like:

```
func startObservingMLSConversationVerificationStatus() {
   Task {
      for groupID in mlsService.epochChanges {
         mlsConversationVerificationStatusProviderupdateStatus(groupID)
      }
   }
}

**Review by typfel — CHANGES_REQUESTED** [review]

File: wire-ios-data-model/E2EIdentity/MLSConversationVerificationStatusProvider.swift (hunk 1 above)
Comment (KaterinaWire) [reply wire-ios-data-model/E2EIdentity/MLSConversationVerificationStatusProvider.swift]: Yes, it will be. 
Changed

File: wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift (hunk 2 above)
Comment (KaterinaWire) [reply wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift]: I removed EpochChangeObserver

File: wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift (hunk 4 above)
Comment (KaterinaWire) [reply wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift]: Added

File: wire-ios-data-model/Source/E2EIdentity/MLSConversationVerificationStatusProvider.swift [hunk 5]
```diff
@@ -0,0 +1,99 @@
...
+            self.syncContext = syncContext
+        }
+
+    // MARK: - Public interface
+
+    public func updateStatus(_ groupID: MLSGroupID) async {
+        syncContext.performAndWait {
+            guard let conversation = ZMConversation.fetch(with: groupID, in: syncContext) else {
+                return
+            }
```
typfel: It's not allowed to use `performAndWait` inside a async function.

```suggestion
       guard let conversation = await syncContext.perform({
            ZMConversation.fetch(with: groupID, in: syncContext)
       }) else {
            return
       }
```

wire-ios-data-model/Source/E2EIdentity/MLSConversationVerificationStatusProvider.swift
```diff
@@ -0,0 +1,99 @@
...
+            Task {
```
typfel: No need to create a `Task ` you are already in an async context.

wire-ios-data-model/Source/E2EIdentity/MLSConversationVerificationStatusProvider.swift
```diff
@@ -0,0 +1,99 @@
...
+    public init(
+        e2eIVerificationStatusService: E2eIVerificationStatusServiceInterface,
+        syncContext: NSManagedObjectContext) {
+            self.e2eIVerificationStatusService = e2eIVerificationStatusService
+            self.syncContext = syncContext
+        }
+
+    // MARK: - Public interface
+
+    public func updateStatus(_ groupID: MLSGroupID) async {
```
typfel: does it make sense let the function throw?

wire-ios-data-model/Source/E2EIdentity/MLSConversationVerificationStatusProvider.swift
```diff
@@ -0,0 +1,99 @@
...
+                }
+            }
+        }
+    }
+
+    // MARK: - Helpers
+
+    private func updateStatusAndNotifyUserIfNeeded(newStatusFromCC: MLSVerificationStatus,
+                                                   conversation: ZMConversation) {
+        syncContext.performAndWait {
```
typfel: I'd suggest to to remove this `syncContext.performAndWait` and it's the responsibility of the caller to do a `await syncContext.perform {}`

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
```diff
@@ -550,6 +550,17 @@ public class ZMUserSession: NSObject {
...
+    private func startObservingMLSConversationVerificationStatus() {
+        guard e2eiFeature.isEnabled else {
+            return
+        }
+        Task {
+            for try await groupID in mlsService.epochChanges() {
+                try await mlsConversationVerificationStatusProvider?.updateStatus(groupID)
+            }
+        }
+    }
```
typfel: Ok create, but now move this function into some "manager" class that makes it a lot easier to unit test since you don't need to deal with the whole user session.

typfel (changes requested):

wire-ios-data-model/Source/E2EIdentity/MLSConversationVerificationStatusProvider.swift
KaterinaWire: fixed

wire-ios-data-model/Source/E2EIdentity/MLSConversationVerificationStatusProvider.swift
KaterinaWire: yes

wire-ios-data-model/Source/E2EIdentity/MLSConversationVerificationStatusProvider.swift
KaterinaWire: fixed

wire-ios-data-model/Source/E2EIdentity/MLSConversationVerificationStatusProvider.swift
```diff
@@ -0,0 +1,97 @@
...
+    private var syncContext: NSManagedObjectContext
+
+    // MARK: - Life cycle
+
+    public init(
+        e2eIVerificationStatusService: E2eIVerificationStatusServiceInterface,
+        syncContext: NSManagedObjectContext) {
+            self.e2eIVerificationStatusService = e2eIVerificationStatusService
+            self.syncContext = syncContext
+        }
```
David-Henner:
```suggestion
    public init(
        e2eIVerificationStatusService: E2eIVerificationStatusServiceInterface,
        syncContext: NSManagedObjectContext
    ) {
        self.e2eIVerificationStatusService = e2eIVerificationStatusService
        self.syncContext = syncContext
    }
```

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMessagingMigrationVersion.swift
```diff
@@ -64,8 +65,10 @@ enum CoreDataMessagingMigrationVersion: String, CaseIterable {
 
     var nextVersion: Self? {
         switch self {
-        case .version2_111:
+        case .version2_112:
             return nil
+        case .version2_111:
+            return .version2_110
```
David-Henner: this looks wrong. should it be like this instead?

```swift
        case .version2_112:
            return nil
        case .version2_111:
            return .version2_112
```

wire-ios-data-model/Source/E2EIdentity/MLSConversationVerificationStatusProvider.swift
```diff
@@ -0,0 +1,97 @@
...
+
+    // MARK: - Helpers
+
+    private func updateStatusAndNotifyUserIfNeeded(newStatusFromCC: MLSVerificationStatus,
+                                                   conversation: ZMConversation) {
+        guard let currentStatus = conversation.mlsVerificationStatus else {
+            return conversation.mlsVerificationStatus = newStatusFromCC
+        }
+
+        let newStatus = getActualNewStatus(newStatusFromCC: newStatusFromCC, currentStatus: currentStatus)
```
David-Henner:
```suggestion
        let newStatus = resolveNewStatus(newStatusFromCC: newStatusFromCC, currentStatus: currentStatus)
```

wire-ios-data-model/Source/E2EIdentity/MLSConversationVerificationStatusProvider.swift
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    private func getActualNewStatus(newStatusFromCC: MLSVerificationStatus,
+                                    currentStatus: MLSVerificationStatus) -> MLSVerificationStatus {
+        switch (newStatusFromCC, currentStatus) {
+        case (.notVerified, .verified):
+            return .degraded
+        case(.notVerified, .degraded):
+            return .degraded
```
David-Henner, line 88:
```suggestion
        case (.notVerified, .verified), (.notVerified, .degraded):
            return .degraded
```

David-Henner: I kind of understand why the status is `degraded` when going from `verified` to `notVerified`. But why is it `degraded` when going from `degraded` to `notVerified` ?

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMessagingMigrationVersion.swift
KaterinaWire: Yes, my mistake. Fixed

wire-ios-data-model/Source/E2EIdentity/MLSConversationVerificationStatusProvider.swift
KaterinaWire: I would prefer to leave them on separate lines. In my opinion, it’s more readable.
The new state comes from the CC, and there is no `degraded` state on the CC. This means that CC only knows whether the conversation is verified or not. We calculate the degraded state on the client side. If it happens that the conversation becomes degraded, we should only change the status to verified.

David-Henner: I understand that we have two different statuses:
- The core crypto status -> (`.verified, .notVerified, .notEnabled`)
- The conversation status -> (`.verified, .notVerified, .degraded`)

And we choose to represent them both by the same type: `MLSVerificationStatus`. I find that a bit confusing when reading through the code, but unfortunately I can't think of a better way to do it. Any thoughts on this?

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
```diff
@@ -112,6 +112,8 @@ public class ZMUserSession: NSObject {
 
     public var appLockController: AppLockType
 
+    var mlsConversationVerificationManager: MLSConversationVerificationManagerInterface?
```
typfel: Do we rely need to have this class optional? Isn't enough to always create it an only call `startObservingMLSConversationVerificationStatus()` when e2ei is enabled?

typfel (changes requested):

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
KaterinaWire: Changed it to non optional

David-Henner (approved):

typfel (approved):

PR #847, opened 2024-01-11
fix: joining MLS conversation after login

typfel:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Already MLS conversations aren't always joined via external commit after login.

### Causes

1. The slow sync ends without having processed MLS conversation payload since that's done inside a Task
2. The conversation we should join is tracked in memory inside the MLSService, this is unreliable.

### Solutions

1. Wait for the Tasks to complete
2. Fetch conversation we need to join using a fetch request.

----

----

----
##### References

### DIFF
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -820,7 +820,12 @@ public final class MLSService: MLSServiceInterface {
             return
         }
 
-        let pendingGroups = await context.perform { self.fetchPendingGroups(in: context) }
+        let pendingGroups = try await context.perform {
+            try ZMConversation.fetchConversationsWithMLSGroupStatus(
+                mlsGroupStatus: .pendingJoin,
+                in: context
+            ).compactMap(\.mlsGroupID)
+        }
 
         logger.info("joining \(pendingGroups.count) group(s)")
 
@@ -835,15 +840,6 @@ public final class MLSService: MLSServiceInterface {
                 }
             }
         }
-    }
-
-    private func fetchPendingGroups(in context: NSManagedObjectContext) -> [MLSGroupID] {
-        logger.info("fetching list of groups pending join")
-
-        return ZMConversation.fetchConversationsWithMLSGroupStatus(
-            mlsGroupStatus: .pendingJoin,
-            in: context
-        ).compactMap(\.mlsGroupID)
     }
 
     // MARK: - Out-of-sync conversations
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -840,12 +840,10 @@ public final class MLSService: MLSServiceInterface {
     private func fetchPendingGroups(in context: NSManagedObjectContext) -> [MLSGroupID] {
         logger.info("fetching list of groups pending join")
 
-        let pendingConversations = ZMConversation.fetchConversationsWithMLSGroupStatus(
+        return ZMConversation.fetchConversationsWithMLSGroupStatus(
             mlsGroupStatus: .pendingJoin,
             in: context
-        )
-
-        return pendingConversations.compactMap(\.mlsGroupID)
+        ).compactMap(\.mlsGroupID)
     }
 
     // MARK: - Out-of-sync conversations
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -42,8 +42,6 @@ public protocol MLSServiceInterface: MLSEncryptionServiceInterface, MLSDecryptio
 
     func removeMembersFromConversation(with clientIds: [MLSClientID], for groupID: MLSGroupID) async throws
 
-    func registerPendingJoin(_ group: MLSGroupID)
-
     func performPendingJoins() async throws
 
     func wipeGroup(_ groupID: MLSGroupID) async
@@ -115,7 +113,6 @@ public final class MLSService: MLSServiceInterface {
     private let staleKeyMaterialDetector: StaleMLSKeyDetectorProtocol
     private let userDefaults: PrivateUserDefaults<Keys>
     private let logger = WireLogger.mls
-    private var groupsPendingJoin = Set<MLSGroupID>()
     private let groupsBeingRepaired = GroupsBeingRepaired()
     private let syncStatus: SyncStatusProtocol
 
@@ -814,12 +811,6 @@ public final class MLSService: MLSServiceInterface {
 
     typealias PendingJoin = (groupID: MLSGroupID, epoch: UInt64)
 
-    /// Registers a group to be joined via external commit once the app has finished processing events
-    /// - Parameter groupID: the identifier for the MLS group
-    public func registerPendingJoin(_ groupID: MLSGroupID) {
-        groupsPendingJoin.insert(groupID)
-    }
-
     /// Request to join groups still pending
     ///
     /// Generates a list of groups for which the `mlsStatus` is `pendingJoin`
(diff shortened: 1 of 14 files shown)


wire-ios-data-model/Source/Model/Conversation/ZMConversation+MLS.swift
```diff
@@ -159,6 +159,24 @@ public extension ZMConversation {
...
+        let matchingGroupStatus = NSPredicate(
+            format: "%K == \(mlsGroupStatus.rawValue)",
+            argumentArray: [Self.mlsStatusKey]
+        )
+
+        request.predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [
+            matchingGroupStatus, .isMLSConversation
+        ])
+
+        return context.executeFetchRequestOrAssert(request) as? [ZMConversation] ?? []
```
markusfassbender: Can we use replace this Objc-C version with assertions by Swift using `throws` here? `func fetch(_ request: NSFetchRequest<NSFetchRequestResult>) throws -> [Any]`

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
```diff
@@ -496,23 +496,23 @@ class ConversationByIDTranscoder: IdentifierObjectSyncTranscoder {
...
 
     func didReceive(response: ZMTransportResponse, for identifiers: Set<UUID>, completionHandler: @escaping () -> Void) {
-        defer { completionHandler() }
 
         guard response.result != .permanentError else {
             if response.httpStatus == 404 {
-                Task {
+                WaitingGroupTask(context: context) { [self] in
                     await deleteConversations(identifiers)
+                    await context.perform { completionHandler() }
```
markusfassbender: It looks wrong to me, that `completionHandler()` is called inside the `context.perform`. The handler should can manage itself, if core data operations are needed, or not?

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -828,32 +819,33 @@ public final class MLSService: MLSServiceInterface {
...
+        logger.info("fetching list of groups pending join")
 
-            return (groupID, conversation.epoch)
+        let pendingConversations = ZMConversation.fetchConversationsWithMLSGroupStatus(
+            mlsGroupStatus: .pendingJoin,
+            in: context
+        )
 
-        }
+        return pendingConversations.compactMap(\.mlsGroupID)
```
markusfassbender: One could write this also in a functional style like this, but I think it's personal preference 😉
```swift
return ZMConversation.fetchConversationsWithMLSGroupStatus(
    mlsGroupStatus: .pendingJoin,
    in: context
)
.compactMap(\.mlsGroupID)
```

markusfassbender (commented):
Not sure if I understand everything in this context, but it makes sense to me 👍

wire-ios-request-strategy/Sources/Payloads/Processing/ConversationEventPayloadProcessor.swift
```diff
@@ -687,15 +687,6 @@ final class ConversationEventPayloadProcessor {
...
         )
-
-        if source == .slowSync {
-            await context.perform {
-                mlsEventProcessor.joinMLSGroupWhenReady(
-                    forConversation: conversation,
-                    context: context
-                )
-            }
-        }
```
netbe, line 698: not needed anymore?

wire-ios-request-strategy/Sources/Request Strategies/MLS/MLSRequestStrategy.swift
```diff
@@ -49,6 +49,8 @@ public final class MLSRequestStrategy: AbstractRequestStrategy {
             withManagedObjectContext: managedObjectContext,
             applicationStatus: applicationStatus
         )
+
+        configuration = [.allowsRequestsDuringSlowSync, .allowsRequestsWhileOnline]
```
netbe, line 53: **question** : what was the default configuration before?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
netbe: I think it's to keep the expectation that is called on the syncContext. As the implementation of this method is in multiple places changing the behaviour could involve more work for now.

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -828,32 +819,33 @@ public final class MLSService: MLSServiceInterface {
...
-                logger.warn("conversation not found for group (\(groupID.safeForLoggingDescription))")
-                return nil
+        await withTaskGroup(of: Void.self) { group in
+            for pendingGroup in pendingGroups {
+                group.addTask {
+                    do {
+                        try await self.joinByExternalCommit(groupID: pendingGroup)
+                    } catch {
+                        WireLogger.mls.error("Failed to join pending group (\(pendingGroup): \(error)")
+                    }
```
netbe, line 839: **suggestion**: I guess the method does not throw then, we can remove the throws annotation.

What will happen with the failed join, it will be tried forever at each start up?

netbe (commented):
Just have some questions to fully understand the changes.

wire-ios-request-strategy/Sources/Payloads/Processing/ConversationEventPayloadProcessor.swift
typfel: The end result of this call was to add the conversation to the list of `pendingJoins` in the `MLSService`. I removed this property and now we're getting the pending joins via a fetch request instead.

wire-ios-request-strategy/Sources/Request Strategies/MLS/MLSRequestStrategy.swift
typfel: `[.allowsRequestsWhileOnline]` is the default, which wasn't allowing us complete the slow sync since we try to join/create the MLS self conversation during the slow sync.

This was "working" before since previously the slow sync wasn't waiting for this to complete previously.

wire-ios-data-model/Source/MLS/MLSService.swift
typfel: > suggestion: I guess the method does not throw then, we can remove the throws annotation.

👍 I'll remove it

> What will happen with the failed join, it will be tried forever at each start up?

Yep, after each quick sync even.

I want to refactor joining MLS conversation some more and probably break it out of the `MLSService`. Currently there's nothing stopping it from starting multiple "joining pending conversations" operations if two quick sync happens in quick succession.

I'll create ticket for this

wire-ios-data-model/Source/Model/Conversation/ZMConversation+MLS.swift
typfel: I'll try

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
typfel: Yeah, it's a bit weird but the handler currently expects the block to be called on the context. I can move the the perform into the handler, it would unnecessary in some cases but this isn't really a performance critical code path and it's easy to forget to to put the `completionHandler()` inside a perform.

wire-ios-data-model/Source/Model/Conversation/ZMConversation+MLS.swift
typfel: Done

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
typfel: I moved it the handler then we also only do this in one place.

wire-ios-data-model/Source/MLS/MLSService.swift
typfel: Done

wire-ios-data-model/Source/MLS/MLSService.swift
typfel: > > suggestion: I guess the method does not throw then, we can remove the throws annotation.
> 
> 👍 I'll remove it

Had to keep because the fetch request can throw now.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
markusfassbender: happy we could revert this! 🙌

netbe (approved):

markusfassbender (approved):

PR #850, opened 2024-01-11
chore: Integrate SwiftGen to Wire-Share-Extension target 

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

In order to continue with migrating from `.localized` to `L10n` we need to integrate SwiftGen to Wire Share Extension Target so we can generate the strings from the appropriate file into `L10n` struct. With this PR:

- Moved all `Localizable.strings` under `Resources` and now the Xcode project is in sync with the file directory
- Created a new `swiftgen.yml` for the share extension
- Added a `Run SwiftGen` build phase for the Wire Share Extension target
- Adjusted `run-swiftgen.sh` to accept a path, so the script can be executed for both targets individually.

----
##### References


scripts/run-swiftgen.sh
```diff
@@ -32,7 +34,25 @@ if [[ ! -f "$SWIFTGEN" ]]; then
     xcrun --sdk macosx swift package --package-path "$SCRIPTS_DIR" resolve
 fi
 
+# Run SwiftGen for main app
 (
     cd "$REPO_ROOT/wire-ios"
-    "$SWIFTGEN"
```
caldrian, line 37: I suggest to keep the script clean and independent of the actual config files, so we don't have to touch it for each new target that supports SwiftGen.

Therefore I would pass the script's arguments on to SwiftGen
```
"$SWIFTGEN" "$@"
```

and specify the config file in the Xcode run phases configuration, like we did for Sourcery:

https://github.com/wireapp/wire-ios/blob/e01287092627a9ea3f5ac7aa204515e7d575ab01/wire-ios-request-strategy/WireRequestStrategy.xcodeproj/project.pbxproj#L1912C7-L1912C7

caldrian (approved):

markusfassbender (approved):

scripts/run-swiftgen.sh
agisilaos: @caldrian Addressed it. Thank you!

PR #880, opened 2024-01-19
fix: failing tests on sync engine WPB-6196

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6196" title="WPB-6196" target="_blank">TaskWPB-6196</a>  [iOS] Fix failing sync engine tests
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Failing tests on sync engine needs fixing

The main reason for tests failing seems to be due to using `WaitingGroupTask` with a managed object context different from the context used by the tests. We wait for all groups to be empty but since we're using different context, the changes are not in sync. As a solution, we can save the context when we're done. 

There seems to be another issue related to Xcode and defining tests within extensions. Not sure what happens but it looks like some failing tests are not always being detected by Xcode.


caldrian (approved):

netbe:
> The main reason for tests failing seems to be due to using WaitingGroupTask with a managed object context different from the context used by the tests. 

wait the groups are not shared by contexts? it's one set of groups per context?

typfel (approved):

typfel:
> > The main reason for tests failing seems to be due to using WaitingGroupTask with a managed object context different from the context used by the tests.
> 
> wait the groups are not shared by contexts? it's one set of groups per context?

They are using the same managed object contexts.

From looking at the changes I was thinking it can be that the code was previously relying on implicit saves which are made after a http response has been processed. Now if during in the processing you do core data changes in a task it would happen after the implicit save.

wire-ios-sync-engine/Tests/Source/E2EE/ConversationTests+OTR.m
```diff
@@ -284,10 +284,7 @@ - (void)testThatItOTRAssetCanExpire
     WaitForAllGroupsToBeEmpty(0.5);
 }
 
-@end
-
 #pragma mark - Trust
-@implementation ConversationTestsOTR (Trust)
```
markusfassbender: Nice 👍

wire-ios-sync-engine/Source/Data Model/Conversation+MessageDestructionTimeout.swift
```diff
@@ -48,15 +48,21 @@ extension ZMConversation {
...
             return completion(.failure(WirelessLinkError.unknown))
         }
 
-        let request = MessageDestructionTimeoutRequestFactory.set(timeout: Int(timeout.rawValue), for: self, apiVersion: apiVersion)
+        let request = MessageDestructionTimeoutRequestFactory.set(
+            timeout: Int(timeout.rawValue),
+            for: self,
+            apiVersion: apiVersion
+        )
+
```
markusfassbender: Linter is complaining here:
> Lines should not have trailing whitespace (trailing_whitespace)

markusfassbender (approved):

netbe (approved):

PR #902, opened 2024-01-26
fix: create mls group no keypackages - WPB-6082

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6082" title="WPB-6082" target="_blank">BugWPB-6082</a>  [iOS] MLS: can't create MLS group with no participants
  </td></table>
  <br />
 

# What's new in this PR?

Changes

* Fixes not retying uploading keypackages if it fails
* Sending commit via updateKeyMaterial in case we're adding only the selfUser (with no packages) to the MLS group.
* Cleanup logs

#### Test Coverage (Optional)

### Notes (Optional)

* Still potentially we don't know why the upload to keypackages is sometimes failing


wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -513,12 +513,13 @@ public final class MLSService: MLSServiceInterface {
             guard !users.isEmpty else { throw MLSAddMembersError.noMembersToAdd }
             let keyPackages = try await claimKeyPackages(for: users)
 
-            guard keyPackages.count > 0 else {
-                throw MLSAddMembersError.noInviteesToAdd
+            let events = if keyPackages.isEmpty {
+                try await mlsActionExecutor.updateKeyMaterial(for: groupID)
```
johnxnguyen, line 520: suggestion: add a comment here explaining why we update the key material.

wire-ios-request-strategy/Sources/Request Strategies/MLS/Actions/ClaimMLSKeyPackageActionHandler.swift
```diff
@@ -68,8 +68,9 @@ class ClaimMLSKeyPackageActionHandler: ActionHandler<ClaimMLSKeyPackageAction> {
             let keyPackagesExcludingSelfClient = payload.keyPackages.filter {
                 $0.client != action.excludedSelfClientId
             }
-
-            guard keyPackagesExcludingSelfClient.isNonEmpty else {
+            let selfUser = ZMUser.selfUser(in: context)
+            let isSelfUserRequesting = selfUser.remoteIdentifier == action.userId && selfUser.domain == action.domain
+            guard isSelfUserRequesting || keyPackagesExcludingSelfClient.isNonEmpty else {
```
johnxnguyen, line 73: Why if we're requesting key packages for the self user and none are found, then we consider it a success?

netbe: @typfel

typfel: Because the self user not having key packages is not an error.

For other users claiming key packages is successful if there's at least one MLS client with key packages available. This isn't true when claiming key packages for the self user, it would throw `.emptyKeyPackages` unless the self user had at least two MLS clients since we exclude the self client when claiming key packages.

We can deal with it in two ways:
1. Allow returning an empty list of key packages for the self user
2. Catch `.emptyKeyPackages` and check if it was for the self user and then handle the special case

I don't have a strong opinion on how we do it.

wire-ios-data-model/Source/MLS/MLSService.swift
netbe: done

typfel (approved):

wire-ios-request-strategy/Sources/Request Strategies/MLS/Actions/ClaimMLSKeyPackageActionHandler.swift
johnxnguyen: Right, so it's for the case when the self user with a single mls client creates a group with no one else.

johnxnguyen (approved):

PR #916, opened 2024-01-31
refactor: make UserStatusView a passive view [WPB-765]

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-765" title="WPB-765" target="_blank">StoryWPB-765</a>  [iOS] 5.3.16 Indicate verified user
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The `UserStatusView` is subscribing to two types of notifications and updating itself. However, soon we want to display the user's verification status which will require to call async code.

### Solutions

This PR introduces a model `UserStatus` which contains the information to display on the `UserStatusView`.
The view receives a public property and only updates, when the property is set.
Any logic to update the view based on external changes is moved out of the view.

#### How to Test

Snapshot tests are updated.

----

----

----
##### References


wire-ios-data-model/Source/Notifications/ObjectObserverTokens/Helpers/ClosureBasedUserChangeObserver.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// Allows for handling user change events via closure.
+@objc(ZMClosureBasedUserChangeObserver)
+public final class ClosureBasedUserChangeObserver: NSObject, UserObserver {
```
markusfassbender: I started here with the question: why does it need to inherit from `NSObject`? Can't `ClosureBasedUserChangeObserver` be a struct?

- The `protocol UserObserver: NSObjectProtocol` requires it, but is it really necessary?
- I replaced it with `protocol UserObserver: AnyObject` and the compiler doesn't complain.
- Then I removed it like `protocol UserObserver` and the compiler doesn't complain 

From my understanding the inheritance from `NSObjectProtocol` and `AnyObject` is required when the code requires a class implementing it and sometimes it's important to assume so. In this particular case I think it's not necessary and we can remove the constraint.

caldrian: I added it because the protocol required it, but you're right, we can just remove Obj-C support: 55ed881a7eb24642f8dee154f63aa63a56766770

markusfassbender: Oh I see because of `@objc` we need to have a class conformance. Thanks!

wire-ios-data-model/Source/Notifications/ObjectObserverTokens/Helpers/ClosureBasedUserChangeObserver.swift
```diff
@@ -0,0 +1,31 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+/// Allows for handling user change events via closure.
+public final class ClosureBasedUserChangeObserver: UserObserver {
```
theswiftway: **[suggestion(blocking)]:** This could confirm to `ZMUserObserver` instead of the concrete class `UserObserver`.

theswiftway: Or could have a nil property closure on existing `UserObserver` instead of a separate class doing the same thing. Consider this like the `URLSession` API where you could use closure way or async way
`var userDidChange:((UserChangeInfo) -> Void)?` add assign it where you need to observe it.

caldrian: Okay I deleted this class.

wire-ios/Wire-iOS Tests/UserStatusViewSnapshotTests.swift
```diff
@@ -130,13 +130,19 @@ final class UserStatusViewSnapshotTests: ZMSnapshotTestCase {
         updateAvailability(for: user, newValue: availability)
 
         let sut = UserStatusView(
-            user: user,
             options: options,
             userSession: userSession
         )
         sut.overrideUserInterfaceStyle = userInterfaceStyle
-        sut.backgroundColor = userInterfaceStyle == .dark ? .black : .white
+        sut.backgroundColor = .systemBackground
```
theswiftway, line 137: I see snapshots for only light mode here. May be we should do the test for dark mode as well?

wire-ios/Wire-iOS/Sources/UserInterface/Availability/UserStatusViewController.swift
```diff
@@ -88,4 +90,41 @@ final class AvailabilityTitleViewController: UIViewController {
...
+            object: nil
+        )
+
+        // refresh view when some user changes
+        userChangeObservation = userSession.addUserObserver(self, for: user)
+    }
+
+    @objc
+    private func updateUserStatusView() {
+        let view = view as! UserStatusView
```
markusfassbender: I know this struggle... what one can do create a `var userStatusView: UIView!` (next to `view: UIView!`)and assign the view to both vars. 
Or implement it like this `var userStatusView: UIView { view as! UserStatusView }`.
Even though you have a force cast, it's quite clear that it happens only in one place in the whole view controller.

markusfassbender (approved):

theswiftway (approved):

wire-ios/Wire-iOS Tests/UserStatusViewSnapshotTests.swift
caldrian: Sorry, the [updated reference image](https://github.com/wireapp/wire-ios/blob/090f9159f8adfdb6fa4e7569525cf2799538bc51/wire-ios/Wire-iOS%20Tests/ReferenceImages/ConversationCellBurstTimestampViewSnapshotTests/testForIncludeDayOfWeekAndDot.1.png) doesn't actually belong to this test.

This test (`UserStatusViewSnapshotTests`) does have dark and light reference images: https://github.com/wireapp/wire-ios/tree/develop/wire-ios/Wire-iOS%20Tests/ReferenceImages/UserStatusViewSnapshotTests

PR #844, opened 2024-01-10
chore: Get rid of .localized and replace it with L10n.Localizable struct - P7

agisilaos:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [ ] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [ ] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [ ] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

This PR is following #713, #730, #732, #735, #754, #783, #803, #815

This PR addresses the task of standardizing localization practices across the codebase. We have replaced all instances of the .localized method with the newly generated L10n.Localizable struct, which provides type-safe access to localized strings.

## Changes

Updated localization calls from .localized to `L10n.Localizable` to ensure consistency and safety.
Enhanced the maintainability of localization by utilizing `SwiftGen` to generate `L10n.Localizable`, reducing the risk of typos and simplifying the process of updating localized strings.

**Note**: We're down to 87 results in 30 files

## Affected files

- `Settings+ColorScheme.swift`
- `SettingsPropertyFactory.swift`
- `UIAlertController+DegradedCall.swift`
- `BitRateLabel.swift`
- `UIAlertController+CompanyLogin.swift`
- `ConversationButtonMessageCell.swift`
- `ParticipantsStringFormatter.swift`
- `ConversationTitleView.swift`
- `CameraKeyboardPermissionsCell.swift`
- `ConversationInputBarViewController+Camera.swift`
- `ConversationInputBarViewController+DragAndDrop.swift`
- `ConversationInputBarViewController.swift`
- `ConversationListTopBarViewController.swift`
- `ConversationAction.swift`
- `LocationSelectionViewController.swift`
- `ProfileHeaderViewController.swift`
- `SelfProfileViewController+NewDeviceAlert.swift`
- `SettingsClientViewController.swift`
- `SettingsTableViewController.swift`

----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #923, opened 2024-02-01
fix: Localized String in NewLoginAlert we show in SelfProfile 

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

In #844 I did a mistake and passed the wrong parameter in the conctruction of a localized string which led to show the wrong text inside an alert. While I figured that out I did some more testing and realized that even before that change we were showing the text but in the wrong format. 

Screenshot 2024-02-01 at 12 15 43

With this PR I fix both of those mistakes and add some unit tests. 

----
##### References


caldrian (approved):

wire-ios/Wire-iOS Tests/SelfProfileViewControllerTests.swift
agisilaos, line 1: I know that in this file we have both Unit and SnapshotTests. I would like to separate those in a different PR.

wire-ios-data-model/Source/Model/UserClient/UserClientType.swift
```diff
@@ -37,6 +37,8 @@ import Foundation
     /// True if the self user has verfied the client
     var verified: Bool { get }
 
+    var needsToNotifyUser: Bool { get set }
+
```
markusfassbender, line 41: question: I don't see the variable `needsToNotifyUser` used in the changes, why was it added?

agisilaos: Good question! When I changed the type from UserClient to UserClientType I would get an error in this method:

```swift
fileprivate func presentNewLoginAlertController(_ clients: [UserClientType]) {
        let newLoginAlertController = UIAlertController(forNewSelfClients: clients)

        let actionManageDevices = UIAlertAction(title: L10n.Localizable.Self.NewDeviceAlert.manageDevices, style: .default) { _ in
            self.openControllerForCellWithIdentifier(SettingsCellDescriptorFactory.settingsDevicesCellIdentifier)
        }

        newLoginAlertController.addAction(actionManageDevices)

        let actionTrustDevices = UIAlertAction(title: L10n.Localizable.Self.NewDeviceAlert.trustDevices, style: .default) { [weak self] _ in
            self?.presentUserSettingChangeControllerIfNeeded()
        }

        newLoginAlertController.addAction(actionTrustDevices)

        present(newLoginAlertController, animated: true, completion: .none)

        userSession.enqueue {
            clients.forEach {
                $0.needsToNotifyUser = false
            }
        }
    }
```

More specifically in line:

**$0.needsToNotifyUser = false**

needsToNotifyUser wasn't part of UserClientType and I needed to add it.

markusfassbender: thanks!

markusfassbender (approved):

PR #934, opened 2024-02-05
feat: establish mls team 1-1 WPB-2176

johnxnguyen:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR adds support for establishing 1-1 mls conversations with team members (from the same team as the self user). Changes include:

- Introduce `CreateTeamOneOnOneConversationUseCase` which is responsible for creating the team one on one depending on the supported protocols
- Invoke the use case from the user session, taking care of invoking the use case on a background thread and reporting the result on the main thread
- Using the new use case when selecting a user from the search ui or opening the conversation from their user profile
- adjust predicate for 1-1 conversations
- refactor predicates out into `ConversationPredicateFactory` to make it easier to generate predicates with input (in this case, self team)

#### Test Coverage

- Updated existing tests for predicates
- Unit tests for new use case

#### How to Test

1. User A and B belong to Team T.
2. User A does not have an existing conversation with B.
3. User A searches for B and selects them.
4. Assert that a conversation was created on both sides, and the protocol is mls if both sides support mls, else proteus.

### Notes

The conversation predicates are quite hard to understand. I'd like to attempt to clarify them and clean them up a bit if possible, but I'd do that in another PR.

----

----

----
##### References


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.swift
```diff
@@ -369,10 +369,10 @@ final class ConversationViewController: UIViewController {
     // MARK: Resolve 1-1 conversations
 
     private func resolveConversationIfOneOnOne() {
-        let isOneOnOneConversation = conversation.conversationType == .oneOnOne
-            || conversation.conversationType == .group && (conversation.localParticipants.count == 2)
-
-        guard isOneOnOneConversation && conversation.messageProtocol == .proteus else {
```
johnxnguyen, line 375: If this is a fake one on one, then the conversation type will be computed as `.oneOnOne` by a custom accessor, even if the underlying value is `.group`.

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Predicates.swift
```diff
@@ -77,102 +77,124 @@ extension ZMConversation {
...
+extension String {
 
-    @objc(predicateForGroupConversations)
-    class func predicateForGroupConversations() -> NSPredicate {
-        let groupConversationPredicate = NSPredicate(format: "\(ZMConversationConversationTypeKey) == \(ZMConversationType.group.rawValue)")
-        let notInFolderPredicate = NSCompoundPredicate(notPredicateWithSubpredicate: predicateForConversationsInFolders())
+    var words: [String] {
+        var words: [String] = []
+        enumerateSubstrings(in: self.startIndex..., options: .byWords) { substring, _, _, _ in
+            words.append(String(substring!))
```
markusfassbender: I've read in the doc that the string is non-optional if we don't pass a certain option. Though to avoid force-cast I would suggest to use:
```swift
guard let substring else {
    assertionFailure("expected substring value, since we skip option '.substringNotRequired'!")
}
```

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Predicates.swift
```diff
@@ -77,102 +77,124 @@ extension ZMConversation {
...
 
-        return NSCompoundPredicate(andPredicateWithSubpredicates: [notClearedTimestamp, predicateForValidConversations()])
-    }
+extension String {
 
-    @objc(predicateForGroupConversations)
-    class func predicateForGroupConversations() -> NSPredicate {
-        let groupConversationPredicate = NSPredicate(format: "\(ZMConversationConversationTypeKey) == \(ZMConversationType.group.rawValue)")
-        let notInFolderPredicate = NSCompoundPredicate(notPredicateWithSubpredicate: predicateForConversationsInFolders())
+    var words: [String] {
```
markusfassbender: Is it worth to create an extension on string, that is available in all places for all strings? Or can it be a normal `func extractWords(from string: String)` in this class?

If `.words` should be extension on `String`, it would be better searchable and also easier to move to other modules in future as dedicated file.

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Predicates.swift
```diff
@@ -77,102 +77,124 @@ extension ZMConversation {
...
+
+        return words
     }
+}
 
-    @objc(predicateForLabeledConversations:)
-    class func predicateForLabeledConversations(_ label: Label) -> NSPredicate {
-        let labelPredicate = NSPredicate(format: "%@ IN \(ZMConversationLabelsKey)", label)
+@objc
+public final class ConversationPredicateFactory: NSObject {
```
markusfassbender: Please let's move this factory to a dedicated file as it is better searchable and also easier to move to other modules in future.

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Predicates.swift
```diff
@@ -77,102 +77,124 @@ extension ZMConversation {
...
 
-        return NSCompoundPredicate(andPredicateWithSubpredicates: [predicateForConversationsExcludingArchived(), labelPredicate])
+    private let selfTeam: Team?
+
+    @objc
+    init(selfTeam: Team? = nil) {
```
markusfassbender: I ask myself why we need retain team as object here. Is it required for all funcs or only certain ones? In my opinion a factory should not depend on any state but provide pure functions.

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Predicates.swift
```diff
@@ -77,102 +77,124 @@ extension ZMConversation {
...
 
-    private class func predicateForValidConversations() -> NSPredicate {
-        let basePredicate = predicateForFilteringResults()
+    private func predicateForValidConversations() -> NSPredicate {
+        let basePredicate = ZMConversation.predicateForFilteringResults()
         return .all(of: [basePredicate, isProtocolReady(), isValidConversation()])
     }
 
-    private class func isProtocolReady() -> NSPredicate {
+    private func isProtocolReady() -> NSPredicate {
```
markusfassbender: Nice that it's not `class func` anymore 👍

wire-ios-data-model/Support/Sources/ModelHelper.swift
```diff
@@ -0,0 +1,129 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireDataModel
+
+/// A helper object to make it easy to create and relate model objects.
+
+public struct ModelHelper {
```
markusfassbender, line 24: Nice extraction, thanks!

wire-ios-data-model/Tests/Source/Model/ConversationList/ZMConversationListDirectoryTests.m
```diff
@@ -65,9 +65,20 @@ - (void)setUp
     Team *team = [Team insertNewObjectInManagedObjectContext:self.uiMOC];
     team.remoteIdentifier = [NSUUID createUUID];
 
+    ZMUser *selfUser = [ZMUser selfUserInContext:self.uiMOC];
+    selfUser.remoteIdentifier = [NSUUID createUUID];
```
markusfassbender: Do we need to use `createUUID` or could it be `[NSUUID new]` aka `[[NSUUID alloc] init]`?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationService.swift
```diff
@@ -32,6 +32,11 @@ public protocol ConversationServiceInterface {
         completion: @escaping (Result<ZMConversation, ConversationCreationFailure>) -> Void
     )
 
+    func createFakeOneOnOneProteusConversation(
```
markusfassbender: Fake sounds for me like a mock or stub... does it make sense to choose a slightly different word? Maybe something like `Artificial` or `Imitating`.. native speakers welcome 🙃

wire-ios-sync-engine/Source/Use cases/CreateTeamOneOnOneConversationUseCase.swift
```diff
@@ -0,0 +1,154 @@
...
+    case failedToCreateProteusConversation(Error)
+    case noCommonProtocols
+    case conversationNotFound
+
+}
+
+/// Creates the team one on one conversation with a particlar user, depending
+/// on the currently supported protocols of the self and other users.
+
+class CreateTeamOneOnOneConversationUseCase: CreateTeamOneOnOneConversationUseCaseProtocol {
```
markusfassbender: Would it be possible to use a `struct`?

wire-ios-sync-engine/Source/UserSession/ZMUserSession+OneOnOne.swift
```diff
@@ -0,0 +1,76 @@
...
+                return
+            }
+
+            let useCase = CreateTeamOneOnOneConversationUseCase(
+                protocolSelector: OneOnOneProtocolSelector(),
+                migrator: self.syncContext.mlsService.map(OneOnOneMigrator.init),
+                service: ConversationService(context: self.syncContext)
+            )
+
+            Task {
```
markusfassbender: I think we discussed this before.. is it a good idea to run swift concurrency inside a `context.perform` block? it would be better to fetch required values and later in order run the `Task`.

wire-ios-sync-engine/Tests/Source/Use cases/CreateTeamOneOnOneConversationUseCaseTests.swift
```diff
@@ -0,0 +1,177 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+import WireDataModelSupport
+import WireRequestStrategySupport
+import WireSyncEngineSupport
+@testable import WireSyncEngine
+
+final class CreateTeamOneOnOneConversationUseCaseTests: XCTestCase {
```
markusfassbender, line 25: Grea, thank you for avoid class inheritance here 😍

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel+StartUIDelegate.swift
```diff
@@ -45,14 +45,22 @@ extension ConversationListViewController.ViewModel: StartUIDelegate {
...
+        guard let userSession = ZMUserSession.shared() else {
+            return
+        }
+
+        if let conversation = user.oneToOneConversation {
+            onConversationCreated(.success(conversation))
+        } else {
+            viewController?.setState(.conversationList, animated: true) {
+                userSession.createTeamOneOnOne(with: user) {
+                    switch $0 {
```
markusfassbender: For reading it's easier to keep the same level of abstraction like line 53, so I would suggested to extract a func and call it in line 55.

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Predicates.swift
johnxnguyen: Moved to a new file.

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Predicates.swift
johnxnguyen: In this PR I just moved existing code into a the factory, refraining from refactoring also code that was moved. I'd like to defer that to another PR.

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Predicates.swift
johnxnguyen: In this PR I just moved existing code into a the factory, refraining from refactoring also code that was moved. I'd like to defer that to another PR.

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Predicates.swift
johnxnguyen: It's used in pretty much all of the predicate functions because it's used in one of the lowest level predicates (the one that filters out valid one on one conversations). I chose to retain it here to avoid injected it in every function of the factory, but I'm happy to reconsider. I would though defer that for a follow up PR that could simplify these predicates.

wire-ios-data-model/Tests/Source/Model/ConversationList/ZMConversationListDirectoryTests.m
johnxnguyen: I was aiming to be consistent, but I suppose it wouldn't hurt to avoid the create function.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationService.swift
johnxnguyen: I can also just take "fake" out and leave a comment.

johnxnguyen: done

wire-ios-sync-engine/Source/Use cases/CreateTeamOneOnOneConversationUseCase.swift
johnxnguyen: done

wire-ios-sync-engine/Source/UserSession/ZMUserSession+OneOnOne.swift
johnxnguyen: Right, I forgot :) I moved the sync block inside the task.

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel+StartUIDelegate.swift
johnxnguyen: done

wire-ios-data-model/Tests/Source/Model/ConversationList/ZMConversationListDirectoryTests.m
johnxnguyen: done

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationService.swift
markusfassbender: That works for me, thanks!

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Predicates.swift
markusfassbender: Okay fine

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Predicates.swift
markusfassbender: I understand, it's okay for me

markusfassbender (approved):

wire-ios-sync-engine/Source/Use cases/CreateTeamOneOnOneConversationUseCase.swift
```diff
@@ -0,0 +1,154 @@
...
+    case missingUserQualifiedID
+    case mlsMigratorNotFound
+    case failedToCreateMLSConversation(Error)
+    case failedToCreateProteusConversation(Error)
+    case noCommonProtocols
+    case conversationNotFound
+
+}
+
+/// Creates the team one on one conversation with a particlar user, depending
```
netbe:
```suggestion
/// Creates the team one on one conversation with a particular user, depending
```

wire-ios-sync-engine/Source/Use cases/CreateTeamOneOnOneConversationUseCase.swift
```diff
@@ -0,0 +1,154 @@
...
+        service: ConversationServiceInterface
+    ) {
+        self.protocolSelector = protocolSelector
+        self.migrator = migrator
+        self.service = service
+    }
+
+    func invoke(
+        with user: ZMUser,
+        syncContext: NSManagedObjectContext
```
netbe, line 66: **suggestion**: what about passing syncContext in the init instead?

wire-ios-sync-engine/Source/Use cases/CreateTeamOneOnOneConversationUseCase.swift
```diff
@@ -0,0 +1,154 @@
...
+            )
+
+        case .proteus:
+            return try await createProteusConversation(
+                with: user,
+                in: syncContext
+            )
+
+        case .none, .mixed:
+            throw Error.noCommonProtocols
```
netbe, line 103: **question**: what does this mean noCommonProtocols, are you supposed to have both when mixed?

why does `getProtocolForUser` return nil?

wire-ios-sync-engine/Source/UserSession/ZMUserSession+OneOnOne.swift
```diff
@@ -0,0 +1,75 @@
...
+    ///
+    /// - Parameters:
+    ///   - user: the other user.
+    ///   - completion: a result handler.
+
+    public func createTeamOneOnOne(
+        with user: UserType,
+        completion: @escaping (Swift.Result<ZMConversation, CreateTeamOneOnOneConversationError>) -> Void
+    ) {
+        guard let viewUser = user.materialize(in: viewContext) else {
```
netbe, line 33: I guess we can't materialize directly in syncContext, right?

netbe (approved):
LGTM;) just some suggestions

wire-ios-sync-engine/Source/UserSession/ZMUserSession+OneOnOne.swift
johnxnguyen: No, there's an assert that it can only be done in the view context.

wire-ios-sync-engine/Source/Use cases/CreateTeamOneOnOneConversationUseCase.swift
johnxnguyen: If there are no common protocols it will return nil. The "mixed" protocol doesn't really make sense here, I think it happens now because previously there were two distinct message protocol types and they got merged into one. One of them was for the conversation protocol, the other was for which protocols a user supports.

wire-ios-sync-engine/Source/Use cases/CreateTeamOneOnOneConversationUseCase.swift
johnxnguyen: I passed it in the function to avoid having state. Do you see a benefit to retaining it in init?

netbe: only advantage would be not to pass it to all methods, but I don't have a strong objection to change this, just observation.

wire-ios-sync-engine/Source/Use cases/CreateTeamOneOnOneConversationUseCase.swift
netbe: ok thks for the explanation

PR #955, opened 2024-02-07
fix: Replace `verifyInAllPhoneWidths(view: UIView` with `verifyInAllPhoneWidths(matching: UIView` 

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we get rid of the method that was being used by the old deprecated snapshot library and replace it using the method we currently would like to use where it belongs to the framework that we would like to keep. 

----
##### References


markusfassbender (approved):
Thanks! Looks good to me, but I did not compare all snapshots manually.

johnxnguyen (approved):

PR #986, opened 2024-02-14
fix: video toggle disabled in MLS calls - WPB-6204

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6204" title="WPB-6204" target="_blank">BugWPB-6204</a>  [iOS] MLS: Cannot turn on camera when joining video call on bund-next
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

On a MLS conference calls, sometimes the video button is disabled

### Causes (Optional)

When receiving the call, we set the wrong value of `isConferenceCall`  to the callSnapshot. We check only for conference value and not mlsConference. This results on caller being able to enable the video but not receivers.

### Solutions

Use helper method from AVSIdentiferType to get the right value of `isConferenceCall`

#### Test Coverage (Optional)

#### How to Test

- MLS Call with multiple people
- answer on iphone
- you should be able to enable video


caldrian (approved):

wire-ios-sync-engine/Source/Calling/CallCenterSupport.swift
```diff
@@ -43,6 +43,19 @@ public enum AVSConversationType: Int32 {
 
 }
 
+extension AVSConversationType {
```
markusfassbender: Suggestion: move this extension to a file `AVSConversationType+Conference.swift`. It makes it easier to discover in the project and the files smaller.

markusfassbender (approved):

theswiftway (approved):
LGTM

theswiftway:
**question:** Not related but I saw that `AVSConversationType` & `AVSCallType` are explicitly numbered. They would be numbered the same even without that. Is it for any particular reason?

netbe:
> **question:** Not related but I saw that `AVSConversationType` & `AVSCallType` are explicitly numbered. They would be numbered the same even without that. Is it for any particular reason?

@theswiftway yes you're right, maybe at the time they just ported to objc without having this in mind

PR #990, opened 2024-02-15
fix: When dark mode (or vice versa) is toggled while on conversation view, not all buttons in Input Bar adapt their colors - WPB-5671

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5671" title="WPB-5671" target="_blank">BugWPB-5671</a>  [iOS] When dark mode is toggled while on conversation view, not all buttons switch mode
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we fix an issue with the input bar and the colors not being updating when switching from light to dark mode and vice versa.

### Issues

Not all buttons in the input bar would update their colors when switching from light to dark mode and vice versa.

### Solution

The reason removing resolvedColor(with: traitCollection) fixed the issue with certain buttons not updating their color in response to changes between light and dark modes lies in how UI elements in iOS adapt to appearance changes. When you use resolvedColor(with: traitCollection), you are explicitly resolving a color based on the current traitCollection at the time of execution. This means that the color is determined once based on the current mode (light or dark) and the traitCollection specifics at that moment. Once resolved, this color does not automatically update when the user interface style changes (e.g., from light to dark mode or vice versa) because it has been fixed to a specific appearance.

The traitCollection contains information about the environmental and interface attributes of a UI component, including the user interface style (light or dark mode). When you resolve a color using resolvedColor(with: traitCollection), it becomes tied to the traitCollection it was resolved with. If the system's appearance changes after this resolution (the user switches from light to dark mode), the resolved color remains unchanged because it does not re-evaluate against a new traitCollection reflecting the updated interface style.

By removing the call to resolvedColor(with: traitCollection), we allow the color to remain dynamic or "unresolved." UIKit then automatically resolves this color at the point of use, taking into account the current traitCollection at that time, which includes the current user interface style. This means that every time the system's appearance changes, UIKit will resolve the dynamic color according to the new traitCollection, ensuring that the color reflects the current mode correctly.

--

**Note** In the past we implemented this resolving as a solution to this exact problem. My guess is that something may have changed between iOS versions so that wouldn't be needed as a workaround. 

| BEFORE | AFTER |
|---|---|
| Simulator Screen Recording - iPhone 15 - 2024-02-15 at 14 44 43 | Simulator Screen Recording - iPhone 15 - 2024-02-15 at 14 47 49 |

----

##### References


netbe:
> Note In the past we implemented this resolving as a solution to this exact problem.
@agisilaos so is there a risk of regression somewhere else ?

johnxnguyen (commented):
I thought we did this because the color can't be dynamic because it gets converted to a cgcolor or it's drawn as an image of something like that. In these cases I thought we listen to trait changes and redraw the UI elements that can't support dynamic colors, using the `resolvedColor` method?

agisilaos:
> > Note In the past we implemented this resolving as a solution to this exact problem.
> > @agisilaos so is there a risk of regression somewhere else ?

I manually checked other places the input bar and I didn't see any regressions.

agisilaos:
> I thought we did this because the color can't be dynamic because it gets converted to a cgcolor or it's drawn as an image of something like that. In these cases I thought we listen to trait changes and redraw the UI elements that can't support dynamic colors, using the `resolvedColor` method?

Yes but as it turns out that's not working anymore. Something I found when moving to xcode 15 I think is that things like overrideUserInterfaceStyle started only affecting the next draw, rather than taking affect immediately.

wire-ios/Wire-iOS/Sources/Components/Buttons/IconButton.swift
```diff
@@ -182,7 +182,6 @@ class IconButton: ButtonWithLargerHitArea {
 
     func setBackgroundImageColor(_ color: UIColor,
                                  for state: UIControl.State) {
-        let color = color.resolvedColor(with: traitCollection)
```
johnxnguyen, line 185: So is this method invoked automatically when the theme changes?

agisilaos: @johnxnguyen Exactly.

agisilaos: I've also tested it with an iOS 15, 16 and 17 simulator and had no issues!

johnxnguyen: I'm curious where it's being invoked from? Perhaps we can look at the debugger together?

johnxnguyen (approved):

markusfassbender (approved):

PR #991, opened 2024-02-15
fix: request loop team - WPB-6724

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6724" title="WPB-6724" target="_blank">BugWPB-6724</a>  request loop on /teams/XXX/members
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

During the playtest, we inserted a duplicate of team and after the migration, the slowsync was looping on the fetching team members phase.

### Causes (Optional)

Deleting the duplicate team deleted the members (which makes the link to User) in cascade and nulling the team relationship on User (here the selfUser).
The TeamDownloadMembers strategy depends on the selfuser team to update the members, as it was nil it was not able to go on.

Note that in the duplicate Team action we moved the members which is in reality not a likely scenario. Also having multiple teams is possible in theory in reality we always have only one.

### Solutions

* Update the method that update or get the team for a user to update properties at the same time. See `getOrUpdateMember(for:in:context:)`
* Report request loops in Datadog

#### How to Test

* Restore backup with team duplicate
-> it should not have a request loop

### DIFF
--- a/wire-ios-sync-engine/Source/Synchronization/Strategies/TeamMembersDownloadRequestStrategy.swift
+++ b/wire-ios-sync-engine/Source/Synchronization/Strategies/TeamMembersDownloadRequestStrategy.swift
@@ -63,7 +63,7 @@ public final class TeamMembersDownloadRequestStrategy: AbstractRequestStrategy, 
             let rawData = response.rawData,
             let payload = MembershipListPayload(rawData)
         else {
-            syncStatus.failCurrentSyncPhase(phase: .fetchingTeamMembers)
+            failSyncPhase()
             return
         }
 
@@ -76,8 +76,11 @@ public final class TeamMembersDownloadRequestStrategy: AbstractRequestStrategy, 
         completeSyncPhase()
     }
 
+    func failSyncPhase() {
+        syncStatus.failCurrentSyncPhase(phase: .fetchingTeamMembers)
+    }
+
     func completeSyncPhase() {
         syncStatus.finishCurrentSyncPhase(phase: .fetchingTeamMembers)
     }
-
 }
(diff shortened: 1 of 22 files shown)


agisilaos (approved):

caldrian (approved):

wire-ios-sync-engine/Source/Synchronization/Strategies/TeamMembersDownloadRequestStrategy.swift
```diff
@@ -63,6 +63,7 @@ public final class TeamMembersDownloadRequestStrategy: AbstractRequestStrategy,
             let rawData = response.rawData,
             let payload = MembershipListPayload(rawData)
         else {
+            syncStatus.failCurrentSyncPhase(phase: .fetchingTeamMembers)
```
markusfassbender: The success case happens with a `func` on a different abstraction level, see screenshot. I would suggest to rather move both `syncStatus` calls in funcs or call directly to make it easier to read.

Screenshot 2024-02-19 at 15 42 22

wire-ios-transport/Source/TransportSession/ZMTransportSession.m
```diff
@@ -238,15 +238,21 @@ - (instancetype)initWithURLSessionsDirectory:(id<URLSessionsDirectory, TearDownC
                                                                           group:group
                                                                         backoff:nil
                                                              initialAccessToken:initialAccessToken];
+
+        self.remoteMonitoring = [[RemoteMonitoring alloc] initWithLevel: LevelInfo];
```
markusfassbender:
```suggestion
        self.remoteMonitoring = [[RemoteMonitoring alloc] initWithLevel:LevelInfo];
```

markusfassbender (approved):
Great finding, thanks!

wire-ios-sync-engine/Source/Synchronization/Strategies/TeamMembersDownloadRequestStrategy.swift
netbe: ok yes, note that for the is phase failCurrentSyncPhase does nothing

PR #1014, opened 2024-02-21
fix: use external senders of parent group when creating subgroup WPB-6665

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6665" title="WPB-6665" target="_blank">BugWPB-6665</a>  [iOS] Use parent conversation `external_sender` in MLS subconversations
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
We are not able to decrypt external proposals in sub conversations belonging to other domains.

### Causes
Unlike normal mls conversations, subconversations (e.g for mls conferences) can be created by any participants of the parent conversation, even if they don't belong to the owning domain (otherwise only some people could start conference calls). 

Consider users A (domain A) and B (domain B). 
- B creates an mls group (belonging to domain B) 
- A starts a conference call, therefore A creates the subconversation. A erroneously passes in the public keys of local backend (domain A) when creating the subconversation in Core Crypto.
- The call works fine
- B leaves the call, leaving the subconversation.
- Domain B sends a remove proposal to A, signed with the public keys of domain B.
- A can't decrypt it because the public keys don't match.

The cause of the problem is that we don't pass in the public keys of the backend owning the parent conversation.

### Solutions

When creating a subconversation, use the keys from the parent conversation. When creating a normal parent conversation, use the keys from the local domain.

#### Test Coverage

- Updated existing unit tests

#### How to Test

- User A from domain A and user B from domain B
- A creates an mls group with B
- B starts an mls conference
- A leaves
- B sees A leaves successfully.

----

----

----
##### References


wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -129,7 +129,6 @@ public final class MLSService: MLSServiceInterface {
         case keyPackageQueriedTime
     }
 
-    var backendPublicKeys = BackendMLSPublicKeys()
```
johnxnguyen, line 132: There's no point storing this as we fetch it every time we create a group anyway.

typfel (approved):

wire-ios-data-model/Source/MLS/MLSService.swift
netbe: I think it was to have an optimization, since we shouldn't need to fetch everytime. But we can do that in a second time if still relevant cc @typfel

netbe (approved):
LGTM;) just wondering if we would be still an optim to fetch the backendKeys only once

PR #944, opened 2024-02-06
feat: [epic] evaluate all 1-1 conversations [WPB-5812]

markusfassbender:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5812" title="WPB-5812" target="_blank">Sub-taskWPB-5812</a>  [iOS] Implement the logic for evaluating all 1:1 conversations
  </td></table>
  <br />
 

----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [x] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [x] contains a reference JIRA issue number like `SQPIT-764`
  - [x] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [x] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

💡 this is an epic branch to collect all changes from sub pull requests:
- [WPB-5812](https://wearezeta.atlassian.net/browse/WPB-5812)
  - #940 
  - #974
  - #994
  - #993
- [WPB-6660](https://wearezeta.atlassian.net/browse/WPB-6660)
  - #995
  - #1005
- [WPB-5814](https://wearezeta.atlassian.net/browse/WPB-5814)
  - #1017
  - Reverted #1034

#### Test Coverage

- [x] I have added automated test to this contribution

#### How to Test

- TBD:

----
#### PR Post Submission Checklist for internal contributors (Optional)

 - [x] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----
#### PR Post Merge Checklist for internal contributors

 - [x] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

[WPB-6660]: https://wearezeta.atlassian.net/browse/WPB-6660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[...]

PR #1017, opened 2024-02-21
feat: Periodically evaluate all 1:1 conversations - WPB-5814 

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5814" title="WPB-5814" target="_blank">Sub-taskWPB-5814</a>  [iOS] Periodically evaluate all 1:1 conversations
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

We periodically evaluate all 1:1 conversations. In this case we do so specifically after updating protocols if needed. I added a Unit Test to verify the behavior as well. 

💡 merges into epic #944

----
##### References


caldrian (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
```diff
@@ -267,6 +267,8 @@ public final class ZMUserSession: NSObject {
     let lastEventIDRepository: LastEventIDRepositoryInterface
     let conversationEventProcessor: ConversationEventProcessor
 
+    var supportedProtocolsService: SupportedProtocolsService?
+
```
netbe, line 271: I don't think we need to store it, we can initialize the service in demand, also solve optionality

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
```diff
@@ -413,8 +423,7 @@ public final class ZMUserSession: NSObject {
             guard let context = self?.syncContext else { return }
 
             context.perform {
-                let service = SupportedProtocolsService(context: context)
-                service.updateSupportedProtocols()
+                self?.supportedProtocolsService?.updateSupportedProtocols()
```
netbe, line 426:
```suggestion
               let service = SupportedProtocolsService(context: context,
                                                                   oneOnOneResolver: OneOnOneResolver(mlsService: self.mlsService))
               service.updateSupportedProtocols()
```

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
```diff
@@ -825,8 +834,7 @@ extension ZMUserSession: ZMSyncStateDelegate {
                 }
 
                 context.perform {
-                    let service = SupportedProtocolsService(context: context)
-                    service.updateSupportedProtocols()
+                    self?.supportedProtocolsService?.updateSupportedProtocols()
```
netbe, line 837:
```suggestion
               let service = SupportedProtocolsService(context: context,
                                                                   oneOnOneResolver: OneOnOneResolver(mlsService: self.mlsService))
               service.updateSupportedProtocols()
```

netbe (commented):
made a suggestion on creating SupportedProtocolsService

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
agisilaos: I will keep it as is for now. I feel otherwise it will make the code harder to read in multiple places since we will need to add a default value for the mls service in two places not just one.

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
agisilaos: I will keep it as is for now. I feel otherwise it will make the code harder to read in multiple places since we will need to add a default value for the mls service in two places not just one.

wire-ios-sync-engine/Source/UserSession/ZMUserSession.swift
agisilaos: I will keep it as is for now. I feel otherwise it will make the code harder to read in multiple places since we will need to add a default value for the mls service in two places not just one.

wire-ios-sync-engine/Source/Services/SupportedProtocolsService.swift
```diff
@@ -31,29 +31,48 @@ final class SupportedProtocolsService: SupportedProtocolsServiceInterface {
...
         let selfUser = userRepository.selfUser()
+        let previousProtocols = selfUser.supportedProtocols
+
         selfUser.supportedProtocols = calculateSupportedProtocols()
+
+        if previousProtocols != selfUser.supportedProtocols {
+            Task {
+                try await oneOnOneResolver.resolveAllOneOnOneConversations(in: context)
+            }
+        }
```
johnxnguyen, line 75: By resolving the one on one conversations here we're adding a new hidden responsibility of the `SupportedProtocolService`. In my opinion, this class is responsible only for calculating the current supported protocols of the self user and storing it on the self user.

I suggest to create a delegate method so that we can report externally that the supported protocols of the self user changed, something like:

```swift
delegate?.supportedProtocolsService(self, didCaluculateNewSupportedProtocols: selfUser.supportedProtocols)
```

johnxnguyen (changes requested):

wire-ios-sync-engine/Source/Services/SupportedProtocolsService.swift
markusfassbender: I agree this is not a good idea and turns around responsibilities as it was implemented by the slow sync for example right now... ❌

agisilaos: I started with that approach initially but after talking with @netbe he suggested to go another route. @johnxnguyen Using the delegate it will make a bit tricky to pass the resolver up to many classes. I remember having this problem as I initially implemented it that way. Thoughts?

markusfassbender:
Please revert the merge for now, it didn't have 2 approvals and had objections. We can find solutions afterwards.

agisilaos:
Apologies I thought @netbe approved. I'm working on the delegate solution right now.

PR #1028, opened 2024-02-21
feat: display other users' verification status in conversation details [WPB-765]

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-765" title="WPB-765" target="_blank">StoryWPB-765</a>  [iOS] 5.3.16 Indicate verified user
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Implements the displaying of other user's E2EI certification status on group conversation details and other users' profiles.

----

----

----
##### References


typfel (commented):
Looks good

wire-ios/Wire-iOS/Sources/Helpers/SemanticColors.swift
```diff
@@ -62,6 +62,8 @@ public enum SemanticColors {
         static let textInactive = UIColor(light: Asset.Colors.gray60, dark: Asset.Colors.gray70)
         static let textParticipantDisconnected = UIColor(light: Asset.Colors.red300Light, dark: Asset.Colors.red300Dark)
 
+        // UserCell: e.g. "Paul Nagel (You)"
+        static let youSuffix = UIColor(light: Asset.Colors.gray70, dark: Asset.Colors.gray60)
```
KaterinaWire: We have a format for semantic colors:  "<\usage>.<context/role>.<state?>"
Documentation at the top of the class

wire-ios/Wire-iOS/Sources/UserInterface/Availability/AvailabilityStringBuilder.swift
```diff
@@ -74,6 +79,12 @@ enum AvailabilityStringBuilder {
             interactive: false,
             color: textColor
         )
+        if appendYouSuffix, !L10n.Localizable.UserCell.Title.youSuffix.isEmpty {
```
KaterinaWire: Is it possible that` L10n.Localizable.UserCell.Title.youSuffix` isEmpty?

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsViewController.swift
```diff
@@ -58,6 +65,11 @@ final class GroupDetailsViewController: UIViewController, ZMConversationObserver
                 }
             }
         }
+
+        Task { @MainActor in
+            await updateUserE2EICertificationStatuses()
+            collectionViewController.sections = computeVisibleSections()
+        }
```
netbe: is this the same as :

```
Task {
            await updateUserE2EICertificationStatuses()
      let sections = computeVisibleSections()
       await MainActor.run {
            collectionViewController.sections = sections
       }
}
```

I moved the `computeVisibleSections` out because it's quite consuming and it *should* not be needed to be done on mainThread

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsViewController.swift
```diff
@@ -276,6 +298,11 @@ final class GroupDetailsViewController: UIViewController, ZMConversationObserver
 
         if changeInfo.participantsChanged, !conversation.isSelfAnActiveMember {
             navigationController?.popToRootViewController(animated: true)
+        } else {
+            Task { @MainActor in
+                await updateUserE2EICertificationStatuses()
+                collectionViewController.sections = computeVisibleSections()
+            }
```
netbe: can factor this logic to one function

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/Sections/ParticipantsSectionController.swift
```diff
@@ -216,19 +228,54 @@ final class ParticipantsSectionController: GroupDetailsSectionController {
...
+            cell.sectionName = viewModel.accessibilityTitle
+            cell.cellIdentifier = "participants.section.participants.cell"
+        }
+
+        lazy var unexpectedCellHandler: () -> UICollectionViewCell = {
+            let message = "Unexpected collection view cell type: \(String(describing: cell.self))"
+            WireLogger.conversation.error(message)
+            assertionFailure(message)
+            return cell
+        }
```
netbe: question: what's the benefit to have a lazy var in a function?

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
```diff
@@ -190,16 +193,41 @@ extension ConversationListViewController.ViewModel {
...
+                    conversation: selfConversation,
+                    user: selfUser
+                )
+                await MainActor.run {
+                    selfUserStatus.isCertified = isE2EICertified
+                }
+            } catch {
+                WireLogger.e2ei.error("failed to get E2EI certification status: \(error)")
+            }
+        }
```
typfel: We shouldn't be writing code like this in UI layer. This can become a use case?

wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/ProfileHeaderViewController.swift
```diff
@@ -353,6 +332,30 @@ final class ProfileHeaderViewController: UIViewController {
...
+                    user: selfUser
+                )
+                await MainActor.run {
+                    userStatus.isCertified = isE2EICertified
+                }
+            } catch {
+                WireLogger.e2ei.error("failed to get E2EI certification status: \(error)")
+            }
+        }
+    }
```
netbe: isn't it the same logic as in ConversationListViewController.ViewModel?

wire-ios-data-model/Source/Repositories/FeatureRepository.swift
```diff
@@ -21,6 +21,8 @@ import Foundation
 // sourcery: AutoMockable
 public protocol FeatureRepositoryInterface {
 
+    var context: NSManagedObjectContext { get }
```
typfel: Why do we need to expose the context in the protocol?

typfel (changes requested):

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsViewController.swift
caldrian: I don't think that would work that easily. That just brings us back to our ubiquitous issue of accessing core data instances from the wrong thread. I haven't tried it yet, but just from observing the code I expect it to run into issues as early as the filtering between admins and members happens:
```
let admins = participants.filter { $0.isGroupAdmin(in: conversation) }
```

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/Sections/ParticipantsSectionController.swift
caldrian: The benefit of lazy vars is the same in functions as well as for member variables: if it's not accessed, it will never consumer execution time or memory. The `unexpectedCellHandler` is only called from the two `guard`'s `else` block.

wire-ios-data-model/Source/Repositories/FeatureRepository.swift
caldrian: It's ugly, I agree. I was scared of changing the FeatureRepository's implementation.
I used it like this:
```
let isE2EIEnabled = await featureRepository.context.perform {
    featureRepository.fetchE2EI().isEnabled
}
```
But I will have to change the implementation of `fetchE2EI()`, because the initial guard is not concurrency safe:
```
        guard
            let feature = Feature.fetch(name: .e2ei, context: context),
            let featureConfig = feature.config
        else {
            return .init()
        }
```

wire-ios/Wire-iOS/Sources/Helpers/SemanticColors.swift
caldrian: I'm not sure what the correct name of this property would be, I'll just prefix it with `text` like all the others.

wire-ios-data-model/Source/Repositories/FeatureRepository.swift
caldrian: > Why do we need to expose the context in the protocol?

After thinking again, I'd answer your question with yes.
There is a requirement that the `FeatureRepository`'s internal implementation is only called from the same context's queue that was passed into the `FeatureRepository`'s initialiser. 
If that context is not part of the protocol, the requirement cannot be satisfied.

wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/ProfileHeaderViewController.swift
caldrian: Now, `ConversationListViewController.ViewModel` contains only logic for the self user (different use case type).

wire-ios/Wire-iOS/Sources/UserInterface/Availability/AvailabilityStringBuilder.swift
caldrian: Who knows what strings all the localisation files contain ...
I removed it since the implementation works with an empty string as well.

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
caldrian: I created a type `IsSelfUserE2EICertifiedUseCase` which does all the heavy lifting now.

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/Sections/ParticipantsSectionController.swift
caldrian: I guess the benefit is so little, that it's not worth talking about.
Changed it to `let`.

typfel (approved):

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsViewController.swift
netbe: ok sorry maybe my point was two actually. You answered the second one about performance of `computeVisibleSections()`.

The first one is: Does this
``` 
Task { @MainActor in
            await updateUserE2EICertificationStatuses()
            collectionViewController.sections = computeVisibleSections()
}
```
equals
``` 
Task {
            await updateUserE2EICertificationStatuses()
            await MainActor.run {
               collectionViewController.sections = computeVisibleSections()
             }
}
```
I wonder if in the first case we're blocking mainthread somehow since we're waiting for the updateUserE2EICertificationStatuses to be finished?

wire-ios-sync-engine/Source/UserSession/UserSession.swift
```diff
@@ -213,7 +213,7 @@ public protocol UserSession: AnyObject {
     var networkState: ZMNetworkState { get }
 
     var getUserClientFingerprint: GetUserClientFingerprintUseCaseProtocol { get }
-    var isSelfUserProteusVerifiedUseCase: IsSelfUserProteusVerifiedUseCaseProtocol { get }
+    var isUserE2EICertifiedUseCase: IsUserE2EICertifiedUseCaseProtocol { get }
```
KaterinaWire, line 216: Do we still need both `isUserE2EICertifiedUseCase` and `isSelfUserE2EICertifiedUseCase` (the next line)?

caldrian: Yes, that is intended.

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsViewController.swift
caldrian: In my current understanding about Swift Concurrency I consider the two snippets identical regarding run-time behaviour.
`await` never blocks, the second statement is just scheduled to be run on the main thread after the first statement is done. However, in the meantime, the main thread is running as normal.

wire-ios-mono.xcworkspace/xcshareddata/xcdebugger/Breakpoints_v2.xcbkptlist
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
KaterinaWire, line 1: Do we want to add these changes in `Breakpoints_v2.xcbkptlist` ?

wire-ios-data-model/WireDataModel.xcodeproj/project.pbxproj
```diff
@@ -3933,7 +3937,7 @@
 				63B1335A29A503D100009D84 /* MLSActionExecutor.swift in Sources */,
 				16460A46206544B00096B616 /* PersistentMetadataKeys.swift in Sources */,
 				5E67168E2174B9AF00522E61 /* LoginCredentials.swift in Sources */,
-				069BCC5A2B30945500DF4EC2 /* E2EIClientID.swift in Sources */,
+				069BCC5A2B30945500DF4EC2 /* E2eIClientID.swift in Sources */,
```
KaterinaWire: It should be `E2EIClientID.swift` (from the epic branch)

wire-ios-mono.xcworkspace/xcshareddata/xcdebugger/Breakpoints_v2.xcbkptlist
caldrian: Yes, I'd like to keep it until an issue is fixed.
It breaks into the debugger whenever a viewController's present method is called while the controller is already presenting.

wire-ios-data-model/WireDataModel.xcodeproj/project.pbxproj
caldrian: Not sure where this change came from, but it's gone since the last merge.

KaterinaWire (approved):

PR #1055, opened 2024-02-28
fix: certificate validation WPB-6882

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6882" title="WPB-6882" target="_blank">BugWPB-6882</a>  Valid E2EI certificates are set to Invalid Status on Device List is device details screens
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Valid certificates were shown as invalid.

### Causes

We were using the mls client id in the WireIdentity to fetch the local UserClient by remote identifier. These are different ids so we were never able to fetch the client.

Also, the handle inside the WireIdentity has a specific format that doesn't match the handle in the ZMUser, so validating the handle would always fail.

### Solutions

Parse the mls client id from the identity, then extract the UserClient remote identifier so we can fetch the client. 

Also, adjust the handle validation by checking the handle in the identity _contains_ the handle of the user.

#### Test Coverage

- Updated existing tests

----

----

----
##### References

### DIFF
--- a/wire-ios-sync-engine/Tests/Source/Use cases/GetE2eIdentityCertificatesUseCaseTests.swift
+++ b/wire-ios-sync-engine/Tests/Source/Use cases/GetE2eIdentityCertificatesUseCaseTests.swift
@@ -63,22 +63,23 @@ final class GetE2eIdentityCertificatesUseCaseTests: XCTestCase {
     ) -> WireCoreCrypto.WireIdentity {
         return .init(
             clientId: clientID,
-            handle: handle,
+            handle: "wireapp://%40\(handle)",
             displayName: name,
             domain: "local.com",
             certificate: mockCertificate,
             status: status,
-            thumbprint: "thumbprint",
-            serialNumber: "serialNumber",
-            notBefore: 0,
-            notAfter: 0
+            thumbprint: "QrsvPI0PDiJyAgsF-p3HoSyWLGWjyKwMdqlL0zWZOew",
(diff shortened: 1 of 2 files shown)


KaterinaWire (approved):

netbe (approved):

wire-ios-sync-engine/Tests/Source/Use cases/GetE2eIdentityCertificatesUseCaseTests.swift
```diff
@@ -63,22 +63,23 @@ final class GetE2eIdentityCertificatesUseCaseTests: XCTestCase {
     ) -> WireCoreCrypto.WireIdentity {
         return .init(
             clientId: clientID,
-            handle: handle,
+            handle: "wireapp://%40\(handle)",
```
theswiftway, line 66: question: Shouldn't this be `wireapp://%40\(handle)@\(domain)`?

theswiftway (approved):

wire-ios-sync-engine/Tests/Source/Use cases/GetE2eIdentityCertificatesUseCaseTests.swift
johnxnguyen: The handle being passed in is already "\(handle)@\(domain)"

PR #1074, opened 2024-03-01
fix: slow sync stuck resolving 1-1s WPB-6861

johnxnguyen:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When logging in, the slow sync gets stuck resolving 1-1 conversations indefinitely. Restarting the app doesn't help.

### Causes

If the user logging in has many 1-1s, then there are many 1-1s to resolve. Resolving a conversation may result in several network requests, for instance to fetch the mls one conversation. We were attempting to resolve each 1-1 in parallel within a task group, which resulted in a flood of network requests to the same endpoint (one for each conversation) in a very short amount of time. Quickly the BE returned back error responses with code 420 indicating that we hit rate limits. This code is not handled by the action handlers and the phase to resolve 1-1s failed but did not (for some reason) end the slow sync.

### Solutions

~The simplest solution is to resolve the 1-1s serially. This avoids the rate limits, but is fairly slow.~

When the action handler receives a rate limit error, push the action back in the list of pending actions. This ensures that the next time the operation loop runs again (after a backoff, handled elsewhere) it can try to perform the action again.

Another issue was found where the transport session tries to notify that new requests are available after backing off but the operation loop never starts again. This was because it was firing a "ZMReqestAvailableNotification" notification but the operation loop listens for "RequestAvailableNotification". I got rid of the former so there's only one notification.

#### How to Test

1. Log in with an MLS capable client with an account that has many contacts (either a large team or many accepted connection requests.
2. Assert that the slow sync finishes and the 1-1s are resolved.

----

----

----
##### References


typfel (approved):

wire-ios-request-strategy/Sources/Entity Actions/ActionHandler.swift
```diff
@@ -58,7 +58,13 @@ class ActionHandler<T: EntityAction>: NSObject, EntityActionHandler, ZMRequestGe
         let request = self.request(for: action, apiVersion: apiVersion)
 
         request?.add(ZMCompletionHandler(on: context, block: { [weak self] (response) in
-            self?.handleResponse(response, action: action)
+            if response.httpStatus.isOne(of: TooManyRequestsStatusCode, EnhanceYourCalmStatusCode) {
+                // We're being rate limited, put the action back so we can try again
+                // next time the operation loop polls.
```
netbe, line 63: 🤔 pulls or polls but again I am not native speaker
```suggestion
                // next time the operation loop pulls.
```

netbe (approved):
LGTM;) just a small comment

typfel (approved):
Nice! Not that much changed!

wire-ios-request-strategy/Sources/Entity Actions/ActionHandler.swift
typfel: I think polling is the correct term

typfel:
@johnxnguyen does it take a long to time to recover after you've been rate limited?

johnxnguyen:
> @johnxnguyen does it take a long to time to recover after you've been rate limited?

@typfel yes about 1min 20sec for an account with 94 one on one conversations, and we get limited about 5 times.

wire-ios-request-strategy/Sources/Entity Actions/ActionHandler.swift
johnxnguyen: Polls in the correct term

PR #1082, opened 2024-03-04
fix: migration hangs - WPB-6816

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6816" title="WPB-6816" target="_blank">BugWPB-6816</a>  [iOS] Wire stuck at "one moment, please" screen
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

During last db migration update, for some users the update hangs and never finishes (black shield screen).

### Causes (Optional)

This can happen during db migration of *large* dataset with a manual/heavy migration (the one with a mapping model) - 2.111.0.
In that case, the data is not copied from one location to another but loaded on context and then saved to a new persistent store, which results in big memory consumption and large core data saves.

### Solutions

Relies on lightweight migration instead and do manual pre and post actions.

* remove MappingModel for 2.111.0
* move policy code to CoreDataActions 

#### Test Coverage (Optional)

#### How to Test

Restore large db, it should takes seconds (Jacob's db took 7,5 sec - 100 MB)

### Note

This does not touch the database models (zmessaging).

This would need to be applied later to release 3.112.

### TODO

- [ ] This affects also 2.106-2.107 migration and will need to be addressed in another PR.


wire-ios-data-model/Source/ManagedObjectContext/Migration/2.110-2.111/PrefillPrimaryKeyAction.swift
```diff
@@ -0,0 +1,53 @@
...
+    override func execute(in context: NSManagedObjectContext) throws {
+        entityNames.forEach { entityName in
+            fillPrimaryKeys(for: entityName, context: context)
+        }
+    }
+
+    private func fillPrimaryKeys(for entityName: String, context: NSManagedObjectContext) {
+        do {
+            let request = NSFetchRequest<NSManagedObject>(entityName: entityName)
+            request.fetchBatchSize = 200
```
netbe: I defined a batchSize here, but that could be adjusted

wire-ios/Wire-iOS.xcodeproj/xcshareddata/xcschemes/Wire-iOS.xcscheme
```diff
@@ -261,7 +265,7 @@
       </EnvironmentVariables>
    </LaunchAction>
    <ProfileAction
-      buildConfiguration = "Release"
+      buildConfiguration = "Debug"
```
netbe, line 268: this can be left, as we don't have a release provisioning profile to profile the app

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataAction.swift
```diff
@@ -0,0 +1,60 @@
...
+            }
+        }
+        if let savedError {
+            throw savedError
+        }
+    }
+
+
+    func execute(in context: NSManagedObjectContext) throws {
+        // to be overiden by subclasses
```
typfel:
```suggestion
        // to be overriden by subclasses
```

wire-ios-data-model/Source/ManagedObjectContext/Migration/2.110-2.111/CoreDataAction2111.swift
```diff
@@ -0,0 +1,47 @@
...
+//
+
+
+import Foundation
+
+protocol CoreDataAction2111 {
+    func primaryKey(for object: NSManagedObject, entityName: String) -> String
+}
+
+extension CoreDataAction2111 {
```
typfel: I first thought this was the primary core data action then I realised this is a helper for calculating the primary key. It could also be a static func?

typfel (commented):
Looks OK, just some nitpicking.

How long does the migration take with these changes?

netbe:
> How long does the migration take with these changes?
@typfel (Jacob's db took 7,5 sec - 100 MB)

wire-ios-data-model/Source/ManagedObjectContext/Migration/2.110-2.111/CoreDataAction2111.swift
netbe: yes I can make it as a static func, is it the name of the protocol that is confusing?

typfel: Yes, it was the name.

wire-ios-data-model/Source/ManagedObjectContext/Migration/2.110-2.111/CoreDataAction2111.swift
```diff
@@ -0,0 +1,47 @@
...
+    func primaryKey(for object: NSManagedObject, entityName: String) -> String {
+
+        let remoteIdentifierData = object.value(forKey: ZMManagedObject.remoteIdentifierDataKey()) as? Data
+
+        switch entityName {
+        case ZMUser.entityName(), ZMConversation.entityName():
+            let path = entityName == ZMUser.entityName() ? #keyPath(ZMUser.domain) : #keyPath(ZMConversation.domain)
+
+            let domain = object.value(forKeyPath: path) as? String
+            return ZMManagedObject.primaryKey(from: remoteIdentifierData.flatMap(UUID.init(data: )), domain: domain)
```
markusfassbender: opinion: it's better understandable and easier to maintain to use the cases individually and concrete key paths and funcs:
```suggestion
        case ZMUser.entityName():
            let domain = object.value(forKeyPath: #keyPath(ZMUser.domain)) as? String
            let remoteIdentifier = remoteIdentifierData.flatMap(UUID.init(data:))
            return ZMUser.primaryKey(from: remoteIdentifier, domain: domain)

        case ZMConversation.entityName():
            let domain = object.value(forKeyPath: #keyPath(ZMConversation.domain)) as? String
            let remoteIdentifier = remoteIdentifierData.flatMap(UUID.init(data:))
            return ZMConversation.primaryKey(from: remoteIdentifier, domain: domain)
```

wire-ios-data-model/Source/ManagedObjectContext/Migration/2.110-2.111/CoreDataAction2111.swift
markusfassbender: Can it be a struct that implements the func without protocol and be used as object?
For me it's not easy to understand why the Actions like `PrefillPrimaryKeyAction` conform to the protocol, but don't implement it and then rely on `self.primaryKey(...)` by the default implementation, this feels obfuscated.

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMessagingMigrationVersion.swift
```diff
@@ -130,4 +129,70 @@ enum CoreDataMessagingMigrationVersion: String, CaseIterable {
...
+        }
+    }
+}
+
+extension CoreDataMessagingMigrationStep {
+    enum Error: Swift.Error {
+        case wrongModel
+    }
+
+    func runPreMigrationStep(for storeURL: URL) throws {
```
markusfassbender: I think the `enum CoreDataMessagingMigrationStep` is not the right place to implement a migration, it should not be its responsibility.

wire-ios-data-model/Source/ManagedObjectContext/Migration/2.110-2.111/PrefillPrimaryKeyAction.swift
```diff
@@ -0,0 +1,53 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+class PrefillPrimaryKeyAction: CoreDataAction, CoreDataAction2111 {
```
markusfassbender: Can we use composition over inheritance with error-prune overriding methods etc.?

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataAction.swift
```diff
@@ -0,0 +1,60 @@
...
+            loadError =  error
+        }
+        if let loadError {
+            throw loadError
+        }
+    }
+
+    func perform(with persistentContainer: NSPersistentContainer) throws {
+
+        try loadStores(for: persistentContainer)
```
markusfassbender: Should `loadStores` be executed each time the `perform` method is called?

wire-ios-data-model/Source/ManagedObjectContext/Migration/2.110-2.111/CoreDataAction2111.swift
netbe: @typfel I moved this with @markusfassbender to a struct

wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack.swift
```diff
@@ -233,6 +234,9 @@ public class CoreDataStack: NSObject, ContextProvider {
                     }
                     return
                 }
+                if tp?.warnIfLongerThanInterval() != true {
```
markusfassbender: I've added one commit to remove optionality in 785082f05f4aadedd5e1d1f6813274b62ee624cf, because I didn't understand the result of the `optional != true`.

markusfassbender (approved):

wire-ios-data-model/Source/ManagedObjectContext/Migration/2.110-2.111/PrefillPrimaryKeyAction.swift
markusfassbender: We keep the class as return type for now

typfel (approved):

PR #1097, opened 2024-03-06
fix(mls-migration): attempt finalisation directly after migration started

David-Henner:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Suggesting change to attempt to finalise the migration directly after it has been started instead of waiting 24 hours.
If there are any conversation that are ready to migrate from `mixed` to `mls`, they will migrate. The conversations that aren't ready will stay in `mixed` mode until every client in the conversation supports MLS or the migration finalisation time has been reached.

### DIFF
--- a/wire-ios-data-model/Source/MLS/Migration/ProteusToMLSMigrationCoordinator.swift
+++ b/wire-ios-data-model/Source/MLS/Migration/ProteusToMLSMigrationCoordinator.swift
@@ -91,6 +91,7 @@ public class ProteusToMLSMigrationCoordinator: ProteusToMLSMigrationCoordinating
         switch storage.migrationStatus {
         case .notStarted:
             try await startMigrationIfNeeded()
+            try await finaliseMigrationIfNeeded()
         case .started:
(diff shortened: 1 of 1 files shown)


markusfassbender:
In general I think it makes sense. I'm just thinking about the time consumed doing this for larger data sets... maybe the user opens the app only for 5 seconds and closes it again, can we run all of this in the background?
I know that this can happen also in the current implementation, but we split the single steps so potentially when doing both right away we have longer run times.

typfel:
> In general I think it makes sense. I'm just thinking about the time consumed doing this for larger data sets... maybe the user opens the app only for 5 seconds and closes it again, can we run all of this in the background? I know that this can happen also in the current implementation, but we split the single steps so potentially when doing both right away we have longer run times.

It's fine if we don't manage to complete this for all conversations, we'll try again later.

typfel (approved):

markusfassbender (approved):

agisilaos (approved):

PR #1120, opened 2024-03-12
fix: registration of intermediate certificates WPB-7077

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7077" title="WPB-7077" target="_blank">BugWPB-7077</a>  Not able to add federated accounts to group or have messages or conversations
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

In a federated context where both teams use E2EI. inability to add federated users to MLS groups, resulting in:
- Inability to migrate federated 1:1 conversations from proteus to mls
- Inability add federated users to MLS conversations
- Inability to exchange messages in MLS conversations

### Causes 

1. Misconfiguration of intermediate certificates registration resulting in `invalid identity` errors from core crypto. Intermediate certificate registration should happen after registering trust anchors and periodically every 24h

2. Incorrect parsing of the certificates fetched from the acme server. We receive a list of certificates that we need to parse and pass one by one to core crypto for registration. We erroneously passed the whole list to core crypto without any parsing

### Solutions

1. Add registration of intermediate certificates after registering trust anchors
2. Parse response from the federation certificates endpoint and register intermediate certificates one by one


KaterinaWire (approved):

wire-ios-request-strategy/Sources/E2EIdentity/E2EIRepository.swift
```diff
@@ -71,9 +71,11 @@ public final class E2EIRepository: E2EIRepositoryInterface {
...
     }
 
-    public func fetchFederationCertificate() async throws {
-        let federationCertificate = try await acmeApi.getFederationCertificate()
-        try await e2eiSetupService.registerFederationCertificate(federationCertificate)
+    public func fetchFederationCertificates() async throws {
+        let federationCertificates = try await acmeApi.getFederationCertificates()
+        for certificate in federationCertificates {
+            try await e2eiSetupService.registerFederationCertificate(certificate)
+        }
```
netbe: if one fail, we fail all? it could be interesting to catch the one failing, no?

markusfassbender: Maybe we can find a way to report to the caller, like return an array of errors instead of throwing right away or pass a fail closure?

wire-ios-request-strategy/Sources/E2EIdentity/ACME/AcmeAPITests.swift
```diff
@@ -272,16 +272,27 @@ class AcmeAPITests: ZMTBaseTest {
...
+        let mockCertificates = [
+            "certificate_1",
+            "certificate_2",
+            "certificate_3"
+        ]
+        let mockPayload = [
+            "crts": mockCertificates
+        ]
+
+        let mockData = try JSONSerialization.data(withJSONObject: mockPayload)
```
markusfassbender: Do we need to use serialization here? Midterm I would like to replace it with Data and Decodable.
In other tests usually define a json string and use .utf8 to make Data object.

David-Henner: It was simply faster to write it this way rather than using a json string. I could also create a `FederationCertificates` object and encode it to data. What do you think?
Also, what's the reason against using serialisation?

wire-ios-request-strategy/Sources/E2EIdentity/E2EIRepository.swift
David-Henner: Added a catch and logged the error, please check https://github.com/wireapp/wire-ios/pull/1120/commits/c3d1d3a6dcbb02ee4c98be0a37b8d237862ffa66

wire-ios-request-strategy/Sources/E2EIdentity/ACME/AcmeAPITests.swift
David-Henner: Updated https://github.com/wireapp/wire-ios/pull/1120/commits/c3d1d3a6dcbb02ee4c98be0a37b8d237862ffa66

markusfassbender: Thanks!

markusfassbender (approved):

netbe (approved):

PR #1122, opened 2024-03-13
feat: e2ei certificate renewal logic [WPB-3324]

theswiftway:
<!--do not remove this marker, its needed to replace info when ticket title is updated -->
<!--jira-description-action-hidden-marker-start-->

<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-3324" title="WPB-3324" target="_blank"><img alt="Story" src="https://wearezeta.atlassian.net/rest/api/2/universal_avatar/view/type/issuetype/avatar/10815?size=medium" />WPB-3324</a>  [iOS] 5.3.28 E2EI-certificate renewal logic
  </td></table>
  <br />
 

<!--jira-description-action-hidden-marker-end-->
<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Solutions

Update E2EI certificate when expired.
In order to get `renewalNudgingDate` we should deduct standard deductions from Validity Period (VP) and add it to `notValidBefore` date
 Standard deductions are : Server storage time(HT) , Grace period set by team admin(GP),  Random time in a day(UT)
  `Renewal nudging date = VP - (HT + GP + UT)`
 Here we calculate it from the other way where we deduct the standard deductions from the expiry date to get the renewal nudging date. This is done so as to be in sync with Android codebase

We nudge the user with an alert to update, if user has not been nudged.

[WPB-6805](https://wearezeta.atlassian.net/browse/WPB-6805)


### Testing

#### Test Coverage (Optional)

#### How to Test

**Steps**
1. Make sure your current client has E2EI certificate is expired or near expiry.
2. Go to Settings -> Devices -> Current Client
3. Tap on Update Certificate

Alternate
Upon opening the app, you should receive an update certificate alert when the certificate is near expiry.


### Attachments (Optional)

_Attachments like images, videos, etc. (drag and drop in the text box)_

----

----

----
##### References


[WPB-6805]: https://wearezeta.atlassian.net/browse/WPB-6805?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KaterinaWire (dismissed):
@theswiftway Don't forget to trigger `Crowdin action` to send new strings to Crowdin and pull translations back

wire-ios-sync-engine/Source/E2EI/LastE2EIdentityUpdateDateProtocol.swift
```diff
@@ -0,0 +1,66 @@
...
+@objc
+public protocol LastE2EIdentityUpdateDateProtocol {
+
+    func fetchLastAlertDate() -> Date?
+    func storeLastAlertDate(_ date: Date?)
+
+}
+
+@objc
+public final class LastE2EIdentityUpdateDate: NSObject, LastE2EIdentityUpdateDateProtocol {
```
KaterinaWire: We usually call it ` ...Repository`. Examples: CRLExpirationDatesRepository, GracePeriodRepository, SubconversationGroupIDRepository, etc.

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
```diff
@@ -0,0 +1,114 @@
...
+    private let serverStoragePeriod: TimeInterval
+    private let randomPeriod: TimeInterval
+    private let lastAlertDate: Date?
+    private let mlsGroupID: MLSGroupID
+    private let mlsClientID: MLSClientID
+
+    public init(
+        e2eCertificateForCurrentClient: GetE2eIdentityCertificatesUseCaseProtocol,
+        gracePeriod: TimeInterval,
+        serverStoragePeriod: TimeInterval = 28 * TimeInterval.oneDay,
```
KaterinaWire: I would move `serverStoragePeriod` and `randomPeriod` to constants.

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
```diff
@@ -0,0 +1,114 @@
...
+        self.gracePeriod = gracePeriod
+        self.lastAlertDate = lastAlertDate
+        self.serverStoragePeriod = serverStoragePeriod
+        self.randomPeriod = randomPeriod
+        self.mlsGroupID = mlsGroupID
+        self.mlsClientID = mlsClientID
+    }
+
+    public func invoke() async throws -> E2EIdentityCertificateUpdateStatus {
+        if let certificate = try await e2eCertificateForCurrentClient.invoke(
```
KaterinaWire: I think we should check `e2eiFeature.isEnabled`, no?

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
```diff
@@ -122,45 +156,44 @@ final class E2EINotificationActionsHandler: E2EINotificationActions {
...
-            typealias Alert = L10n.Localizable.FailetToUpdateCertificate.Alert
-            typealias Button = L10n.Localizable.FailetToUpdateCertificate.Button
+            typealias UpdateAlert = L10n.Localizable.FailedToUpdateCertificate.Alert
+            typealias Alert = L10n.Localizable.FailedToGetCertificate.Alert
+            typealias Button = L10n.Localizable.FailedToGetCertificate.Button
 
-            let title = Alert.title
-            let message = canCancel ? Alert.message : Alert.forcedMessage
+            let title = isUpdateMode ? UpdateAlert.title : Alert.title
+            let message = canCancel ? (isUpdateMode ? UpdateAlert.message : Alert.message) : Alert.forcedMessage
```
KaterinaWire: Let's make it simple, it's not easy to read
let message = canCancel ? (isUpdateMode ? UpdateAlert.message : Alert.message) : Alert.forcedMessage

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
```diff
@@ -31,28 +31,37 @@ public protocol E2EINotificationActions {
...
 
     // MARK: - Properties
-
     private var enrollCertificateUseCase: EnrollE2EICertificateUseCaseProtocol
     private var snoozeCertificateEnrollmentUseCase: SnoozeCertificateEnrollmentUseCaseProtocol
     private var stopCertificateEnrollmentSnoozerUseCase: StopCertificateEnrollmentSnoozerUseCaseProtocol
     private let gracePeriodRepository: GracePeriodRepository
     private let targetVC: UIViewController
-
+    weak var userSession: UserSession?
```
KaterinaWire: userSession is very `heavy` object, why not pass `e2eIdentityUpdateCertificateUpdateStatus` instead?

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
```diff
@@ -0,0 +1,114 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public enum E2EIdentityCertificateUpdateStatus {
```
caldrian, line 21: Why is that enum at the top of the file `E2EIdentityCertificateUpdateStatusUseCase.swift`? I'd expect a type `E2EIdentityCertificateUpdateStatusUseCase` first.

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
```diff
@@ -0,0 +1,114 @@
...
+    case noAction, reminder, block
```
caldrian: Please add documentation to these cases.

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// sourcery: AutoMockable
+public protocol E2EIdentityCertificateUpdateStatusProtocol {
```
caldrian: Maybe rename the protocol?
```suggestion
public protocol E2EIdentityCertificateUpdateStatusUseCaseProtocol {
```

Also put it into a separate file please.

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
```diff
@@ -0,0 +1,114 @@
...
+    func invoke() async throws -> E2EIdentityCertificateUpdateStatus
+}
+public extension Notification.Name {
+    static let checkForE2EICertificateStatus = NSNotification.Name("CheckForE2EICertificateStatus")
```
caldrian: Please add documentation for this notification name.
Also, why is it declared in `E2EIdentityCertificateUpdateStatusUseCase.swift`?

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
```diff
@@ -0,0 +1,114 @@
...
+}
+
+final public class E2EIdentityCertificateUpdateStatusUseCase: E2EIdentityCertificateUpdateStatusProtocol {
```
caldrian:
```suggestion
public struct E2EIdentityCertificateUpdateStatusUseCase: E2EIdentityCertificateUpdateStatusProtocol {
```

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
```diff
@@ -0,0 +1,114 @@
...
+    private let randomPeriod: TimeInterval
+    private let lastAlertDate: Date?
+    private let mlsGroupID: MLSGroupID
+    private let mlsClientID: MLSClientID
+
+    public init(
+        e2eCertificateForCurrentClient: GetE2eIdentityCertificatesUseCaseProtocol,
+        gracePeriod: TimeInterval,
+        serverStoragePeriod: TimeInterval = 28 * TimeInterval.oneDay,
+        randomPeriod: TimeInterval = Double((0..<Int(TimeInterval.oneDay)).randomElement() ?? 0), // Random time in a day
```
caldrian:
```suggestion
        randomPeriod: TimeInterval = .random(in: 0 ..< TimeInterval.oneDay), // Random offset
```

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
```diff
@@ -0,0 +1,114 @@
...
+public protocol E2EIdentityCertificateUpdateStatusProtocol {
+    func invoke() async throws -> E2EIdentityCertificateUpdateStatus
+}
+public extension Notification.Name {
+    static let checkForE2EICertificateStatus = NSNotification.Name("CheckForE2EICertificateStatus")
+}
+
+final public class E2EIdentityCertificateUpdateStatusUseCase: E2EIdentityCertificateUpdateStatusProtocol {
+
+    private let e2eCertificateForCurrentClient: GetE2eIdentityCertificatesUseCaseProtocol
```
caldrian:
```suggestion
    private let e2eiCertificateForCurrentClient: GetE2eIdentityCertificatesUseCaseProtocol
```

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+UserSession.swift
```diff
@@ -294,6 +294,29 @@ extension ZMUserSession: UserSession {
         let featureRepository = FeatureRepository(context: syncContext)
         return GetMLSFeatureUseCase(featureRepository: featureRepository)
     }
+
+    @MainActor
```
caldrian, line 298: Why `@MainActor`? I think it should be removed.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+UserSession.swift
```diff
@@ -294,6 +294,29 @@ extension ZMUserSession: UserSession {
...
+            return ZMConversation.fetchSelfMLSConversation(in: self.syncContext)?.mlsGroupID
+        }
+    }
+
+    @MainActor
+    public func e2eIdentityUpdateCertificateUpdateStatus() async -> E2EIdentityCertificateUpdateStatusProtocol? {
+        guard let mlsGroupID = await self.fetchSelfConversationMLSGroupID(), let selfMLSClientID else {
+            return nil
+        }
+        var updateE2EICertificate: E2EIdentityCertificateUpdateStatusProtocol = E2EIdentityCertificateUpdateStatusUseCase(
```
caldrian:
```suggestion
        var updateE2EICertificate = E2EIdentityCertificateUpdateStatusUseCase(
```

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+UserSession.swift
```diff
@@ -294,6 +294,29 @@ extension ZMUserSession: UserSession {
...
     }
+
+    @MainActor
+    public func fetchSelfConversationMLSGroupID() async -> MLSGroupID? {
+        return await syncContext.perform {
+            return ZMConversation.fetchSelfMLSConversation(in: self.syncContext)?.mlsGroupID
+        }
+    }
+
+    @MainActor
```
caldrian, line 305: Please remove this attribute.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+UserSession.swift
```diff
@@ -294,6 +294,29 @@ extension ZMUserSession: UserSession {
...
+    public func e2eIdentityUpdateCertificateUpdateStatus() async -> E2EIdentityCertificateUpdateStatusProtocol? {
+        guard let mlsGroupID = await self.fetchSelfConversationMLSGroupID(), let selfMLSClientID else {
+            return nil
+        }
+        var updateE2EICertificate: E2EIdentityCertificateUpdateStatusProtocol = E2EIdentityCertificateUpdateStatusUseCase(
+            e2eCertificateForCurrentClient: getE2eIdentityCertificates,
+            gracePeriod: Double(e2eiFeature.config.verificationExpiration),
```
caldrian: `e2eiFeature` must be accessed from the view context's queue.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -20,6 +20,7 @@ import Foundation
 import WireDataModel
 import WireSystem
 import WireRequestStrategy
+import UIKit
```
caldrian: There shouldn't be any need for UIKit in this file.
For NotificationCenter please import Foundation.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -336,6 +337,15 @@ public final class ZMUserSession: NSObject {
             context: syncContext)
     }()
 
+    public lazy var selfMLSClientID: MLSClientID? = {
+        guard let selfUserClient  else {
+            return nil
+        }
+       return  MLSClientID(userClient: selfUserClient)
+    }()
```
caldrian: Please remove this and just call `MLSClientID(userClient:)` directly wherever it is needed.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -336,6 +337,15 @@ public final class ZMUserSession: NSObject {
...
+
+    public var lastE2EIUpdateDate: LastE2EIdentityUpdateDateProtocol?
```
caldrian: This shouldn't be modifiable from outside.
```suggestion
    public private(set) var lastE2EIUpdateDate: LastE2EIdentityUpdateDateProtocol?
```

wire-ios-sync-engine/Tests/Source/E2EI/SelfClientCertificateProviderTests.swift
```diff
@@ -141,7 +141,7 @@ class SelfClientCertificateProviderTests: MessagingTest {
 
 }
 
-private extension String {
+extension String {
```
caldrian: Actually this whole extension can be removed, there is the exact same code in `String+Mock.swift`.

wire-ios-sync-engine/Tests/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCaseTests.swift
```diff
@@ -0,0 +1,204 @@
...
+
+    func testThatItReturnsNoAction_WhenExpiryDateIsBeyondSevenDays() async {
+        update(certificate: certificate(with: .oneWeek + .oneDay))
+        do {
+            let result = try await e2eIdentityCertificateStatus.invoke()
+            XCTAssertEqual(result, .noAction)
+        } catch {
+            XCTFail(error.localizedDescription)
+        }
+    }
```
caldrian: No need for `do/catch`, please adjust the other tests as well.

```suggestion
    func testThatItReturnsNoAction_WhenExpiryDateIsBeyondSevenDays() async throws {
        update(certificate: certificate(with: .oneWeek + .oneDay))
        let result = try await e2eIdentityCertificateStatus.invoke()
        XCTAssertEqual(result, .noAction)
    }
```

wire-ios-sync-engine/Tests/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCaseTests.swift
```diff
@@ -0,0 +1,204 @@
...
+        do {
+            let result = try await e2eIdentityCertificateStatus.invoke()
+            XCTAssertEqual(result, .reminder)
+        } catch {
+            XCTFail(error.localizedDescription)
+        }
+    }
+
+    func testThatItReturnsNoAction_WhenExpiryDateIsWithInSevenDaysAndLastAlertWasDisplayedToday() async {
+        e2eIdentityCertificateStatus = E2EIdentityCertificateUpdateStatusUseCase(
```
caldrian: Instead of repeating the whole initialisation I'd rather make the user case properties `public var` (assuming the use case is a `struct`) and just set the `lastAlertDate`.

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/SuccessfulCertificateEnrollmentViewController.swift
```diff
@@ -183,8 +195,8 @@ final class SuccessfulCertificateEnrollmentViewController: AuthenticationStepVie
         wrapNavigationController.presentTopmost()
     }
 
-    @objc
-    private func okTapped() {
+    @MainActor
+    @objc private func okTapped() {
```
caldrian: This can be reverted.

wire-ios/Wire-iOS/Sources/UserInterface/Helpers/UIAlertAction+Convenience.swift
```diff
@@ -39,13 +39,15 @@ extension UIAlertAction {
...
-        presenter: UIViewController?
+        presenter: UIViewController?,
+        onDismiss: (() -> Void)? = nil
     ) -> Self {
         return .init(
             title: title,
             style: .default
         ) { [weak presenter] _ in
             let browserViewController = BrowserViewController(url: url)
+            browserViewController.onDismiss = onDismiss
```
caldrian, line 50: Is this really required?
The `BrowserViewController.onDismiss` closure is called in `UIViewController.dismiss(animated:completion:)` but I don't think any `BrowserViewController/SFSafariViewController` should present other view controllers modally.

caldrian (approved):

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
```diff
@@ -31,28 +31,37 @@ public protocol E2EINotificationActions {
...
+        userSession: UserSession?,
         targetVC: UIViewController) {
             self.enrollCertificateUseCase = enrollCertificateUseCase
             self.snoozeCertificateEnrollmentUseCase = snoozeCertificateEnrollmentUseCase
             self.stopCertificateEnrollmentSnoozerUseCase = stopCertificateEnrollmentSnoozerUseCase
             self.gracePeriodRepository = gracePeriodRepository
+            self.userSession = userSession
             self.targetVC = targetVC
+
+            NotificationCenter.default.addObserver(forName: .checkForE2EICertificateStatus, object: nil, queue: .main) { _ in
```
caldrian: We must store the return value of this method and use it to unregister!

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
```diff
@@ -31,28 +31,37 @@ public protocol E2EINotificationActions {
...
+    weak var userSession: UserSession?
+    private var isUpdateMode: Bool = false
     // MARK: - Life cycle
 
     init(
         enrollCertificateUseCase: EnrollE2EICertificateUseCaseProtocol,
         snoozeCertificateEnrollmentUseCase: SnoozeCertificateEnrollmentUseCaseProtocol,
         stopCertificateEnrollmentSnoozerUseCase: StopCertificateEnrollmentSnoozerUseCaseProtocol,
         gracePeriodRepository: GracePeriodRepository,
+        userSession: UserSession?,
```
caldrian: We've been trying hard to get rid of all the user session variables in views and view controller.
Please don't inject any user session here.
It seems to be only needed for accessing two use cases, if I'm not mistaken.
Please inject the use cases instead.

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
caldrian:
```suggestion
    private weak var userSession: UserSession?
```

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
netbe: we can add comment specified in the use case to match what is what : VP - (HT + GP + UT)

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/SuccessfulCertificateEnrollmentViewController.swift
```diff
@@ -116,6 +125,9 @@ final class SuccessfulCertificateEnrollmentViewController: AuthenticationStepVie
         super.viewDidLoad()
 
         view.backgroundColor = SemanticColors.View.backgroundDefault
+        if isUpdateMode {
+            self.lastE2EIdentityUpdateDate?.storeLastAlertDate(Date.now)
+        }
```
caldrian: This is a side effect nobody expects.

Don't inject any use case into this VC. It's purely for presenting information.
Let the presenter of this view controller invoke the use case instead, please.

caldrian (changes requested):

wire-ios/Wire-iOS Tests/UserInterface/E2EIdentity/SuccessfulCertificateEnrollmentViewControllerSnapshotTests.swift
```diff
@@ -22,19 +22,13 @@ import XCTest
...
-
-    override func setUp() {
-        super.setUp()
-        sut = .init()
-    }
-
-    override func tearDown() {
-        sut = nil
-        super.tearDown()
+    func testThatItShouldShowAppriateMessage_WhenEnrolE2eIdentityIsSuccessful() {
```
netbe:
```suggestion
    func testThatItShouldShowAppropriateMessage_WhenEnrolE2EIdentityIsSuccessful() {
```

wire-ios/Wire-iOS Tests/UserInterface/E2EIdentity/SuccessfulCertificateEnrollmentViewControllerSnapshotTests.swift
```diff
@@ -22,19 +22,13 @@ import XCTest
...
+        let sut = SuccessfulCertificateEnrollmentViewController(lastE2EIdentityUpdateDate: nil, isUpdateMode: false)
+        verify(matching: sut)
     }
 
-    func testLayout() {
+    func testThatItShouldShowAppriateMessage_WhenUpdateE2eIdentityIsSuccessful() {
```
netbe:
```suggestion
    func testThatItShouldShowAppropriateMessage_WhenUpdateE2EIdentityIsSuccessful() {
```

wire-ios/Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -10837,6 +10838,7 @@
 				F15650C321DD0A9B00210504 /* Bundle+Backend.swift in Sources */,
 				F15650B221DD018700210504 /* Bundle+Identifiers.swift in Sources */,
 				F15650B021DCFF5300210504 /* AutomationHelper.swift in Sources */,
+				259AAB0E2B9F4EAF00B13A7C /* MLSClientResolver.swift in Sources */,
```
netbe: question: was this not removed ?

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EIFeatureChange.swift
```diff
@@ -24,34 +24,47 @@ enum E2EIChangeAction: CaseIterable {
 }
 
 extension UIAlertController {
+    private typealias MlsE2eiStrings = L10n.Localizable.FeatureConfig.Alert.MlsE2ei
```
netbe: nitpic: 🤓
```suggestion
    private typealias MLSE2EIStrings = L10n.Localizable.FeatureConfig.Alert.MlsE2ei
```

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EIFeatureChange.swift
```diff
@@ -24,34 +24,47 @@ enum E2EIChangeAction: CaseIterable {
...
-            handler(.remindLater)
-        }
 
         controller.addAction(learnMoreAction)
         controller.addAction(getCertificateAction)
-        controller.addAction(remindLaterAction)
+
+        if canRemindLater {
+            let remindLaterAction = UIAlertAction(title: MlsE2eiStrings.Button.remindMeLater,
+                                                  style: .destructive) {_ in
```
netbe: question: shouldn't it be `.cancel` instead of `.destructive`?

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/SuccessfulCertificateEnrollmentViewController.swift
theswiftway: Crashes at certain times otherwise

wire-ios/Wire-iOS/Sources/UserInterface/Helpers/UIAlertAction+Convenience.swift
theswiftway: I need this to soft block the app upon certificate exipiration

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
theswiftway: https://developer.apple.com/documentation/foundation/notificationcenter/1413994-removeobserver

No need to this from iOS 9.0

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/SuccessfulCertificateEnrollmentViewController.swift
theswiftway: This gets invoked at many places. That would result in duplicate code everywhere.

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/SuccessfulCertificateEnrollmentViewController.swift
```diff
@@ -20,34 +20,41 @@ import SwiftUI
...
+    typealias LocalizedEnrollE2eiCertificate = L10n.Localizable.EnrollE2eiCertificate
+    typealias LocalizedUpdateE2eiCertificate = L10n.Localizable.UpdateE2eiCertificate
+
     // MARK: - Properties
 
+    var certificateDetails: String = ""
     public var onOkTapped: ((_ viewController: SuccessfulCertificateEnrollmentViewController) -> Void)?
 
-    private let titleLabel: UILabel = {
+    private let isUpdateMode: Bool
```
netbe, line 33: suggestion: I am not sure to follow what this mode is, maybe rename it to have a case with enroll and update ?

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
```diff
@@ -0,0 +1,114 @@
...
+    public func invoke() async throws -> E2EIdentityCertificateUpdateStatus {
+        if let certificate = try await e2eCertificateForCurrentClient.invoke(
+            mlsGroupId: mlsGroupID,
+            clientIds: [mlsClientID]
+           ).first {
+
+            if certificate.expiryDate.isInThePast {
+                return .block
+            }
+            let timeLeft = certificate.expiryDate.timeIntervalSinceNow - serverStoragePeriod - gracePeriod - randomPeriod
```
netbe: question: is timeLeft the  "Renewal nudging time"?

netbe: or the Time left before to end of grace period date?

netbe (commented):
adding some questions to clarify

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
caldrian: You are referring to this, I assume:

"If your app targets iOS 9.0 and later or macOS 10.11 and later, and you used [addObserver(_:selector:name:object:)](https://developer.apple.com/documentation/foundation/notificationcenter/1415360-addobserver), you do not need to unregister the observer. If you forget or are unable to remove the observer, the system cleans up the next time it would have posted to it."

But it's wrong. You're not using `addObserver(_:selector:name:object:)`!

You are using `addObserver(forName:object:queue:using:)`, see https://developer.apple.com/documentation/foundation/notificationcenter/1411723-addobserver.

There you can read:

"Return Value: An opaque object to act as the observer. Notification center strongly holds this return value until you remove the observer registration."

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
```diff
@@ -0,0 +1,119 @@
...
+    ) {
+        self.e2eCertificateForCurrentClient = e2eCertificateForCurrentClient
+        self.gracePeriod = gracePeriod
+        self.lastAlertDate = lastAlertDate
+        self.mlsGroupID = mlsGroupID
+        self.mlsClientID = mlsClientID
+        self.comparedDate = comparedDate
+    }
+
+    @MainActor
```
caldrian: There is no point in putting a MainActor attribute on a use case method. Please remove!

caldrian (changes requested):

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EIFeatureChange.swift
theswiftway: done

wire-ios/Wire-iOS Tests/UserInterface/E2EIdentity/SuccessfulCertificateEnrollmentViewControllerSnapshotTests.swift
theswiftway: done

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/SuccessfulCertificateEnrollmentViewController.swift
theswiftway: This is to show update message

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
theswiftway: updated the code to better understand the logic

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/SuccessfulCertificateEnrollmentViewController.swift
theswiftway: removed it

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
theswiftway: Not applicable anymore

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
theswiftway: removed it

wire-ios-sync-engine/Tests/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCaseTests.swift
theswiftway: done

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
theswiftway: no longer applicable

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
theswiftway: done

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
theswiftway: It is just a preference. I prefer the components required for a class to top first. consider this like a constant. Constants are usually declared on the top

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
theswiftway: done

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
theswiftway: Not sure if we are going to gain anything by putting this dependency in a separate file. It will just take more time to compile.

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
theswiftway: Moved this ZMUserSession for now. Suggest where do you want this ?

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
theswiftway: can't do this as this is a dependency in UsersessionType which is accessed by Objective C members

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+UserSession.swift
theswiftway: It crashes otherwise

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+UserSession.swift
theswiftway: It crashes otherwise

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
theswiftway: done

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
theswiftway: done

wire-ios/Wire-iOS.xcodeproj/project.pbxproj
theswiftway: removed

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EIFeatureChange.swift
theswiftway: done

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+UserSession.swift
theswiftway: done. Good catch

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+UserSession.swift
theswiftway: not applicable anymore

wire-ios-data-model/Source/Model/UserClient/E2eIdentityCertificate.swift
```diff
@@ -63,3 +77,33 @@ public enum E2EIdentityCertificateStatus: CaseIterable {
...
+    }
+
+    func shouldUpdate(with gracePeriod: TimeInterval) -> Bool {
+        let renewalNudgingDate = renewalNudgingDate(with: gracePeriod)
+        return isExpired || (isActivated && comparedDate >= renewalNudgingDate)
+    }
+
+    /// In order to get `renewalNudgingDate` we should deduct standard deductions from Validity Period (VP) and add it to `notValidBefore` date
+    /// Standard deductions are : Server storage time(HT) , Grace period set by team admin(GP),  Random time in a day(UT)
+    /// Renewal nuding date = VP - (HT + GP + UT)
```
netbe:
```suggestion
    /// Renewal nudging date = VP - (HT + GP + UT)
```

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
theswiftway: done

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
```diff
@@ -19,96 +19,108 @@
...
 }
-public extension Notification.Name {
-    static let checkForE2EICertificateStatus = NSNotification.Name("CheckForE2EICertificateStatus")
-}
 
-final public class E2EIdentityCertificateUpdateStatusUseCase: E2EIdentityCertificateUpdateStatusProtocol {
+final public class E2EIdentityCertificateUpdateStatusUseCase: E2EIdentityCertificateUpdateStatusUseCaseProtocol {
 
-    private let e2eCertificateForCurrentClient: GetE2eIdentityCertificatesUseCaseProtocol
+    private let e2eiCertificateForCurrentClient: GetE2eIdentityCertificatesUseCaseProtocol
```
netbe: as convention we call the useCases variables the same name but we drop the UseCase suffix

```suggestion
    private let getE2eIdentityCertificates: GetE2eIdentityCertificatesUseCaseProtocol
```

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
```diff
@@ -19,96 +19,108 @@
...
+        ).first else {
+            return .noAction
+        }
+
+        if certificate.expiryDate.isInThePast {
+            return .block
+        }
+
+        let renewalNudgingDate = certificate.renewalNudgingDate(with: gracePeriod)
+        if  renewalNudgingDate > comparedDate && renewalNudgingDate < certificate.expiryDate {
```
netbe:
```suggestion
        if renewalNudgingDate > comparedDate && renewalNudgingDate < certificate.expiryDate {
```

wire-ios-sync-engine/Source/UserSession/UserSession.swift
```diff
@@ -234,11 +234,11 @@ public protocol UserSession: AnyObject {
 
     var updateMLSGroupVerificationStatus: UpdateMLSGroupVerificationStatusUseCaseProtocol { get }
 
-    var lastE2EIUpdateDate: LastE2EIdentityUpdateDateProtocol? { get }
+    var lastE2EIUpdateDate: LastE2EIdentityUpdateDateRepositoryInterface? { get }
```
netbe: without the repository suffix, I would tend to read it's a date
```suggestion
    var lastE2EIUpdateDateRepository: LastE2EIdentityUpdateDateRepositoryInterface? { get }
```

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
theswiftway: done

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -337,14 +336,7 @@ public final class ZMUserSession: NSObject {
...
 
-    public lazy var selfMLSClientID: MLSClientID? = {
-        guard let selfUserClient  else {
-            return nil
-        }
-       return  MLSClientID(userClient: selfUserClient)
-    }()
-
-    public var lastE2EIUpdateDate: LastE2EIdentityUpdateDateProtocol?
+    private(set) public var lastE2EIUpdateDate: LastE2EIdentityUpdateDateRepositoryInterface?
```
netbe:
```suggestion
    private(set) public var lastE2EIUpdateDateRepository: LastE2EIdentityUpdateDateRepositoryInterface?
```

wire-ios/Tests/Mocks/UserSessionMock.swift
```diff
@@ -25,13 +25,13 @@ import WireRequestStrategySupport
 @testable import Wire
 
 final class UserSessionMock: UserSession {
-    var lastE2EIUpdateDate: WireSyncEngine.LastE2EIdentityUpdateDateProtocol?
+    var lastE2EIUpdateDate: LastE2EIdentityUpdateDateRepositoryInterface?
```
netbe:
```suggestion
    var lastE2EIUpdateDateRepository: LastE2EIdentityUpdateDateRepositoryInterface?
```

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
```diff
@@ -36,32 +36,42 @@ final class E2EINotificationActionsHandler: E2EINotificationActions {
     private var stopCertificateEnrollmentSnoozerUseCase: StopCertificateEnrollmentSnoozerUseCaseProtocol
     private let gracePeriodRepository: GracePeriodRepository
     private let targetVC: UIViewController
-    weak var userSession: UserSession?
+    private var lastE2EIdentityUpdateAlertDate: LastE2EIdentityUpdateDateRepositoryInterface?
```
netbe:
```suggestion
    private var lastE2EIdentityUpdateAlertDateRepository: LastE2EIdentityUpdateDateRepositoryInterface?
```

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
```diff
@@ -36,32 +36,42 @@ final class E2EINotificationActionsHandler: E2EINotificationActions {
...
+    private var e2eIdentityCertificateUpdate: E2EIdentityCertificateUpdateStatusUseCaseProtocol?
```
netbe:
```suggestion
    private var e2eIdentityCertificateUpdateStatus: E2EIdentityCertificateUpdateStatusUseCaseProtocol?
```

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
```diff
@@ -106,40 +117,39 @@ final class E2EINotificationActionsHandler: E2EINotificationActions {
...
                         authenticate: oauthUseCase.invoke)
                     await self.confirmSuccessfulEnrollment(certificateDetails)
                 } catch {
                     WireLogger.e2ei.error("failed to \(self.isUpdateMode ? "update" : "get") E2EI certification status: \(error)")
-                    isUpdateMode = false
                 }
+                self.isUpdateMode = false
             }
+        } cancelled: {[weak self] in
+            self?.isUpdateMode = false
```
netbe, line 150: here we use weak self but not on the other closure , normal?

netbe (commented):
looks almost there, just some last naming issues.
@theswiftway Plus I don't see the crowdin PR merge commit in it, we should merge it before merging this.

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
theswiftway: self is captured inside a Task. Let me fix that for other closures in the file.

theswiftway:
> looks almost there, just some last naming issues. @theswiftway Plus I don't see the crowdin PR merge commit in it, we should merge it before merging this.

Yet to get an update from Astrid. She said she would try it today. If not on Monday.

wire-ios-sync-engine/Source/UserSession/UserSession.swift
theswiftway: done

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
theswiftway: done

wire-ios/Tests/Mocks/UserSessionMock.swift
theswiftway: done

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
theswiftway: done

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
theswiftway: done

wire-ios-data-model/Source/E2EIdentity/GracePeriodRepository.swift
```diff
@@ -18,15 +18,16 @@
...
+// sourcery: AutoMockable
 public protocol GracePeriodRepositoryInterface {
 
     func fetchGracePeriodEndDate() -> Date?
     func storeGracePeriodEndDate(_ date: Date)
 
 }
 
+/// The repository is responsible for storing the grace period during which the user must enroll the end-to-end identity certificate.
+/// The grace period starts at the moment where the client fetches and processes the team feature flag.
```
caldrian: Actually the documentation should be at the protocol. It describes what a conforming type should be capable of, independently of the actual implementation. 

If you are worried about the `sourcery` comment, it worked for me like that:
```
// sourcery: AutoMockable
/// Determines if a user has a valid E2EI certificate on all clients.
public protocol IsUserE2EICertifiedUseCaseProtocol {
```

wire-ios-data-model/Source/Model/UserClient/E2eIdentityCertificate.swift
```diff
@@ -23,6 +23,14 @@ public enum E2EIdentityCertificateStatus: CaseIterable {
     case notActivated, revoked, expired, invalid, valid
 }
 
+public enum E2eIdentityCertificateConstants {
```
caldrian, line 26: Since the constants are only used in this file, shouldn't they be `private` instead?
Also, couldn't we move this enum below the main content of this file (`class E2eIdentityCertificate`)?

wire-ios-data-model/Source/Model/UserClient/E2eIdentityCertificate.swift
```diff
@@ -52,6 +64,8 @@ public enum E2EIdentityCertificateStatus: CaseIterable {
         self.status = certificateStatus
         self.serialNumber = serialNumber
         self.comparedDate = comparedDate
+        self.serverStoragePeriod = serverStoragePeriod
+        self.randomPeriod = randomPeriod
     }
 
     public struct DateProvider: DateProviding {
```
caldrian, line 71: There is one `SystemDateProvider` and two mock-date providers. This implementation should be removed.

wire-ios-data-model/Source/Model/UserClient/E2eIdentityCertificate.swift
```diff
@@ -42,7 +52,9 @@ public enum E2EIdentityCertificateStatus: CaseIterable {
         expiryDate: Date,
         certificateStatus: E2EIdentityCertificateStatus,
         serialNumber: String,
-        comparedDate: Date = DateProvider(now: .now).now
+        comparedDate: Date = DateProvider(now: .now).now,
```
caldrian: This line doesn't really make sense.
Either you inject a date provider (type `DateProviding`) and access the `now` property whenever you need a current date/time, or you simply inject a date:
```suggestion
        comparedDate: Date = .now,
```

Opinion: 
However, in my opinion a certificate shouldn't have any `compareDate` property.
The operations which are defined in the extension below, could be moved to a separate type, like a `E2EIdentityCertificateValidator`, but with a better name.

wire-ios-sync-engine/Source/Use cases/SnoozeCertificateEnrollmentUseCase.swift
```diff
@@ -20,7 +20,7 @@ import Foundation
 
 // sourcery: AutoMockable
 public protocol SnoozeCertificateEnrollmentUseCaseProtocol {
-    func invoke() async
+    func invoke(isUpdateMode: Bool) async
```
caldrian, line 23: Please add a description for the parameter.
```suggestion
    /// <#Description#>
    /// - Parameter isUpdateMode: <#isUpdateMode description#>
    func invoke(isUpdateMode: Bool) async
```

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
```diff
@@ -31,28 +31,47 @@ public protocol E2EINotificationActions {
...
     private var stopCertificateEnrollmentSnoozerUseCase: StopCertificateEnrollmentSnoozerUseCaseProtocol
     private let gracePeriodRepository: GracePeriodRepository
     private let targetVC: UIViewController
-
+    private var lastE2EIdentityUpdateAlertDateRepository: LastE2EIdentityUpdateDateRepositoryInterface?
+    private var e2eIdentityCertificateUpdateStatus: E2EIdentityCertificateUpdateStatusUseCaseProtocol?
+    private var isUpdateMode: Bool = false
     // MARK: - Life cycle
-
+    private var observer: Any?
```
caldrian: The return value of `NotificationCenter.default.addObserver` is of type `NSObjectProtocol`.

```suggestion
    private var observer: NSObjectProtocol?
```

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
```diff
@@ -31,28 +31,47 @@ public protocol E2EINotificationActions {
...
+                queue: .main
+            ) { _ in
+                Task { [weak self] in
+                    await self?.updateCertificate()
+                }
+            }
         }
 
+    deinit {
+        self.observer = nil
```
caldrian: Setting this property to `nil` here is neither sufficient nor required (every member variable will be released on `deinit`).
Please unregister from the notification center:
```suggestion
        if let observer {
            NotificationCenter.default.removeObserver(observer)
        }
```

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
```diff
@@ -0,0 +1,126 @@
...
+    // certificate expired so soft block user to update certificate
+    case block
+}
+
+// sourcery: AutoMockable
+public protocol E2EIdentityCertificateUpdateStatusUseCaseProtocol {
+    func invoke() async throws -> E2EIdentityCertificateUpdateStatus
+}
+
+final public class E2EIdentityCertificateUpdateStatusUseCase: E2EIdentityCertificateUpdateStatusUseCaseProtocol {
```
caldrian:
```suggestion
public struct E2EIdentityCertificateUpdateStatusUseCase: E2EIdentityCertificateUpdateStatusUseCaseProtocol {
```

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
caldrian: If we don't have a dedicated place for notification names, I think it should be close to the place where it is sent.

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
caldrian: I checked out your branch in order to try making it a struct and it worked.

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
theswiftway: done

wire-ios-data-model/Source/Model/UserClient/E2eIdentityCertificate.swift
theswiftway: done

wire-ios-data-model/Source/Model/UserClient/E2eIdentityCertificate.swift
theswiftway: Enum 'E2eIdentityCertificateConstants' is internal and cannot be referenced from a default argument value.
I like to keep the constants which are dependencies on top of the file as it s easier to reference.

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
theswiftway: done

wire-ios-data-model/Source/Model/UserClient/E2eIdentityCertificate.swift
theswiftway: compared date is for aiding unit testing

wire-ios-data-model/Source/E2EIdentity/GracePeriodRepository.swift
theswiftway: done

wire-ios-data-model/Source/Model/UserClient/E2eIdentityCertificate.swift
theswiftway: done

wire-ios-data-model/Source/Model/UserClient/E2eIdentityCertificate.swift
theswiftway: I tried extracting it out. It leads to undesirable changes across the code. May be thing for future if this extension grows

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
theswiftway: done

wire-ios-sync-engine/Source/Use cases/E2EIdentityCertificateUpdateStatusUseCase.swift
theswiftway: done

wire-ios-sync-engine/Source/Use cases/SnoozeCertificateEnrollmentUseCase.swift
theswiftway: done

wire-ios-system/Source/DateProviding/SystemDateProvider.swift
```diff
@@ -26,4 +26,6 @@ extension DateProviding where Self == SystemDateProvider {
 /// Provides date values based on the system clock.
 public struct SystemDateProvider: DateProviding {
     public var now: Date { .now }
+
+    public init() {}
```
netbe, line 30: why public, isn't it DateProviding.system accessible?

netbe (approved):

wire-ios-sync-engine/Source/Use cases/SnoozeCertificateEnrollmentUseCase.swift
```diff
@@ -72,10 +74,14 @@ final class SnoozeCertificateEnrollmentUseCase: SnoozeCertificateEnrollmentUseCa
...
+                NotificationCenter.default.post(name: .checkForE2EICertificateExpiryStatus, object: nil)
+            } else {
+                // We save the end of the grace period once and should not update it.
+                let notificationObject = FeatureRepository.FeatureChange.e2eIEnabled(gracePeriod: nil)
+                NotificationCenter.default.post(name: .featureDidChangeNotification,
+                                                object: notificationObject)
+            }
         }
 
         recurringActionService.registerAction(recurringAction)
```
caldrian, line 86: All the other actions are registered on the main thread. I wonder if we should add something like ...
```suggestion
        await MainActor.run {
            recurringActionService.registerAction(recurringAction)
        }
```

wire-ios-sync-engine/Source/Use cases/SnoozeCertificateEnrollmentUseCase.swift
```diff
@@ -72,10 +74,14 @@ final class SnoozeCertificateEnrollmentUseCase: SnoozeCertificateEnrollmentUseCa
...
-            NotificationCenter.default.post(name: .featureDidChangeNotification,
-                                            object: notificationObject)
+            if isUpdateMode {
+                NotificationCenter.default.post(name: .checkForE2EICertificateExpiryStatus, object: nil)
+            } else {
+                // We save the end of the grace period once and should not update it.
+                let notificationObject = FeatureRepository.FeatureChange.e2eIEnabled(gracePeriod: nil)
+                NotificationCenter.default.post(name: .featureDidChangeNotification,
+                                                object: notificationObject)
+            }
```
caldrian, line 83: Could you please explain to me briefly, when this code might be executed?
When is `RecurringActionService.performActionsIfNeeded()` called?

wire-ios/Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -345,25 +346,29 @@ extension AppRootRouter {
...
-                account: selectedAccount,
-                userSession: userSession
-            )
-        else {
-            completion()
-            return
-        }
-
-        self.authenticatedRouter = authenticatedRouter
+        Task {
```
caldrian: I wouldn't have expected changes to AppRootRouter.swift. Is this really required for your feature?

wire-ios-system/Source/DateProviding/SystemDateProvider.swift
theswiftway: Used `SystemDateProvider` directly.  Not sure if it is a good practice to use `DateProviding.system`

wire-ios/Wire-iOS/Sources/AppRootRouter.swift
theswiftway: Yes. As the usecases that we set as dependencies are fetched asynchronously. this has led to this change.

wire-ios-sync-engine/Source/Use cases/SnoozeCertificateEnrollmentUseCase.swift
theswiftway: This is a reminder service. There's a calculation in `SnoozeTimeProvider` based on which the next reminder alert would be shown for enrolling or updating a certificate.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientListViewController.swift
```diff
@@ -207,11 +207,14 @@ final class ClientListViewController: UIViewController,
             contextProvider: contextProvider
         )
         viewModel.showCertificateUpdateSuccess = {[weak self] certificateChain in
-            self?.updateAllClients {
-                self?.updateE2EIdentityCertificateInDetailsView()
+            guard let self = self, let lastE2EIUpdateDateRepository = userSession.lastE2EIUpdateDateRepository else {
```
caldrian: `lastE2EIUpdateDateRepository` doesn't seem to be used, please remove it.

theswiftway: done

wire-ios/Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -428,7 +434,7 @@ extension AppRootRouter {
         }
 
         let needToShowDialog = appStateCalculator.wasUnauthenticated && !isTeamMember
-
+        let e2eIdentityCertificateUpdate = await userSession.e2eIdentityUpdateCertificateUpdateStatus()
```
caldrian: I think most of the changes in this file should be reverted.
Please try to remove `async` from `userSession.e2eIdentityUpdateCertificateUpdateStatus()`.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+UserSession.swift
```diff
@@ -294,6 +294,31 @@ extension ZMUserSession: UserSession {
...
+    public func fetchSelfConversationMLSGroupID() async -> MLSGroupID? {
+        return await syncContext.perform {
+            return ZMConversation.fetchSelfMLSConversation(in: self.syncContext)?.mlsGroupID
+        }
+    }
+
+    @MainActor
+    public func e2eIdentityUpdateCertificateUpdateStatus() async -> E2EIdentityCertificateUpdateStatusUseCaseProtocol? {
+        guard let selfUserClient,
+              let mlsGroupID = await self.fetchSelfConversationMLSGroupID(),
```
caldrian: If we always need the id of the self-mls-conversation, there is no point in injecting it into the use case.
It should rather be part of the use case to fetch the conversation.

Then the `async` can be removed from this method.

theswiftway: done

wire-ios/Wire-iOS/Sources/AppRootRouter.swift
theswiftway: done

wire-ios-sync-engine/Source/Use cases/SnoozeCertificateEnrollmentUseCase.swift
theswiftway: done

wire-ios-sync-engine/Source/UserSession/MLSGroupIDProvider.swift
```diff
@@ -0,0 +1,41 @@
...
+
+    public init(context: NSManagedObjectContext) {
+        self.context = context
+    }
+
+    @MainActor
+    public func fetchMLSGroupID() async -> MLSGroupID? {
+        precondition(context.zm_isSyncContext)
+        return await context.perform {
+            return ZMConversation.fetchSelfMLSConversation(in: self.context)?.mlsGroupID
```
caldrian: Since we only fetch the "self"-mls-conversation, should we rename the use case accordingly?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+UserSession.swift
```diff
@@ -294,6 +294,30 @@ extension ZMUserSession: UserSession {
         let featureRepository = FeatureRepository(context: syncContext)
         return GetMLSFeatureUseCase(featureRepository: featureRepository)
     }
+
+    @MainActor
+    public func fetchSelfConversationMLSGroupID() async -> MLSGroupID? {
+        return await syncContext.perform {
+            return ZMConversation.fetchSelfMLSConversation(in: self.syncContext)?.mlsGroupID
```
caldrian, line 301: What about using `MLSGroupIDProviding`/`MLSGroupIDProvider`?

wire-ios/Tests/Mocks/UserSessionMock.swift
```diff
@@ -25,6 +25,16 @@ import WireRequestStrategySupport
 @testable import Wire
 
 final class UserSessionMock: UserSession {
+    var lastE2EIUpdateDateRepository: LastE2EIdentityUpdateDateRepositoryInterface?
+
+    func fetchSelfConversationMLSGroupID() async -> WireDataModel.MLSGroupID? {
+        return MLSGroupID(Data())
+    }
+
+    func e2eIdentityUpdateCertificateUpdateStatus() async -> WireSyncEngine.E2EIdentityCertificateUpdateStatusUseCaseProtocol? {
```
caldrian:
```suggestion
    func e2eIdentityUpdateCertificateUpdateStatus() -> E2EIdentityCertificateUpdateStatusUseCaseProtocol? {
```

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+UserSession.swift
```diff
@@ -294,6 +294,30 @@ extension ZMUserSession: UserSession {
         let featureRepository = FeatureRepository(context: syncContext)
         return GetMLSFeatureUseCase(featureRepository: featureRepository)
     }
+
+    @MainActor
+    public func fetchSelfConversationMLSGroupID() async -> MLSGroupID? {
```
caldrian, line 299: Actually, now that we have a "use case" `MLSGroupIDProviding`/`MLSGroupIDProvider`, can't this be removed?

caldrian (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+UserSession.swift
caldrian: Will be done in a future refactoring.

PR #924, opened 2024-02-01
fix: Make initializer in OneOnOneResolver non optional - WPB-5815

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5815" title="WPB-5815" target="_blank">Sub-taskWPB-5815</a>  [iOS] Periodically evaluate a single 1:1 conversation
  </td></table>
  <br />
 

----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [x] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [x] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [x] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

With this PR we make the initializer of the OneOnOneResolver non optional and instead we make the Migrator to be optional. The reason for this change is that sometimes we want to use the OneOnOneResolver with the migrator and sometimes we don't. 

----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #1126, opened 2024-03-13
fix: conversation not visible [WPB-6203]

markusfassbender:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6203" title="WPB-6203" target="_blank">BugWPB-6203</a>  [iOS] MLS: Conversation visible on Web but not on iOS
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

See ticket.

### Causes (Optional)

- The `oneOnOneUser` property was never set and thus the conversation was filtered out as invalid
- The 1-1 conversation mls were not joined correctly, when `epoch > 0` from other clients

### Solutions

- Link 1-1 conversations with other user
- when `epoch > 0` we join the conversation via `MLSService`
- Move `syncMLSConversationFromBackend` from migrator to resolver
- Inject MLSService instead of syncContext to access mlsService
- Change tests from uiMOC to syncContext

#### Test Coverage (Optional)

#### How to Test

- See ticket

----

----

----
##### References


wire-ios-request-strategy/Sources/Payloads/Processing/ConversationEventPayloadProcessor.swift
```diff
@@ -665,25 +669,33 @@ struct ConversationEventPayloadProcessor {
...
-        updateConversationStatus(from: payload, for: conversation)
-        conversation.needsToBeUpdatedFromBackend = false
+            conversation.conversationType = self.conversationType(for: conversation, from: conversationType)
+            updateAttributes(from: payload, for: conversation, context: context)
+            updateMessageProtocol(from: payload, for: conversation, in: context)
+            updateMetadata(from: payload, for: conversation, context: context)
+            updateMembers(from: payload, for: conversation, context: context)
+            updateConversationTimestamps(for: conversation, serverTimestamp: serverTimestamp)
+            updateConversationStatus(from: payload, for: conversation)
+            linkOneOnOneUserIfNeeded(for: conversation)
```
markusfassbender, line 685: Hint: the link to `oneOnOneUser` was missing to filter via predicates correctly.

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
```diff
@@ -92,65 +92,98 @@ public final class OneOnOneResolver: OneOnOneResolverInterface {
...
+    private func resolveCommonUserProtocolMLS(
+        with userID: QualifiedID,
+        in context: NSManagedObjectContext
+    ) async throws -> OneOnOneConversationResolution {
+        WireLogger.conversation.debug("should resolve to mls 1-1 conversation")
 
-            guard let migrator else {
-                throw OneOnOneResolverError.migratorNotFound
-            }
+        guard let migrator else {
```
caldrian, line 146: Could the `migrator` be made non-optional?

caldrian (approved):

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
markusfassbender: @agisilaos please help me here :)
I had the same question when it became optional, but remember that there was case where the migrator was not initialised but the resolver should work for the other cases (proteus, mixed)... so the answer was no, it needs to be optional.

wire-ios-sync-engine/Source/Use cases/CreateTeamOneOnOneConversationUseCase.swift
```diff
@@ -112,11 +114,15 @@ struct CreateTeamOneOnOneConversationUseCase: CreateTeamOneOnOneConversationUseC
             throw Error.mlsMigratorNotFound
         }
 
-        let groupID: MLSGroupID
+        let groupID = try await syncMLSConversationFromBackend(
+            userID: userID,
+            in: context
+        )
```
markusfassbender, line 120: Does it make sense that we sync from backend in case we `createMLSConversation`?
If not how to retrieve the groupID, that is needed for `migrateToMLS`?

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
markusfassbender: It was in #924 and I think we can not change it for now, unfortunately.

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
```diff
@@ -43,30 +44,28 @@ public final class OneOnOneResolver: OneOnOneResolverInterface {
 
     private let protocolSelector: OneOnOneProtocolSelectorInterface
     private let migrator: OneOnOneMigratorInterface?
+    private let mlsService: MLSServiceInterface?
```
typfel: Why is it optional? When does it make sense to initialise a OneOnOneResolve without a `MLSService`?

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
```diff
@@ -92,68 +91,151 @@ public final class OneOnOneResolver: OneOnOneResolverInterface {
...
-                else {
-                    return
-                }
+            return await resolveCommonUserProtocolNone(with: userID, in: context)
+        case .mls:
+            return try await resolveCommonUserProtocolMLS(with: userID, in: context)
+        case .proteus:
+            return resolveCommonUserProtocolProteus()
+        case .mixed:
+            return resolveCommonUserProtocolMixed()
```
typfel: This can never happen a user cannot support `.mixed` as a protocol. Only conversations have the protocol set to `.mixed`.

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
```diff
@@ -92,68 +91,151 @@ public final class OneOnOneResolver: OneOnOneResolverInterface {
...
+    ) async throws -> OneOnOneConversationResolution {
+        WireLogger.conversation.debug("should resolve to mls 1-1 conversation")
 
-            let mlsGroupIdentifier = try await migrator.migrateToMLS(
+        guard let migrator else {
+            throw OneOnOneResolverError.migratorNotFound
+        }
+
+        guard let mlsService else {
+            throw OneOnOneResolverError.migratorNotFound
```
David-Henner:
```suggestion
            throw OneOnOneResolverError.mlsServiceNotFound
```

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
```diff
@@ -92,68 +91,151 @@ public final class OneOnOneResolver: OneOnOneResolverInterface {
...
+        if epoch == 0 {
+            // migrate to a new conversation
+            try await migrator.migrateToMLS(
                 userID: userID,
+                mlsGroupID: mlsGroupID,
                 in: context
             )
+        } else {
+            // join existing conversation via external commit
+            try await mlsService.joinGroup(with: mlsGroupID)
```
typfel: This is not enough we also need to perform what's done inside `migrator.migrateToMLS()` with the difference that we join the group instead of establishing it.

typfel (changes requested):

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
David-Henner: When looking at `migrateToMLS` above, migrating to MLS if epoch is 0 involves two steps:
- Establishing the MLS group  (in `establishLocalMLSConversationIfNeeded`)
- Switching from the proteus `ZMConversation` to the mls `ZMConversation` (in `switchLocalConversationToMLS`)

In the case where epoch is > 0, wouldn't we only join the MLS group but miss switching from the proteus conversation to the mls conversation?

wire-ios-data-model/Source/Model/User/ZMUser.swift
johnxnguyen, line 1: I know this may be out of the scope of the PR, but it bugs me that we have this logic on a ZMUser instance and that it's in the data model.

If we extracted this out into a use case, perhaps we would be able to:
- create (and possibly retain) the resolver on the user session
- avoid the optional mls service (and also migrator) in the resolver since the user session has the mls service
- create the use case from the user session with all necessary dependencies

What do you think?

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
```diff
@@ -92,68 +91,151 @@ public final class OneOnOneResolver: OneOnOneResolverInterface {
...
         // Users can only support proteus and mls protocols.
         // Mixed protocol is used by conversations to represent
         // the migration state when migrating from proteus to mls.
-        case .mixed:
-            assertionFailure("users should not have mixed protocol")
-            return .noAction
-        }
+        assertionFailure("users should not have mixed protocol")
+        return .noAction
     }
```
David-Henner: in this case I would prefer not to have this in a separate method, I think it's clearer if the comment and the assertion are shown directly within the `mixed` case

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
markusfassbender: My idea was to have the same level of abstraction in the switch, but I can revert it :)

Screenshot 2024-03-21 at 12 47 42

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
markusfassbender: My idea was to have the same level of abstraction in the switch, but I can revert it :)

Screenshot 2024-03-21 at 12 47 42

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
markusfassbender: What exactly? If we just join locally a 1-1 conversation because it didn't exist, we have no history to copy

markusfassbender: Can you please describe a case where this happen? I thought either:
- we have a proteus conversation the epoch is always 0 when mls conversation is created, then the migration is executed
- we have an existing mls conversation, so the epoch is > 0 and we just join with the new device

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
markusfassbender: You can init a OneOnOneResolver if you have only proteus and the other user only mls. The resolver then sets the 1-1 conversation to read only.

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
typfel: The epoch being zero or not is not dependent upon there being an existing  proteus conversation.

If epoch > 0 some else already established the MLS group and it already exists on the backend:
    1. Join MLS group via external commit
    2. Set `oneOnOneConversation` to MLS 1-1 conversations
    3. Migrate messages from proteus 1-1 conversation

if epoch == 0 MLS group doesn't exist on backend so we need to create it
  1. Establish MLS group with other user
  2. Set `oneOnOneConversation` to MLS 1-1 conversations
  3. Migrate messages from proteus 1-1 conversation

The only difference between the two cases is what we do in step 1.

markusfassbender: Thanks, that helped. With David I discussed it similar yesterday and did the changes here: a5b2e7beb6da359e4e1597f14a69365aeeaec213

And there are more things to consider:
- if the epoch can not be found in core data locally, there is another mistake that e.g. the preconditions to sync the conversation from the backend didn't work properly
- when no proteus conversation exists, because the user login on the new device the first time, `step 3` must be skipped (instead of throwing an error, because the conversation can not be found etc).

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
markusfassbender: Solved by other changes

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
markusfassbender: See other discussion below

wire-ios-data-model/Source/Model/User/ZMUser.swift
markusfassbender: Yes it would be better, but I doubt I can this change now... will take a note to look at it after the deadline! thanks

David-Henner (approved):

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
```diff
@@ -92,68 +87,95 @@ public final class OneOnOneResolver: OneOnOneResolverInterface {
...
-                    let conversation = otherUser.oneOnOneConversation
-                else {
-                    return
-                }
+            return await resolveCommonUserProtocolNone(with: userID, in: context)
+        case .mls:
+            return try await resolveCommonUserProtocolMLS(with: userID, in: context)
+        case .proteus:
+            return resolveCommonUserProtocolProteus()
+        case .mixed:
```
typfel, line 86: This comes from the fact that we are using the same type for two unrelated things. We should probably split it into something like  `ConversationProtocol` and `UserSupportedProtocol`.

Not now of course.

typfel (approved):

caldrian (approved):
Superficially it looks good to me, however, I'm not very familiar with MLS yet.

PR #1134, opened 2024-03-14
feat: OAuthUseCase improvements

KaterinaWire:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

 1. Change the keycloak redirectURI, we don't need to have it different for every app.
 2. Set `prefersEphemeralSession` to `true`.

----

----

----
##### References


wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/OAuthUseCase.swift
```diff
@@ -42,7 +42,7 @@ public class OAuthUseCase: OAuthUseCaseInterface {
         logger.info("invoke authentication flow")
 
         guard let bundleID = Bundle.main.bundleIdentifier,
-              let redirectURI = URL(string: "\(bundleID):/oauth2redirect")
+              let redirectURI = URL(string: "wire://e2ei/oauth2redirect")
```
netbe: did you test it on C builds? as wire is not a scheme, does that work?
you can change the WIRE_URL_SCHEME to check locally?

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/OAuthUseCase.swift
```diff
@@ -109,7 +109,10 @@ public class OAuthUseCase: OAuthUseCaseInterface {
 
     @MainActor
     private func execute(authorizationRequest: OIDAuthorizationRequest) async throws -> OAuthResponse {
-        guard let userAgent = OIDExternalUserAgentIOS(presenting: rootViewController) else {
+        guard let userAgent = OIDExternalUserAgentIOS(
+            presenting: rootViewController,
+            prefersEphemeralSession: true)
+        else {
```
markusfassbender: 💅 Somehow I would rather format this way, but it's up to you:
```suggestion
        guard let userAgent = OIDExternalUserAgentIOS(
            presenting: rootViewController,
            prefersEphemeralSession: true
        ) else {
```

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/OAuthUseCase.swift
markusfassbender: Can it happen that users have multiple wire apps installed and then another app is called? 🤓

markusfassbender (approved):

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/OAuthUseCase.swift
KaterinaWire: `ASWebAuthenticationSession ensures that only the calling app’s session receives the authentication callback, even when more than one app registers the same callback URL scheme.`

https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession

KaterinaWire: yes, with `WIRE_URL_SCHEME = wire-c3-test`

netbe (approved):

PR #1148, opened 2024-03-19
fix: system message when added to group WPB-7094

typfel:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7094" title="WPB-7094" target="_blank">BugWPB-7094</a>  [iOS] Missing Added to Group System Messages
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

There's an incorrect system message about MLS migration when you're added an MLS group and there's no "memberJoined" or "newConversation" system message.

### Causes

1, The MLS migration message gets inserted because when processing the memberJoin event we insert an incomplete conversation which assigns proteus as the message protocol. Afterwards we sync the conversation metadata from the backend which thinks conversation was migration since message protocol is then updated to MLS.
2. The "newConversation" system message is missing because with MLS we don't get any `conversation.create` events and by the time we fetch metadata the conversation is not considered new since a incomplete conversation by the memberJoin event.
3. The "memberJoined" system message doesn't get inserted but are missing then necessary metadata when processing it `conversationType`.

### Solutions

Fetch conversation metadata on `memberJoined` and "welcomeMessage" if and only if we don't already know about it.

#### Test Coverage

----

----

----
##### References

### DIFF
--- a/wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationEventProcessor.swift
+++ b/wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationEventProcessor.swift
@@ -150,6 +150,9 @@ public class ConversationEventProcessor: NSObject, ConversationEventProcessorPro
             from: event.payload
         ) else { return }
 
+        if let conversationID = payload.qualifiedID {
+            await conversationService.syncConversationIfMissing(qualifiedID: conversationID)
+        }
         await context.perform {
             self.processor.processPayload(
                 payload,
@@ -157,7 +160,6 @@ public class ConversationEventProcessor: NSObject, ConversationEventProcessorPro
                 in: self.context
             )
         }
-        await syncConversationForMLSStatus(payload: payload)
     }
 
     private func processConversationRename(_ event: ZMUpdateEvent) async {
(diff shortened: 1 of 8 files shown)


wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationEventProcessor.swift
```diff
@@ -150,14 +150,16 @@ public class ConversationEventProcessor: NSObject, ConversationEventProcessorPro
...
+            await conversationService.syncConversationIfMissing(qualifiedID: conversationID)
+        }
         await context.perform {
             self.processor.processPayload(
                 payload,
                 originalEvent: event,
                 in: self.context
             )
         }
-        await syncConversationForMLSStatus(payload: payload)
```
typfel, line 160: @johnxnguyen idea what purpose this served? I don't see why we would need to update any MLS status upon receiving a `memberJoined` event, that's done when the welcome message arrives.

David-Henner: IIRC This was done because the `memberJoin` and `welcome` events don't always arrive in the same order.
- `memberJoin` creates the `ZMConversation`  
- `welcome` creates the mls group

So when we receive a memberJoin event we need to set the correct `mlsStatus` depending on wether or not the mls group already exists.

typfel: Ok I think this is not true anymore because inside `process(welcomeMessage:)` the first we do is to call

```
await conversationService.syncConversation(qualifiedID: conversationID)
```

I think this was logic was from the time when `conversationID` inside the WelcomeMessage event didn't reference the correct conversation.

See https://wearezeta.atlassian.net/browse/WPB-3435

David-Henner (approved):

johnxnguyen (approved):

PR #1149, opened 2024-03-19
fix: core data migration errors on 3.111.8 - WPB-7174

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7174" title="WPB-7174" target="_blank">BugWPB-7174</a>  [iOS] core data crashes on startup 3.111.8
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

On 3.111.8 version, there are some reports of migration errors with Cocoa error `133021`
`// merge policy failed - unable to complete merging due to multiple conflicting constraint violations`

### Causes (Optional)

On that version we add constraints on uniqueness of primary key composed of remoteIdentifier and domain. During the migration finding the duplicates it seems to have a case where we won't detect duplicate objects having no remoteIdentifier and/or domain. (primaryKey is a concatenation of remoteIdentifier_data as a string and the domain).

In the unit test, it appears the error `133021` appears only for the `remoteIdentifier_data` property.

### Solutions

Add extra step before removeDuplicate to remove all objects with no remoteIdentifier.

#### Test Coverage (Optional)

### Notes (Optional)

It is not clear if how we could end up in that situation.
Another solution to treat the objects with no remoteIdentifier could be to set a value to these objects. Not sure how backend will handle it and other side effects.


wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack.swift
```diff
@@ -235,7 +235,7 @@ public class CoreDataStack: NSObject, ContextProvider {
                     return
                 }
                 if tp.warnIfLongerThanInterval() == false {
-                    WireLogger.localStorage.debug("time spent in migration only: \(tp.elapsedTime)")
+                    WireLogger.localStorage.debug("time spent in migration only: \(tp.elapsedTime)", attributes: .safePublic)
```
markusfassbender: Is the `debug` level filtered out for production builds (although the safe public attribute) is set?

wire-ios-data-model/Source/ManagedObjectContext/Migration/2.110-2.111/RemoveDuplicatePreAction.swift
```diff
@@ -29,10 +29,26 @@ class RemoveDuplicatePreAction: CoreDataMigrationAction {
...
+    // Method to cleanup objects without remoteIdentifierDataKey before removeDuplicates
+    private func removeNilPrimaryKey(for entityName: String, context: NSManagedObjectContext) {
+        let request = NSFetchRequest<NSManagedObject>(entityName: entityName)
+        request.predicate = NSPredicate(format: "%K = nil", ZMManagedObject.remoteIdentifierDataKey())
+        do {
+            let objectsWithNil = try context.fetch(request)
+
+            for object in objectsWithNil {
+                context.delete(object)
+            }
```
markusfassbender, line 46: Maybe we can add a log how many objects will be deleted to make this step visible in the happy case.

wire-ios-data-model/Source/ManagedObjectContext/Migration/2.110-2.111/PrefillPrimaryKeyAction.swift
```diff
@@ -38,11 +38,15 @@ class PrefillPrimaryKeyAction: CoreDataMigrationAction {
             let request = NSFetchRequest<NSManagedObject>(entityName: entityName)
             request.fetchBatchSize = batchSize
             let objects = try context.fetch(request)
-
+            var index = 0
             objects.forEach { object in
```
markusfassbender: Maybe we can add a log how many objects will be updated to make this step visible in the happy case.

wire-ios-data-model/Tests/Source/ManagedObjectContext/DatabaseRemoteIdentifierUniquenessTests.swift
```diff
@@ -0,0 +1,75 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+@testable import WireDataModel
+
+final class DatabaseRemoteIdentifierUniquenessTests: XCTestCase {
```
markusfassbender, line 22: Nice it inherits only from `XCTestCase` 👍

markusfassbender (approved):

wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack.swift
netbe: no the safePublic here is for os_log to not redact the log

netbe: actually you're right I'll increase to info

wire-ios-data-model/Source/ManagedObjectContext/Migration/2.110-2.111/RemoveDuplicatePreAction.swift
netbe: done

johnxnguyen (approved):

PR #1150, opened 2024-03-20
fix(e2ei): revoked certificates not updated WPB-3230

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-3230" title="WPB-3230" target="_blank">StoryWPB-3230</a>  [iOS] 5.3.18 Client with revoked certificate
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

I do not see when users have revoked certificates (including my own user). Revoked certificates continue to show as valid, unless I do a fresh login.

### Causes

Due to caching, we don't pass the regenerated CRL containing the newly revoked certificate to CC. Instead we pass the old cached CRL.

### Solutions

Use an `URLSession` with an `ephemeral` configuration and disable caching

**Note:** I'm also making `updateAllStatuses` non-throwing to avoid returning early in case we fail to update the status of one of the conversation


wire-ios-data-model/Source/E2EIdentity/MLSConversationVerificationStatusUpdater.swift
```diff
@@ -44,7 +44,9 @@ public class MLSConversationVerificationStatusUpdater: MLSConversationVerificati
 
     // MARK: - Public interface
 
-    public func updateAllStatuses() async throws {
+    public func updateAllStatuses() async {
+        WireLogger.ear.info("updating all MLS conversations verification status")
```
netbe: is it the right logger?

netbe (approved):
just logger to be changed

wire-ios-data-model/Source/E2EIdentity/MLSConversationVerificationStatusUpdater.swift
David-Henner: which logger would you use instead?

David-Henner: ah no I see my mistake

David-Henner: good catch

typfel (approved):

PR #1153, opened 2024-03-20
fix: flaky test in MLSServiceTests

David-Henner:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

`test_CommitPendingProposals_ForSubconversation()` is flaky 

### Causes 

we don't wait for all tasks in the task group of `commitPendingProposals`

### Solutions

wait for all tasks

### DIFF
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -1382,9 +1382,7 @@ public final class MLSService: MLSServiceInterface {
                         logger.error("failed to commit pending proposals: \(String(describing: error))")
                     }
                 }
-
             }
-            await taskGroup.waitForAll()
         }
     }
 
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -1343,15 +1343,11 @@ public final class MLSService: MLSServiceInterface {
         }
 
         WaitingGroupTask(context: context) { [self] in
-            do {
-                try await commitPendingProposals()
-            } catch {
-                WireLogger.mls.error("Failed to commit pending proposals: \(String(describing: error))")
-            }
+            await commitPendingProposals()
         }
     }
 
-    func commitPendingProposals() async throws {
+    func commitPendingProposals() async {
         guard context != nil else {
             return
         }
@@ -1386,6 +1382,8 @@ public final class MLSService: MLSServiceInterface {
                         logger.error("failed to commit pending proposals: \(String(describing: error))")
                     }
                 }
+
+                await taskGroup.waitForAll()
             }
         }
     }
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -1383,8 +1383,8 @@ public final class MLSService: MLSServiceInterface {
                     }
                 }
 
-                await taskGroup.waitForAll()
             }
+            await taskGroup.waitForAll()
         }
     }
 
--- a/wire-ios-data-model/Tests/MLS/MLSServiceTests.swift
+++ b/wire-ios-data-model/Tests/MLS/MLSServiceTests.swift
@@ -1115,7 +1115,7 @@ final class MLSServiceTests: ZMConversationTestsBase, MLSServiceDelegate {
         }
 
         // When
-        try await self.sut.commitPendingProposals()
+        await self.sut.commitPendingProposals()
 
         // Then we asked for the subgroup id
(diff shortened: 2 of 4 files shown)


wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1386,6 +1382,8 @@ public final class MLSService: MLSServiceInterface {
                         logger.error("failed to commit pending proposals: \(String(describing: error))")
                     }
                 }
+
+                await taskGroup.waitForAll()
```
markusfassbender: Hmm I am not really sure how to understand this. In the documentation it says for `withTaskGroup(of:returning:body:)`:
> A group waits for all of its child tasks to complete or be canceled before it returns. After this function returns, the task group is always empty.

So why is this `await taskGroup.waitForAll()` needed?

caldrian: 1. I agree with Markus, the `waitForAll()` is not required, if there are no more statements in the block afterwards
2. More importantly, the `waitForAll()` is executed on each iteration of the loop, right after the `.addTask {}` call. So either it is placed in the wrong line here or there is no point in having a task group at all, since there will not be any parallel execution (`waitForAll` right after `add`)

caldrian (approved):

wire-ios-data-model/Source/MLS/MLSService.swift
David-Henner: 1. In theory, I agree with what the documentation says. But in practice it doesn't seem to work like that: without waiting for all tasks, the tasks simply do not complete before `await commitPendingProposals()` finishes executing. [This example](https://www.avanderlee.com/concurrency/task-groups-in-swift/) also suggests that to collect all values returned by a task group, one musts `await` for them all to return within `withTaskGroup(of:returning:body:)`.

2. Good catch. I meant to do that outside of the loop

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1386,7 +1382,9 @@ public final class MLSService: MLSServiceInterface {
                         logger.error("failed to commit pending proposals: \(String(describing: error))")
                     }
                 }
+
             }
+            await taskGroup.waitForAll()
```
typfel: Oh I thought a `TaskGroup` was always waiting for all tasks to complete before returning.

"A group waits for all of its child tasks to complete or be canceled before it returns. After this function returns, the task group is always empty."

https://developer.apple.com/documentation/swift/withtaskgroup(of:returning:body:)

wire-ios-data-model/Source/MLS/MLSService.swift
markusfassbender: @David-Henner do you have a way to reproduce the flakiness to validate?
When I run the test 200 times it always passes, if the line `await taskGroup.waitForAll()` exists or not. So maybe there are other conditions that make it flake, but I would question this solution.

wire-ios-data-model/Source/MLS/MLSService.swift
johnxnguyen: it doesn't wait unless you invoke `waitForAll()`?

wire-ios-data-model/Source/MLS/MLSService.swift
David-Henner: If I run the test in isolation it systematically fails

markusfassbender: For me it doesn't 🤯
```
Test Case '-[WireDataModelTests.MLSServiceTests test_CommitPendingProposals_ForSubconversation]' passed (0.060 seconds).
Test Suite 'MLSServiceTests' passed at 2024-03-21 10:51:29.574.
	 Executed 1 test, with 0 failures (0 unexpected) in 0.060 (0.060) seconds
Test Suite 'WireDataModelTests.xctest' passed at 2024-03-21 10:51:29.574.
	 Executed 1 test, with 0 failures (0 unexpected) in 0.060 (0.060) seconds
```

Did you try to reset your simulator and clean derived data and so on? Do you have any other setup than me, like an Intel processor instead of ARM based chip?

wire-ios-data-model/Source/MLS/MLSService.swift
David-Henner: discussed with @markusfassbender and updated https://github.com/wireapp/wire-ios/pull/1153/commits/40b46ed2b876d8adcba28f8d9a5ca591d6a5d88e

wire-ios-data-model/Source/MLS/MLSService.swift
David-Henner: discussed with @markusfassbender and updated https://github.com/wireapp/wire-ios/pull/1153/commits/40b46ed2b876d8adcba28f8d9a5ca591d6a5d88e

markusfassbender (approved):

PR #1175, opened 2024-03-26
chore: enable datadog for C builds

netbe:
# What's new in this PR?

### Issues

In order to test the apps with QA and in playtest, we temporary enable datadog.


caldrian (approved):

markusfassbender:
Is there a better solution than changing the build assets every release again? Maybe we can just pass a env var to the build command or something?

typfel (approved):

netbe:
> Is there a better solution than changing the build assets every release again? Maybe we can just pass a env var to the build command or something?

I mean a better solution would be great, I'll create a ticket to look into it

PR #1183, opened 2024-03-27
fix: various issues with encrypted assets WPB-7249 WPB-7257 WPB-7251

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7249" title="WPB-7249" target="_blank">BugWPB-7249</a>  [iOS] Can't open files after bring app back to foreground
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR fixes three issues:
### [[iOS] Can't open files after bring app back to foreground](https://wearezeta.atlassian.net/browse/WPB-7249)
**Cause:** when accessing file assets, we create a temporary url to the decrypted file data in a temp directory. When the app goes to the background, this entire temp directory is deleted. When we open the app again and try to view the file, the old url will be broken, so the file will be decrypted again and stored temporary in the temp directory. However, the temp directory was deleted so we can't store the decrypted file. Because of this, it appears to the UI that there is no file data.

**Fix:** after wiping the temporary directory, create it again so that new files can be written to it.

### [[iOS] Request loop on /assets](https://wearezeta.atlassian.net/browse/WPB-7257)
**Cause:** when downloading the link preview image, we store the encrypted data directly in the file cache. When the UI wants to load the image, we only try to fetch the unencrypted data from the file cache. Since it doesn't exist, if appears as if we have no image data for the link preview and will try to download it again, causing the loop.

**Fix:** when fetching the asset data, fetch the encrypted asset data and decrypt it before returning it back to the UI.

### [[iOS] Viewing old assets freezes and crashes the app](https://wearezeta.atlassian.net/browse/WPB-7251)
**Cause:** Assets downloaded prior to the update with encrypted assets will have unencrypted versions stored in the file cache. After updating, a change to how the processing state for each asset was made where an asset is considered in the preprocessing state if it does not have its encrypted version in the file cache. However, older assets don't have their encrypted versions (only the unencrypted version), so these assets were erroneously considered to be "preprocessing". This in turn caused them to go through the assets preprocessor indefinitely, which froze the app.

**Fix:** redefine a "preprocessing" asset as one that is not uploaded and doesn't have an encrypted version. This allows older assets to fail the condition because even if they don't have an encrypted version, they are already uploaded.

#### How to Test

See the attached tickets.

----

----

----
##### References


wire-ios-data-model/Source/Model/Message/FileAssetCache.swift
```diff
@@ -975,11 +966,21 @@ private struct FileCache: Cache {
...
+            do {
+                // Delete the entire cache.
+                try FileManager.default.removeItem(at: cacheFolderURL)
+            } catch {
+                logger.error("failed to wipe cache: \(error)")
+            }
+        }
+
+        // Create it again so we can write files to it.
+        FileManager.default.createAndProtectDirectory(at: cacheFolderURL)
```
netbe: in case the remove the cache folder fail, would that method fail too?

johnxnguyen: I'm not sure, but if it does it may crash since there's a fatal inside. I'll make a throwing version and throw the errors up

johnxnguyen: done

netbe (approved):

wire-ios-utilities/Source/FileManager+Protection.swift
```diff
@@ -38,24 +47,24 @@ extension FileManager {
...
         }
 
-        // Make sure it's not accessible until first unlock
-        self.setProtectionUntilFirstUserAuthentication(url)
+        // Make sure it's not accessible until first unlock.
+        try setProtectionUntilFirstUserAuthentication(url)
 
-        // Make sure this is not backed up:
+        // Make sure this is not backed up.
         try? url.excludeFromBackup()
```
typfel: Shouldn't we let this throw now?

```suggestion
        try url.excludeFromBackup()
```

wire-ios-data-model/Source/Model/AccountStore.swift
```diff
@@ -41,7 +41,7 @@ public final class AccountStore: NSObject {
     public required init(root: URL) {
         directory = root.appendingPathComponent(AccountStore.directoryName)
         super.init()
-        FileManager.default.createAndProtectDirectory(at: directory)
+        try! FileManager.default.createAndProtectDirectory(at: directory)
```
typfel, line 44: Should we create ticket for cleaning up all these `try!`?

wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage.swift
```diff
@@ -518,9 +524,9 @@ extension ZMAssetClientMessage: AssetMessage {
     public var processingState: AssetProcessingState {
         let assets = self.assets
 
-        // There is an asset that needs to be encrypted.
+        // There is an asset that still needs encrypting and uploading.
         if assets.contains(where: {
-            !$0.hasEncrypted
+            !$0.isUploaded && !$0.hasEncrypted
```
typfel, line 529: So this is the fix for the app freeze due to preprocessing which never ends.

wire-ios-data-model/Source/Model/Message/ZMClientMessage+LinkPreview.swift
```diff
@@ -78,13 +78,51 @@ extension ZMClientMessage {
         NotificationInContext(name: ZMClientMessage.linkPreviewImageDownloadNotification, context: moc.notificationContext, object: self.objectID).post()
     }
 
-    public func fetchLinkPreviewImageData(with queue: DispatchQueue, completionHandler: @escaping (_ imageData: Data?) -> Void) {
-        guard let cache = managedObjectContext?.zm_fileAssetCache else { return }
-        let originalKey = FileAssetCache.cacheKeyForAsset(self, format: .original)
-        let mediumKey = FileAssetCache.cacheKeyForAsset(self, format: .medium)
+    public func fetchLinkPreviewImageData(
```
typfel, line 81: This is the fix for the request loop

typfel (changes requested):

wire-ios-data-model/Source/Model/AccountStore.swift
markusfassbender: Before it has crashed inside the func with `fatal`, so I think it's better now to have it visible. But yeah please let's try to avoid these force casts via refactoring ticket 👍

markusfassbender (approved):

typfel (approved):
Since, I'm not available in the morning I'll approve this now.

wire-ios-data-model/Source/Model/AccountStore.swift
johnxnguyen: yeah I want to get rid of them but didn't want to do it in this PR. I can make a general ticket to clean up all force trys

wire-ios-data-model/Source/Model/Message/ZMClientMessage+LinkPreview.swift
johnxnguyen: Correct

wire-ios-utilities/Source/FileManager+Protection.swift
johnxnguyen: yep

wire-ios-data-model/Source/Model/Message/ZMAssetClientMessage.swift
johnxnguyen: Correct

wire-ios-data-model/Source/Model/Message/FileAssetCache.swift
```diff
@@ -975,11 +978,17 @@ private struct FileCache: Cache {
...
+    func wipeCaches() throws {
         logger.debug("wiping cache")
-        _ = try? FileManager.default.removeItem(at: cacheFolderURL)
+        if FileManager.default.fileExists(atPath: cacheFolderURL.path) {
+            // Delete the entire cache.
+            try FileManager.default.removeItem(at: cacheFolderURL)
+        }
+
+        // Create it again so we can write files to it.
+        try FileManager.default.createAndProtectDirectory(at: cacheFolderURL)
```
johnxnguyen, line 991: @typfel this is the fix for assets not being available after BG -> FG

wire-ios-data-model/Source/Model/AccountStore.swift
johnxnguyen: created a ticket https://wearezeta.atlassian.net/browse/WPB-7331

PR #1200, opened 2024-03-29
fix: update certificate - WPB-7268

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7268" title="WPB-7268" target="_blank">BugWPB-7268</a>  Update certificate popup failure not showing because of presented alert
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

Errors while updating certificate after the popup are not shown because update certificate popup is displayed.

### Solutions

Wait for alert to be dismissed before showing alert with error.


wire-ios-request-strategy/Sources/E2EIdentity/EnrollE2EICertificateUseCase.swift
```diff
@@ -90,7 +91,9 @@ public final class EnrollE2EICertificateUseCase: EnrollE2EICertificateUseCasePro
         do {
             try await e2eiRepository.fetchTrustAnchor()
         } catch {
+            // if you fail here, why go further
             logger.warn("failed to register trust anchor: \(error.localizedDescription)")
+            throw error
```
netbe: @KaterinaWire what do you think? should go forward if this fails here or throw?

KaterinaWire: The initial agreement was to not throw an error and to continue enrollment. Android doesn't throw an error either.

netbe: ok I saw this fix too in https://github.com/wireapp/wire-ios/pull/1208

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
```diff
@@ -174,39 +179,45 @@ final class E2EINotificationActionsHandler: E2EINotificationActions {
...
-    private let durationFormatter: DateComponentsFormatter = {
-        let formatter = DateComponentsFormatter()
-        formatter.allowedUnits = [.day, .hour, .minute]
-        formatter.unitsStyle = .abbreviated
-        return formatter
-    }()
+    @MainActor
+    private func presentScreen(viewController: UIViewController) {
+        let vc = UIApplication.shared.topmostViewController(onlyFullScreen: false) ?? targetVC
+        vc.present(viewController, animated: true)
```
markusfassbender, line 188: question: is this the wanted behaviour that we present just on top of everything?

My high level idea would be rather to replace the current window with a simple new screen, where the user has no possibilities for any interactions, so to say a different area of the app. Once everything is validated again, we would move him back to where he left.

markusfassbender (approved):

KaterinaWire (approved):

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/E2EINotificationActionsHandler.swift
netbe: it's better than to fail presenting the view.

We have something similar to your idea that we could use `BlockerViewController` with the screencurtain (UIWindow) though it would differ from what design provided.

I'll keep as it is for now

PR #1218, opened 2024-04-04
refactor: ColorPickerController to SwiftUI

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we refactor `ColorPickerController`  from `UIKit` to `SwiftUI`. 

----
##### References


wire-ios/Wire-iOS/Sources/UserInterface/Settings/ColorPickerView.swift
```diff
@@ -0,0 +1,83 @@
...
+
+struct ListBackgroundStyleModifier: ViewModifier {
+    func body(content: Content) -> some View {
+        if #available(iOS 16.0, *) {
+            content.scrollContentBackground(.hidden)
+        } else {
+            content.background(Color(SemanticColors.View.backgroundDefault))
+        }
+    }
+}
```
markusfassbender: Can we make the view modifier `private` and it can be simply replaced when iOS 16 is the minimum deployment target?
`.scrollContentBackground(.hidden)`

wire-ios/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Account.swift
```diff
@@ -284,7 +285,24 @@ extension SettingsCellDescriptorFactory {
...
+                    get: {
+                        AccentColor(rawValue: zmAccentColor ?? .min)
+                    },
+                    set: { newColor in
+                        ZMUserSession.shared()?.perform {
+                            let defaultZMAccentColor = ZMAccentColor.default
+                            ZMUser.selfUser()?.accentColorValue = newColor?.rawValue ?? .min
+                        }
+                    }
+                )
```
markusfassbender: I looked at this in detail and don't get why we need a `Binding` at all. The SwiftUI view just receives a value and has a closure, so we can use it directly.
Hope it's fine but not to pollute your PR too much, I've made a new branch with my suggestions: https://github.com/wireapp/wire-ios/pull/1416

wire-ios/Wire-iOS/Sources/UserInterface/Settings/ColorPickerView.swift
```diff
@@ -0,0 +1,83 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import SwiftUI
+import WireCommonComponents
+import WireSyncEngine
+
+struct ColorPickerView: View {
```
johnxnguyen: suggestion: `AccentColorPicker`

wire-ios/Wire-iOS/Sources/UserInterface/Settings/ColorPickerView.swift
```diff
@@ -0,0 +1,83 @@
...
+    @State var selectedColor: AccentColor?
+
+    let colors: [AccentColor]
+    var onColorSelect: ((AccentColor) -> Void)?
+
+    private let colorViewSize: CGFloat = 28
+    private let colorViewCornerRadius: CGFloat = 14
+
+    var body: some View {
+        NavigationView {
```
johnxnguyen: Why do we have a navigation view?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/ColorPickerView.swift
```diff
@@ -0,0 +1,83 @@
...
+                    Text(color.name)
+
+                    Spacer()
+
+                    // Checkmark view
+                    if selectedColor == color {
+                        Image(systemName: "checkmark")
+                            .foregroundColor(Color(SemanticColors.Icon.foregroundDefaultBlack))
+                    }
+                }
```
johnxnguyen: suggestion: move this to a private view builder function, something like:

```swift
@ViewBuilder
private func cell(for color: AccentColor) -> some View {
  // ...
}
```

In my opinion this makes it more readable when glancing at the `body`:

```swift
List(colors, id: \.self) { color in
  cell(for: color)
    .onTapGesture { ... }
}
```

wire-ios/Wire-iOS/Sources/UserInterface/Settings/ColorPickerView.swift
```diff
@@ -0,0 +1,83 @@
...
+                .listRowBackground(Color(SemanticColors.View.backgroundUserCell))
```
johnxnguyen: Should this modifier go on the cell or the list?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/ColorPickerView.swift
```diff
@@ -0,0 +1,83 @@
...
+}
+
+struct ListBackgroundStyleModifier: ViewModifier {
+    func body(content: Content) -> some View {
+        if #available(iOS 16.0, *) {
+            content.scrollContentBackground(.hidden)
+        } else {
+            content.background(Color(SemanticColors.View.backgroundDefault))
+        }
+    }
```
johnxnguyen: Is this only for lists? Perhaps we should move it to separate file and add a doc comment.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPickerController.swift
```diff
@@ -16,217 +16,38 @@
...
 
-        var color: AccentColor? {
-            didSet {
-                if let color {
-                    colorView.backgroundColor = color.uiColor
+        super.init(rootView: ColorPickerView(selectedColor: initialSelectedColor, colors: allAccentColors, onColorSelect: { selectedColor in
+            selectedAccentColor.wrappedValue = selectedColor
+            if let colorIndex = allAccentColors.firstIndex(of: selectedColor) {
+                ZMUserSession.shared()?.perform {
+                    ZMUser.selfUser()?.accentColorValue = allAccentColors[colorIndex].rawValue
```
johnxnguyen: Can we avoid the closure and just pass the binding to the picker?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/ColorPickerView.swift
agisilaos: Done! https://github.com/wireapp/wire-ios/commit/ad8e341bfc0bd19a13f830d305cbe3905436a55b

wire-ios/Wire-iOS/Sources/UserInterface/Settings/ColorPickerView.swift
agisilaos: So we can provide the navigation title with the correct font size. There isn't another workaround for this atm. If we don't do it this way the font size for the navigation bar title is too small.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/ColorPickerView.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/e96b58d70b88e6a8fb4fbafbbe3da6cd5dc0f37b

wire-ios/Wire-iOS/Sources/UserInterface/Settings/ColorPickerView.swift
agisilaos: Yep it's only for lists. Done: https://github.com/wireapp/wire-ios/commit/c4aa244cd88c6a297b02ba691379dab5137ac8ce

wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPickerController.swift
agisilaos: @johnxnguyen Not sure I follow. Can you please expand on the idea?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/ColorPickerView.swift
agisilaos: I think that's also done.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/ColorPickerView.swift
johnxnguyen: I suggest then to add a comment explaining this.

agisilaos: Done: https://github.com/wireapp/wire-ios/commit/310ae500a563d65eb3ecb9982330f024d6184876

wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPickerController.swift
johnxnguyen: I think the code changed, but when I reviewed it there was a `Binding` available that you could have passed to the picker. 

`AccentColorPicker` is a control, and typically (according to Apple) controls use bindings to get and set values rather than closures. See for instance the [`init` for `ColorPicker`](https://developer.apple.com/documentation/swiftui/colorpicker/init(selection:supportsopacity:label:)-1ijcv):

```swift
init(
    selection: Binding<Color>,
    supportsOpacity: Bool = true,
    @ViewBuilder label: () -> Label
)
```

I think it would be nice to stick with this convention and have a similar init for `AccentColorPicker`:

```swift
init(selection: Binding<AccentColor>)
```

markusfassbender: I think it depends a bit what one wants to achieve in SwiftUI:
- Button for example just calls a closure for [tap action](https://developer.apple.com/documentation/swiftui/button/init(_:action:)-1asy).
- To communicate a value out of the view a `Binding` seems the right choice.

So in this case I agree we can use a `Binding` here for the view itself. In our case with this architecture to go through a factory currently, we "misuse" the `AccentColorPickerController` as wrapper for business logic, so I suggest the binding can be created there and pass and receive the value to/from the view.

markusfassbender (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPicker.swift
```diff
@@ -0,0 +1,81 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import SwiftUI
+import WireCommonComponents
+import WireSyncEngine
+
+struct AccentColorPicker: View {
+
+    @State var selectedColor: AccentColor
```
johnxnguyen: I think state should be private.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPicker.swift
```diff
@@ -0,0 +1,81 @@
...
+
+    private let colorViewSize: CGFloat = 28
+    private let colorViewCornerRadius: CGFloat = 14
+
+    var body: some View {
+        // The NavigationView is needed here to properly set the font size for the navigation bar title.
+        // Currently, there's no other way to set the font size of the navigation bar title to the desired value.
+        // Without the NavigationView, the title would be too small.
+        NavigationView {
+            List(colors, id: \.self) { color in
```
johnxnguyen: Also if we don't need to show a subset of colors, I suggest to use `AccentColor.allCases`.

johnxnguyen (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPickerController.swift
agisilaos: I added a comment why Binding is hard to use, so I'm marking this comment as resolved.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPicker.swift
agisilaos: I don't think it can be private since  @State properties drive the view’s state and must be mutable and accessible within the view’s body, thus cannot be private.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPicker.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/UserInterface/Settings/ColorPickerView.swift
agisilaos: Moved it to its own file therefore can't be private atm. It's on my list to replace it when IOS 16 is our minimum version.

PR #1220, opened 2024-04-04
fix: 1-1s not resolved after client deletion WPB-7112

johnxnguyen:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When the self client removes one of the self users other clients, it should but does not resolve 1-1 conversations.

### Solutions

Invoke the use case to resolve 1-1s.

#### Test Coverage

- Updated unit test

#### How to Test

1. Log in user A with an MLS client on iOS
2. Log in with user A on a non MLS client on another device
3. Be logged in with user B on a single MLS client.
4. A has a proteus 1-1 with user B
5. While A's MLS client is open, log out of A's non MLS client
6. Assert that the proteus 1-1 with user B is migrated to MLS.

----

----

----
##### References


wire-ios-sync-engine/Source/Synchronization/Strategies/UserClientEventConsumer.swift
```diff
@@ -25,38 +25,50 @@ import Foundation
...
+    private let useCaseFactory: UseCaseFactoryProtocol
 
-    public init (managedObjectContext: NSManagedObjectContext,
-                 clientRegistrationStatus: ZMClientRegistrationStatus,
-                 clientUpdateStatus: ClientUpdateStatus) {
+    public init(
+        managedObjectContext: NSManagedObjectContext,
+        clientRegistrationStatus: ZMClientRegistrationStatus,
+        clientUpdateStatus: ClientUpdateStatus,
+        useCaseFactory: UseCaseFactoryProtocol
```
typfel: Why do we need a use case factory? Why can't we inject `resolveOneOnOneUseCase` directly?

markusfassbender (approved):

wire-ios-sync-engine/Source/Synchronization/Strategies/UserClientEventConsumer.swift
johnxnguyen: yeah good point

johnxnguyen: ah I think because I was thinking the use case should be created and invoked once, rather than injecting an instance to be invoked repeatedly. What's the best practice?

markusfassbender: Not completely sure about the best practice, but would agree with @typfel here. The use case should be a light-weight `struct` or `class` without states. So it should be save to retain it for long time also and pass around.

Additionally then I would question `UseCaseFactoryProtocol`, the factory can be used to create use cases with their protocols, but having another protocol here feels for me like a too complex construction.

markusfassbender (approved):

wire-ios-sync-engine/Source/Synchronization/Strategies/UserClientEventConsumer.swift
```diff
@@ -25,38 +25,50 @@ import Foundation
...
+        _ events: [ZMUpdateEvent],
+        liveEvents: Bool,
+        prefetchResult: ZMFetchRequestBatchResult?
+    ) async {
         for event in events {
-            await processUpdateEvent(event)
+            do {
+                try await processUpdateEvent(event)
+            } catch {
+                WireLogger.updateEvent.error("failed to process user client event: \(event): \(error)")
```
netbe: I would expose error to logfile on prod, what do you think something like:
```suggestion
                WireLogger.updateEvent.error("failed to process user client event: \(event.safeForLogging): \(error)", attributes: .safePublic)
```

wire-ios-sync-engine/Source/Synchronization/Strategies/UserClientEventConsumer.swift
typfel: I'm on the same line as @markusfassbender,  since the use cases doesn't hold any state it fine to hold on to it an re-use it.

johnxnguyen: Ok got it, thanks!

johnxnguyen: done

typfel (approved):

markusfassbender (approved):

PR #1228, opened 2024-04-05
fix: passcode during call - reuse `LAContext` [task][WPB-7111]

markusfassbender:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7111" title="WPB-7111" target="_blank">BugWPB-7111</a>  [iOS] C builds asks for passcode during call
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

> [!NOTE]
> Merges into feature branch #1188

This also solves the ticket https://wearezeta.atlassian.net/browse/WPB-87 ✅

# What's new in this PR?

### Issues

So now `LAContext` is accessed through the `AuthenticationContext` which makes sure we that we only have one instance around and the instance is reused in order not to prompt the user to enter the passcode. The `LAContext` object is released whenever the app goes into the background. Previously an LAContext was created implicitly when calling certain methods.

This change doesn't change any properties of how we can access the database since the `LAContext` is still only used to unlock the database keys and after that we hold on to the database keys and this hasn't changed.

This solves the existing issue, that the user had to enter the passcode two times on app startup - one time for the custom AppLock and one time for the access to the keychain.

Unfortunately the existing issue [WPB-7111] is not solved and the prompt still appears during calls. I will make another follow up to address this issue.

#### Test Coverage (Optional)

#### How to Test

- Build app that enforces encryption at rest:
  - In `session_manager.json` change the value to `"encryptionAtRestEnabledByDefault": true`
  - In `Security-Flags.xcconfig` change the value to `FORCE_ENCRYPTION_AT_REST_ENABLED=1`
- Login with a user
- The prompt for authentication should appear only once

----

----

----
##### References

[WPB-7111]: https://wearezeta.atlassian.net/browse/WPB-7111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wire-ios/Wire-iOS/Sources/UserInterface/Calling/Haptic Feedback/CallHapticsController.swift
```diff
@@ -83,7 +83,7 @@ final class CallHapticsController {
 
     private func updateParticipantsVideoStateList(_ newParticipants: [CallParticipant]) {
         let newVideoStates = createVideoStateMap(using: newParticipants)
-        Log.haptics.debug("updating video state map: \(newVideoStates), old: \(videoStates)")
+        Log.haptics.debug("updating video state map")
```
markusfassbender, line 86: Those video changes are very large objects in the log with hundreds of lines of the core data object array. To make the debug life easier, I have removed the states from the log.

wire-ios-data-model/Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -236,6 +241,7 @@ public final class AppLockController: AppLockType {
 
 // MARK: - TEST ONLY!
 
+@_spi(AppLockControllerState)
```
markusfassbender, line 244: https://github.com/apple/swift/blob/main/docs/ReferenceGuides/UnderscoredAttributes.md#_spispiname

wire-ios-data-model/Source/Utilis/EAR/PrivateEARKeyDescription.swift
```diff
@@ -62,26 +67,23 @@ public final class PrivateEARKeyDescription: BaseEARKeyDescription, KeychainItem
...
+            label: Constant.labelPrivatePrimary,
             context: context
         )
     }
 
     static func secondaryKeyDescription(accountID: UUID) -> PrivateEARKeyDescription {
         return PrivateEARKeyDescription(
             accountID: accountID,
-            label: "secondary-private"
+            label: Constant.labelPrivatePrimary
```
markusfassbender: This is a bug!

wire-ios-data-model/Source/Utilis/EAR/EARKeyRepository.swift
```diff
@@ -34,6 +35,7 @@ protocol EARKeyRepositoryInterface {
 
 }
 
+/// Caches keys for reuse and avoid prompting the user to authenticate for each key access.
```
johnxnguyen, line 38: I wonder how useful the key cache is now if we are instead caching the authenticated context?

wire-ios-data-model/Source/Utilis/EAR/EARService.swift
```diff
@@ -184,12 +190,12 @@ public class EARService: EARServiceInterface {
...
 
         let enableEAR: (NSManagedObjectContext) throws -> Void = { [weak self] context in
-            guard let `self` = self else { return }
+            guard let self else { return }
 
             do {
                 try self.deleteExistingKeys()
                 try self.generateKeys()
-                let databaseKey = try self.fetchDecyptedDatabaseKey(context: LAContext())
+                let databaseKey = try self.fetchDecyptedDatabaseKey()
```
johnxnguyen: typo "Decypted" instead of "Decryptded"

typfel (commented):
Writing this down to double check for myself that I've understood the change:

So now `LAContext` is accessed through the `AuthenticationContext` which makes sure we that we only have one instance around. The LAContext is dellocated whenever the app goes into the background. Previously an `LAContext` was created implicitly when calling certain methods.

This change doesn't change any properties of how we can access the database since the `LAContext` is still only used to unlock the database keys and after that we hold on to the database keys and this hasn't changed.

johnxnguyen (commented):
Looks good so far

wire-ios-data-model/Source/Utilis/EAR/EARKeyRepository.swift
markusfassbender: Good point! I have played with the sample project from Apple and unfortunately after a few minutes even a cached `LAContext` prompts again. So it would be counteracting to our wanted behavior if we would remove the local key cache.
https://developer.apple.com/documentation/localauthentication/accessing-keychain-items-with-face-id-or-touch-id

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -123,7 +123,8 @@ public final class ZMUserSession: NSObject {
 
     let earService: EARServiceInterface
 
-    public var appLockController: AppLockType
+    public internal(set) var appLockController: AppLockType
+    private let contextStorage = LAContextStorage()
```
markusfassbender, line 127: Can I assume that the storage is correctly deallocated when the user logsout?
Otherwise I am not sure if `func tearDown()` is the right choice to release the store, but then the contextStorage needs to be optional.

typfel: Yes, we can assume  that ZMUserSession gets deallocated along with all its objects. I'm not a fan of the tearDown pattern and I hope we can get rid of it.

typfel (approved):

johnxnguyen (approved):

PR #1240, opened 2024-04-08
fix: mls migration after updating from old client WPB-7163

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7163" title="WPB-7163" target="_blank">BugWPB-7163</a>  [iOS] Migration: When migration happens when using old client, group breaks for user on old client even after updating
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When the MLS migration happens while on an old client, the conversation breaks for the user on the old client (expected) and doesn't get fixed after updating to the latest version (not expected)

### Causes

Because the migration happens while the user is still on the old client, the conversation protocol updates are never processed by the client. After the update, the client doesn't sync its conversations and remains in a conversation that still uses `proteus` instead of `mls`

### Solutions

Once the client has registered itself as an MLS client, we should trigger a slow sync to pull in the latest changes to conversations and be aware that the protocol has changed to `mls`. The client will then automatically join the MLS groups as needed.


wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -998,6 +998,10 @@ extension ZMUserSession: ZMSyncStateDelegate {
 
             self?.delegate?.clientRegistrationDidSucceed(accountId: accountId)
         }
+
+        if userClient.hasRegisteredMLSClient {
+            syncStatus.forceSlowSync()
```
typfel, line 1005: This doesn't interfere when registering an MLS client on the initial login because the slow sync is done after registering the client?

typfel (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: suggestion: add a comment here to explain why the slow sync is necessary

johnxnguyen (approved):

netbe (approved):
just add the comment and good to go

PR #1248, opened 2024-04-09
fix: update mls conversation verification status WPB-7404

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7404" title="WPB-7404" target="_blank">BugWPB-7404</a>  [iOS] Adding non E2EI user to a verified conversation didn’t degrade it
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

1. Adding non E2EI user (or with expired certificate) to a verified conversation didn’t degrade it.
2. Removing non E2EI user (or with expired certificate) from the conversation didn’t update the status.

### Causes (Optional)

Observer was setup in the `ZMUserSession` initializer before we got the `e2eiFeature` status, so it was always skipped.

### Solutions

1. Move the observer setting to the `didFinishQuickSync()`.
2. Observe `mlsVerificationStatusChanged` on the `ConversationViewController` and update the conversation title.

----

----

----
##### References

### DIFF
--- a/wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
+++ b/wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
@@ -563,6 +563,7 @@ public final class ZMUserSession: NSObject {
             self.applicationStatusDirectory.clientRegistrationStatus.determineInitialRegistrationStatus()
             self.hasCompletedInitialSync = self.applicationStatusDirectory.syncStatus.isSlowSyncing == false
 
+            self.mlsGroupVerificationStatusObserver.invoke()
             self.cRLsDistributionPointsObserver.startObservingNewCRLsDistributionPoints(
                 from: self.mlsService.onNewCRLsDistributionPoints()
             )
@@ -932,7 +933,6 @@ extension ZMUserSession: ZMSyncStateDelegate {
 
         managedObjectContext.performGroupedBlock { [weak self] in
             self?.notifyThirdPartyServices()
-            self?.observeMLSGroupVerificationStatus()
         }
 
     }
@@ -1037,14 +1037,6 @@ extension ZMUserSession: ZMSyncStateDelegate {
     func checkE2EICertificateExpiryStatus() {
         guard e2eiFeature.isEnabled else { return }
         NotificationCenter.default.post(name: .checkForE2EICertificateExpiryStatus, object: nil)
-    }
-
-    // MARK: - Observers
-
-    private func observeMLSGroupVerificationStatus() {
-        guard e2eiFeature.isEnabled else { return }
-
-        mlsGroupVerificationStatusObserver.invoke()
     }
 
 }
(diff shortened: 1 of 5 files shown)


wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -935,7 +928,10 @@ extension ZMUserSession: ZMSyncStateDelegate {
 
         managedObjectContext.performGroupedBlock { [weak self] in
             self?.notifyThirdPartyServices()
+            self?.observeMLSGroupVerificationStatus()
+            self?.observeNewCRLsDistributionPoints()
```
typfel: I'm not sure that it's good idea to move this to `didFinishQuickSync`, we can potentially discover new CRLs during the quick syncing which would then be lost.

cc @David-Henner

typfel: What bad things happens if we always observe regardless if `e2eiFeature` is enabled or not?

David-Henner: > I'm not sure that it's good idea to move this to `didFinishQuickSync`, we can potentially discover new CRLs during the quick syncing which would then be lost.
> 
> cc @David-Henner

Just to clarify, we can potentially discover new CRL **distribution points** during the quick sync. Then we fetch the CRLs from the new distribution points.

We should start observing the new distribution points as soon as possible. During the quick sync, we may forward mls events to CC and CC may return new distribution points. If we start observing after the quick sync then we'd miss these DPs.

> What bad things happens if we always observe regardless if e2eiFeature is enabled or not?

Nothing, I guess CC would never discover any CRL distribution points

KaterinaWire: I moved back the observing of CRL distribution points.

David-Henner (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -935,7 +932,9 @@ extension ZMUserSession: ZMSyncStateDelegate {
 
         managedObjectContext.performGroupedBlock { [weak self] in
             self?.notifyThirdPartyServices()
+            self?.observeMLSGroupVerificationStatus()
```
typfel: Doesn't the same thing also apply to verification status? What if the epoch changes during the quick sync? Wouldn't we miss the epoch change and therefore not update the verifications status?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1039,6 +1038,15 @@ extension ZMUserSession: ZMSyncStateDelegate {
         guard e2eiFeature.isEnabled else { return }
         NotificationCenter.default.post(name: .checkForE2EICertificateExpiryStatus, object: nil)
     }
+
+    // MARK: - Observers
+
+    private func observeMLSGroupVerificationStatus() {
+        guard e2eiFeature.isEnabled else { return }
```
caldrian: Don't we waste energy with this check. It looks like a simple property, but in fact it uses the feature repository to access core data. Can it really change during the runtime of the app?

KaterinaWire: removed

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
KaterinaWire: Moved it to init() based on our discussion in the office.

typfel (approved):

PR #1257, opened 2024-04-10
fix: changelog and upload artifacts in github workflows - WPB-8615

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8615" title="WPB-8615" target="_blank">BugWPB-8615</a>  Changelog generation on C builds
  </td></table>
  <br />
 

### Issue

artifact/upload github action v4 does not allow to upload the same name of artifact in the same run which breaks the job on C builds (the notification part does not work then)

* add specific name for artifact upload
* add option to select one build
* generate changelog from proper tag

Screenshot 2024-04-10 at 18 51 02

---

### Checklist

- [x] Title conforms to [semantic commits messages](https://github.com/wireapp/.github#usage).

---


.github/workflows/_reusable_app_release.yml
```diff
@@ -4,6 +4,10 @@ on:
       fastlane_action:
         required: true
         type: string
+      cloud:
+        required: true
+        type: boolean
+        default: true
```
markusfassbender: I think the name confused me, that we want to load the changeling to/from the cloud like AWS?
But I think what is meant is "this s a build to use the wire cloud as backend", right?

Maybe we can use a name like `is_cloud_build`?

netbe: relates to this configuration: https://github.com/wireapp/wire-ios-build-assets/tree/master/Cloud/AppStore

netbe: done

markusfassbender (approved):

caldrian (approved):

PR #1265, opened 2024-04-11
feat: add enable datadog option to C builds github workflow - WPB-8660

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8660" title="WPB-8660" target="_blank">BugWPB-8660</a>  Move datadog enabled to CI
  </td></table>
  <br />
 

### Issue

When testing C builds, we often require datadog to be enabled but need it disabled for release which involve to change the commit of wire-ios-build-assets.

The solution is now to pass info for datadog from github secrets to fastlane and pass it as xcargs in xcodebuild instead of relying on xcconfig values.

### Checklist

See Build Wire C1 3.112.2 (11340)


fastlane/Fastfile
```diff
@@ -541,6 +553,7 @@ class Build
             @git_branch = ENV['GITHUB_REF_NAME']
         end
 
+        @enable_datadog = ENV['ENABLE_DATADOG'] != nil
```
johnxnguyen: What is `@enabled_datadog`? And will it be true of `ENV['ENABLED_DATADOG']` is false?

netbe: it's a property of Build just to wrap the env variable. Good point maybe we need to change it to `ENV['ENABLE_DATADOG'] == 1 || ENV['ENABLE_DATADOG'] == true`

typfel (approved):

netbe:
See both builds:
```
Column 1 Restricted (version: 3.112.2 build: 11366) is ready to test  🚀
DATADOG_ENABLED: false
```

```
Column 1 Restricted (version: 3.112.2 build: 11375) is ready to test  🚀
DATADOG_ENABLED: true
```

johnxnguyen (approved):

PR #1268, opened 2024-04-11
chore: SwiftLint add `yoda_condition`

markusfassbender:
### Issue

Add SwiftLint opt-in rule `yoda_condition`:
https://realm.github.io/SwiftLint/yoda_condition.html

### Testing

- Build
- Run unit test suite

---

### Checklist


wire-ios-data-model/Source/Notifications/ObjectObserverTokens/Helpers/StringKeyPath.swift
```diff
@@ -53,7 +53,7 @@ public final class StringKeyPath: Hashable {
     }
 
     public lazy var decompose: (head: StringKeyPath, tail: StringKeyPath?)? = {
-        if 1 <= self.count {
```
netbe, line 56: this one is really messing with my head

netbe (approved):

wire-ios-data-model/Source/Notifications/ObjectObserverTokens/Helpers/StringKeyPath.swift
markusfassbender: Yes 😆 I first rewrote it to `self.count > 1`, but it must be `self.count >= 1`, but then I thought it's an `Int` so it can be `self.count > 0`, what is actually meant 😬

agisilaos (approved):

PR #1284, opened 2024-04-12
fix: get rid of inconvenient helper functions

caldrian:
Removes Sequence+Async.swift.

---

### Checklist


wire-ios-sync-engine/Source/Use cases/GetE2eIdentityCertificatesUseCase.swift
```diff
@@ -89,14 +94,16 @@ final public class GetE2eIdentityCertificatesUseCase: GetE2eIdentityCertificates
...
-                return (identity, .invalid)
+                validatedIdentities += [(identity, .invalid)]
+                continue
             }
 
             let hasValidDisplayName = identity.x509Identity?.displayName == name
             let hasValidHandle = identity.x509Identity?.handle.contains("\(handle)@\(domain)") ?? false
             let isValid = hasValidDisplayName && hasValidHandle
-            return (identity, isValid ? .valid : .invalid)
+            validatedIdentities += [(identity, isValid ? .valid : .invalid)]
```
markusfassbender: In many places you now use `+= [object]`, I would rather use the `.append(object)` function. But it's rather personal preference.

markusfassbender (approved):
thank you!

agisilaos (approved):

PR #1289, opened 2024-04-15
feat: Send the password to POST - WPB-4641

agisilaos:
<!--do not remove this marker, its needed to replace info when ticket title is updated -->
<!--jira-description-action-hidden-marker-start-->

<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4641" title="WPB-4641" target="_blank"><img alt="Task" src="https://wearezeta.atlassian.net/rest/api/2/universal_avatar/view/type/issuetype/avatar/10818?size=medium" />WPB-4641</a>  [iOS] Send the password to POST 
  </td></table>
  <br />
 

<!--jira-description-action-hidden-marker-end-->
<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove the jira markers to link tickets automatically -->


This PR introduces a significant feature: the ability to add a password to a guest link. This addition is critical for allowing more secure and controlled access to conversations. The implementation of this feature necessitated several modifications and the introduction of new components to accommodate the variability in API support for password-protected links.

## Key Changes:

### Modifications and Additions:

**API Version Compatibility**: Acknowledging the varied support for password protection across different API versions, we've implemented a solution that adapts dynamically. For APIs lacking password support, alternative endpoints are utilized to maintain functionality without compromising the user experience.

### Architectural Enhancements:

**CreateConversationGuestLinkActionHandler**: Serves as the orchestrator for creating password-protected guest links, ensuring the process aligns with API capabilities and user requirements.

**CreateConversationGuestLinkAction**: Directly responsible for the generation of guest links, this component manages the inclusion of passwords based on the API's support level.

**CreateSecuredGuestLinkUseCase**: Dedicated to the scenario where enhanced security through password protection is desired, encapsulating the logic for secure link creation.

**SetAllowGuestsAndServicesActionHandler** and SetAllowGuestsAndServicesAction: These components are integral to setting up the necessary permissions for guests and services

-- 

One of the goals working on this ticket was to move the creation of the wireless link as well as setting guest and services away from the conversation and into the UseCase. Due to time constraint that's not possible atm but it's something @johnxnguyen and I discussed and we can tackle on a later stage during the refactoring. 


### Checklist


wire-ios-data-model/Source/Model/Conversation/Actions/CreateConversationGuestLinkAction.swift
```diff
@@ -0,0 +1,66 @@
...
+    public init?(response: ZMTransportResponse) {
+        switch (response.httpStatus, response.payloadLabel()) {
+        case (403, "invalid-op"?): self = .invalidOperation
+        case (404, "no-conversation-code"?): self = .noCode
+        case (404, "no-conversation"?): self = .noConversation
+        case (409, "guest-links-disabled"?): self = .guestLinksDisabled
+        case (400..<499, _): self = .unknown
+        default: return nil
+        }
+    }
```
johnxnguyen: I think this should be private and moved to the action handler since it concerns parsing the transport response.

wire-ios-data-model/Source/Model/Conversation/Actions/CreateConversationGuestLinkAction.swift
```diff
@@ -0,0 +1,66 @@
...
+
+public struct CreateConversationGuestLinkParameters {
+    public let password: String?
+    public let conversationID: UUID
+
+    public init(password: String?, conversationID: UUID) {
+        self.password = password
+        self.conversationID = conversationID
+    }
+}
```
johnxnguyen: non-blocking: you could also add `password` and `conversationID` as properties to the action and avoid creating a new struct.

wire-ios-data-model/Source/Model/Conversation/Actions/CreateConversationGuestLinkAction.swift
```diff
@@ -0,0 +1,66 @@
...
+
+public final class CreateConversationGuestLinkAction: EntityAction {
```
johnxnguyen: suggestion: use a struct if there is no need for a class.

wire-ios-data-model/Source/Model/Conversation/Actions/SetAllowGuestsAndServicesAction.swift
```diff
@@ -0,0 +1,41 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+public final class SetAllowGuestsAndServicesAction: EntityAction {
```
johnxnguyen: suggestion: use a struct if a class is not necessary.

wire-ios-data-model/Source/Model/Conversation/Actions/SetAllowGuestsAndServicesAction.swift
```diff
@@ -0,0 +1,41 @@
...
+
+    public typealias Result = Void
+    public typealias Failure = Error
```
johnxnguyen: Are we able to define a more specific error, like `CreateConversationGuestLinkError`?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
```diff
@@ -0,0 +1,92 @@
...
+
+    override func handleResponse(_ response: ZMTransportResponse, action: CreateConversationGuestLinkAction) {
+
+        var action = action
+
+        switch response.httpStatus {
+        case 201:
+            guard let payload = response.payload?.asDictionary(),
+                  let data = payload["data"] as? [String: Any],
+                  let uri = data["uri"] as? String else {
```
johnxnguyen, line 78: Generally it's better to define a decodable struct representing the payload, rather than using dictionaries, but I think it's fine to keep it like this since we'll be refactoring this very soon.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
```diff
@@ -0,0 +1,92 @@
...
+
+                let errorInfo = CreateConversationGuestLinkError(response: response)
+                action.fail(with: errorInfo ?? .unknown)
```
johnxnguyen: I think the error you want to return here is `.failedToDecodeResponsePayload`.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
```diff
@@ -0,0 +1,92 @@
...
+            }
+
+            action.succeed(with: uri)
+
+        case 200:
+            guard let payload = response.payload?.asDictionary(),
+                  let uri = payload["uri"] as? String else {
+
+                let errorInfo = CreateConversationGuestLinkError(response: response)
+                action.fail(with: errorInfo ?? .unknown)
```
johnxnguyen: I think the error you want to return here is `.failedToDecodeResponsePayload`.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandlerTests.swift
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    override func tearDown() async throws {
+        stack = nil
+        try coreDataStackHelper.cleanupDirectory()
+        try await super.tearDown()
+    }
+
+    // create tests for the rest of the API versions
```
johnxnguyen: Is this todo?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandlerTests.swift
```diff
@@ -0,0 +1,99 @@
...
+
+        let payload: [AnyHashable: Any] = [
+            "code": "SOME-UNIQUE-CODE",
+            "has_password": false,
+            "key": "sampleKey123",
+            "uri": "https://fakeurlfortest.com"
+        ]
+
+        // WHEN && THEN
+        test_itHandlesSuccess(
```
johnxnguyen: I think we should check the result and assert it's the expected uri.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandlerTests.swift
```diff
@@ -0,0 +1,99 @@
...
+        let statusCode = 400
+
+        let payload: [AnyHashable: Any] = [
+            "code": "SOME-UNIQUE-CODE",
+            "has_password": false,
+            "key": "sampleKey123"
+        ]
+
+        // WHEN && THEN
+        test_itHandlesFailure(.failure(status: statusCode, error: .unknown))
```
johnxnguyen: What does status 400 represent? Is it really an unknown error?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/SetAllowGuestsAndServicesActionHandler.swift
```diff
@@ -0,0 +1,108 @@
...
+        var payload: [String: Any] = [
+            "access": ConversationAccessMode.value(forAllowGuests: action.allowGuests).stringValue,
+            "access_role": accessRoles.map(\.rawValue)
+        ]
+
+        let path: String
+        switch apiVersion {
+        case .v3, .v4, .v5, .v6:
+            guard let domain = conversation.domain.nonEmptyValue ?? BackendInfo.domain else {
+                fatal("no domain associated with conversation, can't make the request")
```
johnxnguyen: Rather than crash the app, we should return an error.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/SetAllowGuestsAndServicesActionHandler.swift
```diff
@@ -0,0 +1,108 @@
...
+        for action: SetAllowGuestsAndServicesAction,
+        apiVersion: APIVersion
+    ) -> ZMTransportRequest? {
+
+        var action = action
+
+        guard
+            let conversation = ZMConversation.existingObject(for: action.conversationID, in: context),
+            let identifier = conversation.remoteIdentifier?.transportString() else {
+            return nil
```
johnxnguyen, line 38: Don't forget to report an error for the action.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/SetAllowGuestsAndServicesActionHandler.swift
```diff
@@ -0,0 +1,108 @@
...
+
+    // MARK: - Request Handling
+
+    override func handleResponse(_ response: ZMTransportResponse, action: SetAllowGuestsAndServicesAction) {
+
+        var action = action
+
+        guard let payload = response.payload,
+              let updateEvent = ZMUpdateEvent(fromEventStreamPayload: payload, uuid: nil) else {
+            action.fail(with: SetAllowGuestsAndServicesError.unknown)
```
johnxnguyen: I suggest to return `.failedToDecodeResponsePayload` error instead of `.unknown`

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/SetAllowGuestsAndServicesActionHandler.swift
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        let success = {
+            action.succeed()
+        }
+
+        Task {
+            await eventProcessor.processConversationEvents([updateEvent])
+            await context.perform { _ = self.context.saveOrRollback() }
+            success()
```
johnxnguyen, line 102: Perhaps the instead of processing the events inside the action handler, the handler can return the update events as the result. Supposing we perform the action in a use case, we could make that use case async and process the events there. Something like:

```swift
struct SetAllowGuestsAndServicesUseCase {

  func invoke(allowGuests: Bool, allowServices: Bool, conversationID: QualifiedID) async throws {
    let action = ...
    let events = try await action.perform()
    await eventProcessor.processConversationEvents(events)
  }

}
```

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
```diff
@@ -47,6 +47,8 @@ public class ConversationRequestStrategy: AbstractRequestStrategy, ZMRequestGene
     let addParticipantActionHandler: AddParticipantActionHandler
     let removeParticipantActionHandler: RemoveParticipantActionHandler
     let updateAccessRolesActionHandler: UpdateAccessRolesActionHandler
+    let createGuestLinkActionHandler: CreateConversationGuestLinkActionHandler
+    let setAllowGuestsAndServicesActionHandler: SetAllowGuestsAndServicesActionHandler
```
johnxnguyen: suggestion: if there's no need to retain them, we can just initialize them directly in the entity sync init below, like other action handlers.

wire-ios-sync-engine/Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -228,30 +217,32 @@ extension ZMConversation {
...
-                self.allowServices = allowServices
+        var action = SetAllowGuestsAndServicesAction(
+            allowGuests: allowGuests,
+            allowServices: allowServices,
+            conversationID: self.objectID
+        )
 
-                // Process `conversation.access-update` event
-                userSession.processUpdateEvents([event])
+        action.perform(in: self.managedObjectContext!.notificationContext) { result in
```
johnxnguyen: suggestion: safely unwrap the context and return an error if needed.

wire-ios-sync-engine/Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -275,8 +266,8 @@ internal struct WirelessRequestFactory {
         }
         return .init(getFromPath: "/conversations/\(identifier)/features/conversationGuestLinks", apiVersion: apiVersion.rawValue)
     }
-
-    static func createLinkRequest(for conversation: ZMConversation, apiVersion: APIVersion) -> ZMTransportRequest {
+    // Optional password in this method as a reference and then pass it all the way up to the cell
```
johnxnguyen: I don't quite understand this comment, could you clarify it please?

johnxnguyen: Do we still need this method if we have the action now?

wire-ios-sync-engine/Source/Use cases/SecuredGuestLinkUseCase.swift
```diff
@@ -0,0 +1,54 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+// sourcery: AutoMockable
+public protocol SecuredGuestLinkUseCaseProtocol {
```
johnxnguyen: suggestion: `CreateConversationGuestLinkUseCaseProtocol`

wire-ios-sync-engine/Source/Use cases/SecuredGuestLinkUseCase.swift
```diff
@@ -0,0 +1,54 @@
...
+                switch result {
+                case .failure(let error):
+                    completion(.failure(error))
+                case .success:
+                    conversation.createWirelessLink(password: password, completion)
+                }
+            }
+        } else {
+            conversation.createWirelessLink(password: password, completion)
+        }
```
johnxnguyen: This appears to duplicate:

```swift
public func updateAccessAndCreateWirelessLink(password: String?, _ completion: @escaping (Result<String, Error>) -> Void)
```

Are you planning to remove the old method?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreateSecureGuestLink/CreateSecureGuestLinkViewController.swift
```diff
@@ -32,6 +32,7 @@ class CreateSecureGuestLinkViewController: UIViewController, CreatePasswordSecur
     weak var delegate: ValidatedTextFieldDelegate?
 
     private lazy var viewModel: CreateSecureGuestLinkViewModel = {
+        // Inject the useCase here, have the userSession, get the factory and inject it
```
johnxnguyen: Is this a todo?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
markusfassbender: Right, though it's easier to do refactorings if you don't need to handle all details but some things like a decodable struct can be just moved around without having one more thread in the head :)

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreateSecureGuestLink/CreateSecureGuestLinkViewModel.swift
```diff
@@ -62,6 +72,31 @@ final class CreateSecureGuestLinkViewModel {
...
     }
 
+    func createSecuredGuestLinkIfValid(conversation: ZMConversation, passwordField: ValidatedTextField, confirmPasswordField: ValidatedTextField) {
+        guard validatePassword(for: passwordField, against: confirmPasswordField) else {
+            delegate?.viewModel(self, didFailToValidatePasswordWithReason: "Password validation failed.")
+            return
+        }
+
+        let password = passwordField.text ?? ""
+        UIPasteboard.general.string = password
```
markusfassbender: Why is it needed to write the password to the user's system wide pasteboard?

wire-ios-data-model/Source/Model/Conversation/Actions/SetAllowGuestsAndServicesAction.swift
agisilaos: Done!

wire-ios-data-model/Source/Model/Conversation/Actions/CreateConversationGuestLinkAction.swift
agisilaos: Done!

wire-ios-data-model/Source/Model/Conversation/Actions/CreateConversationGuestLinkAction.swift
agisilaos: Done!

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/SetAllowGuestsAndServicesActionHandler.swift
agisilaos: Done!

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/SetAllowGuestsAndServicesActionHandler.swift
agisilaos: Done!

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/SetAllowGuestsAndServicesActionHandler.swift
agisilaos: Done!

wire-ios-sync-engine/Source/Data Model/Conversation+AccessMode.swift
agisilaos: We don't I deleted it as well as their respected tests. https://github.com/wireapp/wire-ios/commit/de595db36a92c05a68dd639e47b26fb27d8e869c

wire-ios-sync-engine/Source/Use cases/SecuredGuestLinkUseCase.swift
agisilaos: Done! https://github.com/wireapp/wire-ios/commit/24ab6eb7b12bc68847bfc881a07c455d4e46d384

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreateSecureGuestLink/CreateSecureGuestLinkViewController.swift
agisilaos: Nope, it's already done, so I'm planning to remove the comment.

wire-ios-sync-engine/Source/Use cases/SecuredGuestLinkUseCase.swift
agisilaos: I did, as well as their respected tests.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreateSecureGuestLink/CreateSecureGuestLinkViewModel.swift
agisilaos: We had this discussion 1:1, it's according to design.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreateSecureGuestLink/CreateSecureGuestLinkViewController.swift
agisilaos: Done!

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
agisilaos: Done!

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/SetAllowGuestsAndServicesActionHandler.swift
johnxnguyen: Also, if all you need is the conversation id, why not pass that in the action rather than going through the extra step of fetching the conversation?

johnxnguyen: Actually I see that you do need the conversation.

wire-ios-sync-engine/Source/Data Model/Conversation+AccessMode.swift
agisilaos: Done!

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
agisilaos: Done!

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandlerTests.swift
agisilaos: I removed it and added more tests, done!

wire-ios-data-model/Source/Model/Conversation/Actions/SetAllowGuestsAndServicesAction.swift
agisilaos: Done!

wire-ios-data-model/Source/Model/Conversation/Actions/CreateConversationGuestLinkAction.swift
agisilaos: For the rest of the actions we define the errors inside the Action and not the handler. Should we change that now @johnxnguyen?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandlerTests.swift
agisilaos: Yep it is unknown I ported over the error structure from the ZMConversation to the Action

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandlerTests.swift
agisilaos: Done!

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
agisilaos: Done!

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
agisilaos: Done!

wire-ios-data-model/Source/Model/Conversation/Actions/CreateConversationGuestLinkAction.swift
agisilaos: Done!

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
agisilaos: Done!

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
```diff
@@ -0,0 +1,121 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireDataModel
+
+class CreateConversationGuestLinkActionHandler: ActionHandler<CreateConversationGuestLinkAction> {
```
markusfassbender: Can it become `final`?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
```diff
@@ -0,0 +1,121 @@
...
+
+    // MARK: - Request generation
+
+    override func request(
+        for action: Action,
+        apiVersion: APIVersion
+    ) -> ZMTransportRequest? {
```
markusfassbender, line 29: Maybe this can be extracted and tested in a builder, similar to `SelfSupportedProtocolsRequestBuilder`?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
markusfassbender: Currently I still see the dictionary code, so what am I missing? 😁
```swift
guard let payload = response.payload?.asDictionary(),
                  let data = payload["data"] as? [String: Any],
                  let uri = data["uri"] as? String else {
```

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/SetAllowGuestsAndServicesActionHandler.swift
```diff
@@ -0,0 +1,111 @@
...
+class SetAllowGuestsAndServicesActionHandler: ActionHandler<SetAllowGuestsAndServicesAction> {
+
+    private lazy var eventProcessor = ConversationEventProcessor(context: context)
+
+    // MARK: - Request Generation
+
+    override func request(
+        for action: SetAllowGuestsAndServicesAction,
+        apiVersion: APIVersion
+    ) -> ZMTransportRequest? {
```
markusfassbender, line 30: Maybe this can be extracted and tested in a builder, similar to `SelfSupportedProtocolsRequestBuilder`?

wire-ios-sync-engine/Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -29,6 +29,10 @@ public enum SetAllowServicesError: Error {
     case invalidOperation
 }
 
+public enum ContextError: Error {
+    case contextUnavailable
+}
```
markusfassbender: Why is this enum declared in this file, if it's never used here? Can it be a single new file?

wire-ios-sync-engine/Source/Use cases/CreateConversationGuestLinkUseCase.swift
```diff
@@ -0,0 +1,84 @@
...
+
+struct CreateConversationGuestLinkUseCase: CreateConversationGuestLinkUseCaseProtocol {
+
+    let setGuestsAndServicesUseCase: SetAllowGuestAndServicesUseCaseProtocol
+
+    public func invoke(
+        conversation: ZMConversation,
+        password: String?,
+        completion: @escaping (Result<String?, Error>) -> Void
+    ) {
```
markusfassbender: From my use case understanding I would expect rather to init with values and then just call `invoke()`. This makes it also easier to mock, because the init can be different for the mock.
```swift
let useCase = CreateConversationGuestLinkUseCase(
        conversation: ZMConversation,
        password: String?,
        completion: @escaping (Result<String?, Error>) -> Void
)
useCase.invoke()
```

wire-ios-sync-engine/Source/Use cases/SetAllowGuestAndServicesUseCase.swift
```diff
@@ -0,0 +1,65 @@
...
+}
+
+struct SetAllowGuestAndServicesUseCase: SetAllowGuestAndServicesUseCaseProtocol {
+
+    func invoke(
+        conversation: ZMConversation,
+        allowGuests: Bool,
+        allowServices: Bool,
+        completion: @escaping (Result<Void, Error>) -> Void
+    ) {
```
markusfassbender: Same comment as other use case

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -272,7 +272,7 @@ public final class ZMUserSession: NSObject {
         }
     }
 
-    let useCaseFactory: UseCaseFactoryProtocol
+   public let useCaseFactory: UseCaseFactoryProtocol
```
markusfassbender: 💅 indentation looks off

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
```diff
@@ -0,0 +1,121 @@
...
+
+    }
+
+    override func handleResponse(_ response: ZMTransportResponse, action: CreateConversationGuestLinkAction) {
+
+        var action = action
+
+        switch (response.httpStatus, response.payloadLabel()) {
+        case (201, _):
+            // This is an update event and therefore this is why we leave it as is
```
johnxnguyen: Someone reading this comment would probably not understand what you mean by "leave it as it is".

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
```diff
@@ -0,0 +1,121 @@
...
+            action.succeed(with: payload.uri)
+
+            guard let payload = response.payload?.asDictionary(),
+                  let uri = payload["uri"] as? String else {
+
+                action.fail(with: .failedToDecodePayload)
+                return
+            }
+
+            action.succeed(with: uri)
```
johnxnguyen: I think you forgot to delete this part

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandlerTests.swift
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    override func tearDown() async throws {
+        stack = nil
+        try coreDataStackHelper.cleanupDirectory()
+        try await super.tearDown()
+    }
+
+    func testCreateConversationGuestLinkRequestGeneration_APIV4() throws {
```
johnxnguyen, line 96: It would be good to also assert the body payload since that is part of the request.

johnxnguyen: I see tests for V4 and up but not V3 and below. Are you planning to add them?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandlerTests.swift
```diff
@@ -0,0 +1,124 @@
...
+            apiVersion: .v4
+        )
+
+        // THEN
+        XCTAssertNotNil(result, "Result should not be nil")
+        if let resultData = result {
+            XCTAssertEqual(resultData, expectedUri, "The URI should match the expected URI")
+        } else {
+            XCTFail("Result does not contain URI or is of unexpected type")
+        }
```
johnxnguyen: You could replace this with a single line:

```swift
XCTAssertEqual(result, expectedUri)
```

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandlerTests.swift
```diff
@@ -0,0 +1,124 @@
...
+
+    func testCreateConversationGuestLinkFailure() throws {
+        // GIVEN
+        let statusCode = 400
+
+        let payload: [AnyHashable: Any] = [
+            "code": "SOME-UNIQUE-CODE",
+            "has_password": false,
+            "key": "sampleKey123"
+        ]
```
johnxnguyen: A 400 with a payload isn't something that we expect from the backend. I suggest to write a more realistic test or tests, with known error codes and labels.

wire-ios-sync-engine/Source/Use cases/CreateConversationGuestLinkUseCase.swift
```diff
@@ -0,0 +1,84 @@
...
+                case .success:
+                    createWirelessLink(conversation: conversation, password: password, completion)
+                }
+            }
+        } else {
+            createWirelessLink(conversation: conversation, password: password, completion)
+        }
+    }
+
+    func createWirelessLink(
```
johnxnguyen: I suggest to avoid having different terms like "WirelessLink", let's just stick with "GuestLink"

wire-ios-sync-engine/Source/Use cases/CreateConversationGuestLinkUseCase.swift
```diff
@@ -0,0 +1,84 @@
...
+            password: password,
+            conversationID: conversation.remoteIdentifier
+        )
+
+        action.perform(in: context.notificationContext) { result in
+            switch result {
+            case .success(let link):
+                completion(.success(link))
+            case .failure(let error):
+                completion(.failure(error))
```
johnxnguyen: If you're not mapping the result in any way, you can just pass the completion as a parameter:

```swift
action.perform(in: context.notificationContext, resultHandler: completion)
```

wire-ios-sync-engine/Source/Use cases/CreateConversationGuestLinkUseCase.swift
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+
+    func createWirelessLink(
+        conversation: ZMConversation,
+        password: String?,
+        _ completion: @escaping (Result<String?, Error>) -> Void
+    ) {
+        guard conversation.canManageAccess else {
+            return completion(.failure(WirelessLinkError.invalidOperation))
```
johnxnguyen: I suggest to have a `CreateConversationGuestLinkUseCaseError` instead of `WirelessLinkError` and `ContextError`.

wire-ios-sync-engine/Source/Use cases/SetAllowGuestAndServicesUseCase.swift
```diff
@@ -0,0 +1,65 @@
...
+        )
+
+        action.perform(in: context.notificationContext) { result in
+            switch result {
+            case .success:
+                completion(.success(()))
+            case .failure(let error):
+                completion(.failure(error))
+            }
+        }
```
johnxnguyen: Same, you could just pass the completion as a parameter.

wire-ios-sync-engine/Source/UserSession/UserSession.swift
```diff
@@ -26,6 +26,8 @@ public typealias SelfUserType = UserType & SelfLegalHoldSubject
 /// layer.
 public protocol UserSession: AnyObject {
 
+    var useCaseFactory: UseCaseFactoryProtocol { get }
```
johnxnguyen: Does this need to be exposed? we could (like we already do) expose protocol requirements to get the use case, but internal the implementation can use the use case factory.

wire-ios/Wire-iOS Tests/ConversationOptionsViewControllerTests.swift
```diff
@@ -64,17 +66,30 @@ final class MockOptionsViewModelConfiguration: ConversationGuestOptionsViewModel
 
 final class ConversationOptionsViewControllerTests: BaseSnapshotTestCase {
 
+    var mockConversation: MockConversation!
+    var mockUserSession: UserSessionMock!
+    var mockUseCaseFactory: MockUseCaseFactoryProtocol!
```
johnxnguyen: Why do you need the factory?

wire-ios/Wire-iOS Tests/CreateSecureGuestLinkViewModelTests.swift
```diff
@@ -19,22 +19,30 @@
...
 class CreateSecureGuestLinkViewModelTests: XCTestCase {
 
     // MARK: - Properties
 
-    var viewModel: CreateSecureGuestLinkViewModel!
+    var viewModel: CreateSecureConversationGuestLinkViewModel!
     var mockDelegate: MockCreatePasswordSecuredLinkViewModelDelegate!
     var textField: ValidatedTextField!
     var confirmPasswordField: ValidatedTextField!
+    var userSession: UserSessionMock!
```
johnxnguyen: Why do you need the user session?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreateSecureGuestLink/CreateSecureConversationGuestLinkViewModel.swift
```diff
@@ -62,6 +76,31 @@ final class CreateSecureGuestLinkViewModel {
         return true
     }
 
+    func createSecuredGuestLinkIfValid(conversation: ZMConversation, passwordField: ValidatedTextField, confirmPasswordField: ValidatedTextField) {
+        guard validatePassword(for: passwordField, against: confirmPasswordField) else {
+            delegate?.viewModel(self, didFailToValidatePasswordWithReason: "Password validation failed.")
+            return
+        }
+
+        let password = passwordField.text ?? ""
```
johnxnguyen: I suggest to unwrap the password rather than default to empty.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreateSecureGuestLink/CreateSecureConversationGuestLinkViewModel.swift
```diff
@@ -18,26 +18,40 @@
...
+    func viewModel(_ viewModel: CreateSecureConversationGuestLinkViewModel, didFailToCreateLinkWithError error: Error)
 }
 
 // MARK: - CreateSecureGuestLinkViewModel
 
-final class CreateSecureGuestLinkViewModel {
+final class CreateSecureConversationGuestLinkViewModel {
+
+    enum LinkCreationError: Error {
+        case underfinedLink
```
johnxnguyen, line 39: What does this mean?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreateSecureGuestLink/CreateSecureGuestLinkViewController.swift
```diff
@@ -340,12 +348,34 @@ class CreateSecureGuestLinkViewController: UIViewController, CreatePasswordSecur
...
+
+    }
+
+    func viewModel(_ viewModel: CreateSecureConversationGuestLinkViewModel, didCreateLink link: String) {
+        print("Link created successfully: \(link)")
+    }
+
+    func viewModel(_ viewModel: CreateSecureConversationGuestLinkViewModel, didFailToCreateLinkWithError error: Error) {
+        print("Failed to create link: \(error.localizedDescription)")
+    }
```
johnxnguyen, line 378: Do these still need to be implemented?

johnxnguyen (commented):
Will you add tests for the use cases?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreateSecureGuestLink/CreateSecureConversationGuestLinkViewModel.swift
agisilaos: Done! https://github.com/wireapp/wire-ios/commit/e03678e15010db6793ed438fc93973ddab433709

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreateSecureGuestLink/CreateSecureConversationGuestLinkViewModel.swift
agisilaos: I needed to pass an error when success is nil:

```swift
conversationGuestLinkUseCase.invoke(conversation: conversation, password: password) { [weak self] result in
            guard let self = self else { return }

            switch result {
            case .success(let link?):
                self.delegate?.viewModel(self, didCreateLink: link)
            case .success(nil):
                self.delegate?.viewModel(self, didFailToCreateLinkWithError: LinkCreationError.underfinedLink)
            case .failure(let error):
                self.delegate?.viewModel(self, didFailToCreateLinkWithError: error)
            }
        }
```

wire-ios/Wire-iOS Tests/CreateSecureGuestLinkViewModelTests.swift
agisilaos: Fixed! Done: https://github.com/wireapp/wire-ios/commit/ed6f88170f0cce8e92b3f41b81cb97417ea823f6

wire-ios/Wire-iOS Tests/ConversationOptionsViewControllerTests.swift
agisilaos: I don't. Fixed: https://github.com/wireapp/wire-ios/commit/9f9a7070ccdc01473c809e1d326b6475aa17d995#diff-7631d025a0150db9f2361daf32ad589a626ff1dd46e7306c4489946501fa3399

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreateSecureGuestLink/CreateSecureGuestLinkViewController.swift
agisilaos: I will do in a separate PR.

wire-ios-sync-engine/Source/Use cases/CreateConversationGuestLinkUseCase.swift
agisilaos: Done! https://github.com/wireapp/wire-ios/commit/1d56097f3c14f4078a0d4326477fdecc44608de9

wire-ios-sync-engine/Source/Use cases/CreateConversationGuestLinkUseCase.swift
agisilaos: Done! https://github.com/wireapp/wire-ios/commit/15b50b7e9f2ea9e899ce8a1b1300bce33a9f8b7a

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/e0a69bfe1568fa50179ef2c82e882edc764902de

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
agisilaos: Done! https://github.com/wireapp/wire-ios/commit/e0a69bfe1568fa50179ef2c82e882edc764902de

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandlerTests.swift
agisilaos: Done! https://github.com/wireapp/wire-ios/commit/d12c2aacf79f6ffd9958bab817dc06a7556134b9

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandlerTests.swift
agisilaos: Done!

https://github.com/wireapp/wire-ios/commit/e5af88a4afb5a6fa75bec2f01e798549d9a5e19b

wire-ios-sync-engine/Source/Data Model/Conversation+AccessMode.swift
agisilaos: Got rid of it!

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
agisilaos: Done! https://github.com/wireapp/wire-ios/commit/64897966ddd28b74a1b4342e82c536ab19082662

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
agisilaos: Not anymore!

wire-ios-sync-engine/Source/Use cases/SetAllowGuestAndServicesUseCase.swift
agisilaos: Fixed it here: https://github.com/wireapp/wire-ios/commit/07232abe05c5fc55bbeb6c649d92ec01e8d027d0

wire-ios-sync-engine/Source/Use cases/CreateConversationGuestLinkUseCase.swift
agisilaos: I did for the other use case. Do you think it's worth doing so for this one too? In this case it's going to be more complex since the two useCase need to share a common error type for the completions to work as expected.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
agisilaos: @markusfassbender I updated that part with a comment why needs to stay that way.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandlerTests.swift
agisilaos: Fixed!

wire-ios-sync-engine/Source/Use cases/CreateConversationGuestLinkUseCase.swift
agisilaos: Fixed!

wire-ios-sync-engine/Source/UserSession/UserSession.swift
agisilaos: Fixed it!

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
```diff
@@ -0,0 +1,117 @@
...
+            action.fail(with: .invalidOperation)
+
+        case (404, "no-conversation-code"?):
+            action.fail(with: .noCode)
+
+        case (404, "no-conversation"?):
+            action.fail(with: .noConversation)
+
+        case (409, "guest-links-disabled"?):
+            action.fail(with: .guestLinksDisabled)
```
johnxnguyen: I don't think the `?` are needed here.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/SetAllowGuestsAndServicesActionHandler.swift
```diff
@@ -0,0 +1,111 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireDataModel
+
+public enum SetAllowGuestsAndServicesError: Error {
+    case unknown
+    case invalidOperation
+}
```
johnxnguyen: Are these errors for the action handler or something else?

wire-ios-sync-engine/Tests/Source/Use cases/CreateConversationGuestLinkUseCaseTests.swift
```diff
@@ -0,0 +1,127 @@
...
+                case .success(let link):
+                    XCTAssertNotNil(link)
+                case .failure(let error):
+                    XCTFail("Test failed with error: \(error)")
+                }
+
+                expectation.fulfill()
+            }
+
+            wait(for: [expectation], timeout: 4.0)
```
johnxnguyen: 4 seconds is a bit long, usually we wait for 0.5 seconds.

wire-ios-sync-engine/Tests/Source/Use cases/CreateConversationGuestLinkUseCaseTests.swift
```diff
@@ -0,0 +1,127 @@
...
+            let mockHandler = MockActionHandler<CreateConversationGuestLinkAction>(result: .failure(.unknown), context: syncContext.notificationContext)
+
+            let expectation = XCTestExpectation(description: "completion should be called")
+
+            sut.invoke(conversation: mockConversation, password: nil) { result in
+                switch result {
+                case .success(let success):
+                    XCTFail("Expected operation to fail, but it succeeded.")
+                case .failure(let error):
+                    print("Operation failed with \(error)")
```
johnxnguyen: suggestion: simply `break`

wire-ios-sync-engine/Tests/Source/Use cases/CreateConversationGuestLinkUseCaseTests.swift
```diff
@@ -0,0 +1,127 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireDataModelSupport
+import XCTest
+
+@testable import WireSyncEngine
+
+final class CreateConversationGuestLinkUseCaseTests: XCTestCase {
```
johnxnguyen, line 25: Are you planning to add tests for the legacy path?

wire-ios-sync-engine/Tests/Source/Use cases/SetAllowGuestsAndServicesUseCaseTests.swift
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    func testGuestEnablementFails_WithInsufficientPermissions() async {
+        await syncContext.perform { [self] in
+            // GIVEN
+            let role = Role.insertNewObject(in: syncContext)
+            let action = Action.insertNewObject(in: syncContext)
+
+            action.name = "modify_conversation_access"
+            role.actions = [action]
```
johnxnguyen: The only difference I see with the test where the user has permission is the role name. Is that gives the user permission?

wire-ios-sync-engine/Tests/Source/Use cases/SetAllowGuestsAndServicesUseCaseTests.swift
```diff
@@ -0,0 +1,180 @@
...
+
+    func testServicesEnablementFails_WithInsufficientPermissions() async {
+
+        await syncContext.perform { [self] in
+            // GIVEN
+            let role = Role.insertNewObject(in: syncContext)
+            let action = Action.insertNewObject(in: syncContext)
+
+            action.name = "modify_conversation_access"
+            role.actions = [action]
```
johnxnguyen: This looks the same as the test where the user has permission...

wire-ios-sync-engine/Tests/Source/Use cases/SetAllowGuestsAndServicesUseCaseTests.swift
```diff
@@ -0,0 +1,180 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireDataModelSupport
+import XCTest
+
+@testable import WireSyncEngine
+
+final class SetAllowGuestsAndServicesUseCaseTests: XCTestCase {
```
johnxnguyen, line 24: It would be nice to factor out the // GIVEN code to reduce the duplication.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/CreateSecureGuestLink/CreateSecureConversationGuestLinkViewModel.swift
johnxnguyen: Did you find out from backend when we can expect to have a nil link?

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/SetAllowGuestsAndServicesActionHandler.swift
agisilaos: Leftovers I removed them thank you!

wire-ios-sync-engine/Tests/Source/Use cases/CreateConversationGuestLinkUseCaseTests.swift
agisilaos: Done!

wire-ios-sync-engine/Tests/Source/Use cases/CreateConversationGuestLinkUseCaseTests.swift
agisilaos: Done!

wire-ios-sync-engine/Tests/Source/Use cases/SetAllowGuestsAndServicesUseCaseTests.swift
agisilaos: Yes, as far as I understood.

wire-ios-sync-engine/Tests/Source/Use cases/SetAllowGuestsAndServicesUseCaseTests.swift
agisilaos: Clean up the entire test class. I think now it's much more clear: https://github.com/wireapp/wire-ios/commit/80a04a5d4aaf473f97eee2a7d49763ac74f136c2

wire-ios-sync-engine/Tests/Source/Use cases/SetAllowGuestsAndServicesUseCaseTests.swift
agisilaos: Clean up the entire test class. I think now it's much more clear: https://github.com/wireapp/wire-ios/commit/80a04a5d4aaf473f97eee2a7d49763ac74f136c2

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandlerTests.swift
```diff
@@ -0,0 +1,155 @@
...
+            self.action = CreateConversationGuestLinkAction(password: nil, conversationID: conversationID)
+            self.handler = CreateConversationGuestLinkActionHandler(context: syncContext)
+        }
+    }
+
+    override func tearDown() async throws {
+        stack = nil
+        try coreDataStackHelper.cleanupDirectory()
+        try await super.tearDown()
+    }
```
markusfassbender: There is one memory problem with unit tests, that is not obvious: they are not deallocated and exist in memory until end of executing the whole test suite / testplan. https://qualitycoding.org/xctestcase-teardown/

Therefore all properties should be released (except some minor constants etc.), especially because it still can have side effects if they listen to notifications or similar.
Please release `coreDataStackHelper` and `conversationID` too :)

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/SetAllowGuestsAndServicesActionHandler.swift
```diff
@@ -0,0 +1,106 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireDataModel
+
+class SetAllowGuestsAndServicesActionHandler: ActionHandler<SetAllowGuestsAndServicesAction> {
```
markusfassbender:
```suggestion
final class SetAllowGuestsAndServicesActionHandler: ActionHandler<SetAllowGuestsAndServicesAction> {
```

wire-ios-sync-engine/Source/Use cases/CreateConversationGuestLinkUseCase.swift
```diff
@@ -0,0 +1,93 @@
...
+    case contextUnavailable
+    case networkError(Error)
+    case failedToEnableGuestAccess(Error)
+
+}
+
+// sourcery: AutoMockable
+public protocol CreateConversationGuestLinkUseCaseProtocol {
+
+    func invoke(conversation: ZMConversation, password: String?, completion: @escaping (Result<String?, CreateConversationGuestLinkUseCaseError>) -> Void)
```
markusfassbender:
```suggestion
    func invoke(
        conversation: ZMConversation,
        password: String?,
        completion: @escaping (Result<String?, CreateConversationGuestLinkUseCaseError>) -> Void
    )
```

wire-ios-sync-engine/Tests/Source/Use cases/SetAllowGuestsAndServicesUseCaseTests.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/3dcd7085e7f1fac560715becc8888c781e18f971

wire-ios-sync-engine/Source/Use cases/CreateConversationGuestLinkUseCase.swift
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+
+    func createGuestLink(
+        conversation: ZMConversation,
+        password: String?,
+        _ completion: @escaping (Result<String?, CreateConversationGuestLinkUseCaseError>) -> Void
+    ) {
+        guard conversation.canManageAccess else {
+            return completion(.failure(CreateConversationGuestLinkUseCaseError.invalidOperation))
```
markusfassbender: When I read it this way I always ask myself if the `return` doesn't exit already and the `completion` closure is performed or not?
I think it works, but find it easier to read:
```swift
completion(.failure(CreateConversationGuestLinkUseCaseError.invalidOperation))
return
```

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandler.swift
agisilaos: Fixed! https://github.com/wireapp/wire-ios/commit/e46ee2576cd833baf1305d63245b68224b9eb505

wire-ios-sync-engine/Source/Use cases/UseCaseFactory.swift
```diff
@@ -19,9 +19,11 @@
 import Foundation
 
 // sourcery: AutoMockable
-public protocol UseCaseFactoryProtocol {
+protocol UseCaseFactoryProtocol {
```
markusfassbender, line 22: 👍

wire-ios-sync-engine/Tests/Source/Use cases/SetAllowGuestsAndServicesUseCaseTests.swift
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    // MARK: - tearDown
+
+    override func tearDown() async throws {
+        stack = nil
+        sut = nil
+        try coreDataStackHelper.cleanupDirectory()
+        try await super.tearDown()
+    }
```
markusfassbender: Here the same, all properties should be released.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/ConversationGuestOptionsViewController.swift
```diff
@@ -167,6 +163,10 @@ final class ConversationGuestOptionsViewController: UIViewController,
         activityController.configPopover(pointToView: sourceView ?? view)
     }
 
+    func viewModel(_ viewModel: ConversationGuestOptionsViewModel, presentCreateSecureGuestLink viewController: UIViewController, animated: Bool) {
+        self.present(viewController, animated: animated, completion: nil)
```
markusfassbender:
```suggestion
        present(viewController, animated: animated)
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/ConversationGuestOptionsViewModel.swift
```diff
@@ -232,26 +244,37 @@ final class ConversationGuestOptionsViewModel {
...
-        configuration.createConversationLink { [weak self] result in
-            guard let `self` = self else { return }
+        createSecureGuestLinkUseCase.invoke(conversation: conversation, password: nil) { result in
             switch result {
-            case .success(let link): self.link = link
-            case .failure(let error): self.delegate?.viewModel(self, didReceiveError: error)
+            case .success(let link):
+                self.link = link
+            case .failure(let error):
+                self.delegate?.viewModel(self, didReceiveError: error)
```
markusfassbender: Potential retain cycle?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/ConversationGuestOptionsViewModel.swift
```diff
@@ -232,26 +244,37 @@ final class ConversationGuestOptionsViewModel {
...
+                                presentGuestLinkTypeSelection: { [weak self] guestLinkType in
                 guard let `self` = self else { return }
                 switch guestLinkType {
-                case .secure: break
+                case .secure:
+                    delegate?.viewModel(
+                        self,
+                        presentCreateSecureGuestLink: CreateSecureGuestLinkViewController(
+                            conversationSecureGuestLinkUseCase: createSecureGuestLinkUseCase,
+                            conversation: conversation).wrapInNavigationController(),
```
markusfassbender: It would be easier to read if we make a let out of it.
```swift
let viewController = CreateSecureGuestLinkViewController(
    conversationSecureGuestLinkUseCase: createSecureGuestLinkUseCase,
    conversation: conversation
)
let navigationController = viewController.wrapInNavigationController()
```

markusfassbender (approved):

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/CreateConversationGuestLinkActionHandlerTests.swift
agisilaos: Done!

wire-ios-sync-engine/Tests/Source/Use cases/SetAllowGuestsAndServicesUseCaseTests.swift
agisilaos: Done!

wire-ios-sync-engine/Source/Use cases/CreateConversationGuestLinkUseCase.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/ConversationGuestOptionsViewModel.swift
agisilaos: Done!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/ConversationGuestOptionsViewModel.swift
agisilaos: Fixed.

wire-ios-sync-engine/Tests/Source/Use cases/CreateConversationGuestLinkUseCaseTests.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/835d7c6fdcc39951370db163c4ead9551339e466

johnxnguyen (approved):

PR #1293, opened 2024-04-15
fix: add required input parameter to workflow runs

caldrian:
### Issue

GitHub Action runs suddenly failed because the input parameter `is_cloud_build` isn't set.

https://github.com/wireapp/wire-ios/actions/runs/8687385808
https://github.com/wireapp/wire-ios/actions/runs/8686862602
https://github.com/wireapp/wire-ios/actions/runs/8686455273
...

This PR sets the value in development.yml

---

### Checklist

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -14,4 +14,5 @@ jobs:
     uses: ./.github/workflows/_reusable_app_release.yml
     with:
       fastlane_action: development
+      is_cloud_build: true
     secrets: inherit


.github/workflows/development.yml
```diff
@@ -14,4 +14,5 @@ jobs:
     uses: ./.github/workflows/_reusable_app_release.yml
     with:
       fastlane_action: development
+      is_cloud_build: true
```
caldrian, line 17: Actually I don't understand why this is required.
There's a default value provided in `_reusable_app_release.yml`.

caldrian: Also I'm not sure if we should put `true` or `false`, any ideas?

markusfassbender (approved):

.github/workflows/development.yml
markusfassbender: Not sure if the `default` value only sets the value for the interface to run actions from GitHub? `true` should work here the fast lane action is `develop` and thus builds for our environments.

David-Henner (approved):

PR #1295, opened 2024-04-15
refactor: init `ZMUserSession` with builder [WPB-7179]

markusfassbender:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7179" title="WPB-7179" target="_blank">StoryWPB-7179</a>  [iOS / Spike] : App dependencies injection: how components are setup - Plan
  </td></table>
  <br />
 

### Issue

The initialization of `ZMUserSession` is a night mare of complexity and dependencies.

- Introduce a builder pattern to separate the construction from the object itself. This only takes the objects that were passed into the constructor.
- Partially the parameters had default values, that were actually internally creating objects and thus hiding missing mocks in unit tests. Not all tests were updated correctly yet.
- There could be a lot more to extract internal vars and setup logic, that I have skipped for now.

### Testing

Regression test if "everything works as before", means the `ZMUserSession` setup and itself.

---

### Checklist


wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -302,6 +302,8 @@ public final class ZMUserSession: NSObject {
         notificationDispatcher.tearDown()
         callCenter?.tearDown()
 
+        contextStorage.clear()
```
markusfassbender, line 272: The cached is cleared implicitly on object release, but retaining the `var` had the original purpose to call `clear()` explicitly.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSessionBuilder.swift
```diff
@@ -0,0 +1,363 @@
...
+    private func prepareCoreDataStack(
+        _ coreDataStack: CoreDataStack,
+        analytics: (any AnalyticsType)?
+    ) {
+        coreDataStack.syncContext.performGroupedBlockAndWait {
+            coreDataStack.syncContext.analytics = analytics
+            coreDataStack.syncContext.zm_userInterface = coreDataStack.viewContext
+        }
+        coreDataStack.viewContext.zm_sync = coreDataStack.syncContext
+    }
```
markusfassbender: I would expect that this preparation is done before the core data stack is passed to `ZMUserSession`, but not sure where to place it?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSessionBuilder.swift
```diff
@@ -0,0 +1,347 @@
...
+        withObserveMLSGroupVerificationStatusUseCase(observeMLSGroupVerificationStatusUseCase)
+        withProteusToMLSMigrationCoordinator(proteusToMLSMigrationCoordinator)
+        withSharedUserDefaults(sharedUserDefaults)
+        withTransportSession(transportSession)
+        withUpdateMLSGroupVerificationStatusUseCase(updateMLSGroupVerificationStatus)
+        withUseCaseFactory(useCaseFactory)
+        withUserID(userId)
+    }
+
+    mutating func withAnalytics(_ analytics: (any AnalyticsType)?) {
```
caldrian: I suggest to get rid of all these functions and make the properties public.

wire-ios-sync-engine/Tests/Source/SessionManager/APIMigrationManagerTests.swift
```diff
@@ -248,22 +249,40 @@ class APIMigrationManagerTests: MessagingTest {
...
-            userId: .create(),
-            transportSession: mockTransportSession,
-            mediaManager: MockMediaManager(),
-            flowManager: FlowManagerMock(),
+        let mockContextStorable = MockLAContextStorable()
+        mockContextStorable.clear_MockMethod = { }
+        let configuration = ZMUserSession.Configuration()
+
+        var builder = ZMUserSessionBuilder()
+        builder.withAllDependencies(
```
caldrian, line 257: Why not setting the properties directly?

```suggestion
        builder.appVersion = "999"
        builder.application = application
        ...
```

caldrian (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSessionBuilder.swift
markusfassbender: Thanks I've changed it

wire-ios-sync-engine/Tests/Source/SessionManager/APIMigrationManagerTests.swift
markusfassbender: Because we derive new objects from all the passed ones and all are needed. They have some cross-dependencies too.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSessionBuilder.swift
johnxnguyen: I don't see a problem with doing this before passing it to the user session.

johnxnguyen (commented):
looks good to me

johnxnguyen (approved):

PR #1297, opened 2024-04-15
refactor: create api package

johnxnguyen:
### Issue

This PR introduces the WireAPI swift package that will represent the API layer of the app. It will contain all public API models and public API protocols, as well as all private API implementations and various internal helpers.

### Checklist


WireAPI/.swiftpm/xcode/xcshareddata/xcschemes/WireAPITests.xcscheme
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
caldrian: Is a separate scheme for the tests really required? Can't the tests be part of `WireAPI.xcscheme`?

johnxnguyen: Unfortunately I do not know how to achieve this, would you be able to give me some hints?

WireAPI/Sources/WireAPI/Public/HTTP Client/HTTPRequest.swift
```diff
@@ -0,0 +1,73 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// A http request to be sent to a server.
+
+public struct HTTPRequest {
```
caldrian: Don't we replicate the `URLRequest` type here?

WireAPI/Sources/WireAPI/Public/HTTP Client/HTTPResponse.swift
```diff
@@ -0,0 +1,43 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// A http response that is received from a server.
+
+public struct HTTPResponse {
```
caldrian: Don't we have the `HTTPURLResponse` type for those values?

WireAPI/Sources/WireAPI/Public/APIs/BackendInfoAPI.swift
```diff
@@ -0,0 +1,27 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public protocol BackendInfoAPI {
+
+    /// Fetch the info of the local backend.
+
+    func getBackendInfo() async throws -> BackendInfo
```
caldrian: opinion/idea: This declaration looks like its implementation should actually be put into a use case.
I wonder if we could reduce the level of abstraction.

johnxnguyen: I don't think I fully understand what you mean, could you please elaborate more on this idea?

caldrian: If a type conforming to this protocol lives in the API package/module, what would a use case implementation using this look like?
It would probably almost be a one-liner, calling this function.

I wonder if we could use the types of the iOS SDK directly instead of creating levels of abstractions.

caldrian: What about a different approach. Every use of our backend API in the end follows this pattern:
- create an instance of `URLRequest`
- pass it to an `URLSession` instance
- handle the response (status code and bytes)

Why not putting that into a protocol, like:

```
protocol BackendAPI {
    associatedtype APIModel

    func createRequest(baseURL: URL) -> URLRequest
    func parse(data: Data, response: URLResponse) async throws -> APIModel
}
```

If dealing with the `URLSession` is repetitive, we can of course put another convenience implementation around this.

caldrian: Simple APIs like the one you implemented here could be put into an enum.

```
enum BackendInfoAPI: BackendAPI {

    case v0, v1, v2, v3, v4, v5

    func createRequest(baseURL: URL) -> URLRequest {
        /* create the instance */
    }

    func parse(data: Data, response: URLResponse) throws -> BackendInfo {
        /* we know
         - how to parse the bytes into our model
         - which status code to expect and how to translate them into error types
         */
    }

    struct BackendInfo {
        /* API model contents */
    }
}
```

caldrian: APIs that changed with the versions could be put into structs.

```
enum MoreComplexAPI {
    struct MoreComplexAPIModel { /* ... */ }
}

// MARK: V0

extension MoreComplexAPI {

    struct MoreComplexAPIV0: BackendAPI {

        func createRequest(baseURL: URL) -> URLRequest { /* ... */ }
        func parse(data: Data, response: URLResponse) throws -> MoreComplexAPIModel { /* ... */ }
    }
}

// MARK: V1

extension MoreComplexAPI {

    struct MoreComplexAPIV1: BackendAPI {

        func createRequest(baseURL: URL) -> URLRequest { /* ... */ }
        func parse(data: Data, response: URLResponse) throws -> MoreComplexAPIModel { /* ... */ }
    }
}
```

That's what I just came up with, maybe I'm missing a something important.

WireAPI/Sources/WireAPI/Private/Components/ResponsePayloadDecoder.swift
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/// An object responsible for decoding the http response payload into
+/// an api model.
+
+struct ResponsePayloadDecoder {
+
+    let decoder: JSONDecoder
+
+    init(decoder: JSONDecoder = .defaultDecoder) {
```
markusfassbender: Not sure if we want `.defaultDecoder` to be a default parameter, because it's easy to oversee and might be overseen in tests etc.
I tend a little bit more to have it explicit, but no strong opinion.

WireAPI/Sources/WireAPI/Private/Components/ResponsePayloadDecoder.swift
```diff
@@ -0,0 +1,88 @@
...
+
+    func decodePayload<T: Decodable>(
+        from response: HTTPResponse,
+        as type: T.Type
+    ) throws -> Result<
+        T,
+        FailureResponse
+    > {
+        switch response.code {
+        case 100...300:
```
markusfassbender: Shouldn't this include all 3xx codes, like `100..<400`?

WireAPI/Sources/WireAPI/Public/Models/APIVersion.swift
```diff
@@ -0,0 +1,33 @@
...
+
+public enum APIVersion: UInt {
+
+    case v0
+    case v1
+    case v2
+    case v3
+    case v4
+    case v5
+    case v6
```
markusfassbender: Should `v6` be marked as development only somehow? Through the name `v6_development` or some comment?

WireAPI/Sources/WireAPI/Public/Models/BackendInfo.swift
```diff
@@ -0,0 +1,41 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// Metadata about the local backend.
+
+public struct BackendInfo {
```
markusfassbender: My personal preference is to write comments without blank linkes to definition. This is also the formatting one sees in Apple's classes.
I'm open to discuss it, but I think we should align in the team and follow the decision at least for the new code we write :)

Screenshot 2024-04-17 at 09 33 24

wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DebugActions/DeveloperDebugActionsViewModel.swift
```diff
@@ -129,4 +131,19 @@ final class DeveloperDebugActionsViewModel: ObservableObject {
                 .qualifiedID
         }
     }
+
+    private func makeRequest() {
+        guard let userSession = ZMUserSession.shared() else { return }
+        let api = userSession.makeBackendInfoAPI()
+        let logger = WireLogger(tag: "api")
```
markusfassbender: As I understood @netbe we should rather use the logger in a way with static instances like `WireLogger.api.debug(...)`

wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DebugActions/DeveloperDebugActionsViewModel.swift
```diff
@@ -129,4 +131,19 @@ final class DeveloperDebugActionsViewModel: ObservableObject {
                 .qualifiedID
         }
     }
+
+    private func makeRequest() {
```
markusfassbender: Maybe it's due to my English skills that I would expect from the verb `make` to return an object, like `build` or `create`. Does `performRequest()` describes the functionality better?

WireAPI/Sources/WireAPI/Private/Components/ResponsePayloadDecoder.swift
markusfassbender: Actually thinking about it again I don't see why a `Decoder` should be responsible to determine if a http request was successful or not.
In my opinion this should be handled before and then a successful result can be decoded.

WireAPI/Sources/WireAPI/Private/APIs/BackendInfoAPI/BackendInfoAPIV1.swift
```diff
@@ -0,0 +1,25 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+class BackendInfoAPIV1: BackendInfoAPIV0 {
```
markusfassbender: I don't understand why we have this chain of inheritance of api versions, can not each version just conform to the protocol `BackendInfoAPI`?
This would make it easier to understand where a certain logic is implemented and also remove versions in the future.

WireAPI/Sources/WireAPI/Private/Extensions/Date+Transport.swift
markusfassbender, line 1: Those same extensions exists already in `WireTransport` with unit tests. Can we extract it to a new package and reuse in both places instead of duplication?

WireAPI/Sources/WireAPI/Private/APIs/BackendInfoAPI/BackendInfoAPIV0.swift
```diff
@@ -0,0 +1,69 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+class BackendInfoAPIV0: BackendInfoAPI {
+
+    let httpClient: HTTPClient
+
+    init(httpClient: HTTPClient) {
```
markusfassbender: It looks upside down to me to pass the http client into the endpoint description. I would expect to have a http client that is fed with endpoint descriptions and receives a certain result (`Data`?) from the server and decodes the result into a model.

markusfassbender (commented):
@johnxnguyen I left some comments with unfinished thoughts. Probably we can discuss many points and it's not black and white. From my perspective our goal is to build a long-term scalable solution for the whole app and I appreciate your initiative 👍

WireAPI/Sources/WireAPI/Private/Components/ResponsePayloadDecoder.swift
johnxnguyen: Sure, I agree with that

WireAPI/Sources/WireAPI/Private/Components/ResponsePayloadDecoder.swift
johnxnguyen: So in WireTransport, this is determined earlier and is part of the `ZMTransportResponse`. I suppose one way to preserve this is to take the result from the `ZMTransportResponse` and map it into the `HTTPResponse` object that we receive prior to decoding.

WireAPI/Sources/WireAPI/Public/Models/APIVersion.swift
johnxnguyen: Yeah currently the development versions are marked in the SyncEngine in an extension, but you're right that should be done within WireAPI.

WireAPI/Sources/WireAPI/Public/Models/BackendInfo.swift
johnxnguyen: I always add a space to add some vertical separation to avoid crowding the type declaration or signature (makes it easier to read for me) but I'm not precious about it so if there's an agreed convention then I'll happily stick to that.

wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DebugActions/DeveloperDebugActionsViewModel.swift
johnxnguyen: Yes I agree, this was just a temp method so I didn't bother to create a static logger.

wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DebugActions/DeveloperDebugActionsViewModel.swift
johnxnguyen: Again, just a temp method that I'll be deleting.

WireAPI/Sources/WireAPI/Private/APIs/BackendInfoAPI/BackendInfoAPIV1.swift
johnxnguyen: This is one approach that I implemented that is the same that Android does. I'll also implement another approach without inheritance, so we can compare what we prefer.

WireAPI/Sources/WireAPI/Private/Extensions/Date+Transport.swift
johnxnguyen: Yes we could do that, but it might be that those extensions become obsolete once we rely solely on WireAPI.

WireAPI/Sources/WireAPI/Private/APIs/BackendInfoAPI/BackendInfoAPIV0.swift
johnxnguyen: Can you provide a code example?

WireAPI/Sources/WireAPI/Private/Extensions/Date+Transport.swift
markusfassbender: agree, just the migration might take years and if one needs to fix a bug, it might be tricky to remember about both implementations..

WireAPI/Sources/WireAPI/Public/HTTP Client/HTTPRequest.swift
johnxnguyen: Currently `WireTransport` implements our network layer, which exposes `ZMTransportRequest`, `ZMTransportResponse`, and `ZMTransportSession`. These provide a wrapper around `URLRequest`, `URLResponse` and `URLSession`, adding some functionality.

Since this swift package can't depend on `WireTransport`, I created these request and response types, as well as the `HTTPClient` protocol as adapters so I could insert `WireAPI` in our current stack and make use of `WireTransport`. 

Perhaps there is a better way, this was just one that I could think of.

WireAPI/Sources/WireAPI/Public/HTTP Client/HTTPResponse.swift
johnxnguyen: See https://github.com/wireapp/wire-ios/pull/1297#discussion_r1568823728

WireAPI/Sources/WireAPI/Public/APIs/BackendInfoAPI.swift
johnxnguyen: > I wonder if we could use the types of the iOS SDK directly instead of creating levels of abstractions.

@caldrian this would bypass our network stack, so we'd need to reimplement quite a few things to keep existing behavior.

johnxnguyen: > What about a different approach. Every use of our backend API in the end follows this pattern:
> - create an instance of URLRequest
> - pass it to an URLSession instance
> - handle the response (status code and bytes)

This is what @markusfassbender is also suggesting (right?)

WireAPI/Sources/WireAPI/Private/APIs/BackendInfoAPI/BackendInfoAPIV1.swift
typfel: This was debated at length in the Android team. 

**Inheritance**
\+ Minimum code duplication
\+ It's clear from reading the code what changed with each API version
\- Deleting old API versions require some care and code restructuring 

**Without  Inheritance**
\+ No dependencies so very simple to delete old API versions
\- Lot of code duplication, so any fixes must be applied in several places.

johnxnguyen: thanks for the summary

WireAPI/Sources/WireAPI/Public/APIs/BackendInfoAPI.swift
typfel: @johnxnguyen how will this WireApi module  fit in with the rest of the app? Is the plan that the `WireTransport` will go away eventually and that functionality  will live inside the WireAPI module?

johnxnguyen: I haven't thought that much about it yet so but that is certainly an interesting idea. What do you think?

WireAPI/Sources/WireAPI/Private/APIs/BackendInfoAPI/BackendInfoAPIV0.swift
markusfassbender: I've added another package next to this one to compare: https://github.com/wireapp/wire-ios/pull/1309

WireAPI/Sources/WireAPI/Private/Components/ResponsePayloadDecoder.swift
johnxnguyen: I removed this from the decoder and instead check the code in the API to determine how to interpret the response.

WireAPI/Sources/WireAPI/Private/APIs/BackendInfoAPI/BackendInfoAPIV1.swift
netbe: @typfel what did the Android go with 
I would personally go with inheritance since we don't (have never) delete versions that often

johnxnguyen: They use inheritance

WireAPI/Sources/WireAPI/Public/HTTP Client/HTTPRequest.swift
johnxnguyen: From our past discussion, our longer term goal is for WireAPI to also contain the transport code which would replace WireTransport. Until we have built ant tested the new transport code, we still need to rely on WireTransport, which is why I have these simple request and response structs to act as a bridge. But certainly when we no longer need this bridging we can get rid of them.

WireAPI/Sources/WireAPI/Public/APIs/BackendInfoAPI.swift
johnxnguyen: See https://github.com/wireapp/wire-ios/pull/1297#discussion_r1591996064

WireAPI/Sources/WireAPI/Private/APIs/BackendInfoAPI/BackendInfoAPIV2.swift
```diff
@@ -0,0 +1,71 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+class BackendInfoAPIV2: BackendInfoAPIV1 {
```
typfel: How will we decide which `BackendInfoAPI` version we should use? Since the `/api-version` is unversioned doesn't it make to have only one BackendInfoAPI implementation where the `var development: [UInt]` is optional?

typfel (changes requested):

WireAPI/Sources/WireAPI/Private/APIs/BackendInfoAPI/BackendInfoAPIV2.swift
typfel: I get that it serves as an example here so maybe it makes sense to keep initially until we've implemented another versioned API .

johnxnguyen: I will create a ticket to make it an unversioned API once we no longer need it as an example, does that sound ok?

typfel (approved):

WireAPI/Sources/WireAPI/Private/Components/PayloadPager.swift
```diff
@@ -0,0 +1,70 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// An object responsible for facilitating the iteration of
+/// one or more pages returned from an api endpoint.
+
+struct PayloadPager<Payload>: AsyncSequence {
```
markusfassbender: I understand the `PayloadPager` as a showcase example. But should we merge it now, it's never used and probably confuses the reader?
Maybe we can store it on a simple branch or Gist?

WireAPI/Sources/WireAPI/Private/Components/RequestBodyEncoder.swift
```diff
@@ -0,0 +1,54 @@
...
+enum RequestBodyEncoderError: Error {
+
+    case failedToEncodeToJSON(Error)
+    case failedToEncodeToString
+
+}
+
+/// An object responsible for encoding an api model into a json string.
+
+struct RequestBodyEncoder {
```
markusfassbender: Same here the class seems to be unused.

WireAPI/Sources/WireAPI/Private/Components/ResponsePayloadDecoder.swift
```diff
@@ -0,0 +1,56 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+enum ResponsePayloadDecoderError: Error {
```
markusfassbender: Should we move the enum to a new file?

WireAPI/Sources/WireAPI/Private/Extensions/Date+Transport.swift
```diff
@@ -0,0 +1,41 @@
...
+
+    init?(transportString: String) {
+        if let date = ISO8601DateFormatter.default.date(from: transportString) {
+            self = date
+        } else if let date = ISO8601DateFormatter.withoutFractionalSeconds.date(from: transportString) {
+            self = date
+        } else {
+            return nil
+        }
+    }
```
markusfassbender: This duplicates the date decoding code, that has had bugs in the last time, here in the new swift package. Hm.

Wouldn't be the right approach, that we know for all APIs, that we move here to the package, what kind of `Date` format we expect by writing unit tests?

My goal would be that we get rid of this generalized helper here long term and preferably use `jsonDecoder.dateDecodingStrategy = .iso8601`. In rare cases where the backend delivers fractional seconds and it can not be changed, we can use `.custom`  with the `ISO8601DateFormatter` as workaround.

WireAPI/Sources/WireAPI/Private/Extensions/ISO8601DateFormatter+Default.swift
```diff
@@ -0,0 +1,43 @@
...
+        let dateFormatter = ISO8601DateFormatter()
+        dateFormatter.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
+        return dateFormatter
+    }()
+
+    /// ISO8601 date formatter with internet date time but no fractional seconds.
+    ///
+    /// Due to a bug in the backend timestamps without fractional seconds are received.
+    /// /// After the bug is fixed, this date formatter can be removed.
+    /// See [WPB-6529](https://wearezeta.atlassian.net/browse/WPB-6529).
```
markusfassbender: Actually https://wearezeta.atlassian.net/browse/WPB-6529 is fixed, so I am wondering if we need this at all.

WireAPI/Sources/WireAPI/Private/Extensions/JSONDecoder+Default.swift
```diff
@@ -0,0 +1,44 @@
...
+
+            guard let date = Date(transportString: rawDate) else {
+                throw DecodingError.dataCorruptedError(
+                    in: container,
+                    debugDescription: "Expected date string to be ISO8601-formatted with fractional seconds"
+                )
+            }
+
+            return date
+        }
```
markusfassbender: Same question as above - shouldn't we go with `.dateDecodingStrategy = .iso8601` by default?

WireAPI/Sources/WireAPI/Private/Extensions/JSONEncoder+Default.swift
```diff
@@ -0,0 +1,35 @@
...
+
+    static var defaultEncoder: JSONEncoder {
+        let encoder = JSONEncoder()
+        encoder.dateEncodingStrategy = .custom({ date, encoder in
+            var container = encoder.singleValueContainer()
+            try container.encode(date.transportString())
+        })
+
+        return encoder
+    }
```
markusfassbender: Same question as above - shouldn't we go with `.dateDecodingStrategy = .iso8601` by default?

WireAPI/Sources/WireAPI/Public/HTTP Client/HTTPRequest.swift
```diff
@@ -0,0 +1,73 @@
...
+
+    public var body: Data?
+
+    /// The type of the body payload.
+
+    public var contentType: String?
+
+    /// The type of the response payload.
+
+    public var acceptType: String?
```
markusfassbender: Should we have explicit fields for these values or can it be already aligned with `URLSession`?
`var allHTTPHeaderFields: [String : String?] { get set }`

WireAPI/Sources/WireAPI/Public/Models/BackendInfo.swift
```diff
@@ -0,0 +1,41 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// Metadata about the local backend.
+
+public struct BackendInfo: Equatable {
```
markusfassbender: Does it make sense to move this into a swift package `WireAPITypes` (or similar name) and share code? 
Import in `WireTransport` and `WireAPI`.

WireAPI/Sources/WireAPI/Public/Provider/APIProvider.swift
```diff
@@ -0,0 +1,50 @@
...
+        case .v3:
+            BackendInfoAPIV3(httpClient: httpClient)
+        case .v4:
+            BackendInfoAPIV4(httpClient: httpClient)
+        case .v5:
+            BackendInfoAPIV5(httpClient: httpClient)
+        case .v6:
+            BackendInfoAPIV6(httpClient: httpClient)
+        }
+    }
```
markusfassbender: Using the `APIProvider` in such way centralises all APIs in one place. When I wrote the example of an `APIService` it was critiqued it not scaling well.

Isn't it here the same? Can we have a simple object to keep it seperated?
```swift
public struct BackendInfoAPIClient: BackendInfoAPI {
    let httpClient: httpClient

    // maybe even use the `version` as property over parameters.
    public func backendInfoAPI(for version: APIVersion) -> BackendInfoAPI { ... }
}
```

wire-ios/Wire-iOS Tests/UserProfile/ProfileViewControllerTests.swift
```diff
@@ -17,9 +17,9 @@
 //
 
 import SnapshotTesting
-import XCTest
 @testable import Wire
 @testable import WireSyncEngineSupport
+import XCTest
```
markusfassbender: Does this work with the SwiftLint rule to sort imports?

WireAPI/Sources/WireAPI/Private/APIs/BackendInfoAPI/BackendInfoAPIV0.swift
```diff
@@ -0,0 +1,76 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+class BackendInfoAPIV0: BackendInfoAPI {
```
markusfassbender: For me it's confusing we have this separation of folder `Private` and `Public`, because it forces me to scroll through the package and jump in navigation.

What do you think to organise the folders by functionality like `Core`, `APIs/BackendInfo`?

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewControllerViewModel.swift
```diff
@@ -50,7 +50,7 @@ protocol ProfileViewControllerViewModeling {
     func sendConnectionRequest()
     func acceptConnectionRequest()
     func ignoreConnectionRequest()
-    // TODO: Make it async https://wearezeta.atlassian.net/browse/WPB-9028
```
netbe: is it part of this PR on purpose ?

netbe (approved):
Looks like a solid base to go further, thanks 🙏

WireAPI/Sources/WireAPI/Private/Components/PayloadPager.swift
johnxnguyen: I will take it out and whoever needs it first can include it in their PR.

WireAPI/Sources/WireAPI/Private/Components/RequestBodyEncoder.swift
johnxnguyen: I'll take this out too

WireAPI/Sources/WireAPI/Private/Components/ResponsePayloadDecoder.swift
johnxnguyen: done

WireAPI/Sources/WireAPI/Private/Extensions/Date+Transport.swift
johnxnguyen: I think the backend always delivers fractional seconds, which is why the default json decoder uses the custom date decoding strategy.

WireAPI/Sources/WireAPI/Private/Extensions/ISO8601DateFormatter+Default.swift
johnxnguyen: I will remove it then.

WireAPI/Sources/WireAPI/Private/Extensions/JSONDecoder+Default.swift
johnxnguyen: I don't think so because we need the fractional seconds don't we?

WireAPI/Sources/WireAPI/Public/HTTP Client/HTTPRequest.swift
johnxnguyen: My aim for now was to make it easier to bridge over `ZMTransportRequest` which has these explicit fields. These aren't being used at the moment so I'll delete it.

WireAPI/Sources/WireAPI/Public/Models/BackendInfo.swift
johnxnguyen: I think there would only be 2 or so models that would be shared by `WireTransport` and `WireAPI`, which is the `APIVersion` and `BackendInfo` (though this last one isn't quite the same). I don't see much benefit in adding a new package just for these two models at the moment, but certainly if there is more code to share then we could add a new package.

WireAPI/Sources/WireAPI/Public/Provider/APIProvider.swift
johnxnguyen: Yep you're right we can split it up. I'm confused though by the example, is `BackendInfoAPIClient` the client that conforms to `BackendInfoAPI` or is it a provider object that provides a `BackendInfoAPI` object?

Another idea that comes to mind is to create a static function on the protocol to provide the implementation, for instance: 

```swift
extension BackendInfoAPI {

    public static func apiClient(
        for version: APIVersion,
        httpClient: HTTPClient
    ) -> any BackendInfoAPI {
        switch version {
        case .v0:
            BackendInfoAPIV0(httpClient: httpClient)
        case .v1:
            BackendInfoAPIV1(httpClient: httpClient)
        case .v2:
            BackendInfoAPIV2(httpClient: httpClient)
        case .v3:
            BackendInfoAPIV3(httpClient: httpClient)
        case .v4:
            BackendInfoAPIV4(httpClient: httpClient)
        case .v5:
            BackendInfoAPIV5(httpClient: httpClient)
        case .v6:
            BackendInfoAPIV6(httpClient: httpClient)
        }
    }

}
```

Would this work in your opinion?

wire-ios/Wire-iOS Tests/UserProfile/ProfileViewControllerTests.swift
johnxnguyen: When I updated my base branch SwiftLint was complaining these imports were unsorted so I corrected it here.

WireAPI/Sources/WireAPI/Private/APIs/BackendInfoAPI/BackendInfoAPIV0.swift
johnxnguyen: My aim was to try make it as clear as possible what constitutes the public interface and what is an implementation detail. But I agree that there's a bit of jumping back and forth. I'm happy to reorganize by functionality but we should still be vigilant to maintain correct access control.

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewControllerViewModel.swift
johnxnguyen: When I pulled in the base branch I pulled in several swift lint violations which needed to be corrected, including this.

WireAPI/Sources/WireAPI/Public/Models/BackendInfo.swift
markusfassbender: agree, thanks

wire-ios/Wire-iOS Tests/UserProfile/ProfileViewControllerTests.swift
markusfassbender: okay I see, interesting. I think one can break it block between testable and normal imports, but it's just formatting.

WireAPI/Sources/WireAPI/Private/Components/PayloadPager.swift
markusfassbender: thank you

WireAPI/Sources/WireAPI/Private/Extensions/Date+Transport.swift
markusfassbender: Probably I mixed something up here, the implementation in `WireTransport` has also a slightly different naming.
The cleanup etc. is also something for a follow up step, so I resolve the conversation here.

WireAPI/Sources/WireAPI/Public/Provider/APIProvider.swift
markusfassbender: Yes it does. Should we name it `makeAPI(for:httpClient:)`? both would be ok for me

markusfassbender (approved):

WireAPI/Sources/WireAPI/Public/Provider/APIProvider.swift
johnxnguyen: done

WireAPI/Sources/WireAPI/Private/APIs/BackendInfoAPI/BackendInfoAPIV0.swift
johnxnguyen: done

PR #1299, opened 2024-04-16
fix: deadlock on logout WPB-7154

typfel:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7154" title="WPB-7154" target="_blank">BugWPB-7154</a>  Issues on logout
  </td></table>
  <br />
 

### Issue

When a user gets logged out from another device the app gets stuck in `ZMOperationLoop.tearDown()` because the sync ManagedObjectContext dispatch group never becomes gets empty.

So I added trace logging to all dispatch groups and was able to pin point that we fail to leave the dispatch in `ZMUserSession.processEvents()`. This happens because we enter a deadlock like this:

1. Enter sync dispatch  group
2. Process events which includes the client.deleted event which trigger the logout
3. Switch to landing screen
4. Call tearDown() on ZMUserSession which will call tearDown() on ZMOperationLoop
5. Inside ZMOperationLoop.tearDown we spin the main queue until the sync dispatch group is empty
6. We enqueue an operation on the UI ManagedObjectContext which will deadlock since we are spinning the main queue

#### Solution

Make closing a `ZMUserSession` an asynchronous operation in order to break the deadlock.

### Testing

Several integration tests needed to updated because closing / logout of a user session is now asynchronous  so we need to wait for this to complete.

---

### Checklist

### DIFF
--- a/wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack.swift
+++ b/wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack.swift
@@ -127,6 +127,7 @@ public class CoreDataStack: NSObject, ContextProvider {
     let dispatchGroup: ZMSDispatchGroup?
 
     private let migrator: CoreDataMessagingMigrator
+    private var hasBeenClosed = false
 
     public init(account: Account,
                 applicationContainer: URL,
@@ -185,6 +186,16 @@ public class CoreDataStack: NSObject, ContextProvider {
     }
 
     deinit {
+        close()
+    }
+
+    public func close() {
+        guard !hasBeenClosed  else {
+            return
+        }
+
+        defer { hasBeenClosed = true }
+
         viewContext.tearDown()
         syncContext.tearDown()
         searchContext.tearDown()
@@ -193,6 +204,7 @@ public class CoreDataStack: NSObject, ContextProvider {
     }
 
     func closeStores() {
+        WireLogger.localStorage.info("Closing core data stores")
         do {
             try closeStores(in: messagesContainer)
             try closeStores(in: eventsContainer)
--- a/wire-ios-sync-engine/Source/SessionManager/SessionManager+APIVersionResolver.swift
+++ b/wire-ios-sync-engine/Source/SessionManager/SessionManager+APIVersionResolver.swift
@@ -95,23 +95,25 @@ extension SessionManager: APIVersionResolverDelegate {
 
                 // 1. Tear down the user sessions
                 DispatchQueue.main.sync {
-                    self.tearDownBackgroundSession(for: account.userIdentifier)
-                }
-
-                // 2. Migrate users and conversations
-                CoreDataStack.migrateLocalStorage(
-                    accountIdentifier: account.userIdentifier,
-                    applicationContainer: self.sharedContainerURL,
-                    dispatchGroup: dispatchGroup,
-                    migration: {
-                        try $0.migrateToFederation()
-                    },
-                    completion: { result in
-                        if case .failure = result {
-                            log.error("Failed to migrate account for federation")
-                        }
+                    dispatchGroup.enter()
+                    self.tearDownBackgroundSession(for: account.userIdentifier) {
+                        // 2. Migrate users and conversations
+                        CoreDataStack.migrateLocalStorage(
+                            accountIdentifier: account.userIdentifier,
+                            applicationContainer: self.sharedContainerURL,
+                            dispatchGroup: dispatchGroup,
+                            migration: {
+                                try $0.migrateToFederation()
+                            },
+                            completion: { result in
+                                if case .failure = result {
+                                    log.error("Failed to migrate account for federation")
+                                }
+                            }
+                        )
+                        dispatchGroup.leave()
                     }
-                )
+                }
             }
 
             // The migration above will call enter() / leave() on the dispatch group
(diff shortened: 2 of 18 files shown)


johnxnguyen (approved):

wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack.swift
```diff
@@ -185,6 +186,16 @@ public class CoreDataStack: NSObject, ContextProvider {
     }
 
     deinit {
+        close()
+    }
+
+    public func close() {
```
markusfassbender, line 192: As an uninitiated reader I would expect that `close()` sub sequentially also calls `closeStores()`?

wire-ios-sync-engine/Source/SessionManager/SessionManager+APIVersionResolver.swift
```diff
@@ -95,23 +95,25 @@ extension SessionManager: APIVersionResolverDelegate {
 
                 // 1. Tear down the user sessions
                 DispatchQueue.main.sync {
```
markusfassbender: We are waiting for the `dispatchGroup` down the path anyway with `dispatchGroup.wait(forInterval: 5)`, so can this line be `.async` too?

agisilaos (approved):

wire-ios-sync-engine/Source/SessionManager/SessionManager+APIVersionResolver.swift
markusfassbender: In general the documentation states that one should not run sync code on the main queue at all:
https://developer.apple.com/documentation/dispatch/dispatchqueue

Screenshot 2024-04-17 at 10 08 59

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -765,31 +767,37 @@ public final class SessionManager: NSObject, SessionManagerType {
 
         self.createUnauthenticatedSession(accountId: deleteAccount ? nil : account.userIdentifier)
 
+        guard let activeUserSession = activeUserSession else {
```
markusfassbender:
```suggestion
        guard let activeUserSession else {
```

wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests+EncryptionAtRest.swift
```diff
@@ -64,19 +64,9 @@ final class ZMUserSessionTests_EncryptionAtRest: ZMUserSessionTestsBase {
...
         factory.activityManager = activityManager
-
-        setUpEARServiceWorkaround()
     }
 
-    /// This workaround is needed because all tests here are based on assumptions
-    /// that the `managedObjectContext` is changed.
-    /// To remove the workaround simply the `mockEARService` should be used instead of
-    /// a real instance of `EARService`.
-    private func setUpEARServiceWorkaround() {
```
markusfassbender: Thanks, I didn't think about overriding 😅

wire-ios/Wire-iOS/Sources/AppStateCalculator.swift
```diff
@@ -119,7 +119,7 @@ final class AppStateCalculator {
 
         self.appState = appState
         self.pendingAppState = nil
-        ZMSLog(tag: "AppState").debug("transitioning to app state: \(appState)")
+        WireLogger.appState.debug("transitioning to app state: \(appState)")
```
markusfassbender, line 122: 👍

wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests+EncryptionAtRest.swift
markusfassbender: But actually the problem itself is not solved, that the tests are testing a real `EARService` through the user session instead of using a mock.
The `EARService` can be tested itself with dedicated unit tests like `EARServiceTests.swift`.
```swift
    /// This workaround is needed because all tests here are based on assumptions
    /// that the `managedObjectContext` is changed.
    /// To remove the workaround simply the `mockEARService` should be used instead of
    /// a real instance of `EARService`.
```

typfel: Agree, but I fixed a different issue here and that is that the `coreCryptoStack` was being re-used for multiple user sessions which is not allowed anymore since the user session will tear it down when it's closed.

wire-ios-sync-engine/Source/SessionManager/SessionManager+APIVersionResolver.swift
typfel: Yeah I think we can get rid of the `sync()`

wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack.swift
typfel: we do call it

wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests+EncryptionAtRest.swift
typfel: I'll restore the comment so we don't forget about this

wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack.swift
markusfassbender: Sorry!

markusfassbender (approved):

PR #1300, opened 2024-04-16
fix: Navigation Bar Glitch regarding the color of the navigation bar

agisilaos:
**Issue**: We've been experiencing a persistent glitch where the color of the navigation bar exhibits an unintended flicker before correctly displaying the set color. This occurs when presenting a ViewController (VC) that is embedded within a Navigation Controller.

**Root Cause**: The glitch stems from our legacy approach to setting the navigation bar's color. Originally, we implemented a conditional setBackgroundColor method to explicitly set the navigation bar's color. This method included an additional check to determine if the application was running on iOS 15 or later, as setting the navigation bar's color directly was not supported in earlier versions.

**Solution**: Given that our application now targets iOS 15 as its minimum version, the compatibility check for earlier iOS versions has become redundant. Removing this check eliminates the need for the initial setBackgroundColor method to conditionally set the navigation bar's color.

By directly setting the navigation bar's color without the prior check, we successfully eliminate the color glitch, streamlining the UI experience when presenting VCs embedded in a Navigation Controller.

**Demonstration**: Attached are before and after videos showcasing the elimination of the navigation bar color glitch, highlighting the improvement in the UI presentation.

|  BEFORE  | AFTER |
|---|---|
| Simulator Screen Recording - broken | Simulator Screen Recording - iPad Pro (12 9-inch) (6th generation) - 2024-04-16 at 10 44 03 |

---

### Checklist


wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsViewController.swift
```diff
@@ -103,6 +103,7 @@ final class GroupDetailsViewController: UIViewController, ZMConversationObserver
     }
 
     private func setupNavigatiomItem() {
+        navigationController?.navigationBar.backgroundColor  = SemanticColors.View.backgroundDefault
```
caldrian:
```suggestion
        navigationController?.navigationBar.backgroundColor = SemanticColors.View.backgroundDefault
```

caldrian (approved):
👏

markusfassbender (approved):
Thanks!

PR #1301, opened 2024-04-16
fix: Location Address being truncated in default text size (when it's long) and larger text size

agisilaos:
### Issue

This PR addresses a UI issue in the LocationSendViewController where the displayed address would get truncated when the font size was increased. To resolve this, we've implemented a dynamic adjustment of the container size based on the address content length. The main changes include:

- Introduction of sendControllerHeightConstraint to manage the container's height.
- A new delegate method, locationSendViewController(_:shouldChangeHeight:), designed to determine whether the height adjustment is needed and, if so, to update the height constraint accordingly.
- Additionally, this update includes organizational improvements and code cleanup within the class for enhanced readability and maintenance.

| BEFORE | AFTER |
|---|---|
| Simulator Screen Recording - iPhone 15 Pro - 2024-04-16 at 11 33 45 | Simulator Screen Recording - iPhone 15 Pro - 2024-04-16 at 11 37 08 |

### Checklist


wire-ios/Wire-iOS/Sources/UserInterface/Location/LocationSelectionViewController.swift
```diff
@@ -105,6 +107,9 @@ final class LocationSelectionViewController: UIViewController {
             $0.translatesAutoresizingMaskIntoConstraints = false
         }
 
+        sendControllerHeightConstraint = sendController.heightAnchor.constraint(equalToConstant: 56 + UIScreen.safeArea.bottom)
```
caldrian: We should probably use the `safeAreaLayoutGuide` instead of `UIScreen` here, also below.

caldrian (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Location/LocationSelectionViewController.swift
agisilaos: Fixed it thanks!

wire-ios/Wire-iOS/Sources/UserInterface/Location/LocationSendViewController.swift
```diff
@@ -40,48 +52,82 @@ final class LocationSendViewController: UIViewController {
...
 
-        sendButton.setContentCompressionResistancePriority(UILayoutPriority(rawValue: 1000), for: .horizontal)
-        addressLabel.setContentCompressionResistancePriority(UILayoutPriority(rawValue: 750), for: .horizontal)
+        sendButton.setContentCompressionResistancePriority(.required, for: .horizontal)
+        addressLabel.setContentCompressionResistancePriority(.defaultHigh, for: .horizontal)
+        addressLabel.setContentHuggingPriority(.defaultLow, for: .horizontal)
     }
 
-    @objc
-    private func sendButtonTapped(_ sender: LegacyButton) {
```
markusfassbender, line 84: Thanks to remove `@objc` annotations! 👍

markusfassbender (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Location/LocationSendViewController.swift
agisilaos: On my mind every time I touch an already established class, thank you!

PR #1305, opened 2024-04-18
chore: delete protocol ConversationListTabBarControllerDelegate

caldrian:
Cleans up a protocol which is not needed, since the `ConversationListViewController` both owns the `delegate` property and assigns itself as the delegate.

---

### Checklist


markusfassbender (approved):
Good clean up 🗑️

KaterinaWire (approved):

PR #1309, opened 2024-04-18
refactor: api package [draft][markus]

markusfassbender:
### Issue

_Please describe the issue._

### Testing

_Describe how to test._

---

### Checklist


WireAPIService/Sources/WireAPIService/Core/APIServiceProtocol.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+
+import Foundation
+
+public protocol APIServiceProtocol {
+
+    // MARK: BackendInfo
+
+    func getBackendInfo(for version: APIVersion) async throws -> BackendInfoModel
+}
```
johnxnguyen, line 26: Would this contain all the endpoints? If so, I imagine that it would grow to be quite large which could be difficult to mock.

WireAPIService/Sources/WireAPIService/Endpoints/BackendInfo/APIService+BackendInfo.swift
```diff
@@ -0,0 +1,35 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension APIService {
+    public func getBackendInfo(for version: APIVersion) async throws -> BackendInfoModel {
+        let endpoint = Endpoints.BackendInfo()
```
johnxnguyen, line 23: I suppose if there is some different among the versions, we would also need a switch here to get the appropriate endpoint? Something like:

```swift
let endpoint: Endpoint
switch version {
case .v0:
  endpoint = Endpoints.BackendInfoV0()
case .v1:
  endpoint = Endpoints.BackendInfoV1()
case .v2, .v3, .v4, .v5, .v6:
  endpoint = Endpoints.BackendInfoV2()
}
```

WireAPIService/Sources/WireAPIService/Core/APIServiceProtocol.swift
johnxnguyen: The API clients are kind of like smaller API services, in that they provide the same interface but for a smaller collection of endpoints.

WireAPIService/Sources/WireAPIService/Endpoints/BackendInfo/APIService+BackendInfo.swift
typfel: I really think while we refactor the API layer we should also move away from switching over the API version and instead resolve once at launch and then create the suitable API classes at initialisation time.

markusfassbender: To avoid duplication of the endpoints and versions we could also just pass the api versions as configuration:
```swift
    struct BackendInfo: Endpoint {

        let version: APIVersion

        let path: String = "/api-version"

        let method: HTTPMethod = .get

        let queryItems: [URLQueryItem]? = nil

        var body: Data? {
            switch version {
            case .v0, .v1:
                return nil
            case .v2, .v3, .v4, .v5, .v6:
                return Data()
            }
        }
    }
```

markusfassbender: But if it's a completely different endpoint with different models, I would rather create a new class/struct.

markusfassbender:
This was only a showcase and can be closed now.

PR #1315, opened 2024-04-19
fix: showing onboarding hint

caldrian:
### Issue

Currently there is an issue in the initialisation of the `ConversationListViewController` which first tries to set the onboarding hint view to visible if needed in the `init` method, while afterwards in `viewDidLoad` the onboarding view is initialised to be invisible.
This PR reverses the order of these actions.

Screenshot 2024-04-19 at 5 53 29 PM

### Testing

Log into the app with a new user which has no conversations.

---

### Checklist

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
@@ -98,6 +98,9 @@ final class ConversationListViewController: UIViewController {
 
     required init(viewModel: ViewModel) {
         self.viewModel = viewModel
+        defer {
+            viewModel.viewController = self
+        }
 
         topBarViewController = ConversationListTopBarViewController(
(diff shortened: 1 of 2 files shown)


typfel (approved):

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
```diff
@@ -98,6 +98,9 @@ final class ConversationListViewController: UIViewController {
 
     required init(viewModel: ViewModel) {
         self.viewModel = viewModel
+        defer {
+            viewModel.viewController = self
+        }
```
markusfassbender: Using `defer` always makes my brain struggle a bit, because it changes the order of code execution...
I do understand it make sense to always close a file and also when an exception is thrown etc... but here a simple move to line 131 achieves the same result, right?

agisilaos (approved):

PR #1321, opened 2024-04-22
feat: update 1:1 creation flow and handle errors WPB-6938

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6938" title="WPB-6938" target="_blank">BugWPB-6938</a>  [iOS] Handle 1:1 conversations when no key packages are available
  </td></table>
  <br />
 

### Issue

With MLS 1:1 conversations, it is possible that the user we are trying to create a group with doesn't have any key packages available. In such cases, we need to inform the self user and provide a way to retry creating a group at a later time instead of silently failing and displaying a broken conversation. 

As a solution, this PR updates the 1:1 conversation flow for `MLS` conversations. If the conversation does not exist yet or if the mls group is not established, we will redirect the self user to the other user's profile and display a `"Start conversation"` button.

In the case of 1:1 conversations failing to migrate from `proteus` to `mls`, we will set the conversation as read only until the migration succeeds.

In the case of proteus conversations, we will keep the existing behaviour.

### Changes

- Set group to read only if it fails to establish during 1:1 migration
- Created use case to calculate 1:1 conversation creation status
- Used use case to refactor the UI flows for 1:1 conversations 

### Note

Disabled `ProfileViewControllerTests` because I would need to do many ugly fixes in order to make them pass. I'm choosing to refactor the tests in a separate PR instead. see https://wearezeta.atlassian.net/browse/WPB-8901

### Checklist


wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel+StartUIDelegate.swift
```diff
@@ -23,10 +23,49 @@ import WireSyncEngine
...
+
+                    // If the conversation should be using mls and is not established,
+                    // or does not exits, then we open the user profile to let the user
+                    // create the conversation
+                    await openUserProfile(user)
+
+                case .doesNotExist(protocol: .proteus):
+
+                    // If the conversation should be using proteus,
+                    // then we create the conversation and open it. (legacy behaviour)
```
typfel: Why keep the legacy behaviour for proteus? On web they now always ask before creating a 1:1 regardless if it's proteus or MLS, which in IMO is nicer.

typfel (changes requested):

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
```diff
@@ -147,12 +147,37 @@ public final class OneOnOneResolver: OneOnOneResolverInterface {
...
+            guard
+                let otherUser = ZMUser.fetch(with: userID, in: context),
+                let conversation = otherUser.oneOnOneConversation,
+                conversation.isForcedReadOnly != readOnly
+            else {
+                return
+            }
 
-        return .migratedToMLSGroup(identifier: mlsGroupID)
+            conversation.isForcedReadOnly = readOnly
```
johnxnguyen, line 179: Do we append a system message saying why the conversation is read only?

wire-ios-sync-engine/Source/Use cases/OneOnOneConversationCreationStatusUseCase.swift
```diff
@@ -0,0 +1,104 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireDataModel
+
+// sourcery: AutoMockable
+public protocol OneOnOneConversationCreationStatusUseCaseProtocol {
```
johnxnguyen: suggestion: add a doc comment explaining what the use case does.

wire-ios-sync-engine/Source/Use cases/OneOnOneConversationCreationStatusUseCase.swift
```diff
@@ -0,0 +1,104 @@
...
+}
+
+public enum OneOnOneConversationCreationStatus: Equatable {
+
+    case exists(protocol: MessageProtocol, established: Bool?)
+    case doesNotExist(protocol: MessageProtocol?)
+
+}
+
+public struct OneOnOneConversationCreationStatusUseCase: OneOnOneConversationCreationStatusUseCaseProtocol {
```
johnxnguyen: Should this be public? I would consider it an internal implementation.

wire-ios-sync-engine/Source/Use cases/OneOnOneConversationCreationStatusUseCase.swift
johnxnguyen: Also, it may be nice to give it a name that describes the function it performs, such as `GetOneOnOneConversationStatusUseCaseProtocol`

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel+StartUIDelegate.swift
```diff
@@ -23,10 +23,38 @@ import WireSyncEngine
...
+                    // or does not exits, then we open the user profile to let the user
+                    // create the conversation
+                    await openUserProfile(user)
+
+                default:
+                    break
+                }
+
+            } catch {
+
```
johnxnguyen: Is this intentionally empty?

wire-ios-sync-engine/Source/Use cases/OneOnOneConversationCreationStatusUseCase.swift
```diff
@@ -0,0 +1,104 @@
...
+public protocol OneOnOneConversationCreationStatusUseCaseProtocol {
+
+    func invoke(userID: QualifiedID) async throws -> OneOnOneConversationCreationStatus
+
+}
+
+public enum OneOnOneConversationCreationStatus: Equatable {
+
+    case exists(protocol: MessageProtocol, established: Bool?)
+    case doesNotExist(protocol: MessageProtocol?)
```
johnxnguyen: Do you need to return this much info? It seems like we only care if the one to one is ready and can be opened, or if it's not ready and needs to be started. If that's the case, you could simply return a Bool:

```swift
let isReady = try await checkOneOnOneIsReady.invoke(userID: userID)
```

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel+StartUIDelegate.swift
David-Henner: updated [ecc9d41](https://github.com/wireapp/wire-ios/pull/1321/commits/ecc9d414f9e289ed5e42a5aec6fbfceb646d1e73)

wire-ios-sync-engine/Source/Use cases/OneOnOneConversationCreationStatusUseCase.swift
David-Henner: updated https://github.com/wireapp/wire-ios/pull/1321/commits/5bf1d89609b0ea0c3c77278eb4fd6aacc0331a89

wire-ios-sync-engine/Source/Use cases/OneOnOneConversationCreationStatusUseCase.swift
David-Henner: updated https://github.com/wireapp/wire-ios/pull/1321/commits/5bf1d89609b0ea0c3c77278eb4fd6aacc0331a89

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel+StartUIDelegate.swift
David-Henner: I meant to log the error. nice catch https://github.com/wireapp/wire-ios/pull/1321/commits/5bf1d89609b0ea0c3c77278eb4fd6aacc0331a89

wire-ios-sync-engine/Source/Use cases/OneOnOneConversationCreationStatusUseCase.swift
David-Henner: I needed it before to know if the conversation was meant to use `proteus` or `mls` but, since we don't care about this anymore, I'm happy to simplify. Thanks for pointing it out 🙏 

https://github.com/wireapp/wire-ios/pull/1321/commits/5bf1d89609b0ea0c3c77278eb4fd6aacc0331a89

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
David-Henner: We don't. Not sure what copy I should use since it's not a matter of having no common supported protocols. Maybe I can reuse the copy from the alert. @typfel what do you think? Do you know what they do on the other platforms?

typfel (approved):

wire-ios-sync-engine/Source/Use cases/OneOnOneConversationCreationStatusUseCase.swift
```diff
@@ -0,0 +1,98 @@
...
+import WireDataModel
+
+// sourcery: AutoMockable
+public protocol CheckOneOnOneConversationIsReadyUseCaseProtocol {
+
+    /// Checks if there is a one-on-one conversation ready to be used for a given user.
+    /// Will return `false` if there is no conversation or if there's an `mls` conversation that isn't established
+    ///
+    /// - Parameter userID: The qualified ID of the user to check the one on one conversation for
+    /// - Returns: wether or not the one on one conversation is ready
```
johnxnguyen:
```suggestion
    /// - Returns: Whether the one on one conversation is ready.
```

johnxnguyen (approved):

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneResolver.swift
David-Henner: Since we're waiting for feedback from design, I'll resolve this conversation

PR #1332, opened 2024-04-23
chore: remove redundant AccentColor enum [WPB-7307]

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7307" title="WPB-7307" target="_blank">StoryWPB-7307</a>  [iOS] : Remove "Contacts" "Folder"
  </td></table>
  <br />
 

### Issue

We have two accent color related enums. This PR converts the `ZMAccentColor` into an Objective C class which serves as a bridge to the `AccentColor?` type.

---

### Checklist


wire-ios-utilities/Source/Public/AccentColor/AccentColor.swift
```diff
@@ -16,14 +16,19 @@
...
+    case amber
+    case turquoise
+    case purple
+}
+
+// MARK: - Default and random value
+
+extension AccentColor {
+    public static var `default`: Self { .blue }
+    public static var random: Self! { allCases.randomElement() }
```
markusfassbender, line 33: Shouldn't the force cast be inside?
```suggestion
    public static var random: Self { allCases.randomElement()! }
```

wire-ios/Wire-iOS Tests/Utils/XCTestCase+AccentColor.swift
```diff
@@ -22,30 +22,21 @@ import XCTest
 @testable import Wire
 
 private extension UIColor {
-    class var accentOverrideColor: WireDataModel.AccentColor? {
-        return ZMUser.selfUser()?.accentColorValue
+    class var accentOverrideColor: AccentColor? {
+        ZMUser.selfUser().map { .init(rawValue: $0.accentColorValue) ?? .default }
     }
```
markusfassbender: Using `.default` changes the behaviour here, right?

wire-ios/Wire-iOS/Sources/Helpers/SemanticColors.swift
```diff
@@ -269,39 +269,16 @@ enum SemanticColors {
...
-            self.init(light: Asset.Colors.red500Light, dark: Asset.Colors.red500Dark)
-        case .amber:
-            self.init(light: Asset.Colors.amber500Light, dark: Asset.Colors.amber500Dark)
-        case .turquoise:
-            self.init(light: Asset.Colors.turquoise500Light, dark: Asset.Colors.turquoise500Dark)
-        case .purple:
-            self.init(light: Asset.Colors.purple500Light, dark: Asset.Colors.purple500Dark)
+    convenience init(light: ColorAsset, dark: ColorAsset) {
+        self.init { traits in
+            traits.userInterfaceStyle == .dark ? dark.color : light.color
```
markusfassbender, line 281: One next step could be to use Xcode asset catalog directly, because it supports dark mode out of the box.

wire-ios-utilities/Source/Public/AccentColor/AccentColor.swift
caldrian: Actually I like it this way because of the following reason:
It's implicitly force-unwrapping when used like a non-optional property:
```
let randomAccentColor: AccentColor = .random
```
However, in unit tests you can test that it's not nil:
```
XCTAssertNotNil(AccentColor.random)
```
That way the tests don't crash if the implementation is wrong and nil is returned.

wire-ios/Wire-iOS/Sources/Helpers/SemanticColors.swift
caldrian: I'm not sure the mapping is 1:1 for color uses. 
Text for example could be black in light mode, but grey in dark mode. This might not apply to view backgrounds.

wire-ios-utilities/Source/Public/AccentColor/AccentColor.swift
markusfassbender: I see your points, fine for me in this case. Often I tend more to have a bit aggressive behaviour and crash right away to see the problem (maybe the unit test gets removed).

markusfassbender (approved):

wire-ios-data-model/Source/Public/ZMEditableUser.h
```diff
@@ -16,18 +16,15 @@
...
+@import Foundation;
 
 @class ZMEmailCredentials;
 @class ZMPhoneCredentials;
 
 @protocol ZMEditableUser <NSObject>
 
 @property (nonatomic, copy, nullable) NSString *name;
-@property (nonatomic) ZMAccentColor accentColorValue;
+@property (nonatomic, nullable) ZMAccentColor *zmAccentColor;
```
KaterinaWire: Why did you make it `nullable`? Do you have an example of a use case when AccentColor is nil?

wire-ios/Wire-iOS/Sources/UserInterface/Helpers/AccentColor+Name.swift
```diff
@@ -16,19 +16,17 @@
...
+
     var name: String {
         typealias AccentColor = L10n.Localizable.Self.Settings.AccountPictureGroup.AccentColor
         switch self {
         case .blue:
             return AccentColor.blue
         case .green:
             return AccentColor.green
-        case .deprecatedYellow:
-            return AccentColor.yellow
```
KaterinaWire, line 31: suggestion: I think we should also remove the localisable string

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Devices/ProfileDeviceDetailsView.swift
```diff
@@ -85,13 +85,13 @@ struct ProfileDeviceDetailsView: View {
         HStack {
             Text(L10n.Localizable.Profile.Devices.Detail.ShowMyDevice.title)
                 .padding(.all, ViewConstants.Padding.standard)
-                .foregroundColor(SemanticColors.Label.textDefault.swiftUIColor)
```
KaterinaWire, line 88: Maybe I didn't notice, but did you also remove swiftUIColor property from the UIColor extension?

wire-ios-data-model/Source/Public/ZMEditableUser.h
caldrian: Yes, I removed the enum case `undefined` which is now replaced by `nil`.

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Devices/ProfileDeviceDetailsView.swift
caldrian: Yes, it doesn't make sense, we can just call `Color(:uiColor)` directly.
```
var swiftUIColor: Color {
    return Color(uiColor: self)
}
```

wire-ios/Wire-iOS/Sources/UserInterface/Helpers/AccentColor+Name.swift
caldrian: https://github.com/wireapp/wire-ios/pull/1332/commits/1b48e0e8080709035043388307cf85db2fd12f2a

KaterinaWire (approved):

PR #1343, opened 2024-04-25
fix: quick sync never completing on a cold start WPB-8897

typfel:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8897" title="WPB-8897" target="_blank">BugWPB-8897</a>  Endless syncing bar on cold start
  </td></table>
  <br />
 

### Issue

The quick sync never completes on a cold start of the app (until you put the app into the background/foreground).

#### Cause

On a cold start we begin fetching the `/notifications` immediately before the push channel has been established. In order to safely end the quick sync the push channel has to have been connected before we start fetching `/notifications` in order to ensure that we don't miss any events. The correct thing to do in this scenario is to fetch the `/notification` again but that wasn't happening and the sync thus never completes.

#### Solution

Restart the quick sync if the push channel wasn't already open when the fetch `/notifications` began.

---

### Checklist

### DIFF
--- a/wire-ios-sync-engine/Tests/Source/UserSession/SyncStatusTests.swift
+++ b/wire-ios-sync-engine/Tests/Source/UserSession/SyncStatusTests.swift
@@ -320,6 +320,25 @@ final class SyncStatusTests: MessagingTest {
         XCTAssertEqual(sut.currentSyncPhase, .fetchingLastUpdateEventID)
     }
 
+    func testThatItRestartsQuickSyncWhenPushChannelWasOpenedAfterNotificationFetchBegan() {
+        // given
+        lastEventIDRepository.storeLastEventID(UUID.timeBasedUUID() as UUID)
+        sut.pushChannelDidOpen()
+        sut.determineInitialSyncPhase()
+        XCTAssertEqual(sut.currentSyncPhase, .fetchingMissedEvents)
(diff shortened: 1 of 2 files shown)


wire-ios-sync-engine/Tests/Source/UserSession/SyncStatusTests.swift
```diff
@@ -320,6 +320,25 @@ final class SyncStatusTests: MessagingTest {
...
+    func testThatItRestartsQuickSyncWhenPushChannelWasOpenedAfterNotificationFetchBegan() {
+        // given
+        lastEventIDRepository.storeLastEventID(UUID.timeBasedUUID() as UUID)
+        sut.pushChannelDidOpen()
+        sut.determineInitialSyncPhase()
+        XCTAssertEqual(sut.currentSyncPhase, .fetchingMissedEvents)
+        XCTAssertFalse(sut.needsToRestartQuickSync)
+
+        // then
+        XCTAssertEqual(sut.currentSyncPhase, .fetchingMissedEvents)
```
netbe: this was assert on line 328, and I don't see any side effect on sut in between?

typfel: Deleted

netbe (approved):

KaterinaWire (approved):

PR #1346, opened 2024-04-25
chore: remove contacts and folders from tab bar [WPB-7307]

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7307" title="WPB-7307" target="_blank">StoryWPB-7307</a>  [iOS] : Remove "Contacts" "Folder"
  </td></table>
  <br />
 

This PR removes contacts and folders tabs from the tab bar.
New conversations can now be created with a top right navigation bar item.
Another navigation bar item for filtering is added, but doesn't work yet.

---

### Checklist


wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Devices/UserClientCell.swift
```diff
@@ -88,7 +88,7 @@ final class UserClientCell: SeparatorCollectionViewCell {
         addBorder(for: .bottom)
         accessoryIconView.translatesAutoresizingMaskIntoConstraints = false
         accessoryIconView.contentMode = .center
-        accessoryIconView.image = .init(resource: .rightChevron).withRenderingMode(.alwaysTemplate)
```
agisilaos, line 91: This needs to stay here so the color changes the right way between light and dark mode.

wire-ios/Tests/Mocks/SwiftMockConversation+ConversationCreation.swift
```diff
@@ -21,7 +21,7 @@ extension SwiftMockConversation {
     static func oneOnOneConversation(
         otherUser: UserType = MockUser.mockUsers().first!
     ) -> SwiftMockConversation {
-        let selfUser = (MockUser.mockSelf() as Any) as! ZMUser
+        _ = (MockUser.mockSelf() as Any) as! ZMUser
```
markusfassbender: Can it be only `MockUser.mockSelf()` without type casts as they should have no side effects?

wire-ios/Wire-iOS Tests/Routers/URLActionRouterTests.swift
```diff
@@ -110,7 +110,7 @@ final class MockAuthenticatedRouter: AuthenticatedRouterProtocol {
 
     func updateActiveCallPresentationState() { }
 
-    func minimizeCallOverlay(animated: Bool, withCompletion completion: Completion?) { }
+    func minimizeCallOverlay(animated: Bool, completion: () -> Void) { }
```
markusfassbender, line 113: 🤘

wire-ios/Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -136,10 +136,9 @@
 "conversation_list.header.self_team.accessibility_value.has_new_messages" = "Has new messages.";
 
 // Conversation List Bottom Bar
-"conversation_list.bottom_bar.contacts.title" = "Contacts";
```
markusfassbender, line 139: Do we need to delete keys manually from Crowding? @netbe?

wire-ios/Wire-iOS/Sources/AuthenticatedRouter.swift
```diff
@@ -192,18 +195,8 @@ struct AuthenticatedWireFrame {
     }
 }
 
-private extension UIViewController {
-
-    func presentAlert(_ alert: UIAlertController) {
-        present(alert, animated: true, completion: nil)
-    }
-
-}
```
markusfassbender, line 201: Thanks 🚮

wire-ios/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
```diff
@@ -37,7 +37,7 @@ protocol AVSMediaManagerInterface {
 extension AVSMediaManager: AVSMediaManagerInterface {
 }
 
-typealias SettingsSelfUser = EditableUserType & UserType
+typealias SettingsSelfUser = UserType & EditableUserType
```
markusfassbender, line 40: Does the order make any difference?

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListTopBarViewController.swift
```diff
@@ -23,36 +23,26 @@ import WireSyncEngine
...
-    private weak var userStatusViewController: UserStatusViewController?
     private weak var titleViewLabel: UILabel?
 
-    /// init a ConversationListTopBarViewController
-    ///
-    /// - Parameters:
-    ///   - account: the Account of the user
-    ///   - selfUser: the self user object. Allow to inject a mock self user for testing
-    ///   - selfProfileViewControllerBuilder: a builder for the self profile view controller
     init(
```
markusfassbender, line 41: Was it not helpful to have the docs?

wire-ios/WireCommonComponents/ViewControllerBuilder.swift
```diff
@@ -19,6 +19,5 @@
 import UIKit
 
 public protocol ViewControllerBuilder {
-    associatedtype ViewController where ViewController: UIViewController
-    func build() -> ViewController
+    func build() -> UIViewController
 }
```
markusfassbender, line 23: I assume this change was made because you didn't need the have the specific return type?
In this case it seems good and not over engineered 👍
But if one needs to set a delegate or property on the view controller, doesn't make sense to get the correct type?

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListTopBarViewController.swift
caldrian: I don't consider it helpful. The names of the arguments are self-explanatory and the description here didn't actually add more infos.

wire-ios/WireCommonComponents/ViewControllerBuilder.swift
caldrian: Exactly, I didn't want to over-engineer it. This can of course be adjusted whenever something different is needed.

wire-ios/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
caldrian: No, but `UserType & EditableUserType` is used on other places in the code. As such this place can be found more easily when using a text-based search.

wire-ios/Wire-iOS/Sources/AuthenticatedRouter.swift
caldrian: 🙏

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Devices/UserClientCell.swift
caldrian: Since the rendering mode was set each time this image asset is used, I set the rendering mode in the asset catalog instead and removed it from code.
The end result should be the same, please correct me if I'm wrong.

wire-ios/Wire-iOS/Resources/Base.lproj/Localizable.strings
agisilaos: No @markusfassbender.

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/Devices/UserClientCell.swift
agisilaos: Ok I missed that then it's good to go! 🎉

agisilaos (approved):

markusfassbender (approved):

PR #1363, opened 2024-04-30
docs: add documentation about certificate pinning - WPB-8675

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8675" title="WPB-8675" target="_blank">TaskWPB-8675</a>  Check Certificate pinning
  </td></table>
  <br />
 

### Issue

Add documentation related to certificate pinning, referring to https://github.com/wireapp/wire-ios/pull/1352

---

### Checklist


wire-ios-transport/WireTransport.docc/Certificate Pinning.md
```diff
@@ -0,0 +1,23 @@
+# Certificate pinning
+
+Verify identity of Backend servers by checking certificates.
```
johnxnguyen:
```suggestion
Verify the identity of backend servers by checking certificates.
```

wire-ios-transport/WireTransport.docc/Certificate Pinning.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Overview
+
+In order to make sure we're talking to the right server, we use certificate pinning. This means that we check the certificate of the server we're talking to against a known certificate. If the certificates match, we know we're talking to the right server.
```
johnxnguyen:
```suggestion
Certificate pinning is a process to ensure a client is communicating with a trusted server. To achieve this, the client compares a certificate obtained from a server with a known certificate that it has stored locally. If the certificates match then the server is considered trustworthy.
```

wire-ios-transport/WireTransport.docc/Certificate Pinning.md
```diff
@@ -0,0 +1,23 @@
...
+
+
+Here, we only verify the public key of the certificate that way when server changes its certificate (and keeping the same public key), we won't have to update all clients. We also check the certificate downloaded from the server is not expired.
```
johnxnguyen:
```suggestion
One approach of certificate pinning involves comparing the entire certificate. The downside to this is that when the server eventually updates their certificate, then also the client must update its local certificate to continue to trust that server. This, in effect, means that older clients will no longer trust the server.

To avoid this issue, Wire employs a second approach, namely to verify only the public key of the certificate. In this way, if the server renews its certificate with the same public-private key pair, then all existing clients that have the same public key will continue to trust the server, even with new certificates. Specifically, a server is considered trusted when its certificate is not expired and its public key matches the known and trusted public key.
```

johnxnguyen (approved):
Just added some suggestions

markusfassbender (approved):

PR #1365, opened 2024-04-30
refactor: ProfileViewController tests WPB-8901

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8901" title="WPB-8901" target="_blank">TaskWPB-8901</a>  [iOS] Refactor snapshot testing logic of `ProfileViewControllerTests`
  </td></table>
  <br />
 

### Issue

`ProfileViewController` is coupled with its view model `ProfileViewControllerViewModel`. There is a way to inject the view model but there is no way to mock it. This results in the snapshot tests testing both the view and the view model. 

### Changes

- Refactoring of `ProfileViewController`
  - Decoupled `ProfileViewController` from `ProfileViewControllerViewModel` 
  - Rendered the view model mockable by making it conform to `ProfileViewControllerViewModeling`
  - Mocked `ProfileViewControllerViewModeling` with Sourcery
  - Updated the snapshots to test several UI states of the `ProfileViewController` instead of both the view controller and view model
  - Removed unused methods of `ProfileViewController`
- Refactoring of `ProfileViewControllerViewModel`
  - Made `ProfileActionsFactory` conform to `ProfileActionsFactoryProtocol` with Sourcery
  - Injected it as dependency into the view model
  - Migrated additional dependencies Mocks to Sourcery
  - Added partial test coverage to `ProfileViewControllerViewModel`
  - Documented further potential changes in  https://wearezeta.atlassian.net/browse/WPB-8959

### Checklist


wire-ios/Wire-iOS Tests/UserProfile/ProfileViewControllerViewModelTests.swift
```diff
@@ -0,0 +1,180 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import XCTest
+@testable import Wire
+@testable import WireDataModelSupport
+
+final class ProfileViewControllerViewModelTests: XCTestCase {
```
markusfassbender, line 24: `: XCTestCase` nice 🙌

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController.swift
```diff
@@ -47,26 +47,17 @@ extension ZMConversationType {
 }
 
 final class ProfileViewController: UIViewController {
-    let viewModel: ProfileViewControllerViewModel
     weak var viewControllerDismisser: ViewControllerDismisser?
+    var delegate: ProfileViewControllerDelegate?
```
markusfassbender: Usually delegates are `weak` and not retained, can we apply it here?

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController.swift
```diff
@@ -101,7 +101,7 @@ final class ProfileViewController: UIViewController {
         self.viewControllerDismisser = viewControllerDismisser
     }
 
-    required init(viewModel: ProfileViewControllerViewModel) {
+    required init(viewModel: ProfileViewControllerViewModeling) {
```
markusfassbender:
```suggestion
    required init(viewModel: any ProfileViewControllerViewModeling) {
```

markusfassbender:
Should the PR merge to the release branch or maybe better `develop`?

wire-ios/Tests/TestPlans/AllTests.xctestplan
```diff
@@ -52,7 +52,6 @@
         "AVURLAsset_conversionTests\/testThatVideoIsConvertedToUploadFormat()",
         "ConversationListTopBarViewControllerSnapshotTests\/testForLongName()",
         "MessageReplyPreviewViewTests\/testThatItRendersImageMessagePreview()",
-        "ProfileViewControllerTests",
```
markusfassbender, line 55: Shouldn't the new tests be added to the testplan?

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController.swift
```diff
@@ -624,4 +616,8 @@ extension ProfileViewController: ProfileViewControllerViewModelDelegate {
     func stopAnimatingActivity() {
         activityIndicator.stopAnimating()
     }
+
+    func transition(to conversation: ZMConversation) {
+        delegate?.profileViewController(self, wantsToNavigateTo: conversation)
```
markusfassbender: I think this `ProfileViewControllerDelegate` is needed to manage navigation, isn't it?

From my understanding this might be better to have the logic in view model and call it from here:
`viewModel.navigateTo(conversation)`.

The view model can have a closure `var onNavigation(to conversation: Conversation)`, that is set from the outer world, or another delegate for navigation only.

wire-ios/Tests/TestPlans/AllTests.xctestplan
David-Henner: I've just removed them from the skipped tests list

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController.swift
David-Henner: Agreed. I originally just wanted the view model to delegate the navigation _somewhere_ in order to mock it (the delegate) for tests. 

Where would you set the closure from? I guess from where we have an instance of the VM, but that would be the VC. Wouldn't it be the same as having that navigation logic in the VC with extra steps?

Or, having another delegate could work, I was thinking of something like a router. WDYT?

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController.swift
David-Henner: Good catch

David-Henner:
> Should the PR merge to the release branch or maybe better `develop`?

It's a follow-up PR to https://github.com/wireapp/wire-ios/pull/1321 which was merge to release branch. I should cherry-pick the commit to develop first, then rebase this PR on develop

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController.swift
```diff
@@ -47,26 +47,17 @@ extension ZMConversationType {
 }
 
 final class ProfileViewController: UIViewController {
-    let viewModel: ProfileViewControllerViewModel
     weak var viewControllerDismisser: ViewControllerDismisser?
+    weak var delegate: ProfileViewControllerDelegate?
 
+    private var viewModel: ProfileViewControllerViewModeling
```
netbe:
```suggestion
    private let viewModel: ProfileViewControllerViewModeling
```

Do we replace the viewModel at some point?

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController.swift
netbe: without more structure, I would use convenience init to setup the viewModel and the closure for navigation.

Otherwise, move the logic to a separate object like a Builder

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewControllerViewModel.swift
```diff
@@ -30,38 +30,69 @@ enum ProfileViewControllerContext {
...
+    var blockTitle: String? { get }
+    var allBlockResult: [BlockResult] { get }
+    var hasLegalHoldItem: Bool { get }
+    var incomingRequestFooterHidden: Bool { get }
+
+    func updateActionsList()
+    func sendConnectionRequest()
+    func acceptConnectionRequest()
+    func ignoreConnectionRequest()
+    func cancelConnectionRequest(completion: @escaping Completion)
```
netbe, line 54: can we mordernize this using async?

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController.swift
markusfassbender:
```
protocol ProfileViewControllerViewModeling {
// ...
/// set from the outer world to communicate
var onNavigationToConversation: ((ZMConversation) -> Void)? { get set }
/// called from the view to forward to the closure
func navigate(to conversation: ZMConversation)
}
```

But I see the problem now, the caller side doesn't create a `viewModel` and runs `ProfileViewController(viewModel: viewModel)`, but we have a lot of convenience init and stuff around, that is not easy to refactor. 
An example is `UserDetailViewControllerFactory.createUserDetailViewController`.

Not sure if we just keep it as it is (one can understand it well) or pass a closure through all the inits instead of the delegate...

markusfassbender (approved):

johnxnguyen (approved):

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewControllerViewModel.swift
David-Henner: seems a bit out of the scope so I created a ticket instead https://wearezeta.atlassian.net/browse/WPB-9028

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController.swift
David-Henner: No but the compiler didn't like that I was setting the delegate property directly. I chose to use a `setDelegate` method instead

David-Henner: https://github.com/wireapp/wire-ios/pull/1365/commits/c8e57081d54b03ce21f7cd4b9c318faab9519978

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController.swift
David-Henner: https://github.com/wireapp/wire-ios/pull/1365/commits/c8e57081d54b03ce21f7cd4b9c318faab9519978

David-Henner: went with the closure solution

PR #1366, opened 2024-04-30
refactor: remove UseCaseFactory

markusfassbender:
### Issue

Remove `UseCaseFactory`. I would like to avoid that the `UseCaseFactory` becomes another big collection of dependencies that are required to create all use cases.

I think it's better to inject the mockable root dependencies or a use case protocol (like `ResolveOneOnOneConversationsUseCaseProtocol`) directly to the place where the use case is invoked. A use case shouldn't be too complex and be created with 1, 2 or 3 dependencies in place.

### Testing

- Build
- Run unit tests suite

---

### Checklist


caldrian (approved):

wire-ios-sync-engine/Source/Synchronization/Strategies/UserClientEventConsumer.swift
```diff
@@ -34,12 +34,13 @@ public class UserClientEventConsumer: NSObject, ZMEventAsyncConsumer {
         managedObjectContext: NSManagedObjectContext,
         clientRegistrationStatus: ZMClientRegistrationStatus,
         clientUpdateStatus: ClientUpdateStatus,
-        resolveOneOnOneConversations: ResolveOneOnOneConversationsUseCaseProtocol
+        resolveOneOnOneConversations: any ResolveOneOnOneConversationsUseCaseProtocol
```
johnxnguyen, line 37: What is the difference when adding `any`?

johnxnguyen (approved):

wire-ios-sync-engine/Source/Synchronization/Strategies/UserClientEventConsumer.swift
markusfassbender: The best documentation is actually the original proposal with the downsides and solution approach:
https://github.com/apple/swift-evolution/blob/main/proposals/0335-existential-any.md#introduction

Short summary I copied from multiple paragraphs:
> A plain protocol name in type context means an existential type.
> Existential types in Swift have significant limitations and performance implications.
> Existential types are also significantly more expensive than using concrete types.
> The cost of using existential types should not be hidden, and programmers should explicitly opt into these semantics.
> I propose to make existential types syntactically explicit in the language using the any keyword.

It got introduced as optional in Swift 5.6, the plan was to enforce it by the compiler with Swift 6, but it seems that it was deferred to a later release.

caldrian (approved):

caldrian:
CI got stuck in the SyncEngine tests and was aborted after running more than 6 hours. Something might be wrong.

PR #1369, opened 2024-04-30
chore: clean up unused launch-screen xibs [WPB-7306]

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7306" title="WPB-7306" target="_blank">StoryWPB-7306</a>  [iOS] : Add “Conversation” , “Archive” , “Settings”
  </td></table>
  <br />
 

This PR 
- removes launch screen files which aren't needed
- configures the main app to not need any launch screen xib or storyboard, but use the `UILaunchScreen` configuration
- replaces the old wire logo shield image with an updated one

---

### Checklist


johnxnguyen (approved):

agisilaos (approved):

wire-ios/Wire-iOS/Wire-Info.plist
```diff
@@ -154,8 +152,15 @@
...
-	<string>LaunchScreen</string>
+	<key>UILaunchScreen</key>
+	<dict>
+		<key>UIColorName</key>
+		<string>Black</string>
+		<key>UIImageName</key>
+		<string>Wire/Shield</string>
+		<key>UIImageRespectsSafeAreaInsets</key>
+		<true/>
+	</dict>
```
markusfassbender, line 163: I didn't know it's possible, so it just shows a black screen with a centred image?
https://developer.apple.com/documentation/bundleresources/information_property_list/uilaunchscreen

markusfassbender (approved):

wire-ios/Wire-iOS/Sources/UserInterface/LaunchImage/LaunchImageViewController.swift
```diff
@@ -23,7 +23,7 @@ class LaunchImageViewController: UIViewController {
 
     private var shouldShowLoadingScreenOnViewDidLoad = false
 
-    private var contentView: UIView!
+    private weak var contentView: UIView!
```
netbe, line 26: why weak here?

caldrian: Probably it doesn't make any difference, but it is not needed either. When added as subview the superview keeps a strong reference. ([Line 51](https://github.com/wireapp/wire-ios/pull/1369/files#diff-c3caaff90f1eab72fef9db70f02d046dff570c04d6c0ca7cc28083f68b96015eR51))

markusfassbender: I agree with Christoph, when using programatically created `UIKit` in the past, I always used `weak var contentView: UIView?` or sometimes `weak var contentView: UIView!` (you know you set it up on init or loadView).

PR #1370, opened 2024-05-02
chore: increase snapshot precision to 1

johnxnguyen:
### Issue

Our snapshot tests precision setting 0.9 which means when comparing a snapshot to a reference, only 90% of pixels need to match to consider the snapshot as valid. This is very low and as a result a snapshot that appears quite different can still pass the test. 

I have removed this setting so the default of 1 is used, meaning all pixels must match. We still have a `peceptualPrecision` of 0.98 which compares each pixel with 98% accuracy.

### Testing

No new tests need, just ensure the CI passes.

---

### Checklist


markusfassbender:
> Failed Tests (160)
I guess we need to re-record all of them now?

johnxnguyen:
> > Failed Tests (160)
> > I guess we need to re-record all of them now?

@markusfassbender it seems like the failing tests are recording images larger than the reference. The references seem to be 2/3 the size of the failed snapshot, so I guess it's somehow related to 2x and 3x resolutions?

johnxnguyen:
closing because of conflicts, will open a new PR.

PR #1372, opened 2024-05-02
chore: add option to skip security_tests on C builds workflows - WPB-8998

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8998" title="WPB-8998" target="_blank">BugWPB-8998</a>  CI failing on C builds
  </td></table>
  <br />
 

### Issue

In order to unblock the [error](https://github.com/wireapp/wire-ios/actions/runs/8920070925) (caused by the change of image from [cirrus-ci](https://github.com/cirruslabs/macos-image-templates/issues/147)) from workflow  to build apps with security_tests, we add an option to skip them.

The option to skip is set to true in the template for build_app.
For C Restricted builds, we add an option to skip or not the security_tests, the option is false by default.
For C Production builds, we never skip the security_tests

Screenshot 2024-05-02 at 12 26 13

### Notes

I think the part on the fastlane file could be improved, but that would need another PR. The concerns about the other workflows (playground) are resolved now
---

### Checklist


netbe:
Superseeds #1373

markusfassbender (approved):

caldrian (approved):

PR #1375, opened 2024-05-03
fix: calls crash on simulator video [WPB-8954]

markusfassbender:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8954" title="WPB-8954" target="_blank">BugWPB-8954</a>  iOS: calls crash on iOS 17 simulator
  </td></table>
  <br />
 

### Issue

The app crashes on iOS 17 simulators.

The solution creates a video container as simple view, that is added initially. Only when the video setup is required, the real `AVSVideoView` is created and embedded.

So then it's possible to do audio only calls, but video calls still crash when the video rendering activates.

### Testing

- Tests call between iOS 16 and 17 simulators
- Test real devices and simulators
- Test real between devices
- Start calls with audio only
- Start calls with video
- Receive calls with audio only
- Receive calls with video

---

### Checklist


wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallParticipantViews/CallParticipantView.swift
```diff
@@ -161,4 +157,29 @@ final class CallParticipantView: BaseCallParticipantView {
...
+    // MARK: Override Base
+
+    override func updateVideoShouldFill(_ shouldFill: Bool) {
+        if shouldFill, videoView == nil {
+            // [WPB-8954] Setup video only when the video really starts to avoid
+            // calls crashing on the iOS 17 simulator.
+            let videoView = makeVideoView()
+            self.videoView = videoView
+
+            videoContainerView?.setupVideoView(videoView)
```
johnxnguyen, line 170: So will this still crash when video is enabled?

markusfassbender: Unfortunately yes. It just makes it possible to have audio only calls in the simulator again (as on iOS 16 simulators it was working).

wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallParticipantViews/AVSVideoContainerView.swift
```diff
@@ -0,0 +1,58 @@
...
+import UIKit
+import struct WireSystem.WireLogger
+
+/// A placeholder container for AVSVideo to start the rendering only if the view is instantiated and setup.
+final class AVSVideoContainerView: UIView {
+
+    private weak var videoView: UIView?
+
+    init() {
+        super.init(frame: .zero)
```
caldrian: I think the default implementation of `init()` does exactly the same.

wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallParticipantViews/AVSVideoContainerView.swift
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @available(*, unavailable)
+    required init?(coder: NSCoder) {
```
caldrian: Maybe both inits can be removed.

wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallParticipantViews/BaseCallParticipantView.swift
```diff
@@ -224,7 +224,11 @@ class BaseCallParticipantView: OrientableView, AVSIdentifierProvider {
     private func updateFillMode() {
         // Reset scale if the view was zoomed in
         scalableView?.resetScale()
-        videoView?.shouldFill = shouldFill
+        updateVideoShouldFill(shouldFill)
+    }
+
+    func updateVideoShouldFill(_ shouldFill: Bool) {
+        // TODO: override in subsclasses
```
caldrian:
```suggestion
        fatalError("override in subsclass")
```

caldrian (approved):
👏

wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallParticipantViews/BaseCallParticipantView.swift
markusfassbender: I had this before, but some unit tests uses this base view through several layers 🤷‍♂️ So I turned it into a comment..

johnxnguyen (approved):

PR #1379, opened 2024-05-03
feat: configurable cipher-suite [follow up] WPB-8591

typfel:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8591" title="WPB-8591" target="_blank">StoryWPB-8591</a>  [iOS] Add support for ECDSA ciphersuites
  </td></table>
  <br />
 

### Issue

There a number of issues with the configurable cipher-suite:

1. Using wrong tags when encoding/decoding mls public keys
2. Didn't claim key-packages with correct cipher-suite 
3. We should only pass the external sender matching the cipher-suite (and not all available).

---

### Checklist


wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -544,7 +548,8 @@ public final class MLSService: MLSServiceInterface {
         do {
             logger.info("adding members to group (\(groupID.safeForLoggingDescription)) with users: \(users)")
             guard !users.isEmpty else { throw MLSAddMembersError.noMembersToAdd }
-            let keyPackages = try await claimKeyPackages(for: users)
+            guard let ciphersuite = MLSCipherSuite(rawValue: await featureRepository.fetchMLS().config.defaultCipherSuite.rawValue) else { throw MLSAddMembersError.noMembersToAdd}
```
netbe:
```suggestion
            let mlsConfig = await featureRepository.fetchMLS().config
            guard let ciphersuite = MLSCipherSuite(rawValue: mlsConfig.defaultCipherSuite.rawValue) else {
                throw MLSAddMembersError.noMembersToAdd
            }
```

it might be easier to read on multiple lines

wire-ios-data-model/Source/MLS/SafeCoreCrypto.swift
```diff
@@ -38,11 +38,9 @@ public class SafeCoreCrypto: SafeCoreCryptoProtocol {
     private let databasePath: String
 
     public convenience init(path: String, key: String) async throws {
-        // NOTE: the ciphersuites argument is not used here and will eventually be removed. WPB-8921
```
netbe, line 41: 🎉

wire-ios-data-model/Source/Model/UserClient/UserClient.swift
```diff
@@ -454,6 +450,14 @@ public extension UserClient {
             selfUser.needsToAcknowledgeLegalHoldStatus = true
         }
 
+        if !client.isSelfClient() {
```
netbe, line 453: why don't we pass the keys on selfClient?

netbe (commented):
added some questions

johnxnguyen (approved):

wire-ios-data-model/Source/Model/UserClient/UserClient.swift
typfel: Because we generate them locally and send them to the backend, it never makes sense to retrieve then from the backend that only opens up for things going out of sync.

netbe: makes sense

netbe (approved):

PR #1385, opened 2024-05-04
chore: delete extensions to Optional [WPB-8907]

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8907" title="WPB-8907" target="_blank">TaskWPB-8907</a>  Migrate Objective C code to Swift
  </td></table>
  <br />
 

This PR removes various helpers which are actually confusing and don't provide much value.

---

### Checklist


wire-ios-data-model/Source/MLS/MLSUserID.swift
```diff
@@ -48,18 +48,8 @@ public struct MLSUserID {
...
-    ) {
-        guard
-            userID.isNonEmpty,
-            domain.isNonEmpty
-        else {
-            return nil
-        }
-
+    public init?(userID: String, domain: String) {
+        guard !userID.isEmpty, !domain.isEmpty else { return nil }
```
markusfassbender: Maybe that's easier to read?
```suggestion
        if userID.isEmpty || domain.isEmpty { return nil }
```

wire-ios-request-strategy/Sources/Request Strategies/Assets/Helpers/AssetDownloadRequestFactory.swift
```diff
@@ -23,21 +23,21 @@ public final class AssetDownloadRequestFactory: NSObject {
 
     public func requestToGetAsset(withKey key: String, token: String?, domain: String?, apiVersion: APIVersion) -> ZMTransportRequest? {
         let path: String
+        let domain = domain?.isEmpty == false ? domain! : BackendInfo.domain
```
markusfassbender: `domain?.isEmpty == false` can it be `true` if domain is `nil`? I guess no, but not completely clear to me..

Do you like this variant?
```suggestion
        let resolvedDomain: String
        if let domain, !domain.isEmpty {
            resolvedDomain = domain
        } else {
            resolvedDomain = BackendInfo.domain
        }
```

wire-ios-request-strategy/Sources/Request Strategies/Client Message/ClientMessageRequestFactory.swift
```diff
@@ -51,17 +52,21 @@ public final class ClientMessageRequestFactory: NSObject {
                 nativePush: false,
                 recipients: []
             )
+
         case .v1, .v2, .v3, .v4, .v5, .v6:
-            guard let domain = domain.nonEmptyValue ?? BackendInfo.domain else {
+            let domain = domain?.isEmpty == false ? domain! : BackendInfo.domain
```
markusfassbender: Same as above

wire-ios-request-strategy/Sources/Request Strategies/Connection/Actions/ConnectToUserActionHandler.swift
```diff
@@ -54,18 +54,19 @@ class ConnectToUserActionHandler: ActionHandler<ConnectToUserAction> {
     }
 
     func federatedRequest(for action: ActionHandler<ConnectToUserAction>.Action, apiVersion: APIVersion) -> ZMTransportRequest? {
-        guard
-            apiVersion > .v0,
-            let domain = action.domain.nonEmptyValue ?? BackendInfo.domain
-        else {
+
+        let domain = action.domain?.isEmpty == false ? action.domain! : BackendInfo.domain
```
markusfassbender: Same as above

wire-ios-utilities/Source/Public/Map+KeyPath.swift
```diff
@@ -43,13 +43,3 @@ extension Sequence {
...
-
-extension Optional {
-    public func map<Value>(_ keyPath: KeyPath<Wrapped, Value>) -> Value? {
-        return map { $0[keyPath: keyPath] }
-    }
-
-    public func flatMap<Value>(_ keyPath: KeyPath<Wrapped, Value?>) -> Value? {
-        return flatMap { $0[keyPath: keyPath] }
-    }
-}
```
markusfassbender, line 55: Thanks 🙌

wire-ios-utilities/Source/Public/Optional+Apply.swift
```diff
@@ -1,30 +0,0 @@
...
-extension Optional {
-
-    /// Like map, but intended to be used to perform side effects.
-    /// Basically what `forEach` on `Collection` is compared to `map`, but for `Optional`.
-    /// - parameter block: The closure to be executed in case self holds a value.
-    public func apply(_ block: (Wrapped) -> Void) {
-        if case .some(let unwrapped) = self {
-            block(unwrapped)
-        }
-    }
```
markusfassbender, line 28: finally! 👍

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/SimpleTextField.swift
```diff
@@ -163,8 +159,14 @@ final class SimpleTextField: UITextField, DynamicTypeCapable {
...
 // MARK: SimpleTextField Extension
 extension SimpleTextField: SimpleTextFieldValidatorDelegate {
-    func textFieldValueChanged(_ value: String?) {
-        textFieldDelegate?.textField(self, valueChanged: value.value ?? .error(.empty))
+
+    func textFieldValueChanged(_ text: String?) {
+        let validator = SimpleTextFieldValidator()
+        let newValue = { () -> SimpleTextField.Value in
+            guard let text else { return .error(.empty) }
+            return if let error = validator.validate(text: text) { .error(error) } else { .valid(text) }
```
markusfassbender: This formatting would align with line 46:
```suggestion
            if let error = validator.validate(text: text) {
                return .error(error)
            } else { 
                return .valid(text)
            }
```

markusfassbender (approved):
🚮
Just some formatting and the possible decision between `if` and `guard`.

wire-ios-request-strategy/Sources/Request Strategies/Assets/Helpers/AssetDownloadRequestFactory.swift
```diff
@@ -19,25 +19,25 @@
...
-        let path: String
 
+        let domain = if let domain, !domain.isEmpty { domain } else { BackendInfo.domain }
+        let path: String
         switch apiVersion {
         case .v0:
             path = "/assets/v3/\(key)"
         case .v1:
-            guard let domain = domain.nonEmptyValue ?? BackendInfo.domain else { return nil }
+            guard let domain else { return nil }
```
KaterinaWire, line 32: I think it's not 100% the same. We also check `BackendInfo.domain` if domain is nil.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/UpdateAccessRolesActionHandler.swift
```diff
@@ -51,19 +51,24 @@ final class UpdateAccessRolesActionHandler: ActionHandler<UpdateAccessRolesActio
...
                                       payload: payloadAsString as ZMTransportData?,
                                       apiVersion: apiVersion.rawValue)
+
         case .v1, .v2, .v3, .v4, .v5, .v6:
-            guard let domain = conversation.domain.nonEmptyValue ?? BackendInfo.domain else { return nil }
-            return ZMTransportRequest(path: "/conversations/\(domain)/\(conversationID)/access",
-                                      method: .put,
-                                      payload: payloadAsString as ZMTransportData?,
-                                      apiVersion: apiVersion.rawValue)
+            let domain = conversation.domain?.isEmpty == false ? conversation.domain! : BackendInfo.domain
```
KaterinaWire: Could we avoid force unwrapping a` conversation.domain`?

markusfassbender: I think the best would be to make sure when we set the `domain` to make sure it is not empty and if it is, set it to nil. Then we could skip all the validations on reading, but not sure if this is feasible to refactor now.

Alternatively I think this could be a valid case for an extension `var resolvedDomain: String`, how is your opinion?

caldrian: My opinion is that we shouldn't use `Optional` if the actual type already has a proper special value. In this case an empty string could be our way to express "no domain". Then these helpers converting empty string to nil wouldn't be required.

KaterinaWire (approved):

PR #1386, opened 2024-05-05
Xcode selective testing

mikeger:
### Issue

Draft integration of https://github.com/mikeger/XcodeSelectiveTesting for the benefit of Wire-iOS.

I've not been able yet to properly test those changes on the CI (which means they are probably going to fail).

### Testing

---

### Checklist


.github/workflows/test_pr_changes.yml
```diff
@@ -12,6 +12,129 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  detect-changes-new:
```
mikeger, line 15: This workflow will run in parallel with the existing one in order to compare them. It's a copy of `run-tests` workflow from `_reusable_run_tests.yaml`.

.github/workflows/test_pr_changes.yml
```diff
@@ -12,6 +12,129 @@ concurrency:
...
+        run: |
+          ./setup.sh
+      
+      - name: Setup simulator
+        run: |
+          bundle exec fastlane prepare_for_tests
+      
+      - name: Setup selective testing
+        run: |
+          mint run mikeger/XcodeSelectiveTesting@0.9.3
```
mikeger: Running this would update the `AllTests` test plan, disabling unnecessary test targets.

.xcode-selective-testing.yml
```diff
@@ -0,0 +1,2 @@
+basePath: wire-ios-mono.xcworkspace
+testPlan: AllTests.xctestplan
```
mikeger, line 2: Configuration file for XcodeSelectiveTesting is not a necessity but a convenience. Same parameters can be passed via CLI

wire-ios/Wire-iOS.xcodeproj/xcshareddata/xcschemes/Wire-iOS.xcscheme
```diff
@@ -20,6 +20,244 @@
                ReferencedContainer = "container:Wire-iOS.xcodeproj">
             </BuildableReference>
          </BuildActionEntry>
+         <BuildActionEntry
```
mikeger, line 23: This is the only change that might cause trouble. In Wire-iOS scheme, I am marking all test targets to be compiled for the test build action. This would slow down running local tests. An alternative would be creating a new scheme specifically for selective testing. I was not bold enough to start managing schemes in the project.

.github/workflows/test_pr_changes.yml
mikeger: In a final integration (if you find this useful) it would be necessary to cache the mint build folder to save the tool's compilation time.

.github/workflows/test_pr_changes.yml
```diff
@@ -12,6 +12,129 @@ concurrency:
...
+        run: |
+          ./setup.sh
+      
+      - name: Setup simulator
+        run: |
+          bundle exec fastlane prepare_for_tests
+      
+      - name: Setup selective testing
+        run: |
+          mint run mikeger/XcodeSelectiveTesting@0.9.3 --base-branch $GITHUB_BASE_REF
```
mikeger, line 68: Running this would update the AllTests test plan, disabling unnecessary test targets.
In a final integration (if you find this useful) it would be necessary to cache the mint build folder to save the tool's compilation time.

.xcode-selective-testing.yml
markusfassbender: Interesting tool, thanks! How does `XcodeSelectiveTesting` calculate the dependency graph? As far as I can see from the [Documentation](https://github.com/mikeger/XcodeSelectiveTesting?tab=readme-ov-file#2-build-the-dependency-graph) it uses SPM, but we have most third-party dependencies in `Carthage` and the local ones not split in Packages.

mikeger: Hey @markusfassbender , thanks for the reply! The tool has support for SPM packages and for Xcode projects. It's working well with Wire monorepo setup. I will probably create a short video to showcase the integration in the upcoming days.

netbe:
thks @mikeger, for now we'll go with #1707, we'll re-evaluate later if needed

PR #1391, opened 2024-05-06
fix: assign cipher-suite when creating MLS groups WPB-8591

typfel:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8591" title="WPB-8591" target="_blank">StoryWPB-8591</a>  [iOS] Add support for ECDSA ciphersuites
  </td></table>
  <br />
 

### Issue

The cipher-suite is fixed after the first commit when creating an MLS conversation. Due to this and the fact that in [API V5 the backend always returns a cipher-suite when fetching a conversation](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/1035632650/API+changes+v5+v6) we need to:

1. Only assign the cipher-suite if the epoch > 0
2. Assign the cipher-suite locally when we are creating the underlying MLS group.

### Testing

1. Create an MLS conversation
2. Check in the conversation details that the cipher-suite is correct

---

### Checklist

### DIFF
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -25,16 +25,16 @@ public protocol MLSServiceInterface: MLSEncryptionServiceInterface, MLSDecryptio
 
     func uploadKeyPackagesIfNeeded() async
 
-    func createSelfGroup(for groupID: MLSGroupID) async
+    func createSelfGroup(for groupID: MLSGroupID) async throws -> MLSCipherSuite
 
     func joinGroup(with groupID: MLSGroupID) async throws
 
     /// Join group after creating it if needed
     func joinNewGroup(with groupID: MLSGroupID) async throws
 
-    func establishGroup(for groupID: MLSGroupID, with users: [MLSUser]) async throws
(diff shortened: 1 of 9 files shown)


wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -829,8 +835,11 @@ public final class MLSService: MLSServiceInterface {
             return
         }
 
+        // TODO: [WPB-9029] jacob this looks wrong,
+        // why would we create the MLS group if doesn't exist? We are about
+        // to join it via external commit.
         if await !conversationExists(groupID: groupID) {
```
netbe, line 841: isn't it when we join a group via deeplink this is needed if I recall correctly?

netbe (approved):

wire-ios-data-model/Source/MLS/MLSService.swift
typfel: Yes, this method is only used when joining via a deep link. But how can you share a deep link to a conversation which hasn't already been created?

johnxnguyen (approved):

PR #1394, opened 2024-05-07
feat: Opening a secure guest link  - Knowing if this is a secure guest link - WPB-4650

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4650" title="WPB-4650" target="_blank">TaskWPB-4650</a>  [iOS] Opening a secure guest link  - Knowing if this is a secure guest link
  </td></table>
  <br />
 

This pull request introduces the ability to check whether a conversation is password protected when joining through a guest link, applicable to API version 4 and above.

### Changes Made

- API Version Check: Added a conditional check for the API version (V4 and above) to incorporate password verification.
- Made changes to the response from the endpoint regarding joining group conversations to incorporate has_Password field. 

### Checklist

### DIFF
--- a/wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
+++ b/wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
@@ -126,6 +126,7 @@ extension ZMConversation {
                                completion: @escaping (Result<(conversationId: UUID, conversationName: String, hasPassword: Bool?), Error>) -> Void) {
 
         guard let apiVersion = BackendInfo.apiVersion else {
+            completion(.failure(ConversationFetchError.unknown))
             return
         }
 
--- a/wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
+++ b/wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
@@ -123,7 +123,7 @@ extension ZMConversation {
                                code: String,
                                transportSession: TransportSessionType,
                                contextProvider: ContextProvider,
-                               completion: @escaping (Result<(conversationId: UUID, conversationName: String, hasPassword: Bool?), Error>) -> Void) {
+                               completion: @escaping (Result<(conversationId: UUID, conversationName: String, hasPassword: Bool), Error>) -> Void) {
 
         guard let apiVersion = BackendInfo.apiVersion else {
             completion(.failure(ConversationFetchError.unknown))
@@ -146,9 +146,11 @@ extension ZMConversation {
                     return
                 }
 
-                var passwordProtected: Bool?
+                let passwordProtected: Bool
                 if apiVersion.rawValue >= 4 {
-                    passwordProtected = payload["has_password"] as? Bool
+                    passwordProtected = payload["has_password"] as? Bool ?? false
+                } else {
+                    passwordProtected = false
                 }
 
                 let fetchResult = (conversationId, conversationName, passwordProtected)
--- a/wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
+++ b/wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
@@ -141,9 +141,9 @@ extension ZMConversation {
                     return
                 }
 
-                let passwordProtected = payload["has_password"] as? Bool ?? false
+                let hasPassword = payload["has_password"] as? Bool ?? false
 
-                let fetchResult = (conversationId, conversationName, passwordProtected)
+                let fetchResult = (conversationId, conversationName, hasPassword)
                 completion(.success(fetchResult))
             default:
                 let error = ConversationFetchError(response: response)
--- a/wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
+++ b/wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
@@ -187,12 +187,10 @@ struct ConversationJoinRequestFactory {
         var url = URLComponents()
         url.path = joinConversationsPath
 
-        let queryItems = [
+        url.queryItems = [
             URLQueryItem(name: URLQueryItem.Key.conversationKey, value: key),
             URLQueryItem(name: URLQueryItem.Key.conversationCode, value: code)
         ]
-
-        url.queryItems = queryItems
 
         guard let urlString = url.string else {
             return nil
--- a/wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
+++ b/wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
@@ -182,7 +182,7 @@ struct ConversationJoinRequestFactory {
         var url = URLComponents()
         url.path = joinConversationsPath
 
-        var queryItems = [
+        let queryItems = [
             URLQueryItem(name: URLQueryItem.Key.conversationKey, value: key),
             URLQueryItem(name: URLQueryItem.Key.conversationCode, value: code)
         ]
--- a/wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
+++ b/wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
@@ -147,7 +147,7 @@ extension ZMConversation {
                 }
 
                 var passwordProtected: Bool?
-                if apiVersion.rawValue >= 4 && apiVersion.rawValue <= 6 {
+                if apiVersion.rawValue >= 4 {
                     passwordProtected = payload["has_password"] as? Bool
                 }
 
--- a/wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
+++ b/wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
@@ -123,7 +123,7 @@ extension ZMConversation {
                                code: String,
                                transportSession: TransportSessionType,
                                contextProvider: ContextProvider,
-                               completion: @escaping (Result<(conversationId: UUID, conversationName: String, hasPassword: Bool), Error>) -> Void) {
+                               completion: @escaping (Result<(conversationId: UUID, conversationName: String, hasPassword: Bool?), Error>) -> Void) {
 
         guard let apiVersion = BackendInfo.apiVersion else {
             return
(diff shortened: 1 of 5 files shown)


wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
```diff
@@ -123,7 +123,11 @@ extension ZMConversation {
                                code: String,
                                transportSession: TransportSessionType,
                                contextProvider: ContextProvider,
-                               completion: @escaping (Result<(conversationId: UUID, conversationName: String), Error>) -> Void) {
+                               completion: @escaping (Result<(conversationId: UUID, conversationName: String, hasPassword: Bool?), Error>) -> Void) {
+
+        guard let apiVersion = BackendInfo.apiVersion else {
+            return
```
caldrian: Please ensure `completion` is always called.

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
```diff
@@ -123,7 +123,11 @@ extension ZMConversation {
                                code: String,
                                transportSession: TransportSessionType,
                                contextProvider: ContextProvider,
-                               completion: @escaping (Result<(conversationId: UUID, conversationName: String), Error>) -> Void) {
+                               completion: @escaping (Result<(conversationId: UUID, conversationName: String, hasPassword: Bool?), Error>) -> Void) {
```
caldrian: Why does the `hasPassword` argument have to be optional?
Isn't `true` and `false` sufficient or do we need a `"N/A"` state?

caldrian (approved):

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
agisilaos: Because hasPassword isn't available to all api versions therefore we make it optional.

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
agisilaos: Done!

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
```diff
@@ -140,11 +145,17 @@ extension ZMConversation {
                     completion(.failure(ConversationFetchError.unknown))
                     return
                 }
-                let fetchResult = (conversationId, conversationName)
+
+                var passwordProtected: Bool?
+                if apiVersion.rawValue >= 4 && apiVersion.rawValue <= 6 {
+                    passwordProtected = payload["has_password"] as? Bool
+                }
```
markusfassbender: Would this break for upcoming api versions 7 and later?
A second point is that `APIVersion` is `Comparable`, which can be used without `rawValue`:
`apiVersion >= .v4`

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
```diff
@@ -175,8 +186,14 @@ struct ConversationJoinRequestFactory {
...
         url.path = joinConversationsPath
-        url.queryItems = [URLQueryItem(name: URLQueryItem.Key.conversationKey, value: key),
-                          URLQueryItem(name: URLQueryItem.Key.conversationCode, value: code)]
+
+        let queryItems = [
+            URLQueryItem(name: URLQueryItem.Key.conversationKey, value: key),
+            URLQueryItem(name: URLQueryItem.Key.conversationCode, value: code)
+        ]
+
+        url.queryItems = queryItems
```
markusfassbender: Can it be formatted without extra `let`?
```suggestion
        url.queryItems = [
            URLQueryItem(name: URLQueryItem.Key.conversationKey, value: key),
            URLQueryItem(name: URLQueryItem.Key.conversationCode, value: code)
        ]
```

wire-ios-sync-engine/Tests/Source/UserSession/DeepLinkURLActionProcessorTests.swift
```diff
@@ -49,92 +49,92 @@ final class DeepLinkURLActionProcessorTests: DatabaseTest {
...
+        // GIVEN
         let action: URLAction = .joinConversation(key: "test-key", code: "wrong-code")
 
-        // when
+        // WHEN
         sut.process(urlAction: action, delegate: presentationDelegate)
         XCTAssertTrue(waitForAllGroupsToBeEmpty(withTimeout: 0.5))
 
-        // then
+        // THEN
```
markusfassbender, line 137: Do we have aligned something about it? I always write ne tests in lower case 😬

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
markusfassbender: Does it make any problems if we would send the value to older backend versions, that do not read the value?

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
agisilaos: I guess what I can do is only have `apiVersion >= .v4` and remove the `apiVersion.rawValue <= 6`

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
agisilaos: It will be wrong to do so, no? the previous api versions aren't aware of the has_Password field

wire-ios-sync-engine/Tests/Source/UserSession/DeepLinkURLActionProcessorTests.swift
agisilaos: I don't think we aligned. I usually do all caps

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
agisilaos: Done! https://github.com/wireapp/wire-ios/commit/f5b99ad83f67dd76eb9f09cc9456ad33f9d7bcc8

wire-ios-sync-engine/Source/SessionManager/URLActions.swift
```diff
@@ -199,5 +199,6 @@ extension URLQueryItem.Key {
 
     static let conversationKey = "key"
     static let conversationCode = "code"
+    static let hasPassword = "has_password"
```
KaterinaWire: I think we don't need this anymore

wire-ios-sync-engine/Source/UserSession/URLActionProcessors/DeepLinkURLActionProcessor.swift
```diff
@@ -51,7 +51,7 @@ class DeepLinkURLActionProcessor: URLActionProcessor {
                 let viewContext = strongSelf.contextProvider.viewContext
 
                 switch response {
-                case .success((let conversationId, let conversationName)):
+                case .success((let conversationId, let conversationName, _)):
```
KaterinaWire, line 54: Maybe you can add ToDo with a Jira link here? I believe there should be `let hasPassword` that will be sent to the next step

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
```diff
@@ -140,11 +145,17 @@ extension ZMConversation {
                     completion(.failure(ConversationFetchError.unknown))
                     return
                 }
-                let fetchResult = (conversationId, conversationName)
+
+                var passwordProtected: Bool?
+                if apiVersion.rawValue >= 4 {
```
KaterinaWire: If `passwordProtected` will have a default value as false, do we really need to check the apiVersion?

wire-ios-sync-engine/Source/UserSession/URLActionProcessors/DeepLinkURLActionProcessor.swift
agisilaos: I will implement that on the next PR.

wire-ios-sync-engine/Source/SessionManager/URLActions.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/e0dbef64b93c7934e891c85eb0f6e59519ac4905

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
agisilaos: Done! https://github.com/wireapp/wire-ios/commit/a61f00b86a9735f59ad070604397b413a034d0c2

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
agisilaos: Fixed it after all!

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
```diff
@@ -123,7 +123,12 @@ extension ZMConversation {
                                code: String,
                                transportSession: TransportSessionType,
                                contextProvider: ContextProvider,
-                               completion: @escaping (Result<(conversationId: UUID, conversationName: String), Error>) -> Void) {
+                               completion: @escaping (Result<(conversationId: UUID, conversationName: String, hasPassword: Bool), Error>) -> Void) {
+
+        guard let apiVersion = BackendInfo.apiVersion else {
```
KaterinaWire: we don't need apiVersion, right?

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
```diff
@@ -140,11 +145,14 @@ extension ZMConversation {
                     completion(.failure(ConversationFetchError.unknown))
                     return
                 }
-                let fetchResult = (conversationId, conversationName)
+
+                let passwordProtected = payload["has_password"] as? Bool ?? false
```
KaterinaWire: Why don't you use the same name as `hasPassword` ?

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
agisilaos: Yep removed it.

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
agisilaos: Fixed it thanks!

KaterinaWire (approved):

markusfassbender (approved):

PR #1401, opened 2024-05-07
fix: unhandled throwing tasks

caldrian:
### Issue

It's easy to miss throwing code within Task closures.

```
Task {
    try someThrowingFunction()
}
```

This PR enables a SwiftLint rule in order to prevent this.

---

### Checklist


agisilaos (approved):
Thank you! 🙌

David-Henner (approved):
nice 👍 
could you share your reasoning behind using `assertionFailure` when catching errors?

caldrian:
> could you share your reasoning behind using assertionFailure when catching errors?

I usually choose `assertionFailure` so that the program holds when debugging with Xcode, but doesn't crash in production.
Now that I'm thinking again about it, we should introduce something in-between, a mix between `fatalError` and `assertionFailure` which crashes hard in our test-builds.

PR #1404, opened 2024-05-08
refactor: remove optional comparison override

markusfassbender:
### Issue

Removes override of comparison operators `<` and `>` for optional values. How to handle optionality should be up to the specific use case. With this change the compiler works again as expected.

To give some convenience like `array.sort(by:)` there is a new generic func `OptionalComparison.prependingNilAscending(lhs:rhs:). One needs to use it explicitly instead of a hidden operator override.

Review `OptionalComparisonTests.swift` first for exampels of old and new behaviour.

### Testing

- Run unit tests suite, passed all suites [`Test Development`](https://github.com/wireapp/wire-ios/actions/runs/9002732692) ✅
- I think it's not possible to test all cases in runtime explicitly, but we have regression tests.

---

### Checklist


wire-ios/Wire-iOS/Sources/UserInterface/Calling/Helper/Array+CallParticipant.swift
```diff
@@ -22,13 +22,19 @@ import WireSyncEngine
...
     var hasMoreThanTwoConnectedParticipants: Bool {
-        return filter(\.state.isConnected).count > 2
+        filter(\.state.isConnected).count > 2
     }
 
     mutating func sortByName(selfStreamId: AVSClient?) {
-        self = self.sorted {
-            $0.streamId == selfStreamId ||
-                $0.user.name?.lowercased() < $1.user.name?.lowercased()
+        sort {
```
markusfassbender, line 29: I think `sort` us equal in the result and using `self = self.sorted` just additionally copies arrays around.

wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/AccountSelectorController.swift
```diff
@@ -46,8 +53,14 @@ final class AccountSelectorController: UIViewController {
...
-        let showAccounts = SessionManager.shared?.accountManager.accounts.count > 1
-        guard showAccounts != showAccounts else { return }
+        let showAccounts: Bool
+        if let count = SessionManager.shared?.accountManager.accounts.count, count > 1 {
+            showAccounts = true
+        } else {
+            showAccounts = false
+        }
+
+        guard showAccounts != self.showAccounts else { return }
```
markusfassbender: This `self` is a bugfix as discussed in the `iOS developers` chat in Wire.

markusfassbender:
> ### Issue
> Removes override of comparison operators `<` and `>` for optional values. How to handle optionality should be up to the specific use case.
> 
> To give some convenience like `array.sort(by:)` there is a new generic func `OptionalComparison.prependingNilAscending(lhs:rhs:). One needs to use it explicitly instead of a hidden operator override.
> 
> ### Testing
> * Build
> * Run unit tests suite
> * I think it's not possible to test all cases in runtime explicitly
> 
> ### Checklist
> * [ ]  Title contains a reference JIRA issue number like `[WPB-XXX]`.
> * [ ]  Description is filled and free of optional paragraphs.
> * [ ]  Adds/updates automated tests.

wire-ios-utilities/Source/Public/OptionalComparison.swift
```diff
@@ -18,17 +18,31 @@
...
+    ///
+    /// - Parameters:
+    ///   - lhs: An optional comparable value on the left hand side.
+    ///   - rhs: An optional comparable value on  the right hand side.
+    ///
+    /// - Returns:
+    ///  `true` if left hand side is smaller than right hand side.
+    /// `nil` is  considered as smaller than any value.
+    /// If both paratemeters are `nil`, the method returns `false`.
+    public static func prependingNilAscending<T: Comparable>(lhs: T?, rhs: T?) -> Bool {
```
markusfassbender, line 32: The naming can be discussed.. it was not easy for me to express what it does in a comprehensive way. Thanks also to @caldrian for the proposal.

wire-ios-utilities/Source/Public/OptionalComparison.swift
```diff
@@ -18,17 +18,31 @@
...
-    case (nil, _?): return true
-    default: return false
-    }
-}
+public enum OptionalComparison {
+    /// Compares two optional values in ascending order.
+    ///
+    /// - Parameters:
+    ///   - lhs: An optional comparable value on the left hand side.
+    ///   - rhs: An optional comparable value on  the right hand side.
```
caldrian:
```suggestion
    ///   - rhs: An optional comparable value on the right hand side.
```

wire-ios-utilities/Source/Public/OptionalComparison.swift
caldrian: I actually thought about an extension to `Collection`, something like this:

```
extension Collection {

    func sortedAscendingPrependingNil<Wrapped>() -> [Element] where Element == Optional<Wrapped>, Wrapped: Comparable {
        sorted { lhs, rhs in

            if lhs == nil, rhs == nil {
                return false
            }

            guard let lhs else {
                return true
            }

            guard let rhs else {
                return false
            }

            return lhs < rhs
        }
    }
}
```

caldrian: Alternatively, to keep the syntax similar and let the user provide a closure:
```
extension Collection {

    func sortedPrependingNil<Wrapped>(
        by areInIncreasingOrder: (Wrapped, Wrapped) throws -> Bool
    ) rethrows -> [Element]
    where Element == Wrapped?, Wrapped: Comparable {
        try sorted { lhs, rhs in

            if lhs == nil, rhs == nil {
                return false
            }

            guard let lhs else {
                return true
            }

            guard let rhs else {
                return false
            }

            return try areInIncreasingOrder(lhs, rhs)
        }
    }
}
```

typfel (approved):

wire-ios-utilities/Source/Public/OptionalComparison.swift
```diff
@@ -18,17 +18,31 @@
...
+    ///
+    /// - Returns:
+    ///  `true` if left hand side is smaller than right hand side.
+    /// `nil` is  considered as smaller than any value.
+    /// If both paratemeters are `nil`, the method returns `false`.
```
netbe:
```suggestion
    /// If both parameters are `nil`, the method returns `false`.
```

wire-ios-utilities/Source/Public/OptionalComparison.swift
netbe: I am not sure to grasp what we gain compared to the custom operator? clarity?

markusfassbender: @netbe the reason why this change is important is, that one can not find a general solution to the question "how to compare nil with a value".

But with the custom operator it implicitly "works" without any compiler warning, which can lead to interesting cases like here:

1. Does `epoch=nil` enter the if with `true` or `false`

Screenshot 2024-05-14 at 16 41 29

2. What if on of the dates is nil? what if both values are `nil`?
Screenshot 2024-05-14 at 16 46 39

caldrian: Nice, `sortedAscendingPrependingNil` is very clear about the handling of `nil`.

caldrian (approved):

PR #1410, opened 2024-05-08
fix: missing conversation WPB-9054

typfel:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9054" title="WPB-9054" target="_blank">BugWPB-9054</a>  [iOS] Conversation disappears after creation with cipher suite p256
  </td></table>
  <br />
 

### Issue

Team 1:1 MLS conversations which are created by the self user doesn't appear in the conversation list until the app is put into background/foreground.

#### Why

They don't appear in the conversation because we don't update the `mlsStatus` to `.ready`, which effectively hides the conversation.

It appears after putting the app into background/foreground because this downloads the notification stream which will include the newly created conversation event which when processed updates the `mlsStatus` to `.ready`.

### Testing

1. Sign in with two fresh team accounts
2. In the contacts start a new 1:1 with the other team user

1:1 is not visible the conversation list.

---

### Checklist


johnxnguyen (approved):

markusfassbender (approved):
Thanks for also updating the unit tests.

PR #1412, opened 2024-05-09
refactor: TeamsAPI WPB-8694

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8694" title="WPB-8694" target="_blank">TaskWPB-8694</a>  [iOS] Create TeamsAPI
  </td></table>
  <br />
 

### Issue

This PR implements the `TeamAPI` to:
- fetch team metadata
- fetch team roles
- fetch team members
- fetch legalhold status for a member

### Checklist


WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
```diff
@@ -0,0 +1,367 @@
...
+
+    func getTeam(for teamID: Team.ID) async throws -> Team {
+        let request = HTTPRequest(
+            path: basePath(for: teamID),
+            method: .get
+        )
+
+        let response = try await httpClient.executeRequest(request)
+
+        switch response.code {
```
typfel: The response handling / error handling is repeated for most of these endpoints. Have you considered extracting it into a generic function?

typfel (changes requested):

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
johnxnguyen: Yeah I'm looking into ways to reduce the duplication and boilerplate code.

johnxnguyen: I factored out the boilerplate code into a response parser.

typfel (approved):

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    func basePath(for teamID: Team.ID) -> String {
+        switch apiVersion {
+        case .v0:
+            "/teams/\(teamID.transportString())"
+        default:
+            "/v\(apiVersion.rawValue)/teams/\(teamID.transportString())"
+        }
+    }
```
markusfassbender: Why is this switch in this place? From our plannings before I understood that `TeamsAPIV1` should override this method to see the differences instead of doing it inline?

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
```diff
@@ -0,0 +1,289 @@
...
+    // MARK: - Get team members
+
+    func getTeamMembers(
+        for teamID: Team.ID,
+        maxResults: UInt
+    ) async throws -> [TeamMember] {
+        var components = URLComponents(string: "\(basePath(for: teamID))/members")
+        components?.queryItems = [URLQueryItem(name: "maxResults", value: "2000")]
+
+        guard let path = components?.url?.absoluteString else {
```
markusfassbender, line 77: Good idea to use `URLComponents` here!

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
```diff
@@ -0,0 +1,289 @@
...
+
+}
+
+struct TeamResponseV0: Decodable, ToAPIModelConvertible {
+
+    let id: UUID
+    let name: String
+    let creator: UUID
+    let icon: String
+    let icon_key: String?
```
markusfassbender: Should we use snake case for vars here? We have the optionns set the ison decoder [`keydecodingstrategy`](https://developer.apple.com/documentation/foundation/jsondecoder/keydecodingstrategy/convertfromsnakecase) or custom [`CodingKey`](https://developer.apple.com/documentation/swift/codingkey).

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
```diff
@@ -0,0 +1,289 @@
...
+                $0.toAPIModel()
+            })
+        )
+    }
+
+}
+
+enum ConversationActionResponseV0: String, Decodable {
+
+    case add_conversation_member
```
markusfassbender: Same question: should we use snake case for vars here?

WireAPI/Sources/WireAPI/Models/Team.swift
markusfassbender, line 1: Should we move the Models in a subfolder `Teams`? I imagine this otherwise to become difficult to overview.

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
johnxnguyen: yeah so currently `ZMTransportRequest` is adding this prefix, but it didn't feel right that the request should mutate the path, so I thought it'd be better to pass in the correct path when creating the request.

Of course though this means that every request for a different api version is different which would require every single method of the `TeamsAPI` to be overridden, even if no other changes to the API were present.

To try strike a balance between being explicit and avoiding duplication, I thought to have this inherited base path function.

While implementing this API I found that even a small change to an endpoint means to override the whole method and have some duplication. Sometimes I was wishing that I could only override part of the implementation, such as the request only, keeping the existing response handling, or vice versa. I'm trying to imagine some kind of hybrid solution where we have versioned API implementations but those can be composed from smaller units like Endpoint request generators and response parsers.

On the other hand, I also think the duplication that comes from more explicit code can also be tolerable given that duplication is minimal and not noisy with boilerplate code. That what I've tried to achieve with the `ResponseParser`, but I also was researching to see if other swift language features like result builder or macros could help here.

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
johnxnguyen: I'm not sure though if the backend consistently uses underscores, but we could check. I wanted to keep the declarations as simple as possible which is why I went with underscore so I didn't need to define CodingKeys and instead just rely on the generated conformance. I feel like this is acceptable given that these models are not public.

WireAPI/Tests/WireAPITests/Helpers/RequestSnapshotHelper.swift
```diff
@@ -0,0 +1,97 @@
...
+            )
+        }
+    }
+
+    /// Snapshot test the request generated by the given block for a
+    /// specific api version.
+    ///
+    /// Note: this is run on the main thread as a workaround of a crash in
+    /// the SnapshotTesting framework when snapshotting in async environments.
+    /// See https://github.com/pointfreeco/swift-snapshot-testing/issues/822
```
markusfassbender, line 60: The issue is open since the beginning of January.. 😬 will it be fixed? with the `@MainActor` attribute we avoid possibility of concurrency here.

While I see the usefulness here in this case I'm always a bit careful to embed third-party libraries.. what do others think?

WireAPI/Tests/WireAPITests/TeamsAPITests.swift
```diff
@@ -0,0 +1,446 @@
...
+                            "self": 456
+                        },
+                        "created_by": "\(creatorID.transportString())",
+                        "created_at": "\(ISO8601DateFormatter.default.string(from: creationDate))",
+                        "legalhold_status": "pending"
+                    }
+                ]
+            }
+            """
+        )
```
markusfassbender: To keep readability in the tests one can move larger json to the bottom of the file and use a static var:
```swift
private enum JSON {
     static let testGetMembersV0 = """
     """
}
```

Or even create a new json file :)

WireAPI/Sources/WireAPI/Models/Team.swift
johnxnguyen: Do you mean like all team related models in /teams, all user related models in /users, etc?

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
markusfassbender: Can't we configure it for each endpoint/decoding?

WireAPI/Sources/WireAPI/Models/Team.swift
markusfassbender: Yes, in the same naming as the API subfolders. Does it make sense?

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
johnxnguyen: Do you mean something like for every endpoint?

```swift
let response = try await httpClient.executeRequest(request)
var decoder = JSONDecoder.default
decoder.keyDecodingStrategy = ...

return try ResponseParser(decoder: decoder)
    .success(code: 200, type: LegalholdStatusResponseV0.self)
    .failure(code: 404, error: TeamsAPIError.invalidRequest)
    .failure(code: 404, label: "no-team-member", error: TeamsAPIError.teamMemberNotFound)
    .parse(response)
```

WireAPI/Sources/WireAPI/Models/Team.swift
johnxnguyen: Some models appear across different APIs so it may be tricky to know which folder to put them in, but sure I will try a better organization.

WireAPI/Tests/WireAPITests/Helpers/RequestSnapshotHelper.swift
johnxnguyen: It's always a trade off and I think it comes down to a decision if it's worth it. We already use this snapshot library and it does feel helpful. Could we live without it? Probably. We could also try to fix the issue for them.

WireAPI/Tests/WireAPITests/TeamsAPITests.swift
johnxnguyen: Yeah I was thinking about having these canned payloads in separate files that we load.

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
netbe: I would use CodingKeys to keep the convention or the strategy

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
netbe: all the API classes have the same init, what about having a base class for all APIs that provides this functionality. Or going the protocol way with a common protocol

The disadvantage here I see is that when we'll drop v0 or v1 we need to move common methods around versions for example

WireAPI/Tests/WireAPITests/Helpers/RequestSnapshotHelper.swift
markusfassbender: True, we already have it in the project for view snapshots. I just wanted to raise the attention here, but for me it's acceptable too, thanks.

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
markusfassbender: I am not sure if I understand the issue, should we have a chat today after standup?

Do you mean because of an implementation like this, it uses `basePath(for: teamID)`? From my understanding it should be resolved correctly in subclasses if `basePath` has been overridden, because we have dynamic dispatch for each method.
```swift
func getTeam(for teamID: Team.ID) async throws -> Team {
        let request = HTTPRequest(
            path: basePath(for: teamID),
            method: .get
        )
```

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
johnxnguyen: done 77f2f3fad0fd7d002b837cc5caddd332ebc3f9a3

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
johnxnguyen: done 77f2f3fad0fd7d002b837cc5caddd332ebc3f9a3

WireAPI/Sources/WireAPI/Models/Team.swift
johnxnguyen: done 204b236b79532731127e0322731cb49bfb621ce8

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
johnxnguyen: From my point of view there isn't really an issue, I was just saying that it's tricky knowing where the balance should be between being explicit and having duplication vs being implicit and having less duplication. I'm happy to override the base path method for each subclass.

johnxnguyen: done 4e4840800f873dc3ad743af161c93d82b25a2727

WireAPI/Tests/WireAPITests/TeamsAPITests.swift
johnxnguyen: done f67733603197f30a86f22b090ad13eef0ac96925

WireAPI/Sources/WireAPI/Models/Team.swift
markusfassbender: Thanks!

markusfassbender (approved):

PR #1416, opened 2024-05-10
refactor: ColorPickerController to SwiftUI [suggestion markus]

markusfassbender:
### Issue

Refactor suggestion on top of #1218.

### Testing

- Run the app as logged in user
- Tap on your icon in the top left bar
- Go to Settings
- Go to Account
- Tap on "Profile Color"
- Test the color picker

---

### Checklist

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPickerController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPickerController.swift
@@ -17,41 +17,30 @@
 //
 
 import SwiftUI
-import UIKit
-import WireCommonComponents
-import WireDataModel
-import WireSyncEngine
+import protocol WireDataModel.EditableUserType
+import enum WireUtilities.AccentColor
+import protocol WireSyncEngine.UserSession
 
-class AccentColorPickerHostingController: UIHostingController<ColorPickerView> {
+final class AccentColorPickerController: UIHostingController<ColorPickerView> {
 
     @available(*, unavailable)
     required init?(coder aDecoder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
 
-    init(selectedAccentColor: Binding<AccentColor?>) {
-        let allAccentColors = AccentColor.allCases
-        var initialSelectedColor: AccentColor?
+    init(selfUser: EditableUserType, userSession: UserSession) {
+        let accentColor = AccentColor(rawValue: selfUser.accentColorValue) ?? .default
 
-        if let bindingValue = selectedAccentColor.wrappedValue {
-            // If the binding has an initial value, use it
-            initialSelectedColor = bindingValue
-        } else if let firstColor = allAccentColors.first {
-            // If no initial value, use the first available color as a default
-            initialSelectedColor = firstColor
-        }
-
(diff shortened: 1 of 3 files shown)


wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPickerController.swift
```diff
@@ -17,37 +17,30 @@
 //
 
 import SwiftUI
-import UIKit
-import WireCommonComponents
-import WireDataModel
-import WireSyncEngine
+import protocol WireDataModel.EditableUserType
```
caldrian: Is there any specific reason why you changed the imports to import single types only?

caldrian (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPickerController.swift
```diff
@@ -17,37 +17,30 @@
...
-        super.init(rootView: ColorPickerView(selectedColor: initialSelectedColor, colors: allAccentColors, onColorSelect: { selectedColor in
-            selectedAccentColor.wrappedValue = selectedColor
-            if let colorIndex = allAccentColors.firstIndex(of: selectedColor) {
-                ZMUserSession.shared()?.perform {
-                    ZMUser.selfUser()?.accentColorValue = allAccentColors[colorIndex].rawValue
+        let colorPickerView = ColorPickerView(
+            selectedColor: accentColor,
+            colors: AccentColor.allCases,
+            onColorSelect: { accentColor in
+                userSession.perform {
```
caldrian, line 38: What about moving this code out? Then `AccentColorPickerController` could probably be deleted (replaced by `UIHostingController<ColorPickerView>` instead of subclassed from).

markusfassbender: I've thought about it, where should the logic exist? What is responsible to init the SwiftUI with the right closure and handle the result on the user session?
I am not sure if the creating factory is the right place.

Maybe it would be better to create a `ViewModel`?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPickerController.swift
markusfassbender: No.. I just thought that each uses a single type so if this type disappears at some point in future the import can be removed as well.
But I can revert it, what do you prefer?

agisilaos (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPickerController.swift
markusfassbender: Updated 1d828917a769e3fa0f3a10cdc5a7815e90c3295b

wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPickerController.swift
markusfassbender: After talking we decided to keep it like it is for now.

PR #1420, opened 2024-05-10
chore: remove int even extension

markusfassbender:
### Issue

Remove general `Int` extension that is used only in one place.

### Testing

- Run app and navigate to the grid view

---

### Checklist


wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallGridView/Grid/GridView.swift
```diff
@@ -168,13 +168,18 @@ private extension GridView {
 
         switch (participantAmount, splitType) {
         case (.moreThanTwo, .proportionalSplit):
-            return numberOfItemsInPage.evenlyCeiled / 2
+            let isEven = numberOfItemsInPage.isMultiple(of: 2)
+            let ceiledNumberOfItemsInPage = isEven ? numberOfItemsInPage : (numberOfItemsInPage + 1)
+            return ceiledNumberOfItemsInPage / 2
```
markusfassbender: Isn't it equal to `return (numberOfItemsInPage + 1) / 2`, because Integer division always cut fractions (round to floor)?

typfel: True, https://github.com/apple/swift/blob/477f37ffdcbb8ab12ac8705390b03bd3fe2fc0ed/stdlib/public/core/Integers.swift#L742

So it should be the same.

typfel (approved):

agisilaos (approved):

PR #1423, opened 2024-05-10
refactor: mls group verification in `ZMUserSession` [WPB-9149]

markusfassbender:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9149" title="WPB-9149" target="_blank">BugWPB-9149</a>  iOS: refactor mls group verification in ZMUserSession
  </td></table>
  <br />
 

### Motivation

The `ZMUserSession`'s initialiser takes about 25 arguments. They can be categorised in two:
1. primary dependencies: foundational services, caches etc.. Ideally they should not use other dependencies.
2. secondary dependencies: derived objects that can be initialised by the primary dependencies.

The `ObserveMLSGroupVerificationStatusUseCase` belongs to category 2 and thus should not be passed to the initialiser of `ZMUserSession` to avoid a infinite growing of responsibilities.

The second point was, that the use case has a state. Therefore it was required to retain the whole use case and keep a task reference alive, again a responsibility of `ZMUserSession` until now. This is implicit and cannot be easily understand from the outer scope.

### Refactoring

1.  While working on it, I figured out that the logic broke into several smaller files, that kind of belong together, and overlapped in some points, logs etc.:
- `ObserveMLSGroupVerificationStatusUseCase`
- `MLSConversationVerificationStatusUpdater`
- `UpdateMLSGroupVerificationStatusUseCase`

`ObserveMLSGroupVerificationStatusUseCase` and `MLSConversationVerificationStatusUpdater` got merged into `MLSGroupVerification`. Extract extension to user session. Simplify mocking and unit tests.

2. Extract `CertificateRevocationLists` as extension to user session.

### Review

> [!NOTE]
> Start your review in `ZMUserSession.swift` to see the benefit there.

Maybe it's also helpful to checkout the whole branch and read classes in code.

### Testing

- Simple test: breakpoints, run the app and login
- For the whole ticket we need to test the MLS group verification with @KaterinaWire and QA.

---

### Checklist


wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -424,6 +427,12 @@ public final class ZMUserSession: NSObject {
         applicationStatusDirectory.syncStatus.syncStateDelegate = self
         applicationStatusDirectory.clientRegistrationStatus.registrationStatusDelegate = self
 
+        let observeMLSGroupVerificationStatusUseCase = observeMLSGroupVerificationStatusUseCase ?? ObserveMLSGroupVerificationStatusUseCase(
+            mlsService: mlsService,
+            updateMLSGroupVerificationStatusUseCase: updateMLSGroupVerificationStatus,
+            syncContext: coreDataStack.syncContext
+        )
+
```
johnxnguyen: Is this only injected for tests? The tests also seem to just pass an empty implementation, suggesting that we don't want any observation during tests.

I feel like `ObserveMLSGroupVerificationStatusUseCase` shouldn't be a use case, but rather just an object that observes epoch changes and invokes another use case. I'm wondering if we even need to have this here and instead setup and tear down the observation when needed.

markusfassbender: 1. Yes it's only for tests. My assumption is that the original idea was tests using the fat `ZMUserSession` with all kinds of side effects should not perform them and thus causing suddenly unexpected behaviour.

2. I agree that it's not really a use case. The reason why it lives in `ZMUserSession` is probably the life cycle and that it must be retained as an observer. @David-Henner maybe you can give us some additional insight?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -274,7 +259,8 @@ public final class ZMUserSession: NSObject {
             onNewCRLsDistributionPointsSubject: onNewCRLsDistributionPointsSubject
         )
 
-        cRLsDistributionPointsObserver.startObservingNewCRLsDistributionPoints(
+        // TODO: why is this observer started two times? first time is 'setupCertificateRevocationLists'!
+        cRLsDistributionPointsObserver?.startObservingNewCRLsDistributionPoints(
             from: onNewCRLsDistributionPointsSubject.eraseToAnyPublisher()
         )
```
markusfassbender: @KaterinaWire @David-Henner can you help me with this question?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
markusfassbender: @johnxnguyen I worked on this and chose a different solution to the use case. Would be happy to get some feedback :) thanks

wire-ios-data-model/Source/MLS/MLSGroupVerification.swift
```diff
@@ -0,0 +1,106 @@
...
+public final class MLSGroupVerification: MLSGroupVerificationProtocol {
+
+    // MARK: - Properties
+
+    private let updateUseCase: any UpdateMLSGroupVerificationStatusUseCaseProtocol
+
+    private let mlsService: MLSServiceInterface
+    private let syncContext: NSManagedObjectContext
+
+    private var observingTask: Task<Void, Never>?
```
caldrian: optional
```suggestion
    private var observationTask: Task<Void, Never>?
```

wire-ios-data-model/Source/MLS/MLSGroupVerification.swift
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    deinit {
+        observingTask?.cancel()
+    }
+
+    // MARK: Observing
+
+    public func startObserving() {
+        observingTask = .detached { [mlsService, weak self] in
```
caldrian: Why is `mlsService` in the `capture list`? Can't it be accessed via `self`?

caldrian (approved):
Nice cleanup 🧹

wire-ios-data-model/Source/MLS/MLSGroupVerification.swift
markusfassbender: I thought it was not possible, but found a way now. Thanks 👍

wire-ios-data-model/Source/MLS/MLSGroupVerification.swift
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        await updateConversation(conversation, with: groupID)
+    }
+
+    public func updateConversation(_ conversation: ZMConversation, with groupID: MLSGroupID) async {
+        do {
+            try await updateUseCase.invoke(for: conversation, groupID: groupID)
+        } catch {
+            WireLogger.e2ei.warn("failed to update MLS group: \(groupID.safeForLoggingDescription) verification status: \(String(describing: error))")
```
KaterinaWire, line 87: I would keep "failed to update conversation's verification status", because it's more accurate. We only update the verification status. 
I would also rename `updateUseCase` to something like `updateVerificationStatus`

wire-ios-data-model/Source/MLS/MLSGroupVerification.swift
```diff
@@ -0,0 +1,108 @@
...
+
+    deinit {
+        observationTask?.cancel()
+    }
+
+    // MARK: Observing
+
+    public func startObserving() {
+        observationTask = .detached { [weak self] in
+            guard let asyncStream = self?.mlsService.epochChanges() else { return}
```
KaterinaWire:
```suggestion
            guard let asyncStream = self?.mlsService.epochChanges() else { return }
```

wire-ios-data-model/Source/MLS/MLSGroupVerification.swift
```diff
@@ -0,0 +1,108 @@
...
+        }) else {
+            return WireLogger.e2ei.warn("failed to fetch the conversation by mlsGroupID \(groupID)")
+        }
+
+        await updateConversation(conversation, with: groupID)
+    }
+
+    public func updateConversation(_ conversation: ZMConversation, with groupID: MLSGroupID) async {
+        do {
+            try await updateUseCase.invoke(for: conversation, groupID: groupID)
```
KaterinaWire: I would rename `updateUseCase` to something like `updateVerificationStatus`. What do you think?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
David-Henner: we're adding a different publisher. in `setupCertificateRevocationLists` the publisher is the publisher from `mlsService`. Here it's another publisher, `onNewCRLsDistributionPointsSubject` that is passed to `E2EIKeyPackageRotator` and `E2EIRepository`.

Technically we could have only one publisher and pass it down to all the dependencies

markusfassbender: Okay sounds good so far, I think we need someway to make sure that `setupCertificateRevocationLists()` and this are executed in the right order. Maybe we can have a short pairing session?

markusfassbender: @David-Henner since I am not really able to resolve the timing issue now, I've added just an assertion in 0624c4fd98dab07bad0e162c4a6da5efe9c30beb. This whole construction with lazy vars, that are dependent one each other is quite fragile.
Probably we can make a follow up PR here to clean this up, what do you think? cc @KaterinaWire

johnxnguyen (approved):
Nice

KaterinaWire (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
KaterinaWire: Agree

PR #1444, opened 2024-05-15
fix: Send invisible messages in a degraded conversation WPB-7402

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7402" title="WPB-7402" target="_blank">BugWPB-7402</a>  [iOS] Message cannot be sent for confirmation and last read messages once conversation is degraded
  </td></table>
  <br />
 

### Issue
Invisible messages (hidden in the conversation) were not sent in a degraded conversation.
In the `MessageDependencyResolver` we check if the conversation has degraded and throw an error if so. This is ok for all visible messages because from a UI perspective we are informing the user and asking for confirmation. There is no UI interaction for hidden messages.

### Solution
I've added `shouldIgnoreTheSecurityLevelCheck` to the message and ignore these messages for security check in the `MessageDependencyResolver`.

### Testing

1. Degrade mls conversation.
2. Receive a message in a group where Read receipts are enabled.

### Checklist


typfel (approved):
Nice

caldrian (approved):

wire-ios-request-strategy/Sources/Protocols/OTREntity.swift
```diff
@@ -37,6 +37,9 @@ private let zmLog = ZMSLog(tag: "Dependencies")
     /// This message entity has expired and no more attempt will be made to sent it
     var isExpired: Bool { get }
 
+    /// This message entity should be ignored for security level check
```
netbe, line 40: 👍

netbe (approved):

PR #1445, opened 2024-05-15
refactor: create TeamRepository WPB-8739

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8739" title="WPB-8739" target="_blank">TaskWPB-8739</a>  [iOS] Create TeamRepository
  </td></table>
  <br />
 

### Issue

This PR implements the `TeamRepository` with an initial set of methods to pull team data from the server via the `TeamsAPI` and store them locally in the database via Core Data.

I've also added sourcery to `WireAPI` and a new `WireAPISupport` module to contain the mocks.

As a first step I've kept the existing code to store the core data entities, but we may wish to introduce an abstraction around the data layer in the future.

---

### Checklist


WireAPI/Package.swift
```diff
@@ -21,12 +25,19 @@ let package = Package(
         .package(
             url: "https://github.com/pointfreeco/swift-snapshot-testing",
             from: "1.16.0"
+        ),
+        .package(
+            url: "https://github.com/krzysztofzablocki/Sourcery.git",
+            from: "2.2.4"
```
markusfassbender: This pulls a lot of sub-dependencies in the project that all need to be build. Until now we have used another way with a binary file, @caldrian has some knowledge there.
Can we use it here too?

wire-ios-data-model/Tests/Source/Model/Conversation/ZMConversationTests+SecurityLevel.swift
```diff
@@ -519,10 +519,10 @@ class ZMConversationTests_SecurityLevel: ZMConversationTestsBase {
             selfUser.name = "MYSELF"
             conversation.addParticipantAndUpdateConversationState(user: selfUser, role: nil)
 
-            let mainTeam = Team.fetchOrCreate(with: UUID.create(),
-                                              create: true,
-                                              in: self.syncMOC,
-                                              created: nil)!
+            let mainTeam = Team.fetchOrCreate(
+                with: UUID.create(),
```
markusfassbender: Is it possible here to use the default init instead of the customised func?
```suggestion
                with: UUID,
```

wire-ios-sync-engine/Source/Repositories/TeamRepository.swift
```diff
@@ -0,0 +1,261 @@
...
+
+    func pullSelfTeamRoles() async throws
+
+    /// Pull team members for the self team from the server and store locally.
+
+    func pullSelfTeamMembers() async throws
+
+    /// Fetch the legalhold status for the self user from the server.
+
+    func fetchSelfLegalholdStatus() async throws -> LegalholdStatus
```
markusfassbender, line 45: Do I understand correctly, that `pull` should describe also the locale store? I like the idea

wire-ios-sync-engine/Source/Repositories/TeamRepository.swift
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    private func storeTeamLocally(_ teamAPIModel: WireAPI.Team) async {
+        await context.perform { [context] in
+            let team = WireDataModel.Team.fetchOrCreate(
+                with: teamAPIModel.id,
+                in: context
+            )
+
+            let selfUser = ZMUser.selfUser(in: context)
```
markusfassbender: Using this func is kind of a smelly piece of code, because it has many implicit conditions. Can we avoid it somehow in new code?
Maybe one idea could be to write a `SelfUserRepository` and inject it here to make it easier to refactor in future?

WireAPI/Package.swift
johnxnguyen: Yeah it'd be nice to pull it once and share it, not sure how though. We could also just not add the dependency to the package and instead generate the mocks via the command line.

wire-ios-sync-engine/Source/Repositories/TeamRepository.swift
johnxnguyen: I'm not sure what you mean by "describe also the locale store".

wire-ios-sync-engine/Source/Repositories/TeamRepository.swift
johnxnguyen: Certainly!

agisilaos (approved):
Looking forward to many of such changes! Thank you @johnxnguyen. 🙌

wire-ios-sync-engine/Source/Repositories/TeamRepository.swift
johnxnguyen: Hmmm so we already have a `UserRepository` define in WireDataModel (because it's used in WireRequestStrategy). The way it's used now is problematic so I'll try to fix that in a separate PR first so we can use it here.

johnxnguyen: https://github.com/wireapp/wire-ios/pull/1451

wire-ios-sync-engine/Source/Repositories/TeamRepository.swift
markusfassbender: I tried to rephrase and see what `and store locally` from the docuementation means 😅 it's fine I looked into the code

wire-ios-sync-engine/Source/Repositories/TeamRepository.swift
johnxnguyen: done in 3f9ef3da990450c0801e62bcc32fa94459d1ef25

wire-ios-data-model/Tests/Source/Model/Conversation/ZMConversationTests+SecurityLevel.swift
johnxnguyen: yes, done c20f80c1b73c1667c7d15bf5a50a7831398ed8da

WireAPI/Package.swift
caldrian: I don't know about a quick solution now.
We have the sourcery binary downloaded by the "scripts" package here:
`./scripts/.build/artifacts/scripts/sourcery/sourcery/bin/sourcery`
Maybe there is a way to execute a shell script during the build process?
Maybe writing a custom plugin which executes the existing script?

johnxnguyen: I removed the plugin for now, we can still generate mocks from the command line by running `sourcery` from the `WireAPI` directory.

WireAPI/Package.resolved
```diff
@@ -1,6 +1,114 @@
...
+    },
+    {
+      "identity" : "sourcery",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/krzysztofzablocki/Sourcery.git",
+      "state" : {
+        "revision" : "3b84a205f25c58d03369e3775703c7c7ee115a77",
+        "version" : "2.2.4"
+      }
+    },
```
markusfassbender: This still contains all sub-dependencies, is that right?

markusfassbender (approved):

WireAPI/Package.resolved
johnxnguyen: It seems like the resolved file in the workspace is used rather than this local file, I've deleted it and everything runs fine and a new resolved file wasn't regenerated.

PR #1451, opened 2024-05-16
refactor: move UserRepository to WireSyncEngine

johnxnguyen:
### Issue

We're now starting to refactor code and introduce repositories as objects that aggregate remote and local datasources. Currently we have an existing `UserRepository` which was defined in WireDataModel and used in WireRequestStrategy, but we actually want to have the repository in the sync engine so it can depend on a future `UsersAPI` from the WireAPI module. This PR moves the `UserRepository` as well as few objects that depend on it to WireSyncEngine.

---

### Checklist


typfel:
Do we really want to move it to the WireSyncEngine? If we ever intend to use it from the share ext of the notification ext it must live somewhere else.

caldrian (approved):

johnxnguyen:
> Do we really want to move it to the WireSyncEngine? If we ever intend to use it from the share ext of the notification ext it must live somewhere else.

@typfel we need a layer between the engines and WireAPI. I envision this as WireDomain or something like that. We'll have to figure that out when we want to start using the new code we're writing now, perhaps we should meet to discuss it soon.

typfel (approved):

PR #1482, opened 2024-05-23
refactor: add conversations api [WPB-8696]

markusfassbender:
<table>
<td>
  </td></table>
  <br />
 

### Issue

Introduces `WireAPI/ConversationsAPI` to access the conversations via the new api.

It adds a single func `getConversationIdentifiers()` with unit tests, but more will follow in new pull requests.

### Testing

- Run unit tests
- Not used in production code yet

---

### Checklist


WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
```diff
@@ -0,0 +1,88 @@
...
+        case hasMore = "has_more"
+    }
+
+    let conversationUUIDs: [UUID]
+    let pagingState: String
+    let hasMore: Bool
+
+    func toAPIModel() -> PayloadPager<QualifiedID>.Page {
+        let qualifiedIDs = conversationUUIDs.map {
+            QualifiedID(uuid: $0, domain: "")
```
markusfassbender: Does this work with backend to map domain to empty string?

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIBuilder.swift
```diff
@@ -0,0 +1,56 @@
...
+    private let httpClient: any HTTPClient
+
+    /// Create a new builder for the conversations API.
+    ///
+    /// - Parameter httpClient: A http client.
+    public init(httpClient: any HTTPClient) {
+        self.httpClient = httpClient
+    }
+
+    /// Make a versioned`ConnectionsAPI`.
```
netbe:
```suggestion
    /// Make a versioned`ConversationsAPI`.
```

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIBuilder.swift
```diff
@@ -0,0 +1,56 @@
...
+    ///
+    /// - Parameter version: An api version.
+    /// - Returns: A versioned `ConnectionsAPI`.
```
netbe:
```suggestion
    /// - Returns: A versioned `ConversationsAPI`.
```

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
netbe: maybe better map to BackendInfo.domain

netbe: It's more with what we store than backend, we would need to check for empty string

johnxnguyen (approved):
Looks good, nice work

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
markusfassbender: Starting to work on this I got into trouble with `APIBuilder` and other things arounds.. here is a suggestion to discuss in a stacked PR:
https://github.com/wireapp/wire-ios/pull/1512/files

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIError.swift
```diff
@@ -0,0 +1,27 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+/// Errors originating from `ConversationsAPI`.
+
+public enum ConversationsAPIError: Error {
+
+    /// Invalid body
```
johnxnguyen: This comment should be updated.

johnxnguyen (approved):
Still looks good 👍🏻

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIError.swift
markusfassbender: Good point, will do so in the follow up PR and merge now as it passed all tests and is ready. thanks

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPI.swift
```diff
@@ -0,0 +1,29 @@
...
+import Foundation
+
+/// Access to conversations API.
+public protocol ConversationsAPI {
+
+    /// Fetch all conversation identifiers in batches for ``APIVersion`` v0.
+    func getLegacyConversationIdentifiers() async throws -> PayloadPager<UUID>
+
+    /// Fetch all conversation identifiers in batches available from ``APIVersion`` v1.
+    func getConversationIdentifiers() async throws -> PayloadPager<QualifiedID>
```
netbe, line 28: I think we need a comment here to remember why we ended up with a two methods for same need.

Normally we would have one method that abstract the api version away.

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
```diff
@@ -0,0 +1,90 @@
...
+
+    let httpClient: HTTPClient
+
+    // MARK: - Initialize
+
+    init(httpClient: HTTPClient) {
+        self.httpClient = httpClient
+    }
+
+    func getLegacyConversationIdentifiers() async throws -> PayloadPager<UUID> {
```
netbe, line 41:
```suggestion
    public func getLegacyConversationIdentifiers() async throws -> PayloadPager<UUID> {
```
isn't it missing the visibility?

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
```diff
@@ -0,0 +1,90 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+class ConversationsAPIV0: ConversationsAPI, VersionedAPI {
+
+    // MARK: - Constants
+
+    enum Constant {
```
netbe:
```suggestion
    private enum Constants {
```
nitpic: but what we stick with plural or singular on ConnectionsAPI it's plural and private

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPI.swift
markusfassbender: Okay, I can add a comment in the implementation that mapping was a problem, but I think here for the caller it's clear how to use the API object, isn't it?

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV2.swift
```diff
@@ -18,10 +18,6 @@
...
 
-public protocol APIBuilder {
-    associatedtype API
-
-    init(httpClient: any HTTPClient)
-
-    func makeAPI(for version: APIVersion) -> API
+class ConversationsAPIV2: ConversationsAPIV1 {
+    override var apiVersion: APIVersion { .v2 }
 }
```
netbe, line 23: was the only need of this to make it work on snaphot tests?

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPI.swift
markusfassbender: Normally yes, but we have different return types that can not be simply mapped to each other... and passing the backend version was an egg dance... as you can see in the last commits. So I would prefer to keep it up to the caller now how to process the results, ok?

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
markusfassbender: good point, but I think it doesn't need to be `public` because the whole `ConversationsAPI0` isn't. the outer caller just uses the `ConversationsAPIBuilder` and sees the protocol.

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
markusfassbender: Me personally prefer singular when you use it it's `Constant.batchSize`. But I can change it to plural again. It can't be private, because `ConversationsAPIV1` uses the constants too.

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV2.swift
markusfassbender: I thought the pattern is like this, that we create one class for every version? cc @typfel 
From my point of view we can remove all classes, that didn't make any changes...
e.g. if you have changes in V6 it can be `ConversationsAPIV6: ConversationsAPIV1` and the `ConversationsAPIBuilder` takes care of the right instantiation?

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIError.swift
markusfassbender: done here

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPI.swift
netbe: yes make sense, my point was more about the why, adding a comment in the implementation sounds good :)

markusfassbender: 0c74973272eb45186bfbdf666a262354cdebfecd

agisilaos (approved):

PR #1487, opened 2024-05-24
fix: duplicate link notifications - WPB-9417

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9417" title="WPB-9417" target="_blank">BugWPB-9417</a>  Notification containing a link is displayed twice
  </td></table>
  <br />
 

### Issue

When sending message with a link, the receiver on iOS will get two notifications. This seems to happen because we get two updateEvents for the same message id: one for the message and one that contains the link preview.

Solution

Filter notifications that come from the same message id.

### Testing

- [x] tried to send message with a link, it should display only 1 notifiaction with the link - not 2 same notifications nor 1 *grouped* notification  

---

### Checklist


johnxnguyen (commented):
Can we check the contents of the decoded protobuf messages of the text and the link preview? I think we can filter out at a more appropriate step.

wire-ios/Wire Notification Service Extension/LegacyNotificationService.swift
```diff
@@ -141,6 +143,19 @@ public final class LegacyNotificationService: UNNotificationServiceExtension, No
         contentHandler(content)
     }
 
+    private func removeNotification(withSameMessageId messageNonce: UUID?) {
```
netbe, line 146: this will remove any notification with the same message id already delivered, maybe during another fetch of notifications stream.

This will be also useful for another upcoming PR to replace edited messages

wire-ios-request-strategy/Sources/Notifications/Push Notifications/Notification Types/LocalNotificationContentType.swift
```diff
@@ -80,6 +80,7 @@ public enum LocalNotificationContentType: Equatable {
             self = .knock
 
         case .image:
+            // NOTE: is this v2 asset ?
```
markusfassbender, line 83: Does this need to become solved first?

wire-ios-notification-engine/Sources/NotificationSession.swift
```diff
@@ -409,18 +409,24 @@ extension NotificationSession: PushNotificationStrategyDelegate {
...
+
         for event in events {
             if let callEventPayload = callEventPayloadForCallKit(from: event) {
                 WireLogger.calling.info("detected a call event")
                 // Only store the last call event.
                 callEvent = callEventPayload
             } else if let notification = notification(from: event, in: context) {
                 WireLogger.notifications.info("generated a notification from an event")
-                localNotifications.append(notification)
+                tempNotifications[notification.contentHashValue] = notification
```
markusfassbender, line 423: I think it works this way as we need it. It feels a bit complex so I wondered if we can override the NSObject hash value and use a `Set<ZMLocalNotification>` over this dictionary? 
```swift
    public override var hash: Int {
        var hash = Hasher()
        hash.combine(messageNonce)
        hash.combine(title)
        hash.combine(body)
        return hash.finalize()
    }
```

Would it have other side effects maybe?

netbe: it could be weird to have a custom implementation in other cases where we use ZMLocalNotification.

wire-ios-request-strategy/Sources/Notifications/Push Notifications/Notification Types/LocalNotificationContentType.swift
netbe: no, it's just an annotation, at some point we might need to clean

markusfassbender (approved):

KaterinaWire (approved):

David-Henner (approved):

PR #1488, opened 2024-05-24
fix: enable speaker button for video calls - WPB-9217

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9217" title="WPB-9217" target="_blank">BugWPB-9217</a>  [iOS] Speaker not enabled when video is enabled during a call
  </td></table>
  <br />
 

### Issue

Speaker not enabled when video is enabled during a call.

### Solution

This issue occurred after replacing the old calling UI with the new one. In the old version, we replaced the speaker with a flip camera button for video calls. We no longer need to do this and the speaker button should always be on. I also tested the logic using an Android app.
I also discovered that we still have some "old" code from the calling UI migration (i.e. `CallActionsView`) and I can't just remove it in this PR. We are currently using `CallingActionsView`. I decided not to do any refactoring here, perhaps @David-Henner   will do it as part of a calling performance improvement.

### Testing

- join a video call
- turn on/off speaker button

### Checklist


agisilaos (approved):

David-Henner (approved):

netbe (approved):
Great, thks for fixing this, let's make sure that old call ui code is cleaned up

PR #1500, opened 2024-05-27
fix: deleteClient crash on missing managedObjectContext

netbe:
### Issue

Fix crash (nil found) on debug builds where a client is deleted

---

### Checklist


markusfassbender (approved):
Is it possible to reproduce those steps for testing?

netbe:
> Is it possible to reproduce those steps for testing?

@markusfassbender not sure it's second time it happens to me in different times, I'll check

wire-ios-data-model/Source/Model/UserClient/UserClient.swift
```diff
@@ -265,7 +265,9 @@ public class UserClient: ZMManagedObject, UserClientType {
     }
 
     private func deleteClient() {
-        assert(self.managedObjectContext!.zm_isSyncContext, "clients can only be deleted on syncContext")
+        guard let managedObjectContext else { return }
+
+        assert(managedObjectContext.zm_isSyncContext, "clients can only be deleted on syncContext")
```
caldrian, line 270: I had this crash too, but couldn't reproduce it.
But I wonder, how could the `managedObjectContext` become `nil` while/before executing the `managedObjectContext.perform` closure?

caldrian: > I wonder, how could the managedObjectContext become nil while/before executing the managedObjectContext.perform closure?

If that's really the case, what about holding a strong reference on the caller side?
```
if let managedObjectContext {
    await managedObjectContext.perform { self.deleteClient() }
}
```

netbe: it seems to happen when i change branch

netbe: the nil part is either we don't have the coredatastack or the userclient was deleted already therefore not attached to the MOC

johnxnguyen (approved):

PR #1512, opened 2024-05-30
refactor: create conversations api with `BackendInfo` [WPB-8696]

markusfassbender:
<table>
<td>
  </td></table>
  <br />
 

### Issue

_Please describe the issue._

### Testing

_Describe how to test._

---

### Checklist


WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
```diff
@@ -54,15 +57,29 @@ class ConversationsAPIV0: ConversationsAPI, VersionedAPI {
...
+            return payload.toAPIModel(domain: backendDomain)
+        default:
+            throw try decoder.decode(FailureResponse.self, from: data)
         }
     }
 
 }
 
-private struct PaginatedConversationIDsV0: Decodable, ToAPIModelConvertible {
+private struct PaginatedConversationIDsV0: Decodable {
```
markusfassbender: It was not possible to use `ToAPIModelConvertible`, because of the additional required value. Then I implemented the decoding manually again.
Can we find a better design?

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIBuilder.swift
```diff
@@ -37,19 +39,20 @@ public struct ConversationsAPIBuilder: APIBuilder {
     public func makeAPI(for version: APIVersion) -> any ConversationsAPI {
         switch version {
         case .v0:
-            ConversationsAPIV0(httpClient: httpClient)
+            ConversationsAPIV0(httpClient: httpClient, backendDomain: backendDomain)
         case .v1:
-            ConversationsAPIV1(httpClient: httpClient)
+            // TODO: we only need backend info for V0, but now pass to all?
```
markusfassbender: This is due to the inheritance of classes. Any ideas how to avoid it and make it clear, that it can be refactored simply in future?

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
johnxnguyen: Perhaps we can return back an empty domain here, and on the caller side where we have the local domain info, we could use it wherever we find an empty domain?

markusfassbender: thanks, I choose a different implementation where `ToAPIModelConvertible` works.

PR #1515, opened 2024-05-30
feat: Display Active Filters in Conversation List View - WPB-7309

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7309" title="WPB-7309" target="_blank">StoryWPB-7309</a>  [iOS] : Header : Add filter Button
  </td></table>
  <br />
 

With this PR we introduce a new view to show which filter is applied to our conversation list and a way to remove it with the remove button. More specifically the changes look like:

1. Filter Management:
- Added a filterContainerView to display the currently applied filter.
- Implemented logic to show and hide the filterContainerView based on the selected filter.
- Updated filter labels to use localized strings for consistency across the app.
- Added a removeButton to allow users to clear the selected filter.

2. Code organization
- Organized the class in sections so it's easier to navigate

### Screenshot

Screenshot 2024-05-30 at 15 49 04

--

**TODO**
- [ ] Add snapshot tests

### Checklist

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
@@ -247,7 +247,7 @@ final class ConversationListViewController: UIViewController {
         filterLabel = UILabel()
         filterLabel.font = UIFont.font(for: .h5)
         filterLabel.textColor = SemanticColors.Label.baseSecondaryText
-        filterLabel.text = "Filtered by \(selectedFilterLabel)"
+        filterLabel.text = L10n.Localizable.ConversationList.FilterLabel.text(selectedFilterLabel)
 
         removeButton = UIButton(type: .system)
         removeButton.setTitle("Remove", for: .normal)
@@ -352,7 +352,7 @@ final class ConversationListViewController: UIViewController {
         if filter == .allConversations {
             filterContainerView.isHidden = true
         } else {
-            filterLabel.text = "Filtered by \(selectedFilterLabel)"
+            filterLabel.text = L10n.Localizable.ConversationList.FilterLabel.text(selectedFilterLabel)
             filterContainerView.isHidden = false
         }
 
--- a/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
@@ -84,6 +84,7 @@ final class ConversationListViewController: UIViewController {
         return label
     }()
 
+    /// Arranges the filterContainerView (if visible) and the contentContainer below each other.
     private var stackView: UIStackView!
 
     let contentContainer: UIView = {
--- a/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
@@ -35,15 +35,16 @@ final class ConversationListViewController: UIViewController {
     private var removeButton: UIButton!
 
     var selectedFilterLabel: String {
+        typealias FilterMenuLocale = L10n.Localizable.ConversationList.Filter
         switch listContentController.listViewModel.selectedFilter {
         case .allConversations:
             return ""
         case .favorites:
-            return "Favorites"
+            return FilterMenuLocale.Favorites.title
         case .groups:
-            return "Groups"
+            return FilterMenuLocale.Groups.title
         case .oneToOneConversations:
-            return "1:1 Conversations"
+            return FilterMenuLocale.OneOnOneConversations.title
         }
     }
 
--- a/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
@@ -253,7 +253,12 @@ final class ConversationListViewController: UIViewController {
         removeButton.setTitle(L10n.Localizable.ConversationList.Filter.RemoveButton.title, for: .normal)
         removeButton.titleLabel?.font = UIFont.font(for: .h5)
         removeButton.setTitleColor(UIColor.accent(), for: .normal)
-        removeButton.addTarget(self, action: #selector(removeFilter), for: .touchUpInside)
+
+        let action = UIAction { [weak self] _ in
+            self?.removeFilter()
+        }
+
+        removeButton.addAction(action, for: .touchUpInside)
 
         filterContainerStackView.addArrangedSubview(filterLabel)
         filterContainerStackView.addArrangedSubview(removeButton)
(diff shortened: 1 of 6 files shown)


wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
```diff
@@ -251,35 +336,52 @@ final class ConversationListViewController: UIViewController {
...
+            filterContainerView.isHidden = false
+        }
 
-        let navigationController = UINavigationController(rootViewController: viewController)
-        navigationController.transitioningDelegate = self
-        navigationController.modalPresentationStyle = .currentContext
+        // Trigger a layout update to ensure the correct positioning
+        // of the add conversation button and filter button
+        // when the filter button is tapped.
+        view.setNeedsLayout()
```
caldrian: This line is redundant since `setupRightNavigationBarButtons` calls `setNeedsLayout`.

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
```diff
@@ -184,6 +211,66 @@ final class ConversationListViewController: UIViewController {
...
+
+        filterContainerStackView.addArrangedSubview(filterLabel)
+        filterContainerStackView.addArrangedSubview(removeButton)
+
+        // Initially hide the filter container view
+        filterContainerView.isHidden = true
+    }
+
+    func removeFilter() {
+        applyFilter(.allConversations)
```
caldrian: optional: Since `.allConversations` is not filtering anything, we might also remove it and use `Optional<ConversationFilterType>` instead.
Then we wouldn't "apply" an "actually-nothing-filtering-filter", but just "remove" the filter.

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
```diff
@@ -44,6 +63,8 @@ final class ConversationListViewController: UIViewController {
         return label
     }()
 
+    private var stackView: UIStackView!
```
caldrian, line 85: Could you please add a comment what this stack view is for?
Something like "arranges the `filterContainerView` (if visible) and the `contentContainer` below each other.

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
```diff
@@ -251,35 +336,52 @@ final class ConversationListViewController: UIViewController {
...
+        view.setNeedsLayout()
+    }
 
-        tabBarController?.present(navigationController, animated: true) {
-            viewController.showKeyboardIfNeeded()
-        }
+    // MARK: - Selection Management
+
+    /// Scroll to the current selection
+    /// - Parameter animated: Perform animation or not
```
caldrian: opinion
Thank you for commenting code.
In some cases like this one, however, I think comments don't add any value/information and so I suggest to not add them at all. (since it enlarges the file and they need to be maintained)

In this case for example:
- "Scroll to the current selection" is almost the same as the method name `scrollToCurrentSelection`. (the method name is actually self-explanatory)
- `animated` I would consider convention, so every iOS developer knows what it means.

I would consider a comment valuable for example if it contained non-obvious background information, infos about side effects, things to keep in mind etc.
Otherwise we should try to choose self-explanatory names/identifiers.

I'm happy to learn about other opinions in this regard.

johnxnguyen: I agree with you @caldrian , the comment here doesn't provide any additional information that I can't already extract (quite easily) from the well-named method.

I would add though, in the case of documentation of a public api (not the case here), I like to have doc comments to be explicit and transparent. I guess that is more of a habit and / or attempt to be consistent.

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
```diff
@@ -23,16 +23,35 @@ import WireSyncEngine
...
+
     let viewModel: ViewModel
 
-    /// private
     private var viewDidAppearCalled = false
     private static let contentControllerBottomInset: CGFloat = 16
 
+    private var filterContainerView: UIView!
+    private var filterLabel: UILabel!
+    private var removeButton: UIButton!
```
johnxnguyen: Can we avoid implicitly unwrapped types by initializing them (also lazily?) and setting up later when the view loads?

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
markusfassbender: Indeed it's hard to write high quality documentation, that doesn't age too quick.. maybe it's worth to take a workshop or online course for that? Just an idea

agisilaos: I agree it's redundant in this case. I removed it: https://github.com/wireapp/wire-ios/commit/57aef7f884f37c30c43ee13726b89a8db7fe1c8a

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/57aef7f884f37c30c43ee13726b89a8db7fe1c8a

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/6a9b3d3c3cdae9243a9e600dc5a8ce7508ba52c6

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/49b1d92e6efffbf2ec50248358e51ffdbfa8e974

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/821cfe38f1fe7b203124526f1f2a166da383cf8e

caldrian (approved):
👏

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
```diff
@@ -23,16 +23,56 @@ import WireSyncEngine
...
     let viewModel: ViewModel
 
-    /// private
     private var viewDidAppearCalled = false
     private static let contentControllerBottomInset: CGFloat = 16
 
+    private lazy var filterContainerView: UIView = {
+        let view = UIView()
+        return view
+    }()
```
caldrian:
```suggestion
    private lazy var filterContainerView = UIView()
```

johnxnguyen (approved):

PR #1518, opened 2024-05-31
refactor: api conversations by ids [WPB-8696]

markusfassbender:
<table>
<td>
  </td></table>
  <br />
 

### Issue

Adds fetch conversation list with qualified identifiers:
`func getConversations(for identifiers: [QualifiedID]) async throws -> ConversationList`

### Testing

- Adds unit tests
- Not used in production code yet

---

### Checklist


WireAPI/Sources/WireAPI/APIs/ConversationsAPI/APIModel/QualifiedConversationMember.swift
```diff
@@ -0,0 +1,81 @@
...
+
+struct QualifiedConversationMember: Decodable, ToAPIModelConvertible {
+
+    struct Service: Decodable, ToAPIModelConvertible {
+        let id: UUID
+        let provider: UUID
+
+        func toAPIModel() -> WireAPI.Service {
+            // TODO: in WireRequestStrategy the object had also UUIDs
+            // but we want to unify the model classes. So does it work with Strings like that?
```
markusfassbender: This is an open question. @typfel I think you added the `WireAPI.Service` with string, do you have some opinion or knowledge here?

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
```diff
@@ -96,3 +124,115 @@ private struct PaginatedConversationIDsV0: Decodable, ToAPIModelConvertible {
...
+        mlsGroupID = try container.decodeIfPresent(String.self, forKey: .mlsGroupID)
+        name = try container.decodeIfPresent(String.self, forKey: .name)
+        qualifiedID = try container.decodeIfPresent(QualifiedID.self, forKey: .qualifiedID)
+        readReceiptMode = try container.decodeIfPresent(Int.self, forKey: .readReceiptMode)
+        teamID = try container.decodeIfPresent(UUID.self, forKey: .teamID)
+        type = try container.decodeIfPresent(Int.self, forKey: .type)
+
+        // parsing for v0
+        legacyAccessRole = try container.decodeIfPresent(String.self, forKey: .accessRole)
+        accessRoles = try container.decodeIfPresent([String].self, forKey: .accessRoleV2)
```
johnxnguyen: Is this custom decoding necessary? If we rely on the generated conformance, then we'd still understand which of the `accessRole` and `accessRoleV2` fields get mapped to `legacyAccessRole` and `acessRoles` when converting to the api model.

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV3.swift
```diff
@@ -20,4 +20,143 @@ import Foundation
...
+        name = try container.decodeIfPresent(String.self, forKey: .name)
+        qualifiedID = try container.decodeIfPresent(QualifiedID.self, forKey: .qualifiedID)
+        readReceiptMode = try container.decodeIfPresent(Int.self, forKey: .readReceiptMode)
+        teamID = try container.decodeIfPresent(UUID.self, forKey: .teamID)
+        type = try container.decodeIfPresent(Int.self, forKey: .type)
+
+        // v3 replaces the field "access_role_v2" with "access_role".
+        // However, since the format of update events does not depend on versioning,
+        // we may receive conversations from the `conversation.create` update event
+        // which still have both "access_role_v2" and "access_role" fields
```
johnxnguyen: Perhaps we shouldn't use this struct for decoding update events containing conversations. If we didn't, would we still need this custom decoding?

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV5.swift
```diff
@@ -20,4 +20,151 @@ import Foundation
...
+        // Remove in v5: remove handling of error code 400
+        return try ResponseParser()
+            .success(code: 200, type: QualifiedConversationListV5.self) // Change in v5
+            .parse(response)
+    }
+}
+
+// MARK: Decodables
+
+private struct QualifiedConversationListV5: Decodable, ToAPIModelConvertible {
```
johnxnguyen, line 45: As I understand, the success payload didn't change at all right? If that's the case, couldn't we keep using the previous version?

WireAPI/Sources/WireAPI/Models/Conversation/Conversation.swift
```diff
@@ -0,0 +1,41 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public struct Conversation {
```
johnxnguyen, line 22: It'd be nice to add some doc comments to this public type. Also I think we could also have  more stricter types to replace the strings. What do you think?

WireAPI/Sources/WireAPI/Models/Conversation/ConversationList.swift
```diff
@@ -0,0 +1,23 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+public struct ConversationList {
```
johnxnguyen, line 20: Doc comments would be nice on this public type.

WireAPI/Sources/WireAPI/Models/Conversation/ConversationMember.swift
```diff
@@ -0,0 +1,36 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension Conversation {
```
johnxnguyen, line 21: Doc comments would be nice on this public type.

WireAPI/Sources/WireAPI/Models/Conversation/ConversationMembers.swift
```diff
@@ -0,0 +1,24 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+extension Conversation {
+    public struct Members {
```
johnxnguyen, line 22: Doc comments would be nice on this public type.

WireAPI/Tests/WireAPITests/APIs/ConversationsAPI/ConversationsAPITests.swift
```diff
@@ -210,4 +214,345 @@ final class ConversationsAPITests: XCTestCase {
             XCTFail("expected error 'FailureResponse'")
         }
     }
+
+    // MARK: getConversations
+
+    func testGetConversations_givenV0_thenVerifyRequests() async throws {
```
johnxnguyen: I only see snapshots for V0, V2, V3, and V5. I think though we should have snapshots for all versions, since they all have different paths.

Perhaps we should have a `verifyRequestForAllVersions` test like we do in other API tests?

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV5.swift
markusfassbender: The issue I faced here is that each of these `QualifiedConversationList*` objects refers to a conversation like `let found: [ConversationV5]`. So V4 would use `ConversationV4`, which doesn't contain `cipherSuite` and `epochTimestamp`.

I've tried to make a generic version of `QualifiedConversationList<Conversation: Decodable & ToAPIModelConvertible>`, but as far as I got I didn't make the compiler understand that the generic is always transformable into `found: [Conversation]` for the final `public ConversationList`.

Do you know another approach in mind?

WireAPI/Tests/WireAPITests/APIs/ConversationsAPI/ConversationsAPITests.swift
markusfassbender: right, done in 1a5ca4a42ba9d1b4d63d8ca5b2b49378531a9797 and ecf8b97cb348016943f1eaed9db93c5db66e12de

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV3.swift
markusfassbender: right, good catch 03fd6767da32b530f841978ffcafbdfa58dcaf9b

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
markusfassbender: right, good catch https://github.com/wireapp/wire-ios/commit/03fd6767da32b530f841978ffcafbdfa58dcaf9b

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV3.swift
markusfassbender: Not sure I understand your suggestion, could you please explain a bit more? thanks

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
```diff
@@ -72,9 +72,37 @@ class ConversationsAPIV0: ConversationsAPI, VersionedAPI {
...
+            .success(code: 200, type: QualifiedConversationListV0.self)
+            .failure(code: 400, error: ConversationsAPIError.invalidBody)
+            .parse(response)
+    }
 }
 
-// MARK: -
+// MARK: Encodables
+
+struct GetConversationsParametersV0: Encodable {
```
netbe, line 97: can it be private ?

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
```diff
@@ -96,3 +124,91 @@ private struct PaginatedConversationIDsV0: Decodable, ToAPIModelConvertible {
...
+            notFound: notFound,
+            failed: failed
+        )
+    }
+}
+
+// MARK: -
+
+struct ConversationV0: Decodable, ToAPIModelConvertible {
+    enum CodingKeys: String, CodingKey {
```
netbe, line 153: same for those if it's in this file, i would expect to be private otherwise to be in a separate file

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV5.swift
```diff
@@ -20,4 +20,151 @@ import Foundation
...
+        let resourcePath = "\(pathPrefix)/conversations/list"
+
+        let request = HTTPRequest(
+            path: resourcePath,
+            method: .post,
+            body: body
+        )
+        let response = try await self.httpClient.executeRequest(request)
+
+        // Remove in v5: remove handling of error code 400
```
netbe:
```suggestion
        // Removed in v5: remove handling of error code 400
```

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
markusfassbender: No, it's also used in V2, because only the path has changed but the enconding/decoding stayed the same. To avoid not necessary duplication I use the same structs.

I think the root is that we decided to use inheritance to make this dance around api versions instead of other solutions. This way we need to implement the whole function again even if a tiny bit changes:
1. create request
2. execute request
3. parse response

Any idea how to solve this case using inheritance in another way? cc @johnxnguyen @typfel

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
markusfassbender: Please see the answer/discussion in the other conversation: https://github.com/wireapp/wire-ios/pull/1518#discussion_r1628874021

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
johnxnguyen: For me it's fine to have this duplication, as long as it's readable and concise. For this reason we added the `ResponseParser` so most of the duplication is factored out. Perhaps we could extend this idea to a `RequestBuilder`, which could look something like:

```swift
let request = RequestBuilder(apiVersion: .v2)
    .path("/conversations/list/v2")
    .post(GetConversationsParametersV0(qualifiedIdentifiers: identifiers))
    .build()
```

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV5.swift
johnxnguyen: Ah right I missed that the conversation payload changed too. Makes sense and we should leave it as it is 👍🏻

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV5.swift
```diff
@@ -20,4 +20,151 @@ import Foundation
...
+        // we may receive conversations from the `conversation.create` update event
+        // which still have both "access_role_v2" and "access_role" fields
+
+        if container.contains(CodingKeys.accessRoleV2) {
+            legacyAccessRole = try container.decodeIfPresent(String.self, forKey: .accessRole)
+            accessRoles = try container.decodeIfPresent([String].self, forKey: .accessRoleV2)
+        } else {
+            legacyAccessRole = nil
+            accessRoles = try container.decodeIfPresent([String].self, forKey: .accessRole)
+        }
```
johnxnguyen: Do I understand that this logic is necessary in order to use this `ConversationsAPIV5` for both payloads from /conversations as well as payloads in update events?

WireAPI/Tests/WireAPITests/APIs/ConversationsAPI/__Snapshots__/ConversationsAPITests/testGetConversations_givenAllAPIVersions_thenVerifyRequests.v1.txt
```diff
@@ -0,0 +1,4 @@
+- HTTPRequest
+  - path: "/conversations/list/v2"
```
johnxnguyen: I believe the path here should be `/v1/conversations/list/v2`

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV3.swift
johnxnguyen: I left a question at the same code in the V5 version of this, but I'll repeat it here. Do I understand correctly that this logic allows us to reuse the payload in both /conversation responses as well as in update event payloads? If so, then I would say they should have separate payload models, and here we just have the payload only for versioned api responses, meaning you could rely on the generated decodable conformance.

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/APIModel/QualifiedConversationMember.swift
typfel: So the ID of an service is a UUID according to Swagger so we can definitely change it to that. I guess I was porting the service definition from android where they don't decode the id into a UUID but refer to it as an opaque string.

I think the reason why iOS client historically has been decoding UUIDs is because they can be persisted more efficiently.

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
markusfassbender: Thanks for the suggestion John. I would like to do this idea in a separate PR. @netbe is your question clarified so far?

WireAPI/Tests/WireAPITests/APIs/ConversationsAPI/__Snapshots__/ConversationsAPITests/testGetConversations_givenAllAPIVersions_thenVerifyRequests.v1.txt
markusfassbender: Right, thanks! f116a8049a2397553417de483d31cf6af551fe04

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV5.swift
markusfassbender: Thanks, I didn't understand the comment initially and it was copied from the original decoding. Now it became clear that "update events" is a different channel and should do it's own decoding 👍
57bf86d46884e9109257b3060ded1b0d9efc9900

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV3.swift
markusfassbender: Alright, fixed (see comment on V5): 57bf86d46884e9109257b3060ded1b0d9efc9900

WireAPI/Sources/WireAPI/Models/Conversation/Conversation.swift
markusfassbender: I agree, added some comments on the structs: d7df938c01aa26a0fbe2ca8b3093591a88cc1c55
Unfortunately I don't have knowledge to document all properties.

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/APIModel/QualifiedConversationMember.swift
markusfassbender: Thanks @typfel, updated with b0e55e5c26bce46a9614970acce6c6c1ff1b714a

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV3.swift
```diff
@@ -20,4 +20,131 @@ import Foundation
...
+        members = try container.decodeIfPresent(QualifiedConversationMembers.self, forKey: .members)
+        messageProtocol = try container.decodeIfPresent(String.self, forKey: .messageProtocol)
+        messageTimer = try container.decodeIfPresent(TimeInterval.self, forKey: .messageTimer)
+        mlsGroupID = try container.decodeIfPresent(String.self, forKey: .mlsGroupID)
+        name = try container.decodeIfPresent(String.self, forKey: .name)
+        qualifiedID = try container.decodeIfPresent(QualifiedID.self, forKey: .qualifiedID)
+        readReceiptMode = try container.decodeIfPresent(Int.self, forKey: .readReceiptMode)
+        teamID = try container.decodeIfPresent(UUID.self, forKey: .teamID)
+        type = try container.decodeIfPresent(Int.self, forKey: .type)
+    }
```
johnxnguyen: Is this still necessary?

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV5.swift
```diff
@@ -20,4 +20,135 @@ import Foundation
...
+        members = try container.decodeIfPresent(QualifiedConversationMembers.self, forKey: .members)
+        messageProtocol = try container.decodeIfPresent(String.self, forKey: .messageProtocol)
+        messageTimer = try container.decodeIfPresent(TimeInterval.self, forKey: .messageTimer)
+        mlsGroupID = try container.decodeIfPresent(String.self, forKey: .mlsGroupID)
+        name = try container.decodeIfPresent(String.self, forKey: .name)
+        qualifiedID = try container.decodeIfPresent(QualifiedID.self, forKey: .qualifiedID)
+        readReceiptMode = try container.decodeIfPresent(Int.self, forKey: .readReceiptMode)
+        teamID = try container.decodeIfPresent(UUID.self, forKey: .teamID)
+        type = try container.decodeIfPresent(Int.self, forKey: .type)
+    }
```
johnxnguyen: Is this still necessary?

johnxnguyen (approved):

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV3.swift
markusfassbender: No, thanks removed 9fe392f45956b5508304aa87df73e39e47d79463

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV5.swift
markusfassbender: No, thanks removed 9fe392f45956b5508304aa87df73e39e47d79463

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV0.swift
netbe: yes

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV3.swift
```diff
@@ -67,8 +67,8 @@ private struct QualifiedConversationListV3: Decodable, ToAPIModelConvertible {
 private struct ConversationV3: Decodable, ToAPIModelConvertible {
     enum CodingKeys: String, CodingKey {
         case access
-        case accessRole = "access_role"
-        case accessRoleV2 = "access_role_v2"
+        // Change: replace "access_role_v2" with "access_role".
```
netbe:
```suggestion
        // Changed: replace "access_role_v2" with "access_role".
```

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV3.swift
```diff
@@ -147,7 +114,7 @@ private struct ConversationV3: Decodable, ToAPIModelConvertible {
             id: id,
             lastEvent: lastEvent,
             lastEventTime: lastEventTime,
-            legacyAccessRole: legacyAccessRole,
+            legacyAccessRole: nil, // Remove: `var legacyAccessRole`
```
netbe:
```suggestion
            legacyAccessRole: nil, // Removed: `var legacyAccessRole`
```

netbe (approved):

PR #1523, opened 2024-06-03
refactor: decode federation events WPB-9523

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9523" title="WPB-9523" target="_blank">TaskWPB-9523</a>  [iOS] Decode federation events
  </td></table>
  <br />
 

### Issue

Decode federation events:
- federation connection removed
- federation delete

---

### Checklist


WireAPI/Tests/WireAPITests/EventDecoding/EventDataResource.swift
```diff
@@ -0,0 +1,40 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+struct EventDataResource {
```
markusfassbender: Maybe It's easier to add something to identify this as test/mock?
```suggestion
struct EventDataMockResource {
```

WireAPI/Tests/WireAPITests/Helpers/String+Error.swift
```diff
@@ -0,0 +1,21 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension String: Error {}
```
markusfassbender, line 21: Hm is this a solution for `WireAPITests` in general? not sure if Apple has intended error handling in this way..

WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/UpdateEvent+Federation.swift
```diff
@@ -24,13 +24,57 @@ extension UpdateEvent {
...
+        return FederationConnectionRemovedEvent(domains: payload.domains)
+    }
+
+    private struct FederationConnectionRemovedEventPayload: Decodable {
+
+        let domains: Set<String>
+
+    }
+
+}
```
markusfassbender: Hm should this be an extension with a single func used in a single place? We can do it also without extension quite simple:

```swift
    // switch 
    case .connectionRemoved:
            let payload = try container.decode(FederationConnectionRemovedEventPayload.self, forKey: .payload)
            let event = FederationConnectionRemovedEvent(domains: payload.domains)

// ...

private struct FederationConnectionRemovedEventPayload: Decodable {

    let domains: Set<String>

}
```

WireAPI/Tests/WireAPITests/Helpers/String+Error.swift
johnxnguyen: Often test helpers can throw but the errors are just informative rather than something to handle in the calling code. With this extension you can just throw the error message rather than wrap it in some container struct.

WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/UpdateEvent+Federation.swift
johnxnguyen: When I started with the conversation events, there are a lot of events to decode, and often the event payloads contain more than a simple payload. To avoid having a super long switch and to try improve readability, I created these private extensions, one for each event type.

There are only two federation events, so it may seem unnecessary here, but I kept the same approach to be consistent. I'm happy to also do it inline here.

WireAPI/Tests/WireAPITests/EventDecoding/EventDataResource.swift
johnxnguyen: done 25d8903fbb19a28fffff4e8ed546277255cf03c9

WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/UpdateEvent+Federation.swift
markusfassbender: I see, consistency is also important so either way is fine for me.

WireAPI/Tests/WireAPITests/EventDecoding/EventDataResource.swift
markusfassbender: Unfortunately I can't find the change, maybe something went wrong?

johnxnguyen: strange, clicking on the commit in the previous comment shows the change for me

johnxnguyen: I made the change again 0c8af4a9fba6b8fa355d08181eb6b2ac036dfe34

markusfassbender: Hm yes, now it works. thanks!

markusfassbender (approved):

typfel (approved):

PR #1530, opened 2024-06-04
fix: request loop add to group [WPB-7392]

markusfassbender:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7392" title="WPB-7392" target="_blank">BugWPB-7392</a>  [iOS] Request loop when searching and adding many people to group on group creation
  </td></table>
  <br />
 

### Issue

The issue didn't have real request loops, but in two cases it had false positives:
1. `/v5/search/contacts?q=&size=10`
2. `/v5/teams/<id>/get-members-by-ids-using-post`

#### Case 1
The search makes quite a lot of empty string calls to backend to load the full list of available users. This request will be ignored now and doesn't trigger loop detections.

Add unit tests for excluded paths.

#### Case 2
The call loads information about the members while typing in search member list. This was not so easy to fix, because we don't want many URLs to be ignored by loop detection.

The root cause is, that the content of the request was not considered as unique in the loop detection. This came from an Objective-C `hash` value on `NSDictionary`, which seems to return only the number of entries instead of a unique hash.

To fix this root cause I have changed the content hint to a string and pass a unique value from Swift `hashValue`. 

### Testing

See ticket.

---

### Checklist

### DIFF
--- a/wire-ios-transport/Source/Requests/ZMTransportRequest.m
+++ b/wire-ios-transport/Source/Requests/ZMTransportRequest.m
@@ -138,8 +138,8 @@ typedef NS_ENUM(NSUInteger, ZMTransportRequestSessionType) {
 @property (nonatomic) ZMTransportAccept acceptedResponseMediaTypes; ///< C.f. RFC 7231 section 5.3.2 <http://tools.ietf.org/html/rfc7231#section-5.3.2>
 @property (nonatomic) NSDate *timeoutDate;
 @property (nonatomic) NSMutableArray<NSString *>* debugInformation;
-/// Hash of the content debug information. This is used to identify the content of the request (e.g. detect repeated requests with the same content)
-@property (nonatomic) NSUInteger contentDebugInformationHash;
+
+@property (nonatomic) NSString *contentHintForRequestLoop;
 @property (nonatomic) BOOL shouldCompress;
 @property (nonatomic) NSURL *fileUploadURL;
 @property (nonatomic) NSDate *startOfUploadTimestamp;
@@ -179,7 +179,7 @@ typedef NS_ENUM(NSUInteger, ZMTransportRequestSessionType) {
         self.acceptedResponseMediaTypes = ZMTransportAcceptTransportData;
         self.shouldCompress = shouldCompress;
         self.debugInformation = [NSMutableArray array];
-        self.contentDebugInformationHash = payload.hash;
+        self.contentHintForRequestLoop = payload.contentHintForRequestLoop;
         self.apiVersion = apiVersion;
     }
     return self;
@@ -265,7 +265,7 @@ typedef NS_ENUM(NSUInteger, ZMTransportRequestSessionType) {
         self.acceptedResponseMediaTypes = acceptHeaderType;
         self.shouldCompress = shouldCompress;
         self.debugInformation = [NSMutableArray array];
-        self.contentDebugInformationHash = data.hash;
+        self.contentHintForRequestLoop = [NSString stringWithFormat:@"%lu", data.hash];
         self.apiVersion = apiVersion;
     }
     return self;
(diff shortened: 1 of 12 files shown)


wire-ios-transport/Source/Requests/ZMTransportRequest.m
```diff
@@ -853,7 +853,6 @@ - (void)setDebugInformationState:(NSObject *)state;
 - (void)addContentDebugInformation:(NSString *)debugInformation
 {
     [self.debugInformation addObject:debugInformation];
-    self.contentDebugInformationHash ^= debugInformation.hash;
 }
```
markusfassbender, line 856: This was a side effect, that needs to be done carefully. Therefore I mimic the old behaviour in the places calling `addContentDebugInformation `, see 36100be721cb014522b3eaf31f1bb31cac751ccd.

wire-ios-transport/Source/URLSession/RequestLoopDetection.swift
```diff
@@ -105,24 +102,46 @@ extension RequestLoopDetection: RequestRecorder {
...
+
+struct IdentifierDate {
     let identifier: String
     let date: Date
     let path: String
 
-    init(path: String, contentHash: UInt, date: Date) {
-        self.identifier = "\(path)[\(contentHash)]"
+    init(path: String, contentHint: String, date: Date) {
+        self.identifier = "\(path)[\(contentHint)]"
```
markusfassbender, line 144: I decided to switch the type to `String`, because the Objective-C hashes were `UInt` while Swift is `Int` and it's actually possible the caller adds other strings too.

wire-ios-transport/Source/TransportSession/ZMTransportSession.m
```diff
@@ -22,6 +22,7 @@
 @import UIKit;
 
 #import <WireTransport/WireTransport-Swift.h>
+#import <libkern/OSAtomic.h>
```
netbe, line 25: ?

wire-ios-transport/Source/URLSession/RequestLoopDetection.swift
netbe: > it's actually possible the caller adds other strings too.
@markusfassbender what do you mean by that?

netbe:
@markusfassbender for case 1, is it the exact query each time, if so maybe we should cache the result to avoid always querying it all the time, what do you think?

wire-ios-transport/Source/Requests/ZMTransportRequest.m
netbe: what does `^=` do?

markusfassbender: It's a XOR operator

wire-ios-transport/Source/TransportSession/ZMTransportSession.m
markusfassbender: Import was only moved in the list, because it has `import <` and the other imports are `import "`.

wire-ios-transport/Source/URLSession/RequestLoopDetection.swift
markusfassbender: In 3 places like `AssetRequestFactory` I left a comment in code:
```swift
        // [WPB-7392] through a refactoring the `contentHintForRequestLoop` was seperated from `addContentDebugInformation`.
        // Not clear if it is necessary to set `contentHintForRequestLoop` here, but keep the original behavior.
        request.addContentDebugInformation("Uploading full asset to \(path)")
        request.contentHintForRequestLoop += "Uploading full asset to \(path)"
```

Before the string's hash was used to make the request unique, now we just use the string directly instead of having some bit operations.

markusfassbender:
@netbe Regarding your question for Case 1 I created a refactoring ticket in Jira to look closer: [[iOS] refactor empty search network requests](https://wearezeta.atlassian.net/browse/WPB-9573)

David-Henner (approved):

netbe (approved):

PR #1532, opened 2024-06-05
chore: remove log mask name

markusfassbender:
### Issue

The log mask for name `CLIENT_NAME_C1_C2_C3` introduced with https://github.com/wireapp/wire-ios/pull/1213 masked too many places.

### Testing

- Look at looks of a build like `Playground` GitHub action.

---

### Checklist


markusfassbender:
After talk with @netbe we change the name value and thus the removal here is not correct anymore. I close the PR.

PR #1533, opened 2024-06-05
chore: upgrade datadog to 2.x - WPB-7410

markusfassbender:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7410" title="WPB-7410" target="_blank">TaskWPB-7410</a>  iOS: upgrade Datadog SDK to version 2.x and use SPM
  </td></table>
  <br />
 

### Datadog Migration

- Upgrade Datadog from 1.23 to 2.12
- Use Swift Package Manager and remove from Carthage
- Use migration guide https://github.com/DataDog/dd-sdk-ios/blob/2.0.0/MIGRATION.md
- To better understand the old behaviour I have downloaded the [source code version 1.23](https://github.com/DataDog/dd-sdk-ios/releases/tag/1.23.0) and read in Xcode.

### Rewrite as Swift Package
Because the integration via Swift Package compiles and code signs the code, we can not just remove the binaries from Datadog frameworks as it was done before via `datadog-conditional-framework.sh`.

The solution I chose is to define a new Swift Package wrapper around the Datadog frameworks and only include source code, if the env var `DATADOG_APP_ID` is available. See README.md for more details.

Using a Swift Package also made me to take some refactorings:
- Separate logging code in `WireCommonComponents` from the Swift Package.
- Remove `DATADOG_IMPORT` as `Boolean` from Code
- Use singletons like `WireLogger.provider` on a high level and keep the code in the Swift Package clean

### Testing

- Run `Test Develop` https://github.com/wireapp/wire-ios/actions/runs/9384612058
- Build Playground with Datadog enabled, validate in analytics tool
- Build C1 with Datadog disabled, validate in analytics tool

---

### Checklist


netbe:
@markusfassbender did you check the datadog sdks are removed if DATADOG_IMPORT not set ? see `datadog-conditional-framework.sh`

David-Henner (approved):
why is there a lot more swift packages than what was removed from Carthage?

markusfassbender:
- @David-Henner because it has sub-dependencies that need to be compiled. In Carthage the binary it was also an older version and precompiled (as far as I know).
- @netbe will look into this point, good catch!

WireAnalytics/Sources/WireAnalytics/WireAnalyticsVoidTracker.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+public struct WireAnalyticsVoidTracker: WireAnalyticsTracking {
+
+    public init() { }
+
+    public var datadogUserId: String = "NONE"
```
markusfassbender: Better use `nil` and have an transformation in the UI?

WireAnalytics/Sources/WireAnalytics/WireAnalyticsVoidTracker.swift
```diff
@@ -0,0 +1,26 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+public struct WireAnalyticsVoidTracker: WireAnalyticsTracking {
```
markusfassbender: Maybe the name should be different?

WireAnalytics/Sources/WireDatadogTracker/WireDatadogTracker.swift
```diff
@@ -0,0 +1,120 @@
...
+        identifierForVendor: UUID?,
+        environmentName: String
+    ) {
+        applicationID = appID
+
+        if let identifierForVendor {
+            userIdentifier = Self.hashedDatadogUserIdentifier(identifierForVendor)
+        } else {
+            userIdentifier = "none"
+        }
```
markusfassbender: This behavior was migrated from the old implementation. Does it make sense to do enable datadog tracking if the `userIdentifier` will change later once `identifierForVendor` becomes available?

https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor
> If the value is nil, wait and get the value again later. This happens, for example, after the device has been restarted but before the user has unlocked the device.

WireAnalytics/Package.swift
```diff
@@ -0,0 +1,58 @@
...
+
+let package = Package(
+    name: "WireAnalytics",
+    platforms: [
+        .iOS(.v15),
+        .macOS(.v12)
+    ],
+    products: [
+        .library(
+            name: "WireAnalytics",
```
caldrian, line 14: Don't we have to embed/link these libraries into Wire-iOS?

markusfassbender: No, because it's linked to `WireCommonComponents`, that are linked to Wire-iOS.

WireAnalytics/Sources/WireDatadogTracker/WireDatadogTracker.swift
netbe: I didn't catch that the first time... I don't see how the app could run without user unlocked the device, or maybe the NSE, in that case I don't think it can even go further as all files should be restricted access

WireAnalytics/Package.swift
```diff
@@ -0,0 +1,58 @@
+// swift-tools-version: 5.10
+
+import Foundation
+import PackageDescription
+
+let package = Package(
+    name: "WireAnalytics",
```
netbe, line 7: just wondering about the name, as one would expect Countly for example.

WireAnalytics/README.md
```diff
@@ -0,0 +1,11 @@
...
+Provides access to Wire's analytics tools and trackers.
+
+## Datadog
+Datadog is a third-party analytics tool used opt-in for some builds. This Swift package ensures that the final app build does not contain any code or binaries from Datadog for privacy reasons by disabling full compilation.
+
+To enable Datadog, set the environment variable `ENABLE_DATADOG` to `true`/`false`. This variable is already read during Xcode's Swift dependency resolution process, even before building.
+
+### Set Up
+- **CI**: Our CI tooling uses GitHub Actions and Fastlane to build the app. For builds that should include Datadog, set the `ENABLE_DATADOG` value.
+- **Local Machine**: For debugging purposes, edit the `Package.swift` file and modify the line that reads `ENABLE_DATADOG`. Ensure you save the file and freshly resolve dependencies in Xcode.
```
netbe, line 11: 👏

WireAnalytics/Sources/WireAnalytics/WireAnalyticsTracking.swift
```diff
@@ -0,0 +1,23 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+public protocol WireAnalyticsTracking {
+
+    /// SHA256 string to identify current device across app and extensions.
```
netbe:
```suggestion
    /// Unique obfuscated string to identify current device across app and extensions.
```
from outside we don't need to know how the string is generated

WireAnalytics/Sources/WireDatadogTracker/WireDatadogTracker.swift
```diff
@@ -0,0 +1,120 @@
...
+        CrashReporting.enable()
+
+        let logsConfiguration = Logs.Configuration()
+        Logs.enable(with: logsConfiguration)
+
+        let loggerConfiguration = Logger.Configuration(
+            name: "iOS Wire App",
+            networkInfoEnabled: true,
+            remoteLogThreshold: logLevel,
+            consoleLogFormat: .shortWith(prefix: "[iOS App] ")
```
netbe: I think we can remove this line as we don't need to output to console in debug mode, it already go through the WireLogger systemLogger

wire-ios/Wire-iOS/Sources/AppDelegate.swift
```diff
@@ -112,8 +112,8 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
 
         WireLogger.appDelegate.info("application:willFinishLaunchingWithOptions \(String(describing: launchOptions)) (applicationState = \(application.applicationState.rawValue))")
 
-        DatadogWrapper.shared?.startMonitoring()
-        DatadogWrapper.shared?.log(level: .info, message: "start app")
+        WireAnalytics.enable()
+        WireAnalytics.shared.info("start app")
```
netbe: wondering why in a case we have a static func and the other we have shared instance

wire-ios/WireCommonComponents/WireAnalytics+LoggerProtocol.swift
```diff
@@ -0,0 +1,123 @@
...
+extension WireAnalyticsVoidTracker: WireSystem.LoggerProtocol {
+    public func debug(_ message: any LogConvertible, attributes: LogAttributes...) { }
+    public func info(_ message: any LogConvertible, attributes: LogAttributes...) { }
+    public func notice(_ message: any LogConvertible, attributes: LogAttributes...) { }
+    public func warn(_ message: any LogConvertible, attributes: LogAttributes...) { }
+    public func error(_ message: any LogConvertible, attributes: LogAttributes...) { }
+    public func critical(_ message: any LogConvertible, attributes: LogAttributes...) { }
+
+    public func addTag(_ key: LogAttributesKey, value: String?) { }
+}
```
netbe: so this is always defined?

WireAnalytics/Package.swift
netbe: but we could put also Countly here, since for certain builds it's not included

netbe (commented):
added some questions

wire-ios/WireCommonComponents/WireAnalytics+LoggerProtocol.swift
markusfassbender: Yes

markusfassbender: I completely remove `WireAnalyticsVoidTracker` now

WireAnalytics/Package.swift
```diff
@@ -0,0 +1,58 @@
...
+            name: "WireAnalytics",
+            targets: ["WireAnalytics"]
+        ),
+        .library(
+            name: "WireDatadog",
+            targets: ["WireDatadog"]
+        )
+    ],
+    dependencies: [
+        .package(url: "https://github.com/DataDog/dd-sdk-ios.git", exact: "2.12.0")
```
caldrian, line 23: Lot's of dependencies are pulled in by this package.
It would be nice to use a pre-built XCFramework instead.
But we can do this in a later PR.

WireAnalytics/Sources/WireDatadog/WireDatadog.swift
```diff
@@ -0,0 +1,119 @@
...
+import DatadogTrace
+import UIKit
+
+public final class WireDatadog {
+
+    private let applicationID: String
+    private let logLevel: LogLevel = .debug
+    private let _userIdentifier: String
+
+    public var userIdentifier: String? { _userIdentifier }
```
caldrian: Is there a difference between the String "none" (line 48) and `nil`? Maybe the protocol should declare `userIdentifier` as non-optional String?

WireAnalytics/Sources/WireDatadog/WireDatadog.swift
```diff
@@ -0,0 +1,119 @@
...
+        clientToken: String,
+        identifierForVendor: UUID?,
+        environmentName: String
+    ) {
+        applicationID = appID
+
+        if let identifierForVendor {
+            _userIdentifier = Self.hashedDatadogUserIdentifier(identifierForVendor)
+        } else {
+            _userIdentifier = "none"
```
caldrian: My personal preference would be to have an empty string here instead and only in the very last moment, before sending it through DataDogs API replace an empty string by "none".
Feel free to ignore this comment.

WireAnalytics/Sources/WireDatadog/WireDatadog.swift
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    // MARK: Static Helpers
+
+    private static func hashedDatadogUserIdentifier(_ uuid: UUID) -> String {
+        let data = Data(uuid.uuidString.utf8)
+
+        return SHA256.hash(data: data)
+            .compactMap { String(format: "%02x", $0) }
+            .joined()
```
caldrian, line 127: optional: Converting the identifier-bytes to a string representation is also something we could do as last step before passing it into DataDogs API.

wire-ios-system/Source/Logging/WireLogger+Datadog.swift
```diff
@@ -0,0 +1,27 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+extension WireLogger {
+    public static func addDatadog(_ datadog: LoggerProtocol) {
+        if let aggregatedLogger = Self.provider as? AggregatedLogger {
+            aggregatedLogger.addLogger(datadog)
+        } else {
+            Self.provider = datadog
```
caldrian, line 24: I haven't checked where `provider` is used, but just from seeing this line here I wonder if it shouldn't try to not replace the existing provider but again use the `AggregatedLogger`, somehow like
```suggestion
            Self.provider = AggregatedLogger(loggers: [Self.provider, datadog])
```
Maybe a check is needed if the current provider isn't already a or the datadog logger.

wire-ios/WireCommonComponents/Analytics/WireAnalytics+LoggerProtocol.swift
```diff
@@ -0,0 +1,107 @@
...
+        message: any LogConvertible,
+        error: Error? = nil,
+        attributes: [LogAttributes] = []
+    ) {
+        var plainAttributes: [String: any Encodable] = attributes.reduce(into: [:]) { partialResult, logAttribute in
+            logAttribute.forEach { item in
+                partialResult[item.key.rawValue] = item.value
+            }
+        }
+        plainAttributes["build_number"] = Bundle.main.object(forInfoDictionaryKey: "CFBundleVersion") as? String
```
caldrian: Does it make sense to cache the build number? It never changes.
Isn't that line an expensive operation?

wire-ios/WireCommonComponents/Bundle+WireCommonComponents.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension Bundle {
+    /// Provides access to the Bundle 'WireCommonComponents'.
+    static let wireCommonComponents = Bundle(for: WireCommonComponentsBundle.self)
```
caldrian, line 23: I don't see this being used, do we need to keep it?

caldrian (approved):

WireAnalytics/Sources/WireAnalytics/WireAnalyticsVoidTracker.swift
caldrian: 🤷‍♂️

WireAnalytics/Sources/WireDatadogTracker/WireDatadogTracker.swift
caldrian: Isn't that the case when you ran out of battery, turning on the phone again and receiving missing messages?
I would consider it quite common.

WireAnalytics/Package.swift
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func resolveWireAnalyticsDependencies() -> [Target.Dependency] {
+    // You can enable/disable Datadog for debugging by overriding the boolean.
+    if hasEnvironmentVariable("ENABLE_DATADOG", "true") {
+        return ["WireDatadog"]
+    } else {
+        return []
+    }
+}
```
johnxnguyen, line 50: Cool, I didn't know you could do this.

johnxnguyen (approved):

WireAnalytics/Sources/WireDatadog/WireDatadog.swift
markusfassbender: thanks, will not touch it in this PR

wire-ios-system/Source/Logging/WireLogger+Datadog.swift
markusfassbender: thanks but will not touch it in this PR as the code was just moved

wire-ios/WireCommonComponents/Bundle+WireCommonComponents.swift
markusfassbender: it's used in `WireDatadogBuilder`

wire-ios/WireCommonComponents/Analytics/WireAnalytics+LoggerProtocol.swift
markusfassbender: Right, have addressed it

WireAnalytics/Sources/WireDatadog/WireDatadog.swift
markusfassbender: thanks, I've address it

WireAnalytics/Sources/WireDatadog/WireDatadog.swift
markusfassbender: I could simplify it, thanks

WireAnalytics/Sources/WireDatadogTracker/WireDatadogTracker.swift
markusfassbender: let's keep it for now, as this behaviour existed before

netbe (approved):

PR #1535, opened 2024-06-05
refactor: decode user events WPB-9524

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9524" title="WPB-9524" target="_blank">TaskWPB-9524</a>  [iOS] Decode user events
  </td></table>
  <br />
 

### Issue

- Decode all user events.
- De-duplicate event decoding tests.

---

### Checklist


WireAPI/Sources/WireAPI/Models/UpdateEvent/UserEvent/UserDeleteEvent.swift
```diff
@@ -0,0 +1,34 @@
...
+
+public struct UserDeleteEvent: Equatable {
+
+    /// The user's id.
+
+    public let userID: UUID
+
+    /// The user's qualified id.
+
+    public let qualifiedUserID: QualifiedID
```
typfel, line 32: Isn't this optional? And if it's not why even include non qualified ID?

WireAPI/Sources/WireAPI/Models/User/SupportedProtocol.swift
```diff
@@ -18,9 +18,16 @@
 
 import Foundation
 
-/// The protocols which a user can support
+/// The protocols which a user can support.
 
 public enum SupportedProtocol: String, Equatable, Codable {
+
+    /// The Proetus messaging protocol.
```
typfel, line 25:
```suggestion
    /// The Proteus messaging protocol.
```
```suggestion
    /// The Proetus messaging protocol.
```

WireAPI/Sources/WireAPI/Models/UserClient/MLSPublicKeys.swift
```diff
@@ -0,0 +1,55 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// A container of MLS public keys.
+
+public struct MLSPublicKeys: Equatable, Codable {
+
+    /// The key for the ed25519 ciphersuite.
```
typfel, line 25: `ed25519` is not a ciphersuite itself but instead a signature component that can be used in ciphersuite. Same applies for the properties.

```suggestion
    /// The ed25519 signature key.
```

WireAPI/Tests/WireAPITests/EventDecoding/UserEventDecodingTests.swift
```diff
@@ -0,0 +1,258 @@
...
+final class UserEventDecodingTests: XCTestCase {
+
+    private let helper = EventDecodingAssertionHelper()
+
+    func testDecodingClientAddEvent() throws {
+        try helper.assertEventDecodingFromResource(
+            named: "UserClientAdd",
+            to: .user(.clientAdd(Scaffolding.clientAddEvent))
+        )
+    }
```
markusfassbender: I see the helper as a nice way to avoid duplication. Though in unit tests I would like to quickly to understand what is loaded, tested and expected. 

Also dependencies between files/structs/classes in unit tests causes more issues once one starts to modularise and move tests around. Duplication is not bad for unit tests.

So I would prefer to suggest to use it a bit more "plain":
```swift
    func testDecodingClientAddEvent() throws {
        // Given event data.
        let resource = try MockEventDataResource(name: "UserClientAdd")

        // When decode update event.
        let updateEvent = try JSONDecoder.defaultDecoder.decode(UpdateEvent.self, from: resource.jsonData)

        // Then
        XCTAssertEqual(updateEvent, .user(.clientAdd(Scaffolding.clientAddEvent)))
    }
```

WireAPI/Sources/WireAPI/Models/UserClient/MLSPublicKeys.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    public let p512: String?
+
+    enum CodingKeys: String, CodingKey {
+
+        case ed25519
+        case ed448
+        case p256 = "ecdsa_secp256r1_sha256"
+        case p384 = "ecdsa_secp384r1_sha384"
+        case p512 = "ecdsa_secp512r1_sha512"
```
markusfassbender, line 51: I am not sure how to make sure this really works and is the right values etc... but for now I think we just approve it this way and the testing will be done with the real implementation later?

markusfassbender (approved):

WireAPI/Tests/WireAPITests/EventDecoding/UserEventDecodingTests.swift
markusfassbender: maybe JSONDecoder.defaultDecoder can be a `var mockJSONDecoder` in the test case

johnxnguyen: done e04207ee4b3759ed6fb5c52d78191a2c198fdda2

WireAPI/Sources/WireAPI/Models/UserClient/MLSPublicKeys.swift
johnxnguyen: Yeah I think we'll need to do quite a few playtests. One idea I have is to add the decoding alongside existing update event decoding and check for failures, but don't actually use the results.

netbe (approved):

WireAPI/Sources/WireAPI/Models/UpdateEvent/UserEvent/UserDeleteEvent.swift
johnxnguyen: I was taking the Android code (https://github.com/wireapp/kalium/blob/771a409e17561c490bdfc8f842cb8e04f0791865/network/src/commonMain/kotlin/com/wire/kalium/network/api/base/authenticated/notification/EventContentDTO.kt#L345) as a guide, as I understand also there the qualified id is not declared as optional. I can remove the non qualified id.

PR #1541, opened 2024-06-06
refactor: decode feature config events WPB-9522

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9522" title="WPB-9522" target="_blank">TaskWPB-9522</a>  [iOS] Decode feature config events
  </td></table>
  <br />
 

### Issue

Decode all feature config update events:
- It's just a single event, but with different payloads for each feature.
- I've split up the decoding of each feature and added tests for each of the cases.

### Checklist


WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Event decoding/UpdateEvent+FeatureConfig.swift
```diff
@@ -24,10 +24,309 @@ extension UpdateEvent {
...
+    private struct MLSMigrationConfig: Decodable {
+
+        let startTime: Date?
+        let finaliseRegardlessAfter: Date?
+
+    }
+
+}
+
+// MARK: - SelfDeletingMessagesn update
```
markusfassbender:
```suggestion
// MARK: - SelfDeletingMessages update
```

WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Event decoding/UpdateEvent+FeatureConfig.swift
```diff
@@ -24,10 +24,309 @@ extension UpdateEvent {
...
+            status: payload.status,
+            enforcedTimeoutSeconds: payload.config.enforcedTimeoutSeconds
+        )
+    }
+
+    private struct SelfDeletingMessagesConfig: Decodable {
+
+        let enforcedTimeoutSeconds: UInt
+
+    }
```
markusfassbender: I would like to take this example of the pattern used here to question if it's the responsibility of the `KeyedDecodingContainer` to know about the final types and map the decoded values to a non-decodable object?
From my understanding it's rather a concrete container that provides access to the decoded storage.

Therefore my suggestion would be to create a new struct (instead of extending the container)  to provide the required functionality of mapping. It also has the benefits of a pure function without any state or modification from outside:
```swift
private struct SelfDeletingMessagesConfigDecoder {
    private struct Config: Decodable {
        let enforcedTimeoutSeconds: UInt
    }

    func decode(
        from container: KeyedDecodingContainer<FeatureConfigEventCodingKeys>
    ) throws -> SelfDeletingMessagesFeatureConfig {
        let payload = try container.decode(
            FeatureWithConfig<Config>.self,
            forKey: .payload
        )

        return SelfDeletingMessagesFeatureConfig(
            status: payload.status,
            enforcedTimeoutSeconds: payload.config.enforcedTimeoutSeconds
        )
    }
}
```

Calling:
- instead of `try container.decodeSelfDeletingMessagesConfig()`
- one can use `try SelfDeletingMessagesConfigDecoder().decode(from: container)`

johnxnguyen: Good point, I'll make the change across all event decoding.

johnxnguyen: done

WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Event decoding/UpdateEvent+FeatureConfig.swift
johnxnguyen: fixed

WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Event decoding/UpdateEvent+FeatureConfig.swift
markusfassbender: Amazing, thanks!

markusfassbender (approved):

netbe (approved):

PR #1547, opened 2024-06-10
fix: `ConversationViewController` back button hang WPB-8976

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8976" title="WPB-8976" target="_blank">BugWPB-8976</a>  hang during call - ConversationViewController backButton
  </td></table>
  <br />
 

### Issue

We detected a 549ms hang during a call

By tracking it in the time profiler, we found that there were periodic calls to `ConversationViewController.conversationListDidChange` due to receiving messages.

As a result, `updateLeftNavigationBarItems()` is called each time. It refreshes the `backButton`.

`backButton` is a computed property that would always execute a fetch request to determine if there are any unread conversations.

This resulted in a hang of the UI

### Solution

Moved the fetch request to a Task in order to free up the main thread for the UI

### Measurements

To measure the impact of this change, I recorded traces of the current version and the fix version.
The scenario being recorded was a 3 minutes snapshot of a single call running, while a spam bot sent a message every 3 seconds in another conversation. That is a total of 60 messages, i.e: ~60 calls to `updateLeftNavigationBarItems()`

Screenshot 2024-06-10 at 11 45 45

⬆️ The 1st trace, current version ⬆️ 

Screenshot 2024-06-10 at 11 46 22

⬆️ The 2nd trace, fixed version ⬆️ 

**Conclusion:** we can see there is a decrease of 1.9% weight on the main thread

### Checklist


wire-ios/Wire-iOS/Sources/Helpers/UIBarButtonItem+StyleKit.swift
```diff
@@ -37,6 +37,13 @@ extension UIBarButtonItem {
         )
     }
 
+    func setIcon(_ icon: StyleKitIcon) {
+        image = icon.makeImage(
+            size: .tiny,
```
David-Henner:
```suggestion
            size: .tiny,
```

agisilaos (approved):
Looking good. 🙌

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+NavigationBar.swift
```diff
@@ -111,25 +111,45 @@ extension ConversationViewController {
     }
 
     var backButton: UIBarButtonItem {
```
caldrian:
```suggestion
    private var backButton: UIBarButtonItem {
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+NavigationBar.swift
```diff
@@ -111,25 +111,45 @@ extension ConversationViewController {
...
         button.accessibilityIdentifier = "ConversationBackButton"
         button.accessibilityLabel = L10n.Accessibility.Conversation.BackButton.description
 
-        if hasUnreadInOtherConversations {
-            button.tintColor = UIColor.accent()
-            button.accessibilityValue = L10n.Localizable.ConversationList.Voiceover.UnreadMessages.hint
+        // `ZMConversation.hasUnreadMessageInOtherConversations` is expensive because it performs a fetch request,
+        // so we call it from a Task to reduce the load on the UI.
+        // If `true` we update the button icon, If `false` we already have set it up above
+        Task {
```
caldrian: Wouldn't it be cleaner to move this code into `leftNavigationItems(forConversation:)` or `updateLeftNavigationBarItems()`?

caldrian (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+NavigationBar.swift
netbe: @David-Henner I wonder if the task should be retained so if we hit this path rapidly we cancel the ongoing task for example

netbe (commented):
add a question

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+NavigationBar.swift
David-Henner: that would need some refactoring since `backButton` is a computed property defined in an extension

KaterinaWire (changes requested):
@David-Henner Please change the base branch to develop, so these changes will be included in the next public release.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+NavigationBar.swift
```diff
@@ -110,24 +110,38 @@ extension ConversationViewController {
...
+        button.tintColor = hasUnread ? UIColor.accent() : nil
+        button.accessibilityValue = hasUnread ? UnreadMessages.hint : nil
+
+        return button
+    }
+
+    func updateBackButton(hasUnread: Bool) {
+
+    }
+
```
David-Henner:
```suggestion
```

netbe (approved):
lgtm;)

David-Henner:
@netbe should this be rebased on develop?

KaterinaWire (approved):

PR #1561, opened 2024-06-12
refactor: remove notification context `ZMUserSession` - WPB-9862

markusfassbender:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9862" title="WPB-9862" target="_blank">TaskWPB-9862</a>  [iOS] remove notification context ZMUserSession
  </td></table>
  <br />
 

### Issue

Remove `ZMUserSession` conformance to `NotificationContext`.

### Testing

- test network reachability

---

### Checklist


johnxnguyen (approved):

wire-ios-sync-engine/Source/UserSession/NotificationInContext+UserSession.swift
```diff
@@ -18,25 +18,36 @@
 
 import WireDataModel
 
-extension ZMUserSession: NotificationContext { } // Mark ZMUserSession as valid notification context
```
markusfassbender, line 21: This removal was the reason to start the PR 😉

wire-ios-sync-engine/Tests/Source/Integration/SlowSyncTests.m
```diff
@@ -162,13 +162,12 @@ - (void)testThatItDoesAQuickSyncOnStartupIfAfterARestartWithoutAnyPushNotificati
...
 - (void)testThatTheUIIsNotifiedWhenTheSyncIsComplete
 {
-    // given
     NetworkStateRecorder *stateRecoder = [[NetworkStateRecorder alloc] init];
-    id token = [ZMNetworkAvailabilityChangeNotification addNetworkAvailabilityObserver:stateRecoder userSession:self.userSession];
-    
+    [stateRecoder observe];
+
     // when
     XCTAssertTrue([self login]);
```
markusfassbender, line 181: 💡Insights

After my first iteration of changes this test was failing. I found out, that the internal used `userSession.managedObjectContext.notificationContext` was `nil`, because `userSession` was not instantiated and only get instantiated automagically after `[self login]`.

To solve it I couldn't run `[self login]` first, because as side effect the `userSession.init` triggered the network state changes  and add the the observer in the test next line missed the published changes!

The solution I chose is to skip `NotificationInContext` in the chain and add the `NotificationCenter` observer with `nil` value as it has been before implicitly.

> [!WARNING]
Just because of Objective-C compiler doing no optional check it was possible that a nil value was passed to not optional Swift code, that passed it to `NotificationCenter` where the `object` parameter was optional.. and it worked for a long time.

johnxnguyen (approved):

agisilaos (approved):

KaterinaWire (approved):

PR #1563, opened 2024-06-12
refactor: remove initial sync observer

markusfassbender:
### Issue

Remove not needed dependency to `ZMUserSession` and protocols, classes, static funcs.

### Testing

- unit tests `testThatItNotifiesObserverWhenInitialIsSyncCompleted`

---

### Checklist


wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
```diff
@@ -123,7 +123,18 @@ extension ConversationListViewController.ViewModel {
     func setupObservers() {
 
         if let userSession = ZMUserSession.shared() {
-            initialSyncObserverToken = ZMUserSession.addInitialSyncCompletionObserver(self, userSession: userSession)
+            let context = userSession.managedObjectContext
+
+            initialSyncObserverToken = NotificationCenter.default.addObserver(
```
caldrian: From the docs:
You must invoke [removeObserver(_:)](https://developer.apple.com/documentation/foundation/notificationcenter/1413994-removeobserver) or [removeObserver(_:name:object:)](https://developer.apple.com/documentation/foundation/notificationcenter/1407263-removeobserver) before the system deallocates any object that [addObserver(forName:object:queue:using:)](https://developer.apple.com/documentation/foundation/notificationcenter/1411723-addobserver) specifies.

In the closure we have a weak `self` reference, however, the `context` is captured.

Also, if we switch accounts, the `notificationContext` is probably deallocated, which violates the API, if I understand correctly ("before the system deallocates any object that addObserver specifies").

Note: if `setupObservers()` is called more than once, the old token is overwritten.

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsViewController.swift
```diff
@@ -59,9 +59,23 @@ final class GroupDetailsViewController: UIViewController, ZMConversationObserver
...
+                    let context = session.managedObjectContext
+
+                    initialSyncToken = NotificationCenter.default.addObserver(
+                        forName: .initialSync,
+                        object: context.notificationContext,
+                        queue: nil
+                    ) { [weak self] _ in
+                        context.performGroupedBlock {
+                            self?.didCompleteInitialSync = true
+                            self?.initialSyncToken = nil
```
caldrian, line 75: Same here, `initialSyncToken` must be passed to `removeObserver`.
Setting the property to `nil` is not sufficient.

caldrian (changes requested):

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
markusfassbender: You are right, thank you! Missed this when switching from `NotificationInContext`, that uses a self unregister token. Reverted some changes, can you please take a look?

caldrian (approved):

agisilaos (approved):

PR #1567, opened 2024-06-14
refactor: unversion BackendInfoAPI - WPB-9053

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9053" title="WPB-9053" target="_blank">TaskWPB-9053</a>  [iOS] Make BackendInfoAPI unversioned
  </td></table>
  <br />
 

### Issue

- Unversion `BackendInfoAPI` since it does not call a versioned endpoint
- Align the tests with other API tests.

---

### Checklist


WireAPI/Tests/WireAPITests/APIs/BackendInfoAPI/BackendInfoAPITests.swift
```diff
@@ -16,47 +16,48 @@
...
-            httpClient.receivedRequest,
-            HTTPRequest(
-                path: "/api-version",
-                method: .get
-            )
-        )
+    // MARK: - Request generation
+
+    func testGetBackendInfoRequest() async throws {
+        try await apiSnapshotHelper.verifyRequest(for: [.v0]) { sut in
```
markusfassbender: Hmm I think it's bit strange to have the unversioned implementation run with `v0` and also the snapshot contains it in the name `testGetBackendInfoRequest.v0.txt`.

We could add a func to `APISnapshotHelper` that works unversioned or few cases just use `HTTPRequestSnapshotHelper` directly and turn `resourceName: String` optional as the snapshot testing library would accept it optional.

johnxnguyen: done 7f01739c88216e2f37fe33d3d48ed6eace8c3593

WireAPI/Tests/WireAPITests/APIs/BackendInfoAPI/BackendInfoAPITests.swift
```diff
@@ -16,47 +16,40 @@
...
-@testable import WireAPI
+import SnapshotTesting
 import XCTest
 
+@testable import WireAPI
+
 final class BackendInfoAPITests: XCTestCase {
 
-    private var httpClient: HTTPClient!
+    private let snapshotter = HTTPRequestSnapshotHelper()
```
markusfassbender, line 26: Hint: due to a not intuitive memory handling by Xcode properties like this will stay in memory until all test cases finished execution. Therefore usually it's recommended to create properties in `setUp` and nullify in `tearDown`.

From my point of view now in this case keep it as it is.

markusfassbender (approved):

netbe (approved):

PR #1570, opened 2024-06-14
refactor: notification context usersession - WPB-9862

markusfassbender:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9862" title="WPB-9862" target="_blank">TaskWPB-9862</a>  [iOS] remove notification context ZMUserSession
  </td></table>
  <br />
 

### Issue

It's an anti pattern, that the `managedObjectContext` provide access to the notification context, that has nothing to do with core data. Even is has a implicit dependency to `persistentStoreCoordinator`.

To make a first step the `UserSession` protocol provides direct access to the notification context, so that the core data classes are not involved anymore.

### Solution idea
In future the dependency on `persistentStoreCoordinator` can be replaced. One solution to have a unique object per user session could be just to use the account identifier.
```swift
extension NSManagedObjectContext {
    @objc
    var notificationContext: NotificationContext { persistentStoreCoordinator! }
}
```

### Testing

- Run unit test suite

---

### Checklist


David-Henner (approved):

wire-ios-data-model/Source/ManagedObjectContext/NSManagedObjectContext+NotificationContext.swift
```diff
@@ -25,6 +25,7 @@ extension NSPersistentStoreCoordinator: NotificationContext {}
 
 public extension NSManagedObjectContext {
 
+    /// Prefer `UserSession.notificationContext` over this property.
```
caldrian, line 28: What about declaring it deprecated?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+Notifications.swift
```diff
@@ -0,0 +1,26 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
```
caldrian, line 19: not needed

johnxnguyen (approved):

caldrian (approved):

wire-ios-data-model/Source/ManagedObjectContext/NSManagedObjectContext+NotificationContext.swift
markusfassbender: IMO it's not a good idea, because the property can not be simply replaced in many places. it will produce a lot of warnings (hundreds?). see `ZMUserSession.shared`, that pollutes the warnings, but can not be fixed easily.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+Notifications.swift
markusfassbender: I see, it's the only thing not to merge, so I skip it for now and the import can be removed later.

PR #1580, opened 2024-06-17
fix: missing messages - WPB-9221

typfel:
<table>
<td>
  </td></table>
  <br />
 

### Issue

We have observed cases where messages have been missing from conversations. While investigating once such case we could determine this was caused by the application crashing while decrypting messages. When the app was started again it would attempt to decrypt the messages again, but would now fail with a "duplicate message" error since the cryptographic state had already been updated.

The application crashed because while decrypting we are holding a file lock on the cryptographic state which is not allowed when the app gets suspended, which causes the system to kill the app with the `0xdead10cc` (deadlock) error.

#### Solution

Preferably we would only save the cryptographic state after persisting the decrypted message but that's currently not possible with the current API so we do our best to mitigate the issue.

1. Release the file lock and stop the decryption task before the application is about to get suspended.
2. Persist a decrypted message directly after decrypting it (No batch save).
3. Update the last event ID directly after persisting a decrypted message.
4. Avoid decrypting multiple pages of message in parallel in the notification extension.

Other minor fixes:

1. Don't fetch notification if don't have the self client ID, doing so will corrupt the app state since we then get events intended for other self clients.

### Testing

Runt the app and look for missing messages.

---

### Checklist

### DIFF
--- a/wire-ios-system/Source/ExpiringActivity.swift
+++ b/wire-ios-system/Source/ExpiringActivity.swift
@@ -35,7 +35,7 @@ public struct ExpiringActivityNotAllowedToRun: Error { }
 /// calling [Task.checkCancellation](https://developer.apple.com/documentation/swift/task/checkcancellation()) at the appropriate time.
 ///
 /// - Parameters:
-///   - reason: Discription of what the activity does, helpful for debugging purposes.
+///   - reason: Description of what the activity does, helpful for debugging purposes.
 ///   - block: async operation which supports cancellation.
 
 public func withExpiringActivity(reason: String, block: @escaping () async throws -> Void) async throws {
--- a/wire-ios-system/Source/ExpiringActivity.swift
+++ b/wire-ios-system/Source/ExpiringActivity.swift
@@ -18,6 +18,18 @@
 
 import Foundation
 
+protocol ExpiringActivityInterface {
+
+    func performExpiringActivity(withReason reason: String, using block: @escaping @Sendable (Bool) -> Void)
+
+}
+
+extension ProcessInfo: ExpiringActivityInterface { }
+
+/// The expiring activity is not allowed to run possibly because the background execution time has already expired.
+
+public struct ExpiringActivityNotAllowedToRun: Error { }
+
 /// Execute an async function inside an [performExpiringActivity](https://developer.apple.com/documentation/foundation/processinfo/1617030-performexpiringactivity)
 /// which cancels the task when the activity expires. It's up to the async function to handle the cancellation by for example
 /// calling [Task.checkCancellation](https://developer.apple.com/documentation/swift/task/checkcancellation()) at the appropriate time.
@@ -32,11 +44,21 @@ public func withExpiringActivity(reason: String, block: @escaping () async throw
 }
 
 actor ExpiringActivityManager {
+
+    let api: ExpiringActivityInterface
     var task: Task<Void, Error>?
+
+    init() {
+        self.init(api: ProcessInfo.processInfo)
+    }
+
+    init(api: ExpiringActivityInterface) {
+        self.api = api
+    }
 
     func withExpiringActivity(reason: String, block: @escaping () async throws -> Void) async throws {
         try await withCheckedThrowingContinuation { continuation in
-            ProcessInfo.processInfo.performExpiringActivity(withReason: reason) { expiring in
+            api.performExpiringActivity(withReason: reason) { expiring in
                 if !expiring {
                     let semaphore = DispatchSemaphore(value: 0)
                     Task {
@@ -54,7 +76,11 @@ actor ExpiringActivityManager {
                 } else {
                     WireLogger.backgroundActivity.warn("Background activity is expiring: \(reason)")
                     Task {
-                        await self.stopWork()
+                        do {
+                            try await self.stopWork()
+                        } catch {
+                            continuation.resume(throwing: error)
+                        }
                     }
                 }
             }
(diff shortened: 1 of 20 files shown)


wire-ios-system/Source/ExpiringActivity.swift
```diff
@@ -0,0 +1,107 @@
...
+/// The expiring activity is not allowed to run possibly because the background execution time has already expired.
+
+public struct ExpiringActivityNotAllowedToRun: Error { }
+
+/// Execute an async function inside an [performExpiringActivity](https://developer.apple.com/documentation/foundation/processinfo/1617030-performexpiringactivity)
+/// which cancels the task when the activity expires. It's up to the async function to handle the cancellation by for example
+/// calling [Task.checkCancellation](https://developer.apple.com/documentation/swift/task/checkcancellation()) at the appropriate time.
+///
+/// - Parameters:
+///   - reason: Discription of what the activity does, helpful for debugging purposes.
```
netbe:
```suggestion
///   - reason: Description of what the activity does, helpful for debugging purposes.
```

wire-ios-system/Tests/ExpiringActivityTests.swift
```diff
@@ -0,0 +1,116 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import XCTest
+@testable import WireSystem
+
+class MockExpiringActivityAPI: ExpiringActivityInterface {
```
netbe:
```suggestion
private class MockExpiringActivityAPI: ExpiringActivityInterface {
```

let's move it at bottom of file instead

wire-ios-system/Tests/ExpiringActivityTests.swift
```diff
@@ -0,0 +1,116 @@
...
+
+        // when
+        do {
+            try await sut.withExpiringActivity(reason: "Expiring test activity") {
+                while true {
+                    await Task.yield()
+                    try Task.checkCancellation()
+                }
+            }
+            XCTFail("Expected an cancellation error to be thrown")
```
netbe:
```suggestion
            XCTFail("Expected a cancellation error to be thrown")
```

wire-ios-system/Tests/ExpiringActivityTests.swift
```diff
@@ -0,0 +1,116 @@
...
+}
+
+class ExpiringActivityTests: XCTestCase {
+
+    let concurrentQueue = DispatchQueue(label: "activity queue", attributes: [.concurrent])
+
+    func testThatTaskIsCancelled_WhenActivityExpires() async throws {
+
+        // given
+        let api = MockExpiringActivityAPI()
```
netbe, line 30: Do I understand correctly that this mock compared to ProcessInfo just execute the method directly instead of waiting 30 sec?

typfel: The individual test cases determinate how it behave, but yes we execute directly since we don't want to wait for 30 sec.

netbe (approved):

johnxnguyen:
> Preferably we would only save the cryptographic state after persisting the decrypted message but that's currently not possible with the current API

@typfel why is it not possible?

johnxnguyen:
> The notification extension would attempt decrypt multiple pages of message in parallel.

@typfel I thought proteus messages needed to be decrypted in order (with some out of order tolerance). Don't we risk to increase out of order decryption if we decrypt pages in parallel?

typfel:
> > Preferably we would only save the cryptographic state after persisting the decrypted message but that's currently not possible with the current API
> 
> @typfel why is it not possible?

In theory it is possible and we will do it but requires a new API from CC

johnxnguyen:
> > The notification extension would attempt decrypt multiple pages of message in parallel.
> 
> @typfel I thought proteus messages needed to be decrypted in order (with some out of order tolerance). Don't we risk to increase out of order decryption if we decrypt pages in parallel?

I think I misunderstood... you mean we were decrypting pages in parallel and we actually shouldn't?

typfel:
> > The notification extension would attempt decrypt multiple pages of message in parallel.
> 
> @typfel I thought proteus messages needed to be decrypted in order (with some out of order tolerance). Don't we risk to increase out of order decryption if we decrypt pages in parallel?

I was unclear. Previously we would decrypt pages parallel with my fix we don't.

wire-ios-system/Source/ExpiringActivity.swift
```diff
@@ -0,0 +1,107 @@
...
+
+    init(api: ExpiringActivityInterface) {
+        self.api = api
+    }
+
+    func withExpiringActivity(reason: String, block: @escaping () async throws -> Void) async throws {
+        try await withCheckedThrowingContinuation { continuation in
+            api.performExpiringActivity(withReason: reason) { expiring in
+                if !expiring {
+                    let semaphore = DispatchSemaphore(value: 0)
```
johnxnguyen, line 63: Just checking my understanding is correct: the semaphore ensures that we don't return from the `performExpiringActivity` block until the work block starts? Why is that important?

johnxnguyen (approved):

wire-ios-system/Source/ExpiringActivity.swift
typfel: It ensures that we don't return from the `performExpiringActivity` block until the work block has completed. This is important because the activity is considered completed when you return.

PR #1581, opened 2024-06-18
chore: stop using NSManagedObjectContext category method in Swift - WPB-8907

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8907" title="WPB-8907" target="_blank">TaskWPB-8907</a>  Migrate Objective C code to Swift
  </td></table>
  <br />
 

This PR removes calls from Swift code to the custom NSManagedObjectContext extension method `executeFetchRequestOrAssert` in order to reduce dependencies from Swift to Objective C.

---

### Checklist

### DIFF
--- a/wire-ios-data-model/Source/Model/Conversation/ZMConversation+MLS.swift
+++ b/wire-ios-data-model/Source/Model/Conversation/ZMConversation+MLS.swift
@@ -197,7 +197,7 @@ public extension ZMConversation {
             argumentArray: [Self.mlsGroupIdKey, groupID.data]
         )
 
-        let result = context.executeFetchRequestOrAssert(request)
+        let result = try! context.fetch(request)
         require(result.count <= 1, "More than one conversation found for a single group id")
         return result.first as? ZMConversation
     }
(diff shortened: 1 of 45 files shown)


wire-ios-data-model/Source/Model/Conversation/ZMConversation+MLS.swift
```diff
@@ -197,7 +197,7 @@ public extension ZMConversation {
             argumentArray: [Self.mlsGroupIdKey, groupID.data]
         )
 
-        let result = context.executeFetchRequestOrAssert(request)
+        let result = try! context.fetch(request)
```
johnxnguyen, line 200: Do I understand correctly the objc method also crashes on any error? So the behavior hasn't changed?

johnxnguyen: So it looks like it the previous method also crashes upon any error. I'm not sure how I feel though about crashing, but since it's not a change in behavior I'm fine to discuss it outside of the PR, and at least now it's more obvious that there is a crash.

johnxnguyen (approved):

wire-ios-data-model/Source/Model/Conversation/ZMConversation+MLS.swift
caldrian: > at least now it's more obvious that there is a crash.

I totally agree, this now catches everybody's eye.

agisilaos (approved):

PR #1566, opened 2024-06-14
refactor: decode conversation events - WPB-9520

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9520" title="WPB-9520" target="_blank">TaskWPB-9520</a>  [iOS] Decode conversation events
  </td></table>
  <br />
 

### Issue

Decode all conversation events, remove deprecated events.

---

### Checklist

- [x] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [x] Description is filled and free of optional paragraphs.
- [x] Adds/updates automated tests.

PR #1597, opened 2024-06-19
refactor: distinguish between fractional and non-fractional date strings - WPB-9793

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9793" title="WPB-9793" target="_blank">TaskWPB-9793</a>  [iOS] Make date string encoding and decoding more explicit
  </td></table>
  <br />
 

### Issue

In https://github.com/wireapp/wire-ios/pull/1566#discussion_r1645631952 we discussed whether it was safe to always encode date strings with fractional seconds. The backend doesn't have a general decoding strategy to accept difference date formats, but rather each endpoint may specify what format it expects.

For this reason, it's necessary to encode and decode date strings according to the proper format. Since a single payload may have two date strings of different formats, I have opted to create two different types:

- `UTCTime` (name taken from the Swagger docs) encodes to and decode from date strings **without** fractional seconds.
- `UTCTimeMillis` (also taken from Swagger) encodes to and decodes from date strings **with** fractional seconds.

These types can be used in payloads where necessary. The benefit of this approach is that we don't need to add custom encoding and decoding of payloads containing date strings, nor do we need to use different encodes and decoders (which would only work if all date strings are the same format).

---

### Checklist


agisilaos (approved):

typfel (approved):

WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Event decoding/Conversation/ConversationCreateEventDecoder.swift
```diff
@@ -34,21 +34,161 @@ struct ConversationCreateEventDecoder {
...
+                access: payload.access,
+                accessRoles: payload.accessRoles,
+                legacyAccessRole: payload.legacyAccessRole,
+                lastEvent: payload.lastEvent,
+                lastEventTime: payload.lastEventTime?.date
+            )
         )
     }
 
+    private struct Payload: Decodable {
```
johnxnguyen, line 74: Previously I was just using the API model directly, but that model uses `Date` instances and not the `UTCTime` and `UTCTimeMillis`.  I wasn't sure what was the best approach:

- have a separate private payload struct for the update event, this is inline with other event decoding, comes at the cost of a fair bit of duplication, but perhaps it's warranted.
- key decoding the `Conversation` API model, but change the instances of `Date` to the corresponding `UTCTime` and `UTCTimeMillis`.

If you have some thoughts please let me know.

WireAPI/Sources/WireAPI/Components/UTCTimeMillis.swift
```diff
@@ -0,0 +1,51 @@
...
+
+import Foundation
+
+/// A timestamp that encodes to and decodes from ISO8601 date with
+/// fractional secconds, i.e yyyy-mm-ddThh:MM:ss.qqqZ
+///
+/// See `UTCTime` if encoding and decoding without fractional seconds
+/// is required.
+
+struct UTCTimeMillis: Codable {
```
caldrian, line 27: I might have picked a more descriptive name like `ISO8601TimestampWithFractionalSeconds` but I'm fine with this name also.

WireAPI/Sources/WireAPI/Extensions/JSONEncoder+Default.swift
```diff
@@ -24,12 +24,9 @@ extension JSONEncoder {
 
     static var defaultEncoder: JSONEncoder {
         let encoder = JSONEncoder()
+
+        // This is needed so we can deterministically snapshot request bodies.
         encoder.outputFormatting = .sortedKeys
```
caldrian, line 30: Can't we put `#if DEBUG` around this line, so that in production we don't waste resources?

WireAPI/Sources/WireAPI/Components/UTCTimeMillis.swift
```diff
@@ -0,0 +1,51 @@
...
+            throw DecodingError.dataCorruptedError(
+                in: container,
+                debugDescription: "Expected date string to be ISO8601-formatted with fractional seconds"
+            )
+        }
+    }
+
+    func encode(to encoder: any Encoder) throws {
+        var container = encoder.unkeyedContainer()
+        let string = ISO8601DateFormatter.fractionalInternetDateTime.string(from: date)
```
caldrian, line 47: Nice solution with the two custom timestamp types!

There ist just a little disadvantage over setting a dateDecoding strategy, but it's probably a minor risk and we've been doing it like this before:
If somewhere else in the code we accidentally modify the formatter, since it's a mutable singleton, it will affect encoding and decoding payloads.

But this approach is still a lot better than initializing a new instance in this method.

caldrian (approved):
👏

WireAPI/Sources/WireAPI/Components/UTCTimeMillis.swift
johnxnguyen: I may have too, but I chose to use the same name as in the swagger doc to make it clearer which type to reach for when modeling the payloads.

WireAPI/Sources/WireAPI/Extensions/JSONEncoder+Default.swift
johnxnguyen: Good idea, done! bd4ba7fdad6ec9aa2a5708c648af98a3b5cf1cd3

WireAPI/Sources/WireAPI/Components/UTCTimeMillis.swift
johnxnguyen: Ah right, because it's a reference type. I understand that it can be quite expensive to initialize formatters, maybe there's a way to wrap it to make it immutable.

PR #1601, opened 2024-06-20
chore: skip bundle install in setup.sh on failure - no ticket

caldrian:
### Issue

Currently setup.sh fails hard when no proper ruby version is installed.
This PR instead adds a printed warning and continues with the script.

Screenshot 2024-06-20 at 09 37 59

### Testing

Gems should be installed when a proper ruby version is available, otherwise a clear warning must be shown.

---

### Checklist

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -84,7 +84,7 @@ if [[ -n "${CI-}" ]]; then # skip cache bootstrap for CI
     echo "Skipping install since CI is defined"
 else
     which bundle || gem install bundler
-    bundle check || bundle install || echo "⚠️  Skipping bundle install since no or an old ruby version is installed"
+    bundle check || bundle install || echo "⚠️  Skipping bundle install, please check if you have the required ruby version installed"
 fi
 echo ""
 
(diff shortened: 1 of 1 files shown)


agisilaos (approved):

setup.sh
```diff
@@ -84,7 +84,7 @@ if [[ -n "${CI-}" ]]; then # skip cache bootstrap for CI
     echo "Skipping install since CI is defined"
 else
     which bundle || gem install bundler
-    bundle check || bundle install || echo "⚠️  Skipping bundle install since no or an old ruby version is installed"
+    bundle check || bundle install || echo "⚠️  Skipping bundle install, please check if you have the required ruby version installed"
```
netbe, line 87: @caldrian I am not sure we should just skip it, I'd rather like that the one executing the script match the requirements

caldrian: I don't agree because the gems are not needed for anything locally.
Only for devs who like to execute fastlane and other dependencies need it.

caldrian:
postponed

PR #1602, opened 2024-06-20
fix(actions): prevent command injection - WPB-9709

lwille:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9709" title="WPB-9709" target="_blank">TaskWPB-9709</a>  Fix GHA pipeline command injection vulnerabilities
  </td></table>
  <br />
 

### Issue

Unsanitized user input could be used to execute arbitrary code in the context of the workflow.

related:
- https://wearezeta.atlassian.net/wiki/spaces/SC/pages/1216053328/How+to+avoid+Command+Injection+via+Github+Actions+or+CI+jobs+in+general
- https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#example-of-a-script-injection-attack

---

### Checklist

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -12,11 +12,11 @@ on:
       changelog-name:
         description: "The name of the changelog file"
         value: ${{ jobs.changelog.outputs.changelog-name }}
-    
+
 
 env: # https://docs.fastlane.tools/getting-started/ios/setup/
   LC_ALL: en_US.UTF-8
-  LANG: en_US.UTF-8     
+  LANG: en_US.UTF-8
 
 jobs:
   changelog:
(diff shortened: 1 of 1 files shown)


caldrian (approved):

netbe:
@lwille is it the only place?
what about inputs from:
 https://github.com/wireapp/wire-ios/blob/develop/.github/workflows/_reusable_app_release.yml#L81 or 
https://github.com/wireapp/wire-ios/blob/fac2e0444a30399304802a61ded4659a2d9b79bc/.github/workflows/_reusable_run_tests.yml#L132 for example?

lwille:
@netbe using inputs in `if`, `where`, `env` etc. Is completely fine. Only in a `run` step, templating any kind of user-defined input could be used to directly inject a command.

This is the only place I could spot.
I used [code search](https://github.com/search?q=%28org%3Awireapp+OR+org%3Awearezeta+OR+org%3Azinfra%29+path%3A.github%2Fworkflows%2F*.y*ml+%22run%3A%22+%22%24%7B%7B%22&type=code) to identify files with `run` steps that use templating notation, i.e. `${{`.

The GitHub docs have [some examples](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#example-of-a-script-injection-attack) on how such injections work.

netbe (approved):

netbe:
@lwille feel free to merge, thks again for fixing this.

PR #1604, opened 2024-06-20
refactor: add UserRepository - WPB-8704

typfel:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8704" title="WPB-8704" target="_blank">TaskWPB-8704</a>  [iOS] Create UserRepository
  </td></table>
  <br />
 

### Issue

Add `UserRepository` with functions for fetching & persisting all known users.

*Note:* there's already a `UserRepository` which I assume we want to merge with this once at some point, or this problem goes away when we move the repositories to their own module.

---

### Checklist


wire-ios-sync-engine/Source/Repositories/UserRepository_.swift
```diff
@@ -0,0 +1,147 @@
...
+import WireAPI
+import WireDataModel
+
+/// Facilitate access to users related domain objects.
+///
+/// A repository provides an abstraction for the access and storage
+/// of domain models, concealing how and where the models are stored
+/// as well as the possible source(s) of the models.
+
+protocol UserRepositoryProtocol_ {
```
johnxnguyen: I think we can merge this with the existing user repository, which should just mean to add an accessor to the self user.

wire-ios-sync-engine/Source/Repositories/UserRepository_.swift
```diff
@@ -0,0 +1,147 @@
...
+        persistedUser.expiresAt = user.expiresAt
+        persistedUser.serviceIdentifier = user.service?.id.transportString()
+        persistedUser.providerIdentifier = user.service?.provider.transportString()
+        persistedUser.supportedProtocols = user.supportedProtocols?.toDomainModel() ?? [.proteus]
+        persistedUser.needsToBeUpdatedFromBackend = false
+    }
+
+}
+
+extension Collection where Element == WireDataModel.QualifiedID {
```
johnxnguyen: These extensions should probably go into separate files.

wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
```diff
@@ -149,6 +149,7 @@ actor EventProcessor: UpdateEventProcessor {
             let prefetchResult = await syncContext.perform { self.syncContext.executeFetchRequestBatchOrAssert(fetchRequest) }
 
             let eventDescriptions = decryptedUpdateEvents.map {
+                $0.messageNonce
```
johnxnguyen: What does this do?

johnxnguyen (approved):
Nice! I like how clean it looks.

wire-ios-sync-engine/Source/Repositories/UserRepository_.swift
typfel: Ok, but that also means we need to start using the UserRepository with all its dependencies in production code, but yeah shouldn't have any side effect if the new methods aren't used.

I'll give it a try later

wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
typfel: Doesn't belong here I'll revert it

WireAPI/Sources/WireAPI/APIs/UsersAPI/UsersAPI.swift
```diff
@@ -18,8 +18,8 @@
 
 import Foundation
 
+// sourcery: AutoMockable
 /// An API access object for endpoints concerning users.
-
 public protocol UsersAPI {
```
markusfassbender, line 23: 😁👍

wire-ios-sync-engine/Source/Repositories/UserRepository_.swift
```diff
@@ -0,0 +1,105 @@
...
+    /// Fetch and persist a list of users
+    ///
+    /// - parameters:
+    ///     - userIDs: IDs of users to fetch
+
+    func pullUsers(userIDs: [WireDataModel.QualifiedID]) async throws
+
+}
+
+final class UserRepository_: UserRepositoryProtocol_ {
```
markusfassbender: I am not sure why the underscores are required?

wire-ios-sync-engine/Source/Repositories/UserModelMappings.swift
```diff
@@ -0,0 +1,62 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireAPI
+
+extension Collection where Element == WireDataModel.QualifiedID {
```
markusfassbender: As long as these mapping helpers stay `internal`, I'm ok with it.
For public scope I would rather repeat the mapping in place to make the legacy issues visible.

wire-ios-sync-engine/Source/Repositories/UserRepository_.swift
typfel: Because there's was already a UserRepository in `WireSyncEngine` the two should be merged but I didn't want to do it in this PR.

But since the `WireDomain` module has now been merged I can probably drop the _. I'll try

wire-ios-sync-engine/Source/Repositories/UserModelMappings.swift
typfel: The mapping should stay internal.

For base structs like `QualifiedID` I don't think we need separate type in each layer we could share them. The same way we don't have separate types for String and Int in each layer.

markusfassbender: I agree mostly, just keep in mind that this create a dependency to the module where `QualifiedID` is defined.
One could accept it for few types and live with it or think about a refactoring, e.g. introducing a light-weight module `WireDataTypes` or `WireDomainTypes` etc...

markusfassbender: One other thing could be to split feature wise, e.g. `WireSearch` that imports `WireSearchTypes`. Just my thoughts, nothing to change in this PR 👍

wire-ios-sync-engine/Source/Repositories/UserRepository_.swift
typfel: Alright, I merged the functionality from the old UserRepository into the new one, but it turns it's not so easy to adopt the new one in `SupportedProtocolsService` since the new one depends on `UsersAPI` and we currently don't have good way to create instances of APIs.

We need to know the API version to build api structs but we fetch it lazily. Until we have a solution for that I'll not try to bring in these new repositories into our old system.

johnxnguyen: Thanks of trying, we can figure it out later.

wire-ios-sync-engine/Source/Repositories/UserModelMappings.swift
johnxnguyen: Perhaps it could be `WireFoundation` and remain lightweight?

johnxnguyen (approved):

agisilaos (approved):

PR #1608, opened 2024-06-21
fix: refilling prekeys always start from index 0 - WPB-9806

typfel:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9806" title="WPB-9806" target="_blank">BugWPB-9806</a>  When refilling prekeys the iOS client is always starting from index 0
  </td></table>
  <br />
 

### Issue

When re-filling prekeys we were always starting from index 0. If not all local prekeys have been used they would be over-written, and if someone had been holding on to an old prekey and tries use it this will result in a decryption error.

#### Solution

Generate new prekeys after the previous range to avoid any change of overwriting old prekeys.

### Testing

Run out of prekeys and watch the logs to see that the new prekeys have an index higher than the previously generated prekeys.

---

### Checklist

### DIFF
--- a/wire-ios-sync-engine/Source/Synchronization/Strategies/UserClientRequestStrategy.swift
+++ b/wire-ios-sync-engine/Source/Synchronization/Strategies/UserClientRequestStrategy.swift
@@ -323,10 +323,11 @@ public final class UserClientRequestStrategy: ZMObjectSyncStrategy, ZMObjectStra
                 return false
             } else if clientUpdateStatus?.currentPhase == .waitingForPrekeys {
                 clientUpdateStatus?.willGeneratePrekeys()
+                let nextPrekeyIndex = UInt16(userClient.preKeysRangeMax) + 1
                 let groups = managedObjectContext?.enterAllGroupsExceptSecondary()
                 Task {
                     do {
-                        let prekeys = try await prekeyGenerator.generatePrekeys()
+                        let prekeys = try await prekeyGenerator.generatePrekeys(startIndex: nextPrekeyIndex)
                         await managedObjectContext?.perform {
                             self.clientUpdateStatus?.didGeneratePrekeys(prekeys)
                         }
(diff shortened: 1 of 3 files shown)


wire-ios-sync-engine/Source/Synchronization/Strategies/UserClientRequestStrategy.swift
```diff
@@ -323,10 +323,11 @@ public final class UserClientRequestStrategy: ZMObjectSyncStrategy, ZMObjectStra
                 return false
             } else if clientUpdateStatus?.currentPhase == .waitingForPrekeys {
                 clientUpdateStatus?.willGeneratePrekeys()
+                let nextPrekeyIndex = UInt16(userClient.preKeysRangeMax) + 1
```
johnxnguyen, line 326: If anyone is wondering: the `preKeysRangeMax` is set on a successful upload of preKeys.

johnxnguyen (approved):

wire-ios-sync-engine/Source/Synchronization/Strategies/UserClientRequestStrategy.swift
netbe: @typfel question is there a risk the index we arrive at this point before the previous generation task below is finished (async) ?

typfel: > @typfel question is there a risk the index we arrive at this point before the previous generation task below is finished (async) ?

You are asking if it's possible that we have overlapping tasks generating new prekeys? No, because It's controlled to happen serially by the `ClientUpdateStatus.currentPhase`. We only enter this code path if the currentPhase == `.waitingForPrekeys` and while the task is running currentPhase == `.generatingPrekeys` and afterwards currentPhase == `.done` until the prekeys have been uploaded.

netbe (approved):

PR #1614, opened 2024-06-24
feat: CRL Proxy support if needed - WPB-8795

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8795" title="WPB-8795" target="_blank">TaskWPB-8795</a>  [iOS]:E2EI: CRL Proxy for iOS
  </td></table>
  <br />
 

### Issue
For some clients,  we need to use the proxy to fetch CRLs (certificate revocation lists) from different domains.

### Solution
The `mlsE2EId` feature flag has been expanded with two new fields: : `crlProxy` and `useProxyOnMobile`.
```
	"mlsE2EId": {
		"config": {
			"acmeDiscoveryUrl": "",
			"crlProxy": "",
			"useProxyOnMobile": true,
			"verificationExpiration": 3600
		}
	}
```
- **crlProxy** is a mandatory field, that contains the URL to the CRL proxy;
- **useProxyOnMobile** is an optional field. If `true`, clients should use the CRL proxy. If missing, null or false, clients should not use the CRL proxy.

### Testing
E2EI certificate revocation if CRL proxy is enabled.

### TODO
- not sure about the base branch cc @netbe 

---

### Checklist


wire-ios-sync-engine/Source/E2EI/CRL/CertificateRevocationListsChecker.swift
```diff
@@ -32,6 +32,7 @@ public class CertificateRevocationListsChecker: CertificateRevocationListsChecki
     private let crlAPI: CertificateRevocationListAPIProtocol
     private let mlsConversationsVerificationUpdater: MLSConversationVerificationStatusUpdating
     private let selfClientCertificateProvider: SelfClientCertificateProviderProtocol
+    private let featureRepository: FeatureRepositoryInterface
```
johnxnguyen: It seems like you only need the e2ei feature config. Could you just inject that instead of the entire feature repository?

wire-ios-sync-engine/Source/E2EI/CRL/ProxyCRLProvider.swift
```diff
@@ -0,0 +1,54 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// Depending on the `mlsE2EId` feature flag configuration, we should use a CRL proxy or fetching the CRL directly.
+/// https://wearezeta.atlassian.net/wiki/spaces/PAD/pages/1147666542/2024-04-22+CRL+proxy+for+mobile+apps
+struct ProxyCRLProvider {
```
johnxnguyen: From the outside, it  seems like this object will always provide a proxy URL because of the name `ProxyCRLProvider`, but actually it may or may not depending on private flag.

I would suggest one of two things:
1. Rename the struct to `CRLURLProvider` or something like that
2. Build the crl url in a private method in the `CertificateRevocationListsChecker`, since it appear that is the only place that needs it, and delete this struct.

David-Henner (commented):
Same comments as John

wire-ios-sync-engine/Source/E2EI/CRL/ProxyCRLProvider.swift
KaterinaWire: @johnxnguyen @David-Henner 
1. I agree with the name suggestion, it would be better to make it clearer.

2. I'd like to encapsulate the logic in a structure to make it easier to test. The URL builder is not that simple (in my opinion), I would like to have tests for it. If you check the `CertificateRevocationListsCheckerTests`, it contains many tests regarding the functionality of the CertificateRevocationListsChecker. CRL proxy tests may get lost there.

KaterinaWire: Renamed to `CRLURLBuilder`

wire-ios-sync-engine/Source/E2EI/CRL/CertificateRevocationListsChecker.swift
KaterinaWire: yes, changed

johnxnguyen (approved):

David-Henner (approved):

wire-ios-sync-engine/Source/E2EI/CRL/CertificateRevocationListsChecker.swift
```diff
@@ -32,6 +32,7 @@ public class CertificateRevocationListsChecker: CertificateRevocationListsChecki
     private let crlAPI: CertificateRevocationListAPIProtocol
     private let mlsConversationsVerificationUpdater: MLSConversationVerificationStatusUpdating
     private let selfClientCertificateProvider: SelfClientCertificateProviderProtocol
+    private let fetchE2eiFeatureConfig: (() -> Feature.E2EI.Config?)
```
netbe:
```suggestion
    private let fetchE2EIFeatureConfig: (() -> Feature.E2EI.Config?)
```

netbe (approved):

PR #1615, opened 2024-06-24
refactor: create UpdateEventsRepository - WPB-8742

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8742" title="WPB-8742" target="_blank">TaskWPB-8742</a>  Create UpdateEventRepository
  </td></table>
  <br />
 

### Issue

This PR introduces the `UpdateEventsRepository`, which is responsible for retrieving, decrypting, and storing pending update events from the backend. It's intended use will be to:
- retrieve events from the notification service extension when a push notification is received, 
- retrieve events during quick sync in the main app.

#### The changes in this PR include:

Add new ZMEventModel database version:
- Introduce a new entity `StoredEventEnvelope` which will contain the encoded `UpdateEventEnvelope` for later processing
- Eventually we can delete the existing `StoredUpdateEvent` entity.

Make `UpdateEventEnvelope` and `UpdateEvent` `Codable`:
- This allows us to encode and decode to `Data` so we can store it in the event database.
- Move the custom `Decodable` implementation (from backend JSON payloads)  of `UpdateEvent` into `UpdateEventDecodingProxy`, this frees `UpdateEvent` to have the generated `Codable` conformance necessary for the previous point.

Introduce `UpdateEventDecryptor`:
- It receives update event envelopes
- Inspects the event types, forwards to `ProteusMessageDecryptor` for decryption
- Repackages the event with the decrypted content and returns it
- In case of decryption errors, append a system message

Introduce `ProteusMessageDecryptor`:
- It decrypts proteus messages using Core Crypto

### Note
Things that will come in followup PRs:
- Decrypting MLS messages
- Encryption at Rest of stored event envelopes.

### Testing

In its current state we can't real-world test this. To do so we need to integrate into the existing sync mechanism or a new one.

---

### Checklist


WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Event decoding/Conversation/ConversationProteusMessageAddEventDecoder.swift
```diff
@@ -47,19 +47,19 @@ struct ConversationProteusMessageAddEventDecoder {
...
+            messageRecipientClientID: payload.recipient
         )
     }
 
     private struct Payload: Decodable {
 
         let text: String
         let data: String?
-        let sender: UUID
-        let recipient: UUID
```
johnxnguyen, line 62: This was an error, these are client ids not user ids.

wire-ios-data-model/Source/Proteus/ProteusError.swift
```diff
@@ -25,32 +25,32 @@ public enum ProteusError: Int, Error, Equatable {
     /// An error occurred while loading or peristing key material. The
     /// operation may be retried a limited number of times.
 
-    case storageError
+    case storageError = 1
```
johnxnguyen, line 28: These error codes should correspond to error codes defined within the proteus implementation, but they're all wrong. Many were off by one, others were completely wrong because won't don't list every single error (not sure why). Here I've corrected the errors that we seem to care about.

WireDomain/Sources/WireDomain/Event Decryption/ProteusMessageDecryptor.swift
```diff
@@ -0,0 +1,152 @@
...
+                domain: eventData.senderID.domain,
+                in: managedObjectContext
+            )
+
+            guard let senderClient = UserClient.fetchUserClient(
+                withRemoteId: eventData.messageSenderClientID,
+                forUser: senderUser,
+                createIfNeeded: true
+            ) else {
+                throw ProteusMessageDecryptorError.selfClientNotFound
```
typfel, line 136:
```suggestion
                throw ProteusMessageDecryptorError.senderClientNotFound
```

WireAPI/Sources/WireAPI/Models/Connection/Connection.swift
```diff
@@ -18,7 +18,7 @@
 
 import Foundation
 
-public struct Connection: Equatable {
+public struct Connection: Equatable, Codable {
```
markusfassbender, line 21: I thought that the public objects should only transfer data between layers and not be `Codable`. Adding or modifying properties can lead to unexpected failure on different sides.

If some layer write those objects to disk, for example, it should use it's own internal type.

WireDomain/Sources/WireDomain/Event Decryption/ProteusMessageDecryptor.swift
```diff
@@ -0,0 +1,152 @@
...
+    ) async throws -> ConversationProteusMessageAddEvent
+
+}
+
+struct ProteusMessageDecryptor: ProteusMessageDecryptorProtocol {
+
+    let proteusService: any ProteusServiceInterface
+    let managedObjectContext: NSManagedObjectContext
+
+    private let maxCiphertextSize = Int(12_000 * 1.5)
```
markusfassbender, line 43: 💅 typo?
```suggestion
    private let maxCipherTextSize = Int(12_000 * 1.5)
```

WireDomain/Sources/WireDomain/Event Decryption/ProteusMessageDecryptorError.swift
```diff
@@ -0,0 +1,28 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
```
markusfassbender: @caldrian would say the import is not needed 😅

WireDomain/Sources/WireDomain/Repositories/UpdateEventsRepository.swift
```diff
@@ -0,0 +1,122 @@
...
+
+final class UpdateEventsRepository: UpdateEventsRepositoryProtocol {
+
+    private let selfClientID: String
+    private let updateEventsAPI: any UpdateEventsAPI
+    private let updateEventDecryptor: any UpdateEventDecryptorProtocol
+    private let eventContext: NSManagedObjectContext
+    private let jsonEncoder = JSONEncoder()
+
+    private let lastEventIDRepository: any LastEventIDRepositoryInterface
```
markusfassbender, line 44: Great to see more places using `any` as Swift 6 preperation.

WireDomain/Sources/WireDomain/Repositories/UpdateEventsRepository.swift
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+    }
+
+    private func indexOfLastEventEnvelope() async throws -> Int64 {
+        try await eventContext.perform { [eventContext] in
+            let request = StoredUpdateEventEnvelope.sortedFetchRequest(asending: false)
+            request.fetchBatchSize = 1
+            let lastEnvelope = try eventContext.fetch(request).first
+            return lastEnvelope?.sortIndex ?? 0
```
markusfassbender, line 100: Is it right to return `0` instead of an error?

WireAPI/Sources/WireAPI/Models/Connection/Connection.swift
johnxnguyen: Yeah makes sense, but I'm not sure we can add automatic codable conformance from an extension on the consumer side. Either we add manual conformance (would be painful) or we add mirror models (also painful) that are Codable. Or perhaps there's another solution I'm not seeing right now.

WireDomain/Sources/WireDomain/Event Decryption/ProteusMessageDecryptor.swift
johnxnguyen: [Ciphertext](https://en.wikipedia.org/wiki/Ciphertext) and Plaintext are both single words, hence I didn't capitalize mid-word.

WireDomain/Sources/WireDomain/Event Decryption/ProteusMessageDecryptorError.swift
johnxnguyen: Is it always implicit? It's always there when I create a file so I just leave it. 🤷🏻‍♂️

WireAPI/Sources/WireAPI/Models/Connection/Connection.swift
markusfassbender: My personal way would be to add mirror models and be explicit. But I understand that the simpler way is fine in many cases. It's not a blocker as long as the decision is taken conciously :)

WireDomain/Sources/WireDomain/Repositories/UpdateEventsRepository.swift
johnxnguyen: if there is no last envelope, then it's not an error, we just start with a zero index. The important thing isn't where the index starts, but that it increases monotonically.

WireAPI/Sources/WireAPI/Models/Connection/Connection.swift
johnxnguyen: I'm leaning toward mirror models too, but I think I'll address this in a separate PR in order to avoid bloating this one even more.

johnxnguyen: I create a ticket for this: https://wearezeta.atlassian.net/browse/WPB-9989

WireDomain/Sources/WireDomain/Event Decryption/ProteusMessageDecryptorError.swift
johnxnguyen: removed 31a6fbe7096f2af8ceb7eef257ed1ff75fccaea9

WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Event decoding/UpdateEventDecodingProxy.swift
```diff
@@ -18,7 +18,27 @@
 
 import Foundation
 
-extension UpdateEvent: Decodable {
+/// A wrapper that provides custom decoding of update events from
+/// JSON payloads received from the backed.
```
David-Henner:
```suggestion
/// JSON payloads received from the backend.
```

WireDomain/Sources/WireDomain/Event Decryption/ProteusMessageDecryptorError.swift
markusfassbender: Thanks and sorry for the late reply. Foundation is just needed for types (like UUID) that are not included in Swift itself (String, etc.).

markusfassbender (approved):

David-Henner (approved):

PR #1631, opened 2024-06-28
fix: cancel all queued processing tasks when processing throws WPB-9221

typfel:
<table>
<td>
  </td></table>
  <br />
 

### Issue

If we we call 

```
EventProcessor.processEvent([event1, event2])
EventProcessor.processEvent([event3, event4])
```

and we run out of processing time while processing `[event1, event2]` we would immediately continue with processing the second batch (`[event3, event4]`).

This was happening because we were swallowing any error thrown while waiting for the previous task to finish.

---

### Checklist

### DIFF
--- a/wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
+++ b/wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
@@ -91,8 +91,10 @@ actor EventProcessor: UpdateEventProcessor {
     }
 
     private func enqueueTask(_ block: @escaping @Sendable () async throws -> Void) async throws {
+        defer { processingTask = nil }
+
         processingTask = Task { [processingTask] in
-            _ = await processingTask?.result
+            _ = try await processingTask?.value
             return try await block()
         }
 


wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
```diff
@@ -91,8 +91,10 @@ actor EventProcessor: UpdateEventProcessor {
     }
 
     private func enqueueTask(_ block: @escaping @Sendable () async throws -> Void) async throws {
+        defer { processingTask = nil }
+
         processingTask = Task { [processingTask] in
-            _ = await processingTask?.result
+            _ = try await processingTask?.value
```
typfel, line 97: awaiting the value instead of the result will make it throw on any error.

wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
```diff
@@ -91,8 +91,10 @@ actor EventProcessor: UpdateEventProcessor {
     }
 
     private func enqueueTask(_ block: @escaping @Sendable () async throws -> Void) async throws {
+        defer { processingTask = nil }
```
typfel, line 94: We need to nil out the processing task otherwise we'll keep re-throwing the first which occurs.

KaterinaWire (approved):

netbe (commented):
can we add a test to verify it? or is there manual steps to repoduce this?

johnxnguyen (approved):

typfel:
> can we add a test to verify it? or is there manual steps to repoduce this?

We reproduced it by creating a sample project. We don't have any simple steps to reproduce it in the app. I'll see if it's feasible to write a test for it even though `EventProcessor` isn't very testable atm.

PR #1347, opened 2024-04-25
feat: configurable ciphersuite WPB-8591

typfel:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8591" title="WPB-8591" target="_blank">StoryWPB-8591</a>  [iOS] Add support for ECDSA ciphersuites
  </td></table>
  <br />
 

### Issue

Previously we hardcoded the ciphersuite we use for MLS to `MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519` now we allow it be configurable via the MLS feature config.

### Testing

If you change the default ciphersuite in the team setting this should now be respected and reflected everywhere.

Under the hood this means we now need to agnostic about the ciphersuite when:
- Initialising an new MLS client
- Creating new key packages
- Creating an MLS conversation
- Joining an MLS conversation via external commit
- Displaying which ciphersuite an MLS conversation uses on the conversation details screen
- Displaying an MLS thumbprint on the device details screen

---

### Checklist

- [ ] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [ ] Description is filled and free of optional paragraphs.
- [ ] Adds/updates automated tests.

PR #1636, opened 2024-07-01
fix: expiration timer fails to send messages - WPB-9517

markusfassbender:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9517" title="WPB-9517" target="_blank">BugWPB-9517</a>  [iOS] Assets frequently fail to send
  </td></table>
  <br />
 

### Issue

The simple description would be that an expiration timer for messages in the interface starts, that is not synced with networking. On sending delays the user already sees "retry"  button but can't send messages, for example when a quick sync is still running in parallel.

Solution is that the expiration date is calculated and timer is started before the networking begins and not earlier when the app was waiting for quick sync.

Also I renamed the core data model versions in code, because it was so difficult to read. In the past we had also version 1, but now only 2.x is available in the app.

### Testing

My own testing:
1. the expiration time is 15 seconds, the quick sync takes 5 and network 12 seconds. messages got sent ✅
2. the expiration time is 15 seconds, the quick sync takes 16 and network 1 seconds. messages got sent ✅
3. expiration time is 15 seconds, the quick sync takes 0 and network 16 seconds. messages didn't get sent ✅
4. tested retry button where the expiration time is 15 seconds and network took randomly between 10-20 seconds. depending on the random value messages got sometimes send and sometimes retry was needed ✅

Developers modification to test:
- `ZMDefaultMessageExpirationTime`: expiration time interval
- `QuickSyncObserver.waitForQuickSyncToFinish()`: add sleep to fake a delay
- `MessageAPIV1.sendProteusMessage(..)`: : add sleep to fake a delay

Group Testing
‼️ We should do regression testing with bad network conditions

---

### Checklist


wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
```diff
@@ -112,17 +83,27 @@ public class MessageSender: MessageSenderInterface {
 
     public func sendMessage(message: any SendableMessage) async throws {
         await context.perform {
-            WireLogger.messaging.debug("send message", attributes: message.logInformation)
+            WireLogger.messaging.debug(
+                "send message - start wait for quick sync to finish",
+                attributes: .safePublic
+            )
```
netbe: you should add message.logInformation as then we can filter for a given message

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMessagingMigrationVersion.swift
```diff
@@ -29,92 +29,94 @@ enum CoreDataMessagingMigrationVersion: String, CaseIterable {
...
+    case v84 = "zmessaging2.84.0"
+    case v83 = "zmessaging2.83.0"
+    case v82 = "zmessaging2.82.0"
+    case v81 = "zmessaging2.81.0"
+    case v80 = "zmessaging2.80.0"
 
     var nextVersion: Self? {
         switch self {
-        case .version2_116,
-                .version2_117:
```
markusfassbender: This was a bug (introduced [here](https://github.com/wireapp/wire-ios/pull/1347/files#diff-34674e147f7b2974213af0e23acb90b04a3ac3ed482fd2f9d316885d8be4fb09)) returning `nil` and not doing the migration from 116 to 117.
It got fixed now both versions migrating directly to 118 (automatic light weight):
```swift
case .v116, .v117:
    return .v118
```

wire-ios-request-strategy/Sources/Object Syncs/Helpers/MLSMessage.swift
```diff
@@ -20,14 +20,31 @@ import Foundation
...
+
     public func encryptForTransport(using encrypt: (Data) async throws -> Data) async throws -> Data {
         try await message.encryptForTransport(using: encrypt)
     }
+
+    public func setExpirationDate() {
+        // Just required for protocol conformance.
+        // Generic messages are used as underlying messages in proteus and mls,
+        // so they don't need to have their own expiration date.
+    }
```
markusfassbender, line 49: At least this is @johnxnguyen 's and my assumption... please add knowledge if you know more :)

wire-ios-request-strategy/Sources/Object Syncs/Helpers/ProteusMessage.swift
```diff
@@ -18,7 +18,14 @@
 
 import Foundation
 
-public protocol ProteusMessage: OTREntity, EncryptedPayloadGenerator {}
+public protocol ProteusMessage: OTREntity, EncryptedPayloadGenerator {
+
+    /// Messages can expire, e.g. if network conditions are too slow to send.
+    var shouldExpire: Bool { get }
+
+    /// Sets the expiration date with the default time interval and returns the date.
```
samwyndham:
```suggestion
    /// Sets the expiration date with the default time interval.
```

wire-ios-request-strategy/Sources/Object Syncs/Helpers/MLSMessage.swift
```diff
@@ -20,14 +20,31 @@ import Foundation
 
 /// A message that can be sent in an mls group.
 
-public protocol MLSMessage: OTREntity, MLSEncryptedPayloadGenerator {}
+public protocol MLSMessage: OTREntity, MLSEncryptedPayloadGenerator {
+
+    /// Messages can expire, e.g. if network conditions are too slow to send.
+    var shouldExpire: Bool { get }
+
+    /// Sets the expiration date with the default time interval and returns the date.
```
samwyndham:
```suggestion
    /// Sets the expiration date with the default time interval.
```

The current comment is a bit confusing because `setExpirationDate()` doesn't return a date.

wire-ios-data-model/Source/Model/Message/ZMMessage+Internal.h
```diff
@@ -84,6 +84,9 @@ extern NSString * _Nonnull const ZMMessageNeedsLinkAttachmentsUpdateKey;
 
 @property (nonatomic) BOOL delivered;
 
+/// Sets the expiration date with the default time interval and returns the date.
```
samwyndham:
```suggestion
/// Sets the expiration date with the default time interval.
```

samwyndham (commented):
Just commenting as I don't know enough about the code base yet to approve but looks good to me. It took me some time to understand the fix, which is basically not setting it on `ZMMessage.resend` but waiting until actual send time right?

wire-ios-request-strategy/Sources/Object Syncs/Helpers/ProteusMessage.swift
markusfassbender: thanks, done with 30f9c41fbddf02f59eaf6825dd46716aada528d2

markusfassbender:
Thanks Sam, the logic here is quite complex and only thanks to @johnxnguyen and @netbe I was able to work on it.
Basically first the message objects are set in CoreData and saved by `MessageSender` now. Kind of "magically" another object, here `MessageExpirationTimer` receives the changes from CoreData and starts a timer.

The fix is to start the timer only in the moment the network also starts running. Before it was started too early and other side effects could interfere.

If there are questions, I'm happy to pair with you and go through the relevant parts.

samwyndham:
> If there are questions, I'm happy to pair with you and go through the relevant parts.

@markusfassbender No that's cool. Thanks for the explanation!

johnxnguyen (approved):

wire-ios-data-model/Tests/Source/Model/Conversation/ZMConversationTests+Messages.swift
```diff
@@ -445,7 +446,7 @@ class ZMConversationMessagesTests: ZMConversationTestsBase {
         // then
         XCTAssertNil(message.underlyingMessage?.imageAssetData)
         XCTAssertNotNil(message.underlyingMessage?.locationData)
-        XCTAssertNotNil(message.expirationDate)
+        XCTAssert(message.shouldExpire)
```
netbe:
```suggestion
        XCTAssertTrue(message.shouldExpire)
```

wire-ios-data-model/Tests/Source/Model/Conversation/ZMConversationTests.m
```diff
@@ -629,9 +629,7 @@ - (void)testThatItSetsTheExpirationDateOnATextMessage
     ZMMessage *message = (id)[sut appendMessageWithText:@"Quux"];
 
     // then
-    XCTAssertNotNil(message.expirationDate);
-    NSDate *expectedDate = [NSDate dateWithTimeIntervalSinceNow:[ZMMessage defaultExpirationTime]];
-    XCTAssertLessThan(fabs([message.expirationDate timeIntervalSinceDate:expectedDate]), 1);
+    XCTAssert(message.shouldExpire);
```
netbe:
```suggestion
    XCTAssertTrue(message.shouldExpire);
```

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Message.swift
```diff
@@ -384,11 +384,7 @@ extension ZMConversation {
         }
 
         message.sender = ZMUser.selfUser(in: moc)
-
-        if expires {
-            let expirationDate = message.setExpirationDate()
-            WireLogger.messaging.debug("set expiration date \(expirationDate)", attributes: logAttributes)
-        }
+        message.shouldExpire = expires
```
netbe, line 387: @markusfassbender should we toggle shouldExpire again once it's sent ?
What's the default value of shouldExpire?

netbe (commented):
looks good, added a question in order to understand it

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Message.swift
markusfassbender: Good points:
- the default value is `false` and defined this way (and non-optional) in the CoreData model.
- in my opinion it is not required to reset the `shouldExpire` as it is just a flag to let the other piece of code know when a timer should be started. @johnxnguyen what do you think?

johnxnguyen: I agree with @markusfassbender, once it's sent it becomes irrelevant.

netbe (approved):

PR #1647, opened 2024-07-03
fix: can't see screen share - WPB-9909

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9909" title="WPB-9909" target="_blank">BugWPB-9909</a>  Cannot see screen share
  </td></table>
  <br />
 

### Issue

While in a call, if another person shares their screen (without having first enabled their video) then the self user would not see the screen share.

This is due to another fix (https://wearezeta.atlassian.net/browse/WPB-8954) that only created the video view (used for both camera and screen share) for another participant when they enabled their camera. This was to prevent a crash on iOS 17 simulators.

The fix is to always create the video view. 

In my testing, I was not able to reproduce any crash on iOS 17 simulators. I did observe though that when answering a call on simulator, the call is immediately dismissed the first time, but succeeds the second time. Also, if another user enables video or screen share, then this is never visible on the simulator, even if we always create the video view.

### Testing

1. Be in a group call.
2. Other participants enable screen share.
3. Assert eh self user can see the screen share.

---

### Checklist

I'm not sure how to add a unit test for this, if you have any ideas let me know.

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallParticipantViews/CallParticipantView.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallParticipantViews/CallParticipantView.swift
@@ -122,6 +122,10 @@ final class CallParticipantView: BaseCallParticipantView {
         self.videoContainerView?.removeFromSuperview()
         self.videoContainerView = videoContainerView
 
+        let videoView = makeVideoView()
+        self.videoView = videoView
+        videoContainerView.setupVideoView(videoView)
+
         // Adding the preview into a container allows smoother scaling
         let scalableView = ScalableView(isScalingEnabled: shouldEnableScaling)
         scalableView.addSubview(videoContainerView)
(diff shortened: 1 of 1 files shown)


wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallParticipantViews/CallParticipantView.swift
```diff
@@ -162,15 +166,6 @@ final class CallParticipantView: BaseCallParticipantView {
     // MARK: Override Base
 
     override func updateVideoShouldFill(_ shouldFill: Bool) {
-        if shouldFill, videoView == nil {
-            // [WPB-8954] Setup video only when the video really starts to avoid
-            // calls crashing on the iOS 17 simulator.
```
johnxnguyen, line 167: I was a bit confused by this condition... `shouldFill` is `true` when the user enables camera, but not screen share. I would expect the condition to check the participant state to see if they have either camera or screen share enabled.

Also... I wasn't able to reproduce the issue described, maybe my testing wasn't rigorous enough. Do you think we should still avoid setting up the video view for simulators? In that case, couldn't we do something like:

```swift
#if targetEnvironment(simulator)
  // don't create video view
#else
  // create video view
#endif
```

netbe:
@johnxnguyen is a potential fix could be : create the video view when the other user share it's screen ?

markusfassbender (approved):

KaterinaWire (approved):

PR #1651, opened 2024-07-03
fix: name status update - WPB-9993

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9993" title="WPB-9993" target="_blank">BugWPB-9993</a>  [iOS] Status not synced across app after update from another client
  </td></table>
  <br />
 

### Issue

When user change its status on another client and the iOS app is in background, the new status will not get updated in the topbar.

The cause is that we don't detect it as a change in the modifiedKeys. 

### Testing

1.Be logged into two clients one iOS and one other platform
2.put iOS in background
3.Update name or status in other client
4.open iOS app in foreground

Expected

Status is changed in both the conversation list view and settings

---

### Checklist


agisilaos (approved):
Looking good. 🙌 Can we somehow have tests for this? To make sure it doesn't happen again in the future?

wire-ios-request-strategy/Sources/Request Strategies/Availability/AvailabilityRequestStrategy.swift
```diff
@@ -72,8 +76,11 @@ extension AvailabilityRequestStrategy: ZMEventConsumer {
                 continue
             }
 
-            let user = ZMUser.fetch(with: senderUUID, in: context)
-            user?.updateAvailability(from: message)
+            guard let user = ZMUser.fetch(with: senderUUID, in: context) else {
+                assertionFailure("User not found to updateAvailability")
+                continue
+            }
+            user.updateAvailability(from: message)
```
markusfassbender, line 83: What happens here in the last line, if the user is `nil` and `continue` runs? 🤔

wire-ios-data-model/Source/Notifications/ObjectObserverTokens/UserChangeInfo.swift
```diff
@@ -132,7 +132,7 @@ extension ZMUser: ObjectInSnapshot {
     }
 
     public var availabilityChanged: Bool {
-        return changedKeys.contains(#keyPath(ZMUser.availability))
+        return changedKeysContain(keys: #keyPath(ZMUser.availability))
```
samwyndham, line 135: I assume this works because `ObjectChangeInfo.considerAllKeysChanged` is `true` when returning from the background?

wire-ios-request-strategy/Sources/Request Strategies/Availability/AvailabilityRequestStrategy.swift
markusfassbender: All good, the loop continues without going to line 83.

markusfassbender (approved):

wire-ios-data-model/Source/Model/User/Availability/Availability.swift
```diff
@@ -20,3 +20,18 @@
 public enum Availability: Int, CaseIterable {
     case none, available, busy, away
 }
+
+extension Availability: CustomStringConvertible {
+    public var description: String {
```
samwyndham, line 25: What is this conformance for? Don't we get this behavior (printing a representative value) free these days in many places? Or is that only within same target maybe?

samwyndham (approved):
I've confirmed the fix works on device

wire-ios-data-model/Source/Model/User/Availability/Availability.swift
samwyndham: Please ignore this comment. I now see behavior is different when called from a different target.

wire-ios-data-model/Source/Notifications/ObjectObserverTokens/UserChangeInfo.swift
netbe: exactly

PR #1652, opened 2024-07-03
feat: format code in WireAnalytics, WireDomain and WireUI - WPB-6234

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6234" title="WPB-6234" target="_blank">TaskWPB-6234</a>  Use a formatter to fix formatting issues on all swift files including SwiftUI files
  </td></table>
  <br />
 

### Issue

This PR enables SwiftFormat listing and formatting for the directories
- WireAnalytics
- WireDomain
- WireUI
- SourceryPlugin

---

### Checklist


WireDomain/Sources/WireDomain/Repositories/TeamRepository.swift
```diff
@@ -101,7 +101,7 @@ final class TeamRepository: TeamRepositoryProtocol {
                 domain: nil,
                 in: context
             )
-            team.pictureAssetId = teamAPIModel.logoID
+            team.pictureAssetID = teamAPIModel.logoID
```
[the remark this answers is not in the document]
johnxnguyen: Looks like you'll need to rename this from data model.

johnxnguyen (approved):

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor.swift
```diff
@@ -36,7 +36,7 @@ protocol ConversationEventProcessorProtocol {
 
 struct ConversationEventProcessor {
 
-    func processEvent(_ event: ConversationEvent) async throws {
+    func processEvent(_: ConversationEvent) async throws {
```
markusfassbender, line 39: I forgot that this is valid Swift syntax 😁

WireDomain/Sources/WireDomain/Event Processing/UpdateEventProcessor.swift
```diff
@@ -44,22 +44,22 @@ struct UpdateEventProcessor {
 
     func processEvent(_ event: UpdateEvent) async throws {
         switch event {
-        case let .conversation(event):
+        case .conversation(let event):
```
markusfassbender, line 47: ❤️

WireDomain/Sources/WireDomain/Repositories/TeamRepository.swift
```diff
@@ -236,9 +236,9 @@ final class TeamRepository: TeamRepositoryProtocol {
 
 }
 
-private extension ConversationAction {
+extension ConversationAction {
 
-    var name: String {
+    fileprivate var name: String {
```
markusfassbender: Hmm when I write code I always prefer this `private extension`, because it means automatically nothing is available outside this file which makes it easier to refactor etc.
Now with the scope next to the property, one could just add a property `public var displayName: String`, but then I would move extension to a new file to have it clear.

Is it a special case here or shall we modify the rule?

markusfassbender (approved):

WireDomain/Sources/WireDomain/Repositories/TeamRepository.swift
caldrian: I restored the default behaviour of this rule: https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#extensionAccessControl

WireDomain/Sources/WireDomain/Repositories/TeamRepository.swift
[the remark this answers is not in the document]
caldrian: I disabled the acronyms rule, it's hard and risky to enable it on an existing codebase.

PR #1656, opened 2024-07-04
fix: Don't try to share non-existing logs - WPB-9948

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9948" title="WPB-9948" target="_blank">BugWPB-9948</a>  Sending logs - when email client is not configured
  </td></table>
  <br />
 

### Issue

Currently when sending logs via a share sheet we attempt to attach all logging URLs derived from `LogFileDestination` and `ZMSLog`. However in some cases no logs have actually been created, which unsurprisingly can cause sending logs to fail.

This PR:

- Updates `DebugLogSender.debugLogs` to filter out urls for files that do not exist. It also renames it to `existingDebugLogs` to emphasize this and makes it a function for testing purposes.
- Makes `cachesDirectory` non optional which simplifies a bunch of dependent code as previously discussed. My assumption here is that `FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask)[0]` will always succeed. This is based on my experience and Apples implementation of [URL.cachesDirectory](https://github.com/apple/swift-foundation/blob/046bb36763c420545f8a14f77bb1325954394f8d/Sources/FoundationEssentials/URL/URL.swift#L2491).
- Adds a test to ensure `existingDebugLogs()` filters out non existing urls. I don't like this test very much but it is hard to improve without more significant changes.

### Testing

0. Clean install app, don't have Mail.app configured, have Gmail installed.
1. Login.
2. Go to Account > Settings > Advanced > Debug Report.
3. Tap "Send report to Wire".
4. Tap "OK" on alert informing you that no mail clients are set up.
5. Select Gmail.
6. Gmail compose email sheet appears with bug reports attached.

### Checklist

- [X] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [X] Description is filled and free of optional paragraphs.
- [X] Adds/updates automated tests.

PR #1658, opened 2024-07-04
feat: share logs through Wire - WPB-9135

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9135" title="WPB-9135" target="_blank">StoryWPB-9135</a>  Share logs through Wire
  </td></table>
  <br />
 

This PR introduces a new feature: **log sharing through Wire**

Users now have the option to share logs via their mail provider, or via Wire directly, by choosing which conversations they wish to send their log files to.

### Changes

I removed the `SettingsTechnicalReportViewController` and replaced it with a new module. It uses a **MVVM + Router** pattern. 

I used the existing `ShareViewController` to present a list of conversations to choose from to share the logs. I also created a `ShareableDebugReportView` to represent the log file to be shared. It only displays the file but could be extended further if we ever want to include text.

I added two use cases:
- `FetchShareableConversationsUseCase` 
- `ShareFileUseCase`

I use them in the view model to fetch conversations and share the log files. One of the advantages of using use cases here is that it helps with writing unit tests for the view model. The use cases are also straightforward to test.

I extracted some of the log file generation logic out of the `MFMailComposeViewController+Logs` extension into a new struct `LogFilesProvider`. And added a functionality to generate a URL to the log files archive.

I updated `FileMetaDataGenerator.metadataForFileAtURL` from a static func to a non-static func in order to facilitate mocking and dependency injection. Some class extensions were using the static func: since class extensions cannot have stored properties, I introduced a `shared` property in `FileMetaDataGenerator` to allow the extensions to use it as a singleton. I immediately deprecated the property to also ensure we instantiate the class whenever possible.

### Tests

Added unit / snapshot tests for:
- `SettingsDebugReportViewModel`
- `SettingsDebugReportViewController`
- `FetchShareableConversationsUseCase` 
- `ShareFileUseCase`

### Testing

1. Go to `Settings` > `Advanced` > `Debug Report`
2. Tap `Send Report` button and verify that you're able to send logs via email (or share sheet if email isn't configured)
3. Tap `Share Report Via Wire` button, select the conversation(s) you want to share the logs to, tap send. Verify the logs were sent

### Checklist


wire-ios/Wire-iOS/Sources/UserInterface/Settings/MFMailComposeViewController+Logs.swift
```diff
@@ -60,22 +60,12 @@ extension MFMailComposeViewController {
...
-
-        if let data = FileManager.default.zipData(from: logFiles) {
+    func attachLogs() {
+        do {
+            let data = try LogFilesProvider().generateLogFilesData()
             addAttachmentData(data, mimeType: "application/zip", fileName: "logs.zip")
-        } else {
-            WireLogger.system.debug("no logs for WireLogger to send \(logFiles.description)")
+        } catch {
+            WireLogger.system.debug("no logs for WireLogger to send")
```
samwyndham: Maybe it is also worth logging the error here as it might fail for an unexpected reason.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/LogFilesProvider.swift
```diff
@@ -0,0 +1,103 @@
...
+    func generateLogFilesData() throws -> Data {
+        defer {
+            // because we don't rotate file for this one, we clean it once sent
+            // this regenerated from os_log anyway
+            if let url = LogFileDestination.main.log {
+                try? FileManager.default.removeItem(at: url)
+            }
+        }
+
+        let urls = logFilesURLs
```
samwyndham: Sanity check. Will there always be files on disk at these URLs. My fix [here](https://github.com/wireapp/wire-ios/pull/1656) related to files not being there.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+
+    private typealias Strings = L10n.Localizable.Self.Settings
+
+    // MARK: - Properties
+
+    private let viewModel: SettingsDebugReportViewModelProtocol
+
+    // MARK: - Views
+
+    private let infoLabel: UILabel = {
```
samwyndham: I think it is good practice to make this lazy to be a good citizen in UIViewController lazy instantiation.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import MessageUI
+import UIKit
+import WireDataModel
+import WireDesign
+
+class SettingsDebugReportViewController: UIViewController {
```
samwyndham, line 24: Just out of interest why are these new views not SwiftUI?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewModel.swift
```diff
@@ -0,0 +1,102 @@
...
+
+        let logsURL = logsProvider.generateLogFilesZip()
+
+        fileMetaDataGenerator.metadataForFileAtURL(
+            logsURL,
+            UTI: logsURL.UTI(),
+            name: logsURL.lastPathComponent
+        ) { [weak self] metadata in
+
+            guard let `self` else { return }
```
samwyndham: Not sure you need the `` ` `` around the self here. Is it the style in the codebase?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReport.swift
```diff
@@ -0,0 +1,72 @@
...
+        logFileMetadata: ZMFileMetadata,
+        userSession: UserSession
+    ) {
+        self.logFileMetadata = logFileMetadata
+        self.userSession = userSession
+    }
+
+    // MARK: - Interface
+
+    func share<I>(to: [I]) {
```
samwyndham, line 46: Out of interest why do you use `I` instead of `ZMConversation` directly? especially considering you use `ZMConversation` within the implementation. I believe the compiler will be happy with Shareable implementation even if delete all the `I`'s from this file.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Advanced.swift
```diff
@@ -43,11 +43,26 @@ extension SettingsCellDescriptorFactory {
...
     // MARK: - Sections
-    private var troubleshootingSection: SettingsSectionDescriptor {
+    private func troubleshootingSection(userSession: UserSession) -> SettingsSectionDescriptor {
         let submitDebugButton = SettingsExternalScreenCellDescriptor(
             title: SelfSettingsAdvancedLocale.Troubleshooting.SubmitDebug.title,
             presentationAction: { () -> (UIViewController?) in
-                return SettingsTechnicalReportViewController()
+
+                // TODO: How to avoid this?
+                guard let contextProvider = userSession as? ContextProvider else {
```
samwyndham: Are `ContextProvider` and `UserSession` both protocols. Maybe change 

```swift
private func troubleshootingSection(userSession: UserSession) -> SettingsSectionDescriptor
``` 
to
```swift
private func troubleshootingSection(userSession: UserSession & ContextProvider) -> SettingsSectionDescriptor
```

wire-ios-sync-engine/Tests/Source/Use cases/FetchShareableConversationUseCaseTests.swift
```diff
@@ -0,0 +1,81 @@
...
+    func testInvoke() {
+        // Given
+        let selfUser = ZMUser.selfUser(in: coreDataStack.viewContext)
+
+        // create a shareable conversation
+        let shareable = createConversation()
+        shareable.addParticipantAndUpdateConversationState(user: selfUser)
+
+        // create a non-shareable conversation
+        let nonShareable = createConversation()
```
samwyndham: Do you get a warning here for an unused type?

samwyndham (commented):
Commenting rather than approving as I'm new to the project

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewModel.swift
markusfassbender: No, it's not needed indeed 👍

wire-ios/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Advanced.swift
```diff
@@ -43,11 +43,19 @@ extension SettingsCellDescriptorFactory {
...
-    private var troubleshootingSection: SettingsSectionDescriptor {
+    private func troubleshootingSection(userSession: UserSession) -> SettingsSectionDescriptor {
         let submitDebugButton = SettingsExternalScreenCellDescriptor(
             title: SelfSettingsAdvancedLocale.Troubleshooting.SubmitDebug.title,
             presentationAction: { () -> (UIViewController?) in
-                return SettingsTechnicalReportViewController()
+                let router = SettingsDebugReportRouter()
+                let viewModel = SettingsDebugReportViewModel(
+                    router: router,
+                    shareFile: userSession.shareFileUseCase,
```
markusfassbender: Actually we have discussed recently that we don't want to provide use cases through the user session, as it leads to a uncontrollable amount of dependencies.

You can create the use case right here in place with the dependencies from user session, like:
`ShareFileUseCase(contextProvider: userSession)` or `ShareFileUseCase(contextProvider: userSession.contextProvider)` (not sure what is correct right now).

wire-ios-sync-engine/Source/UserSession/UserSession.swift
```diff
@@ -254,6 +254,10 @@ public protocol UserSession: AnyObject {
 
     var lastE2EIUpdateDateRepository: LastE2EIdentityUpdateDateRepositoryInterface? { get }
 
+    var fetchShareableConversationsUseCase: FetchShareableConversationsUseCaseProtocol { get }
+
+    var shareFileUseCase: ShareFileUseCaseProtocol { get }
```
markusfassbender: As described in the other comment, please let's reduce the user session as provider for everything and instantiate the use cases in the places they are used.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
```diff
@@ -96,7 +96,7 @@ extension ZMMessage: Shareable {
             }
         } else if isVideo || isAudio || isFile {
             guard let url = fileMessageData!.temporaryURLToDecryptedFile() else { return }
-            FileMetaDataGenerator.metadataForFileAtURL(url, UTI: url.UTI(), name: url.lastPathComponent) { fileMetadata in
+            FileMetaDataGenerator.shared.metadataForFileAtURL(url, UTI: url.UTI(), name: url.lastPathComponent) { fileMetadata in
```
markusfassbender, line 99: There are only two places that use the deprecated singleton `FileMetaDataGenerator.shared`. Can we change it to instantiate an object right in place `FileMetaDataGenerator`, as you already did in some changes in this PR? 👍

And FileMetaDataGenerator can become a `struct`.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
```diff
@@ -134,18 +134,6 @@ extension ZMConversationMessage {
     }
 }
 
-// swiftlint:disable:next todo_requires_jira_link
-extension ConversationList { // TODO: mv to DM
```
markusfassbender, line 138: Thanks for moving it to a new file!

wire-ios/Wire-iOS Tests/Settings/Debug Report/SettingsDebugReportViewModelTests.swift
```diff
@@ -0,0 +1,121 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import XCTest
+
+@testable import Wire
+@testable import WireDataModelSupport
+@testable import WireSyncEngineSupport
```
agisilaos: For this test Foundation isn't needed. As far as I understand you don't use any internal members or interfaces of WireDataModelSupport or WireSyncEngine so you don't need to mark them as testable. 

```suggestion
import WireDataModelSupport
import WireSyncEngineSupport
import XCTest

@testable import Wire
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
agisilaos, line 99: Why do we use a singleton here?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import MessageUI
+import UIKit
+import WireDataModel
+import WireDesign
```
agisilaos: I don't think Foundation is needed here. 

```suggestion
import MessageUI
import UIKit
import WireDataModel
import WireDesign
```

wire-ios/Wire-iOS Tests/Settings/Debug Report/SettingsDebugReportViewControllerSnapshotTests.swift
```diff
@@ -22,19 +22,20 @@ import XCTest
...
 @testable import Wire
 
-final class SettingsTechnicalReportViewControllerSnapshotTests: XCTestCase {
+final class SettingsDebugReportViewControllerSnapshotTests: XCTestCase {
 
     // MARK: - Properties
 
     var snapshotHelper: SnapshotHelper!
-    var sut: SettingsTechnicalReportViewController!
+    var sut: SettingsDebugReportViewController!
```
agisilaos: suggestion: Those two properties can be private

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Files.swift
```diff
@@ -86,7 +86,7 @@ extension ConversationInputBarViewController {
             return
         }
 
-        FileMetaDataGenerator.metadataForFileAtURL(
+        FileMetaDataGenerator.shared.metadataForFileAtURL(
```
agisilaos, line 89: Same question regarding the singleton.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/LogFilesProvider.swift
```diff
@@ -0,0 +1,104 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireCommonComponents
+import WireSystem
+import ZipArchive
```
agisilaos: Foundation is not needed here:

```suggestion
import WireCommonComponents
import WireSystem
import ZipArchive
```

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportRouter.swift
```diff
@@ -0,0 +1,124 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import MessageUI
+import WireDataModel
```
agisilaos: Same comment regarding Foundation

```suggestion
import MessageUI
import WireDataModel
```

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewModel.swift
```diff
@@ -0,0 +1,100 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import MessageUI
+import WireCommonComponents
+import WireSyncEngine
```
agisilaos:
```suggestion
import MessageUI
import WireCommonComponents
import WireSyncEngine
```

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReport.swift
```diff
@@ -0,0 +1,72 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireDataModel
+import WireDesign
+import WireSyncEngine
```
agisilaos:
```suggestion
import WireDataModel
import WireDesign
import WireSyncEngine
```

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReportView.swift
```diff
@@ -0,0 +1,127 @@
...
+        topLabel.numberOfLines = 1
+        topLabel.lineBreakMode = .byTruncatingMiddle
+        return topLabel
+    }()
+
+    private let bottomLabel: UILabel = {
+        let bottomLabel = UILabel()
+        bottomLabel.numberOfLines = 1
+        return bottomLabel
+    }()
```
agisilaos: By setting the numbers of lines to 1 we don't allow the text to grow when font size increases. I would recommend setting it to 0 and modify constraints to make sure the text will grow.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReportView.swift
```diff
@@ -0,0 +1,127 @@
...
+
+    private let labelsView = UIView()
+
+    private let documentIconView: UIImageView = {
+        let documentIconView = UIImageView()
+        documentIconView.tintColor = SemanticColors.Icon.backgroundDefault
+        documentIconView.contentMode = .center
+        documentIconView.setTemplateIcon(.document, size: .small)
+        return documentIconView
```
agisilaos: We want to move away from StyleKit. I would recommend checking the Figma file, export the icon directly from there as SVG. And then use it in the Xcode Project.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReportView.swift
```diff
@@ -0,0 +1,127 @@
...
+            labelsView.leftAnchor.constraint(equalTo: documentIconView.rightAnchor, constant: 12),
+            labelsView.rightAnchor.constraint(equalTo: rightAnchor, constant: -12),
+            labelsView.centerYAnchor.constraint(equalTo: centerYAnchor),
+
+            documentIconView.topAnchor.constraint(equalTo: topAnchor, constant: 12),
+            documentIconView.bottomAnchor.constraint(equalTo: bottomAnchor, constant: -12),
+            documentIconView.widthAnchor.constraint(equalToConstant: 32),
+            documentIconView.heightAnchor.constraint(equalToConstant: 32),
+            documentIconView.centerYAnchor.constraint(equalTo: centerYAnchor),
+            documentIconView.leftAnchor.constraint(equalTo: leftAnchor, constant: 12)
```
agisilaos: suggestion: Create an enum to hold all of those numbers, so we can avoid using magic numbers. 

```swift
enum LayoutConstants {
    static let spacing: CGFloat = 2
    static let padding: CGFloat = 12
    static let iconSize: CGFloat = 32
}

NSLayoutConstraint.activate([
    topLabel.topAnchor.constraint(equalTo: labelsView.topAnchor),
    topLabel.leftAnchor.constraint(equalTo: labelsView.leftAnchor),
    topLabel.rightAnchor.constraint(equalTo: labelsView.rightAnchor),

    bottomLabel.topAnchor.constraint(equalTo: topLabel.bottomAnchor, constant: LayoutConstants.spacing),
    bottomLabel.leftAnchor.constraint(equalTo: labelsView.leftAnchor),
    bottomLabel.rightAnchor.constraint(equalTo: labelsView.rightAnchor),
    bottomLabel.bottomAnchor.constraint(equalTo: labelsView.bottomAnchor),

    labelsView.leftAnchor.constraint(equalTo: documentIconView.rightAnchor, constant: LayoutConstants.padding),
    labelsView.rightAnchor.constraint(equalTo: rightAnchor, constant: -LayoutConstants.padding),
    labelsView.centerYAnchor.constraint(equalTo: centerYAnchor),

    documentIconView.topAnchor.constraint(equalTo: topAnchor, constant: LayoutConstants.padding),
    documentIconView.bottomAnchor.constraint(equalTo: bottomAnchor, constant: -LayoutConstants.padding),
    documentIconView.widthAnchor.constraint(equalToConstant: LayoutConstants.iconSize),
    documentIconView.heightAnchor.constraint(equalToConstant: LayoutConstants.iconSize),
    documentIconView.centerYAnchor.constraint(equalTo: centerYAnchor),
    documentIconView.leftAnchor.constraint(equalTo: leftAnchor, constant: LayoutConstants.padding)
])
```

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReportView.swift
```diff
@@ -0,0 +1,127 @@
...
+        let fileSize = ByteCountFormatter.string(
+            fromByteCount: Int64(fileMetadata.size),
+            countStyle: .binary
+        )
+
+        let firstLine = fileMetadata.filename.uppercased()
+        let secondLine = "\(fileSize) \(dot) \(ext)".uppercased()
+
+        topLabel.attributedText = firstLine && UIFont.smallSemiboldFont && SemanticColors.Label.textDefault
+        bottomLabel.attributedText = secondLine && UIFont.smallLightFont && SemanticColors.Label.textCollectionSecondary
```
agisilaos: We want to move away from using Fontspec. I would recommend using UIFont.font(for: `WireTextStyle`)

WireTextStyle = Text styles defined in Wire's design system.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReportView.swift
```diff
@@ -0,0 +1,127 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import UIKit
+import WireDataModel
+import WireDesign
```
agisilaos:
```suggestion
import UIKit
import WireDataModel
import WireDesign
```

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+            shareReportButton.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: 20),
+            shareReportButton.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -20),
+            shareReportButton.heightAnchor.constraint(equalToConstant: 48),
+
+            sendReportButton.bottomAnchor.constraint(equalTo: shareReportButton.topAnchor, constant: -8),
+            sendReportButton.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: 20),
+            sendReportButton.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -20),
+            sendReportButton.heightAnchor.constraint(equalToConstant: 48)
+        ])
+    }
```
agisilaos: Same comment regarding the magic numbers here.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+
+            shareReportButton.bottomAnchor.constraint(equalTo: view.safeBottomAnchor, constant: -30),
+            shareReportButton.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: 20),
+            shareReportButton.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -20),
+            shareReportButton.heightAnchor.constraint(equalToConstant: 48),
+
+            sendReportButton.bottomAnchor.constraint(equalTo: shareReportButton.topAnchor, constant: -8),
+            sendReportButton.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: 20),
+            sendReportButton.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -20),
+            sendReportButton.heightAnchor.constraint(equalToConstant: 48)
```
agisilaos: I would recommend not setting a height for the buttons because that would block from the buttons to expand when the font size grows

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+    @available(*, unavailable)
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        view.backgroundColor = SemanticColors.View.backgroundDefault
+        navigationItem.setDynamicFontLabel(title: Strings.TechnicalReportSection.title.capitalized)
```
agisilaos: Regarding Navigation Bar Titles isn't recommend it to make the text grow. I know that we already do in many places but I'm planning to fix that. Instead there's a feature called [Large Content Viewer](https://www.blog.kevin-hirsch.com/dynamic-type-large-content-viewer/)

Here's how that would work in this case:

```swift
func setupNavigationBarTitle() {
        // Set the navigation item title
        navigationItem.title = "Your Title"

        // Set the navigation bar title color and font
        let titleTextAttributes: [NSAttributedString.Key: Any] = [
            .foregroundColor: SemanticColors.Label.textDefault,
            .font: UIFont.font(for: .h3)
        ]
        navigationController?.navigationBar.titleTextAttributes = titleTextAttributes

        // Ensure title supports LargeContentViewer
        navigationItem.titleView?.isAccessibilityElement = true
        navigationItem.titleView?.accessibilityTraits = .header
        navigationItem.titleView?.accessibilityLabel = navigationItem.title

        if let titleView = navigationItem.titleView {
            titleView.showsLargeContentViewer = true
            titleView.largeContentTitle = navigationItem.title
        }
    }
```

It's also recommend it to setup the navigationBar in viewWillAppear. More info can be found [viewWIllAppear docs](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621510-viewwillappear)

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+    private func createButton(title: String, action: Selector) -> UIButton {
+        let button = ZMButton(
+            style: .secondaryTextButtonStyle,
+            cornerRadius: 16,
+            fontSpec: .buttonBigSemibold
+        )
+        button.setTitle(title, for: .normal)
+        button.addTarget(self, action: action, for: .touchUpInside)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        return button
```
agisilaos: 1. I would recommend create a new init for ZMButton so we can start using our new Font Styles instead of the old ones. 

```swift
    convenience init(
        style: ButtonStyle,
        cornerRadius: CGFloat,
        textStyle: WireTextStyle = .body1,
    ) {
        self.style = style
        self.font = .font(for: textStyle)
        self.adjustsFontForContentSizeCategory = true
        textTransform = .none
        layer.cornerRadius = cornerRadius
        contentEdgeInsets = UIEdgeInsets(top: 4, left: 16, bottom: 4, right: 16)

        applyStyle(style)
    }
```
    
2.  We would like to move away from using the #selector pattern. What we're using instead is UIAction. It can look something like this:

```swift
let action = UIAction { [weak self] _ in                                 
   self?.handleScrollToBottomTapped()
}
 button.addAction(action, for: .touchUpInside)
```

That allow us to remove the the @objc prefix before each method related to button actions

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReportView.swift
agisilaos: I would recommend setting up those as DynamicFontLabels

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+
+    private let infoLabel: UILabel = {
+        let label = UILabel()
+        label.text = Strings.TechnicalReport.info
+        label.textColor = SemanticColors.Label.textDefault
+        label.numberOfLines = 0
+        label.textAlignment = .left
+        label.translatesAutoresizingMaskIntoConstraints = false
+        return label
+    }()
```
agisilaos: I would recommend setting this label as DynamicFontLabel using the new init so we allow the text to grow.

agisilaos (changes requested):
Left some comments regarding setting up the UI, DynamicType, updated APIs.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReport.swift
David-Henner: this is for conformance to the `Shareable` protocol

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewModel.swift
David-Henner: not strictly necessary, but it helps keep the syntax highlighting of `self`

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
David-Henner: I initially made this view with SwiftUI but ran into complications conforming to the `MFMailComposeViewControllerDelegate` that would've required a lot of scaffolding. (Note that after clean up I moved the conformance to the router)

Additionally, I'm using the existing UIKit-based `ShareViewController`. It is being presented from this `SettingsDebugReport` module. Using SwiftUI for `SettingsDebugReportViewController` would've also required extra scaffolding for SwiftUI - UIKit compatibility.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
David-Henner: If I instantiate the object in place, it risks being deallocated before being able to call the completion block. And since we're in an extension, I cannot save it as a property

David-Henner: > If I instantiate the object in place, it risks being deallocated before being able to call the completion block. And since we're in an extension, I cannot save it as a property

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Files.swift
David-Henner: same answer as above

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReportView.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
David-Henner: 1. Tried to do, but turned out to be too much work for this scope
2. done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReportView.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReportView.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReportView.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReportView.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReportView.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReport.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewModel.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportRouter.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/LogFilesProvider.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS Tests/Settings/Debug Report/SettingsDebugReportViewControllerSnapshotTests.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS Tests/Settings/Debug Report/SettingsDebugReportViewModelTests.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios-sync-engine/Source/UserSession/UserSession.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Advanced.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios-sync-engine/Tests/Source/Use cases/FetchShareableConversationUseCaseTests.swift
David-Henner: Updated https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/LogFilesProvider.swift
David-Henner: taken care of in https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/MFMailComposeViewController+Logs.swift
David-Henner: done https://github.com/wireapp/wire-ios/pull/1658/commits/8c85d9ddbe3d4f96ae06fb91a9db48865e963819

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/LogFilesProvider.swift
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    // MARK: - Properties
+
+    private let temporaryDirectory = NSTemporaryDirectory()
+
+    private var logsDirectory: URL {
+        let baseURL = URL(
+            fileURLWithPath: temporaryDirectory,
+            isDirectory: false
```
samwyndham: `isDirectory` should be `true` I think

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/LogFilesProvider.swift
```diff
@@ -0,0 +1,109 @@
...
+
+    // MARK: - Types
+
+    enum Error: Swift.Error {
+        case noLogs(description: String)
+    }
+
+    // MARK: - Properties
+
+    private let temporaryDirectory = NSTemporaryDirectory()
```
samwyndham: Following on from @netbe message [here](https://github.com/wireapp/wire-ios/pull/1678#discussion_r1670712841) in my PR, I didn't realize that `NSTemporaryDirectory()` is not unique per call.. i.e calling this multiple times will return the same path. I think `temporaryDirectory` should be unique to `LogsFilesProvider` so I suggest appending  a unique path (e.g. `UUID().uuidString`) to `NSTemporaryDirectory()`.

Maybe this can all be done in `logsDirectory` and you could get rid of this property. Something like:

```swift
private var logsDirectory: URL = {
    let baseURL = URL(
        fileURLWithPath: NSTemporaryDirectory(),
        isDirectory: true
    )
    return baseURL.appendingPathComponent(UUID().uuidString).appendingPathComponent("logs")
}()
```

samwyndham (changes requested):

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
samwyndham: Thanks for the explanation!

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewModel.swift
samwyndham: I see.

In our project I see the have a mix of `self` & `` `self` ``. We should probably agree on a pattern that we all use. Feel free to close this comment!

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReport.swift
samwyndham: I **think** if you do the following the compiler will be happy although I might be wrong. 

```
func share<ZMConversation>(to: [ZMConversation]) {
    // implementation
}
```

Anyway it's a small thing. Feel free to resolve this conversation.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/LogFilesProvider.swift
David-Henner: we already do something similar but instead of having a path such as `tmp/<uuid>/logs/logs.zip` we have a path like `tmp/logs/<uuid>/logs.zip`. in `generateLogFilesZip()` we create an extra directory at the path of `logsDirectory`.

wire-ios-sync-engine/Tests/Source/Use cases/ShareFileUseCaseTests.swift
```diff
@@ -0,0 +1,118 @@
...
+    override func setUp() async throws {
+        try await super.setUp()
+
+        coreDataStackHelper = CoreDataStackHelper()
+        coreDataStack = try await coreDataStackHelper.createStack()
+
+        let cacheLocation = FileManager.default.urls(
+            for: .cachesDirectory,
+            in: .userDomainMask
+        ).first!
```
netbe:
```suggestion
        let cacheLocation = try XCTUnwrap(FileManager.default.urls(
            for: .cachesDirectory,
            in: .userDomainMask
        ).first)
```

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/LogFilesProvider.swift
samwyndham: My concern is that the API is a bit asymmetrical in that `generateLogFilesZip()` creates a log file zip but there is nothing to remove that log file zip... Instead we only have a way to remove all the files of `<user's temp>/logs`. For me this makes things less encapsulated than they could be. It requires us to keep in mind that `<user's temp>/logs` is reserved for this subsystem and could cause conflicts with other sub systems. For example we might find that some 3rd party library also makes use of `<user's temp>/logs` and then we would find that these sub-systems conflict with each other's behavior. 

So my proposed solution would be to make the logs directory unique to any given `LogsFileProvider` instance. 

This would allow for example separate parts of the app to generate zips and clean up simultaneously without affecting each other. 

In practice things will probably be fine but I do think there is a risk of running into a bug at some time with the current implementation.

netbe:
@David-Henner I don't see the body of the mail contained in the conversation when sharing the logs via Wire. That's some information that we would need either as text message or within the logs.zip

David-Henner:
> @David-Henner I don't see the body of the mail contained in the conversation when sharing the logs via Wire. That's some information that we would need either as text message or within the logs.zip

would something like this work?

<div>


</div>

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/LogFilesProvider.swift
```diff
@@ -0,0 +1,175 @@
...
+        try FileManager.default.removeItem(atPath: logsDirectory.path)
+    }
+
+    // MARK: - Helpers
+
+    private func filesToZipURLs(logFilesURLs: [URL], infoFileURL: URL) throws -> [URL] {
+        guard !logFilesURLs.isEmpty else {
+            throw Error.noLogs(description: logFilesURLs.description)
+        }
+
```
David-Henner:
```suggestion

```

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/LogFilesProvider.swift
```diff
@@ -0,0 +1,174 @@
...
+    func generateLogFilesZip() throws -> URL
+
+    /// Clears the logs directory.
+    /// Call once you are done using the URL returned by `generateLogFilesZip` to clean up.
+
+    func clearLogsDirectory() throws
+
+}
+
+struct LogFilesProvider: LogFilesProviding {
```
netbe, line 59: So to recap:
- all logs are now stored in the tmp folder of the user in /randomUUID/logs
- when sending logs, we create a unique folder for the zip then delete in /randomUUID/logs/UUID/logs.zip
- we add info.txt to contain general info inside the logs.

Can we sum up this info inside docC to wrap this all up?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/SendTechnicalReportPresenter.swift
```diff
@@ -0,0 +1,64 @@
...
+
+protocol SendTechnicalReportPresenter: MFMailComposeViewControllerDelegate {
+    func presentMailComposer(withLogs logsIncluded: Bool)
+}
+
+extension SendTechnicalReportPresenter where Self: UIViewController {
+    @MainActor
+    func presentMailComposer(withLogs logsIncluded: Bool) {
+        presentMailComposer(withLogs: logsIncluded, sourceView: nil)
+    }
```
netbe: I think we can remove logsIncluded as sending the report without the logs is no longer an option

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
```diff
@@ -0,0 +1,168 @@
...
+        // Ensure title supports LargeContentViewer
+        navigationItem.titleView?.isAccessibilityElement = true
+        navigationItem.titleView?.accessibilityTraits = .header
+        navigationItem.titleView?.accessibilityLabel = navigationItem.title
+
+        if let titleView = navigationItem.titleView {
+            titleView.showsLargeContentViewer = true
+            titleView.largeContentTitle = navigationItem.title
+        }
+    }
```
agisilaos: a new method has beed introduced inside WireReusableComponents: We can use the `setupNavigationBarTitle(Strings.TechnicalReportSection.title.capitalized)` and we get all of this functionality.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
```diff
@@ -0,0 +1,168 @@
...
+class SettingsDebugReportViewController: UIViewController {
+
+    // MARK: - Constants
+
+    enum LayoutConstants {
+        static let spacing: CGFloat = 8
+        static let padding: CGFloat = 20
+        static let safeBottomPadding: CGFloat = 30
+        static let buttonHeight: CGFloat = 48
+    }
```
agisilaos: We can make this private:

```suggestion
    private enum LayoutConstants {
        static let spacing: CGFloat = 8
        static let padding: CGFloat = 20
        static let safeBottomPadding: CGFloat = 30
        static let buttonHeight: CGFloat = 48
    }
```

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
```diff
@@ -0,0 +1,168 @@
...
+            style: .secondaryTextButtonStyle,
+            cornerRadius: 16,
+            fontSpec: .buttonBigSemibold
+        )
+        button.setTitle(title, for: .normal)
+        button.addAction(action, for: .touchUpInside)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        return button
+    }
+}
```
agisilaos, line 146: I would recommend replacing this init with one using the new WireTextStyle.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReportView.swift
```diff
@@ -0,0 +1,140 @@
...
+
+class ShareableDebugReportView: UIView {
+
+    // MARK: - Constants
+
+    enum LayoutConstants {
+        static let spacing: CGFloat = 2
+        static let padding: CGFloat = 12
+        static let iconSize: CGFloat = 28
+    }
```
agisilaos: We can make this private:

```suggestion
    private enum LayoutConstants {
        static let spacing: CGFloat = 2
        static let padding: CGFloat = 12
        static let iconSize: CGFloat = 28
    }
```

agisilaos (commented):
Good work! Left some comments regarding using a different init for the ZMButton.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
David-Henner: `ZMButton` doesn't have this kind of init, and implementing it is out the scope of this PR

wire-ios-sync-engine/Tests/Source/Use cases/ShareFileUseCaseTests.swift
David-Henner: https://github.com/wireapp/wire-ios/pull/1658/commits/c81926da4e3baa2bf5253ca1bc051109d41a4c71

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/LogFilesProvider.swift
David-Henner: https://github.com/wireapp/wire-ios/pull/1658/commits/c81926da4e3baa2bf5253ca1bc051109d41a4c71

wire-ios/Wire-iOS/Sources/UserInterface/Settings/SendTechnicalReportPresenter.swift
David-Henner: https://github.com/wireapp/wire-ios/pull/1658/commits/c81926da4e3baa2bf5253ca1bc051109d41a4c71

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
David-Henner: https://github.com/wireapp/wire-ios/pull/1658/commits/c81926da4e3baa2bf5253ca1bc051109d41a4c71

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReportView.swift
David-Henner: https://github.com/wireapp/wire-ios/pull/1658/commits/c81926da4e3baa2bf5253ca1bc051109d41a4c71

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportViewController.swift
David-Henner: https://github.com/wireapp/wire-ios/pull/1658/commits/5314a9766a52b49606900f6b32a0fd242067a78e

wire-ios/Wire-iOS Tests/ReferenceImages/ShareViewControllerTests/testThatItRendersCorrectlyShareViewController_FileMessage.iPad-landscape.png
samwyndham, line 1: How come we have all these empty screen snapshots? What are we testing?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReportView.swift
```diff
@@ -0,0 +1,140 @@
...
+
+    private enum LayoutConstants {
+        static let spacing: CGFloat = 2
+        static let padding: CGFloat = 12
+        static let iconSize: CGFloat = 28
+    }
+
+    // MARK: - Views
+
+    private lazy var topLabel: DynamicFontLabel = {
```
samwyndham: In the case of a UIView subclass (vs a UIViewController) I don't think it is worth making its properties `lazy` as they will be created in the init anyway.

samwyndham (approved):

wire-ios/Wire-iOS Tests/ReferenceImages/ShareViewControllerTests/testThatItRendersCorrectlyShareViewController_FileMessage.iPad-landscape.png
David-Henner: I don't know. Opened a ticket https://wearezeta.atlassian.net/browse/WPB-10223

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/ShareableDebugReportView.swift
David-Henner: https://github.com/wireapp/wire-ios/pull/1658/commits/261c26dd45fddd4f82a35ad6c5419f1c0d19acb7

agisilaos (approved):

PR #1664, opened 2024-07-05
fix: unit tests without backend info domain - WPB-10065

markusfassbender:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10065" title="WPB-10065" target="_blank">TaskWPB-10065</a>  iOS: fix unit tests without backend info domain
  </td></table>
  <br />
 

### Issue

We have subtile but serious problem with unit tests and the static BackendInfo:
Some unit tests rely implicitly that BackendInfovalues are set, for example testThatLastReadUpdatesInSelfConversationDontExpire(). It works fine on CI with all tests magically, but if you have a fresh simulator and only run this test, it fails.
The best would be not to use the static BackendInfo in all kind of classes, but that is too big to refactor in once now.

I wrote two solutions that came to my mind to solve this problem:
1. with a struct callAsFunction 
2. with a kind of token (the naming is difficult)

### Testing

---

### Checklist


samwyndham:
@markusfassbender 

I don't no much about the problem but looking at it quickly I was wondering if you could switch out the `BackendInfo.storage` for temporary storage:

```swift
BackendInfo.storage = UserDefaults(suiteName:  "foo")!
BackendInfo.apiVersion = .v6

// Run tests
// ...

// Clean up
BackendInfo.storage.removePersistentDomain(forName: "foo")
```

wire-ios-data-model/Tests/Source/Model/Conversation/ZMConversationTests+Messages.swift
```diff
@@ -268,6 +270,8 @@ class ZMConversationMessagesTests: ZMConversationTestsBase {
     }
 
     func testThatLastReadUpdatesInSelfConversationDontExpire() {
+        let backendInfoToken = TemporaryBackendInfoToken(domain: "wire.com")
```
netbe, line 274: @markusfassbender we could use defer here to put everything in same place

idea 2: use a swiftmacro to specify this test will use backend domain X
+ add swiftlint rule to ban usage of BackendInfo in tests and use this instead

markusfassbender: @samwyndham @netbe thanks for your ideas. The issue is, that one can easily forget to do the reset and then other tests are influenced, e.g. `testThatItReportsToDelegate_WhenFederationHasBeenEnabled`. That it's not reset is the reason some tests are working on CI, but not individually.

So my idea solution was something that does reset automatically..

samwyndham: @markusfassbender You could always fix that issue by doing something like the following although we should still clean up for memory reasons.

```swift
BackendInfo.storage = UserDefaults(suiteName:  UUID().uuidString)!
```

~This should be safe as in XCTest, each test run uses a different static instance of `BackendInfo`. E.g. setting `BackendInfo.storage` in one test shouldn't affect that property in another test.~ (I'm wrong)

samwyndham: Hmmm. Maybe I'm wrong about the about that. This is what I had read but just tried it in reality and that is not the behavior.

samwyndham (commented):
Out of the two proposals I personally prefer the token approach. I might do something like the following instead though, keeping the stubbing of data directly on `BackendInfo`.

```swift
extension BackendInfo {
    static func setTestingStorage() -> AnyCancellable {
        let originalStorage = BackendInfo.storage
        BackendInfo.storage = UserDefaults(suiteName: "something")!
        
        return AnyCancellable {
            BackendInfo.storage = originalStorage
        }
    }
}
```

In usage:

```swift
let token = BackendInfo.setTestingStorage()
BackendInfo.apiVersion = .v1
// ...
withExtendedLifetime(token)
```

But in practice I'm not sure how necessary it is vs the discipline of switching the backing storage in `setUp` and removing the backing storage in `tearDown` as this kind of pattern is probably necessary with other systems we have.

This is just my quick thoughts... I haven't actually delved into the problem.

markusfassbender:
Closed in favor of `@_spi` usage: #1689

PR #1672, opened 2024-07-08
fix: remove filtering events - WPB-9221

netbe:
<table>
<td>
  </td></table>
  <br />
 

### Issue

Here's a fix for a potential scenario for missing messages:
1. The app fetches 100 events 
2. It clears the already fetch event ids before we actually processEvents from ZMMissingUpdateEventsTranscoder 
3. it stores 100 event ids and then decrypt and store up to 50 events (storing the 50th event as lastUpdateEventId)
4. When opening the app, it would start to fetch a new batch of events from last update event (51...), processing the filtered events(51th->100th events) previously fetched

##### Solution:

For now it's easier to remove the filtering of already fetched events, which would resolve the current issue. It will produce decryption errors for duplicate messages but these errors won't make it to the user. 

Note: it would be easier to reimplement a filtering solution if needed after we refactored the quick sync.

### Testing

1. add delay to decryptAndStoreEvents
2. spam 100 events via background fetch or via notification extension
3. let background fetch happen 
4. check we don't have missing events
---

### Checklist


wire-ios-sync-engine/Source/Synchronization/Transcoders/ZMMissingUpdateEventsTranscoder.m
```diff
@@ -330,10 +327,6 @@ - (NSUUID *)nextUUIDFromResponse:(ZMTransportResponse *)response forListPaginato
 
     NSUUID *latestEventId = [self processUpdateEventsAndReturnLastNotificationIDFromResponse:response];
 
-    if (!self.listPaginator.hasMoreToFetch) {
-        [self.previouslyReceivedEventIDsCollection discardListOfAlreadyReceivedPushEventIDs];
-    }
```
netbe, line 335: this is where we may discard already received events before decrypting and storing them because this happens in the `processUpdateEventsAndReturnLastNotificationIDFromResponse:` method above but this is asynchronous

johnxnguyen (approved):

markusfassbender (approved):
Code looks fine for me, even though I'm not sure if I understood the old logic completely ;)

PR #1678, opened 2024-07-09
fix: Sending logs when email client not configured - WPB-9948

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9948" title="WPB-9948" target="_blank">BugWPB-9948</a>  Sending logs - when email client is not configured
  </td></table>
  <br />
 

### Issue

Currently when sending logs via a share sheet we attempt to attach all logging URLs derived from LogFileDestination and ZMSLog. However in some cases no logs have actually been created, which unsurprisingly can cause sending logs to fail.

Originally I created a fix using the current implementation and filtering logging URLs which had no associated file. See https://github.com/wireapp/wire-ios/pull/1656. However I later realized that @David-Henner was working on a related ticket (https://github.com/wireapp/wire-ios/pull/1658) which gathers logs through a different method. I.e currently there are multiple ways of gathering debug logs. In this PR I fix the issue by using the same method of gathering logs as https://github.com/wireapp/wire-ios/pull/1658 uses.

This PR:

- Uses `LogFilesProvider.generateLogFilesZip` to gather logs instead of `DebugLogSender. debugLogs`. This has the side affect of fixing the bug.
- Deletes  `DebugLogSender.debugLogs` and related code.

### Testing

I'm not really sure of any reasonable way to write automated tests for this fix. If you have some ideas let me know. 

To manually test:

0. Ensure that Mail.app is not configured, but Gmail.app is.
1. Clean install and log in
2. Go to Account > Settings > Advanced > Debug Report > Send Report > Clic kOK on alert
3. Select Gmail and send logs
4. Check logs have been sent in Gmail.
5. Clean install and log in
6. Shake phone to go to Developer Tools
7. Select Debug Actions > Send Debug Report > Click OK on alert
8. Select Gmail and send logs
9. Check logs have been sent in Gmail.

### Checklist


wire-ios/Wire-iOS/Sources/Developer/DebugAlert.swift
```diff
@@ -89,19 +89,30 @@ final class DebugAlert {
...
         alert.addAction(UIAlertAction(title: L10n.Localizable.General.ok, style: .default, handler: { _ in
-            let activity = UIActivityViewController(activityItems: logPaths, applicationActivities: nil)
+            let logFilesProvider = LogFilesProvider()
+            let logsFileURL = logFilesProvider.generateLogFilesZip()
+
+            let activity = UIActivityViewController(activityItems: [logsFileURL], applicationActivities: nil)
             activity.configPopover(pointToView: sourceView ?? controller.view)
+            activity.completionWithItemsHandler = { _, _, _, _ in
+                do {
+                    try logFilesProvider.clearTemporaryDirectory()
```
netbe: why do we clear files here? and which files?

netbe (commented):
added one question.
+ Is LogFilesProvider from the targeted branch?

samwyndham:
> added one question.
> 
> * Is LogFilesProvider from the targeted branch?

@netbe Yes it is. See [here](https://github.com/wireapp/wire-ios/blob/08680fcc8ecc169798ba0a1c93a1c380c51c3bf8/wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug%20Report/LogFilesProvider.swift#L46)

wire-ios/Wire-iOS/Sources/Developer/DebugAlert.swift
samwyndham: It is part of the [documentation](https://github.com/wireapp/wire-ios/blob/08680fcc8ecc169798ba0a1c93a1c380c51c3bf8/wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug%20Report/LogFilesProvider.swift#L42C5-L42C42) of `LogFilesProvider` to clean up:

```swift
/// Clears the temporary directory. 
/// Call once you are done using the URL returned by `generateLogFilesZip` to clean up.
func clearTemporaryDirectory() throws
```

Basically `LogFilesProvider` creates a temporary directory for saving a zip file to. Once we are done with the zip `clearTemporaryDirectory()` removes the temporary directory.

However now you mention it perhaps the behavior of `LogFilesProvider` in it's current implementation may be buggy. 

I _expect_ the following might fail and it may be worth looking into it @David-Henner.

```
let provider = LogFilesProvider() // Implicitly creates temp directory
let a = generateLogFilesZip()
try provider. clearTemporaryDirectory() // Deletes temp directory
let b = generateLogFilesZip() // <- This may now fail without error because the temporary directory no longer exists.
```

agisilaos (approved):

wire-ios/Wire-iOS/Sources/Developer/DebugAlert.swift
netbe: Also make sure temporary directory is unique as NSTemporaryDirectory is used elsewhere

samwyndham: Good point @netbe . No it is not unique! I just tried it. Will mention this in https://github.com/wireapp/wire-ios/pull/1658

samwyndham: Please could you re-review @netbe

caldrian (approved):

wire-ios/Wire-iOS/Sources/Developer/DebugAlert.swift
samwyndham: I'm resolving this conversation as I think the points are addressed. Feel free to re-open.

PR #1681, opened 2024-07-10
Parse the web backup file

samwyndham:
### Issue

Create and hook up a method to start the restore process. It allows us to select the backup file and does the parsing. We also make a few updates to the models due to some values being optional etc.

FYI @caldrian 

### Checklist

### DIFF
--- a/wire-ios-data-model/Source/Universal backup/Model/ConversationBackupModel.swift
+++ b/wire-ios-data-model/Source/Universal backup/Model/ConversationBackupModel.swift
@@ -19,14 +19,14 @@
 import Foundation
 
 public struct ConversationBackupModel: Codable {
-    
+
     public let id: UUID
 
     public let domain: String
 
-    public let name: String
+    public let name: String?
 
-    public let otherParticipants: Set<QualifiedID>
+    public let otherParticipants: Set<QualifiedID>?
 
     public init(
         id: UUID,
--- a/wire-ios-data-model/Source/Universal backup/Model/ConversationBackupModel.swift
+++ b/wire-ios-data-model/Source/Universal backup/Model/ConversationBackupModel.swift
@@ -19,24 +19,34 @@
 import Foundation
 
 public struct ConversationBackupModel: Codable {
-    enum CodingKeys: String, CodingKey {
-        case id
-        case domain
-        case name
-        case otherParticipants = "qualified_others"
-    }
-
+    
     public let id: UUID
+
     public let domain: String
 
     public let name: String
 
     public let otherParticipants: Set<QualifiedID>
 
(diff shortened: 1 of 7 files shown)


wire-ios-data-model/Source/Universal backup/Model/ConversationBackupModel.swift
```diff
@@ -18,22 +18,35 @@
 
 import Foundation
 
-public struct ConversationBackupModel {
+public struct ConversationBackupModel: Codable {
 
-    public let qualifiedID: QualifiedID
```
caldrian, line 23: optional: a `qualifiedID` computed property might be convenient

wire-ios-data-model/Source/Universal backup/Model/TextMessageBackupModel.swift
```diff
@@ -18,21 +18,21 @@
...
+public struct MessageAddBackupModel {
 
     public let nonce: UUID
 
     public let content: String
 
     public let conversationID: QualifiedID
 
-    public let senderUserID: QualifiedID
+    public let senderUserID: QualifiedID?
```
caldrian, line 29: Why does it have to be an optional?
Doesn't a message always have a sender?

caldrian (approved):

wire-ios-data-model/Source/Universal backup/Model/ConversationBackupModel.swift
```diff
@@ -18,22 +18,35 @@
...
+    public let id: UUID
 
-    public let name: String
+    public let domain: String
 
-    public let participants: Set<QualifiedID>
+    public let name: String?
```
caldrian, line 27: What is the difference between an empty string and nil?

wire-ios-data-model/Source/Universal backup/Model/ConversationBackupModel.swift
```diff
@@ -18,22 +18,35 @@
...
+
+    public let otherParticipants: Set<QualifiedID>?
```
caldrian, line 29: What is the difference between an empty set and nil?

wire-ios-data-model/Source/Universal backup/Model/ConversationBackupModel.swift
johnxnguyen: We made it nil to fix the decoding errors, but it doesn't matter since we'll remove the name entirely.

wire-ios-data-model/Source/Universal backup/Model/ConversationBackupModel.swift
johnxnguyen: Same here, we'll delete this

PR #1689, opened 2024-07-12
fix: unit tests using backend info `apiVersion` - WPB-10065

markusfassbender:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10065" title="WPB-10065" target="_blank">TaskWPB-10065</a>  iOS: fix unit tests without backend info
  </td></table>
  <br />
 

### Issue

Introduce two new funcs to mock `BackendInfo`, by importing via `@_spi(MockBackendInfo)`. You can find more information about spi [here](https://varanios.com/articles/spi-or-how-to-develop-better-apis-in-swift).

In this PR I just updated the usage of `apiVersion`, but more PRs can follow for `domain`, `preferredAPIVersion` and `isFederationEnabled`.

Remove `XCTestCase+APIVersion` and other helpers, that didn't require too much refactoring. The best is to call the mocking just in the test itself.

### Testing

- Run all unit tests suites

---

### Checklist


wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests+AccessToken.swift
```diff
@@ -16,11 +16,14 @@
 // along with this program. If not, see http://www.gnu.org/licenses/.
 //
 
-import Foundation
-@testable import WireSyncEngine
 import WireTransport
 import XCTest
```
caldrian:
```suggestion
import XCTest
```
see the SwiftLint error below

wire-ios-transport/Source/Public/BackendInfo.swift
```diff
@@ -29,7 +29,10 @@ public enum BackendInfo {
 
     }
 
-    public static var storage = UserDefaults.standard
+    private static let productionStorage = UserDefaults.standard
```
caldrian: Is `.standard` really the "production" value?
In `LaunchSequenceOperation.swift` I see the line
```
BackendInfo.storage = .applicationGroup
```

What about not setting any default value at all?
```suggestion
    private static let productionStorage: UserDefaults!
```

caldrian (approved):
👏🧹

wire-ios-transport/Source/Public/BackendInfo.swift
```diff
@@ -68,5 +71,18 @@ public enum BackendInfo {
         let storedValue = storage.integer(forKey: key.rawValue)
         return APIVersion(rawValue: Int32(storedValue))
     }
+}
+
+// MARK: - Mock
+
```
samwyndham, line 88: Consider putting this code in an `#IF DEBUG` to prevent it from being used for some reason in production

samwyndham (approved):

markusfassbender:
I tried to solve the 94 failing tests, but didn't understand what happens. My assumption is, that there are many implicit expectations about the `BackendInfo` values, also `domain` for example, and introducing here a mock environment and reset to empty values seems to break the tests.

e.g. the helper function does not assert true anymore:
`IntegrationTests.searchAndConnectToUser(withName name: String, searchQuery: String)`

My suggestion is to go with the new change and solve the underlying expectations, I think solving the root might solve all other tests or many at least. Unfortunately I didn't succeed with it today.

wire-ios-transport/Source/Public/BackendInfo.swift
samwyndham: ~Interesting. It looks like that is the only place it is being overridden and it is being done by the `AutomationHelper` which I assume is for UI tests. I'll take a close look.~

I'm wrong. You are right that in production we use `.applicationGroup` which is a bit annoying as it is set in a module at a higher level. I'm going to look into making it `UserDefaults!` as you suggested.

wire-ios-transport/Source/Public/BackendInfo.swift
```diff
@@ -29,44 +29,85 @@ public enum BackendInfo {
...
+    @_spi(MockBackendInfo)
+    public static func resetMocking() {
+        if let storage = mockStorage as? MockUserDefaults {
+            storage.removePersistentDomain(forName: storage.suiteName)
+        }
+        BackendInfo.mockStorage = nil
+    }
+}
+
+private class MockUserDefaults: UserDefaults {
```
samwyndham, line 104: I created a subclass here just to have access to the `suiteName`. This allows us to delete it's data in `resetMocking()`

wire-ios-transport/Source/Public/BackendInfo.swift
samwyndham: I tried using `UserDefaults!` but this causes crashes in tests as we must still be implicitly depending on `BackendInfo` in many places. Therefore, I took a different approach where we return `mockStorage` if it is set, otherwise we return `storage`.

samwyndham:
Closing this PR as I'm implementing these changes in separate PRs per target - so far these are:

- https://github.com/wireapp/wire-ios/pull/1768
- https://github.com/wireapp/wire-ios/pull/1738

More will follow

PR #1694, opened 2024-07-15
feat: introduce WireMacros - no ticket

johnxnguyen:
### Issue

Create a `WireMacros` package:
- `@Provided` generates a "provider" companion protocol for other protocols.
- `#UUID` validates uuid strings at compile time.
- new macros can be created and tested here too.

**Open question:** is the [potential overhead of using macros](https://forums.swift.org/t/swift-macros-build-time-overhead-concerns/70443/10) a concern? 

---

### Checklist

### DIFF
--- a/WireMacros/Package.swift
+++ b/WireMacros/Package.swift
@@ -1,0 +1,51 @@
+// swift-tools-version: 5.10
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+import CompilerPluginSupport
+
+let package = Package(
+    name: "WireMacros",
+    platforms: [.macOS(.v10_15), .iOS(.v13), .tvOS(.v13), .watchOS(.v6), .macCatalyst(.v13)],
+    products: [
+        // Products define the executables and libraries a package produces, making them visible to other packages.
+        .library(
+            name: "WireMacros",
+            targets: ["WireMacros"]
+        ),
+        .executable(
+            name: "WireMacrosClient",
+            targets: ["WireMacrosClient"]
+        ),
+    ],
+    dependencies: [
+        .package(url: "https://github.com/apple/swift-syntax.git", from: "509.0.0"),
+    ],
+    targets: [
+        // Targets are the basic building blocks of a package, defining a module or a test suite.
+        // Targets can depend on other targets in this package and products from dependencies.
+        // Macro implementation that performs the source transformation of a macro.
+        .macro(
+            name: "WireMacrosMacros",
+            dependencies: [
+                .product(name: "SwiftSyntaxMacros", package: "swift-syntax"),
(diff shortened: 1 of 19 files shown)


WireMacros/Sources/WireMacrosClient/main.swift
```diff
@@ -0,0 +1,46 @@
...
+struct FooUseCaseProvider: FooUseCaseProtocolProvider {
+
+    func makeFooUseCaseProtocol() -> any FooUseCaseProtocol {
+        FooUseCase()
+    }
+
+}
+
+let uuid = #UUID("7411ca17-ba08-4905-92d2-0617a8c810ca")
+print(uuid)
```
markusfassbender, line 46: Should these lines here?

markusfassbender (approved):

markusfassbender:
added an example how to use it in `WireAPI` tests: d9661cd21ef561d98cb2fd4bfb1ac07df16618eb
Great job @johnxnguyen , thanks!

WireMacros/Sources/WireMacrosClient/main.swift
johnxnguyen: This main file is kind of a playground to develop the macros.... maybe it's not needed since we can do that in tests and/or in our other modules.

WireMacros/Package.swift
```diff
@@ -0,0 +1,51 @@
...
+            name: "WireMacrosClient",
+            targets: ["WireMacrosClient"]
+        ),
+    ],
+    dependencies: [
+        .package(url: "https://github.com/apple/swift-syntax.git", from: "509.0.0"),
+    ],
+    targets: [
+        // Targets are the basic building blocks of a package, defining a module or a test suite.
+        // Targets can depend on other targets in this package and products from dependencies.
```
samwyndham, line 26: I'd probably remove these two lines of comments

WireMacros/Sources/WireMacros/WireMacros.swift
```diff
@@ -0,0 +1,88 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// A macro that, when attached to a protocol declaration, produces a
+/// companion "provider" protocol that can be used inject dependencies.
```
samwyndham, line 22: This is pretty interesting. I need to take a look to understand the `provider` pattern more

samwyndham (commented):
Nice! I'm super in favor of adding macros if we can keep the compile performance costs are not prohibitive - especially those which can get rid of unnecessary optional values as these tend explode.

I would also consider adding something like `#URL("http://example.com")`

PR #1715, opened 2024-07-17
fix:  Remove dialog and toggle for crash data reports - WPB-10026

agisilaos:
With this PR we get rid of the toggle and cell regarding crash data reports. Since App Center was removed a while ago that toggle and cell has no effect. In addition to that we did of a cleanup and deleted associated tests, strings and public interfaces. 

---

### Checklist


caldrian (approved):

samwyndham (approved):
🧹

PR #1716, opened 2024-07-18
refactor: coordinate quick sync - WPB-9923

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9923" title="WPB-9923" target="_blank">TaskWPB-9923</a>  [iOS] Create SyncManager
  </td></table>
  <br />
 

### Issue

This PR introduces `SyncManager` to coordinate the various syncing mechanisms in the app. For now, it coordinates quick sync and live events. Later, we'll also include slow sync here.

The general flow of quick sync is:
- Open push channel, buffer live events as they come.
- Pull update events from the backend, decrypt and store in the event db.
- Fetch, process, and delete all events from the event db.
- Decrypt and process buffered live events.
- Decrypt and process new live events as they arrive.

The live event buffering is achieved via the `AsyncStream`: all elements that the stream produces before the steam is iterated will be buffered and then delivered as soon as iteration begins.

One difference with this flow compared to the existing flow is that live events aren't stored in the event database and then immediately fetched again, we simply decrypt and process immediately.

The PR also adds the ability to suspend syncing, which will cancel any ongoing tasks. This would likely be called when the app enters the background.

### Testing

This is still not hooked up to real flows so the only testing I've done so far is unit testing.

---

### Checklist


WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
```diff
@@ -0,0 +1,167 @@
...
+
+    init(
+        updateEventsRepository: any UpdateEventsRepositoryProtocol,
+        updateEventProcessor: any UpdateEventProcessorProtocol
+    ) {
+        self.updateEventsRepository = updateEventsRepository
+        self.updateEventProcessor = updateEventProcessor
+    }
+
+    func performQuickSync() async throws {
```
johnxnguyen, line 51: Do we need to make this non-reentrant?

WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
```diff
@@ -0,0 +1,167 @@
...
+                WireLogger.sync.info("live task was cancelled")
+            } catch {
+                WireLogger.sync.error("live task encountered error: \(error)")
+            }
+        }
+
+        syncState = .live(liveTask)
+    }
+
+    func suspend() async throws {
```
johnxnguyen, line 96: Do we need to make this non-reentrant?

WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
```diff
@@ -0,0 +1,167 @@
...
+            do {
+                try await updateEventProcessor.processEvent(event)
+            } catch {
+                WireLogger.sync.error("failed to process live event, dropping: \(error)")
+            }
+        }
+
+        if !envelope.isTransient {
+            updateEventsRepository.storeLastEventEnvelopeID(envelope.id)
+        }
```
johnxnguyen, line 140: Should we persist after processing, or after decryption?

wire-ios-data-model/Source/Model/StoredUpdateEvent/StoreUpdateEvent.swift
```diff
@@ -0,0 +1,80 @@
...
+/// should be persisted as soon as possible so that they can be retrieved
+/// for later processing.
+
+@objc(StoredUpdateEvent)
+public final class StoredUpdateEvent: NSManagedObject {
+
+    public static let entityName = "StoredUpdateEvent"
+    public static let SortIndexKey = "sortIndex"
+
+    /// The event id.
```
samwyndham: Doc comments are so useful for these objects!

WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Event decoding/UpdateEventDecodingProxy.swift
```diff
@@ -18,7 +18,27 @@
...
+/// JSON payloads received from the backed.
+///
+/// The need for this wrapper arises from another need to persist
+/// update events in a database: to persist an update event we
+/// persist its encoded data, from which we need to decode it after
+/// fetching it from the database. This coding and decoding needs
+/// to be symmetric and for this we rely on the automatic conformance
+/// of `UpdateEvent`.
+///
+/// However, we still need to decode the update events from the JSON
```
samwyndham: I don't quite understand the need of this proxy from the comment. Perhaps you could explain a bit further.

WireDomain/Sources/WireDomain/Event Decryption/UpdateEventDecryptor.swift
```diff
@@ -0,0 +1,137 @@
...
+
+                    await appendFailedToDecryptProteusMessage(
+                        eventData: eventData,
+                        error: error
+                    )
+                } catch {
+                    WireLogger.updateEvent.error(
+                        "failed to decrypt proteus event, dropping: \(error.localizedDescription)",
+                        attributes: logAttributes
+                    )
```
samwyndham: Just to check we don't need to append failed to decrypt events in this case right?

WireDomain/Sources/WireDomain/Repositories/UpdateEventsRepository.swift
```diff
@@ -0,0 +1,119 @@
...
+                        id: envelope.id,
+                        index: currentIndex
+                    )
+
+                    currentIndex += 1
+                }
+
+                if !envelope.isTransient {
+                    // Update the last event id so we don't refetch the same events.
+                    lastEventIDRepository.storeLastEventID(envelope.id)
```
samwyndham: As a sanity check it really is the envelope ID we want to store right? Perhaps `storeLastEventID` should be renamed to show this.

If any events in an envelope fail to persist should we delete the ones that were persisted in that envelope. I assume in this case we would get duplicates because as some point we would try to fetch the same envelope again. Probably it is better to handle this situation with de-duplications 🤔

wire-ios-data-model/Source/Model/StoredUpdateEvent/StoreUpdateEvent.swift
```diff
@@ -83,4 +83,14 @@ public final class StoredUpdateEvent: NSManagedObject {
     @NSManaged
     public var sortIndex: Int64
 
+    /// Compute the highest index of all stored events.
+
+    public static func highestIndex(in context: NSManagedObjectContext) -> Int64 {
+        let fetchRequest = NSFetchRequest<StoredUpdateEvent>(entityName: StoredUpdateEvent.entityName)
+        fetchRequest.sortDescriptors = [NSSortDescriptor(key: StoredUpdateEvent.SortIndexKey, ascending: false)]
+        fetchRequest.fetchBatchSize = 1
+        let result = context.fetchOrAssert(request: fetchRequest)
```
samwyndham: If this fails in production we will end with a 0? I'm wondering if this method should instead throw. I'm not sure how we should handle this. Perhaps it is better to crash the app because this is kind of a completely broken state.

samwyndham (commented):
Just commenting here as I don't yet feel comfortable with my understanding of syncing but looks good!

WireDomain/Sources/WireDomain/Repositories/UpdateEventsRepository.swift
samwyndham: Oh I see this has changed - was reviewing commit by commit. Closing.

wire-ios-data-model/Source/Model/StoredUpdateEvent/StoreUpdateEvent.swift
samwyndham: Oh I see this has changed - was reviewing commit by commit. Closing.

WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Event decoding/UpdateEventDecodingProxy.swift
johnxnguyen: This was not meant to be in the PR but am happy to explain it here. Update events come from the backend and need to be (manually) decoded into `UpdateEvent`. We also will store these update events in the event database, and to do so we need to encode them to Data, then decode them when retrieving them. This encoding and decoding uses the default conformance of `Codable`. But you can't have the default conformance if you also explicitly conform to `Decoable` (for the network models), hence a proxy object that provides the manually decoding.

It was discussed though that public models such as `UpdateEvent` shouldn't conform to `Codable` since changing the model would probably break something on the consumer side. Instead we might create mirror objects on the consumer side.

WireDomain/Sources/WireDomain/Event Decryption/UpdateEventDecryptor.swift
johnxnguyen: How do you mean?

WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Event decoding/UpdateEventDecodingProxy.swift
samwyndham: Cool, now I've got it. Thank you!

WireDomain/Sources/WireDomain/Event Decryption/UpdateEventDecryptor.swift
samwyndham: I mean in the case where `error` is a `ProteusError` we append it with `appendFailedToDecryptProteusMessage`. However we don't call `appendFailedToDecryptProteusMessage` in the default error case, even though the error message includes `failed to decrypt proteus event`.

This was just a sanity check comment to make sure you didn't accidentally miss something, but please feel free to resolve this message.

netbe:
> One difference with this flow compared to the existing flow is that live events aren't stored in the event database and then immediately fetched again, we simply decrypt and process immediately.

@johnxnguyen so what if we crash, I guess we loose all live events, right? 
are they fetched again in the quick sync restarting the app?

WireDomain/Sources/WireDomain/Helpers/AsyncSequence+ToStream.swift
```diff
@@ -0,0 +1,30 @@
...
+//
+
+import Foundation
+
+extension AsyncSequence {
+
+    func toStream() -> AsyncStream<Element> {
+        var iterator = makeAsyncIterator()
+        return AsyncStream {
+            try? await iterator.next()
```
netbe: hard to see the impact, but the error is swallowed here ?

WireDomain/Sources/WireDomain/Repositories/UpdateEventsRepository.swift
```diff
@@ -34,20 +34,55 @@ protocol UpdateEventsRepositoryProtocol {
...
+    /// an asynchronous stream.
+    ///
+    /// The envelopes are bufferred until a consumer starts to iterate though
+    /// the stream.
+    ///
+    /// - Returns: An asynchronous stream of `UpdateEventEnvelope`s.
+
+    func startBufferingLiveEvents() async throws -> AsyncStream<UpdateEventEnvelope>
+
+    /// Close the piush channel and stop fnish the asynchronous stream of
```
netbe:
```suggestion
    /// Close the push channel and stop the asynchronous stream of
```

WireDomain/Sources/WireDomain/Repositories/UpdateEventsRepository.swift
```diff
@@ -163,4 +209,46 @@ final class UpdateEventsRepository: UpdateEventsRepositoryProtocol {
...
+            }
+        }.toStream()
+    }
+
+    func stopReceivingLiveEvents() async {
+        await pushChannel.close()
+    }
+
+    func storeLastEventEnvelopeID(_ id: UUID) {
+        WireLogger.sync.debug("storing last event id")
```
netbe: we should continue to log with attributes in order to track event, ie:

```
        WireLogger.sync.info("store last event id", attributes: [LogAttributesKey.lastEventID.rawValue: String(describing: id?.safeForLoggingDescription ?? "<nil>")])
```

WireDomain/Sources/WireDomain/Repositories/UpdateEventsRepository.swift
```diff
@@ -163,4 +209,46 @@ final class UpdateEventsRepository: UpdateEventsRepositoryProtocol {
...
+
+    func startBufferingLiveEvents() async throws -> AsyncStream<UpdateEventEnvelope> {
+        try await pushChannel.open().compactMap {
+            do {
+                WireLogger.sync.debug("decrypting live event")
+                var envelope = $0
+                envelope.events = try await self.updateEventDecryptor.decryptEvents(in: envelope)
+                return envelope
+            } catch {
+                WireLogger.sync.error("failed to decrypt live event, dropping: \(error)")
```
netbe: add event / enveloppe id

WireDomain/Sources/WireDomain/Repositories/UpdateEventsRepository.swift
```diff
@@ -85,11 +123,23 @@ final class UpdateEventsRepository: UpdateEventsRepositoryProtocol {
...
+
             for envelope in envelopes {
+                count += 1
+                WireLogger.sync.debug("decrypting envelope (\(count) of \(batchCount)")
+
                 // We can only decrypt once so store the decrypted events for later retrieval.
                 var decryptedEnvelope = envelope
                 decryptedEnvelope.events = try await updateEventDecryptor.decryptEvents(in: envelope)
 
+                WireLogger.sync.debug("persisting envelope (\(count) of \(batchCount)")
```
netbe: add envelope id

WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
```diff
@@ -0,0 +1,167 @@
...
+            try await quickSyncTask.value
+        } catch {
+            try await suspend()
+            throw error
+        }
+
+        let liveTask = Task {
+            do {
+                for try await envelope in liveEventsStream {
+                    WireLogger.sync.info("received live event")
```
netbe: complete log

WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
```diff
@@ -0,0 +1,167 @@
...
+
+    func performQuickSync() async throws
+
+    /// Stop all syncing activities and prepare to idle.
+
+    func suspend() async throws
+
+}
+
+final class SyncManager: SyncManagerProtocol {
```
netbe, line 35: should it be an actor?

WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
```diff
@@ -0,0 +1,167 @@
...
+                    WireLogger.sync.info("received live event")
+                    try Task.checkCancellation()
+                    await processLiveEvents(in: envelope)
+                }
+            } catch is CancellationError {
+                WireLogger.sync.info("live task was cancelled")
+            } catch {
+                WireLogger.sync.error("live task encountered error: \(error)")
+            }
+        }
```
netbe, line 91: the handling between the quickSyncTask and liveTask is different, any reason why?

johnxnguyen:
> @johnxnguyen so what if we crash, I guess we loose all live events, right?
are they fetched again in the quick sync restarting the app?

@netbe if we crash during event processing, then I suppose yes we would lose it. We could store it in the event db, but instead of retrieving it from the event db for processing, we could pass it immediately to the event processor after storing, then delete after processing.

This would solve the case of having biometric authentication every 10 minutes when EAR is enabled, since that was caused by trying to decrypt the stored update event with a stale LAContext.

WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
```diff
@@ -0,0 +1,167 @@
...
+    /// Stop all syncing activities and prepare to idle.
+
+    func suspend() async throws
+
+}
+
+final class SyncManager: SyncManagerProtocol {
+
+    private(set) var syncState: SyncState = .suspended
+    private var isSuspending = false
```
netbe, line 38: why do we need this bool if the syncState already represent the suspended

WireDomain/Sources/WireDomain/Synchronization/SyncState.swift
```diff
@@ -0,0 +1,37 @@
...
+
+    case quickSync(Task<Void, Error>)
+
+    /// The app is processing live events via the push channel.
+
+    case live(Task<Void, Never>)
+
+    /// The app is neither receiving nor processing any events.
+
+    case suspended
```
netbe, line 35: suggestion: maybe idle

WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
```diff
@@ -0,0 +1,167 @@
...
+
+            for event in envelopes.flatMap(\.events) {
+                do {
+                    try await updateEventProcessor.processEvent(event)
+                } catch {
+                    WireLogger.sync.error("failed to process stored event, dropping: \(error)")
+                }
+            }
+
+            try await updateEventsRepository.deleteNextPendingEvents(limit: batchSize)
```
netbe, line 168: so here we delete what we had in the queue, if we failed with one of the event, is there a risk to remove one enveloppe which was not processed

WireDomain/Sources/WireDomain/Repositories/UpdateEventsRepository.swift
```diff
@@ -85,11 +123,23 @@ final class UpdateEventsRepository: UpdateEventsRepositoryProtocol {
...
             sinceEventID: lastEventID
         ) {
+            let batchCount = envelopes.count
+            var count = 0
+            WireLogger.sync.debug("received batch of \(batchCount) envelopes")
+
+            // If we need to abort, do it before processing the next page.
+            try Task.checkCancellation()
+
             for envelope in envelopes {
```
netbe, line 133: wording: we call it envelope but it comes from getUpdateEvents, which could be confusing, not sure what's better to stick with UpdateEvents or rename to getEnvelopes

netbe (commented):
looks good, add some comments and open to discuss it

WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
johnxnguyen: I'll defer this question to after we do some live testing.

WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
johnxnguyen: I'll defer this question to after we do some live testing.

WireDomain/Sources/WireDomain/Event Decryption/UpdateEventDecryptor.swift
johnxnguyen: Ah I see now. I was preserving the existing behavior where we only appending decryption error system messages if it's a proteus error (because I believe the error is used to determine the message content).

WireDomain/Sources/WireDomain/Helpers/AsyncSequence+ToStream.swift
johnxnguyen: It's because it's an `AsyncStream` so you can't throw. I'll make this an `AsyncThrowingStream` instead.

johnxnguyen: done 3eae8e6ecf4d04a0c301ed2ae1f1b892f12960ce

WireDomain/Sources/WireDomain/Repositories/UpdateEventsRepository.swift
johnxnguyen: done b6d207d49312b9b03ccad859d27fd9b12de6922f

WireDomain/Sources/WireDomain/Repositories/UpdateEventsRepository.swift
johnxnguyen: done b6d207d49312b9b03ccad859d27fd9b12de6922f

WireDomain/Sources/WireDomain/Repositories/UpdateEventsRepository.swift
johnxnguyen: done b6d207d49312b9b03ccad859d27fd9b12de6922f

WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
johnxnguyen: done e54a4fda11615356290288e6cf18afd532a62491

WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
johnxnguyen: I don't see a need for an actor (yet). Do you have something in mind?

WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
johnxnguyen: I guess we can also suspend if the live task fails.

johnxnguyen: done 5d2069a4e841be19a16d6281723681c9b6aa6b3c

WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
johnxnguyen: `isSuspending` is needed to prevent trying to suspend more than once. `syncState` only becomes `suspended` when the suspension completes.

WireDomain/Sources/WireDomain/Synchronization/SyncState.swift
johnxnguyen: I initially named it `idle` but felt that `suspended` more strongly suggested that the app is not live.

WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
johnxnguyen: yes, there is only one chance to process the event. I'm not sure what the strategy should be here though.. stop all event processing? skip the event?, don't delete an retry later?

johnxnguyen: ok so Android stops hard on the event, blocking the app until the error is fixed.

WireDomain/Sources/WireDomain/Repositories/UpdateEventsRepository.swift
johnxnguyen: I would prefer to rename the method `getEventEnvelopes`.

johnxnguyen: Though I'm not sure exactly which method you're referring to.

WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
johnxnguyen: It should be decrypted, then stored, then the id should be persisted. I'll do this in another PR though.

WireDomain/Sources/WireDomain/Synchronization/SyncManager.swift
johnxnguyen: I think we can align to this but I'd do it in another PR.

johnxnguyen: I created a ticket for this https://wearezeta.atlassian.net/browse/WPB-10458

agisilaos (approved):
🙌

samwyndham (approved):

PR #1717, opened 2024-07-18
fix: crash accessing selfClient in NSE - WPB-9939

netbe:
### Issue

When we setup CoreCrypto (CC), we add a safeContext on its database enclosing directory to protect multiple processes to access CC at the same time. The issue is that the directory is the shared account folder containing also the database for messages and events. The protection using flock might block the access to the CoreData stack resulting in a crash.

### Solution

* Change location of CoreCrypto database to another folder `CC` which would be protected by SafeContext.
* Add process name and id to logs to identify multiple processes in Datadog

### Note

Originally, I thought we would need to move the files to a specific folder to protect them with flock, but [moving the CC files](https://github.com/wireapp/wire-ios/pull/1717/commits/11bbe22e0efc633a47dc7caa32b31b1daf2b78a7) always ended up to corrupt the session (Error 204 or 3).

### Testing

1. Install previous build
2. send message
3. update to new build
4. check that sending / receiving message still works

### TODO

- [ ] check no more crash is produced
- [ ] flock check
---

### Checklist

### DIFF
--- a/wire-ios-data-model/Source/MLS/CoreCryptoConfiguration.swift
+++ b/wire-ios-data-model/Source/MLS/CoreCryptoConfiguration.swift
@@ -76,46 +76,6 @@ public class CoreCryptoConfigProvider {
         }
     }
 
-    public func removeOldCoreCryptoFiles(
-        sharedContainerURL: URL,
-        userID: UUID
-    ) {
-        let accountDirectory = CoreDataStack.accountDataFolder(
-            accountIdentifier: userID,
-            applicationContainer: sharedContainerURL
-        )
-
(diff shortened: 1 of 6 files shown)


johnxnguyen:
If I understood correctly, the lock is obtained per process, so shouldn't the Core Data database be accessible within the same process that has the lock?

johnxnguyen (commented):
Since we've been unable to successfully move the CC database, perhaps another fix would be to obtain the self client id before decrypting messages?

netbe:
closing in favor of other fix with selfClientId

PR #1718, opened 2024-07-18
feat: Update/move URLs for C1 and C3 WPB-9748

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9748" title="WPB-9748" target="_blank">BugWPB-9748</a>  [iOS] - C1 and C3 builds -  Links are outdated- add new links as below  , C2 not delivered yet
  </td></table>
  <br />
 

### Issue
We should provide separate URLs for C1 and C3 builds and do not show any links in Wire website. 
The solution to move all the links to **url.json** and configure to every build we need.

**Note:** in a previous implementation, we localized some links (e.g.  https://wire.com/legal or https://wire.com/nutzungsbedingungen), but we won't do that for now. Our website is changing soon and these links will be changing as well.

### Solution
Move all the links to the separate file (url.json) and configure them for C1 and C3 build.
@netbe I've update 'wire-ios-build-asset's version with the last cherry-pick commit to the release branch.

### Testing
 Test main links in C1 and C3 builds.

### Needs releases with:
- [ https://github.com/wireapp/wire-ios-build-assets/pull/115]

### Checklist


wire-ios/Wire-iOS Tests/Bundle/URLsTests.swift
```diff
@@ -0,0 +1,40 @@
...
+            let filePath = Bundle.fileURL(for: "url", with: "json"),
+            let data = try? Data(contentsOf: filePath),
+            let dictionary = try? JSONSerialization.jsonObject(with: data, options: []) as? [String: String]
+        else {
+            XCTFail("Failed to load url.json file")
+            return
+        }
+
+        URLs.CodingKeys.allCases.forEach { urlEnum in
+            XCTAssertNotNil(dictionary[urlEnum.rawValue], "Missing key: \(urlEnum.rawValue)")
```
netbe: nice

wire-ios/Wire-iOS/Sources/Helpers/Bundle/Bundle+URLs.swift
```diff
@@ -0,0 +1,93 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireSystem
+
+struct URLs: Codable {
```
netbe: name suggestion:
```suggestion
struct WireURLs: Codable {
```

netbe (commented):
Looks good, added small suggestion

wire-ios/Wire-iOS/Sources/Helpers/Bundle/Bundle+URLs.swift
```diff
@@ -0,0 +1,93 @@
...
+
+    static var shared: URLs = {
+        guard let urls = URLs(forResource: "url", withExtension: "json") else {
+            fatalError("can't find or decode url.json")
+        }
+
+        return urls
+    }()
+
+    private init?(forResource resource: String, withExtension fileExtension: String) {
```
johnxnguyen: I suggest to make this throwing, that way the caller can print the error and we can see why it failed.

wire-ios/Wire-iOS Tests/Bundle/URLsTests.swift
```diff
@@ -0,0 +1,40 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+import WireCommonComponents
+@testable import Wire
+
+final class URLsTests: XCTestCase {
+
+    func testUrlFileContainsAllKeys() {
```
johnxnguyen: Wouldn't it be better to decode the whole file?

wire-ios/Wire-iOS/Sources/Helpers/Bundle/Bundle+URLs.swift
```diff
@@ -0,0 +1,93 @@
...
+    let askSupportArticle: URL
+    let reportAbuse: URL
+    let pricingInfo: URL
+    let wireEnterpriseInfo: URL
+    let legalHoldInfo: URL
+    let guestLinksInfo: URL
+    let unreachableBackendInfo: URL
+    let federationInfo: URL
+    let mlsInfo: URL
+    let endToEndIdentityInfo: URL
```
johnxnguyen: If any of the names here change, or new keys are added, then it will render existing url config files invalid. I think it would help a lot to make an explicit comment about this, making aware to developers that if this changes, then not only should the default config change, but also all the other configs that are stored elsewhere.

wire-ios/Configuration/url.json
```diff
@@ -1,5 +1,23 @@
 {
-    "wireAppOnItunes": "https://geo.itunes.apple.com/us/app/wire/id930944768?mt=8",
+    "appOnItunes": "https://geo.itunes.apple.com/us/app/wire/id930944768?mt=8",
     "support": "https://support.wire.com",
-    "randomProfilePictureSource": "https://source.unsplash.com/800x800/?landscape"
+    "searchSupport": "https://wire.com/support/username",
```
johnxnguyen: Is this link correct? Shouldn't it be `https://support.wire.com/hc`?

wire-ios/Configuration/url.json
```diff
@@ -1,5 +1,23 @@
...
+    "askSupportArticle": "https://support.wire.com/hc/requests/new",
+    "reportAbuse": "https://support.wire.com/hc/requests/new",
+    "pricingInfo": "https://wire.com/pricing",
+    "wireEnterpriseInfo": "https://wire.com/pricing",
+    "legalHoldInfo": "https://support.wire.com/hc/articles/360002018278-What-is-legal-hold-",
+    "guestLinksInfo": "https://support.wire.com/hc/articles/360000574069-Share-a-link-with-a-person-without-a-Wire-account-to-join-a-guest-room-conversation-in-my-team",
+    "unreachableBackendInfo": "https://support.wire.com/hc/articles/9357718008093-Backend",
+    "federationInfo": "https://support.wire.com/hc/categories/4719917054365-Federation",
+    "mlsInfo": "https://support.wire.com/hc/articles/12434725011485-Messaging-Layer-Security-MLS-",
+    "endToEndIdentityInfo": "https://support.wire.com/hc/articles/9211300150685-End-to-end-identity"
```
johnxnguyen: I think it should be very clear what the purpose of each of these links is so that anyone that needs to fill it out knows what content to put. 

For example, `website` and `legal` seem self explanatory but perhaps not to everyone, or someone who doesn't have all the context. It would be better to have some additional comments:

`website`: link to the app's homepage.
`legal`: link to app's homepage for legal information, e.g terms of use and data processing addendum.
`emailAlreadyInUse`: shown when the user tries to create an account with an email that is already in use. Links to support page explaining issue and what can be done.

wire-ios/Wire-iOS Tests/Bundle/URLsTests.swift
KaterinaWire: I thought it would be clearer and faster to find which key is missing.

wire-ios/Configuration/url.json
KaterinaWire: I took this link from an old implementation and the result is that it redirects to `https://support.wire.com/hc`.
I updated the link.

wire-ios/Wire-iOS/Sources/Helpers/Bundle/Bundle+URLs.swift
KaterinaWire: added

wire-ios/Wire-iOS/Sources/Helpers/Bundle/Bundle+URLs.swift
KaterinaWire: done

wire-ios/Configuration/url.json
KaterinaWire: added

johnxnguyen (approved):

wire-ios/Configuration/url.json
```diff
@@ -1,5 +1,44 @@
...
+        "legal": "Link to app's homepage for legal information, e.g terms of use and data processing addendum.",
+        "licenseInformation": "Link to the license information page.",
+        "passwordReset": "Link to the password reset page.",
+        "askSupportArticle": "Link to the support page where the user can submit a support request for various issues.",
+        "reportAbuse": "Link to the support page where a user can report an abuse issue.",
+        "wireEnterpriseInfo": "Link to an article explaining the various features of Wire.",
+        "legalHoldInfo": "Link to an article explaining legal hold.",
+        "guestLinksInfo": "Link to an article explaining how to create, share, and revoke a group conversation link.",
+        "unreachableBackendInfo": "Shown when a user tries to send a message or create a group with users from different backends when one of the backends is not reachable. Links to a support page that explains the issue when one of the backend is offline.",
+        "federationInfo": "Links to a support page explaining what Federation is.",
```
netbe: so this is to comment what the links are. What about moving this to docC like with a markdown file instead.

Here, I found it confusing as it's not going to be parsed.

netbe: @KaterinaWire as discussed, let's move the comments to the URLs struct so we have the info there and keep the json with only the data needed.

cc @johnxnguyen

KaterinaWire: @netbe and @johnxnguyen updated in the last commit

netbe (approved):

PR #1719, opened 2024-07-18
chore: remove protocol PopoverPresenter - WPB-10853

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10853" title="WPB-10853" target="_blank">Sub-taskWPB-10853</a>  Fix view controller presentation, especially for alerts
  </td></table>
  <br />
 

### Issue

A protocol PopoverPresenter is declared which several view controllers conform to.
However, most of the times the window's root view controller was used for presentation by accessing `AppDelegate.shared.window`.

This PR
- deletes the protocol and fixes the presentation of popovers throughout the app and
- removes the sharing messages from conversations.

---

### Checklist


wire-ios/Wire-iOS Tests/ReferenceImages/ConversationImagesViewControllerTests/testForWrappedInNavigationController.1.png
caldrian, line 1: @agisilaos It's not correct that the first item has a round border, right?

wire-ios/Wire-iOS Tests/EphemeralKeyboardViewControllerTests.swift
```diff
@@ -52,7 +52,6 @@ final class EphemeralKeyboardViewControllerTests: CoreDataSnapshotTestCase {
             .withUserInterfaceStyle(.dark)
             .verify(matching: sut.prepareForSnapshots())
     }
-
```
samwyndham, line 55: I wonder if we can standardize our whitespace via SwiftLint. I feel like we use a few different styles on the team

wire-ios/Wire-iOS Tests/ReferenceImages/ConversationImagesViewControllerTests/testForWrappedInNavigationController.1.png
agisilaos: @caldrian 

In this class the snapshot tests they don't look the same as in the real app. The first and last button for the input bar look like they have 4 rounded corners instead of 2. That's because snapshot tests don't work well with maskedCorners and CI. More on the issue can be found here: https://github.com/pointfreeco/swift-snapshot-testing/issues/358#issuecomment-939854566

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/Sections/ConversationEncryptionProtocolSectionController.swift
```diff
@@ -43,6 +43,7 @@ extension ConversationEncryptionProtocolSectionController {
     }
 
     func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
-        tapAction?()
+        let cell = collectionView.cellForItem(at: indexPath)!
+        tapAction(cell)
```
agisilaos, line 47: Why do force unwrap here? Can we do something like:

```swift
if let cell = collectionView.cellForItem(at: indexPath) {
    tapAction(cell)
} else {
    // Handle the case where the cell is nil, if necessary
}
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Ephemeral.swift
```diff
@@ -77,15 +77,12 @@ extension ConversationInputBarViewController {
...
 
         if let popover = ephemeralKeyboardViewController?.popoverPresentationController,
-            let presentInView = self.parent?.view,
             let backgroundColor = ephemeralKeyboardViewController?.view.backgroundColor {
-                popover.config(from: self,
-                           pointToView: pointToView,
-                           sourceView: presentInView)
-
+            popover.sourceView = pointToView.superview!
+            popover.sourceRect = pointToView.frame.insetBy(dx: -4, dy: -4)
```
agisilaos, line 85: Same here regarding force unwrap

wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/SelfProfileViewController.swift
```diff
@@ -205,10 +205,12 @@ final class SelfProfileViewController: UIViewController {
...
-        let alertViewController = profileImagePicker.selectProfileImage()
-        alertViewController.configPopover(pointToView: profileHeaderViewController.imageView)
-
-        present(alertViewController, animated: true)
+        let alertController = profileImagePicker.selectProfileImage()
+        if let popoverPresentationController = alertController.popoverPresentationController {
+            popoverPresentationController.sourceView = profileHeaderViewController.imageView.superview!
+            popoverPresentationController.sourceRect = profileHeaderViewController.imageView.frame.insetBy(dx: -4, dy: -4)
+        }
+        present(alertController, animated: true)
```
agisilaos, line 213: same here regarding force unwrapping. 

We can do something like:

```swift
let alertController = profileImagePicker.selectProfileImage()
if let popoverPresentationController = alertController.popoverPresentationController {
    if let sourceView = profileHeaderViewController.imageView.superview {
        popoverPresentationController.sourceView = sourceView
        popoverPresentationController.sourceRect = profileHeaderViewController.imageView.frame.insetBy(dx: -4, dy: -4)
    } else {
        // Handle the case where superview is nil
        // For example, set a default sourceView or log an error
        print("Superview is nil")
    }
}
present(alertController, animated: true)
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/Sections/ConversationEncryptionProtocolSectionController.swift
caldrian: Sometimes I consider crashing hard better than failing silently and this is one example:

On tapping a cell the collection view passes the indexPath of that cell as argument to this delegate method. So at this point we know that the cell exists and that force-unwrapping is safe.

If an invalid indexPath is passed it will be an API violation and I consider failing hard the proper consequence so that the issue is detected early, before the app is shipped.

agisilaos (commented):
Left some comments regarding force unwrapping. Other than that it looks good. 🙌

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+ConversationMessageCellDelegate.swift
```diff
@@ -60,8 +60,7 @@ extension ConversationContentViewController: ConversationMessageCellDelegate {
         let actionView = view.targetView(for: message, dataSource: dataSource)
 
         // Do not dismiss Modal for forward since share VC is present in a popover
```
samwyndham: Maybe that comment no longer makes sense?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+MessageAction.swift
```diff
@@ -98,24 +98,22 @@ extension ConversationContentViewController {
         case .present:
             dataSource.selectedMessage = message
             presentDetails(for: message)
+
         case .save:
             if Message.isImage(message) {
                 saveImage(from: message, view: view)
-            } else {
+            } else if let fileURL = message.fileMessageData?.temporaryURLToDecryptedFile() {
```
samwyndham, line 105: Should we handle the `else` only case with an assertion of logging?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Ephemeral.swift
caldrian: Also here I would consider the force-unwrapping safe, but I'm curious about other opinions.

I'm not sure if any edge case like animation/transition or whatever can result to the view's `superview` property being `nil` at the moment when this code is executed, but usually the button has a superview, since it's part of the view hierarchy.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+MessageAction.swift
```diff
@@ -98,24 +98,22 @@ extension ConversationContentViewController {
...
-                } else {
-                    targetView = view
-                }
-
-                if let saveController = UIActivityViewController(message: message, from: targetView) {
-                    present(saveController, animated: true)
+                let activityViewController = UIActivityViewController(activityItems: [fileURL], applicationActivities: nil)
+                if let popoverPresentationController = activityViewController.popoverPresentationController {
+                    let sourceView = (view as? SelectableView)?.selectionView ?? view
+                    popoverPresentationController.sourceView = sourceView.superview
```
samwyndham: This pattern is becoming very common. I wonder if its worth extending `UIPopoverPresentationController`:

```swift
extension UIPopoverPresentationController {
  func setSource(view: UIView, frame: CGRect) {
    // ..
  }
}

So that the call site in this case becomes:

```swift
let sourceView = (view as? SelectableView)?.selectionView ?? view
activityViewController.popoverPresentationController?.setSource(view: sourceView, frame: sourceView.frame)
```

Maybe it's better how it is currently as it is more explicit. Just something to consider.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/Sections/ConversationEncryptionProtocolSectionController.swift
```diff
@@ -43,6 +43,7 @@ extension ConversationEncryptionProtocolSectionController {
     }
 
     func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
-        tapAction?()
+        let cell = collectionView.cellForItem(at: indexPath)!
```
samwyndham, line 46: I like the confidence with the `!` 😄

samwyndham (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/Sections/ConversationEncryptionProtocolSectionController.swift
samwyndham: I personally agree with @caldrian here. I think in some cases it is fine to force unwrap or fatal error so we actually know something is going wrong rather than the issue going unnoticed. 

In this case, I was also thinking about it before I noticed @agisilaos comment. I came to the conclusion that I can't imagine that the cell doesn't exist when this method is called.

agisilaos (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+MessageAction.swift
caldrian: Good point, thanks!
I started a refactoring like this: 73e28aef0c75621b55ff39c47e2510658766d2b6

wire-ios/Wire-iOS Tests/EphemeralKeyboardViewControllerTests.swift
caldrian: I tried to configure SwiftLint and SwiftFormat, however, I couldn't find any option for that.
We should discuss it and agree on one way.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+MessageAction.swift
caldrian: Good idea!

PR #1721, opened 2024-07-18
chore: update pull request template - no ticket

KaterinaWire:
### Issue

This PR contains an additional checklist of UI changes specifically for accessibility. Accessibility changes require more effort to test and are hidden from the average user, so they are sometimes forgotten about.

The checklist will help you remember the accessibility support for UI changes and reduce the number of comments during the review. More info about these things are in Confluence. @agisilaos and I will clean up and update the documentation.

### Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -32,5 +32,5 @@ _Optional: attachments like images, videos, etc._
 
 _If your PR includes UI changes, please utilize this checklist:_
 - [ ] Make sure you use the API for UI elements that support large fonts.
-- [ ] All colors are taken from Semantic Colors.
+- [ ] All colors are taken from WireDesign.ColorTheme or constructed using WireDesign.BaseColorPalette.
 - [ ] New UI elements have Accessibility strings for VoiceOver.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -26,3 +26,11 @@ _Optional: attachments like images, videos, etc._
 - [ ] Description is filled and free of optional paragraphs.
 - [ ] Adds/updates automated tests.
 
+---
+
+### UI accessibility checklist
+
+_If your PR includes UI changes, please utilize this checklist:_
+- [ ] Make sure you use the API for UI elements that support large fonts.
+- [ ] All colors are taken from Semantic Colors.
+- [ ] New UI elements have Accessibility strings for VoiceOver.


agisilaos (approved):

.github/pull_request_template.md
```diff
@@ -26,3 +26,11 @@ _Optional: attachments like images, videos, etc._
 - [ ] Description is filled and free of optional paragraphs.
 - [ ] Adds/updates automated tests.
 
+---
+
+### UI accessibility checklist
+
+_If your PR includes UI changes, please utilize this checklist:_
+- [ ] Make sure you use the API for UI elements that support large fonts.
+- [ ] All colors are taken from Semantic Colors.
```
caldrian: We should discuss if we want to stick to the current SemanticColors type.
In new designs the designer specifies the color theme:
Wire 2024-07-18 at 13-37-55
For this example you would specify: 
```
let color = WireDesign.ColorTheme.Backgrounds.onSurface
```

caldrian (approved):

.github/pull_request_template.md
johnxnguyen: I like this way, rather than colors named by us.

johnxnguyen (approved):

.github/pull_request_template.md
KaterinaWire: @caldrian  I agree that we should move to a new color system since that's what we have from the designers. Does it sound good?
`All colors are taken from WireDesign.ColorTheme`

.github/pull_request_template.md
```diff
@@ -26,3 +26,11 @@ _Optional: attachments like images, videos, etc._
...
 - [ ] Adds/updates automated tests.
 
+---
+
+### UI accessibility checklist
+
+_If your PR includes UI changes, please utilize this checklist:_
+- [ ] Make sure you use the API for UI elements that support large fonts.
+- [ ] All colors are taken from WireDesign.ColorTheme or constructed using WireDesign.BaseColorPalette.
+- [ ] New UI elements have Accessibility strings for VoiceOver.
```
caldrian, line 36: Btw, what about putting the check already?
Most of the times the check should be there so it would save some time.
Or do we prefer it this way?

```suggestion
- [x] Make sure you use the API for UI elements that support large fonts.
- [x] All colors are taken from WireDesign.ColorTheme or constructed using WireDesign.BaseColorPalette.
- [x] New UI elements have Accessibility strings for VoiceOver.
```

caldrian (approved):

.github/pull_request_template.md
agisilaos: @caldrian I think for people who're aware it could be faster BUT it could lead also to situations where it's already ticked and people don't have to think about it, in other words consider it done already.

KaterinaWire: I agree with @agisilaos . I wouldn't put the check by default so as not to ignore it.

PR #1724, opened 2024-07-18
chore: replace protocol SpinnerCapable by class BlockingActivityIndicator - WPB-7306

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7306" title="WPB-7306" target="_blank">StoryWPB-7306</a>  [iOS] : Add “Conversation” , “Archive” , “Settings”
  </td></table>
  <br />
 

### Issue

This PR introduces a type `BlockingActivityIndicator` and removes the `SpinnerCapable` protocol.

---

### Checklist


WireUI/Sources/WireReusableUIComponents/BlockingActivityIndicator/BlockingActivityIndicator.swift
```diff
@@ -0,0 +1,180 @@
...
+        Task {
+            await MainActor.run { [weak view] in
+                view?.unblockAndStopAnimatingIfNeeded(blockingActivityIndicator: nil)
+            }
+        }
+    }
+
+    // MARK: - Methods
+
+    @MainActor
```
samwyndham, line 51: Perhaps make the whole class `@MainActor`, if you think that makes sense

WireUI/Sources/WireReusableUIComponents/BlockingActivityIndicator/BlockingActivityIndicator.swift
```diff
@@ -0,0 +1,180 @@
...
+
+        state.weakReferences = state.weakReferences.filter { $0.reference != nil && $0.reference !== reference }
+        if state.weakReferences.isEmpty {
+            state.activityIndicatorView.superview!.removeFromSuperview()
+            blockingActivityIndicatorState = nil
+        }
+    }
+
+    private var blockingActivityIndicatorState: BlockingActivityIndicatorState? {
+        get { objc_getAssociatedObject(self, &stateKey) as? BlockingActivityIndicatorState }
```
samwyndham, line 136: Sneaky 🧛

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/ConversationGuestOptionsViewController.swift
```diff
@@ -138,10 +138,13 @@ final class ConversationGuestOptionsViewController: UIViewController,
         _ viewModel: ConversationGuestOptionsViewModel,
         didUpdateState state: ConversationGuestOptionsViewModel.State
     ) {
-        tableView.reloadData()
-
-        (navigationController as? SpinnerCapableViewController)?.isLoadingViewVisible = state.isLoading
+        if state.isLoading {
+            activityIndicator.start()
```
samwyndham: Considering how `BlockingActivityIndicator` is used in practice I'm wondering if its API should change to something like the following as it feels like the direction we are moving in swift is reacting to state changes and the current API is less declarative.

```swift
activityIndicator.update(isActive: state.isLoading)
```

samwyndham (approved):
Nice work!

WireUI/Sources/WireReusableUIComponents/BlockingActivityIndicator/BlockingActivityIndicator.swift
caldrian: I had it before, but then a lot more places in the code base need to be adjusted.
Since `init` doesn't need to run on the main actor and `deinit` starts a task, this way I actually like it more.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation Options/ConversationGuestOptionsViewController.swift
caldrian: Very good point, thanks!
Added a method: 4a9448162c3119cb1d710abc26e40e6ebe655cc1

agisilaos (approved):

PR #1725, opened 2024-07-19
fix: duplicate messages - WPB-10251

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10251" title="WPB-10251" target="_blank">BugWPB-10251</a>  [iOS] Duplicate messages in Wire Beta App
  </td></table>
  <br />
 

### Issue

From the logs, we observe that messages are processed twice, one thing we can do to avoid this is to make sure we don't have the same event stored twice in the database.

Solution: check if event is present already in the events database before storing it. We check its hash from the eventId and payload.

~Note: in the future refactoring a database constraint will be added to avoid it instead.~

* refactor and add migration code of event database
* store hash of event id and payload in a new attribute of StoredUpdateEvent entity

---

### Checklist


agisilaos (approved):
Nice! 🎉

wire-ios-request-strategy/Sources/Synchronization/Decoding/StoreUpdateEvent.swift
```diff
@@ -95,6 +101,34 @@ public final class StoredUpdateEvent: NSManagedObject {
...
+        let storedEvents = StoredUpdateEvent.events(eventId: eventId, context: context)
+        guard !storedEvents.isEmpty else {
+            return false
+        }
+
+        let upateEventHash = EventHasher(eventId: eventId, payload: updateEvent.payload as NSDictionary).hashValue
+
+        for storedEvent in storedEvents {
+            let storedEventHash = EventHasher(storedEvent: storedEvent)?.hashValue
+            return storedEventHash == upateEventHash
```
johnxnguyen: This would return in the first iteration. Perhaps you mean something like:

```swift
return storedEvents.contains {
    storedEventHash == EventHasher(storedEvent: storedEvent)?.hashValue
}
```

wire-ios-request-strategy/Sources/Synchronization/Decoding/StoreUpdateEventTests.swift
```diff
@@ -271,6 +271,41 @@ class StoreUpdateEventTests: MessagingTestBase {
         }
     }
 
+    func test_EncryptAndCreate_DoesNotStoreDuplicateEvents() throws {
+        eventMOC.performAndWait {
+            // Given some events.
+            let conversation = self.createConversation(in: self.uiMOC)
+            let event1 = self.createNewConversationEvent(for: conversation)
```
johnxnguyen: Probably good to insert another event with the same event to account for the possibility that we stored multiple events from the same envelope.

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventHasher.swift
```diff
@@ -0,0 +1,38 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// Computes an hash to compare UpdateEvent and StoredUpdateEvent
+struct EventHasher: Hashable {
```
caldrian: The name is confusing in my opinion, since this is just a struct which keeps values/references.
It doesn't do hashing.

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventHasher.swift
```diff
@@ -0,0 +1,38 @@
...
+    var eventId: String
+    var payload: NSDictionary
```
caldrian: NSDictionary does not implement `Hashable` properly, in my opinion. Isn't it dangerous? 
See my comment in the tests.

wire-ios-request-strategy/Sources/Synchronization/Decoding/StoreUpdateEventTests.swift
```diff
@@ -271,6 +271,41 @@ class StoreUpdateEventTests: MessagingTestBase {
...
+            ) else {
+                return XCTFail("Did not create storedEvent")
+            }
+
+            let storedEvent2 = StoredUpdateEvent.encryptAndCreate(
+                event1,
+                context: self.eventMOC,
+                index: 2,
+                publicKeys: nil
+            )
```
caldrian: This is probably the same as @johnxnguyen meant.

By creating a second event in a different NSDictionary instance with the same content you will see that the test fails, because NSDictionary doesn't conform to `Hashable` properly, in my opinion.

```suggestion
            let event1 = self.createNewConversationEvent(for: conversation)
            let event2 = self.createNewConversationEvent(for: conversation) // <---

            // When we store the events with public keys.
            guard let storedEvent1 = StoredUpdateEvent.encryptAndCreate(
                event1,
                context: self.eventMOC,
                index: 1,
                publicKeys: nil
            ) else {
                return XCTFail("Did not create storedEvent")
            }

            let storedEvent2 = StoredUpdateEvent.encryptAndCreate(
                event2, // <---
                context: self.eventMOC,
                index: 2,
                publicKeys: nil
            )
```

wire-ios-request-strategy/Sources/Synchronization/Decoding/StoreUpdateEvent.swift
netbe: changed it

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventHasher.swift
netbe: any suggestion? it does provide hashing via Hashable

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventHasher.swift
netbe: yep, i noticed / recalled that, I changed to use NSKeyedArchiver

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataEventsMigrationVersion.swift
```diff
@@ -0,0 +1,68 @@
...
+    static let current: Self = {
+        guard let current = allCases.first else {
+            fatalError("no model versions found")
+        }
+        return current
+    }()
+
+    // MARK: Store URL
+
+    func managedObjectModelURL() -> URL? {
```
samwyndham, line 64: Is there any situation where this can return `nil` and the app still functions? I think we should fatal error here if the there is no bundle URL.

wire-ios-request-strategy/Sources/Synchronization/Decoding/StoreUpdateEvent.swift
```diff
@@ -95,6 +112,27 @@ public final class StoredUpdateEvent: NSManagedObject {
...
     }
 
+    static private func storedEventExists(for updateEventHash: Int, in context: NSManagedObjectContext) -> Bool {
+
+        let storedEvents = StoredUpdateEvent.events(eventHash: updateEventHash, context: context)
+        guard !storedEvents.isEmpty else {
+            return false
+        }
+        
+        for storedEvent in storedEvents where storedEvent.eventHash == updateEventHash || storedEvent.eventHash == 0 {
```
netbe: note that if the storedEvent has no hash (=0), we fallback to just match by event id, in case we just upgraded to this version and had storedEvents from previous fetch in the database). @johnxnguyen does that make sense to you?

wire-ios-request-strategy/Sources/Synchronization/Decoding/StoreUpdateEventTests.swift
samwyndham, line 1: Do we have a test somewhere that checks that have a `CoreDataEventsMigrationVersion` for all data models... e.g. that we didn't forget to append the new version to `CoreDataEventsMigrationVersion`?

wire-ios-request-strategy/Sources/Synchronization/Decoding/StoreUpdateEvent.swift
johnxnguyen: Yes it makes sense.

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMessagingMigrationStep.swift
```diff
@@ -18,10 +18,10 @@
 
 import CoreData
 
-struct CoreDataMessagingMigrationStep {
+struct CoreDataMessagingMigrationStep<Version: CoreDataMigrationVersion> {
```
johnxnguyen: Perhaps we can rename this `CoreDataMigrationStep`.

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMessagingMigrator.swift
```diff
@@ -18,13 +18,21 @@
 
 import CoreData
 
+protocol CoreDataMigratorProtocol {
+    associatedtype MigrationVersion
```
johnxnguyen: nitpick for the naming: it would be more accurate to call this `DatabaseVersion`.

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMessagingMigrator.swift
```diff
@@ -60,9 +68,11 @@ extension CoreDataMessagingMigratorError: LocalizedError {
     }
 }
 
-final class CoreDataMessagingMigrator: CoreDataMessagingMigratorProtocol {
+extension CoreDataMessagingMigrator<CoreDataMessagingMigrationVersion>: CoreDataMessagingMigratorProtocol {}
```
johnxnguyen: Why do we still have `CoreDataMessagingMigratorProtocol`?

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMessagingMigrator.swift
```diff
@@ -73,17 +83,16 @@ final class CoreDataMessagingMigrator: CoreDataMessagingMigratorProtocol {
...
-    func migrateStore(at storeURL: URL, toVersion version: CoreDataMessagingMigrationVersion) throws {
-        zmLog.safePublic(
-            "migrateStore at: \(SanitizedString(stringLiteral: storeURL.absoluteString)) to version: \(SanitizedString(stringLiteral: version.rawValue))",
-            level: .info
+    func migrateStore(at storeURL: URL, toVersion version: Version) throws {
+        WireLogger.localStorage.info(
+            "migrateStore at: \(SanitizedString(stringLiteral: storeURL.absoluteString)) to version: \(SanitizedString(stringLiteral: version.rawValue))", attributes: .safePublic
         )
 
         try forceWALCheckpointingForStore(at: storeURL)
```
johnxnguyen, line 88: WAL is only enabled for the message DB, not the event DB. I'm not sure if this would be problematic.

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMigrationActionFactory.swift
```diff
@@ -20,6 +20,34 @@ import Foundation
...
 
+    static func createPreMigrationAction<Version: CoreDataMigrationVersion>(for destinationVersion: Version) -> CoreDataMigrationAction? {
+
+        if let version = destinationVersion as? CoreDataMessagingMigrationVersion {
+            return createPreMigrationAction(for: version)
+        }
+
+        if let version = destinationVersion as? CoreDataEventsMigrationVersion {
+            return createPreMigrationAction(for: version)
+        }
```
johnxnguyen, line 31: We should probably refactor this in a way that doesn't rely on casting, but I think it's fine for now.

wire-ios-request-strategy/Sources/Synchronization/Decoding/StoreUpdateEvent.swift
```diff
@@ -95,6 +112,27 @@ public final class StoredUpdateEvent: NSManagedObject {
...
     }
 
+    static private func storedEventExists(for updateEventHash: Int, in context: NSManagedObjectContext) -> Bool {
+
+        let storedEvents = StoredUpdateEvent.events(eventHash: updateEventHash, context: context)
+        guard !storedEvents.isEmpty else {
+            return false
+        }
+
+        for storedEvent in storedEvents where storedEvent.eventHash == updateEventHash || storedEvent.eventHash == 0 {
```
samwyndham: I don't quite understand the need for this check. Are there cases where storedEvents > 0 but we should return false? I was assuming we could just use [countForFetchRequest](https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506868-countforfetchrequest) here but I must be missing something.

wire-ios-request-strategy/Sources/Synchronization/Decoding/StoreUpdateEvent.swift
```diff
@@ -95,6 +112,27 @@ public final class StoredUpdateEvent: NSManagedObject {
         return storedEvent
     }
 
+    static private func storedEventExists(for updateEventHash: Int, in context: NSManagedObjectContext) -> Bool {
```
johnxnguyen: You already when through the trouble of adding event db migrations, perhaps we could go the extra mile and also add a uniqueness constraint on the hash? We could give existing stored events a random hash.

wire-ios-request-strategy/Sources/Synchronization/Decoding/StoreUpdateEvent.swift
johnxnguyen: Why do we have the `where` clause here?  Don't all events in `storedEvents` already have the same event hash?

netbe: right, I forget about updating the fetch query, https://github.com/wireapp/wire-ios/pull/1725/commits/7c8a940c85a45b5d3d60d2b8b9cff3154ec681cc

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataEventsMigrationVersion.swift
netbe: it could happen, if you run the app on an incompatible version db, we would catch this and present an alert to the user to clear the database

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMessagingMigrationStep.swift
netbe: done

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMessagingMigrator.swift
netbe: we need it for importBackup tests, I kept it as it is easier to keep the conformance and be able to inject a mock in the tests for backup for now

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMigrationActionFactory.swift
netbe: yeah, I couldn't figure it out

wire-ios-request-strategy/Sources/Synchronization/Decoding/StoreUpdateEventTests.swift
netbe: added similar check as we had for messagingStore

wire-ios-data-model/Tests/Source/ManagedObjectContext/DatabaseMigrationTests.swift
```diff
@@ -34,6 +34,21 @@ final class DatabaseMigrationTests: DatabaseBaseTest {
         "handle"
     ]
 
+    func testEventsLatestModelHasMigrationVersion() throws {
+        // given
+        let latestMigrationVersion = CoreDataEventsMigrationVersion.allCases.first
```
netbe, line 39: @samwyndham is this the kind of test you were looking for ?

wire-ios-request-strategy/Sources/Synchronization/Decoding/StoreUpdateEventTests.swift
netbe: done in another test

wire-ios-request-strategy/Sources/Synchronization/Decoding/StoreUpdateEvent.swift
netbe: good idea, i replace it with a fetchRequest directly

wire-ios-data-model/Tests/Source/ManagedObjectContext/DatabaseMigrationTests.swift
samwyndham: Looks good!

wire-ios-request-strategy/Sources/Synchronization/Decoding/StoreUpdateEvent.swift
netbe: alright so there was still more changes but I think it's done now :)

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMessagingMigrator.swift
netbe: I haven't seen an issue running it

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationEventProcessor.swift
```diff
@@ -57,6 +57,7 @@ public class ConversationEventProcessor: NSObject, ConversationEventProcessorPro
     // MARK: - Methods
 
     public func processPayload(_ payload: ZMTransportData) {
+        // TODO: [F] should add UUID?
```
johnxnguyen: Assuming this is from some response and not from pulling events or getting it through the push channel, then it's probably fine to init a uuid.

wire-ios-sync-engine/Source/Data Model/ServiceUser.swift
```diff
@@ -317,7 +317,8 @@ public extension ZMConversation {
             guard response.httpStatus == 201,
                   let responseDictionary = response.payload?.asDictionary(),
                   let userAddEventPayload = responseDictionary["event"] as? ZMTransportData,
-                  let event = ZMUpdateEvent(fromEventStreamPayload: userAddEventPayload, uuid: nil) else {
+                  // TODO: [F] check if adding UUID here is ok
```
johnxnguyen: I think it's ok to add the UUID here.

wire-ios-sync-engine/Source/Data Model/Conversation+MessageDestructionTimeout.swift
```diff
@@ -42,7 +42,8 @@ extension ZMTransportResponse {
         guard let payload else {
             return nil
         }
-        return ZMUpdateEvent(fromEventStreamPayload: payload, uuid: nil)
+        // TODO: [F] check if adding an UUID is ok
```
johnxnguyen: I think it's ok to add the UUID here.

wire-ios-sync-engine/Source/Data Model/Conversation+Join.swift
```diff
@@ -74,7 +74,8 @@ extension ZMConversation {
             switch response.httpStatus {
             case 200:
                 guard let payload = response.payload,
-                      let event = ZMUpdateEvent(fromEventStreamPayload: payload, uuid: nil),
+                      // TODO: [F] check if adding UUID is ok
```
johnxnguyen: I think it's ok to add the UUID here.

wire-ios-sync-engine/Source/Data Model/Conversation+AccessMode.swift
```diff
@@ -238,7 +238,8 @@ extension ZMConversation {
 
         request.add(ZMCompletionHandler(on: managedObjectContext!) { response in
             if let payload = response.payload,
-               let event = ZMUpdateEvent(fromEventStreamPayload: payload, uuid: nil) {
+               // TODO: [F] check if addind UUID is ok
```
johnxnguyen: I think it's ok to add the UUID here.

johnxnguyen (approved):

PR #1732, opened 2024-07-22
fix: crash accessing self client id - WPB-9939

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9939" title="WPB-9939" target="_blank">BugWPB-9939</a>  Crash accessing selfUser in NSE - Beta
  </td></table>
  <br />
 

### Issue

There is a crash when accessing the self client in the notification service extension. It's not clear by exactly what is causing the crash. The top of the stack trace is:

```
0
libobjc.A.dylib
object_getMethodImplementation
1
CoreFoundation
_NSIsNSSet
2
CoreFoundation
-[NSMutableSet unionSet:]
3
CoreData
-[_NSFaultingMutableSet willReadWithContents:]
4
CoreData
-[_NSFaultingMutableSet allObjects]
5
WireDataModel
-[ZMUser selfClient] (ZMUser.m:221)
6
WireRequestStrategy
-[ZMSimpleListRequestPaginator requestForSingleRequestSync:apiVersion:] (ZMSimpleListRequestPaginator.m:97)
```

`object_getMethodImplementation` is causing a segmentation fault which is likely because the object it references has been released. There are a couple other threads that appear to wait on a file lock for safe core crypto access. It is possible that this could be interfering with Core Data because the file lock is on the account directory which encloses both the Core Crypto database and the Core Data database.

We are trying to see if we can move the Core Crypto database so the file lock doesn't include the Core Data database, but it's not so easy.

In the meantime however, I thought that one way we could avoid this particular crash is to avoid accessing the self client when trying to fetch events, and instead inject the self client id (just a string) where needed. I don't know if this will fix the crash, so we'll need to monitor crashes and see if it makes a difference.

### Testing

Use the app and check for crashes.

---

### Checklist

---

### DIFF
--- a/wire-ios-request-strategy/Sources/Notifications/NotificationStreamSync.swift
+++ b/wire-ios-request-strategy/Sources/Notifications/NotificationStreamSync.swift
@@ -41,8 +41,11 @@ public class NotificationStreamSync: NSObject, ZMRequestGenerator, ZMSimpleListR
         super.init()
         managedObjectContext = moc
 
-        let selfUser = ZMUser.selfUser(in: moc)
-        let selfClientID = selfUser.selfClient()?.remoteIdentifier
+        var selfClientID: String?
+        moc.performAndWait {
+            let selfUser = ZMUser.selfUser(in: moc)
+            selfClientID = selfUser.selfClient()?.remoteIdentifier
+        }
 
         listPaginator = ZMSimpleListRequestPaginator(
             basePath: "/notifications",
--- a/wire-ios-request-strategy/Sources/Notifications/NotificationStreamSync.swift
+++ b/wire-ios-request-strategy/Sources/Notifications/NotificationStreamSync.swift
@@ -30,6 +30,7 @@ public class NotificationStreamSync: NSObject, ZMRequestGenerator, ZMSimpleListR
     private var managedObjectContext: NSManagedObjectContext!
     private let lastEventIDRepository: LastEventIDRepositoryInterface
     private weak var notificationStreamSyncDelegate: NotificationStreamSyncDelegate?
+    private var clientID: String?
 
     public init(
         moc: NSManagedObjectContext,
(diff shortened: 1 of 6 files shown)


wire-ios-request-strategy/Sources/Notifications/NotificationStreamSync.swift
```diff
@@ -40,12 +40,19 @@ public class NotificationStreamSync: NSObject, ZMRequestGenerator, ZMSimpleListR
...
-        listPaginator = ZMSimpleListRequestPaginator.init(basePath: "/notifications",
-                                                          startKey: "since",
-                                                          pageSize: 500,
-                                                          managedObjectContext: moc,
-                                                          includeClientID: true,
-                                                          transcoder: self)
+
+        let selfUser = ZMUser.selfUser(in: moc)
+        let selfClientID = selfUser.selfClient()?.remoteIdentifier
+
```
netbe: are we sure it's executed on the moc context here ?

netbe (commented):
just left a comment, rest looks good!

wire-ios-request-strategy/Sources/Notifications/NotificationStreamSync.swift
netbe: for me it's executed on the thread from :
```
   public override func didReceive(
        _ request: UNNotificationRequest,
        withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void
```

johnxnguyen: Yep I think you're right, fixed 366b05e4445c45107cc31d79ca1defb49e86173f

netbe (approved):

samwyndham (approved):

PR #1738, opened 2024-07-23
fix: Make `BackendInfo` dependent tests in WireSyncEngine more robust - WPB-10065

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10065" title="WPB-10065" target="_blank">TaskWPB-10065</a>  iOS: fix unit tests without backend info
  </td></table>
  <br />
 

### Issue

This PR improves test reliability in `BackendInfo` dependent tests in the `WireSynceEngine` test targets. It will be one of a many PRs that will replace https://github.com/wireapp/wire-ios/pull/1689. I'm replacing that PR as my implementation has somewhat changed.

This PR only fixes tests in `WireSyncEngine`. I want to get feedback on the strategy before implementing throughout the code base.

#### The problem

`BackendInfo` is an implicit dependency all over our codebase. Mutating it's value in one test can effect the results of later tests. Currently we are manually updating and resetting its values throughout the codebase but there are many places where this is not correct causing test failures when tests run in unexpected order or when individual tests are run locally.

To address this, this PR:

- Sets up `BackendInfo.storage` with some default values in `TestSetup`. `TestSetup` is a [NSPrincipleClass](https://developer.apple.com/documentation/xctest/xctestobservationcenter) which means it is instantiated before any tests are run. Furthermore, for the storage it uses `TestUserDefaults` and configures any test to fail which attempts to mutate `BackendInfo.storage`. This has proven really useful in finding tests which have a hidden dependency on `BackendInfo`. If a test needs to mutate a value it must to so in the mock storage. See next point.
- Updates `BackendInfo` to have both `storage` & `mockStorage`. This allows enabling mocking through `static BackendInfo.enableMocking()` and then returning to the previous state by deleting the mock storage in `static BackendInfo.resetMocking()`

Initially we tried to address the problem by keeping `BackendInfo` empty by default. However I found that because `BackendInfo` is implicitly used all over the place, we also needed to enable mocking and configure it throughout the codebase in unexpected places. The solution here, minimizes this boilerplate and tries to add some safety by failing tests which mutate `BackendInfo` unexpectedly.

#### Questionable decisions

- We mix some testing code into `BackendInfo`. This at least is within `if #DEBUG` so it won't be in the production builds but it is certainly not the most pretty of solutions.
- I subclass `UserDefaults` to provide the ability to fail tests when the storage is unexpectedly mutated. Perhaps we could use protocols + mocks instead. There is however some complexity to mocking `UserDefaults` - it is likely a mock would need to encompass some of its behavior like how an object saved with `NSUserDefaults.set(_:forKey:)` is returned with `NSUserDefaults.bool(forKey:)` - i.e it would return `false` if not set. 
- Nothing stops a test from switching `BackendInfo.storage` to a new value. The tests would not pick up such a situation.

### Testing

This PR principally updates unit tests so run them to see they are working.

### Checklist

---


wire-ios-sync-engine/Tests/Source/Integration/IntegrationTest+BackendInfo.swift
```diff
@@ -0,0 +1,46 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+@_spi(MockBackendInfo)
+import WireTransport
+
+@objc
+extension IntegrationTest {
```
samwyndham, line 23: This exists to allow objective-c code to call BackendInfo mocking functions

caldrian (approved):

agisilaos (approved):

PR #1742, opened 2024-07-23
feat: use `useSFTForOneToOneCalls` for 1:1 calls via SFT - WPB-7152

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7152" title="WPB-7152" target="_blank">StoryWPB-7152</a>  [iOS] Add flag for customer specific 1:1 calls in federated environment
  </td></table>
  <br />
 

### Issue
`conferenceCalling` feature flag will have a config part in **v6** that we should store:
```
conferenceCalling": {
		"config": {
			"useSFTForOneToOneCalls": false
		}
}
```
If `useSFTForOneToOneCalls` set to `true`, clients will use SFT for 1:1 calls (same as conference or mls conference).

This PR also contains a new version of AVS as it contains a fix for this task.

### Testing

1. Setup `conferenceCalling` feature flag, so `useSFTForOneToOneCalls` is true.
2. Make a 1:1 call between users from different backends.

**Verify:**
- call is establishedl
- call is via SFT

### Checklist

---


wire-ios-data-model/Source/Repositories/FeatureRepository.swift
```diff
@@ -110,12 +110,25 @@ public class FeatureRepository: FeatureRepositoryInterface {
...
+            config = try decoder.decode(Feature.ConferenceCalling.Config.self, from: featureConfig)
+        } catch {
+            logger.error("failed to decode Feature.ConferenceCalling.Config: \(error)")
+        }
+
+        return .init(status: feature.status, config: config)
     }
 
     public func storeConferenceCalling(_ conferenceCalling: Feature.ConferenceCalling) {
+        let config = try? encoder.encode(conferenceCalling.config)
```
johnxnguyen: Should we only try to encode if there is a config?

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
```diff
@@ -1079,4 +1034,21 @@ private extension ZMConversation {
...
+            return .oneToOne
+        }
+
+        switch messageProtocol {
+        case .mls:
+            return .mlsConference
+        case .proteus, .mixed:
+            return .conference
+        }
+    }
```
johnxnguyen: I feel like this should be a function `(ZMConversation) -> AvSConversationType`. Could we move this to the call center?

wire-ios-data-model/Tests/Source/Model/FeatureConfiguration/FeatureRepositoryTests.swift
```diff
@@ -327,6 +327,42 @@ class FeatureRepositoryTests: ZMBaseManagedObjectTest {
...
+                XCTFail("failed to decode config data")
+                return
+            }
+
+            XCTAssertEqual(feature.status, .disabled)
+            XCTAssertEqual(featureConfig, config)
+        }
+
+        XCTAssertTrue(waitForAllGroupsToBeEmpty(withTimeout: 0.5))
+    }
```
David-Henner: I think you can use async / await here
```suggestion
    func testThatItStoresConferenceCalling_V6() async {
        await syncMOC.perform {
            // Given
            let sut = FeatureRepository(context: self.syncMOC)
            let config = Feature.ConferenceCalling.Config(
                useSFTForOneToOneCalls: true
            )
            let conferenceCalling = Feature.ConferenceCalling(status: .disabled, config: config)
            self.assertFeatureDoesNotExist(name: .conferenceCalling)

            // When
            sut.storeConferenceCalling(conferenceCalling)

            // Then
            guard let feature = Feature.fetch(name: .conferenceCalling, context: self.syncMOC) else {
                XCTFail("feature not found")
                return
            }

            guard let configData = feature.config else {
                XCTFail("expected config data")
                return
            }

            guard let featureConfig = configData.decode(as: Feature.ConferenceCalling.Config.self) else {
                XCTFail("failed to decode config data")
                return
            }

            XCTAssertEqual(feature.status, .disabled)
            XCTAssertEqual(featureConfig, config)
        }
    }
```

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
```diff
@@ -960,51 +960,6 @@ extension WireCallCenterV3 {
...
-                    parentGroupID: parentGroupID,
-                    subconversationGroupID: subgroupID
-                )
-
-                self.avsWrapper.setMLSConferenceInfo(conversationId: conversationId, info: conferenceInfo)
-            } catch {
-                WireLogger.mls.error("error while setMLSConferenceInfo: \(error.localizedDescription)")
-            }
-        }
-    }
```
David-Henner, line 1006: why has this been removed?

KaterinaWire: I think this is dead code. We don't use it. Do you have any concerns?

wire-ios-data-model/Source/Repositories/FeatureRepository.swift
KaterinaWire: We don't have a config for older versions, so I've made it optional. Do you suggest making separate functions for different versions?

KaterinaWire: Done

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
KaterinaWire: done

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
David-Henner: Ok 👍 Looks fine to remove

David-Henner (approved):

johnxnguyen (approved):

PR #1744, opened 2024-07-23
fix: calling performance - network quality - WPB-9468

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9468" title="WPB-9468" target="_blank">TaskWPB-9468</a>  [iOS] Reduce calling UI refreshes due to network quality updates
  </td></table>
  <br />
 

### Issue

This PR is part of an effort to improve calling performance by reducing the weight on the main thread due to handling events from AVS.

Here we aim to reduce the frequency of UI updates caused by the network quality change handler: `handleNetworkQualityChange` 

Changes:
- Removed the update of the network quality for 1:1 calls, because we don't need it.
- Deleted a bunch of dead code
- Guarded against updating the network quality if its value didn't change
- Added a network quality observer to `CallGridViewController` to decouple the network quality updates from the "configuration" updates which usually update the whole grid in a non-targeted fashion

### Checklist

---


agisilaos (approved):
Great work, Thanks David! 🙌

wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallViewController.swift
```diff
@@ -534,12 +538,6 @@ extension CallViewController: ConstantBitRateAudioObserver {
 
 }
 
-extension CallViewController: NetworkQualityObserver {
```
samwyndham, line 537: Just to check my understanding, we are removing the observer here because the VC if for 1:1 class right?

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
```diff
@@ -445,14 +445,6 @@ extension WireCallCenterV3 {
         }
     }
 
-    /// Call this method when the network quality of a participant changes and avs calls the `wcall_network_quality_h`.
```
samwyndham, line 448: This was dead code right?

samwyndham (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallViewController.swift
David-Henner: `CallViewController` is the root VC of the calling UI. It has two child VCs: `CallGridViewController` & `CallInfoRootViewController`. Both child VCs have a `configuration` property which, when set, triggers an update of their views. This is what `updateConfiguration()` does: it recreates and sets the configuration objects for each child VC.

By removing the observer from `CallViewController` and moving it to `CallGridViewController` we avoid going through this main pipeline of UI updates through configuration objects. Instead, we're directly targeting the UI components that need to be updated

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
David-Henner: yes, it updates the `networkQuality` for individual participants but we don't need it

wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallViewController.swift
samwyndham: That makes sense. Thanks for the explanation!

PR #1086, opened 2024-03-05
feat: Soft lock an app if the certificate wasn't enrolled during the grace period

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6804" title="WPB-6804" target="_blank">Sub-taskWPB-6804</a>  [iOS] Soft blocking an app if the certificate is missing within the grace period
  </td></table>
  <br />
 

----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [x] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [x] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [ ] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

This PR introduces a new app state: `pendingCertificateEnroll`.
In this state the app should be blocked and I reuse the BlockerViewController to implement this.

**When:** If a team admin has enabled E2EI in team settings and the user continued to postpone enrollment. After the grace period finishes, the app should be blocked.

Once the certificate enrollmend is complete, the app will be transferred to an authenticated state.

----
#### PR Post Submission Checklist for internal contributors (Optional)

 - [ ] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----
#### PR Post Merge Checklist for internal contributors

 - [ ] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #1754, opened 2024-07-30
fix: Repeat login when multiple accounts doesn't perform slow sync - WPB-9930

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9930" title="WPB-9930" target="_blank">BugWPB-9930</a>  [iOS] Repeated login for a team user doesn't work
  </td></table>
  <br />
 

### Issue

When multiple accounts are logged into a single device, if a user logs out and tries to re-login to the same account, that account will not be fully synced which results in all kinds of bad state, such as not knowing what team a user is in, missing conversations etc.

The reason is that we have multiple code paths for a user initiated logout depending on whether they are logged in one vs many accounts. 

```swift
func delete(account: Account, reason: ZMAccountDeletedReason) {
    if let secondAccount = accountManager.accounts.first(where: { $0.userIdentifier != account.userIdentifier }) {
        select(secondAccount, tearDownCompletion: { [weak self] in
            self?.tearDownSessionAndDelete(account: account)
        })
    } else if accountManager.selectedAccount != account {
        // Deleted an inactive account, there's no need notify the UI
        self.tearDownSessionAndDelete(account: account)
    } else {
        // Deleted the last account so we need to return to the logged out area
        logoutCurrentSession(deleteAccount: true, error: NSError(userSessionErrorCode: .accountDeleted, userInfo: [ZMAccountDeletedReasonKey: reason]))
    }
}
```

This specific bug was caused because we were not clearing the last event id when multiple accounts are logged in, meaning that when an account re-logged in, it would only sync data since the last event id it saved in its previous logged in state.

To address the issue this PR introduces `static PrivateUserDefaults.removeAll(forUserID:in:)` which takes care of deleting all data saved at a _account_ level in user defaults via `PrivateUserDefaults`. This method is called when logging out no matter if there are single or multiple accounts logged in.

Investigating the issue I felt a little overwhelmed by the multitude of ways to log out. Essentially we have multiple code paths that lead to a _soft_ logout where we don't delete account related information (i.e. when a cookie expires) or _hard_ logout where we delete all account related information.

I was tempted to try and improve the clarity of the whole log out flow but, as this was no small task, instead made a few small changes which I hope will help clarify log out a little. Specifically this PR:

- Fixes the specific bug
- Removes some parameters that represent state that can not happen... E.g. I deleted `deleteCookie ` from `logoutCurrentSession(deleteCookie: deleteCookie, error: nil)` as this method was always called with `deleteCookie == true`
- Removed some duplicate code calls such as attempting to delete a cookie multiple times.
- Removes calls to `UserDefaults.synchronize()` as these are not necessary in modern iOS.
- Fixed `testThatItDestroyedCacheDirectoryAfterLoggedOut()` which was passing but not silently failing.

#### Note ⚠️ 

This PR does introduce 1 behavior change: [Only remove E2EI Activation data in cases where we delete account data](https://github.com/wireapp/wire-ios/commit/bc2d6d1d4249e9286e1ae39b6bed2466b3c46a09). Based on https://github.com/wireapp/wire-ios/pull/1086, I believe that the new behavior is correct - i.e only remove E2EI Activation data when we are _hard_ logging out an account, but would like @KaterinaWire to confirm.

**Please review this commit by commit along with their accompanying comments to review the logic & safety behind the changes**

### Testing

**For the specific fix:**

0. New install - not update
1. Log in to 2 or more accounts
2. Log out of one account
3. Re-login with that account
4. Confirm that the user login seems correct. I.e for team accounts you can check that there is an account name for that account.

**Other tests**
Try to test logging out, logging in with single / multiple accounts to ensure behavior is exspected.

### Checklist

---


wire-ios-utilities/Source/Public/PrivateUserDefaults.swift
```diff
@@ -96,3 +100,22 @@ public protocol DefaultsKey {
...
 
 }
+
+extension Never: DefaultsKey {
+
+    public var rawValue: String { fatalError() }
+
+}
+
+extension PrivateUserDefaults where Key == Never {
```
samwyndham, line 110: `Never` is required so that we can call this static method on this generic type without specifying the type of `Key` as we don't care.

samwyndham:
Interesting to see those failed tests. I will take a look into what is happening

johnxnguyen (approved):
very nice!

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -802,23 +813,12 @@ public final class SessionManager: NSObject, SessionManagerType {
...
-            }
-
-            if deleteAccount {
-                activeUserSession.lastEventIDRepository.storeLastEventID(nil)
-            }
-
             let group = self?.dispatchGroup
             group?.enter()
 
-            activeUserSession.e2eiActivationDateRepository.removeE2EIActivationDate()
```
KaterinaWire, line 814: I think it should be fine do not have it separate and remove all date in user defaults.

KaterinaWire (approved):

PR #1755, opened 2024-07-31
fix: 1-1 conversations locked - WPB-10416

netbe:
### Issue

Due to a race condition, the supported. protocols are calculated after we resolve the 1:1 conversations ending up marking the conversation as readonly as there are no common protocols

Solution: don't run the resolver if mls is off

---

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
+++ b/wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
@@ -943,15 +943,17 @@ extension ZMUserSession: ZMSyncStateDelegate {
             }
         }
 
-        mlsService.commitPendingProposalsIfNeeded()
+        if DeveloperFlag.enableMLSSupport.isOn {
+            mlsService.commitPendingProposalsIfNeeded()
 
-        WaitingGroupTask(context: syncContext) { [self] in
-            do {
-                var getFeatureConfigAction = GetFeatureConfigsAction()
-                try await getFeatureConfigAction.perform(in: syncContext.notificationContext)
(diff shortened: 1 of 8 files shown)


wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -943,15 +943,17 @@ extension ZMUserSession: ZMSyncStateDelegate {
...
-        WaitingGroupTask(context: syncContext) { [self] in
-            do {
-                var getFeatureConfigAction = GetFeatureConfigsAction()
-                try await getFeatureConfigAction.perform(in: syncContext.notificationContext)
-                try await useCaseFactory.createResolveOneOnOneUseCase().invoke()
-            } catch {
-                WireLogger.mls.error("Failed to resolve one on one conversations: \(String(reflecting: error))")
+            WaitingGroupTask(context: syncContext) { [self] in
+                do {
+                    var getFeatureConfigAction = GetFeatureConfigsAction()
```
johnxnguyen: We still want to fetch feature configs

johnxnguyen (commented):
Maybe another safe way (since resolving can happen from different places), is to check the flag inside the one on one resolver, if mls is not enabled, then return a no action result or something like that).

johnxnguyen (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
netbe: resolved

agisilaos (approved):

PR #1756, opened 2024-07-31
fix: crash starting video call - WPB-10382

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10382" title="WPB-10382" target="_blank">BugWPB-10382</a>  [iOS] : Crashes on calls consistently
  </td></table>
  <br />
 

### Issue

In `warnAboutSlowConnection` we instantiate a `NetworkInfo` object. The init takes a `ZMReachability` instance. 
When we reach the end of the method, the `NetworkInfo` instance is deallocated, along with the `ZMReachability` instance.
In the `dealloc` of `ZMReachability` we then hit `RequireString(self.tornDownFlag.rawValue != 0, "Object was never torn down.");` which causes the app to crash intentionally.

### Solution

Call `tearDown` before the `ZMReachability` gets deallocated

### Testing

Start a video call

---

### Checklist

### DIFF
--- a/wire-ios/Wire-iOS/Sources/Helpers/syncengine/ZMConversation+Calling.swift
+++ b/wire-ios/Wire-iOS/Sources/Helpers/syncengine/ZMConversation+Calling.swift
@@ -98,7 +98,8 @@ extension ZMConversation {
             return
         }
 
-        let networkInfo = NetworkInfo(serverConnection: sessionManager.environment.reachability)
+        let reachability = sessionManager.environment.reachability
+        let networkInfo = NetworkInfo(serverConnection: reachability)
         if networkInfo.qualityType() == .type2G {
 
             let badConnectionController = UIAlertController(
@@ -116,6 +117,8 @@ extension ZMConversation {
         } else {
             handler(false)
         }
+
+        reachability.tearDown()
     }
 
     func warnAboutNoInternetConnection() -> Bool {


samwyndham (approved):
Any ideas why we don't tear down the object in its deinit?

wire-ios/Wire-iOS/Sources/Helpers/syncengine/ZMConversation+Calling.swift
```diff
@@ -116,6 +117,8 @@ extension ZMConversation {
         } else {
             handler(false)
         }
+
+        reachability.tearDown()
```
netbe, line 121: but if you tearDown here, will it not crash when we teardown the environment, maybe we never do...

netbe: ah ok, environment.reachability is a computed variable so it's not retained by anything...

what about modifying this inside the networkInfo so one doesn't have to remember to do this explicitly? @David-Henner

David-Henner: `NetworkInfo` doesn't know anything about `ZMReachability`, it expects an object conforming to the `ServerConnection` protocol. I think the problem has more to do with `ZMReachability` itself. Perhaps as @samwyndham suggested we could do the tear down in the deinit, but there may be a reason it has been done differently

netbe (approved):

PR #1757, opened 2024-07-31
fix: prevent duplicate event processing - WPB-10403

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10403" title="WPB-10403" target="_blank">TaskWPB-10403</a>  [iOS] Prevent processing an event more than once
  </td></table>
  <br />
 

### Issue

There's a concurrency bug where we might fetch a stored update event that should have been deleted, and then process that even for a second time. We don't still don't know why this happens, but in the meantime we will avoid to process events more than once.

To achieve this, before processing an event, we use its content hash (which uniquely identifies the event) to check if it has already been added to a thread safe list of hashes. If not, we will process it, then add it to the list. If it is already in the list, we will not process it.

### Testing

The bug is not easy to reproduce, but it's most likely to happen when there are many events to process (i.e the app has been offline and there has been a lot of activity in conversations the self user is a member of) and the phone is under immense load (for me it happened when the device was very hot due to the weather and was likely throttled severely).

I suggest to run this in an internal beta for at least a week and keep an eye out for duplicate messages / logs that an event has skipped processing a second time.

---

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
+++ b/wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
@@ -27,12 +27,14 @@ actor EventProcessor: UpdateEventProcessor {
     private let syncContext: NSManagedObjectContext
     private let eventContext: NSManagedObjectContext
     private var bufferedEvents: [ZMUpdateEvent]
-    private let eventDecoder: EventDecoder
+    private let eventDecoder: any EventDecoderProtocol
     private let eventProcessingTracker: EventProcessingTrackerProtocol
     private let earService: EARServiceInterface
     private var processingTask: Task<Void, Error>?
     private let eventConsumers: [ZMEventConsumer]
     private let eventAsyncConsumers: [ZMEventAsyncConsumer]
+
+    private let processedEventList = ProcessedEventList()
 
     // MARK: Life Cycle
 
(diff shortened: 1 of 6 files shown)


wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
```diff
@@ -225,3 +261,32 @@ extension Notification.Name {
     /// Published after the last event has been processed.
     static let eventProcessorDidFinishProcessingEventsNotification = Self("EventProcessorDidFinishProcessingEvents")
 }
+
+private actor ProcessedEventList {
+
+    private var hashes = Set<Int64>()
+
+    // A full list would coontain approx 80kB.
```
netbe:
```suggestion
    // A full list would contain approx 80kB.
```

netbe (approved):
looks good, just a typo

wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
```diff
@@ -225,3 +261,32 @@ extension Notification.Name {
...
+private actor ProcessedEventList {
+
+    private var hashes = Set<Int64>()
+
+    // A full list would contain approx 80kB.
+    private let capacity = 10_000
+
+    func addEvent(_ event: ZMUpdateEvent) {
+        guard let hash = event.contentHash else {
+            return
```
samwyndham, line 275: The `StoreUpdateEvent`s should all a `contentHash` right and therefore this shouldn't happen? Or am I missing something?

If yes then at a minimum I would assert and log an error here.

wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
```diff
@@ -225,3 +261,32 @@ extension Notification.Name {
...
+        if hashes.count >= capacity, let randomElement = hashes.randomElement() {
+            hashes.remove(randomElement)
+        }
+
+        hashes.insert(hash)
+    }
+
+    func containsEvent(_ event: ZMUpdateEvent) -> Bool {
+        guard let hash = event.contentHash else {
+            return false
```
samwyndham, line 288: Same as above. I would assert and log error here.

samwyndham (approved):
Nice! 🤞

wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
johnxnguyen: yes that's correct, I'll add the assertion

johnxnguyen: done 75842748fdfcc876752f5fa3183f1cd7c8bc095d

wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
johnxnguyen: done 75842748fdfcc876752f5fa3183f1cd7c8bc095d

PR #1762, opened 2024-08-01
refactor: create api service - WPB-10244

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10244" title="WPB-10244" target="_blank">TaskWPB-10244</a>  [iOS] Implement PushChannelProtocol
  </td></table>
  <br />
 

### Issue

The `SyncManager` will receive live update events from a `PushChannelProtocol` which has an async await style api. I am experimenting with a new implementation of the push channel that makes use of async streams. Creating the push channel relies on making a web socket task via a url session. The are two options here:
1. use the existing `ZMTransportSession`, which will require some adapter code
2. give WireAPI some network capabilities.

Our vision is for WireAPI to eventually be able to make network requests and retire WireTransport, but it seemed like too big a task to achieve in the first milestone of the refactoring. The focus now is to implement the SyncManager, which would require minimal network code, so in this PR I'm exploring this by introducing `APIService`.

`APIService` provides functionality to execute url requests. It replaces `HTTPClient` which was provided by WireTransport. In a future iteration it will also all creating web socket tasks. In essence, its functionality resembles that of `ZMTransportSession` but is (so far) much simpler. There are some things missing from the implementation that will need to be added, such as proxy support and access token renewal.

Some things to note:
- It's completely in swift
- It has an async api
- It uses native types for requests and responses

As an example of its usage, I have changed the `BackendInfoAPI` to use the api service instead of the http client.

### Testing

Not testable in a build in its current state.

---

### Checklist

---


WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
```diff
@@ -0,0 +1,131 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+// sourcery: AutoMockable
+/// A service for network communication to a specific backend.
+///
+/// An api service allows api clients to communicate to a target backend.
+/// It may provide additional functionality, such providing authentication
```
samwyndham: Missing an `as` I think

WireAPI/Tests/WireAPITests/APIs/BackendInfoAPI/__Snapshots__/BackendInfoAPITests/testGetBackendInfoRequest.1.txt
```diff
@@ -1,4 +1,3 @@
-- HTTPRequest
-  - path: "/api-version"
-  - method: get
-  - body: none
+curl \
```
samwyndham, line 1: That's pretty interesting that snapshot-testing can convert to `curl`. What are the benefits of `curl` vs the previous format. I'm wondering if the previous format is easier to understand but maybe it misses some data.

WireAPI/Tests/WireAPITests/Helpers/XCTestCase+Assertions.swift
```diff
@@ -20,6 +20,36 @@ import XCTest
...
+    ///   - file: The file name of the invoking test.
+    ///   - line: The line number when this assertion is made.
+
+    func XCTAssertCount(
+        _ collection: some Collection,
+        count: Int,
+        _ message: String? = nil,
+        file: StaticString = #filePath,
+        line: UInt = #line
+    ) throws {
```
samwyndham, line 38: Why do we throw here rather than just relying on `XCTFail`? Is it so we can access values in the collection safely without crashing after the assertion?

WireAPI/Tests/WireAPITests/Mocks/URLSessionMock.swift
```diff
@@ -0,0 +1,41 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+@testable import WireAPI
+
+final class URLSessionMock: URLSessionProtocol {
```
samwyndham: Have you considered mocking `URLProtocol` instead? See https://medium.com/@dhawaldawar/how-to-mock-urlsession-using-urlprotocol-8b74f389a67a. We can use something like https://github.com/WeTransfer/Mocker or roll our own. The advantage here is tests shouldn't break if you change to a different API of URLSession - the mocking surface area is less I think.

WireAPI/Tests/WireAPITests/Network/APIService/APIServiceTests.swift
```diff
@@ -0,0 +1,183 @@
...
+        // Then the request is made against the backend url.
+        let receivedRequest = urlSession.receivedRequests[0]
+        XCTAssertEqual(receivedRequest.url?.absoluteString, backendURL.appendingPathComponent("/foo").absoluteString)
+
+        // Then the request has an access token attached.
+        let authorizationHeader = receivedRequest.value(forHTTPHeaderField: "Authorization")
+        XCTAssertEqual(authorizationHeader, "Bearer some-access-token")
+    }
+
+    func testItThrowsIfAuthenticationIsRequiredByNosAccessTokenIsFound() async throws {
```
samwyndham: Typo

WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
```diff
@@ -0,0 +1,131 @@
...
+
+    /// Execute a request to the backend.
+    ///
+    /// - Parameters:
+    ///   - request: A url request.
+    ///   - requiringAccessToken: Whether the request requires an access token.
+    ///
+    /// - Returns: The response to the request.
+
+    public func executeRequest(
```
samwyndham, line 103: What do we do with the responses of our requests via `APIService`? Do they tend to always be JSON data? I'm wondering if it is worth incorporating JSON parsing here or in a convenience method. I quite like the approach of https://github.com/objcio/tiny-networking (which I'm not sure is production ready) where defining a new endpoint takes very little code. I think this kind generic approach may require less code and testing then the BackendInfoAPIBuilder pattern we are using. But it does give up some flexibility. 

Anyway I'm just thinking out loud and this is not a request for change. Please take a look at the link above. And feel free to close this comment.

samwyndham (approved):
Nice work!

WireAPI/Tests/WireAPITests/APIs/BackendInfoAPI/__Snapshots__/BackendInfoAPITests/testGetBackendInfoRequest.1.txt
johnxnguyen: Before we were snapshotting our own `HTTPRequest` type so we controlled what the snapshot description looks like. Now we're using `URLRequest` and the snapshot library uses a curl description in the snapshot by default. Although the older snapshot is nice and simple, the new one includes the headers without any extra effort which I think is nice.

WireAPI/Tests/WireAPITests/Helpers/XCTestCase+Assertions.swift
johnxnguyen: Yes that's exactly the reason.

WireAPI/Tests/WireAPITests/Mocks/URLSessionMock.swift
johnxnguyen: Ah cool I'll check that out.

WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
johnxnguyen: In most cases they will be json payloads, but not always. I feel like if we include JSON parsing here, then the api service would then become aware of which responses are considered successes and which are not, which to mean is what the api clients are for.

I like the idea of representing endpoints as objects, but I'm not sure how it would handle API versioning. I suppose the most obvious approach would be to switch over an api version inside the endpoint object, but that resembles our existing code that uses a lot of switches to change behavior depending on the version. 

In the end it's a question about how you want to slice the code up, my personal preference is to keep the implementation of a version contained a grouped together. But I'm open to considering alternatives and would be happy to discuss it further on a call.

WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
johnxnguyen:
```suggestion
/// It may provide additional functionality, such as providing authentication
```

WireAPI/Tests/WireAPITests/Network/APIService/APIServiceTests.swift
johnxnguyen:
```suggestion
    func testItThrowsIfAuthenticationIsRequiredButNoAccessTokenIsFound() async throws {
```

WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
johnxnguyen: fixed

WireAPI/Tests/WireAPITests/Network/APIService/APIServiceTests.swift
johnxnguyen: fixed

WireAPI/Tests/WireAPITests/Mocks/URLSessionMock.swift
johnxnguyen: Thanks for this tip, I implemented a mock url protocol and it works nicely! dfa0025f586e4aabf418152f133f6bee439c7306

jullianm (approved):

WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
samwyndham: 👍 Thanks for the detailed response. Good point about API versioning - this is something I have never had to consider.

I think your solution is good and it is probably not worth overthinking this until we see how things work in practice. It is likely that we need a lot of flexibility which your solution provides.

samwyndham (approved):
Looks good! I can imagine we might want to extend `URLProtocolMock` with further convenience functionality in the future but looks perfect for now.

samwyndham (approved):

PR #1767, opened 2024-08-02
feat: Implement manualSessionHandling and connect UI with analytics operations - WPB-10299

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10299" title="WPB-10299" target="_blank">TaskWPB-10299</a>  [iOS] Check if manualSessionHandling needs to be enabled for Countly Configuration
  </td></table>
  <br />
 

With this PR we implement manualSessionHandling and we connect our new analytics implementation with the UI (meaning enabling/disabling) More specifically we introduce two new use cases: 

- `EnableAnalyticsSharingUseCase`
- `DisableAnalyticsSharingUseCase`

In addition to that `TrackingManager` is not a singleton anymore. 

### TODO:
- [x] Push the changes for the alert
- [x] Add tests for the use cases 

---

### Checklist

---


WireAnalytics/Sources/WireAnalytics/AnalyticsManager.swift
```diff
@@ -53,15 +57,42 @@ public struct AnalyticsManager: AnalyticsManagerProtocol {
...
+    /// - Parameter userProfile: The profile of the user to enable tracking for.
+    /// - Returns: An object conforming to AnalyticsSessionProtocol for the new session.
+    public func enableTracking(_ userProfile: AnalyticsUserProfile) -> any AnalyticsSessionProtocol {
+        switchUser(userProfile)
+    }
+
+    // MARK: - Private Helper Methods
+
+    private func updateUserProfile(_ userProfile: AnalyticsUserProfile) {
+        analyticsService.changeDeviceID(userProfile.analyticsIdentifier)
```
johnxnguyen, line 86: The analytics service is the `Countly` instance, and this method is implemented as:

```swift
public func changeDeviceID(_ id: String) {
    changeDeviceID(withMerge: id)
}
```

According to the [docs](https://support.countly.com/hc/en-us/articles/360037753511-iOS-watchOS-tvOS-macOS#h_01HAVHW0RP2DPTREKXC0Q8T6QA):

> The changeDeviceIDWithMerge: method replaces the old device ID on the server with the new one, automatically merging data associated with the old device ID. 

This means that every time you switch user you're not tracking a separate Countly user, but it's all combined into a single user. For example, user A sends 5 events, then we switch to user B and send 1 event. On Countly, we'll see then that user A has no events while user B has 6 events (all that it sent plus those that A sent).

There's a distinction between changing user id and updating an existing user id. I believe when switching users you want to change the id without merge, and when a user logs in and generates a new analytics id, it wants to update the id with a merge.

WireAnalytics/Sources/WireAnalytics/AnalyticsManager.swift
```diff
@@ -53,15 +57,42 @@ public struct AnalyticsManager: AnalyticsManagerProtocol {
...
+        analyticsService.setUserValue(userProfile.teamInfo?.id, forKey: "team_team_id")
+        analyticsService.setUserValue(userProfile.teamInfo?.role, forKey: "team_user_type")
+        analyticsService.setUserValue(userProfile.teamInfo.map { String($0.size.logRound()) }, forKey: "team_team_size")
+    }
+
+    private func clearUserData() {
+        for key in ["team_team_id", "team_user_type", "team_team_size"] {
+            analyticsService.setUserValue(nil, forKey: key)
+        }
+    }
```
johnxnguyen: Two things here:
1. I notice that we're not clearing the id here. I don't think there is a way to remove it, but perhaps a suitable value to set is `CLYTemporaryDeviceID` (see [docs](https://support.countly.com/hc/en-us/articles/360037753511-iOS-watchOS-tvOS-macOS#h_01HAVHW0RP2DPTREKXC0Q8T6QA))
2. The stringly-typed keys is error prone, maybe we can create a `CaseIterable` enum to make sure we remove all values.

WireAnalytics/Sources/WireAnalytics/AnalyticsManager.swift
```diff
@@ -53,15 +57,42 @@ public struct AnalyticsManager: AnalyticsManagerProtocol {
...
         return AnalyticsSession(
             isSelfTeamMember: userProfile.teamInfo != nil,
             service: analyticsService
         )
     }
+
+    /// Disables tracking by ending the current session and clearing user data.
+    public func disableTracking() {
+        clearUserData()
+        analyticsService.endSession()
```
johnxnguyen: It might be better to end the session first, then clear the user data.

WireAnalytics/Sources/WireAnalytics/AnalyticsService.swift
```diff
@@ -20,7 +20,7 @@ import Foundation
 
 // sourcery: AutoMockable
 /// Protocol defining the requirements for an analytics service.
-protocol AnalyticsService {
+public protocol AnalyticsService {
```
johnxnguyen: Why does this need to be public?

wire-ios-sync-engine/Source/SessionManager/SessionManager+AnalyticsUseCases.swift
```diff
@@ -0,0 +1,56 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+extension SessionManager {
+
+    enum SessionManagerError: Error {
```
johnxnguyen: The name of this error suggests it's for general errors within the session manager, but it's all to do with analytics. I suggest a more accurate name.

wire-ios-sync-engine/Source/SessionManager/SessionManager+AnalyticsUseCases.swift
```diff
@@ -0,0 +1,56 @@
...
+
+        case missingAnalyticsManager
+        case missingAnalyticsUserProfile
+        case missingActiveUserSession
+
+    }
+
+    public func makeDisableAnalyticsSharingUseCase() throws -> DisableAnalyticsSharingUseCaseProtocol {
+        guard let analyticsManager else {
+            throw SessionManagerError.missingAnalyticsManager
```
johnxnguyen: Suggestion: `analyticsNotAvailable`

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -891,11 +891,19 @@ public final class SessionManager: NSObject, SessionManagerType {
...
         else {
             return
         }
 
+        let analyticsSession = analyticsManager.switchUser(userProfile)
+
+        activeUserSession.analyticsSession = analyticsSession
+    }
+
+    func getUserAnalyticsProfileForActiveUserSession() -> AnalyticsUserProfile? {
```
johnxnguyen: I suggest to make this an extension on the user session, or a function that accepts the user session as an argument. This would avoid unwrapping the user session twice.

wire-ios-sync-engine/Source/Use cases/DisableAnalyticsSharingUseCase.swift
```diff
@@ -0,0 +1,52 @@
...
+    /// Invokes the use case to disable analytics sharing.
+    func invoke()
+}
+
+/// Concrete implementation of the DisableAnalyticsSharingUseCaseProtocol.
+/// This struct is responsible for disabling analytics sharing.
+public struct DisableAnalyticsSharingUseCase: DisableAnalyticsSharingUseCaseProtocol {
+
+    /// The analytics manager responsible for handling tracking operations.
+    private let analyticsManager: AnalyticsManagerProtocol?
```
johnxnguyen: Why is it optional?

wire-ios-sync-engine/Source/Use cases/DisableAnalyticsSharingUseCase.swift
```diff
@@ -0,0 +1,52 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Countly
+import WireAnalytics
+
+// sourcery: AutoMockable
+/// Protocol defining the interface for disabling analytics sharing.
+public protocol DisableAnalyticsSharingUseCaseProtocol {
```
johnxnguyen: Suggestion: `DisableAnalyticsUseCaseProtocol`

wire-ios-sync-engine/Source/Use cases/EnableAnalyticsSharingUseCase.swift
```diff
@@ -0,0 +1,75 @@
...
+//
+
+import Countly
+import WireAnalytics
+
+// MARK: - EnableAnalyticsSharingUseCaseProtocol
+
+// sourcery: AutoMockable
+/// Protocol defining the interface for enabling analytics sharing.
+public protocol EnableAnalyticsSharingUseCaseProtocol {
```
johnxnguyen: Suggestion: `EnableAnalyticsUseCaseProtocol`

wire-ios-sync-engine/Source/Use cases/EnableAnalyticsSharingUseCase.swift
```diff
@@ -0,0 +1,75 @@
...
+// MARK: - EnableAnalyticsSharingUseCase
+
+/// Concrete implementation of the EnableAnalyticsSharingUseCaseProtocol.
+/// This struct is responsible for enabling analytics sharing for a specific user profile.
+public struct EnableAnalyticsSharingUseCase: EnableAnalyticsSharingUseCaseProtocol {
+
+    // MARK: - Properties
+
+    /// The analytics manager responsible for handling tracking operations.
+    private let analyticsManager: AnalyticsManagerProtocol?
```
johnxnguyen: Why is this optional?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -81,6 +81,8 @@ public final class ZMUserSession: NSObject {
 
     var analyticsSession: (any AnalyticsSessionProtocol)?
 
+    var analyticsManager: (any AnalyticsManagerProtocol)?
```
johnxnguyen: What is this needed?

wire-ios/Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -1160,7 +1160,11 @@
...
 "self.settings.receiveNews_and_offers.description.title" = "Receive news and product updates from Wire via email.";
 
-
+"self.settings.privacy_analytics.alert.title" = "Consent to the Analysis of Your User Behaviour";
+"self.settings.privacy_analytics.alert.message" = "With your consent, we use technologies to analyse the use of Wire Messenger in order to improve and develop it based on user needs. Information about your use of Wire Messenger, such as whether and when an error occurs or a particular function is used (so-called events), and other technical information (e.g., IP address, app version, device type, operating system) is collected using a pseudonymous ID. The collected data cannot be assigned to you. The data is not linked to your personal information or shared with third parties beside Zeta Project Germany GmbH. The ID is deleted from your device after [28] days after the last activity at the latest. The statistical analysis of the collected information helps us to troubleshoot and improve the Wire Messenger. Further information can be found in our privacy policy. You can revoke or change your consent at any time with effect for the future.";
```
johnxnguyen, line 1164: > The ID is deleted from your device after [28] days after the last activity at the latest.

Where does this happen?

wire-ios/Wire-iOS/Sources/Analytics/TrackingManager.swift
```diff
@@ -22,24 +22,47 @@ import WireCommonComponents
...
+            do {
+                if newValue {
+                    let disableUseCase = try sessionManager.makeDisableAnalyticsSharingUseCase()
+                    disableUseCase.invoke()
+                } else {
+                    let enableUseCase = try sessionManager.makeEnableAnalyticsSharingUseCase()
+                    enableUseCase.invoke()
+                }
+            } catch {
+                print("Failed to toggle analytics sharing: \(error)")
```
johnxnguyen: This should be a log, and in case it fails, we need to reset the setting.

wire-ios/Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -421,21 +423,22 @@ extension AppRootRouter {
             return
         }
 
-        TrackingManager.shared.disableAnalyticsSharing = false
-        Analytics.shared.provider?.selfUser = selfUser
+        trackingManager.disableAnalyticsSharing = true
```
johnxnguyen: Won't this override the user's stored preference every time the app is launched?

wire-ios/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
```diff
@@ -231,12 +235,27 @@ final class SettingsPropertyFactory {
...
                     case .number(let number):
-                        tracking.disableAnalyticsSharing = number.boolValue
+                        let newValue = number.boolValue
+
+                        // If newValue is false, it means the user wants to enable analytics sharing
+                        if newValue == false {
+                            self.showAnalyticsConsentAlert { accepted in
+                                if accepted {
+                                    tracking.disableAnalyticsSharing = newValue
+                                    AVSFlowManager.getInstance()?.setEnableMetrics(!newValue)
```
johnxnguyen: Isn't this part of the tracking manager? Why do it here as well?

wire-ios/Wire-iOS/Sources/LaunchSequenceOperation.swift
```diff
@@ -86,7 +86,7 @@ final class TrackingOperation: LaunchSequenceOperation {
             .arguments.contains(AnalyticsProviderFactory.ZMConsoleAnalyticsArgumentKey)
 
         AnalyticsProviderFactory.shared.useConsoleAnalytics = containsConsoleAnalytics
-        Analytics.shared = Analytics(optedOut: TrackingManager.shared.disableAnalyticsSharing)
+       // Analytics.shared = Analytics(optedOut: TrackingManager.shared.disableAnalyticsSharing)
```
johnxnguyen: To remove?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Account.swift
```diff
@@ -384,7 +384,7 @@ extension SettingsCellDescriptorFactory {
     }
 
     func signOutElement() -> SettingsCellDescriptorType {
-        return SettingsSignOutCellDescriptor()
+        return SettingsSignOutCellDescriptor(trackingManager: trackingManager)
```
johnxnguyen: Why does it need the tracking manager? It seems unrelated no?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Developer.swift
```diff
@@ -98,7 +98,7 @@ extension SettingsCellDescriptorFactory {
                            presentationAction: { DatabaseStatisticsController() })
         )
 
-        if !Analytics.shared.isOptedOut && !TrackingManager.shared.disableAnalyticsSharing {
+        if !trackingManager.disableAnalyticsSharing {
```
johnxnguyen: I wonder what role the `Analytics.shared.isOptedOut` played and how it's different to `disableAnalyticsSharing`. Do you know?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsSignOutCellDescriptor.swift
```diff
@@ -54,7 +56,7 @@ final class SettingsSignOutCellDescriptor: SettingsExternalScreenCellDescriptor
             AVSMediaManager.sharedInstance()?.stop(sound: .ringingFromThemSound)
             ZMUserSession.shared()?.logout(credentials: UserEmailCredentials(email: "", password: password ?? "")) { [weak topMostViewController] result in
                 Task { @MainActor in self.activityIndicator.stop() }
-                TrackingManager.shared.disableAnalyticsSharing = false
+                self.trackingManager.disableAnalyticsSharing = false
```
johnxnguyen: Why does logging out change if analytics is enabled?

johnxnguyen (changes requested):
Are you planning to add tests?

wire-ios-sync-engine/Source/Use cases/DisableAnalyticsSharingUseCase.swift
agisilaos: I don't think it needs to be. Only in the case where we have analytics disabled for specific builds... 🤔 Not sure. Removed optionality here: https://github.com/wireapp/wire-ios/commit/d239e12ad69ed903c7dc4527af5b2b1d4d00fc61

wire-ios-sync-engine/Source/Use cases/EnableAnalyticsSharingUseCase.swift
agisilaos: I don't think it needs to be. Only in the case where we have analytics disabled for specific builds... 🤔 Not sure. Removed optionality here: https://github.com/wireapp/wire-ios/commit/d239e12ad69ed903c7dc4527af5b2b1d4d00fc61

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
agisilaos: It is not needed anymore. Good catch: https://github.com/wireapp/wire-ios/commit/92cd5919408c490af8ff75cc74235101e008468b

wire-ios/Wire-iOS/Sources/LaunchSequenceOperation.swift
agisilaos: Yep, removed it here: https://github.com/wireapp/wire-ios/commit/66eb745653a5cab5098e1264ebbb6402d69820dd

wire-ios/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Account.swift
agisilaos: It was needed because we were setting the disableAnalyticsSharing property inside the `private func logout(password: String? = nil)` since I just removed that it's not needed anymore. I removed the trackingManager from the descriptor as well. https://github.com/wireapp/wire-ios/commit/a8288ec6ca5c59eacd5afbfa2b32558f2ac9deac

wire-ios/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsSignOutCellDescriptor.swift
agisilaos: It was there in the previous implementation but for the new one we don't have anything that explicitly says we need to do anything regarding analytics during log out. And because of that I removed the tracking manager entirely from the descriptor: https://github.com/wireapp/wire-ios/commit/a8288ec6ca5c59eacd5afbfa2b32558f2ac9deac

WireAnalytics/Sources/WireAnalytics/AnalyticsService.swift
agisilaos: It doesn't need to be public anymore. It was public because I was trying certain things with the analytics service. Fixed it here: https://github.com/wireapp/wire-ios/commit/03fbc47c5e20620ed828188851f07864159bdebd

wire-ios/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Developer.swift
agisilaos: `isOptedOut` is not used anymore and from what I can found here's how the getter and the setter is looking like:

```swift
import Foundation
import WireDataModel

extension Analytics {

    /// Opt the user out of sending analytics data
    var isOptedOut: Bool {
        get {
            guard let provider else { return true }
            return provider.isOptedOut
        }

        set {
            if newValue && (provider?.isOptedOut ?? false) {
                return
            }

            if newValue {
                provider?.flush {
                    self.provider?.isOptedOut = newValue
                    self.provider = nil
                }
            } else {
                provider = AnalyticsProviderFactory.shared.analyticsProvider()

                if let user = SelfUser.provider?.providedSelfUser {
                    selfUser = user
                } else {
                    assertionFailure("expected available 'user'!")
                    selfUser = nil
                }
            }
        }
    }
}
```

wire-ios-sync-engine/Source/SessionManager/SessionManager+AnalyticsUseCases.swift
agisilaos: You're absolutely right the name doesn't reflect the for those types of errors. Updated it here: https://github.com/wireapp/wire-ios/commit/5690c91fecccb9747ee0e35b7fd6f2ee9012d2de

wire-ios-sync-engine/Source/SessionManager/SessionManager+AnalyticsUseCases.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/5690c91fecccb9747ee0e35b7fd6f2ee9012d2de

wire-ios-sync-engine/Source/Use cases/DisableAnalyticsSharingUseCase.swift
agisilaos: Updated here: https://github.com/wireapp/wire-ios/commit/3ce973cac5f43ee041a4948f91a3257fd15cb357

& here https://github.com/wireapp/wire-ios/commit/7f9d8bc8783614c5e0b0a32abc12b84fdce74b6e

wire-ios-sync-engine/Source/Use cases/EnableAnalyticsSharingUseCase.swift
agisilaos: Updated here: https://github.com/wireapp/wire-ios/commit/3ce973cac5f43ee041a4948f91a3257fd15cb357

& here https://github.com/wireapp/wire-ios/commit/7f9d8bc8783614c5e0b0a32abc12b84fdce74b6e

wire-ios/Wire-iOS/Resources/Base.lproj/Localizable.strings
agisilaos: Provided an answer for you 1:1.

wire-ios/Wire-iOS/Sources/AppRootRouter.swift
agisilaos: Yep! Fixed it here: https://github.com/wireapp/wire-ios/commit/bd2750fe7733ccabe8628e32d9eb1806e67cc176

WireAnalytics/Sources/WireAnalytics/AnalyticsManager.swift
agisilaos: Agreed, fixed it here: https://github.com/wireapp/wire-ios/commit/d5ca76c32819a365b4035eb77982d7323c6b8a7f

wire-ios/Wire-iOS/Sources/Analytics/TrackingManager.swift
agisilaos: Correct! Fixed it here: https://github.com/wireapp/wire-ios/commit/1bd1e64892df204e1cfd17260a6b1ba72a15e3a2

WireAnalytics/Sources/WireAnalytics/AnalyticsManager.swift
agisilaos: Not really sure about the `CLYTemporaryDeviceID` but regarding having a more concrete type for safety. I implemented it here and added documentation: https://github.com/wireapp/wire-ios/commit/fc2ac7c9ab8f7fe91ba20a83a08489a0e522918e

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
agisilaos: I decided to go with the function route: https://github.com/wireapp/wire-ios/commit/bec25454f8ca7ddc9ac4525626707f404f387658

WireAnalytics/Sources/WireAnalytics/AnalyticsManagerProtocol.swift
```diff
@@ -23,4 +23,14 @@ public protocol AnalyticsManagerProtocol {
     /// - Returns: A new analytics session for the switched user.
     func switchUser(_ userProfile: AnalyticsUserProfile) -> any AnalyticsSessionProtocol
 
+    /// Disables analytics tracking and reporting.
+    /// This method should be used when analytics tracking needs to be disables, such as for privacy reasons.
+    func disableTracking()
+
+    /// Enables analytics tracking and reporting.
+    /// This method should be used to re-enable analytics tracking after it has been disabled. Ensure that any necessary state or user context is correctly restored before resuming tracking.
```
samwyndham: Optional: `Ensure that any necessary state or user context is correctly restored before resuming tracking.` seems important. Consider adding it as:

```swift
/// - Warning: Ensure that any necessary state or user context is correctly restored before resuming tracking.
```

That way you get nice scary UI:

Screenshot 2024-08-02 at 16 34 08

wire-ios-sync-engine/Source/Use cases/EnableAnalyticsUseCase.swift
```diff
@@ -0,0 +1,75 @@
...
+
+// MARK: - EnableAnalyticsUseCase
+
+/// Concrete implementation of the EnableAnalyticsUseCaseProtocol.
+/// This struct is responsible for enabling analytics sharing for a specific user profile.
+public struct EnableAnalyticsUseCase: EnableAnalyticsUseCaseProtocol {
+
+    // MARK: - Properties
+
+    /// The analytics manager responsible for handling tracking operations.
```
samwyndham: Optional: Personally I don't think it is necessary to doc comment these private instance properties as they are already well documented in the init method. Keeping them in sync adds some maintenance burden that is not checked by the compiler.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -310,7 +310,7 @@ public final class SessionManager: NSObject, SessionManagerType {
 
     private let minTLSVersion: String?
 
-    private var analyticsManager: (any AnalyticsManagerProtocol)?
+    var analyticsManager: (any AnalyticsManagerProtocol)?
```
samwyndham: I think this can be `private(set) var analyticsManager: (any AnalyticsManagerProtocol)?` as it is not `set` from outside

wire-ios/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory+UIAlertController.swift
```diff
@@ -0,0 +1,54 @@
...
+                self?.presentPrivacyPolicy()
+                completion(false)
+            })
+        ]
+
+        actions.forEach { action in
+            alertController.addAction(UIAlertAction(title: action.title, style: action.style, handler: action.handler))
+        }
+
+        UIApplication.shared.topmostViewController(onlyFullScreen: false)?
```
samwyndham: If there is no `topmostViewController` this would be a serious issue. I think you should guard let for it and log an error if it is not there.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -888,24 +888,29 @@ public final class SessionManager: NSObject, SessionManagerType {
         }
     }
 
-    private func switchAnalyticsUser() {
+    private func switchAnalyticsUser(for userSession: ZMUserSession) {
```
samwyndham: I think this should be `to` instead of `for`. Small thing but I think otherwise it could be confusing.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -888,24 +888,29 @@ public final class SessionManager: NSObject, SessionManagerType {
...
         guard
             let analyticsManager,
-            let activeUserSession
+            let userProfile = getUserAnalyticsProfile(for: userSession)
         else {
             return
```
samwyndham, line 899: Is this a valid state? 

If not I think you should log the error and do an `assertionFailure()` to surface this issue, especially since there are a lot of possible ways `getUserAnalyticsProfile(for:)` can return nil

wire-ios/Wire-iOS/Sources/Analytics/TrackingManager.swift
```diff
@@ -22,24 +22,51 @@ import WireCommonComponents
...
+            using: { _ in
+                AVSFlowManager.getInstance()?.setEnableMetrics(!ExtensionSettings.shared.disableAnalyticsSharing)
+            }
+        )
 
-    static let shared = TrackingManager()
+        super.init()
+    }
 
     var disableAnalyticsSharing: Bool {
```
samwyndham: I think that this property should be tested if it is not already being tested.

samwyndham (approved):

WireAnalytics/Sources/WireAnalytics/AnalyticsManagerProtocol.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/211348f6122f08a19b22bdeccc6b77ae3fd5120e

wire-ios/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory+UIAlertController.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/cd5db3f9625e59c227722780a71e1e219e12d4aa

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/4ded317e123e645bf837c4152f6ec64316ae3089

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/3946e2aeb8cd51f158b993d8cb73c180c15feb38

wire-ios/Wire-iOS/Sources/Analytics/TrackingManager.swift
agisilaos: Out of the scope of this PR but definitely on my mind to do so.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
agisilaos: It is a valid state in some cases analytics are disabled etc.

wire-ios-sync-engine/Source/Use cases/EnableAnalyticsUseCase.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/fe797489d4c9684ecb31b7c52d4168fe7c914477

WireAnalytics/Sources/WireAnalytics/AnalyticsManager.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/3f37bb1867a291670ea2b94d66e4af57383fa402

wire-ios/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
agisilaos: Moved it to the trackingManager: https://github.com/wireapp/wire-ios/commit/508f597ce80e6ab2a0a16c8dfb8b03f4c045901c

WireAnalytics/Sources/WireAnalytics/AnalyticsManager.swift
```diff
@@ -84,14 +84,22 @@ public struct AnalyticsManager: AnalyticsManagerProtocol {
...
+        for key in AnalyticsUserKey.allCases {
+            switch key {
+            case .teamTeamID:
+                analyticsService.setUserValue(userProfile.teamInfo?.id, forKey: key.rawValue)
+            case .teamUserType:
+                analyticsService.setUserValue(userProfile.teamInfo?.role, forKey: key.rawValue)
+            case .teamTeamSize:
+                analyticsService.setUserValue(userProfile.teamInfo.map { String($0.size.logRound()) }, forKey: key.rawValue)
+            }
+        }
```
johnxnguyen: I find this a little confusing, I suggest to be simple:

```swift
analyticsService.setUserValue(userProfile.teamInfo?.id, forKey: AnalyticsUserKey.teamID.rawValue)
analyticsService.setUserValue(userProfile.teamInfo?.role, forKey: AnalyticsUserKey.teamRole.rawValue)
...
```

WireAnalytics/Sources/WireAnalytics/AnalyticsManagerProtocol.swift
```diff
@@ -28,9 +28,11 @@ public protocol AnalyticsManagerProtocol {
     func disableTracking()
 
     /// Enables analytics tracking and reporting.
-    /// This method should be used to re-enable analytics tracking after it has been disabled. Ensure that any necessary state or user context is correctly restored before resuming tracking.
+    /// This method should be used to re-enable analytics tracking after it has been disabled.
     /// - Parameter userProfile: The profile of the user to enable analytics for.
     /// - Returns: A new analytics session for the user.
+    ///
+    /// - Warning: Ensure that any necessary state or user context is correctly restored before resuming tracking.
```
johnxnguyen: Is this warning relevant for the caller of the method?

WireAnalytics/Sources/WireAnalytics/AnalyticsUserKey.swift
```diff
@@ -0,0 +1,33 @@
...
+enum AnalyticsUserKey: String, CaseIterable {
+
+    /// The key for the team ID.
+    case teamTeamID = "team_team_id"
+
+    /// The key for the user type within the team.
+    case teamUserType = "team_user_type"
+
+    /// The key for the size of the team.
+    case teamTeamSize = "team_team_size"
```
johnxnguyen: These names are confusing, I suggest `teamID`, `teamRole`, `teamSize`.

WireAnalytics/Sources/WireAnalyticsSupport/MockAnalyticsManagerProtocol.swift
johnxnguyen, line 1: Is there a reason why this mocked manually and not with sourcery?

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -891,27 +891,24 @@ public final class SessionManager: NSObject, SessionManagerType {
...
-    func getUserAnalyticsProfileForActiveUserSession() -> AnalyticsUserProfile? {
-        guard let activeUserSession else { return nil }
-        let selfUser = ZMUser.selfUser(inUserSession: activeUserSession)
+    func getUserAnalyticsProfile(for userSession: ZMUserSession) -> AnalyticsUserProfile? {
+        let selfUser = ZMUser.selfUser(inUserSession: userSession)
 
         // Set the analytics identifier if it's not present
         if selfUser.analyticsIdentifier == nil {
-            let idProvider = AnalyticsIdentifierProvider(selfUser: activeUserSession.selfUser)
+            let idProvider = AnalyticsIdentifierProvider(selfUser: userSession.selfUser)
```
johnxnguyen: It's confusing that you access the self user it two different ways. I suggest to reuse the self user stored at the beginning of the method.

wire-ios/Wire-iOS/Sources/Analytics/TrackingManager/TrackingManager.swift
```diff
@@ -48,21 +48,36 @@ final class TrackingManager: NSObject, TrackingInterface {
...
         }
         set {
-            ExtensionSettings.shared.disableAnalyticsSharing = newValue
-            AVSFlowManager.getInstance()?.setEnableMetrics(!newValue)
-
-            do {
-                if newValue {
-                    let disableUseCase = try sessionManager.makeDisableAnalyticsSharingUseCase()
-                    disableUseCase.invoke()
+            self.showAnalyticsConsentAlert { userConsented in
```
johnxnguyen: I don't understand... every time we set the value then we should ask for consent?

wire-ios/Wire-iOS/Sources/Analytics/TrackingManager/TrackingManager.swift
```diff
@@ -48,21 +48,36 @@ final class TrackingManager: NSObject, TrackingInterface {
...
+                    // User rejected, so we keep analytics disabled
+                    self.updateAnalyticsSharing(true)
                 }
-            } catch {
-                print("Failed to toggle analytics sharing: \(error)")
             }
         }
     }
 
+    private func updateAnalyticsSharing(_ disable: Bool) {
```
johnxnguyen: I suggest to give the argument a name because it's hard to know what the bool means from the call site. E.g `updateAnalyticsSharing(true)`, what does true mean without looking at the implementation?

wire-ios/Wire-iOS/Sources/Analytics/TrackingManager/TrackingManager.swift
```diff
@@ -48,21 +48,36 @@ final class TrackingManager: NSObject, TrackingInterface {
...
+            } else {
+                let enableUseCase = try sessionManager.makeEnableAnalyticsUseCase()
+                enableUseCase.invoke()
+            }
+        } catch {
+            WireLogger.analytics.error("Failed to toggle analytics sharing: \(error)")
+
+            // Reset the setting
+            ExtensionSettings.shared.disableAnalyticsSharing = !disable
+            AVSFlowManager.getInstance()?.setEnableMetrics(disable)
```
johnxnguyen: This is error prone because you're assuming that the value was previously the opposite of the new value but there's no guard that is actually the case. 

Instead, you could invoke the use case and only after a success update the these settings.

agisilaos: Done: https://github.com/wireapp/wire-ios/pull/1767/commits/4d057e135cfeef971c25594ac2e148f186b7e24f

wire-ios/Wire-iOS/Sources/Analytics/TrackingManager/TrackingManager.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/pull/1767/commits/4d057e135cfeef971c25594ac2e148f186b7e24f

wire-ios/Wire-iOS/Sources/Analytics/TrackingManager/TrackingManager.swift
agisilaos: Fixed it here: https://github.com/wireapp/wire-ios/pull/1767/commits/4d057e135cfeef971c25594ac2e148f186b7e24f

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/edeedc9c3a03de80d18e046fbe7705f22943fe27

WireAnalytics/Sources/WireAnalytics/AnalyticsUserKey.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/70ce8a8bd20ff42f5d9f75590180d1f90dbf6dbf

WireAnalytics/Sources/WireAnalytics/AnalyticsManager.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/e4ab19dde01778aa71df214d8c89191ca7c343e6

WireAnalytics/Sources/WireAnalytics/AnalyticsManagerProtocol.swift
agisilaos: Probably not. I removed it for now. https://github.com/wireapp/wire-ios/commit/fc11fef203a482afe637e069c08381b8d64e956b

WireAnalytics/Sources/WireAnalyticsSupport/MockAnalyticsManagerProtocol.swift
agisilaos: `any` doesn't work well with Sourcery so I needed to create the mock manually.

johnxnguyen (approved):

PR #1770, opened 2024-08-05
fix: Replace old `func verify(matching value: UIImage...` with the new one from the `SnapshotHelper` - WPB-10455

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10455" title="WPB-10455" target="_blank">TaskWPB-10455</a>  [iOS]Replace old `func verify(matching value: UIImage...` with the new one from the `SnapshotHelper`
  </td></table>
  <br />
 

With this PR I replace the old `func verify(matching value: UIImage, named name: String? = nil, file: StaticString = #file, testName: String = #function, line: UInt = #line)` with the new one from `SnapshotHelper`.

In addition to that I renamed `TokenTextAttachmentTests.swift` to `TokenTextAttachmentSnapshotTests.swift` and `TokenSeparatorAttachmentTests` to `TokenSeparatorAttachmentSnapshotTests` to better describe what the tests are about. 

### Checklist

---


samwyndham (approved):
I'm interested in this comment from [42774d60dd3ccfbd4e7fd391ed8a8b55e532fbad](https://github.com/wireapp/wire-ios/pull/1770/commits/42774d60dd3ccfbd4e7fd391ed8a8b55e532fbad)

> The new ones went through ImageOptim to reduce their size

How is this done? Is it a manual process?

agisilaos:
> I'm interested in this comment from [42774d60dd3ccfbd4e7fd391ed8a8b55e532fbad](https://github.com/wireapp/wire-ios/pull/1770/commits/42774d60dd3ccfbd4e7fd391ed8a8b55e532fbad)
> 
> > The new ones went through ImageOptim to reduce their size
> 
> How is this done? Is it a manual process?

@samwyndham It is a manual process for now. I drag and drop the images inside the ImageOptim app and wait until the operation is done. When the operation is done and the size of the images has been reduced I commit the updated and reduced in size reference images.

samwyndham:
> @samwyndham It is a manual process for now. I drag and drop the images inside the ImageOptim app and wait until the operation is done. When the operation is done and the size of the images has been reduced I commit the updated and reduced in size reference images.

@agisilaos I not sure this is a good idea. It's interesting that the tests still pass considering the snapshotted tests and reference images are now different images. But it makes assumptions about how both `snapshot-testing` library and `ImageOptim` work and I can imagine tests breaking with an update to the `snapshot-testing` library. Maybe I'm just missing some context here though. What do you think about it @johnxnguyen?

Also keep in mind that in this case we've actually increased the size of our repo because the old references still exist in history.

caldrian (approved):

samwyndham:
@agisilaos I've looked a little more into `ImageOptim`. As long as we are using it to do lossless optimization I think it should be fine although we should probably agree on a process - maybe this was already discussed before I joined? But I think we should only consider using it if it is indeed lossless otherwise:

- Diff images created between reference and failures would be confusing as they may be less specific to the issue.
- We miss out on the happy path of `swift-snapshot-testing` which tests that the data is the same. If this fails it does other things like perceptual comparison which may add a cost.

PR #1772, opened 2024-08-07
feat: terminate the SFT 1:1 call once the other person hangup the call - WPB-7152

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7152" title="WPB-7152" target="_blank">StoryWPB-7152</a>  [iOS] Add flag for customer specific 1:1 calls in federated environment
  </td></table>
  <br />
 

### Issue
A follow up PR of https://github.com/wireapp/wire-ios/pull/1742

When two participants are in a 1:1 call on SFT, if someone ends the call, the call will stay ongoing for the other person until timeout.
This is because we don't rely on the SFT list, but we have a subconversation that we use for calling.

To resolve that, we agreed on a [short term approch](https://wearezeta.atlassian.net/wiki/spaces/PAD/pages/1314750477/2024-07-29+1+1+calls+over+SFT#Short-term-solution) which aims to :

- Not kick out the participant from subconversation if he leaves the call (MLS)
- Rely on SFT list by checking if the other participant has lost audio, this way the call will be terminated immediately instead of waiting for timeout for MLS and if there was only 1 participant left for Proteus.
- Remove sub conversation after the 1:1 call is ended (MLS)

### Testing

Setup conferenceCalling feature flag, so useSFTForOneToOneCalls is true.
Make a 1:1 call between users from different backends.

### Checklist

---


wire-ios-sync-engine/Source/Calling/WireCallCenterV3+MLS.swift
```diff
@@ -96,8 +96,16 @@ extension WireCallCenterV3 {
...
     private func leaveStaleConferenceIfNeeded(conversationID: AVSIdentifier) {
+        guard let viewContext = uiMOC,
+              let conversation = ZMConversation.fetch(
+                with: conversationID.identifier,
+                domain: conversationID.domain,
+                in: viewContext),
+              conversation.conversationType == .group else {
+            return
+        }
+
```
johnxnguyen: To be consistent with the next guard:

```suggestion
        guard
            let viewContext = uiMOC,
            let conversation = ZMConversation.fetch(
                with: conversationID.identifier,
                domain: conversationID.domain,
                in: viewContext
            ),
            conversation.conversationType == .group
        else {
            return
        }

```

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
```diff
@@ -459,6 +464,48 @@ extension WireCallCenterV3 {
...
+        case .mls:
+            return shouldEndCallForMLS(participants: participants)
+        case .mixed, .proteus:
+            return shouldEndCallForProteus(participants: participants)
+        }
+    }
+
+    private func shouldEndCallForMLS(participants: [AVSCallMember]) -> Bool {
+        guard participants.count == 2,
+              participants[1].audioState == .connecting else {
```
johnxnguyen, line 498: What is special about participants[1]?

David-Henner: I guess we're assuming the 2nd participant is the other user? And if the other user audio state is `connecting` then the call should end? I think it may be better to filter out the self user in that case

David-Henner (commented):
waiting to see the answers to the comments

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
KaterinaWire: As I understood, the self user is always the first member in the payload from AVS. The 2nd participant is the other user

KaterinaWire: Theoretically we can filter out the self-user, but this is a short term solution. Later there will be a proper solution. https://wearezeta.atlassian.net/wiki/spaces/PAD/pages/1314750477/2024-07-29+1+1+calls+over+SFT#Short-term-solution
To be able to filter out self user, we need to fetch it and compare with `AVSCallMember.client`. It can affect performance. @David-Henner  @johnxnguyen Do you insist on this?

David-Henner: Good point about performance, I'd also prefer not to fetch the self user. Though I think it should be possible to avoid a big hit on performance by fetching the self user only once and caching it.
If you'd prefer to avoid it anyway because it's a short term solution, then go for it 👍 But perhaps we could add a comment to explain that we assume the first member to be the self user ?

johnxnguyen: I'm fine with it as is for a temporary fix but a comment would be helpful.

johnxnguyen (approved):

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
KaterinaWire: @David-Henner done

David-Henner (approved):

PR #1780, opened 2024-08-08
refactor: create SelfUserAPI in WireAPI package - WPB-8741

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8741" title="WPB-8741" target="_blank">TaskWPB-8741</a>  [iOS] Create SelfUserAPI
  </td></table>
  <br />
 

### Key points

SelfUserAPI has two endpoints: `getSelfUser` and `pushSupportedProtocols`. The implementation follows the same architecture pattern as the other APIs in the package. 

`getSelfUser` endpoint doesn't return a `supportedProtocols` object in v0, in v5 it does.
`pushSupportedProtocols` is only available from api version >= 5.

### Testing

Testing requests are properly created.
Testing both endpoints `getSelfUser`  and `pushSupportedProtocols` for failure and success.
Making sure calling the `pushSupportedProtocols` in api versions < v5 throws an error.

---

### Checklist

---


WireAPI/Sources/WireAPI/Models/SelfUser/SsoID.swift
```diff
@@ -0,0 +1,37 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// The sso id of the self user
+
+public struct SsoID: Equatable {
```
caldrian:
```suggestion
public struct SSOID: Equatable {
```

WireAPI/Tests/WireAPITests/Mocks/HTTPClientMock.swift
```diff
@@ -97,7 +97,7 @@ final class HTTPClientMock: HTTPClient {
 
     convenience init(
         code: Int,
-        payload: Data
+        payload: Data?
```
caldrian, line 100: Why is this change necessary? We should avoid optionals wherever possible.

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
```diff
@@ -0,0 +1,159 @@
...
+        let request = HTTPRequest(
+            path: "\(pathPrefix)/self",
+            method: .get
+        )
+        
+        let response = try await httpClient.executeRequest(request)
+
+        return try ResponseParser()
+            .success(code: 200, type: SelfUserV0.self)
+            .failure(code: 404, label: "not-found", error: SelfUserAPIError.selfUserNotFound)
```
johnxnguyen: I couldn't see this error response on swagger or in the existing code. Where did you find it?

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
```diff
@@ -0,0 +1,159 @@
...
+    }
+    
+    func pushSupportedProtocols(_ supportedProtocols: Set<SupportedProtocol>) async throws {
+        guard apiVersion >= .v5 else {
+            throw SelfUserAPIError.unsupportedEndpointForAPIVersion
+        }
+        
+        let encoder = JSONEncoder.defaultEncoder
+        let payload = SupportedProtocolsPayloadV5(supportedProtocols: supportedProtocols)
+        let body = try encoder.encode(payload)
```
johnxnguyen: This implementation is for v5 so I suggest to move this to SelfUserAPIV5. Instead in this method we could simply throw the error without checking the api version.

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
```diff
@@ -0,0 +1,159 @@
...
+            path: "\(pathPrefix)/self/supported-protocols",
+            method: .put,
+            body: body
+        )
+        
+        let response = try await self.httpClient.executeRequest(request)
+
+        return try ResponseParser()
+            .success(code: 200)
+            .failure(code: 404, error: SelfUserAPIError.invalidRequest)
```
johnxnguyen: I also didn't find this error anywhere in swagger or the existing code.

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
```diff
@@ -0,0 +1,159 @@
...
+            .parse(response)
+        
+    }
+}
+
+struct SelfUserV0: Decodable, ToAPIModelConvertible {
```
johnxnguyen, line 51: According to Swagger the response model of `GET /self` is the same as some of the `/users` endpoints. This means that we can reuse the response model called `UserResponseV0` and avoid the duplication here.

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
```diff
@@ -0,0 +1,159 @@
...
+    let id: UUID
+    let locale: String
+    let managedBy: ManagedByV0?
+    let name: String
+    let phone: String?
+    let picture: [String]?
+    let qualifiedID: UserID
+    let service: ServiceResponseV0?
+    let ssoID: SsoIDV0?
+    let supportedProtocols: Set<SupportedProtocol>?
```
johnxnguyen: Supported protocols was only introduced in v4, which is also why `UserResponseV4` exists.

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV4.swift
```diff
@@ -0,0 +1,27 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+class SelfUserAPIV4: SelfUserAPIV0 {
```
johnxnguyen: Since the user response changed in v4 we should override the `GET /self` endpoint here to parse the different response.

WireAPI/Sources/WireAPI/Components/ResponseParser.swift
```diff
@@ -63,20 +82,26 @@ struct ResponseParser<Success> {
...
         }
-
-        let failure = try decoder.decode(FailureResponse.self, from: data)
-        throw failure
+        
+        if let data {
+            let failure = try decoder.decode(FailureResponse.self, from: data)
+            throw failure
+        } else {
+            throw ParsingError.nothingToParse
```
johnxnguyen: I think rather than `nothingToParse` it would be more accurate to say the response had no parse result. Usually it should not happen but if it does then it would highlight to us that we have not handled the parsing correctly.

WireAPI/Sources/WireAPI/Models/SelfUser/SelfUser.swift
```diff
@@ -0,0 +1,86 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// User profile for self
+
+public struct SelfUser: Equatable {
```
johnxnguyen, line 23: I believe this duplicates the `User` model.

WireAPI/Tests/WireAPITests/APIs/SelfUserAPI/SelfUserAPITests.swift
```diff
@@ -0,0 +1,232 @@
...
+    // MARK: - Request generation
+
+    func testGetSelfUserRequest() async throws {
+        try await apiSnapshotHelper.verifyRequestForAllAPIVersions { sut in
+            _ = try await sut.getSelfUser()
+        }
+    }
+    
+    func testPushSupportedProtocolsRequest() async throws {
+        try await apiSnapshotHelper.verifyRequest(for: [.v5, .v6]) { sut in
```
johnxnguyen: Are we able to write `verifyRequest(for: .v5...)`? This way if there are new API versions added then they will be tested too.

WireAPI/Tests/WireAPITests/APIs/SelfUserAPI/SelfUserAPITests.swift
```diff
@@ -0,0 +1,232 @@
...
+            _ = try await sut.pushSupportedProtocols([.mls])
+        }
+    }
+    
+    // MARK: - Request unsupported endpoint
+    
+    func testPushSupportedProtocols_UnsupportedVersionError_V0() async throws {
```
johnxnguyen: What do you think about testing this also for v1, v2, and v3?

WireAPI/Tests/WireAPITests/APIs/SelfUserAPI/SelfUserAPITests.swift
```diff
@@ -0,0 +1,232 @@
...
+
+        // When
+        let sut = SelfUserAPIV5(httpClient: httpClient)
+
+        // Then
+        do {
+            try await sut.pushSupportedProtocols([.mls])
+        } catch let error {
+            throw error
+        }
```
johnxnguyen: This do catch is redundant.

WireAPI/Tests/WireAPITests/Mocks/HTTPClientMock.swift
johnxnguyen: I believe because we encountered the first time a response has no payload data.

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
jullianm: response model is slightly different from the UserResponseV0, for instance the SelfUser model has a SSOID object, no legalHoldStatus object and has an id property. Shouldn't we keep both models separated as, while looking similar, they still have these differences ?

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
jullianm: same comment as above, I can however create a SelfUserAPIV4 object that contains the supportedProtocols objects and remove the property from this one.

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
jullianm: makes sense thanks 👍

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
jullianm: yes I haven't quite looked at the errors yet, I will do so and fix them

WireAPI/Tests/WireAPITests/Mocks/HTTPClientMock.swift
jullianm: exactly, we need to handle use cases where API succeeds but no payload is sent back. this is the case for the pushSupportedProtocols API.

WireAPI/Sources/WireAPI/Components/ResponseParser.swift
jullianm: makes sense thx

WireAPI/Sources/WireAPI/Models/SelfUser/SsoID.swift
jullianm: fixed thx

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
jullianm: fixed o there are no explicit errors handling either in Swagger or legacy code with these endpoints, so I removed the explicit `.failure` handling from the ResponseParser and leave it to be decoded freely by the FailureResponse object.

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
jullianm: done

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
jullianm: fixed, see comment above

WireAPI/Tests/WireAPITests/APIs/SelfUserAPI/SelfUserAPITests.swift
jullianm: done. I used a different approach though

WireAPI/Tests/WireAPITests/APIs/SelfUserAPI/SelfUserAPITests.swift
jullianm: done

WireAPI/Tests/WireAPITests/APIs/SelfUserAPI/SelfUserAPITests.swift
jullianm: fixed

WireAPI/Sources/WireAPI/Components/ResponseParser.swift
```diff
@@ -35,26 +40,44 @@ struct ResponseParser<Success> {
         parseBlocks = []
     }
 
+    /// Success with output data
+
     func success<Payload: Decodable & ToAPIModelConvertible>(
         code: Int,
         type: Payload.Type
     ) -> ResponseParser<Success> where Payload.APIModel == Success {
+        precondition(200 ..< 300 ~= code, "Requires a valid success code: 2xx")
```
jullianm, line 49: I think it would make sense to turn the code (Int) into an enum to ensure only success codes can be passed here. Same goes for failure. I will suggest these changes in my next PR.

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV4.swift
```diff
@@ -18,10 +18,76 @@
...
+            path: "\(pathPrefix)/self",
+            method: .get
+        )
+
+        let response = try await httpClient.executeRequest(request)
+
+        return try ResponseParser()
+            .success(code: 200, type: SelfUserV4.self)
+            .parse(response)
+    }
```
johnxnguyen, line 38: nitpick: it'd be nice if this was below the `apiVersion` property.

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
johnxnguyen: ah yes you're right, my mistake. Yes they should be separate.

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
```diff
@@ -0,0 +1,135 @@
...
+    }
+}
+
+struct SelfUserV0: Decodable, ToAPIModelConvertible {
+
+    let accentID: Int
+    let assets: [UserAsset]?
+    let deleted: Bool?
+    let email: String?
+    let expiresAt: UTCTimeMillis?
```
johnxnguyen, line 57: Swagger says this is `UTCTime` not `UTCTimeMillis`.

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV4.swift
```diff
@@ -0,0 +1,93 @@
...
+
+}
+
+struct SelfUserV4: Decodable, ToAPIModelConvertible {
+
+    let accentID: Int
+    let assets: [UserAsset]?
+    let deleted: Bool?
+    let email: String?
+    let expiresAt: UTCTimeMillis?
```
johnxnguyen, line 48: Swagger says this it `UTCTime` not `UTCTimeMillis`.

WireAPI/Tests/WireAPITests/APIs/SelfUserAPI/SelfUserAPITests.swift
```diff
@@ -0,0 +1,277 @@
...
+        // Then
+        await XCTAssertThrowsError {
+            // When
+            try await sut.getSelfUser()
+        }
+    }
+
+    // MARK: - V5
+
+    func testGetSelfUser_SuccessResponse_200_V5() async throws {
```
johnxnguyen: This endpoint didn't change in v5, but it did in v4. I think you only need to rename the test, resource name, and use the v4 api.

johnxnguyen (approved):

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
jullianm: looks like we need UTCTimeMillis as we end up with a decoding error using UTCTime as it expects to decode a date with fractional seconds. the format on Swagger is as follows: 2021-05-12T10:52:02.671Z

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
```diff
@@ -0,0 +1,135 @@
...
+        case managedBy = "managed_by"
+        case name, phone, picture
+        case qualifiedID = "qualified_id"
+        case service
+        case ssoID = "sso_id"
+        case teamID = "team"
+    }
+
+    func toAPIModel() -> SelfUser {
+        SelfUser(id: id,
```
samwyndham: Tiny thing but I guess we use a different style now:

```swift
SelfUser(
    id: id,
    qualifiedID: qualifiedID
    // ..
)
```

It would be nice if we could lint this 🤔

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV1.swift
```diff
@@ -0,0 +1,27 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+class SelfUserAPIV1: SelfUserAPIV0 {
```
samwyndham, line 21: Out of interest is this subclassing pattern for APIs something we are already using?

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV4.swift
```diff
@@ -0,0 +1,93 @@
...
+        case name, phone, picture
+        case qualifiedID = "qualified_id"
+        case service
+        case ssoID = "sso_id"
+        case teamID = "team"
+        case supportedProtocols = "supported_protocols"
+    }
+
+    func toAPIModel() -> SelfUser {
+        SelfUser(id: id,
```
samwyndham: Same comment about formatting as earlier.

WireAPI/Tests/WireAPITests/APIs/SelfUserAPI/SelfUserAPITests.swift
```diff
@@ -0,0 +1,279 @@
...
+    func testPushSupportedProtocolsRequest() async throws {
+        let supportedVersions = APIVersion.v5.andNextVersions
+        try await apiSnapshotHelper.verifyRequest(for: supportedVersions) { sut in
+            _ = try await sut.pushSupportedProtocols([.mls])
+        }
+    }
+
+    // MARK: - Request unsupported endpoint
+
+    func testPushSupportedProtocols_UnsupportedVersionError_V0() async throws {
```
samwyndham: Personally I would do this for all versions where it is unsupported in a _single_ test. 

- As a form of documentation, it makes it easy to understand which API versions behave in the same way.
- It lowers the effort to test all API versions thoroughly.
- It saves a lot of boilerplate.

You can have an array of `APIVersion`s and build the `sut` using `SelfUserAPIBuilder`. While this kind of parameterized testing is not that common in iOS at the moment it will become common when we move to swift testing.

samwyndham (approved):
Nice work!

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV0.swift
jullianm: done thx

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV4.swift
jullianm: done thx

WireAPI/Tests/WireAPITests/APIs/SelfUserAPI/SelfUserAPITests.swift
jullianm: Good point 👍  this way when we introduce a new API version, we make sure new versioned APIs are properly tested on the existing endpoints without having to write additional tests (as you mentioned), I did some changes just for SelfUser for now but it could also be nice to generalize and automatize this behavior to all APIs, we should keep this in mind.

WireAPI/Sources/WireAPI/APIs/SelfUserAPI/SelfUserAPIV1.swift
jullianm: yes

samwyndham (approved):

PR #1785, opened 2024-08-12
fix: Ensure cookie is decryptable when checking if user logged in - WPB-10462

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10462" title="WPB-10462" target="_blank">BugWPB-10462</a>  [iOS] Issues around login & logout behavior
  </td></table>
  <br />
 

### Issue

There are issues when logging out and logging in. From the ticket:

> It Iooks like data wasn't wiped and the app behaves erratically. Sometimes getting session expired, sometimes getting a prompt to enter password again which disappears

This behavior is similar to that of when a cookie expires - which results in erratic behavior. (There is a separate ticket for this).

I have not been able to reproduce the issue. However, we have a lot of code that checks whether a user is logged in by seeing if there is code in the keychain for this user. This is bad because:

1. Data from keychain isn't deleted when the app is deleted. Instead, for security reasons, it is encrypted with a key stored in user defaults. As user defaults is deleted when the app is deleted, while cookie data remains between and app uninstall and re-install cycle, that data is no longer readable so we should no longer use it to determine if a user is logged in.
2. Even if the data is readable it might be expired and so also shouldn't be used to determine if a user is logged in.

I originally addressed both of the above points, so that any time we check whether a user is logged in which check that the cookie exists, can be decrypted and hasn't expired. However I think that was too aggressive a change - perhaps some things only need to know that the cookie exists not that it is still valid.

Therefore this PR addresses point `1` because a cookie that can't be decrypted implies that it was added in a previous install and the app has been uninstalled and reinstalled.

Specifically this PR:

- Creates a single implementation to check whether a user has a cookie - `hasAuthenticationCookie` which checks that the cookie data exists and can be decrypted. Previously we had many implementations which all just checked whether the cookie data exists.
- Makes many changes to tests so that we pass valid cookie data instead of fake data.
- Introduces a new framework, `SwiftTransportSupport`, with some shared code for creating cookie data for tests.

#### Test 1

1. Log in
2. Delete the app
3. Reinstall
4. You should be logged out
5. Try logging in and out repeatably


wire-ios-data-model/Source/Model/AccountManager.swift
```diff
@@ -35,7 +35,7 @@ fileprivate extension UserDefaults {
 }
 
 /// Class used to safely access and change stored accounts and the current selected account.
```
samwyndham: This is no longer being accessed from obj-c 🥳

wire-ios-data-model/Source/Model/AccountStore.swift
```diff
@@ -94,18 +98,11 @@ public final class AccountStore: NSObject {
             try FileManager.default.removeItem(at: root.appendingPathComponent(directoryName))
             return true
         } catch {
-            log.error("Unable to remove all accounts at \(root): \(error)")
```
samwyndham: `root` is a URL. URLs give identifying info about a user so don't think it should be logged. We have a ticket to change how logging works where we can make this safer: https://wearezeta.atlassian.net/browse/WPB-10227

wire-ios-data-model/Tests/Source/Model/AccountStoreTests.swift
```diff
@@ -227,32 +227,6 @@ final class AccountStoreTests: ZMConversationTestsBase {
         XCTAssertNil(store.load(.create()))
     }
 
-    func testThatItReturnsTrueForAccountsContainedInTheStore() {
```
samwyndham: The method that these tests has been removed

wire-ios-transport/Tests/Source/TransportSession/ZMTransportSessionTests.m
```diff
@@ -1648,7 +1648,7 @@ - (void)testThatItStoresCookiesBeforeCallingTheCompletionHandler;
 - (void)setCookieData;
 {
     // Set cookie on cookie storage:
-    NSDictionary *originalHeaders = @{@"Set-Cookie": @"zuid=bar; Expires=Sun, 21-Jul-2024 09:06:45 GMT; Domain=example.com; HttpOnly; Secure"};
```
samwyndham, line 1651: `2024` was a little pessimistic about how long this test would exist 😄

samwyndham:
Looks like there are a lot of tests to fix. I'm removing reviewers until those tests are fixed

wire-ios-sync-engine/Source/SessionManager/BackendEnvironmentProvider+Cookie.swift
```diff
@@ -25,12 +25,10 @@ extension BackendEnvironmentProvider {
...
-        let cookieStorage = self.cookieStorage(for: account)
-
-        if let expirationDate = cookieStorage.authenticationCookieExpirationDate {
-            return expirationDate.timeIntervalSinceNow > 0
-        } else {
-            return cookieStorage.authenticationCookieData != nil
+        guard let expirationDate = cookieStorage(for: account).authenticationCookieExpirationDate else {
+            return false
         }
+
```
samwyndham, line 31: I removed the else branch here because my changes would use `hasAuthenticationCookie` which implicitly check that there exists an expiration date.

wire-ios-transport/WireTransportSupport/HTTPCookie+Testing.swift
```diff
@@ -16,14 +16,16 @@
 // along with this program. If not, see http://www.gnu.org/licenses/.
 //
 
-import Foundation
+@testable import WireTransport
 
-@objc public protocol AuthenticationStatusProvider {
```
samwyndham, line 21: I've no idea why git sees this as a move 🤔 The before and after are 2 unrelated files.

Anyway, deleted the `AuthenticationStatusProvider` **protocol** because it is not used. There is an unrelated class, `AuthenticationStatusProvider` which is still in use.

agisilaos (approved):

wire-ios-transport/Source/Public/ZMPersistentCookieStorage.swift
```diff
@@ -0,0 +1,31 @@
...
+import Foundation
+
+extension ZMPersistentCookieStorage {
+
+    /// Returns true if `self` has an authentication cookie that can be **decrypted**.
+    ///
+    /// - note: This should generally be used in favor of `ZMPersistentCookieStorage.authenticationCookieData` which
+    /// makes no guarantees about whether it's returned value can be decrypted.
+    @objc public var hasAuthenticationCookie: Bool {
+        authenticationCookieExpirationDate != nil
```
johnxnguyen, line 28: I'm assuming that if we can get an expiration date, then it means we have successfully fetched and decrypted the cookie?

samwyndham: Yes exactly. I could have used another method `-[ZMPersistentCookieStorage authenticationCookies]` which is used internally by `-[ZMPersistentCookieStorage authenticationCookieExpirationDate]` and takes care of decrypting the data but that method is currently private and I preferred not to change it. My assumption here is that our cookies will always have an expiration date. Maybe I should confirm this with engineering?

johnxnguyen: I think it's a fair assumption to make, but doesn't hurt to get a confirmation.

johnxnguyen (approved):

jullianm (approved):

PR #1793, opened 2024-08-12
chore: enable running automation tests on PR - WPB-10709

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10709" title="WPB-10709" target="_blank">TaskWPB-10709</a>  [iOS] Enable running automation tests on PR
  </td></table>
  <br />
 

### Issue

**Why?**
To improve the quality of code before it is merged into the develop branch and to detect failed automation tests as soon as possible, we can run them on every PR targeting the **develop** branch. As an improvement, we may add this for release branches.

**How?**
To achieve this, we need to create builds on S3 for each PR, and on the Jenkins side, it will pick up these builds, run the tests, and push back the results.
This PR introduces a new workflow that creates development builds in a folder with the PR number. The rest will be done on Jenkins side, [here is](https://wearezeta.atlassian.net/browse/WPB-9979) the ticket.

**TODOs:**

- remove 'ci/upload-builds-to-s3-on-pr'

### Checklist

---


.github/workflows/build_pr.yml
```diff
@@ -0,0 +1,25 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions: 
+  checks: write
+
+jobs:
+  build_develop:
+    #if: github.event.pull_request.draft == false
```
samwyndham: Is this comment still needed?

.github/workflows/build_pr.yml
```diff
@@ -0,0 +1,25 @@
...
+    uses: ./.github/workflows/_reusable_app_release.yml
+    with:
+      datadog_enabled: false
+      fastlane_action: development_pr
+      is_cloud_build: true
+      pr_s3_path: "PR-${{ github.event.pull_request.number }}"
```
samwyndham, line 23: Something to check... 

1. PR is opened and a build is uploaded to S3
2. PR is updated and a new build is uploaded to same path

Does this work... E.g. overwriting the previous build with the latest build?

.github/workflows/build_pr.yml
```diff
@@ -0,0 +1,25 @@
+name: Build for PR
+
+on:
+  pull_request:
+    types: [ opened, synchronize ]
```
samwyndham: In `Test Pull Request Changes` workflow we have:

```
[opened, synchronize, edited, ready_for_review]
```

Maybe it is worth adding `edited` also. The negative is that this workflow will also run again if the description or title is updated but the positive is that it will run if the target branch changes which I think we want.

fastlane/Fastfile
```diff
@@ -587,6 +600,10 @@ class Build
 
     # Helpers
 
+    def pr_build
```
samwyndham: Optional: There is a [convention](https://github.com/rubocop/ruby-style-guide?tab=readme-ov-file#predicate-methods-suffix) in ruby that methods that return a boolean are named with a question mark. E.g.

```ruby
def pr_build?
```

samwyndham (approved):
Nice. When this is running it will be great!

.github/workflows/build_pr.yml
KaterinaWire: no, removed

.github/workflows/build_pr.yml
KaterinaWire: It's heavy and time consuming work that produces new builds. I'm not sure it would be worse if I added `edited` and `ready_for_review`

fastlane/Fastfile
KaterinaWire: Yes, I see. But it looks like we don't much follow it 🤔

.github/workflows/build_pr.yml
KaterinaWire: It creates a new build with a folder containing the PR number.

KaterinaWire:
```
PR
  PR-1793
   - Wire-development-1793_merge-simulator-12038.ipa
   - Wire-development-1793_merge-simulator-12039.ipa
   - Wire-development-1793_merge-simulator-12043.ipa
  PR-1843
   - Wire-development-1843_merge-simulator-12044.ipa
   - Wire-development-1843_merge-simulator-12045.ipa
```

KaterinaWire: we use the same flow as the Android team

samwyndham: OK, so the build itself will have a unique name, and therefore we keep each build from that particular PR?

fastlane/Fastfile
samwyndham: Yeah it probably doesn't matter that much sticking to conventions considering how much ruby we have. Thanks for making the change anyway!

.github/workflows/build_pr.yml
samwyndham: True. I assume it doesn't cost _us_ anything (our CI time is free for GitHub right?) but isn't great for the planet and might be annoying having to wait for a rebuild if we have just edited the title!

Changing the target branch is probably a rare case so I imagine excluding `edited` if fine. It's just something for us to be aware of. I think we would only need `ready_for_review` if we kept `github.event.pull_request.draft == false` which we are not.

.github/workflows/build_pr.yml
KaterinaWire: yes

KaterinaWire: @samwyndham  but we talk, we discuss it more during the call

samwyndham: I think this is fine. No need to discuss further. I will close my comment.

.github/workflows/build_pr.yml
netbe: it would cancel the current run anyway, editing the target branch does happen, I would add it.

Concerning the time to build, do we need to run these tests each time, I am thinking we probably don't want to run them before it has been reviewed. We could check for that it's approved or use a label to check to trigger the tests. what do you think @KaterinaWire ?

netbe (commented):
I added a comment on a previous conv. + I think we can avoid generating the changelog with another workflow

netbe (approved):

.github/workflows/_reusable_app_release_without_changelog.yml
```diff
@@ -0,0 +1,185 @@
...
+            ~/Library/Logs/DiagnosticReports/**
+            build/**
+
+      - name: Archiving env variables
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: post-build-env-${{ inputs.fastlane_action }}
+          path: |
+            **/.post_build/*.env
```
netbe: remove this

.github/workflows/_reusable_app_release_without_changelog.yml
```diff
@@ -0,0 +1,185 @@
...
+      - name: Load .env file
+        if: always()
+        uses: xom9ikk/dotenv@v2.3.0
+        with:
+          path: fastlane/.post_build
+      - uses: akiojin/decode-base64-github-action@v1.0.2
+        id: base64-decoded-S3_PATHS
+        with:
+          base64: ${{ env.S3_PATHS }}
```
netbe: remove this

.github/workflows/build_pr.yml
KaterinaWire: > [opened, synchronize, edited, ready_for_review]

@samwyndham  added

KaterinaWire: > it would cancel the current run anyway, editing the target branch does happen, I would add it.
> 
> Concerning the time to build, do we need to run these tests each time, I am thinking we probably don't want to run them before it has been reviewed. We could check for that it's approved or use a label to check to trigger the tests. what do you think @KaterinaWire ?

discussed this in a call

.github/workflows/_reusable_app_release_without_changelog.yml
KaterinaWire: done

.github/workflows/_reusable_app_release_without_changelog.yml
KaterinaWire: done

PR #1798, opened 2024-08-13
refactor: create UserPropertiesAPI in WireAPI package - WPB-8748

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8748" title="WPB-8748" target="_blank">TaskWPB-8748</a>  [iOS] Create UserPropertiesAPI
  </td></table>
  <br />
 

### Key points

UserPropertiesAPI has a single endpoint: `getProperty(forKey key: UserProperty.Key) async throws -> UserProperty`. The implementation follows the same architecture pattern as the other APIs in the package.

API can return two different payload formats: an Int or a LabelsPayload object (for the labels).

An API error (bad request 400) is only present in v4.

When defining the parsing success or failure codes for the response, we don't use raw type anymore but cases from  the `HTTPStatusCode` enum.

Instead of : 
```
ResponseParser()
            .success(code: 200, type: payload)
            .failure(code: 404, error: UserPropertiesAPIError.propertyNotFound)
            .parse(response)
```
**We do:**
```
ResponseParser()
            .success(code: .ok, type: payload)
            .failure(code: .notFound, error: UserPropertiesAPIError.propertyNotFound)
            .parse(response)
```
This defines a clear perimeter of what responses codes should we use and is more explicit about what the status code is about.

### Testing

Testing requests are properly created.
Testing endpoint for success and handling the different response formats.
Making sure errors are properly handled in v4 and the other versions.

### Checklist

---


WireAPI/Sources/WireAPI/HTTP Client/HTTPStatusCode.swift
```diff
@@ -0,0 +1,51 @@
...
+
+/// A list of HTTP status codes
+
+public enum HTTPStatusCode: Int {
+
+    // MARK: Success - 2xx
+
+    /// ok - 200
+
+    case ok = 200
```
samwyndham, line 29: This is somewhat outside the scope of this PR but are we certain that a success is always a `200` across all APIs?

WireAPI/Sources/WireAPI/APIs/UserPropertiesAPI/UserPropertiesAPIV0.swift
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    func getProperty<Payload: UserPropertiesResponseAPIV0>(
+        response: HTTPResponse,
+        payload: Payload.Type
+    ) throws -> UserProperty where Payload.APIModel == UserProperty {
+        try ResponseParser()
+            .success(code: .ok, type: payload)
+            .failure(code: .notFound, error: UserPropertiesAPIError.propertyNotFound)
```
samwyndham, line 59: I feel like we could start making some convienence methods here as these kinds of things will be repeated often. E.g.

```swift
func notFoundFailure() -> ResponseParser<Success> 
```

This is not a request for change. I'm just thinking out loud

WireAPI/Sources/WireAPI/HTTP Client/HTTPStatusCode.swift
```diff
@@ -0,0 +1,51 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// A list of HTTP status codes
+
+public enum HTTPStatusCode: Int {
```
johnxnguyen: Nice! I suggest to add a link to all the status codes, such as https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#successful_responses. This will help other devs understand how to add new cases. Also, does this need to be public?

WireAPI/Sources/WireAPI/APIs/UserPropertiesAPI/UserPropertiesAPIV0.swift
```diff
@@ -0,0 +1,113 @@
...
+            return try getProperty(response: response, payload: TypeIndicatorModeResponseV0.self)
+        case .labels:
+            return try getProperty(response: response, payload: LabelsResponseV0.self)
+        }
+    }
+
+    func getProperty<Payload: UserPropertiesResponseAPIV0>(
+        response: HTTPResponse,
+        payload: Payload.Type
+    ) throws -> UserProperty where Payload.APIModel == UserProperty {
```
johnxnguyen: I suggest to make the name a bit more precise:

```suggestion
    func parseResponse<Payload: UserPropertiesResponseAPIV0>(
        _ response: HTTPResponse,
        forPayloadType type: Payload.Type
    ) throws -> UserProperty where Payload.APIModel == UserProperty {
```

WireAPI/Sources/WireAPI/APIs/UserPropertiesAPI/UserPropertiesAPI.swift
```diff
@@ -0,0 +1,31 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+// sourcery: AutoMockable
+/// An API access object for endpoints concerning user properties.
+public protocol UserPropertiesAPI {
+
+    /// Get the proterty value for a given key.
```
samwyndham: Typo

WireAPI/Sources/WireAPI/APIs/UserPropertiesAPI/UserPropertiesAPIV6.swift
```diff
@@ -0,0 +1,37 @@
...
+
+import Foundation
+
+class UserPropertiesAPIV6: UserPropertiesAPIV5 {
+
+    override var apiVersion: APIVersion {
+        .v6
+    }
+
+    override func getProperty<Payload: UserPropertiesResponseAPIV0>(
```
samwyndham: Is this necessary. Isn't it the same as the method in `UserPropertiesAPIV5`?

WireAPI/Sources/WireAPI/APIs/UserPropertiesAPI/UserPropertiesAPIV6.swift
```diff
@@ -0,0 +1,37 @@
...
+        response: HTTPResponse,
+        payload: Payload.Type
+    ) throws -> UserProperty where Payload.APIModel == UserProperty {
+        try ResponseParser()
+            .success(code: .ok, type: payload)
+            .failure(code: .notFound, error: UserPropertiesAPIError.propertyNotFound)
+            .parse(response)
+    }
```
johnxnguyen: I don't see the change compared to v5.

WireAPI/Tests/WireAPITests/APIs/UserPropertiesAPI/UserPropertiesAPITests.swift
```diff
@@ -0,0 +1,142 @@
...
+
+    func testGetUserPropertyRequest() async throws {
+        try await apiSnapshotHelper.verifyRequestForAllAPIVersions { sut in
+            _ = try await sut.getProperty(forKey: .wireTypingIndicatorMode)
+        }
+    }
+
+    // MARK: - Response handling
+
+    // MARK: - V0
```
samwyndham, line 53: How hard would it be to make these tests cover all versions via parametized testing like you did in your other PR?

samwyndham (approved):

WireAPI/Sources/WireAPI/APIs/UserPropertiesAPI/UserPropertiesAPIV0.swift
```diff
@@ -0,0 +1,113 @@
...
+        self.value = .areTypingIndicatorsEnabled(value == 1)
+    }
+}
+
+struct LabelsResponseV0: UserPropertiesResponseAPIV0 {
+    let value: UserProperty
+
+    init(from decoder: any Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        let payload = try container.decode(LabelsPayload.self)
```
johnxnguyen: We should be careful to use this model because it's not explicitly associated with v0, so the risk is that it may change and therefore break v0 or any other version that uses it. What do you think about having an explicit v0 model for this?

WireAPI/Tests/WireAPITests/APIs/UserPropertiesAPI/UserPropertiesAPITests.swift
jullianm: Yep will do thanks, I hadn't merged develop into my branch before

WireAPI/Sources/WireAPI/APIs/UserPropertiesAPI/UserPropertiesAPIV6.swift
jullianm: it is the same but UserPropertiesAPIV5 inherits from UserPropertiesAPIV4 which has an extra error code, this error was then removed from the next versions. it is bit redundant I agree but it is designed in a a way so we know exactly what are the changes introduced in the different api versions. (I'm not sure why this extra error was only introduced in a specific version)

WireAPI/Sources/WireAPI/APIs/UserPropertiesAPI/UserPropertiesAPIV0.swift
jullianm: better 👍  thx

WireAPI/Sources/WireAPI/APIs/UserPropertiesAPI/UserPropertiesAPIV6.swift
jullianm: It's kind of related to my comment to Sam but actually I should have not overridden this method in V6 as it already inherits from V5 that was already changed. So I will most likely remove it.

WireAPI/Sources/WireAPI/HTTP Client/HTTPStatusCode.swift
jullianm: right, should not be public, thanks

WireAPI/Sources/WireAPI/HTTP Client/HTTPStatusCode.swift
jullianm: yes we have Swagger documentation (and legacy codebase as well) to make sure the success / failure codes we use are correct. For now this file is only filled with status codes we use for the refactored APIS but as the refactoring evolves we might add some other success/failure code, this will be done gradually I guess

WireAPI/Sources/WireAPI/APIs/UserPropertiesAPI/UserPropertiesAPIV0.swift
jullianm: Not sure what you mean by that, could you be more specific ? Note that each failure case has a specific error e.g `UserPropertiesAPIError.propertyNotFound` which is an error specific to UserPropertiesAPI

samwyndham: Oh good point that the error is specific to the `UserPropertiesAPI`. I hadn't considered that. I saw that `.failure(code: .notFound, error: UserPropertiesAPIError.propertyNotFound)` was repeated 4 times in this PR and thought it might happen a lot more. I meant we could have a convenience `notFoundFailure()` method so the calling is a little simpler:

```swift
 try ResponseParser()
            .success(code: .ok, type: payload)
            .notFoundFailure()
            .parse(response)
```

But considering that the error is API specific I don't think the change is worth it. It is better to be explicit. Please disregard my comment.

WireAPI/Sources/WireAPI/HTTP Client/HTTPStatusCode.swift
samwyndham: Cool. Perhaps as things evolve it would make sense for the `success` case to take a collection of status codes. Anyway that is for the future. Nothing to change here.

WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Event decoding/User/UserPropertiesSetEventDecoder.swift
```diff
@@ -74,7 +74,7 @@ struct UserPropertiesSetEventDecoder {
 
 }
 
-struct LabelsPayload: Decodable {
+struct LabelsPayloadV0: Decodable {
```
johnxnguyen: Update events are not versioned, so I think making the payload versioned might cause confusion. To maintain separation between versioned APIs and update events, I would leave this payload is it was, and duplicated it as `LabelsPayloadV0` within the user properties api.

johnxnguyen (approved):
Nice work, thanks

WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Event decoding/User/UserPropertiesSetEventDecoder.swift
jullianm: Ok good to know thanks, I'll have to do some changes in my other PR also

PR #1800, opened 2024-08-14
fix: Avoid and log account deletion failures - WPB-10462

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10462" title="WPB-10462" target="_blank">BugWPB-10462</a>  [iOS] Issues around login & logout behavior
  </td></table>
  <br />
 

### Issue

There are issues when logging out and logging in. From the ticket:

> It Iooks like data wasn't wiped and the app behaves erratically. Sometimes getting session expired, sometimes getting a prompt to enter password again which disappears

I have not been able to reproduce the issue. However one danger area that could cause this issue is this:

When deleting an account we delete its file from disc. During this process we check whether it exists on disk via a call to [fileExists(atPath:isDirectory:)](https://developer.apple.com/documentation/foundation/filemanager/1410277-fileexists). Apple states that this is a bad idea in it's documentation:

> Attempting to predicate behavior based on the current state of the file system or a particular file on the file system is not recommended. Doing so can cause odd behavior or race conditions.

To address this issue, this PR:

- No longer checks whether an `Account` file exists on disk before attempting to delete it. I think this is better behavior anyway as attempting to delete an no existent account feels like an error.
- Change `AccountStore` to log using `WireLogger`. This means we will be aware of account deleting issues via DataDog. This is necessary as local logs are deleted during logout so cannot be inspected. 

### Testing

Try to reproduce the original issue by logging in and logging out repeatably. I was unable to reproduce the error original issue before this _fix_ but it is still worth trying.

### DIFF
--- a/wire-ios-data-model/Source/Model/AccountManager.swift
+++ b/wire-ios-data-model/Source/Model/AccountManager.swift
@@ -35,7 +35,7 @@ fileprivate extension UserDefaults {
 }
 
 /// Class used to safely access and change stored accounts and the current selected account.
-@objcMembers public final class AccountManager: NSObject {
+public final class AccountManager: NSObject {
 
     private let defaults = UserDefaults.shared()
     private(set) public var accounts = [Account]()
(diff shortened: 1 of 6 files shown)


wire-ios-data-model/Source/Model/AccountStore.swift
```diff
@@ -116,7 +113,7 @@ public final class AccountStore: NSObject {
             let paths = try fileManager.contentsOfDirectory(atPath: directory.path)
             return Set<URL>(paths.filter(uuidName).map(directory.appendingPathComponent))
         } catch {
-            log.error("Unable to load accounts from \(directory), error: \(error)")
```
samwyndham, line 119: I no longer log the urls as these reveals personal info

wire-ios-data-model/Source/Model/AccountManager.swift
```diff
@@ -35,7 +35,7 @@ fileprivate extension UserDefaults {
 }
 
 /// Class used to safely access and change stored accounts and the current selected account.
-@objcMembers public final class AccountManager: NSObject {
```
agisilaos, line 38: Any idea why those marked with @objc prefix?

agisilaos (approved):

wire-ios-data-model/Source/Model/AccountStore.swift
```diff
@@ -41,7 +42,7 @@ public final class AccountStore: NSObject {
     public required init(root: URL) {
         directory = root.appendingPathComponent(AccountStore.directoryName)
         super.init()
-        try! FileManager.default.createAndProtectDirectory(at: directory)
+        try! fileManager.createAndProtectDirectory(at: directory)
```
jullianm, line 45: we should avoid force unwrap when possible, we could make it safer by making the init throwable and a do a try catch at callsite

jullianm (changes requested):

wire-ios-data-model/Source/Model/AccountManager.swift
samwyndham: I assume for legacy reasons. I've confirmed that these things are no longer used though from obj-c

wire-ios-data-model/Source/Model/AccountStore.swift
samwyndham: I think fixing this is outside the scope of the PR because it won't just be a matter of logging an error at the call site. 

Essentially if this fails the app is unusable because we won't be able to login etc -  we would not be able to recover from this failure.

I think you are right but the only way to handle this correctly I think would be to display a UI at app startup that the app is not usable. Making this work would require big changes.

jullianm: I agree that if it involves too much changes this should not be done in this PR but if the changes don't have too much impact or require little work, I think we should fix these issues when we can.

jullianm (approved):

wire-ios-data-model/Source/Model/AccountStore.swift
samwyndham: Absolutely agree!

wire-ios-data-model/Source/Model/AccountManager.swift
samwyndham: @agisilaos I'm resolving this comment to merge this PR. If you would like to discuss further feel free to re-open the comment.

PR #1806, opened 2024-08-14
fix: link preview image not uploading - WPB-10622

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10622" title="WPB-10622" target="_blank">BugWPB-10622</a>  [iOS] Link preview image sometimes doesn't upload
  </td></table>
  <br />
 

### Issue

When sending a link preview with an image, the image doesn't get uploaded and is not visible to the sender. 

### Cause

The asset is stored encrypted in the file cache, and when we try to retrieve it, we compare it's SHA256 digest with the one passed in (stored on the message) and if it doesn't match we'll delete it from the file cache forever. Most likely due to some concurrency issues, it can happen that when fetching the link preview image to display in the conversation that the decryption key and SHA256 digest are both **empty**. This would result in deleting the encrypted asset, which would prevent it from being uploaded and also from being visible to even the sender.

Why do we pass in an empty key and digest? I don't know, but it may be because these are taken from the message and then cross an async boundary 🤷🏻‍♂️. 

### Solution

The solution is just a workaround: I found that if we simply return nil when the input arguments are empty, it will be invoked again soon after with the non-empty values and succeed. 

I think we should re-visit asset storage and management to try to avoid strange behavior like this.

### Testing

Send several link previews and ensure that the images are uploaded and received on the other side.

---

### Checklist

---

### DIFF
--- a/wire-ios-data-model/Source/Model/Message/FileAssetCache.swift
+++ b/wire-ios-data-model/Source/Model/Message/FileAssetCache.swift
@@ -694,6 +694,19 @@ public final class FileAssetCache: NSObject {
         encryptionKey: Data,
         sha256Digest: Data
     ) -> Data? {
+        // Workaround: when decrypting data for the link preview, the key
+        // and digest are sometimes empty (not sure why). An empty digest
+        // will always fail the digest check and result in deleting the
+        // asset forever. As a workaround, just return nil with these
+        // invalid empty inputs, so next time the asset is fetched with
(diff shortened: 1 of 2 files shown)


agisilaos (approved):
Nice! 🙌

caldrian (approved):

jullianm (approved):

PR #1828, opened 2024-08-16
chore: remove RootViewController class - WPB-10852

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10852" title="WPB-10852" target="_blank">Sub-taskWPB-10852</a>  Remove RootViewController class
  </td></table>
  <br />
 

### Issue

This PR removes the `RootViewController` class in order to reduce complexity of the view(-controller) hierarchy.

### Testing

There shouldn't be any changes noticeable by the user.
Screen transitions affected:
- App start (launch screen to conversation list or login form)
- Transition to and from remove devices screen (authentication flow)
- and more...

---

### Checklist

---


WireUI/Tests/WireReusableUIComponentsTests/AccountImageView/AvailabilityIndicatorViewSnapshotTests.swift
```diff
@@ -26,8 +26,8 @@ final class AvailabilityIndicatorViewSnapshotTests: XCTestCase {
     private var sut: AvailabilityIndicatorView!
     private var snapshotHelper: SnapshotHelper!
 
-    override func setUp() {
-        sut = .init(frame: .init(x: 0, y: 0, width: 20, height: 20))
+    override func setUp() async throws {
+        sut = await .init(frame: .init(x: 0, y: 0, width: 20, height: 20))
```
jullianm, line 30: just wondering why we need to use to async/await here ?

wire-ios/Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -426,8 +419,9 @@ extension AppRootRouter: AppStateCalculatorDelegate {
     ) -> AuthenticatedRouter? {
         guard let userSession = ZMUserSession.shared() else { return  nil }
 
+        let mainWindow = mainWindow
         return AuthenticatedRouter(
-            rootViewController: rootViewController,
+            rootViewController: { mainWindow.rootViewController! },
```
jullianm: would it make sense to create a getter property to easily access the root view controller from here and from the outside  as well ?

`var rootViewController: { mainWindow.rootViewController }`

wire-ios/Wire-iOS/Sources/AuthenticatedRouter.swift
```diff
@@ -57,14 +56,13 @@ final class AuthenticatedRouter {
     // MARK: - Init
 
     init(
-        rootViewController: UIViewController,
+        rootViewController: @escaping () -> UIViewController,
```
jullianm: I may have missed this point but why do we need to make it a closure?

WireUI/Tests/WireReusableUIComponentsTests/AccountImageView/AvailabilityIndicatorViewSnapshotTests.swift
caldrian: Because the async `setUp` method doesn't run on the main actor, but UIKit classes including their initialisers specify the `@MainActor` attribute. Therefore the compiler throws an error without the `await`.

jullianm: right we're dealing with UI stuff, makes sense, thx 👍

wire-ios/Wire-iOS/Sources/AppRootRouter.swift
caldrian: Done: dd4a720fbfcbf4e4724000aa89daa58424c9c8ed

wire-ios/Wire-iOS/Sources/AuthenticatedRouter.swift
caldrian: Before we had a single view controller instance (class `RootViewController`) which was installed on app start and all subsequent changes of screens/flows only changed the child view controllers of this instance.

With this PR I'm removing this layer and the window's `rootViewController` is updated directly.

We need a closure now in order to get the currently installed root view controller from the window.

jullianm: Yep but is the closure required ? I guess we could just do: `init(rootViewcontroller: UIViewController` instead of having `@autoclosure @escaping () -> UIViewController`, from my understanding (from the callsites) we're just passing the rootViewController (which is a `UIViewController`) through the init so this would be sufficient, no?

caldrian: You're right I guess, I reverted the changes for a few classes.
Thank you!

jullianm (approved):

wire-ios/Wire-iOS/Sources/AuthenticatedRouter.swift
caldrian: Actually I had to change it again. We now pass the window, since the time when the window is injected another root view controller is set than later, when the router wants to present some view controller.

jullianm: I am not too comfortable having every router be able to access everything related to the window, what do you think about conforming the window to a `Rootable`protocol that exposes a `rootViewController` that returns the main window root view controller. Routers will only be able to access that property and nothing else.

caldrian: > Routers will only be able to access that property and nothing else.

But that was exactly the purpose of the closure I initially had. It was a way to just get the current `rootViewController` value.

Anyway, we have to refactor this sooner or later.
We should have a proper way for presenting alerts etc.
There should be a single point where it is decided how/where alerts and modals are presented.

wire-ios/Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -440,7 +435,7 @@ extension AppRootRouter: AppStateCalculatorDelegate {
                 lastE2EIdentityUpdateAlertDateRepository: userSession.lastE2EIUpdateDateRepository,
                 e2eIdentityCertificateUpdateStatus: userSession.e2eIdentityUpdateCertificateUpdateStatus(),
                 selfClientCertificateProvider: userSession.selfClientCertificateProvider,
-                targetVC: rootViewController
+                targetVC: { [weak self] in self!.rootViewController }
```
samwyndham, line 438:
```suggestion
                targetVC: { [unowned self] in self.rootViewController }
```

wire-ios/Wire-iOS/Sources/AuthenticatedRouter.swift
jullianm: I guess you still don't need the closure, something like this:

``` 
protocol Rootable {
    var rootController: UIViewController { get }
}
```

```
extension UIWindow: Rootable {
    var rootController: UIViewController { rootViewController }
}

```

```
init(
         rootable: Rootable,
         sessionManager: SessionManager,
         appStateCalculator: AppStateCalculator
     ) {
```

and at callsite you just pass the `mainWindow`

jullianm: But if this will be refactored later, I guess passing the window directly is fine for now

samwyndham (approved):
Looks good! I think the testing section of the description should probably be completed and the pr QA-ed for situations where the root view controller is switched

PR #1361, opened 2024-04-30
feat: navigation overhaul - WPB-6647

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6647" title="WPB-6647" target="_blank">TopicWPB-6647</a>  [iOS] Navigational Overhaul
  </td></table>

### Issue

This PR 
- replaces the custom split view controller by UISplitViewController,
- moves settings into the tab bar, 
- removes the folder-grouped conversation list
- and attempts to decouple the UI code.

---

### Checklist

- [x] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [x] Description is filled and free of optional paragraphs.
- [x] Adds/updates automated tests.

PR #1829, opened 2024-08-16
fix: Unable to tab to Contacts - WPB-10514

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10514" title="WPB-10514" target="_blank">BugWPB-10514</a>  [iOS] Difficulty with contacts screen
  </td></table>
  <br />
 

### Issue

This PR fixes an unreleased bug that prevents showing the contacts view when the navigation is in a certain state. For example the following would occur prior to this fix 100% of the time:

1. Quit app
2. Launch app
3. Tap `Folders`
4. Tap `Contacts` 
5. Tap close on contacts
6. Tap `Contacts` - Nothing happens no matter how many times the user taps `Contacts`.

The root cause of the issue is that currently `ConversationListViewController` is the delegate of the `UITabBarController` and we have two instances of `ConversationListViewController` - used for `Conversations` & `Folders` tabs - which both set themselves as the delegate of `UITabBarController` and respond to taps on the `Contacts` & `Archive` tabs by presenting a corresponding view controller on `self` - i.e whichever `ConversationListViewController` instance is the true delegate of `UITabBarController`. This can lead to some messed up state.

The above is necessary because how we are currently (maybe) misusing our `UITabBarController` - Two tabs (`Conversations` & `Folders`) contain `ConversationListViewController` instances with real content. The other two tabs (`Contacts` & `Archive`) contain **empty** `UIViewController` instances. We use these empty tabs as buttons and have the `UITabBarControllerDelegate` perform actions depending on which tab is selected.

To address this issue, this PR introduces `TabBarChangeHandler` which becomes the **single** delegate of the main `UITabBarController` instance. It tracks the currently selected **real** content tab (i.e `Conversations` or `Folders`) and makes sure to present content for `Contacts` & `Archive` on the correct **real** content tab.

The solution is not pretty but seems to work and is **temporary**. This work will all be discarded in our ongoing navigation overhaul https://github.com/wireapp/wire-ios/pull/1361. I have attempted to minimize changes to avoid conflicts with https://github.com/wireapp/wire-ios/pull/1361.

### Testing

1. Quit app
2. Launch app
3. Tap `Folders`
4. Tap `Contacts` 
5. Tap close on contacts
6. Tap `Contacts` - This should work
7. Keep tapping on all tabs trying to break things.

### Checklist

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
@@ -38,8 +38,6 @@ final class ConversationListViewController: UIViewController {
     /// internal View Model
     var state: ConversationListState = .conversationList
 
-    private var previouslySelectedTabIndex = MainTabBarControllerTab.conversations
-
     /// private
     private var viewDidAppearCalled = false
     private static let contentControllerBottomInset: CGFloat = 16
@@ -184,8 +182,6 @@ final class ConversationListViewController: UIViewController {
 
         if !viewDidAppearCalled {
             viewDidAppearCalled = true
-
-            tabBarController?.delegate = self
 
             zClientViewController?.showAvailabilityBehaviourChangeAlertIfNeeded()
         }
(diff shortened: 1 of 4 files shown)


wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
```diff
@@ -341,31 +337,6 @@ extension ConversationListViewController: ConversationListContainerViewModelDele
     }
 }
 
-// MARK: - UITabBarControllerDelegate
-
-extension ConversationListViewController: UITabBarControllerDelegate {
```
samwyndham, line 346: This logic has moved to `TabBarChangeHandler` with minimal changes.

caldrian (approved):
good work! 👌
thank you for fixing this!

johnxnguyen (approved):

jullianm (approved):

PR #1841, opened 2024-08-20
refactor: create ConversationLabelsRepository - WPB-8749

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8749" title="WPB-8749" target="_blank">TaskWPB-8749</a>  [iOS] Create ConversationLabelRepository
  </td></table>
  <br />
 

### Key points

`ConversationLabelsRepository` acts as an in-between layer between the app’s business logic and data storage.
It contains a method `func pullConversationLabels()` that retrieve labels (and their related conversations) from remote and store them locally. 

Everytime we pull new labels from remote, we ensure the database stays up-to-date by creating or updating (if they already exist) these labels locally and deleting "old" `folder` labels (folder labels that were not fetched from remote).

Note: labels can have two types: `folder` and `favorite` -- ⚠️ `favorite` labels are never deleted from database.

We use a `NSBatchDeleteRequest` as it is faster than deleting objects on the context one by one. (see code comments)

### Testing

Following use cases are covered:
- Creating new labels locally
- Updating existing labels (name, conversations)
- Creating new labels and removing old `folder` labels locally
- Creating  new labels and ensure existing `favorite` label is not deleted.

⚠️ We cannot use the in-memory store to test since `NSBatchDeleteRequest` (and more generally `NSBatchRequest`) are not supported the in-memory store. So we use a persistent store.

---

### Checklist

---


WireAPI/Sources/WireAPI/Models/Conversation/ConversationLabel.swift
```diff
@@ -20,7 +20,7 @@ import Foundation
 
 /// A label used to categorize conversations.
 
-public struct ConversationLabel: Equatable, Codable {
+public struct ConversationLabel: Equatable, Codable, Sendable {
```
johnxnguyen, line 23: Nice!

WireDomain/Sources/WireDomain/Repositories/ConversationLabelsRepository.swift
```diff
@@ -0,0 +1,152 @@
...
+        context: NSManagedObjectContext
+    ) {
+        self.userPropertiesAPI = userPropertiesAPI
+        self.context = context
+    }
+
+    /// Retrieve from backend and store conversation labels locally
+
+    func pullConversationLabels() async throws {
+        let result = try await userPropertiesAPI.getProperty(forKey: .labels)
```
johnxnguyen: I wonder if it would be nicer to have a method like:

```swift
let remoteLabels = try await userPropertiesAPI.getLabels()
```

That way we don't have to deal with other properties? What do you think?

WireDomain/Sources/WireDomain/Repositories/ConversationLabelsRepository.swift
```diff
@@ -0,0 +1,152 @@
...
+        default:
+            throw ConversationLabelsRepositoryError.failedToCollectLabelsRemotely
+        }
+    }
+
+    /// Save label and related conversations objects to local storage.
+    /// - Parameter conversationLabel: conversation label from WireAPI
+
+    private func storeLabelLocally(_ conversationLabel: ConversationLabel) async throws {
+        try await context.perform { [self] in
```
johnxnguyen: do you need to capture `self` or `context`?

WireDomain/Tests/WireDomainTests/Repositories/ConversationLabelsRepositoryTests.swift
```diff
@@ -0,0 +1,332 @@
...
+            userPropertiesAPI: userPropertiesAPI,
+            context: context
+        )
+    }
+
+    override func tearDown() async throws {
+        stack = nil
+        userPropertiesAPI = nil
+        sut = nil
+        try await super.tearDown()
```
johnxnguyen, line 63: Do you need to tear down the core data stack?

WireDomain/Tests/WireDomainTests/Repositories/ConversationLabelsRepositoryTests.swift
```diff
@@ -0,0 +1,332 @@
...
+            let label = Label.fetchOrCreate(remoteIdentifier: Scaffolding.conversationLabel1.id, create: true, in: context, created: &created)
+            label?.name = Scaffolding.conversationLabel1.name
+            context.saveOrRollback()
+        }
+
+        try await context.perform { [context] in
+            let fetchRequest = NSFetchRequest<Label>(entityName: Label.entityName())
+            let a = try context.fetch(fetchRequest)
+            XCTAssertNotNil(a.first)
+            XCTAssertEqual(a.first!.name, Scaffolding.conversationLabel1.name)
```
johnxnguyen, line 105: I think this will crash the test if there is no first. You could use `try AssertCount(a, count: 1)` which will throw if the count is wrong.

WireDomain/Sources/WireDomain/Repositories/ConversationLabelsRepository.swift
jullianm: Good idea 👍

WireDomain/Sources/WireDomain/Repositories/ConversationLabelsRepository.swift
jullianm: good catch, we only need to capture the context.

WireDomain/Tests/WireDomainTests/Repositories/ConversationLabelsRepositoryTests.swift
jullianm: I am using `XCTAssertNotNil(a.first)` right above to make sure of that.

samwyndham: I think @johnxnguyen's point was that even if `XCTAssertNotNil` fails the test will continue so it will crash on `XCTAssertEqual`. `AssertCount` is our custom matcher which will exit early so to prevent following crashes

WireDomain/Sources/WireDomain/Repositories/ConversationLabelsRepository.swift
```diff
@@ -0,0 +1,152 @@
...
+        userPropertiesAPI: any UserPropertiesAPI,
+        context: NSManagedObjectContext
+    ) {
+        self.userPropertiesAPI = userPropertiesAPI
+        self.context = context
+    }
+
+    /// Retrieve from backend and store conversation labels locally
+
+    func pullConversationLabels() async throws {
```
samwyndham, line 47: What happens in the case that this method is called `twice` before the first call is complete? My guess is that we end up in a race condition. Perhaps we need a mechanism to cancel an existing request?

WireDomain/Sources/WireDomain/Repositories/ConversationLabelsRepository.swift
```diff
@@ -0,0 +1,152 @@
...
+            throw ConversationLabelsRepositoryError.failedToCollectLabelsRemotely
+        }
+    }
+
+    /// Save label and related conversations objects to local storage.
+    /// - Parameter conversationLabel: conversation label from WireAPI
+
+    private func storeLabelLocally(_ conversationLabel: ConversationLabel) async throws {
+        try await context.perform { [self] in
+            var created = false
```
samwyndham, line 66: Do we actually need this or is it just because `fetchOrCreate` requires it?

WireDomain/Tests/WireDomainTests/Repositories/ConversationLabelsRepositoryTests.swift
```diff
@@ -0,0 +1,332 @@
...
+
+    var context: NSManagedObjectContext {
+        stack.syncContext
+    }
+
+    override func setUp() async throws {
+        try await super.setUp()
+        /// Batch requests don't work with in-memory store
+        /// so we need to use a persistent store.
+        stack = try await coreDataStackHelper.createStack(inMemoryStore: false)
```
samwyndham, line 48: As this is not operating on an in memory store I think that we should also pass the URL of a random temporary location. That way during `tearDown` we can delete the store at that location. Otherwise say, for instance,   `cleanLabelEntity` fails to delete data , future tests would also be affected.

WireDomain/Tests/WireDomainTests/Repositories/ConversationLabelsRepositoryTests.swift
```diff
@@ -0,0 +1,332 @@
...
+    override func tearDown() async throws {
+        stack = nil
+        userPropertiesAPI = nil
+        sut = nil
+        try await super.tearDown()
+    }
+
+    // MARK: - Tests
+
+    func testPullConversation_Given_Local_Store_Empty_Labels_Are_Created() async throws {
```
samwyndham, line 68: Not a request for change but as a team we should agree on a test name convention as I have seen different things. 🤔

samwyndham (commented):
Looking good! I would just like to understand what happens if we call this repository multiple times. How can we avoid race conditions.

WireDomain/Sources/WireDomain/Repositories/ConversationLabelsRepository.swift
samwyndham: I see myself that it is required for the API. I think we should make that property optional but that is out of scope for this PR so I'll resolve.

WireDomain/Tests/WireDomainTests/Repositories/ConversationLabelsRepositoryTests.swift
jullianm: you're right, I will change that, thx

WireDomain/Tests/WireDomainTests/Repositories/ConversationLabelsRepositoryTests.swift
jullianm: yes it is already done by the CoreDataStackHelper using a random UUID so at every test we're creating a new store at a specific `temporary` location (not removing it though) which should be enough to have the tests passing but they don't anymore, I'll take a better look tomorrow

WireDomain/Sources/WireDomain/Repositories/ConversationLabelsRepository.swift
jullianm: yes that's actually a good point, repositories..writing and accessing data in specific contexts are a good candidate to be actors, no ? We could give it a try. What do you think @johnxnguyen ? (right now repositories are classes so we should indeed provides a mechanism to avoid race condition but I guess now we should use actors for that)

WireDomain/Sources/WireDomain/Repositories/ConversationLabelsRepository.swift
jullianm: fixed in last commit, I also provided getter async throwable var for the other properties

WireDomain/Sources/WireDomain/Repositories/ConversationLabelsRepository.swift
jullianm: fixed in last commit

WireDomain/Tests/WireDomainTests/Repositories/ConversationLabelsRepositoryTests.swift
jullianm: fixed in last commit

WireDomain/Tests/WireDomainTests/Repositories/ConversationLabelsRepositoryTests.swift
jullianm: done

WireDomain/Tests/WireDomainTests/Repositories/ConversationLabelsRepositoryTests.swift
jullianm: So, I added some clean up of the temp directory so we make sure to remove the directory after each test however for some reason we still need to clean up the entity so we have a clean database and have the tests pass. Not sure why tbh, I tried removing the file, closing the stores (that do some reset of the context), it would not work. It behaves in a weird way, I figured cleaning up the entity is the most straightforward approach.

jullianm: you can check in last commit

WireDomain/Tests/WireDomainTests/Repositories/ConversationLabelsRepositoryTests.swift
jullianm: true. I would go with a name convention that describes at best the test using Given, When,Then, so when the tests fails we know exactly what it is about

jullianm: `func someTestGiven..When..Then()`

WireDomain/Tests/WireDomainTests/Repositories/ConversationLabelsRepositoryTests.swift
samwyndham: Maybe we should recreate the `CoreDataStackHelper()` in the setup too. I looked quickly at the implementation and the location is unique per `CoreDataStackHelper` instance. That could be causing the issues you mentioned above.

WireDomain/Sources/WireDomain/Repositories/ConversationLabelsRepository.swift
samwyndham: I agree that repositories could be good candidates for `actor`s. I don't think an `actor` would help in this specific case because they are [reentrant](https://www.donnywals.com/actor-reentrancy-in-swift-explained/)

jullianm: great article. the provided example to handle actor reentrancy is straightforward, by keeping the state of the task but then I don't see the point of using an actor if we still need that kind of mechanism for async/await method. We can just use the same mechanism with a class and still avoid race condition. So in our specific case, as you mentioned, an actor seems useless, but the 'lock' mechanism from the article is good.

jullianm: like maybe we could use this mechanism and introduce some kind of queues : an array of task with a FIFO system

jullianm: actually I don't think there's a need to do that in the case you mentioned, if `pullConversationLabels` is called twice, I guess it will just wait for the first call to finish before doing the second one since the method is async and the repo is not an actor. So there would not be any race condition in that case, what do you think?

WireDomain/Tests/WireDomainTests/Repositories/ConversationLabelsRepositoryTests.swift
jullianm: yeah did that too, still not working

WireDomain/Sources/WireDomain/Repositories/ConversationLabelsRepository.swift
jullianm:
```
// Starts first async work
let labels = await repository.pullConversationLabels() 
// Finishes first async work

// Starts second async work
let labels2 = await repository.pullConversationLabels() 
// Finishes second async work

``` 

There would not be any race condition in that case, however if we were using `async let`, these calls would run concurrently and there might be a race condition.

johnxnguyen: I don't think re-entrancy is a problem here, it just does the same work twice. We should avoid repeated calls, but I don't think any 'pull' method critically needs to be completely under mutual exclusion.

johnxnguyen: I'm also wondering how exactly an actor will help here? What's the data race we would prevent?

jullianm: yes, re-entrancy not a problem here since this is not an actor. However, as mentioned above, if were running two concurrent `pullConversationLabels` using `async let` there might be one but I don't see the cases where it would be needed to do so in the code.

samwyndham: > I'm also wondering how exactly an actor will help here? What's the data race we would prevent?

The possible data race is this:

1. Call `pullConversationLabels`.
2. Call `1` reads from backend but for some reason takes some time to finish.
3. Labels change on backend.
4. Call `pullConversationLabels`
5. Call in `3` finishes.
6. Call in `1` finishes.

In this situation the labels now reflect the old version and not the latest version. 

This might be a very rare situation in practice and whether this is a problem or not dependents on how use `pullConversationLabels`.

For example we call `pullConversationLabels` often then it doesn't matter. We will get the correct values later anyway. But if we call `pullConversationLabels` rarely in response to a notification from backend for example then it is something we should consider.

I think the reason why I raised this concern is that the repositories API doesn't return anything, rather it updates local data as a side affect, and it feels funny that we can get in a situation where the more recent call to `pullConversationLabels` results could be override.

Personally I may consider using `Task` cancellation and have `pullConversationLabels` cancel an existing call if there is one if this is not too complex and if this is a situation we want to avoid

WireDomain/Tests/WireDomainTests/Repositories/ConversationLabelsRepositoryTests.swift
samwyndham: Strange. I still think it is worth creating `CoreDataStackHelper` in setup though.

WireDomain/Sources/WireDomain/Repositories/ConversationLabelsRepository.swift
jullianm: I don't think it would be too complicated to implement but we should consider applying this mechanism to other calls if we want to be consistent in our codebase. We are pulling and storing in many places.

WireDomain/Tests/WireDomainTests/Repositories/ConversationLabelsRepositoryTests.swift
jullianm: fixed

WireDomain/Sources/WireDomain/Repositories/ConversationLabelsRepository.swift
jullianm: I tried playing around with TaskGroup cancellation, it's not as easy as I thought tbh, I used `addTaskUnlessCanceled` and `cancelAll()` but things became messy, so then I figured wrapping it into a single Task and leveraging a `Task.checkCancellation()` (like the code below) to make it more straightforward, seems OK when doing some unit testing although tests are sometimes inconsistent (because we're using concurrent calls).. So I'm not 100% confident pushing this code to the branch, maybe we could leave it as is for now and keep in mind that a task cancellation mechanism may be needed if the future. Any suggestions are welcome.

```
    func pullConversationLabels() async throws {
        let conversationLabels = try await userPropertiesAPI.getLabels()
        
        ongoingTask?.cancel()
        
        ongoingTask = Task {
            let conversationLabels = try await userPropertiesAPI.getLabels()
            
            try await withThrowingTaskGroup(of: Void.self) { [self] taskGroup in
                try Task.checkCancellation()
                
                for conversationLabel in conversationLabels {
                    taskGroup.addTask { [self] in
                        try await storeLabelLocally(conversationLabel)
                    }
                }
                
                try await taskGroup.waitForAll()
            }
            
            try await deleteOldLabelsLocally(excludedLabels: conversationLabels)
        }
        
        try await ongoingTask?.value
    }
```

samwyndham: 👍 Thanks for investigating. OK for now I suggest we don't let this issue block the PR. Instead we can discuss with @johnxnguyen to see if this will be an actual problem in practice.

samwyndham (approved):
Nice work! Regarding _repositories_ and how they handle race conditions, let's discuss outside the context of this PR.

WireDomain/Tests/WireDomainTests/Helpers/XCTestCase+Assertions.swift
```diff
@@ -0,0 +1,120 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+
+extension XCTestCase {
```
johnxnguyen, line 21: We should probably create a test utility package so we avoid this duplication. But we can do it in another PR.

johnxnguyen (approved):

PR #1846, opened 2024-08-22
fix: Fix `NSLinguisticTagger` performance hit - WPB-10722

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10722" title="WPB-10722" target="_blank">TaskWPB-10722</a>  [iOS] Fix performance issue related to Linguistic tagger instantiation
  </td></table>
  <br />
 

### Issue

During profiling I noticed that one of the apps big performance bottlenecks is creating `NSLinguisticTagger`s.

Screenshot 2024-08-22 at 15 32 56

`static PersonName.person(withName:schemeTagger:)` is called any time a cell is loaded that relies on a `PersonName` such as in a conversation or contacts. During scrolling it is called often.

While we do have caching to ensure the we didn't perform the tagger's methods if a `PersonName` has already been created, we check the cache AFTER creating the tagger.

Fixing this is an easy performance win - we just need to avoid re-creating `NSLinguisticTagger` unnecessarily.

This PR:

- Makes the `NSLinguisticTagger` used by `PersonName` a static variable so it is only created once.
- Removes some `Swift 4.2 migrator` functions that really didn't do much apart from converting between structs and their raw values.

### Testing

1. Run the app
2. Go to contacts
3. Scroll
4. Check that users without profile pictures have correct initials

### Checklist


caldrian (approved):
👌

agisilaos (approved):

PR #1849, opened 2024-08-26
refactor: create FeatureConfigRepository - WPB-8740

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8740" title="WPB-8740" target="_blank">TaskWPB-8740</a>  [iOS] Create FeatureConfigRepository
  </td></table>
  <br />
 

### Key Points

`FeatureConfigRepository` acts as an in-between layer between the app’s business logic and data storage.
Its main responsability is to pull feature configs from backend and store them locally.
It provides an `AsyncStream`  that acts as a callback/delegate to let the business logic layer (interactor / usecase) know right away when a config has been stored locally so any business logic can be done in that lower layer (handling a feature change for instance).

The following code was not included in the repository as this is related to some presentation logic (controllers and routers observe this notification) that should be handled by a lower object such as the presentation layer (e.g view model in clean architecture).

```
private func notifyChange(_ change: FeatureChange) {
        NotificationCenter.default.post(name: .featureDidChangeNotification, object: change)
}
```

### Testing

- Testing feature configs are successfully pulled and stored locally
- Testing feature config can be successfully fetched
- Testing needsToNotifyUser flag is correctly set and correctly fetched

---

### Checklist

---


WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -0,0 +1,352 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+import WireAPI
+import WireDataModel
+
+/// Facilitate access to feature configs related domain objects.
```
samwyndham:
```suggestion
/// Facilitates access to feature configs related domain objects.
```

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -0,0 +1,352 @@
...
+
+protocol FeatureConfigRepositoryProtocol {
+
+    /// Pull feature configs from the server and store them locally.
```
samwyndham:
```suggestion
    /// Pulls feature configs from the server and stores them locally.
```

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -0,0 +1,352 @@
...
+    ///
+    /// The `AsyncStream` produces a `FeatureState` value right after the feature is stored locally.
+    /// so actions can be triggered right away at a lower level callsite (e.g an interactor).
+    /// This will allow a user to be notified immediately even if the stream fails afterwards.
+    ///
+    /// - Returns: An async stream of `FeatureState`.
+
+    func pullFeatureConfigs() -> AsyncThrowingStream<FeatureState, Error>
+
+    /// Fetch a feature config locally.
```
samwyndham:
```suggestion
    /// Fetches a feature config locally.
```

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -0,0 +1,352 @@
...
+        featureConfigsAPI: any FeatureConfigsAPI,
+        context: NSManagedObjectContext
+    ) {
+        self.featureConfigsAPI = featureConfigsAPI
+        self.context = context
+    }
+
+    // MARK: - Public
+
+    func pullFeatureConfigs() -> AsyncThrowingStream<FeatureState, Error> {
```
samwyndham: I'm not sure that returning an `AsyncThrowingStream` is necessary here vs `[FeatureState]`.

Is there anything that takes much time between getting each `FeatureState` - for example a network request between each?

Or is this design mostly about enabling the caller to get a `FeatureState` even if an error is thrown later? How likely are errors to be thrown?

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -0,0 +1,352 @@
...
+    func fetchNeedsToNotifyUser(forFeatureName name: Feature.Name) async throws -> Bool {
+        await context.perform { [context] in
+            let feature = Feature.fetch(name: name, context: context)
+            return feature?.needsToNotifyUser ?? false
+        }
+    }
+
+    func storeNeedsToNotifyUser(_ notifyUser: Bool, forFeatureName name: Feature.Name) async throws {
+        await context.perform { [context] in
+            let feature = Feature.fetch(name: name, context: context)
```
samwyndham: I feel this should throw if there is no feature

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -0,0 +1,352 @@
...
+            }
+
+            return LocalFeature(status: feature.status, config: nil)
+        }
+    }
+
+    func fetchNeedsToNotifyUser(forFeatureName name: Feature.Name) async throws -> Bool {
+        await context.perform { [context] in
+            let feature = Feature.fetch(name: name, context: context)
+            return feature?.needsToNotifyUser ?? false
```
samwyndham: I'm not what should happen here but it feels strange to fallback to `false`. Perhaps it should throw? Or return optional `Bool`?

Is is a correct behavior to check this when the Feature hasn't been persisted?

samwyndham (commented):
Just an initial pass with some comments seeking clarification

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
jullianm: No network request between each, it was mostly designed as a callback mechanism to let the caller know right away when a feature has been processed and take action consequently (even if an error is thrown later as you mentioned). I wanted to stick to how it is currently designed in legacy codebase (as soon as a feature is stored it updates some UI related stuff) and process to the next one.

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
jullianm: It was designed like this in legacy codebase so I'm guessing if we fail fetching the `needsToNotifyUser` flag we fallback to false and don't let the user know anything about it. But since the repository should not be responsible of handling that kind of logic, I agree we should just throw here so I'll make the change, thanks

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
jullianm: yep, will fix it, thx

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
samwyndham: Oh I see. @johnxnguyen maybe you have an opinion here? Does the legacy behavior make sense to you?

johnxnguyen: We certainly need a way to inform consumers of changes in data, but we haven't really discussed this topic yet. I think we should always strive to design APIs that are simple and intuitive, and the approach that makes the most sense to me (at the moment) is to separate the "pull" and the "inform" functionality into different calls, something like:

```swift
func observeFeatureStates() -> AnyPublisher<FeatureState, Never>
func pullFeatureConfigs() async throws
```

This way, if the consumer needs to, they can first observe the changes, then pull the feature configs. There is also the possibility that the consumer only wants to pull without observing, or that the observing happens once early on and pulling happens multiple times later on.

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -0,0 +1,352 @@
...
+    /// - Parameter name: The feature name.
+    /// - Returns: `true` if user should be notified.
+
+    func fetchNeedsToNotifyUser(forFeatureName name: Feature.Name) async throws -> Bool
+
+    /// Stores a flag indicating whether the user should be notified of a given feature.
+    /// - Parameter notifyUser: Whether the user should be notified for a given feature.
+    /// - Parameter name: The name of the feature to set the flag for.
+
+    func storeNeedsToNotifyUser(_ notifyUser: Bool, forFeatureName name: Feature.Name) async throws
```
johnxnguyen: I think we can improve the way we keep track of needing to notify the user of things that happen in the app.  For instance, we could introduce a new entity `Alert` and whenever there is a change we can create an instance in the database. It doesn't only have to be for feature config changes, but could be for anything. This would remove the need to acknowledge changes by resetting a flag. Anyway, it's something we can consider later.

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
jullianm: Good idea, it makes sense. However, I would not introduce any Combine code since we can still use async / await code to leverage this behavior while remaining consistent in our refactoring approach. I will give it a try

jullianm: @johnxnguyen my bad, I thought AsyncStream could behave quite similarly as a publisher but for our purpose that is we want to keep our stream 'alive' and receive values over time, a publisher is more suitable. So did the fix you mentioned  🙏

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
jullianm: Yes I guess it could be a good way to regroup all kind of changes into a single category although I think we would still need to handle these alerts from our repositories. This is a just a thought but we could have for instance a BaseRepository (base class for all repositories) that would continuously check for alerts and propagate them to their corresponding interactors and then display them to the user.

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
johnxnguyen: We could also have a long lived task with an infinite stream, whatever makes most sense

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -0,0 +1,363 @@
...
+
+    func pullFeatureConfigs() async throws -> [FeatureState]
+
+    /// Observes feature states.
+    ///
+    /// Each time `pullFeatureConfigs()` is called, a feature config is
+    /// stored locally and a new `FeatureState` value is produced by the publisher.
+    /// It allows the user to be notified of any feature changes over time.
+    ///
+    /// ⚠️ Use this method before calling `pullFeatureConfigs` to receive all emitted values.
```
samwyndham: I recommend using `- Warning:` below the `- Returns:`:

```swift
/// - Warning: Use this method before calling `pullFeatureConfigs` to receive all emitted values.
```

That way we see a nice warning in docs:
Screenshot 2024-08-30 at 15 20 23

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
samwyndham: Unresolving to remind you to change this if you plan to. Feel free to resolve this comment yourself if not.

samwyndham (approved):
Nice work 🍾

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -0,0 +1,363 @@
...
+        featureConfigsAPI: any FeatureConfigsAPI,
+        context: NSManagedObjectContext
+    ) {
+        self.featureConfigsAPI = featureConfigsAPI
+        self.context = context
+    }
+
+    // MARK: - Public
+
+    func pullFeatureConfigs() async throws -> [FeatureState] {
```
johnxnguyen: Do we still need to return something here? In other repositories we usually don't return anything from a pull, and when we want to fetch an object there is a separate fetch method which takes it from the local store.

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
jullianm: fixed now, thx!

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
jullianm: Nice, I didn't know about this one. thx, fixed in last commit

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
jullianm: No, I forgot to remove it, fixed in last commit thx

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
jullianm: what do you mean by long lived task ? I tried using the AsyncStream but we need to call `finish()` at some point to produce elements and after the iteration is over it will not produce any elements ever so I'm definitely interested if there's a way to do the same thing as the publisher with a task

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -0,0 +1,363 @@
...
+            do {
+                try await storeFeatureConfig(featureConfig)
+
+                if let featureState = try await getFeatureState(
+                    forFeatureConfig: featureConfig
+                ) {
+                    featureStateSubject.send(featureState)
+                }
+            } catch {
+                throw FeatureConfigRepositoryError.failedToStoreConfigLocally(error)
```
netbe: so if we fail to store one config we abort on the others, is it intended?

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -0,0 +1,363 @@
...
+    func storeNeedsToNotifyUser(_ notifyUser: Bool, forFeatureName name: Feature.Name) async throws {
+        try await context.perform { [self] in
+            let feature = try fetchFeature(withName: name)
+            feature.needsToNotifyUser = notifyUser
+        }
+    }
+
+    // MARK: - Private
+
+    func fetchFeature(withName name: Feature.Name) throws -> Feature {
```
netbe:
```suggestion
    private func fetchFeature(withName name: Feature.Name) throws -> Feature {
```

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -0,0 +1,363 @@
...
+
+            case .mls(let mLSFeatureConfig):
+
+                try updateOrCreate(
+                    featureName: .mls,
+                    isEnabled: mLSFeatureConfig.status == .enabled,
+                    config: mLSFeatureConfig.toDomainModel()
+                )
+
+            case .mlsMigration(let mLSMigrationFeatureConfig):
```
netbe, line 306:
```suggestion
            case .mlsMigration(let mlsMigrationFeatureConfig):
```

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -0,0 +1,363 @@
...
+
+                try updateOrCreate(
+                    featureName: .mlsMigration,
+                    isEnabled: mLSMigrationFeatureConfig.status == .enabled,
```
netbe, line 310:
```suggestion
                    isEnabled: mlsMigrationFeatureConfig.status == .enabled,
```

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -0,0 +1,363 @@
...
+                    config: mLSMigrationFeatureConfig.toDomainModel()
```
netbe, line 311:
```suggestion
                    config: mlsMigrationFeatureConfig.toDomainModel()
```

WireDomain/Sources/WireDomain/Repositories/ModelMappings/FeatureConfigModelMappings.swift
```diff
@@ -0,0 +1,99 @@
...
+        .init(
+            protocolToggleUsers: Array(protocolToggleUsers),
+            defaultProtocol: defaultProtocol == .mls ? .mls : .proteus,
+            allowedCipherSuites: allowedCipherSuites.map {
+                .init(rawValue: $0.rawValue)!
+            },
+            defaultCipherSuite: .init(rawValue: defaultCipherSuite.rawValue)!,
+            supportedProtocols: Set(supportedProtocols.map {
+                .init(rawValue: $0.rawValue)!
+            })
```
netbe, line 40: just to understand, we transform supportedProtocols to supportedProtocols, it's too different types any risk to get nil here?

netbe (commented):
@jullianm maybe I missed it, but where is the AsyncStream from the description, was it replaced by Combine publisher?

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -0,0 +1,363 @@
...
+    /// - Parameter type: The type of config to retrieve.
+    /// - Returns: A `LocalFeature` object with a status and a config (if any).
+
+    func fetchFeatureConfig<T: Decodable>(withName name: Feature.Name, type: T.Type) async throws -> LocalFeature<T>
+
+    /// Fetches a flag indicating whether the user should be notified of a given feature.
+    /// - Parameter name: The feature name.
+    /// - Returns: `true` if user should be notified.
+
+    func fetchNeedsToNotifyUser(forFeatureName name: Feature.Name) async throws -> Bool
```
netbe: I don't see what type is Feature.Name but if it's a concrete type we can shorten the signature
```suggestion
    func fetchNeedsToNotifyUser(for name: Feature.Name) async throws -> Bool
```

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
jullianm: The error is thrown when a a config could not be encoded which should be an edge case since the passed parameter is of type any Codable. In any cases, you're right that this might be not ideal to abort the others if that happens. I will provide something else to avoid that, thanks 👍

jullianm:
@netbe yes it was replaced with a Combine publisher following John's comment on keeping the API simple and intuitive by providing a separated observable mechanism apart. At first, the `pullFeatureConfigs` method would store the config locally and provides an AsyncStream like a callback mechanism to inform callsite of a feature state right after the config was stored locally. Now, we give the ability to observe any features changes over time in a separated, dedicated method.

WireDomain/Sources/WireDomain/Repositories/ModelMappings/FeatureConfigModelMappings.swift
jullianm: I will provide an exhaustive switch to avoid this, so if we add a new case they will be a compilation error. thanks

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
jullianm: it's an enum, I feel like having an explicit parameter name gives more clarity at call site:

`fetchNeedsToNotifyUser(forFeatureName: .fileSharing)` rather than `fetchNeedsToNotifyUser(for: .fileSharing)` . Looking at the second version we don't right away know what is fileSharing. I'm ok doing both just thinking out loud here

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
jullianm: fixed

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
jullianm: fixed

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
jullianm: fixed thx

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
jullianm: fixed

WireDomain/Sources/WireDomain/Repositories/ModelMappings/FeatureConfigModelMappings.swift
jullianm: fixed thx

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -233,94 +230,100 @@ final class FeatureConfigRepository: FeatureConfigRepositoryProtocol {
             )
 
         case .unknown(let featureName):
-            assertionFailure("Unknown feature: \(featureName)")
+            logger.info(
```
netbe:
```suggestion
            logger.warn(
```
as it should not happen right

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
```diff
@@ -233,94 +230,100 @@ final class FeatureConfigRepository: FeatureConfigRepositoryProtocol {
...
+                updateOrCreate(
                     featureName: .selfDeletingMessages,
                     isEnabled: selfDeletingMessagesFeatureConfig.status == .enabled,
                     config: selfDeletingMessagesFeatureConfig.toDomainModel()
                 )
 
             case .unknown(let featureName):
-                assertionFailure("Unknown feature: \(featureName)")
+
+                logger.info(
```
netbe:
```suggestion
                logger.warn(
```

netbe (approved):
LGTM;) just a minor suggestion on logging level, feel free to merge after

WireDomain/Sources/WireDomain/Repositories/FeatureConfigRepository.swift
netbe: my rule would be if you have a string, it makes sense to be explicit as you don't know what it represents.
but in case of an enum you will know what it represents because its type give you the information

See https://www.swift.org/documentation/api-design-guidelines/#conventions

jullianm: fixed

PR #1855, opened 2024-08-27
fix: can't send messages on API v1 - WPB-10496

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10496" title="WPB-10496" target="_blank">BugWPB-10496</a>  Client experiencing issues with new iOS version
  </td></table>
  <br />
 

### Issue

Given an old backend that only supports API v1, when a user tries to send a message, then it will fail and expire.

### Cause

In API v4, the `MessageSendingStatus` payload that is returned from endpoints that send proteus messages got a new required field `failed_to_confirm_clients`. Although this field was added in v4, it was also added to all previous versions on the cloud backend, but would always be an empty set. As a result of this, the iOS client could in fact parse the response expecting the field to be there and it wouldn't fail for API versions before v4. This is exactly what we were doing.

For an old backend though whose max API version is v1 (but could also be v2 or v3), this field would not be present. When using the public iOS client, that always expected the `failed_to_confirm_clients` field, it would fail with a decoding error and then expire the message.

### Solution

Go through the `MessageAPI` and use decode the correct response payloads for all the endpoints:
- MessageSendingStatusV0
- MessageSendingStatusV1
- MessageSendingStatusV4

### Testing

Log in to an old v1 backend and send a message.

---

### Checklist

---

### DIFF
--- a/wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
+++ b/wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
@@ -197,15 +197,18 @@ class MessageAPIV1: MessageAPIV0 {
         let response = await httpClient.send(request)
 
         if response.httpStatus == 412 {
-            guard
-                let messageSendingStatus = Payload.MessageSendingStatus(response, decoder: .defaultDecoder)
-            else {
+            guard let messageSendingStatus = Payload.MessageSendingStatusV1(
+                response,
+                decoder: .defaultDecoder
+            ) else {
                 throw NetworkError.errorDecodingResponse(response)
             }
-            throw NetworkError.missingClients(messageSendingStatus, response)
+
+            throw NetworkError.missingClients(messageSendingStatus.toAPIModel(), response)
+
         } else {
-            let payload: Payload.MessageSendingStatus = try mapResponse(response)
-            return (payload, response)
+            let payload: Payload.MessageSendingStatusV1 = try mapResponse(response)
+            return (payload.toAPIModel(), response)
         }
     }
 
@@ -236,15 +239,18 @@ class MessageAPIV1: MessageAPIV0 {
         let response = await httpClient.send(request)
 
         if response.httpStatus == 412 {
-            guard
-                let messageSendingStatus = Payload.MessageSendingStatus(response, decoder: .defaultDecoder)
-            else {
+            guard let messageSendingStatus = Payload.MessageSendingStatusV1(
+                response,
+                decoder: .defaultDecoder
+            ) else {
                 throw NetworkError.errorDecodingResponse(response)
             }
-            throw NetworkError.missingClients(messageSendingStatus, response)
+
+            throw NetworkError.missingClients(messageSendingStatus.toAPIModel(), response)
+
         } else {
-            let payload: Payload.MessageSendingStatus = try mapResponse(response)
-            return (payload, response)
+            let payload: Payload.MessageSendingStatusV1 = try mapResponse(response)
+            return (payload.toAPIModel(), response)
         }
     }
 }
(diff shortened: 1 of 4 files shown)


wire-ios-request-strategy/Sources/Protocols/OTREntity.swift
```diff
@@ -149,13 +149,26 @@ extension OTREntity {
                 in: context
             )
 
-        case .v1, .v2, .v3, .v4, .v5, .v6:
-            guard let payload = Payload.MessageSendingStatus(response) else {
+        case .v1, .v2, .v3:
+            guard let payload = Payload.MessageSendingStatusV1(
+                response,
+                decoder: .defaultDecoder
```
samwyndham, line 155: Is it intended that a `decoder` is passed in this version but not the `MessageSendingStatusV4` version?

wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
```diff
@@ -265,6 +271,87 @@ class MessageAPIV4: MessageAPIV3 {
     override var apiVersion: APIVersion {
         .v4
     }
+
+    private let protobufContentType = "application/x-protobuf"
+
+    override func broadcastProteusMessage(message: any ProteusMessage) async throws -> (Payload.MessageSendingStatus, ZMTransportResponse) {
```
samwyndham, line 277: Are these overriden methods be tested in tests?

wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
```diff
@@ -265,6 +271,87 @@ class MessageAPIV4: MessageAPIV3 {
...
+            ) else {
+                throw NetworkError.errorDecodingResponse(response)
+            }
+
+            throw NetworkError.missingClients(messageSendingStatus.toAPIModel(), response)
+
+        } else {
+            let payload: Payload.MessageSendingStatusV4 = try mapResponse(response)
+            return (payload.toAPIModel(), response)
+        }
```
samwyndham, line 310: optional: Personally I would extract this duplicate code from these two methods to a local private method although I'm sure there are arguments against making this too dry.

KaterinaWire (approved):

samwyndham (approved):
Nice work finding the underlying cause!

samwyndham:
> Log in to an old v1 backend and send a message.

As far as I can tell the implementation is changed somewhat on V4 and above backends right? Should this also be tested on those version backends?

wire-ios-request-strategy/Sources/Protocols/OTREntity.swift
johnxnguyen: no probably just a copy paste mistake, but actually the default used anyway when no decoder is passed.

wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
```diff
@@ -265,6 +271,87 @@ class MessageAPIV4: MessageAPIV3 {
     override var apiVersion: APIVersion {
         .v4
     }
+
+    private let protobufContentType = "application/x-protobuf"
+
+    override func broadcastProteusMessage(message: any ProteusMessage) async throws -> (Payload.MessageSendingStatus, ZMTransportResponse) {
+        let path = "/broadcast/proteus/messages"
```
samwyndham, line 278: The path has changed from the current implementation (/broadcast/otr/messages). Just a sanity check that we want this.

wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
johnxnguyen: There are no existing tests for `MessageAPI`, and I'm thinking that since it will soon be replaced in the refactor then perhaps it's not worth writing a whole test suite for the object. But I'm also happy to in a separate PR if you think it's worth it.

wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
johnxnguyen: Yes I would too, but this `MessageAPI` will be replaced with something in `WireAPI` so I don't think it's worth doing just for the short term.

wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
johnxnguyen: Yes the path changed from v0 to v1.

wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
samwyndham: Cool. Makes sense.

wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
samwyndham: I was thinking we could just copy and paste and change API versions so give us some temporary coverage, but if there aren't existing tests already then the overhead is probably too high. 

I would say it would only be worth doing it if those tests can be written in a way that they don't break with the refactor. That way they can give some protection when integrating the new code. If this is not the case then I don't think it is worth the effort. Feel free to close this comment.

wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
samwyndham: Oh yes I see. I was looking at v0 not v1 implementation.

wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
johnxnguyen: With the refactor we'll be rewriting this whole API and ensuring it has comprehensive tests.

PR #1858, opened 2024-08-28
fix: incorrect error message when user cancels sending asset - WPB-9932

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9932" title="WPB-9932" target="_blank">BugWPB-9932</a>  [iOS] Cancelling sending an asset message gives incorrect failure reason (connectivity issues)
  </td></table>
  <br />
 

### Issue

When the user cancels sending an an asset (such as a file or image etc) we show the default error message for when a message can not be sent:

> Message could not be sent due to connectivity issues

This PR:

- introduces changes to inspect the `transferState` of `ZMAssetClientMessage` messages and shows a more precise message in cases where the upload was cancelled: _Message was not send as the upload was cancelled._
- Updates `ZMAssetClientMessage.expire` to only set `transferState` to `uploadingFailed` in cases where the transfer state is currently `uploading`. This was necessary to have the correct `transferState` when deciding the error message.
- Adjusts the trailing anchor between `MessageToolboxView` and its `MessageSendFailureView` subview to ensure there is padding on the trailing edge. Otherwise long messages run right to the edge which is not pretty:

#### Screenshots

| Cancellation message before | Cancellation message after |
| --- | --- |
| Screenshot 2024-08-23 at 14 34 43 |  Screenshot 2024-08-28 at 10 28 47 |

| Long message before | Long message after |
| --- | --- |
| Screenshot 2024-08-28 at 10 26 51 |  Screenshot 2024-08-28 at 10 28 47 |

#### ⚠️ Update ⚠️ 

As suggested by @johnxnguyen [here](https://github.com/wireapp/wire-ios/pull/1858#discussion_r1736056304) I've somewhat changed the implementation of this fix although kept most the changes already mentioned above.

- `MessageSendFailure` has been replaced by `ExpirationReason`. This is implemented in obj-c code to please the compiler as suggested in this [Swift forums comment](https://forums.swift.org/t/swift-objc-enum-interop/71728/3).
- `ZMMessage.expire()` is replaced with `ZMMessage.expire(withReason:)` and additional reasons such as `cancelled` have been added.

### Testing

- Performs steps outlined in the ticket.
- Try causing messages to be not sent by various means (i.e. no internet , cancelling) to test whether this fix has impacted something else.

### Checklist


wire-ios/Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -512,6 +512,7 @@
 "content.system.message_delivered_timestamp" = "Delivered";
 "content.system.message_read_timestamp" = "Seen";
 "content.system.failedtosend_message.general_reason" = "Message could not be sent due to connectivity issues.";
+"content.system.failedtosend_message.user_cancelled_upload_reason" = "Message was not send as the upload was cancelled by the user.";
```
johnxnguyen: It seems like there's a typo:

```suggestion
"content.system.failedtosend_message.user_cancelled_upload_reason" = "Message was not sent as the upload was cancelled by the user.";
```

Also, it feels kind of robotic to say "...was cancelled by the user.". Wouldn't it be more succinct to say something like:

```
upload cancelled
```

or

```
Message was not sent as the upload was cancelled.
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/MessageToolboxDataSource.swift
```diff
@@ -114,14 +112,18 @@ final class MessageToolboxDataSource {
...
         // 2) Failed to send
         else if failedToSend && isSentBySelfUser {
-            var detailsString: String
+            typealias Message = L10n.Localizable.Content.System.FailedtosendMessage
 
+            let detailsString: String
             switch message.failedToSendReason {
             case .unknown, .none:
-                detailsString = FailedToSendMessage.generalReason
+                let userCancelled = message.fileMessageData?.transferState == .uploadingCancelled
```
johnxnguyen: I'm wondering if instead of inspecting the transfer state here, couldn't we do it at the time of failure and set an appropriate `failedToSendReason`, such as

```
.uploadCancelled
```

wire-ios/Wire-iOS/Resources/Base.lproj/Localizable.strings
samwyndham: Thanks for seeing the typo 🤦!

Regarding the copy I asked Wolfgang to provide it - the current copy is taken from some designs for android. `Upload cancelled` doesn't work because it might be displayed twice in the view.. See screenshots above. However, I agree `Message was not sent as the upload was cancelled.` is better. I will make a suggestion to Wolfgang.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/MessageToolboxDataSource.swift
samwyndham: That would certainly make more sense. For some reason I thought that `failedToSendReason` is synced across devices. Just to confirm this is only local and so I can add more cases (as long as I don't change their int values) right?

wire-ios/Wire-iOS/Resources/Base.lproj/Localizable.strings
samwyndham: Changed the copy to `Message was not sent as the upload was cancelled.`

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/MessageToolboxDataSource.swift
samwyndham: I've changed implementation following your suggestion

wire-ios-data-model/Source/Model/Message/ConversationMessage.swift
```diff
@@ -31,12 +31,6 @@ public enum ZMDeliveryState: UInt {
     case failedToSend = 5
 }
 
-@objc
```
samwyndham, line 34: This has been replaced by `ExpirationReason` as it is not necessarily a `failure`

wire-ios-data-model/Source/Model/Message/ZMMessage+Internal.h
```diff
@@ -32,6 +32,13 @@
 @class ZMReaction;
 @class ZMClientMessage;
 
+typedef NS_ENUM(NSInteger, ZMExpirationReason) {
+    ZMExpirationReasonOther = 0,
```
samwyndham, line 36: `other` replaces `unknown` as usually the reason for expiring is known but we don't need so much granularity.

agisilaos (approved):

netbe (approved):
LGTM;)

PR #1859, opened 2024-08-28
chore:  Integrate `HTMLString` and `swift-protobuf` using SPM - WPB-10812

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10812" title="WPB-10812" target="_blank">Sub-taskWPB-10812</a>  [iOS] Integrate `HTMLString` and `swift-protobuf` using SPM instead of Carthage
  </td></table>
  <br />
 

### Issue

Currently `HTMLString` and `swift-protobuf` packages are integrated using `Carthage`. However, `Carthage` has an open issue with `Xcode 16 beta 6` which can cause builds of the above 2 packages to fail -  https://github.com/Carthage/Carthage/issues/3375#issuecomment-2302269918. @caldrian was able to get builds working by modifying the the packages themselves. However as both these packages can be integrated via `SPM`, I think it is worth doing that. As far as I'm away we are migrating away from `Carthage` to `SPM` to I think this makes sense.

This PR reintegrates the following packages using `SPM`:

- https://github.com/wireapp/HTMLString
- https://github.com/apple/swift-protobuf (edited)~https://github.com/wireapp/swift-protobuf~ 

These are automatically integrated as `static` libraries by Xcode. As such I'm only linking them in the targets they are used - `WireLinkPreview` & `WireProtos` - not in the `Wire-iOS` app target - this is correct right?

~One thing of note is that we have two versions of `swift-protobuf` used within our project - one we define https://github.com/wireapp/swift-protobuf) and one that is used by some 3rd party dependencies (https://github.com/apple/swift-protobuf).~ 

As far as I can tell things are working correctly, but please flag if you have any concerns.

I have a playground build [here](https://github.com/wireapp/wire-ios/actions/runs/10612511590) just to ensure that release builds can be submitted to TestFlight - hopefully this will flag any issues with duplicate symbols.

**Note**

While this PR migrates `swift-protobuf` to use the upstream version (from apple), it keeps `HTMLString` using the `wireapp` fork as it is 16 commits ahead of upstream: https://github.com/alexisakers/HTMLString/compare/master...wireapp:master

### Testing

Nothing to test

### Checklist


wire-ios-mono.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -237,11 +246,11 @@
       },
       {
         "package": "SwiftProtobuf",
-        "repositoryURL": "https://github.com/apple/swift-protobuf.git",
+        "repositoryURL": "https://github.com/wireapp/swift-protobuf",
         "state": {
           "branch": null,
-          "revision": "e17d61f26df0f0e06f58f6977ba05a097a720106",
```
samwyndham, line 243: This is a little interesting. This is from the workspace. This was previously pointing to the version used by a `DataDog` in `WireAnalytics` which I don't believe we are explicitly using. However now it points to the version we are explicitly importing which was the same version defined in the Cartfile. I _think_ all is still OK.

wire-ios/Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -3249,8 +3243,8 @@
...
 		BFFE943B1E7839D10025AD75 /* ConversationRenamedCellTests.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = ConversationRenamedCellTests.swift; sourceTree = "<group>"; };
 		BFFE943D1E7839EB0025AD75 /* CoreDataSnapshotTestCase.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = CoreDataSnapshotTestCase.swift; sourceTree = "<group>"; };
+		CB4870F12C7F4FE5001E9151 /* WireTransportSupport.framework */ = {isa = PBXFileReference; explicitFileType = wrapper.framework; path = WireTransportSupport.framework; sourceTree = BUILT_PRODUCTS_DIR; };
 		CB51204A2C6F6C84000C8FEC /* TabBarChangeHandler.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TabBarChangeHandler.swift; sourceTree = "<group>"; };
-		CB79790F2C737AB7006FBA58 /* WireTransportSupport.framework */ = {isa = PBXFileReference; explicitFileType = wrapper.framework; path = WireTransportSupport.framework; sourceTree = BUILT_PRODUCTS_DIR; };
```
samwyndham, line 3253: Xcode made some auto changes to the project putting `WireTransportSupport` in a `Recovered References` folder when integrating HTMLString - I'm not sure why. I had to clean things up slightly, hence the changes here.

caldrian:
We had a similar PR a while ago and it was rejected by @netbe .
https://github.com/wireapp/wire-ios/pull/1700
I wonder if this PR is fine now or should be delayed as well.

samwyndham:
> We had a similar PR a while ago and it was rejected by @netbe . #1700 I wonder if this PR is fine now or should be delayed as well.

@netbe Is it OK in this case as I'm still pointing to wires fork of protobuf?

netbe:
> > We had a similar PR a while ago and it was rejected by @netbe . #1700 I wonder if this PR is fine now or should be delayed as well.
> 
> @netbe Is it OK in this case as I'm still pointing to wires fork of protobuf?

I think we can drop the fork, the issue was if you do update the dependency it should include the privacy manifest.

samwyndham:
@caldrian @netbe OK this PR now also uses the latest `swift-protobuf` and the original repository not a fork

caldrian (approved):

johnxnguyen (approved):

PR #1862, opened 2024-08-29
refactor: create conversation repository - WPB-8701

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8701" title="WPB-8701" target="_blank">TaskWPB-8701</a>  [iOS] Create ConversationRepository
  </td></table>
  <br />
 

### Key points

`ConversationRepository` acts as an in-between layer between the app’s business logic and data storage.
It contains methods to pull conversations, delete conversations, remove a self user from conversations and mark conversations as fetched.

Since the repository contains many CoreData related work (and got pretty massive), the PR introduces a new dedicated object called `ConversationLocalStore` that encapsulates all storage conversation related operations (using an injected context for now).
`ConversationLocalStore` conforms to the `ConversationLocalStoreProtocol` which makes the layer easily testable.

This enables us to isolate the storage layer in the same way we isolated the API layer (e.g `ConversationsAPI`) so that now the repository will not handle storage operations directly but rely on a dedicated local store object.

### Testing

Following use cases are covered:
- Pulling and storing conversations of different types (group, self, connection, 1:1)
- Deleting conversations
- Removing self user from conversations
- Marking conversations as fetched

### Checklist

---


WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore+Group.swift
```diff
@@ -0,0 +1,100 @@
...
+
+/// An extension that encapsulates storage operations related to conversation group (user, role, members).
+
+extension ConversationLocalStore {
+
+    // MARK: - User & Role
+
+    func fetchUserAndRole(
+        from remoteConversationMember: WireAPI.Conversation.Member,
+        for localConversation: ZMConversation
```
johnxnguyen, line 30:
```suggestion
        for remoteConversationMember: WireAPI.Conversation.Member,
        in localConversation: ZMConversation
```

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore+MLS.swift
```diff
@@ -0,0 +1,109 @@
...
+    ) {
+        guard let newMessageProtocol = remoteConversation.messageProtocol else {
+            return
+        }
+
+        localConversation.messageProtocol = newMessageProtocol.toDomainModel()
+    }
+
+    func updateMessageProtocol(
+        from remoteConversation: WireAPI.Conversation,
```
johnxnguyen, line 43: It looks like you can simply pass in the message protocol instead of the whole conversation.

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationModelMappings.swift
```diff
@@ -0,0 +1,49 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireAPI
+import WireDataModel
+
+extension WireAPI.MLSCipherSuite {
+
+    func toDomainModel() -> WireDataModel.MLSCipherSuite {
+        .init(rawValue: rawValue)!
```
samwyndham: I see this can't fail in its current state but I'm wondering if it is worth adding safety here incase one of the enums changes. Consider adding a test for this that at a minimum asserts that all `WireAPI.MLSCipherSuite` cases can be converted to `WireDataModel.MLSCipherSuite`.

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationModelMappings.swift
```diff
@@ -0,0 +1,49 @@
...
+    }
+}
+
+extension WireAPI.ConversationAccessRoleLegacy {
+
+    func toDomainModel() -> WireDataModel.ConversationAccessRole? {
```
samwyndham: Does this need to be `optional`? Can you do:

```swift
    func toDomainModel() -> WireDataModel.ConversationAccessRole {
        switch self {
        case .private:
            .private
        case .team:
            .team
        case .activated:
            .activated
        case .nonActivated:
            .nonActivated
        }
    }
```

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
```diff
@@ -0,0 +1,399 @@
...
+        self.context = context
+    }
+
+    // MARK: - Public
+
+    public func storeConversation(
+        _ conversation: WireAPI.Conversation,
+        isFederationEnabled: Bool
+    ) async throws {
+        guard let id = conversation.id ?? conversation.qualifiedID?.uuid else {
```
samwyndham, line 98: Is a missing ID a valid state? If no we should do something other than failing silently. Consider, throwing or logging the error and asserting. Also consider doing something in the default case below.

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
```diff
@@ -0,0 +1,399 @@
...
+        default:
+            /// conversation type is nil
+            return
+        }
+    }
+
+    public func deleteConversation(
+        withQualifiedId qualifiedId: WireAPI.QualifiedID
+    ) async throws {
+        let conversation = await context.perform { [context] in
```
samwyndham: Do we only care about group conversations? Maybe we should rename the method to show that?

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
```diff
@@ -0,0 +1,399 @@
...
+
+    /// Stores the specified failed conversation
+    /// - Parameter qualifiedId: The conversation qualified ID.
+
+    func storeFailedConversation(
+        withQualifiedId qualifiedId: WireAPI.QualifiedID
+    ) async
+}
+
+public final class ConversationLocalStore: ConversationLocalStoreProtocol {
```
samwyndham, line 60: There are a lot of ways that these methods can return early. As someone looking at the code I'm not sure if this is expected behavior or if this just handling incorrect states. If it is correct behavior I think it should be documented, perhaps in the doc comments of the methods. If it is incorrect behavior, I think it should either throw or log the error & assert. Please revisit areas in this file and others and clarify expected behavior using throws, doc comments, logging etc

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
```diff
@@ -0,0 +1,399 @@
...
+        guard let mlsService, let groupID else {
+            return
+        }
+
+        try await mlsService.wipeGroup(groupID)
+    }
+
+    public func removeSelfUserFromConversation(
+        withQualifiedId qualifiedId: WireAPI.QualifiedID
+    ) async {
```
samwyndham: I think this should throw if we attempt to delete a user that is not the self user.

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore+Group.swift
```diff
@@ -0,0 +1,100 @@
...
+    ) -> (ZMUser, Role?)? {
```
samwyndham: Consider adding parameter names to the tuple so that callers don't have to access results using `1` etc. - e.g. 

```swift
(user: ZMUser, role: Role?)
```

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore+MLS.swift
```diff
@@ -0,0 +1,109 @@
...
+            }
+        }
+    }
+
+    // MARK: - Self / MLS
+
+    func createOrJoinSelfConversation(
+        from localConversation: ZMConversation
+    ) async throws {
+        guard let context = localConversation.managedObjectContext else {
```
samwyndham: Can you not use the context that belongs to self here? That way it can't be `nil`.

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
jullianm: I try to stick at best to what current legacy code is currently doing to avoid introducing breaking changes since these repo will soon be consumed (from my understanding) by existing code. So according to legacy code (`ConversationRequestStrategy` and `ConversationEventPayloadProcessor`), this only concerned `group` conversations however I can rename it `deleteConversationGroup` to be more explicit.

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
jullianm: This code was moved from legacy class `ConversationEventPayloadProcessor ` to our new architecture. This code is, for the most part of it, not commented so I tried not to change too many things, just refactoring repetitive code, cleaning up code, making it more readable, more organized. 

The local store is just an object that encapsulates the already existing storage logic so right now it is supposed to behave like it was in legacy code. So, to answer your question, based on legacy code (given legacy code is correct), this is the expected behavior.

For instance, in legacy code we never handle any errors, just return early and proceed to the next conversation interation. 

```
    func updateOrCreateConverations(
        from payload: Payload.QualifiedConversationList,
        in context: NSManagedObjectContext
    ) async {
        for payload in payload.found {
            await updateOrCreateConversation(
                from: payload,
                source: .slowSync,
                in: context
            )
        }
    }
```

We do have loggers but from my understanding for now we removed these logs from our repositories.

 I will try to do some digging in order to be more explicit about each method's role.

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore+MLS.swift
jullianm: will do thx

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore+Group.swift
jullianm: yep 👍

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
jullianm: I will check that with backend team, thx

WireDomain/Tests/WireDomainTests/Helpers/XCTestCase+Assertions.swift
```diff
@@ -0,0 +1,122 @@
...
+    /// Assert that a collection is of a certain size.
+    ///
+    /// - Parameters:
+    ///   - collection: The collection to assert.
+    ///   - count: The expected number of elements.
+    ///   - message: The error message to show when the assertion fails.
+    ///   - file: The file name of the invoking test.
+    ///   - line: The line number when this assertion is made.
+
+    func XCTAssertCount(
```
johnxnguyen: Perhaps we can move this to WireTesting since it seems to be duplicated a few times.

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationRepository.swift
```diff
@@ -0,0 +1,175 @@
...
+    func removeSelfUserFromConversations(
+        withQualifiedIds qualifiedIds: Set<WireAPI.QualifiedID>
+    ) async
+
+    /// Marks specified conversations as fetched.
+    /// - Parameter qualifiedIds: The qualified conversations IDs.
+
+    func markConversationsAsFetched(
+        qualifiedIds: Set<WireAPI.QualifiedID>
+    ) async
```
johnxnguyen: The ticket only specifies a need for pulling conversations. Is there a need for these right now?

jullianm: actually no, we're only dealing with slow sync so these are not needed at the moment, I will remove them (stash them for the future) and keep only the pullConversations method. 👍

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
jullianm: resolving this comment, as per John's comment, as we're dealing with slow sync in this PR, we don't need these methods for now, so I removed them

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
jullianm: resolving this comment, as per John's comment, as we're dealing with slow sync in this PR, we don't need these methods for now, so I removed this method, we'll deal with them later as the refactoring progress

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationRepository.swift
jullianm: fixed, thanks, now we only have a pullConversations method, resolving this comment

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationModelMappings.swift
jullianm: fixed

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationModelMappings.swift
jullianm: fixed

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationModelMappings.swift
jullianm: I added a full switch self, if we add a case, compilation will fail

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore+Group.swift
jullianm: fixed

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore+MLS.swift
jullianm: fixed

WireDomain/Tests/WireDomainTests/Helpers/XCTestCase+Assertions.swift
jullianm: fixed, these helpers were moved to WireTesting package

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
jullianm: Actually, we are using the WireAPI.Conversation object which is a new object that was recently created as part of the refactoring plan (see ticket https://wearezeta.atlassian.net/browse/WPB-8696 that is now closed) so we can assume that here we are consuming an object that is reliable.

jullianm: Based on legacy codebase, we're not doing anything here:

```
        guard let conversationType = payload.type.map(BackendConversationType.clientConversationType) else {
            return nil
        }
```

and 

```
        default:
            return nil
```

We just proceed to the next conversation iteration. 
I can add some print statement for now, please note that this ticket is just about slow sync and we will most definitely work again on these objects later on which will enable us to make some adjustments if needed.

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
jullianm: I've added some documentation that hopefully will provide some general context on the different conversation types.

samwyndham: Thanks @jullianm !

samwyndham (approved):
🥳

WireAPI/Sources/WireAPI/Models/Conversation/ConversationType.swift
```diff
@@ -20,7 +20,7 @@ import Foundation
 
 /// The various types of conversations.
 
-public enum ConversationType: Int, Codable {
+public enum ConversationType: Int, Codable, Sendable {
```
netbe, line 23: is it for Swift 6?

WireDomain/Package.swift
```diff
@@ -14,11 +14,19 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.1.0"),
         .package(path: "../SourceryPlugin"),
-        .package(name: "WireAPI", path: "../WireAPI")
+        .package(name: "WireAPI", path: "../WireAPI"),
+        .package(name: "WireTestingPackage", path: "../WireTesting")
```
netbe: probably not related to this PR but could the name of the package be WireTesting ?

WireDomain/Package.swift
```diff
@@ -14,11 +14,19 @@ let package = Package(
...
     ],
     targets: [
         .target(name: "WireDomainPkg", dependencies: ["WireAPI"], path: "./Sources/Package", swiftSettings: swiftSettings),
-        .testTarget(name: "WireDomainPkgTests", dependencies: ["WireDomainPkg"], path: "./Tests/PackageTests"),
+        .testTarget(
+            name: "WireDomainPkgTests",
```
netbe: probably not related to this PR but could the name of the package be WireDomainTests ?

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore+Group.swift
```diff
@@ -0,0 +1,100 @@
...
+            in: context
+        )
+
+        func fetchOrCreateRoleForConversation(
+            name: String,
+            conversation: ZMConversation
+        ) -> Role {
+            Role.fetchOrCreateRole(
+                with: name,
+                teamOrConversation: conversation.team != nil ? .team(conversation.team!) : .conversation(conversation),
```
netbe: there's a helper method in DataModel for this
```suggestion
                teamOrConversation: .fromTeamOrConversation(team: team, conversation: conversation),
```

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore+Group.swift
```diff
@@ -0,0 +1,100 @@
...
+
+        let selfUserRole = fetchUserAndRole(
+            from: members.selfMember,
+            for: localConversation
+        )?.role
+
+        localConversation.updateMembers(otherMembers, selfUserRole: selfUserRole)
+    }
+
+    // MARK: 1:1
```
netbe:
```suggestion
    // MARK: - 1:1
```

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore+Status.swift
```diff
@@ -0,0 +1,134 @@
...
+
+    func updateConversationIfNeeded(
+        localConversation: ZMConversation,
+        fallbackGroupID: MLSGroupID?
+    ) async {
+        let (messageProtocol, mlsGroupID, mlsService) = await context.perform { [self] in
+            (
+                localConversation.messageProtocol,
+                localConversation.mlsGroupID,
+                context.mlsService
```
netbe: @jullianm did you discuss injecting mlsService directly with John? this is kind of technical debt we have that the mlsService is passed from NSManagedObjectContext.

But as we refactor this we can think of injecting it to at least the store directly instead

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
```diff
@@ -0,0 +1,437 @@
...
+    ///
+    /// __What is `Federation` ?__
+    ///
+    /// Businesses that use Wire in their organization have 2 choices, either deploy their teams in the Wire cloud or install an on-prem backend.
+    /// On-prem backends are isolated islands that before could not communicate with users from other on-prem islands or Wire cloud users.
+    /// Federation allows islands to be connected. Communication can be established between users from different backends the same way as on a single backend.
+    ///
+    /// - Parameter remoteConversation: The conversation object received from backend.
+    /// - Parameter removeConversationID: The conversation ID received from backend.
+    /// - Parameter isFederationEnabled: A flag indicating whether a federation is enabled.
```
netbe: we're repeating the definition of Federation, maybe we move this back to a docC within WireDomain and point to it instead

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationRepository.swift
```diff
@@ -0,0 +1,111 @@
...
+
+            for conversation in foundConversations {
+                taskGroup.addTask { [self] in
+                    do {
+                        try await conversationsLocalStore.storeConversation(
+                            conversation,
+                            isFederationEnabled: backendInfo.isFederationEnabled
+                        )
+                    } catch {
+                        throw ConversationRepositoryError.failedToStoreConversation(error)
```
netbe: not sure here, does it break the loop at first error?

WireAPI/Sources/WireAPI/Models/Conversation/ConversationType.swift
jullianm: yes, these are some the "easy" Swift 6 concurrency fixes : marking value types as Sendable.

WireDomain/Package.swift
jullianm: No problem, will do so.

jullianm: We could do these specific changes in the PR but looking at the different Swift packages, I think we should streamline our Swift packages (WireUtilities, WireSystem, WireDomain,  WireTesting..) to the same naming convention. Making it more readable.

So, for instance: 

```
let package = Package(
    name: "WireDomainPackage",
    defaultLocalization: "en",
    platforms: [.iOS(.v15), .macOS(.v12)],
    products: [
        .library(name: "WireDomainPackage", type: .dynamic, targets: ["WireDomainPkg"]),
        .library(name: "WireDomainPackageSupport", type: .dynamic, targets: ["WireDomainPkgSupport"])
    ]

{...}
```

could become:

```
let package = Package(
    name: "WireDomain",
    defaultLocalization: "en",
    platforms: [.iOS(.v15), .macOS(.v12)],
    products: [
        .library(name: "WireDomain", type: .dynamic, targets: ["WireDomain"]),
        .library(name: "WireDomainSupport", type: .dynamic, targets: ["WireDomainSupport"])
    ]

{...}
```

And do the same for all packages.. I think it's better to have a dedicated ticket for this task as it concerns many Swift packages not related to this PR (I can work on this dedicated ticket)

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore+Group.swift
jullianm: fixed, we can actually use the public static func `matching(_ conversation: ZMConversation) -> TeamOrConversation` but it does call the method you suggested. thanks.

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore+Status.swift
jullianm: no we didn't discuss it but it makes sense to inject it directly to the store, I fixed it thx

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationRepository.swift
jullianm: Interestingly enough, it would do nothing because we are throwing inside a child task and not inside the body of `withThrowingTaskGroup` itself. If we were throwing inside the body, it would cancel the group and all of its child tasks but here it will do nothing and continue processing. I'm not a 100% sure but looking at Apple's doc and some online resources, it looks like it. Bottom line, if we want to `actually` throw and propagate the error to the call site, it should be done inside the body of `withThrowingTaskGroup` and not inside the child task.

Anyways, this was removed now that I added back the logger, we are catching the underlying error and logging it, so we don't need to have a throwing func anymore.

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
jullianm: done, I've added some documentation for conversations and federation using docc

WireDomain/Package.swift
netbe: added a follow up ticket here https://wearezeta.atlassian.net/browse/WPB-10980

netbe (approved):

PR #1868, opened 2024-09-02
feat: Use proper MLS removal key with federated 1:1 conversations - WPB-10745

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10745" title="WPB-10745" target="_blank">BugWPB-10745</a>  [iOS] Use proper MLS removal key with federated 1:1 conversations
  </td></table>
  <br />
 

### Issue
[Context:](https://wearezeta.atlassian.net/wiki/spaces/PAD/pages/1360953454/2024-08-20+RFC+Solve+removal+key+problem+in+MLS+federated+1+1+conversations)
There's an issue with MLS 1:1 conversations between users on separate backends.

There is a 50% chance that the conversation is created on either backend (this choice is deterministic, but for the purpose of the user experience, it’s a 50% random chance).
If the person that sends the initial commit doesn’t belong to the backend that is hosting the conversation, they will set the wrong removal key in the conversation
They set the removal key of their own backend, no matter what backend actually hosts the conversation
If this happens, then whenever a client leaves the 1:1 conversation, other clients in the conversation start to see decryption errors for any further message or event in that conversation.

[Solution:](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/792920230/Use+case+establish+1+1+conversation+MLS)
If the MLS conversation is initialized with the right removal key, then this problem doesn’t manifest. 
In order to achieve that, when Alice accepts a connection request, and it gets the conversation metadata before creating the initial commit, the backend would return the removal key of the backend that is hosting the conversation (its own, or backend B’s). 
Alice can then initialize the MLS conversation with the right removal key.

 **Steps:**
1. Fetch the conversation metadata by requesting
`GET /conversations/one2one/{domain}/{userId} `
and persist the metadata and in particular remember the returned values for:
- **MLS group ID**, which is part of the `conversation` metadata 
- Conversation **removal key** under `public_keys` and `removal`

2. Call the createConversation(...) function in CoreCrypto and supply the` MLS group ID` and `removal key` retrieved in the previous step.

**TODO:**

- [x] Manual testing
- [ ] `startProteusToMLSMigration()` - need to check

### Testing

- create 1:1 between users from different backends
- login and logout on different devices

**Result**: no decryption errors

### Checklist

---


wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -452,6 +468,55 @@ public final class MLSService: MLSServiceInterface {
         }
     }
 
+    public func createOneToOne(
```
typfel: The split of responsibility is strange, which introduces code duplication with the pre-existing `createGroup(for:parentGroupID:)`.

Maybe instead do something like

```
establishGroup(for:users:removalKey:) -> createGroup(groupID:removalKeys:)
createGroup(groupID:parentGroupID:) -> createGroup(groupID:removalKeys:)
```

typfel (changes requested):

wire-ios-data-model/Source/MLS/MLSService.swift
KaterinaWire: @typfel I've updated the implementation. There are 2 public methods now:
```
func createGroup(
        for groupID: MLSGroupID,
        parentGroupID: MLSGroupID
    ) async throws -> MLSCipherSuite
```
and
   ```
 func createGroup(
        for groupID: MLSGroupID,
        removalKeys: BackendMLSPublicKeys?
    ) async throws -> MLSCipherSuite
```

And one private that calls CC method:
```
private func createGroup(
        for groupID: MLSGroupID,
        externalSenders: [Data],
        ciphersuite: MLSCipherSuite
    ) async throws -> MLSCipherSuite
```
As a result, there is no code duplication.

**Review by typfel — APPROVED** [review]
🚀

**Review by netbe — APPROVED** [review]
LGTM;)

File: wire-ios-request-strategy/Sources/Request Strategies/MLS/Actions/SyncMLSOneToOneConversationActionHandler.swift [hunk 2]
```diff
@@ -114,6 +114,34 @@ final class SyncMLSOneToOneConversationActionHandler: ActionHandler<SyncMLSOneTo
             action.fail(with: .unknown(status: errorInfo.status, label: errorInfo.label, message: errorInfo.message))
         }
     }
+
+    private func updateOrCreateConversation(
+        action: SyncMLSOneToOneConversationAction,
+        payload: Payload.Conversation,
+        publicKeys: BackendMLSPublicKeys? = nil
+    ) {
+        Task { [action] in
```
caldrian, line 123: I'd prefer to remove this line and declare the func async.
Then on the calling side it's obvious that it contains async code which hasn't completed when this call returns.

caldrian (approved):

wire-ios-request-strategy/Sources/Request Strategies/MLS/Actions/SyncMLSOneToOneConversationActionHandler.swift
KaterinaWire: This would require refactoring the `ActionHandler` class, which I would not do in this PR.

PR #1872, opened 2024-09-03
refactor: process team member update event - WPB-10186

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10186" title="WPB-10186" target="_blank">TaskWPB-10186</a>  [iOS] Process TeamMemberUpdateEvent
  </td></table>
  <br />
 

### Key points

This PR is part of the `quick sync` refactoring plan and is related to processing the multiple events we receive from the backend or the push channel.

Specifically, this PR is about porting the existing implementation of the `TeamMemberUpdate` event.

Legacy code has the following implementation:

```
    private func processUpdatedMember(with event: ZMUpdateEvent) {
        guard event.teamId != nil, let data = event.dataPayload else { return }
        guard let userId = (data[TeamEventPayloadKey.user.rawValue] as? String).flatMap(UUID.init(transportString:)) else { return }
        guard let member = Member.fetch(with: userId, in: managedObjectContext) else { return }
        member.needsToBeUpdatedFromBackend = true
    }

```

New refactored code throws an error when the member could not be found locally.

### Testing

UTs cover the following use case:
- Ensuring the `needsToBeUpdatedFromBackend` flag is set to true after processing the event
- Ensuring an error is thrown when the member could not be found locally.

---

### Checklist

---


WireDomain/Sources/WireDomain/Event Processing/TeamEventProcessor/TeamMemberUpdateEventProcessor.swift
```diff
@@ -32,9 +33,24 @@ protocol TeamMemberUpdateEventProcessorProtocol {
...
 struct TeamMemberUpdateEventProcessor: TeamMemberUpdateEventProcessorProtocol {
 
-    func processEvent(_: TeamMemberUpdateEvent) async throws {
-        // TODO: [WPB-10186]
-        assertionFailure("not implemented yet")
+    enum Error: Swift.Error, Equatable {
+        case failedToFindMember(_ membershipId: UUID)
+    }
+
+    let context: NSManagedObjectContext
```
jullianm: To align with what John did for the RenameEvent [here](https://github.com/wireapp/wire-ios/commit/f3b06e8f640e3e8eb1e66ad6047e103ede1c7886#diff-a6cf300bceb56a8dcc352452381cf8121e41c3aec7b3687b96df76c5ec742799), we are passing the context directly to the processor however since we have repositories that are responsible of fetching / storing data locally, I wonder if we should be passing repositories instead of context (and expose only what is necessary to the processor using a protocol that the repository would conform to).

WireDomain/Sources/WireDomain/Event Processing/TeamEventProcessor/TeamMemberUpdateEventProcessor.swift
```diff
@@ -32,9 +33,24 @@ protocol TeamMemberUpdateEventProcessorProtocol {
 
 struct TeamMemberUpdateEventProcessor: TeamMemberUpdateEventProcessorProtocol {
 
-    func processEvent(_: TeamMemberUpdateEvent) async throws {
-        // TODO: [WPB-10186]
-        assertionFailure("not implemented yet")
+    enum Error: Swift.Error, Equatable {
+        case failedToFindMember(_ membershipId: UUID)
```
samwyndham:
```suggestion
        case failedToFindMember(_ membershipID: UUID)
```

samwyndham: According to https://www.swift.org/documentation/api-design-guidelines/#follow-case-conventions ID should be all uppercase (I think)

WireDomain/Sources/WireDomain/Event Processing/TeamEventProcessor/TeamMemberUpdateEventProcessor.swift
samwyndham: So limit that the processor has knowledge of core data? I think that makes sense in principle - it is nice to prevent core data encroachment!

samwyndham (approved):
Nice work. Agree it is much better to throw here.

WireDomain/Sources/WireDomain/Event Processing/TeamEventProcessor/TeamMemberUpdateEventProcessor.swift
caldrian: There is a SwiftFormat rule for that, but it's dangerous to apply it on an existing codebase.
I would actually prefer to have it enabled, but we would have to be careful.
(it can break Codable for example)
https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#acronyms

WireDomain/Sources/WireDomain/Event Processing/TeamEventProcessor/TeamMemberUpdateEventProcessor.swift
caldrian: Abstracting Core Data has its own difficulties, for example the requirement that entities must only be accessed from the context's queue. Since the protocol probably won't allow you to specify this requirement you would need to declare any method async.
That, however, has it's own side effects, for example that you cannot batch several calls inside one `context.perform`.

Generally I think we should fight less against Apple's SDK. I'm not even a great fan of applying MVVM for UIKit, btw.

caldrian (approved):

WireDomain/Sources/WireDomain/Event Processing/TeamEventProcessor/TeamMemberUpdateEventProcessor.swift
netbe: @jullianm wasn't it because the repository was not ready here ?

jullianm: I guess so. I just pushed some changes to use the dedicated `TeamRepository` instead of the context.

WireDomain/Sources/WireDomain/Repositories/TeamRepository.swift
```diff
@@ -72,6 +77,20 @@ final class TeamRepository: TeamRepositoryProtocol {
...
+        try await context.perform { [context] in
+
+            guard let member = Member.fetch(
+                with: membershipID,
+                in: context
+            ) else {
+                throw TeamRepositoryError.failedToFindTeamMember(membershipID)
+            }
+
+            member.needsToBeUpdatedFromBackend = true
```
netbe: I think we should save here no?
In legacy code, it's not obvious when we save the context. But here I think the repository or its localstorage should have the responsability to save it

WireDomain/Sources/WireDomain/Event Processing/TeamEventProcessor/TeamMemberUpdateEventProcessor.swift
jullianm: I fixed the issue thx

WireDomain/Sources/WireDomain/Repositories/TeamRepository.swift
jullianm: fixed, thx

PR #1877, opened 2024-09-03
fix: video calls fail immediately - WPB-10762

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10762" title="WPB-10762" target="_blank">BugWPB-10762</a>  [iOS] Video calls fail immediately
  </td></table>
  <br />
 

### Issue

The video call in 1:1 ends after a couple of seconds.

 **Solution**
The issues was introduces when we implemented "[1:1 calls over SFT](https://wearezeta.atlassian.net/wiki/spaces/PAD/pages/1314750477/2024-07-29+1+1+calls+over+SFT)".  
Since 1:1 calls should be treated as group calls (if the feature flag for the team is enabled), we implemented a way to drop the call if one (out of two, since it is a 1:1 call) of the participants has already ended it.
One requirement was missing, we need to check that the previous participant count was 2, the new one is 1 and the call is established. This requirement has been added in this PR.

### Testing

- Be logged into a user with e2ei and connected to another e2ei user on web
- Try to video call 1:1

### Checklist

---


samwyndham (approved):
Nice work!

netbe (approved):

PR #1881, opened 2024-09-05
fix: video calls fail immediately - WPB-10762

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10762" title="WPB-10762" target="_blank">BugWPB-10762</a>  [iOS] Video calls fail immediately
  </td></table>
  <br />
 

### Issue

This PR is cherry-picked based on the following PR:

https://github.com/wireapp/wire-ios/pull/1877

**Original PR description:**
The video call in 1:1 ends after a couple of seconds.

**Solution**
The issues was introduces when we implemented "[1:1 calls over SFT](https://wearezeta.atlassian.net/wiki/spaces/PAD/pages/1314750477/2024-07-29+1+1+calls+over+SFT)".
Since 1:1 calls should be treated as group calls (if the feature flag for the team is enabled), we implemented a way to drop the call if one (out of two, since it is a 1:1 call) of the participants has already ended it.
One requirement was missing, we need to check that the previous participant count was 2, the new one is 1 and the call is established. This requirement has been added in this PR.

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
+++ b/wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
@@ -433,7 +433,11 @@ extension WireCallCenterV3 {
     /// Call this method when the callParticipants changed and avs calls the handler `wcall_participant_changed_h`
     func callParticipantsChanged(conversationId: AVSIdentifier, participants: [AVSCallMember]) {
         guard isEnabled else { return }
-        guard !shouldEndCall(conversationId: conversationId, participants: participants) else {
+        let shouldEndCall = shouldEndCall(
+            conversationId: conversationId,
+            previousParticipants: callSnapshots[conversationId]?.callParticipants.members.array ?? [],
+            newParticipants: participants)
+        guard !shouldEndCall else {
             endAllCalls()
             return
         }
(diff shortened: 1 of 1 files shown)


wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
```diff
@@ -478,23 +485,36 @@ extension WireCallCenterV3 {
...
+        newParticipants: [AVSCallMember]
+    ) -> Bool {
         /// We assume that the 2nd participant is the other user, and if the other user's audio state is connecting, the call should end.
-        guard participants.count == 2,
-              participants[1].audioState == .connecting else {
+        guard
+            previousParticipants.count == 2,
+            newParticipants.count == 2,
+            newParticipants[1].audioState == .connecting
+        else {
```
jullianm, line 507: just wondering why the logic differs between MLS and Proteus (why don't we check that newPartipants.count == 1)

jullianm (approved):

samwyndham (approved):

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
KaterinaWire: @jullianm   I will try to answer without going to much into MLS details 😅
We are talking about 1:1 call. Here's an example:
`if Alice and Bob are calling in a 1:1 conversation over SFT, and Alice hangs up, Bob will receive callParticipantsChanged event from AVS.` (The more info about SFT you can find in the Confluence documentation.)
Now let's look at the differences between the protocols.
**Proteus:**
When Alice hangs up, SFT knows that the new amount of participants is 1 and sends this info to Bob.

**MLS:**
We use subgroups for calls. This means that Alice and Bob join a subgroup. When Alice hangs up, SFT still thinks the number of participants is 2 and sends this information to Bob. And Bob is responsible for leaving the subgroup and ending the call.

jullianm: got it, thanks! I will take a look at SFT doc to learn more about it 👍

PR #1883, opened 2024-09-05
refactor: process team delete event - WPB-10184

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10184" title="WPB-10184" target="_blank">TaskWPB-10184</a>  [iOS] Process TeamEvent.delete event
  </td></table>
  <br />
 

### Key points

Specifically, this PR is about porting the existing implementation of the TeamDelete event.

Legacy code has the following implementation:

```
    private func deleteAccount() {
        let notification = AccountDeletedNotification(context: managedObjectContext)
        notification.post(in: managedObjectContext.notificationContext)
    }
```

### Testing

UTs cover the following use cases:
- Ensuring an account deletion request is triggered

### Checklist

---


caldrian (approved):

WireDomain/Sources/WireDomain/Event Processing/TeamEventProcessor/TeamDeleteEventProcessor.swift
```diff
@@ -30,9 +32,11 @@ protocol TeamDeleteEventProcessorProtocol {
 
 struct TeamDeleteEventProcessor: TeamDeleteEventProcessorProtocol {
 
+    let context: NSManagedObjectContext
+
     func processEvent() async throws {
-        // TODO: [WPB-10184]
-        assertionFailure("not implemented yet")
+        let notification = AccountDeletedNotification(context: context)
```
netbe, line 40: maybe we add a TODO to revisit the implementation: Although the legacy implementation does this, I wonder if our new code should continue to do this.

We basically forward the event to SessionManager to delete the account:

```
  func delete(account: Account, reason: ZMAccountDeletedReason) {
        if let secondAccount = accountManager.accounts.first(where: { $0.userIdentifier != account.userIdentifier }) {
            select(secondAccount, tearDownCompletion: { [weak self] in
                self?.tearDownSessionAndDelete(account: account)
            })
        } else if accountManager.selectedAccount != account {
            // Deleted an inactive account, there's no need notify the UI
            self.tearDownSessionAndDelete(account: account)
        } else {
            // Deleted the last account so we need to return to the logged out area
            logoutCurrentSession(deleteCookie: true, deleteAccount: true, error: NSError(userSessionErrorCode: .accountDeleted, userInfo: [ZMAccountDeletedReasonKey: reason]))
        }
    }
    ```
I would have not expect this processor to do just the forwarding

netbe (approved):
left a comment, the rest looks good

WireDomain/Sources/WireDomain/Event Processing/TeamEventProcessor/TeamDeleteEventProcessor.swift
jullianm: No, this is not ideal and should be revisited to avoid sending a notification, this is a just a thought but we could consider injecting a SessionManager to the SyncManager (though I don't like the idea of the SyncManager having too much dependencies either..)  and use a callback mechanism to trigger these operations. Maybe add a protocol to SessionManager to only expose to the SyncManager what it needs.

jullianm: I will add a TODO for that

jullianm: done, resolving and merging

PR #1888, opened 2024-09-06
fix: App still sends analytics data after analytics is disabled - WPB-10714

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10714" title="WPB-10714" target="_blank">BugWPB-10714</a>  [iOS] App still sends analytics data after analytics is disabled
  </td></table>
  <br />
 

This PR addresses an issue where events were still being sent to Countly even when the user had disabled analytics. Previously, the analytics manager was initialized regardless of whether the user had analytics enabled or disabled. This behavior led to events being tracked even when analytics were supposed to be turned off.

With this fix:

- `EnableAnalyticsUseCase`: The analytics manager is now only initialized when the user explicitly enables analytics.
- `DisableAnalyticsUseCase`: The analytics manager is set to nil when the user disables analytics, ensuring no events are tracked when analytics are turned off.

This approach guarantees that no analytics manager is available when the user opts out, effectively preventing any further event tracking.

---

### Checklist

---


wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -641,23 +641,28 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+    private func selectInitialAccount(
+        _ account: Account,
+        launchOptions: LaunchOptions,
+        completion: @escaping () -> Void
+    ) {
         if let url = launchOptions[UIApplication.LaunchOptionsKey.url] as? URL {
             if (try? URLAction(url: url))?.causesLogout == true {
                 // Do not log in if the launch URL action causes a logout
-                return
+                return completion()
```
netbe: do we need to enable tracking yet if the user logged out? Is it useful for the caller to know if it was successful or not. In that case we could pass a bool in the completion

agisilaos: > do we need to enable tracking yet if the user logged out?

@netbe No, which brings the question do we need the completion there after all. cc: @caldrian

netbe: @agisilaos well I guess you wanted to execute the tracking after the session is loaded

agisilaos: That was the intention yes, because fore that it wouldn't be possible.

agisilaos: @netbe Done https://github.com/wireapp/wire-ios/pull/1888/commits/80a8fed39470c67e77190d5dceea7d573e8a96db

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -641,23 +643,31 @@ public final class SessionManager: NSObject, SessionManagerType {
...
 
         guard !shouldPerformPostRebootLogout() else {
             performPostRebootLogout()
-            return
+            return completion(false)
         }
 
         loadSession(for: account) { [weak self] session in
-            guard let self, let session else { return }
+            guard let self = self, let session = session else {
```
caldrian: Please run the formatter, it will fix it automatically.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
caldrian: I'm not sure we handle login/logout/multiple accounts gracefully.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
agisilaos: Done!

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
agisilaos: I already done that on a previous ticket if I remember correctly. Making sure the analytics user profile is updated and we start a new session if there's a different user e.g has multiple accounts.

wire-ios-sync-engine/Source/SessionManager/SessionManager+AnalyticsUseCases.swift
```diff
@@ -16,26 +16,39 @@
...
 
     public func makeDisableAnalyticsUseCase() throws -> DisableAnalyticsUseCaseProtocol {
-        guard let analyticsManager else {
-            throw AnalyticsSessionError.analyticsNotAvailable
+
+        guard let userSession = self.activeUserSession else {
+            throw AnalyticsSessionError.missingActiveUserSession
+        }
+
+        let didDisableAnalyticsClosure: () -> Void = {
```
KaterinaWire: Where do we use `didDisableAnalyticsClosure`? I didn't find in the code

wire-ios-sync-engine/Source/SessionManager/SessionManager+AnalyticsUseCases.swift
```diff
@@ -16,26 +16,39 @@
...
+            userSession.analyticsSession = nil
         }
-        return DisableAnalyticsUseCase(analyticsManager: analyticsManager)
+
+        return DisableAnalyticsUseCase(
```
KaterinaWire, line 38: We are passing 2 huge objects to `DisableAnalyticsUseCase`: `sessionManager` and `userSession`? Is it so necessary? Especially that `sessionManager` already has a reference to the `activeUserSession`.

KaterinaWire: I can guess that you wanted to pass `didDisableAnalyticsClosure` instead of `sessionManager`, no?

KaterinaWire: Same question about `EnableAnalyticsUseCase`.  @netbe  and @caldrian  I would like to ask you opinions as well

wire-ios-sync-engine/Source/SessionManager/SessionManager+AnalyticsUseCases.swift
caldrian: It was left-over, deleted now.

wire-ios-sync-engine/Source/SessionManager/SessionManager+AnalyticsUseCases.swift
caldrian: > We are passing 2 huge objects to DisableAnalyticsUseCase: sessionManager and userSession?

I don't get the point about "huge objects".
The use cases depend on protocols, one which declares an `analyticsManager` property, where the `AnalyticsManager` reference is set or cleared, and another protocol where the `analyticsSession` can be stored.
It doesn't depend on anything from `ZMUserSession` or `SessionManager`.

If/why the two instances need to be stored separately as well as any implementation detail around `WireAnalytics` I haven't looked into yet.

> Especially that sessionManager already has a reference to the activeUserSession.

Is there always one user session at a time which is available through `SessionManager. activeUserSession`? Then it might make sense to combine the protocols.

netbe: > I can guess that you wanted to pass didDisableAnalyticsClosure instead of sessionManager, no?
yes that was a way I suggested to not mock userSession in SyncEngine

wire-ios-sync-engine/Source/Use cases/DisableAnalyticsUseCase.swift
```diff
@@ -29,21 +29,35 @@ public protocol DisableAnalyticsUseCaseProtocol {
...
-    public func invoke() {
-        analyticsManager.disableTracking()
+    func invoke() {
+        sessionManager.analyticsManager?.disableTracking()
+        sessionManager.analyticsManager = nil
+        userSession.analyticsSession = nil
     }
 }
+
+protocol DisableAnalyticsUseCaseUserSession: AnyObject {
```
samwyndham: `DisableAnalyticsUseCaseUserSession` & `EnableAnalyticsUseCaseUserSession` are identical right? It is also not clear from their names what they do. I think you should combine them into a single protocol. If they are really needed I think you should combine them into a single protocol and rename it to `AnalyticsSessionProviding`

wire-ios-sync-engine/Source/Use cases/DisableAnalyticsUseCase.swift
```diff
@@ -29,21 +29,35 @@ public protocol DisableAnalyticsUseCaseProtocol {
 
 /// Concrete implementation of the DisableAnalyticsUseCaseProtocol.
 /// This struct is responsible for disabling analytics sharing.
-public struct DisableAnalyticsUseCase: DisableAnalyticsUseCaseProtocol {
+struct DisableAnalyticsUseCase<UserSession>: DisableAnalyticsUseCaseProtocol
```
samwyndham: I don't think you need generics here. In the initializer you can just do:

`userSession: DisableAnalyticsUseCaseUserSession`

wire-ios-sync-engine/Source/Use cases/DisableAnalyticsUseCase.swift
```diff
@@ -29,21 +29,35 @@ public protocol DisableAnalyticsUseCaseProtocol {
...
     ///
-    /// - Parameter analyticsManager: The analytics manager to use for disabling tracking.
-    public init(analyticsManager: AnalyticsManagerProtocol) {
-        self.analyticsManager = analyticsManager
+    /// - Parameters:
+    ///   - sessionManager: The session manager that conforms to `AnalyticsManagerProviding` for managing analytics sessions.
+    ///   - didDisableAnalytics: A closure to be executed after analytics have been disabled.
+    init(
+        sessionManager: AnalyticsManagerProviding,
+        userSession: UserSession
```
samwyndham: Consider renaming `userSession` here and else where to something like `analyticsSessionProvider` as we don't really care about the `userSession` functionality - only that it provides the analytics session. It could be any object.

wire-ios-sync-engine/Source/Use cases/EnableAnalyticsUseCase.swift
```diff
@@ -33,15 +33,18 @@ public protocol EnableAnalyticsUseCaseProtocol {
 
 /// Concrete implementation of the EnableAnalyticsUseCaseProtocol.
 /// This struct is responsible for enabling analytics sharing for a specific user profile.
```
samwyndham, line 36: Same comments in this file as I left in `EnableAnalyticsUseCase`

wire-ios-sync-engine/Tests/Source/Use cases/EnableAnalyticsUseCaseTests.swift
```diff
@@ -0,0 +1,70 @@
...
+        sut = nil
+        mockAnalyticsManagerProvider = nil
+        mockAnalyticsManager = nil
+        sessionConfiguration = nil
+        userProfile = nil
+    }
+
+    // MARK: - Unit Tests
+
+    func testTODO() {
```
samwyndham: Will you complete this in this PR?

wire-ios-sync-engine/Source/SessionManager/SessionManager+AnalyticsUseCases.swift
caldrian: The closure is gone.

wire-ios-sync-engine/Tests/Source/Use cases/EnableAnalyticsUseCaseTests.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/pull/1888/commits/e0ee1127c879ce3173717e82559483f4b54a33fe

wire-ios-sync-engine/Source/Use cases/DisableAnalyticsUseCase.swift
samwyndham: I guess you are using Generics to have static vs dynamic dispatch like @caldrian in some of his PRs. Is that correct?

samwyndham (commented):
@agisilaos There are some comments I would like to understand before I continue review. Once those are answered I will complete the review. Thank you!

wire-ios-sync-engine/Source/Use cases/DisableAnalyticsUseCase.swift
agisilaos: Removed the generics from the use cases: https://github.com/wireapp/wire-ios/commit/c82f5d3b1e3b7847902aa182978145fcd4e241d5

wire-ios-sync-engine/Source/Use cases/EnableAnalyticsUseCase.swift
agisilaos: Removed the generics from the use cases: https://github.com/wireapp/wire-ios/commit/c82f5d3b1e3b7847902aa182978145fcd4e241d5

wire-ios-sync-engine/Source/Use cases/DisableAnalyticsUseCase.swift
agisilaos: Done: https://github.com/wireapp/wire-ios/commit/97b486e3536ba60882aa7ac9a80cea7c6f76e7de

wire-ios-sync-engine/Source/Use cases/DisableAnalyticsUseCase.swift
agisilaos: I thought about it and I would like to keep those two separate even though they're containing the same property so we have better separation of concerns if we decide to add anything else inside those protocols.

samwyndham: 👍 Yeah I guess there are always tradeoffs - I would probably take the other approach but it is all subjective! 

Maybe still consider renaming as they don't explicitly need a user session. For example, `DisableAnalyticsUseCaseAnalyticsSession`

samwyndham (approved):
Nice work. Approving with the knowledge that you are still working on the UI toggle issue.

wire-ios-sync-engine/Source/Use cases/DisableAnalyticsUseCase.swift
caldrian: optional
```suggestion
protocol DisableAnalyticsUseCaseAnalyticsSessionProviding: AnyObject {
```

wire-ios-sync-engine/Source/Use cases/EnableAnalyticsUseCase.swift
```diff
@@ -66,8 +78,19 @@ public struct EnableAnalyticsUseCase: EnableAnalyticsUseCaseProtocol {
...
+
+        sessionManager.analyticsManager = analyticsManager
+
         let analyticsSession = analyticsManager.enableTracking(analyticsUserProfile)
-        userSession.analyticsSession = analyticsSession
+        analyticsSessionProvider.analyticsSession = analyticsSession
     }
 }
+
+protocol EnableAnalyticsUseCaseUserSession: AnyObject {
```
caldrian: optional
```suggestion
protocol EnableAnalyticsUseCaseAnalyticsSessionProviding: AnyObject {
```

wire-ios-sync-engine/Source/Use cases/EnableAnalyticsUseCase.swift
```diff
@@ -33,31 +34,42 @@ public protocol EnableAnalyticsUseCaseProtocol {
...
     /// - Parameters:
-    ///   - analyticsManager: The analytics manager to use for enabling tracking.
+    ///   - analyticsManagerBuilder: A closure that provides an `AnalyticsManagerProtocol` implementation.
+    ///   - sessionManager: The session manager that conforms to `AnalyticsManagerProviding` for managing analytics sessions.
+    ///   - analyticsSessionConfiguration: The configuration for the analytics session.
     ///   - analyticsUserProfile: The user profile for which to enable analytics sharing.
-    public init(
-        analyticsManager: AnalyticsManagerProtocol,
+    ///   - analyticsSessionProvider: An instance conforming to `EnableAnalyticsUseCaseUserSession` that manages the user's analytics session state.
+    init(
```
caldrian, line 66: Since the `init` doesn't have to be `public`, it can be omitted:
Just declare the properties like this:
```
/// The configuration for the analytics session.
private(set) var analyticsSessionConfiguration: AnalyticsSessionConfiguration

/// A closure that provides an `AnalyticsManagerProtocol` implementation.
private(set) var analyticsManagerBuilder: (_ appKey: String, _ host: URL) -> any AnalyticsManagerProtocol

/// The session manager that conforms to `AnalyticsManagerProviding` for managing analytics sessions.
private(set) var sessionManager: AnalyticsManagerProviding

/// The user profile for which to enable analytics sharing.
private(set) var analyticsUserProfile: AnalyticsUserProfile

/// An instance conforming to `EnableAnalyticsUseCaseUserSession` that manages the user's analytics session state.
private(set) var analyticsSessionProvider: EnableAnalyticsUseCaseUserSession
```

caldrian (approved):

wire-ios-sync-engine/Source/Use cases/EnableAnalyticsUseCase.swift
agisilaos: Done!

wire-ios-sync-engine/Source/Use cases/EnableAnalyticsUseCase.swift
agisilaos: Done!

wire-ios-sync-engine/Source/Use cases/DisableAnalyticsUseCase.swift
```diff
@@ -29,21 +29,34 @@ public protocol DisableAnalyticsUseCaseProtocol {
...
+    ) {
+        self.sessionManager = sessionManager
+        self.analyticsSessionProvider = analyticsSessionProvider
     }
 
     /// Invokes the use case to disable analytics sharing.
     ///
     /// This method calls the `disableTracking` method on the analytics manager if it exists.
-    public func invoke() {
-        analyticsManager.disableTracking()
```
johnxnguyen, line 47: Why wasn't this enough before? It seems like a bug that even if this is called, that the events are still being sent, regardless of the `AnalyticsManager` existing.

netbe: @agisilaos ?

agisilaos: @johnxnguyen Not really sure to be honest. As soon as we took the analytics manager out of the equation we stopped receiving events when disabled. After we merge everything to the feature branch. I want to take another look at the whole refactor and set of changes as a good practice to see what we can improve. 

For now is it ok to go ahead and merge it?

johnxnguyen: Your solution introduces some complexity that would be nice to avoid. Since this is somewhat urgent, I'm happy to approve it but I think the time should be taken to get to the bottom of the issue.

johnxnguyen: But then... if we know that `disableAnayltics()` doesn't work, why do we still call it?

agisilaos: Sounds good! 👍

agisilaos: We're going to revisit with @johnxnguyen after everything is merged.

PR #1905, opened 2024-09-09
feat: Remove custom self delete timer option - WPB-10861

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10861" title="WPB-10861" target="_blank">TaskWPB-10861</a>  Remove Custom option in self delete timer
  </td></table>
  <br />
 

Also fix memory link when assigning `didTapViewClosure`

### Issue

On **development builds only** we have the option to select a custom time for the self delete timer both when sending a message or in the self deleting messages group options. 3rd party QA found a collection of bugs related to this.

As this feature is likely not used much this PR removes the feature to avoid future confusion with QA.

Additionally this PR fixes a memory leak due to explicitly capturing `self` when configuring `picker.didTapViewClosure`:

Screenshot 2024-09-09 at 15 15 39

### Testing

1. Go to a group which you have admin permissions.
2. Select the `timer`.
3. Scroll through the picker.
4. Ensure there is no `Custom` options.
5. Select 10 seconds.
6. Send a message.
7. Ensure the message is deleted after 10 seconds.
8. Go to group details > `Self-deleting messages`.
9. Ensure there is no `Custom` option.

---

### Checklist


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/EphemeralKeyboardViewController/EphemeralKeyboardViewController.swift
```diff
@@ -95,7 +93,9 @@ final class EphemeralKeyboardViewController: UIViewController {
         picker.backgroundColor = .clear
         picker.tintColor = .red
         picker.selectorColor = separatorColor
-        picker.didTapViewClosure = dismissKeyboardIfNeeded
+        picker.didTapViewClosure = { [weak self] in
```
samwyndham, line 96: Fixes a memory leak

netbe: nice work

netbe (approved):

caldrian (approved):
Nice also that you kept backwards compatibility.
For conversations which have already a custom timeout set, the value is still shown until a new value is chosen.

Btw, not related to these changes:
If a self-deletion timeout is set in the conversation settings, we should hide the option in the input accessory view. It seems like the conversation settings will always override the one set per message, no matter if it's longer or shorter.

agisilaos (approved):
Great! Thank you @samwyndham!

PR #1912, opened 2024-09-10
chore: remove unnecessary SnapshotTesting import - no ticket

jullianm:
no-ticket

### Key points

- Removes deprecated `verify` method from `XCTestCase+SnapshotTesting.swift` extension file, use `SnapshotHelper` from WireTestingPackage (part of WireFoundation local package) instead.
- Removes unnecessary `import SnapshotTesting` from codebase.

### Checklist

---


wire-ios/Wire-iOS Tests/Calling/CallActionsViewSnapshotTests.swift
```diff
@@ -67,6 +68,7 @@ final class CallActionsViewSnapshotTests: XCTestCase {
 
     private var sut: CallActionsView!
     private var widthConstraint: NSLayoutConstraint!
+    private let snapshotHelper = SnapshotHelper()
```
samwyndham:
```suggestion
    private var snapshotHelper: SnapshotHelper!
```
I think it is good practice to create this in setup and delete in tear down to help high memory usage due to how XCTest works - otherwise, as far as I know, a new `SnapshotHelper` instance is created for every test in this test case and kept in memory until the whole test bundle completes testing.

To be honest it probably makes little difference in the case of `SnapshotHelper` but I think it is good for us to be consistent with all kinds of objects we use.

samwyndham (approved):
Nice work!

caldrian (approved):

wire-ios/Wire-iOS Tests/Calling/CallActionsViewSnapshotTests.swift
jullianm: fixed, thx. resolving and merging

PR #1900, opened 2024-09-09
refactor: process federation connection removed event - WPB-10187

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10187" title="WPB-10187" target="_blank">TaskWPB-10187</a>  [iOS] Process FederationConnectionRemovedEvent
  </td></table>
  <br />
 

### Key points

This PR is part of the quick sync refactoring plan and is related to processing the multiple events we receive from the backend or the push channel.

Specifically, this PR is about porting the existing implementation of the `FederationConnectionRemovedEventProcessor` event.

Legacy code was using a FederationTerminationManager part of WireRequestStrategy framework to manage the storage logic. 

```
 federationTerminationManager.handleFederationTerminationBetween(firstDomain, otherDomain: secondDomain)
```

To avoid using this internal dependency, we now directly perform the storage logic operations in our dedicated ConnectionRepository object.

### Testing

UTs cover the following use cases:
- Federated participants are removed from the conversations that are not hosted on the given domains.
- A federated participant on a hosted conversation domain is removed from that conversation if he's not part of the same domain.

### Checklist

- [ ] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [ ] Description is filled and free of optional paragraphs.
- [ ] Adds/updates automated tests.

---

### UI accessibility checklist

_If your PR includes UI changes, please utilize this checklist:_
- [ ] Make sure you use the API for UI elements that support large fonts.
- [ ] All colors are taken from WireDesign.ColorTheme or constructed using WireDesign.BaseColorPalette.
- [ ] New UI elements have Accessibility strings for VoiceOver.

PR #1918, opened 2024-09-11
refactor: process federation delete event - WPB-10188

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10188" title="WPB-10188" target="_blank">TaskWPB-10188</a>  [iOS] Process FederationDeleteEvent
  </td></table>
  <br />
 

### Key Points

Specifically, this PR is about porting the existing implementation of the `FederationDeleteEventProcessor` event.

Legacy code was using a FederationTerminationManager part of WireRequestStrategy framework to manage the storage logic.

```
federationTerminationManager.handleFederationTerminationWith(payload.domain)

```

To avoid using this internal dependency, we now directly perform the storage logic operations in our dedicated ConnectionRepository object.

Note: The following ongoing PR [here](https://github.com/wireapp/wire-ios/pull/1900) was merged into this one as they are closely related.

### Testing

- a 1:1 conversation is marked as read-only.
- `pending` connections requests are `ignored`.
- `sent` connection requests are `cancelled`.
- connections for connected users are removed.
- selfUser is removed from conversation hosted by defederated domain.
- it removes a user from a conversation hosted by self user domain.
- it removes selfUser and another user from a conversation hosted by another domain.

### Checklist

---


netbe (approved):
should we close https://github.com/wireapp/wire-ios/pull/1900 then? @jullianm 
the rest looks ok apart from swiftlint warnings

jullianm:
@netbe yep I closed the other PR since changes are already part of this PR

WireDomain/Sources/WireDomain/Event Processing/FederationEventProcessor/FederationConnectionRemovedEventProcessor.swift
```diff
@@ -32,9 +32,20 @@ protocol FederationConnectionRemovedEventProcessorProtocol {
...
+    }
+
+    let repository: any ConnectionsRepositoryProtocol
+
+    func processEvent(_ event: FederationConnectionRemovedEvent) async throws {
+        guard let domain = Array(event.domains).first,
+              let otherDomain = Array(event.domains).last
+        else {
+            throw Error.missingDomains(event.domains)
+        }
```
johnxnguyen: Since `event.domains` is a `Set`, the order is not defined when it is converted to an `Array`, which means it's possible that `domain` and `otherDomain` are the same. I think we should also validate that there are 2 domains (if that is indeed the case).

```suggestion
        let domains = Array(event.domains)
        guard 
            domains.count == 2,
            let domain = domains.first,
            let otherDomain = domains.last
        else {
            throw Error.missingDomains(event.domains)
        }
```

WireDomain/Sources/WireDomain/Repositories/Connections/ConnectionsLocalStore.swift
```diff
@@ -0,0 +1,389 @@
...
+    ) async throws
+
+    func deleteFederationConnection(
+        with domain: String
+    ) async throws
+
+    func removeFederationConnection(
+        between domain: String,
+        and otherDomain: String
+    ) async
```
johnxnguyen: These connections are not quite the same. `Connection` is the connection between two users (via a connection request). The connection between federating domains is the ability to communicate between those domains.

What do you think about moving the federation logic up into the event processor?

WireDomain/Sources/WireDomain/Event Processing/FederationEventProcessor/FederationConnectionRemovedEventProcessor.swift
jullianm: fixed in https://github.com/wireapp/wire-ios/pull/1918/commits/c77f7a4babb947cd5f43ea7e091187dbddd08c68

WireDomain/Sources/WireDomain/Repositories/Connections/ConnectionsLocalStore.swift
jullianm: It makes sense, fixed in https://github.com/wireapp/wire-ios/pull/1918/commits/c77f7a4babb947cd5f43ea7e091187dbddd08c68

johnxnguyen (approved):

PR #1922, opened 2024-09-12
chore: use case for getting account image - WPB-11052

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11052" title="WPB-11052" target="_blank">StoryWPB-11052</a>  Extract getting the image of the user account use case
  </td></table>
  <br />
 

### Issue

Since for the sidebar on iPad we need to get and show an account image, this PR extracts the logic into a use case.

A new target `WireAccountImage` is added which doesn't have any dependencies beside `WireFoundation`.
It contains the view code as well as use case code, so basically everything that belongs to that feature.

For that a few protocols abstract the model structure of `WireDataModel`.

---

### Checklist

---


.github/workflows/test_pr_changes.yml
```diff
@@ -20,6 +20,7 @@ jobs:
     # Set job outputs to values from filter step
     outputs:
       wire-ios: ${{ steps.filter.outputs.wire-ios == 'true' || steps.filter.outputs.carthage == 'true' }}
+      wire-user-profile: ${{ steps.filter.outputs.wire-user-profile }}
```
netbe: is that a new package?

WireUI/Package.swift
```diff
@@ -10,7 +10,8 @@ let package = Package(
     products: [
         .library(name: "WireDesign", targets: ["WireDesign"]),
         .library(name: "WireReusableUIComponents", targets: ["WireReusableUIComponents"]),
-        .library(name: "WireUIFoundation", targets: ["WireUIFoundation"])
+        .library(name: "WireUIFoundation", targets: ["WireUIFoundation"]),
+        .library(name: "WireUserProfile", targets: ["WireUserProfile"])
```
netbe: Why is this a new target? Is it only UI?

.github/workflows/test_pr_changes.yml
caldrian: A new target in WireUI.

caldrian: Do you prefer to combine all targets of a package and run the unit tests together?

WireUI/Package.swift
caldrian: I combined all WireUI targets into a single WireUI test plan which is run by the CI.

.swiftlint.yml
```diff
@@ -83,20 +83,21 @@ excluded:
   - "**/Package.swift"
   - "**/Package@swift-*.swift"
   - Carthage
-  - "WireUI/Sources/WireDesign/Icons/Autogenerated/StyleKitIcons.generated.swift"
-  - "WireUI/Sources/WireDesign/Icons/Autogenerated/WireStyleKit.swift"
+  - DerivedData
```
caldrian, line 86: For some reason specifying `**/DerivedData` doesn't prevent SwiftLint to exclude any `DerivedData` folder at the root.

WireUI/Sources/WireAccountImage/UseCases/GetTeamAccountImageUseCaseProtocol.swift
```diff
@@ -0,0 +1,27 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
+
+/// If the team has an image set, this use cases retrieves it.
+/// If no team image data is available, an image will be generated using the team name's initials.
+public protocol GetTeamAccountImageUseCaseProtocol {
+
+    func invoke<User, Account>(user: User, account: Account) async throws -> UIImage
```
samwyndham: How come you use generics here instead of 

```swift
 func invoke(
    user: GetAccountImageUseCaseUserProtocol, 
    account: GetAccountImageUseCaseAccountProtocol
) async throws -> UIImage
```

johnxnguyen (approved):

samwyndham (approved):
Nice work extracting this functionality. I left a comment asking about the reason for using generics in the protocols that I would like to understand.

Also, I would say that this PR is pretty big and I do think my reviews are not to good on big PRs, but I understand it is not always easy to keep PR sizes small.

WireUI/Sources/WireAccountImage/UseCases/GetTeamAccountImageUseCaseProtocol.swift
caldrian: Because I prefer static dispatching.
Actually I forgot to change it to the compact syntax using `some` in the two protocols.
Thanks for pointing that out.
Fixed: 3ef1d0cfc236667efc07a833cf416ccda1a71f81

samwyndham: OK I see. Yeah I think this is worthwhile and the direction Swift is pushing us. Perhaps this is worth some discussion (maybe it has been discussed already) so have this as a principle in our project

PR #1927, opened 2024-09-12
refactor: calculate and push supported protocols use case - WPB-10734

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10734" title="WPB-10734" target="_blank">TaskWPB-10734</a>  [iOS] Port calculating and push self supported protocols to WireDomain
  </td></table>
  <br />
 

### Key points

This PR, part of the slow sync refactoring plan, aims at porting the calculation of the supported protocols to be pushed remotely from the WireSyncEngine `SupportedProtocolsService` to our Domain package.

When a slow sync is triggered, the newly created `PushSupportedProtocolsUseCase` will calculate and push the supported protocols to the backend using `UserRepository`.

### Testing

- Supported protocols are correctly calculated.

### Checklist

---


WireDomain/Sources/WireDomain/Synchronization/PushSupportedProtocolsUseCase.swift
```diff
@@ -0,0 +1,169 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireAPI
+import WireDataModel
+import WireSystem
+
+/// A domain strictly related use case.
```
samwyndham: What do you mean by this?

WireDomain/Sources/WireDomain/Synchronization/PushSupportedProtocolsUseCase.swift
```diff
@@ -0,0 +1,169 @@
...
+/// Helps calculating the supported protocols.
+struct PushSupportedProtocolsUseCase {
```
samwyndham: I'm still not sure what a use case is 🙈. I feel like they are mainly used by android developers so probably I should ask some. I had a Quick Look at [UseCase Red Flags and Best Practices in Clean Architecture](https://engineering.teknasyon.com/usecase-red-flags-and-best-practices-in-clean-architecture-76e2f6d921eb) which I don't fully agree with but perhaps point `2` about naming in it makes sense. So in that sense I would rename this struct to `CalculatePushSupportedProtocolsUseCase` and rename `func calculateSupportedProtocols()` to `func invoke()`.

WireDomain/Sources/WireDomain/Synchronization/PushSupportedProtocolsUseCase.swift
```diff
@@ -0,0 +1,169 @@
...
+
+        var result = Set<WireAPI.MessageProtocol>()
+
+        /// All clients are proteus ready so we support it if the backend does.
+        if remoteProtocols.contains(.proteus) {
+            result.insert(.proteus)
+        }
+
+        /// All clients are mls ready so we support it if the backend does.
+        if remoteProtocols.contains(.mls), allClientsMLSReady {
```
samwyndham: I guess the  comment above is incorrect? .. as we only support it if `allClientsMLSReady` returns true.

WireDomain/Sources/WireDomain/Synchronization/PushSupportedProtocolsUseCase.swift
```diff
@@ -0,0 +1,169 @@
...
+        guard
+            let startDate = mlsMigration.config.startTime,
+            startDate <= now
+        else {
+            return .notStarted
+        }
+
+        guard
+            let endDate = mlsMigration.config.finaliseRegardlessAfter,
+            endDate <= now
```
samwyndham: Sanity check: Is this logic correct? I imagined that if now is after endDate we should already be finalized.

WireDomain/Tests/WireDomainTests/Synchronization/PushSupportedProtocolsUseCaseTests.swift
```diff
@@ -0,0 +1,357 @@
...
+
+        await mock(migrationState: .notStarted)
+        let supportedProtocols_NotSupported = await sut.calculateSupportedProtocols()
+        XCTAssertEqual(supportedProtocols_NotSupported, [.proteus, .mls])
+
+        await mock(migrationState: .ongoing)
+        let supportedProtocols_OnGoing = await sut.calculateSupportedProtocols()
+        XCTAssertEqual(supportedProtocols_OnGoing, [.proteus, .mls])
+
+        await mock(migrationState: .finalised)
```
samwyndham: Aer you missing an assertion?

WireDomain/Tests/WireDomainTests/Synchronization/PushSupportedProtocolsUseCaseTests.swift
```diff
@@ -0,0 +1,357 @@
...
+        try coreDataStackHelper.cleanupDirectory()
+        coreDataStackHelper = nil
+
+        try await super.tearDown()
+    }
+
+    // MARK: - Tests
+
+    func test_CalculateSupportedProtocols_AllActiveMLSClients_RemoteProteus() async throws {
+        // Given
```
samwyndham, line 76: OPTIONAL: Consider using parameterized testing here where you define the test cases:

```swift
let testCases: [(migrationState:Scaffolding.MigrationState, supportedProtocols: Set<WireAPI.MessageProtocol>)] = [
   (migrationState: .disabled, supportedProtocols: [.proteus]),
   (migrationState: .notStarted, supportedProtocols: [.proteus]),
   (migrationState: .ongoing, supportedProtocols: [.proteus]),
   (migrationState: .finalised, supportedProtocols: [.proteus])
]
```

... then loop over them an assert the expected result. It _might be_ easier to read

samwyndham (approved):
Looks good. The hard part was that I'm not familiar with the business requirements so this is more just commenting on the code not the correctness of it.

WireDomain/Sources/WireDomain/Synchronization/PushSupportedProtocolsUseCase.swift
jullianm: I meant that this UseCase was not supposed to be exposed (and used) outside of this package but actually since UseCases (which handle business logic) should be part of the Domain layer, I removed that comment.

WireDomain/Sources/WireDomain/Synchronization/PushSupportedProtocolsUseCase.swift
jullianm: Thanks, I revisited the implementation using invoke() and also pushing the calculated supported protocols using the dedicated UserRepository.

WireDomain/Sources/WireDomain/Synchronization/PushSupportedProtocolsUseCase.swift
jullianm: a little bit confusing indeed, I fixed it thx

WireDomain/Sources/WireDomain/Synchronization/PushSupportedProtocolsUseCase.swift
jullianm: it looks like `finaliseRegardlessAfter` is the deadline, if we have not reached this date yet, the migration is still ongoing, so theoretically, the above logic is incorrect (inversed actually) however this is the legacy code so I assume there is some reason we are doing so

WireDomain/Tests/WireDomainTests/Synchronization/PushSupportedProtocolsUseCaseTests.swift
jullianm: fixed thx

WireDomain/Tests/WireDomainTests/Synchronization/PushSupportedProtocolsUseCaseTests.swift
jullianm: done thx

WireDomain/Sources/WireDomain/Synchronization/PushSupportedProtocolsUseCase.swift
jullianm: will double check on Android, out of curiosity

jullianm: on Android, logic is the same

```
 fun MLSMigrationModel.hasMigrationEnded(): Boolean {
     return status == Status.ENABLED && endTime?.let { Clock.System.now() > it } ?: false
}
```

WireDomain/Tests/WireDomainTests/Synchronization/PushSupportedProtocolsUseCaseTests.swift
```diff
@@ -0,0 +1,362 @@
...
+
+        let testCases: [(migrationState: Scaffolding.MigrationState, supportedProtocols: Set<WireAPI.MessageProtocol>)] = [
+            (migrationState: .disabled, supportedProtocols: [.mls]),
+            (migrationState: .notStarted, supportedProtocols: [.proteus]),
+            (migrationState: .ongoing, supportedProtocols: [.proteus]),
+            (migrationState: .finalised, supportedProtocols: [.mls])
+        ]
+
+        for testCase in testCases {
+            await mock(migrationState: testCase.migrationState)
```
netbe: nit-pick: are we really mocking or is it more setup?
```suggestion
            await setup(migrationState: testCase.migrationState)
```
same for all mock methods

netbe (approved):
minor comment on naming, the rest looks good

samwyndham (approved):

WireDomain/Tests/WireDomainTests/Synchronization/PushSupportedProtocolsUseCaseTests.swift
jullianm: fixed thx

PR #1928, opened 2024-09-13
feat: Determine our `SwiftFormat` rules - WPB-10848

samwyndham:
# ⚠️ ⚠️ ⚠️ This PR will never be merged to `develop`. You don't need to review it closely. Please read the description carefully.

### Issue

As discussed in [Linting & Formatting Improvements](https://wearezeta.atlassian.net/wiki/spaces/IOS/pages/1391755277/Linting+Formatting+Improvements) (@johnxnguyen please take a look at that document as you were away during the meeting) we plan to improve our linting and formatting of Swift code to cover the entire codebase with more rules.

This PR is a place of discussion of `SwiftFormat` rules only. A similar PR will be created in the future to discuss `SwiftLint` rules. Once we have consensus on both `SwiftFormat` and `SwiftLint` rules we can move forward. 

Nearly every rule is enabled in this PR including the ones the make me 🤮 - there are two exceptions:

- [acronyms](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#acronyms) as this is hard to use with coding keys.
- [docComments](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#doccomments) as this won't work with our _floating_ style comments.

Personally I'm of the opinion that generally the more rules the better that lead to a standard style across the team, but if we have preferences which these rules go against perhaps we can assert them via linting instead.

#### Task

Each commit (approximately) in this PR introduces a new rule to the codebase. The rule name will be in the title of the commit and there may be some additional info about it in the commit description.

As a reviewer your job is to go through the PR **commit by commit** and note down any rules that you dislike or have concerns about. 

⚠️ Please leave you review here: https://wearezeta.atlassian.net/wiki/spaces/IOS/pages/1454800899/SwiftFormat+review

I will leave this PR up for some time. Please enjoy bike shedding.

### Checklist


netbe:
> [docComments](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#doccomments) as this won't work with our floating style comments.
what do you mean by that?

wire-ios-request-strategy/Sources/Helpers/ZMSound.swift
```diff
@@ -76,7 +76,9 @@ public enum ZMSound: String, CustomStringConvertible {
         switch self {
         case .None:
             return nil
-        case .WireCall, .WirePing, .WireText:
+        case .WireCall,
```
netbe, line 79: I'll go with like or not enforces, worst case a line length would enforce going to the next line

wire-ios-cryptobox/WireCryptobox/EncryptionSessionsDirectory.swift
```diff
@@ -31,14 +31,15 @@ enum EncryptionSessionError: Int {
...
-        }
+        var info =
+            switch self {
+            case .unknown:
+                "Unknown EncryptionSessionError"
+            case .encryptionFailed:
+                "Encryption Failed"
+            case .decryptionFailed:
+                "Decryption Failed"
+            }
```
netbe, line 42: don't like this

samwyndham:
> > [docComments](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#doccomments) as this won't work with our floating style comments.
> > what do you mean by that?

@netbe I mean that we often have a line of whitespace between the doc comment and the declaration:

```swift
/// Adds two numbers

func add(a: Int, b: Int) {
  // ...
}
```

Whereas the rule expects the following and behaves incorrectly without it.

```swift
/// Adds two numbers
func add(a: Int, b: Int) {
  // ...
}

**Review by samwyndham — COMMENTED** [review]
Here is my review of the rules. I am generally happy with most rules. I have reservations about the following:

- [conditionalAssignment](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#conditionalAssignment) - I have mixed feelings about this one. I'm not super against it but feel that in some cases the content can become overly dense.
- [preferForLoop](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#preferForLoop) - Mixed feelings. I would prefer if it ignored one liner `forEach` statements which can be done with `--onelineforeach` option.
- [redundantType](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#redundantType) - This can be bad for compile times. If we do enable it we should also measure our compile times and add exceptions where necessary.
- [markTypes](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#markTypes) - Creates a lot of marks and I don't feel they are that useful for anything. If everything becomes a mark what is the point.
- [organizeDeclarations](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#organizeDeclarations) - Not against this, but I think this one requires a closer look and discussion. It moves our code around a lot.
- [wrapConditionalBodies](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#wrapConditionalBodies) - I don't like this one as I like single line early exits.
- [wrapSwitchCases](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#wrapSwitchCases) - I don't like this one. It takes up unnecessary space in my opinion.

File: wire-ios-data-model/Source/MLS/MLSGroup.swift [hunk 3]
```diff
@@ -60,11 +60,9 @@ public class MLSGroup: ZMManagedObject {
 
     static let idKey = "id"
 
-    @NSManaged
-    private var primitiveId: Data?
+    @NSManaged private var primitiveId: Data?
```
netbe: weird it does not behave as @discardableResult or @objc

netbe (commented):
Reviewed all the rules

KaterinaWire (approved):

.swiftformat
```diff
@@ -20,7 +20,8 @@
     enumNamespaces, \
     extensionAccessControl, \
     genericExtensions, \
-    hoistAwait
+    hoistAwait, \
+    hoistPatternLet
```
caldrian: This rule I would probably disable and decide on a case-by-case basis.

WireUI/Sources/WireReusableUIComponents/AccessoryTextField/AccessoryTextField.swift
```diff
@@ -154,12 +154,12 @@ extension AccessoryTextField {
 // MARK: - Custom edge insets
 
 extension AccessoryTextField {
-    public override func textRect(forBounds bounds: CGRect) -> CGRect {
+    override public func textRect(forBounds bounds: CGRect) -> CGRect {
```
caldrian, line 173: I suggest to configure the `modifierOrder` rule with a different order. At least `override` I would want to move to the right.

.swiftformat
```diff
@@ -28,7 +28,8 @@
     leadingDelimiters, \
     linebreakAtEndOfFile, \
     linebreaks, \
-    modifierOrder
+    modifierOrder, \
+    numberFormatting
```
caldrian: I like the rule, however, we should discuss about the configuration.

.swiftformat
```diff
@@ -87,6 +87,7 @@
     # OPTIONAL
     # acronyms, \ - Difficult to use because it messes with coding keys etc. So a lot of disabling this rule would be necessary
     blankLineAfterSwitchCase, \
+    blankLinesBetweenImports, \
```
caldrian, line 90: I would disable this rule. I really like to have `import` and `@testable import` separated and sorted within their blocks.

.swiftformat
```diff
@@ -91,6 +91,7 @@
     blockComments, \
     # docComments - Cannot enable due to our floating doc comment style
     isEmpty, \ # Rare false positives in situations where a type provides a `count` but not a `isEmpty` property.
+    markTypes, \
```
caldrian, line 94: I don't consider this rule beneficial. Xcode would then show the name of the type twice, right? 
Screenshot 2024-10-21 at 7 22 42 PM

.swiftformat
```diff
@@ -96,6 +96,7 @@
     organizeDeclarations, \
     redundantProperty, \
     sortSwitchCases, \
+    wrapConditionalBodies, \
```
caldrian, line 99: I wouldn't enable this rule.

.swiftformat
```diff
@@ -97,6 +97,7 @@
     redundantProperty, \
     sortSwitchCases, \
     wrapConditionalBodies, \
+    wrapEnumCases, \ # In rare cases this messes up enums that fail the rule and has to be manually fixed - affected 4 enums over the entire project
```
caldrian, line 100: I wouldn't enable this rule.

.swiftformat
```diff
@@ -99,6 +99,7 @@
     wrapConditionalBodies, \
     wrapEnumCases, \ # In rare cases this messes up enums that fail the rule and has to be manually fixed - affected 4 enums over the entire project
     wrapMultilineConditionalAssignment, \
+    wrapSwitchCases, \
```
caldrian, line 102: I probably wouldn't enable this rule.

caldrian (approved):
Great work! Looking forward to enabling many of these rules.

.swiftformat
```diff
@@ -14,11 +14,14 @@
...
+    elseOnSameLine
 
 # Options
 
 --lineaftermarks true # blankLinesAroundMark
 --typeblanklines remove # blankLinesAtStartOfScope
 --allman false # braces
 --condassignment after-property # conditionalAssignment
+--elseposition same-line # elseOnSameLine
+--guardelse auto # elseOnSameLine
```
jullianm, line 111: 👍 I'm surprised there's not many fixes in codebase after adding this rule

WireAPI/Tests/WireAPITests/APIs/TeamsAPI/TeamsAPITests.swift
```diff
@@ -55,7 +55,7 @@ final class TeamsAPITests: XCTestCase {
 
     func testGetTeamMembersRequest() async throws {
         try await apiSnapshotHelper.verifyRequestForAllAPIVersions { sut in
-            _ = try await sut.getTeamMembers(for: .mockID1, maxResults: 2_000)
+            _ = try await sut.getTeamMembers(for: .mockID1, maxResults: 2000)
```
jullianm, line 57: there's another rule that does the exact opposite right ? like formatting 2000 into 2_000, I think it makes sense for big numbers, 1_000_000 is more readable but for small numbers 1000 is fine

wire-ios-data-model/Source/MLS/CoreCryptoKeyProvider.swift
```diff
@@ -87,7 +87,7 @@ struct CoreCryptoKeychainItem: KeychainItemProtocol {
         ]
     }
 
-    func setQuery<T>(value: T) -> [CFString: Any] {
+    func setQuery(value: some Any) -> [CFString: Any] {
```
jullianm, line 100: it doesn't make sense, I think it should resolve this to `Any`, not `some Any`

.swiftformat
```diff
@@ -30,7 +30,8 @@
     linebreaks, \
     modifierOrder, \
     numberFormatting, \
-    opaqueGenericParameters
+    opaqueGenericParameters, \
+    preferForLoop
```
jullianm: not particularly fan of this rule, in many cases it makes sense to use ForEach() over a for-in loop.

wire-ios-data-model/Source/Model/Conversation/ConversationLike.swift
```diff
@@ -88,7 +88,7 @@ extension ZMConversation: ConversationLike {
 
     public var sortedServiceUsers: [UserType] {
         localParticipants
-            .filter { $0.isServiceUser }
+            .filter(\.isServiceUser)
```
jullianm, line 100: 👍

.swiftformat
```diff
@@ -50,3 +50,4 @@
 --extensionacl on-declarations # extensionAccessControl
 --patternlet hoist # hoistPatternLet
 --indent 4 # indent
+--anonymousforeach ignore # preferForLoop
```
jullianm, line 117: not particularly fan of enforcing the usage of one or the other, II would not add any of these two rules regarding loops.

.swiftformat
```diff
@@ -67,3 +68,4 @@
 --indent 4 # indent
 --anonymousforeach ignore # preferForLoop
 # --self init-only # redundantSelf TODO
+--redundanttype inferred
```
jullianm: hmmm interesting, I feel like in some cases we might want the type to be explicit

```
var number = 2 // inferred as Int
var number: Double = 2
```

or 

```
var id: UUID!
let copy = id // inferred as UUID? but we know this should have a value
let copy: UUID = id // not having to deal with optional unwrapping anymore
```

.swiftformat
```diff
@@ -69,7 +69,8 @@
     spaceInsideComments, \
     spaceInsideGenerics, \
     spaceInsideParens, \
-    strongOutlets
+    strongOutlets, \
```
jullianm, line 72: 🤔

.swiftformat
```diff
@@ -74,7 +74,8 @@
     todos, \
     trailingClosures, \
     trailingCommas, \
-    trailingSpace
+    trailingSpace, \
+    typeSugar
```
jullianm: ❤️

.swiftformat
```diff
@@ -77,8 +77,9 @@
     trailingSpace, \
     typeSugar, \
     unusedArguments, \
-    void
-
+    void, \
+    wrap, \
+    wrapArguments
```
jullianm: definitely yes on this rule

.swiftformat
```diff
@@ -82,7 +82,8 @@
     wrapArguments, \
     wrapAttributes, \
     wrapLoopBodies, \
-    wrapSingleLineComments
+    wrapSingleLineComments, \
+    yodaConditions
```
jullianm: let's add this rule just for its name.

.swiftformat
```diff
@@ -85,6 +85,7 @@
     wrapSingleLineComments, \
     yodaConditions, \
     # OPTIONAL
+    # acronyms, \ - Difficult to use because it messes with coding keys etc. So a lot of disabling this rule would be necessary
```
jullianm: so maybe let's not add this rule.

.swiftformat
```diff
@@ -93,6 +93,7 @@
     isEmpty, \ # Rare false positives in situations where a type provides a `count` but not a `isEmpty` property.
     markTypes, \
     noExplicitOwnership, \
+    organizeDeclarations, \
```
jullianm, line 96: I find this rule useful, we often see discrepancies in codebase regarding the way we write our // MARK.

.swiftformat
```diff
@@ -98,6 +98,7 @@
     sortSwitchCases, \
     wrapConditionalBodies, \
     wrapEnumCases, \ # In rare cases this messes up enums that fail the rule and has to be manually fixed - affected 4 enums over the entire project
+    wrapMultilineConditionalAssignment, \
```
jullianm, line 101: Seeing the changes below, not a big fan of this rule.

jullianm (approved):
Nice work 👍

johnxnguyen (commented):
Left my review in confluence

WireAPI/Tests/WireAPITests/APIs/TeamsAPI/TeamsAPITests.swift
samwyndham: This is configurable. I'll put you down as in favor but with a different configuration.

wire-ios-data-model/Source/MLS/CoreCryptoKeyProvider.swift
samwyndham: I think that is the new generics API if we don't constrain the generic to a specific type.

.swiftformat
samwyndham: There was a bit of conversation about this. Apparently this is what apple advises in most cases now. However in practice we don't have any weak IBOutlets in our codebase and probably won't get any.

wire-ios-data-model/Source/ManagedObjectContext/Migration/CoreDataMessagingMigrationVersion.swift
```diff
@@ -85,8 +85,8 @@ enum CoreDataMessagingMigrationVersion: String, CoreDataMigrationVersion {
             nil
         case .v116, .v117:
             .v118
-        case .v115,
-             .v114:
+        case .v114,
```
netbe, line 89: here the order is changed

PR #1932, opened 2024-09-13
refactor: pull last event ID - WPB-10733

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10733" title="WPB-10733" target="_blank">TaskWPB-10733</a>  [iOS] Pull last update event id
  </td></table>
  <br />
 

…PB-10733

### Key points

This PR is part of the quick sync refactoring plan and is related to pulling and storing the last event ID in the UserDefaults.

This PR moves the legacy `PrivateUserDefaults` to WireFoundation package and use it from `UpdateEventsRepository` to store the last event envelope ID.

### Testing

The UTs cover the following use cases, ensuring that:
- last event ID is properly stored in user defaults after pulling it from backend

---

### Checklist

---


WireDomain/Sources/WireDomain/Storage/UpdateEventsUserDefaults/UpdateEventsUserDefaults.swift
```diff
@@ -0,0 +1,43 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public enum UpdateEventsUserDefaults: UserDefaultsProtocol {
```
netbe: I wonder why using this as static property instead of wrapping this in a concrete object

WireDomain/Sources/WireDomain/Storage/UserDefaultsWrapper.swift
```diff
@@ -0,0 +1,93 @@
...
+import Combine
+import Foundation
+
+/// Allows to match for optionals with generics that are defined as non-optional.
+protocol AnyOptional {
+    /// Returns `true` if `nil`, otherwise `false`.
+    var isNil: Bool { get }
+}
+
+extension Optional: AnyOptional {
```
netbe: I have seen this kind of syntactic sugar in the code base and other developers joining fighting against it ending removing it for clarity

WireDomain/Sources/WireDomain/Storage/UpdateEventsUserDefaults/UpdateEventsUserDefaults.swift
```diff
@@ -0,0 +1,43 @@
...
+    public nonisolated(unsafe) static var userID: UUID!
+
+    enum Key: String {
+        case lastEventID
+    }
+
+    @UserDefault(key: Key.lastEventID, userID: userID)
```
johnxnguyen: I would expect to see this declared inside the `UpdateEventRepository`. Is there a particular reason to wrap it in the enum?

WireDomain/Sources/WireDomain/Storage/UpdateEventsUserDefaults/UpdateEventsUserDefaults.swift
jullianm: The purpose is to to provide a **scoped** UserDefaults API that is designed to be simple and intuitive to use for clients who consume it : from other packages, no need for initialization, you can call the static property directly and UserDefaults are scoped depending on their domain concern: for instance, we could have a `enum ConversationUserDefaults: UserDefaultsProtocol` or `enum FederationUserDefaults: UserDefaultsProtocols` etc that would also be exposed by WireDomain.

WireDomain/Sources/WireDomain/Storage/UserDefaultsWrapper.swift
jullianm: This is needed by our propertyWrapper to know whether or not we're dealing with an optional so we can call the `container.removeObject(forKey: scopeKey(key))`  method consequently. This is not exposed by WireDomain, I could make the protocol `private` to enforce that.

WireDomain/Sources/WireDomain/Storage/UpdateEventsUserDefaults/UpdateEventsUserDefaults.swift
jullianm: The point of the enum is to provide a **scoped** UserDefaults API depending on the domain concern. For instance, we could have a `enum ConversationUserDefaults` API or a `enum FederationUserDefaults` (with their related user defaults properties) API exposed publicly by WireDomain so clients could consume it in a simple and intuitive way. This helps make a clear separation of concerns between different defaults storage responsibilities.

johnxnguyen (approved):

netbe (approved):

PR #1934, opened 2024-09-15
feat: create MainCoordinator, MainSplitViewController, MainTabBarController - WPB-10903

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10903" title="WPB-10903" target="_blank">StoryWPB-10903</a>  Create abstraction for navigating in the app (MainSplitViewController, MainTabBarController, MainCoordinator etc.)
  </td></table>
  <br />
 

### Issue

This PR creates a `MainCoordinator` class together with subclasses of UISplitViewController and UITabBarController and several protocols. 
These classes are located in their own target and don't depend on anything but `WireFoundation`.

Except for the `MainTabBarController` most of the code is not yet used on `develop`.
So the only way to test this PR manually is to run the app and ensure nothing has changed.

---

### Checklist

---


WireUI/Tests/WireUIFoundationTests/Resources/ReferenceImages/MainTabBarControllerTests/light.light.png
caldrian, line 1: The icons for the contacts and the folders tab changed, but this is temporary because the tabs will be removed soon.

WireUI/Sources/WireDesign/SDKExtensions/UIKit/UITabBarController+applyMainTabBarControllerAppearance.swift
```diff
@@ -0,0 +1,63 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
+
+public extension UITabBarController {
+
+    /// Sets colors so that this instance looks like the `MainTabBarController` should look like.
+    func applyMainTabBarControllerAppearance() {
```
jullianm, line 24: nitpick: it could be parameterized so we could have the possibility to apply different appearances

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
```diff
@@ -0,0 +1,364 @@
...
+    private weak var newConversation: UIViewController?
+
+    private var selfProfileBuilder: SelfProfileBuilder
+    private weak var selfProfile: UIViewController?
+
+    private var mainSplitViewState: MainSplitViewState = .expanded
+
+    // MARK: - Private Helpers
+
+    private var sidebar: SplitViewController.Sidebar! {
```
jullianm: looks like you don't need to force unwrap here since `splitViewController.sidebar` is already forced unwrapped.

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
```diff
@@ -0,0 +1,364 @@
...
+        splitViewController.sidebar
+    }
+
+    private var conversationList: TabBarController.ConversationList! {
+        switch mainSplitViewState {
+        case .collapsed: tabBarController.conversations?.conversationList
```
jullianm: isn't it a bit risky to force unwrap here ? what if we're in collapsed state and conversations is nil ?

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
```diff
@@ -0,0 +1,364 @@
...
+        case .expanded: splitViewController.conversationList
+        }
+    }
+
+    private var archive: TabBarController.Archive! {
+        switch mainSplitViewState {
+        case .collapsed: tabBarController.archive
+        case .expanded: splitViewController.archive
+        }
```
jullianm: are we sure it is safe to force unwrap here ? what if splitViewController archive is nil ?

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    private var settings: TabBarController.Settings! {
+        switch mainSplitViewState {
+        case .collapsed: tabBarController.settings
+        case .expanded: splitViewController.settings
```
jullianm: same comment for splitViewController.settings

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
```diff
@@ -0,0 +1,364 @@
...
+        mainSplitViewController: SplitViewController,
+        mainTabBarController: TabBarController,
+        newConversationBuilder: NewConversationBuilder,
+        selfProfileBuilder: SelfProfileBuilder
+    ) {
+        guard
+            mainSplitViewController.conversationList != nil,
+            mainTabBarController.archive != nil,
+            mainTabBarController.settings != nil
+        else { fatalError("invalid state of container view controllers") }
```
jullianm: I haven't looked all the code yet but maybe we need to check that splitViewController.settings is not nil and splitViewController.archive is not nil

samwyndham:
@caldrian Is there anything that can be manually tested after merging this PR? If so I think it should be added to the description.

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
```diff
@@ -0,0 +1,364 @@
...
+    private var settings: TabBarController.Settings! {
+        switch mainSplitViewState {
+        case .collapsed: tabBarController.settings
+        case .expanded: splitViewController.settings
+        }
+    }
+
+    // MARK: - Life Cycle
+
+    public init(
```
jullianm: nitpick: we could make the init throwable and throw an explicit error when state of container view controllers is invalid (instead of fatalError).

caldrian:
> Is there anything that can be manually tested after merging this PR? If so I think it should be added to the description.

Unfortunately not. The classes will be used in the epic branch for the navigation overhaul. @samwyndham

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
```diff
@@ -0,0 +1,364 @@
...
+
+    // MARK: - Public Methods
+
+    public func showConversationList(conversationFilter: (some MainConversationFilterRepresentable)?) {
+        defer {
+            // switch to the conversation list tab
+            tabBarController.selectedContent = .conversations
+
+            // apply the filter to the conversation list
+            let mainConversationFilter = conversationFilter?.map()
```
samwyndham: Optional: Looking at this via GitHub diff, it is confusing what `map` does here. Consider renaming to something more explicit like `mainConversationFilter()` or `mappedToMainConversationFilter()`

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
```diff
@@ -0,0 +1,364 @@
...
+    SplitViewController.ConversationList == TabBarController.ConversationList,
+    SplitViewController.Archive == TabBarController.Archive,
+    SplitViewController.Settings == TabBarController.Settings
+{
+    // swiftlint:enable opening_brace
+
+    // MARK: - Private Properties
+
+    private weak var splitViewController: SplitViewController!
+    private weak var tabBarController: TabBarController!
```
jullianm: is there a reason to make them weak ?  having them weak + forced unwrapped feels a bit counter intuitive, also wondering if they need be released from memory at any point.

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
```diff
@@ -0,0 +1,364 @@
...
+/// Both, the split view controller as well as the tab view controller actually install `UINavigationController`
+/// instances and then put or remove the content view controllers into/from `viewControllers` array.
+
+@MainActor
+public final class MainCoordinator<
+
+    SplitViewController: MainSplitViewControllerProtocol,
+    TabBarController: MainTabBarControllerProtocol,
+    NewConversationBuilder: MainCoordinatorInjectingViewControllerBuilder,
+    SelfProfileBuilder: MainCoordinatorInjectingViewControllerBuilder
```
jullianm: do we need to have two dedicated builders since they're conforming to the same protocol, we could have just one it would make the method signature easier to understand since the protocol is basically a view controller build builder.

```
ViewControllerBuilder: MainCoordinatorInjectingViewControllerBuilder 
```

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
```diff
@@ -0,0 +1,364 @@
...
+        dismissArchiveIfNeeded()
+        dismissNewConversationIfNeeded()
+        dismissSettingsIfNeeded()
+        dismissSelfProfileIfNeeded()
+
+        // Move the conversation list from the tab bar controller to the split view controller if needed.
+        if let conversationList = tabBarController.conversations?.conversationList {
+            tabBarController.conversations = nil
+            splitViewController.conversationList = conversationList
+        }
```
samwyndham: Just to confirm, `tabBarController.conversations?.conversationList` is a valid state at this point right` - i.e we don't need to log an error etc. Same question for the similar methods

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
```diff
@@ -0,0 +1,364 @@
...
+
+        // move the settings from the tab bar controller to the split view controller
+        if let settings = tabBarController.settings {
+            tabBarController.settings = nil
+            splitViewController.settings = settings
+        }
+    }
+
+    public func showSelfProfile() {
+        guard selfProfile == nil else { return }
```
samwyndham: Optional: I assume if `selfProfile` is not `nil` this method should not have been called. If that is true, consider logging error / asserting in this case and other similar cases.

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
```diff
@@ -0,0 +1,364 @@
...
+    // MARK: - UITabBarControllerDelegate
+
+    public func tabBarController(_ tabBarController: UITabBarController, didSelect viewController: UIViewController) {
+        guard let tabBarController = tabBarController as? TabBarController,
+              tabBarController === self.tabBarController
+        else { return }
+
+        switch tabBarController.selectedContent {
+        case .contacts, .folders:
+            break // `.contacts` and `.folders` are removed for navigation overhaul
```
samwyndham: 🥳

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
```diff
@@ -0,0 +1,364 @@
...
+
+        case .conversations:
+            switch conversationList.conversationFilter?.map() {
+            case .none:
```
samwyndham: Optional: I think there is duplication with earlier in the file. Consider extracting this logic to a private helper.

WireUI/Sources/WireUIFoundation/Main/Models/MainTabBarControllerContent.swift
```diff
@@ -0,0 +1,26 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+/// The type of tabs of the main tab bar controller shows.
+/// Since UIKit has a type `UITab` this type has been suffixed with "Content".
+public enum MainTabBarControllerContent: Int, CaseIterable {
+    case contacts // will be removed in navigation overhaul
+    case conversations
+    case folders // will be removed in navigation overhaul
+    case archive, settings
```
samwyndham: Any reason settings is on the same line here?

WireUI/Sources/WireUIFoundation/Main/Preview/LabelView.swift
```diff
@@ -0,0 +1,38 @@
...
+            HStack {
+                Spacer()
+                Text(content)
+                    .ignoresSafeArea()
+                Spacer()
+            }
+            Spacer()
+        }.background(backgroundColor)
+    }
+}
```
samwyndham: This is pure bike shedding but I guess you could do something like:

```swift
struct LabelView: View {
    var content: String
    var backgroundColor: Color
    var body: some View {
        Text(content)
            .frame(maxWidth: .infinity, maxHeight: .infinity)
            .ignoresSafeArea()
            .background(backgroundColor)
    }
}
```

WireUI/Sources/WireUIFoundation/Main/Preview/PreviewConversationListViewController.swift
```diff
@@ -0,0 +1,57 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import SwiftUI
+
+final class PreviewConversationListViewController: UIHostingController<LabelView>, MainConversationListProtocol {
```
samwyndham: Just to confirm everything in `WireUI/Sources/WireUIFoundation/Main/Preview` is just preview code right?

WireUI/Sources/WireUIFoundation/Main/Protocols/Containers/MainTabBarControllerProtocol.swift
```diff
@@ -0,0 +1,50 @@
...
+
+/// Defines the contract for between the ``MainCoordinator`` and the ``MainTabBarController``.
+///
+/// The MainTabBarControllerProtocol is a protocol designed to define the essential properties and types that
+/// the ``MainCoordinator`` requires to manage the application's tab-based navigation. This protocol
+/// extends UITabBarController and outlines the key content areas (such as conversations, archive, and settings)
+/// that a conforming tab bar controller must manage.
+
+@MainActor
+public protocol MainTabBarControllerProtocol: UITabBarController {
```
samwyndham: WTF! I never knew this was possible. Is this a new feature of swift?

So basically here we are saying that something that conforms to `MainTabBarControllerProtocol` has to be 1. a subclass of `UITabBarController` and 2. implement the protocol methods. 

Is that correct?

samwyndham (approved):

samwyndham (commented):
Nice Work!

WireUI/Sources/WireUIFoundation/Main/Protocols/Containers/MainTabBarControllerProtocol.swift
caldrian: Correct. But I think it has always been possible like that.

WireUI/Sources/WireUIFoundation/Main/Preview/PreviewConversationListViewController.swift
caldrian: Yes, somehow we could agree on a convention. Either a dedicated subfolder or separate file with "preview" in the name like `SomeView.preview.swift`.

WireUI/Sources/WireUIFoundation/Main/Preview/LabelView.swift
caldrian: Nice, thanks!

WireUI/Sources/WireUIFoundation/Main/Models/MainTabBarControllerContent.swift
caldrian: For navigation overhaul it will be `case conversations, archive, settings`.
For `develop` I had to inject the other two cases and left a comment.
Anyway, it will be gone soon, hopefully.

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
caldrian: I wanted to log, however, the `WireLogger` types are in `WireSystem` which is an Xcode project and cannot be imported here.

WireUI/Sources/WireDesign/SDKExtensions/UIKit/UITabBarController+applyMainTabBarControllerAppearance.swift
caldrian: Unfortunately I don't understand this suggestion.
What I was trying to do here is having any styling/appearance related code combined in the `WireDesign` target.
We have one specification of how the main tab bar controller has to look like (consisting of the colors of icons and titles, background colors etc.).
The `applyMainTabBarControllerAppearance` makes the changes to a provided instance of `UITabBarController`.
So in my opinion it fits the single responsibility principle.

Could you maybe provide a little code snippet, please?

WireUI/Sources/WireUIFoundation/Main/Protocols/Containers/MainTabBarControllerProtocol.swift
samwyndham: Interesting. Always good to learn new swift features!

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
samwyndham: 👍 Oh yeah. We should extract it soon then

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
caldrian: You mean to have one builder for various view controllers? I think we would loose flexibility.

WireUI/Sources/WireDesign/SDKExtensions/UIKit/UITabBarController+applyMainTabBarControllerAppearance.swift
jullianm: I understand. if there is only one specification regarding that main tab bar controller, I guess there is no need to have it parameterized. then you can dismiss my comment.

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
jullianm: Yes exactly,  the single builder would be responsible of building the various controllers related to the main coordinator, you would still be able to inject it, the MainCoordinator signature would be a little simpler while still providing a well-defined scoped of what a main coordinator view controller builder should do.

But if you planned it differently by having separated builders, it's fine too 👍.

WireUI/Sources/WireUIFoundation/Main/Protocols/Coordinator/MainCoordinatorProtocol.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+
+public protocol MainCoordinatorProtocol: AnyObject {
+    func showConversationList<ConversationFilter>(conversationFilter: ConversationFilter?) async
+        where ConversationFilter: MainConversationFilterRepresentable
+    func showArchivedConversations() async
+    func showSelfProfile() async
+    func showSettings() async
+    func showNewConversation() async
+}
```
jullianm: related to my comment above, the MainCoordinator builder could have similar builder methods (buildSelfprofile, buildNewConversation..), so we have a consistency between the MainCoordinator protocol methods and MainCoordinator view controller builder methods

jullianm (approved):
looks good, some minor comments. nice work 👍

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
caldrian: Fixed, thanks!

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
caldrian: Done, thanks.

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
caldrian: In my opinion we should throw errors only for issues which could go wrong at runtime due to specific circumstances (network, memory, user input or action), unless we're writing a library for somebody else.

Initializing the MainCoordinator in a wrong way is a mistake on the developer's side. For those case I usually want fail hard so that the issue is discovered early.

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
caldrian: I actually need to remove these checks because of circular dependencies. Some parts of the setup need to be done after the MainCoordinator's init.

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
caldrian: The logic in this class should be consistent so that those cases can't happen.
I sometimes keep a force unwrap type so that a unit test could check for `nil` without failing hard.

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
caldrian: See my comment a few lines below.

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
caldrian: Same argument here: by declaring a property force-unwrapped, I want the user of the property to be sure that a crash never happens.
On the other side before shipping such a code I want to ensure that I'm not wrong by using unit tests.

A little similar example:
```
struct SomeStruct {
    var forceUnwrapped: Int! {
        return nestedForceUnwrapped
    }
    var nestedForceUnwrapped: Int! {
        return nil
    }
}

let sut = SomeStruct()
XCTAssertNotNil(sut.forceUnwrapped) // won't crash here
```

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
caldrian: I'm not 100% sure if it's required but it shouldn't hurt neither. 
What I want to avoid is a retain cycle.
Since we pass the main coordinator into content view controllers (like the conversation list), which themselves a retained by their container view controllers (split/tabbar), if we also keep a strong reference from the main coordinator to the container view controllers, won't it be a cycle?

WireUI/Sources/WireUIFoundation/Main/MainCoordinator.swift
caldrian: Please see the comment below.

PR #1936, opened 2024-09-16
chore:  Remove `nseV2` feature flag and `SimpleNotificationService` related code

samwyndham:
### Issue

A bug was reported where a user on the `Beta` app did not receive any notifications. This user was affected because they had the `nseV2` enabled which makes our notification service use the `SimpleNotifcationService` instead of the `LegacyNotificationService` which is used in production apps.

During investigation I found a bug in `SimpleNotifcationService` in the following code:

```swift
guard httpResponse.value(forHTTPHeaderField: "Content-Type") == request.acceptType.rawValue else {
    throw NetworkError.invalidResponse
}
```

The issue is that `request.acceptType.rawValue == application/json` whereas `httpResponse.value(forHTTPHeaderField: "Content-Type") == application/json;charset=utf-8` and therefore an error is thrown.

Due to the ongoing work related to changing how sync is done, `nseV2` will never be enabled in production.

Therefore in this PR I delete the  `nseV2` feature flag and related code from the "enabled" path, to avoid other beta testers running into this issue.

### Testing

Run the automated tests.

### Checklist


KaterinaWire (approved):

johnxnguyen (approved):
Goodbye! 🧨

PR #1940, opened 2024-09-17
refactor: process user add client event - WPB-10189

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10189" title="WPB-10189" target="_blank">TaskWPB-10189</a>  [iOS] Process UserClientAddEvent
  </td></table>
  <br />
 

### Key points

Specifically, this PR is about porting the existing implementation of the `UserAddClient` event.

### Testing

- a new user client is added if this client doesn't already exist
- a new user client is not added if it already exists

### Checklist

---

### DIFF
--- a/WireDomain/Sources/WireDomain/Repositories/User/UserRepository.swift
+++ b/WireDomain/Sources/WireDomain/Repositories/User/UserRepository.swift
@@ -65,7 +65,7 @@ public protocol UserRepositoryProtocol {
         from remoteClient: WireAPI.UserClient,
         isNewClient: Bool
     ) async throws
-    
+
     /// Adds a legal hold request.
     ///
     /// - parameters:
@@ -247,7 +247,7 @@ public final class UserRepository: UserRepositoryProtocol {
             try context.save()
         }
     }
-    
+
     // MARK: - Private
 
     private func persistUser(from user: WireAPI.User) {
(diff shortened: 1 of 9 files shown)


WireDomain/Sources/WireDomain/Repositories/User/UserRepository.swift
```diff
@@ -89,6 +103,67 @@ public final class UserRepository: UserRepositoryProtocol {
...
+                if let existingClient = UserClient.fetchExistingUserClient(
+                    with: userClient.id,
+                    in: context
+                ) {
+                    return (client: existingClient, isNew: false)
+                } else {
+                    let newClient = UserClient.insertNewObject(in: context)
+                    return (client: newClient, isNew: true)
+                }
+            }()
```
netbe: I wonder if it would be clearer to have one function for inserting and one for updating on the repository and call that from the processor.

Or just rename this method to creatorupdate
What do you think ?

netbe (commented):
LGTM, left a comment

netbe (approved):

WireDomain/Sources/WireDomain/Repositories/User/UserRepository.swift
johnxnguyen: I think it's a good idea to have two separate methods as you describe, one for fetch or create, and one for updating metadata. This is a pattern that we can apply to other repositories too.

johnxnguyen (approved):

WireDomain/Sources/WireDomain/Repositories/User/UserRepository.swift
jullianm: makes sense, I did the fixes + added some UTs, thx 👍

PR #1942, opened 2024-09-18
fix: sending message large group - WPB-11019

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11019" title="WPB-11019" target="_blank">BugWPB-11019</a>  [iOS]: In large group chats messages take long time to deliver
  </td></table>
  <br />
 

### Issue

On big groups, sending a message takes a lot of time with corecrypto via Proteus resulting quite often in the message not getting sent (expired).

This is due to two reasons: 
1. setting the expiration of the message is set before the encryption of the message so in big groups where we have a lot clients to encrypt a message for it can reach the timeout of 30 seconds.
2. For each encrypt data, we acquire and release the lock on corecrypto database to protect from other processes (NSE or Share Extension). During this process we also `restoreFromDisk` the information of corecrypto.

### Solution

1. set expiration date just before the message is sent via the API (after the encryption of message payload)
2. reduce the number of calls to encrypt the data for each clients of each users (of each domain) to a single call where we encrypt all necessary data for all proteusSessions (aka all clients at once)

To achieve this second point:
* extracted the information from CoreData models with a `MessageInfoExtractor`
* encrypt the payload and wrap it in protobuf objects with `ProteusMessagePayloadBuilder`
* pass the payload to the MessageAPI

### TODO

* [x] testing sending message | availability
* [ ] write tests for new objects

### Notes

I left the old encryptForTransport methods in order not to break legacy methods (WireCryptoBox) that could be used in tests - this will be cleaned in another PR.

### Testing

1. Send message in big groups (>100 users)
2. Check time spent looking for " "send message - attemptToSend duration:" logs

---

### Checklist

---


wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
```diff
@@ -177,23 +167,23 @@ class MessageAPIV1: MessageAPIV0 {
         .v1
     }
 
-    override func broadcastProteusMessage(message: any ProteusMessage) async throws -> (Payload.MessageSendingStatus, ZMTransportResponse) {
+    override func broadcastProteusMessage(message encryptedMessage: Data,
+                                          expirationDate: Date?) async throws -> (Payload.MessageSendingStatus, ZMTransportResponse) {
```
netbe: do we need expiration on broadcast messages

wire-ios-request-strategy/Sources/APIs/MessageAPITests.swift
```diff
@@ -0,0 +1,19 @@
+//
```
netbe: to remove

wire-ios-request-strategy/Sources/Message Sending/ProteusMessagePayloadBuilder.swift
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        // Message too big?
+        if  UInt(messageData.count) > ZMClientMessage.byteSizeExternalThreshold && externalData == nil {
+            // The payload is too big, we therefore rollback the session since we won't use the message we just encrypted.
+            // This will prevent us advancing sender chain multiple time before sending a message, and reduce the risk of TooDistantFuture.
+            messageData = try await encryptForTransportExternalDataBlob(message: messageInfo.genericMessage, messageInfo: messageInfo)
+        }
+
+        // TODO: Reset all failed sessions. -> [F] why do we reset the failedToEstablishSession on all clients ??
```
netbe: double check this

wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
typfel: The only usage for broadcast messages is send availabillity status updates, IMO it doesn't make sense to set an explicit expiration date on these messages.

wire-ios-request-strategy/Sources/Object Syncs/Helpers/ProteusMessage.swift
```diff
@@ -25,8 +25,57 @@ public protocol ProteusMessage: OTREntity, EncryptedPayloadGenerator {
 
     /// Sets the expiration date with the default time interval.
     func setExpirationDate()
+
+    /// Updates the underlying message - TODO: check naming
+    func updateUnderlyingMessageIfNeeded() async throws
```
johnxnguyen: Is this called before sending? If so, we could call it `prepareMessageForSending()`

wire-ios-request-strategy/Sources/Request Strategies/Generic Message/GenericMessageEntity.swift
```diff
@@ -68,6 +70,10 @@ import Foundation
         // no-op
     }
 
+    public func updateUnderlyingMessageIfNeeded() async throws {
+        // no-op - TODO: double check this
+    }
```
johnxnguyen: I guess there's nothing specific to do here.

wire-ios-request-strategy/Sources/Message Sending/ProteusMessagePayloadBuilder.swift
johnxnguyen: I don't know off the top of my head but looking at the code...

The flag is set to true when we failed to fetch a prekey for a client, or we failed to establish a session with a prekey (maybe it was invalid).

What happens if  `failedToEstablishSession` is `true`? Then the existing code (on develop) has this check:

```swift
guard await !client.failedToEstablishSession else {
    // If the session is corrupted, we will send a special payload.
    let data = ZMFailedToCreateEncryptedMessagePayloadString.data(using: .utf8)!
    WireLogger.proteus.error("Failed to encrypt payload: session is not established with client: \(await client.loggedId)")
    return await client.proteusClientEntry(with: data)
}
```

This sends a special "bomb" payload and is invoked when trying to create the client entry. This happens directly before trying to encrypt for that client. So my guess is that if we don't reset the flag, then we will never attempt to encrypt for that client but always send the bomb payload. I'm assuming that if we reset the flag, we probably try to re-establish the session again.

wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
netbe: @typfel why because if they fail of bad network, we don't display the error to the user ?

wire-ios-request-strategy/Sources/Message Sending/ProteusMessagePayloadBuilder.swift
```diff
@@ -0,0 +1,166 @@
...
+        let clientEntries = userClientDatas.compactMap { userClientData in
+            let clientId = Proteus_ClientId.with({ $0.client = userClientData.sessionID.clientID.hexRemoteIdentifier })
+
+            if let data = userClientData.data {
+                return Proteus_ClientEntry(withClientId: clientId, data: data)
+            }
+            if let encryptedData = encryptedDatas[userClientData.sessionID.rawValue] {
+                return Proteus_ClientEntry(withClientId: clientId, data: encryptedData)
+            } else {
+                assertionFailure("should not happen")
```
netbe: for some reasons, this does happen, giving X proteusSessions to encryptBatch and receiving only Y [String: Data] values. Y< X.

wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
typfel: It makes sense to have some timeout on the request, isn't there a default timeout?

wire-ios-request-strategy/Sources/Message Sending/ProteusMessagePayloadBuilder.swift
netbe: actually it's when session is not establish, and it's normal as missing clients will be handled in MessageSender

wire-ios-request-strategy/Sources/Object Syncs/Helpers/ProteusMessage.swift
netbe: updated

wire-ios-request-strategy/Sources/Message Sending/ProteusMessagePayloadBuilder.swift
netbe: added a function `resetAllUserClientsFailedSessions` called before the message is sent

.github/workflows/_reusable_app_release.yml
```diff
@@ -82,7 +82,7 @@ jobs:
       
   build_and_release:
     needs: changelog
-    runs-on: ghcr.io/cirruslabs/macos-runner:sonoma
+    runs-on: ghcr.io/cirruslabs/macos-runner:sonoma-xcode:15.4
```
netbe: temp fix for CI

wire-ios-request-strategy/Sources/Message Sending/MessageInfo.swift
```diff
@@ -0,0 +1,72 @@
...
+        }
+        return result
+    }
+    
+    func resetAllUserClientsFailedSessions(in context: NSManagedObjectContext) async {
+        await context.perform {
+            userClients.forEach { $0.failedToEstablishSession = false }
+            context.saveOrRollback()
+        }
+    }
```
netbe: eventually this would be rework to keep MessageInfo free from CoreData

wire-ios-transport/Source/TransportSession/TransportSession.swift
```diff
@@ -68,7 +68,7 @@ public extension ZMTransportSession {
         // If not data is transmitted for this amount of time for a request, it will time out.
         // <https://wearezeta.atlassian.net/browse/MEC-622>.
         // Note that it is ok for the request to take longer, we just require there to be _some_ data to be transmitted within this time window.
-        configuration.timeoutIntervalForRequest = 60
+        configuration.timeoutIntervalForRequest = 30
```
netbe, line 71: changed to match Android value

wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
netbe: yes, the ones from URLSession so 30 sec without data transmission and 12min for the full request to complete

johnxnguyen (approved):
Looks good

wire-ios-request-strategy/Sources/APIs/MessageAPI.swift
```diff
@@ -60,21 +60,13 @@ class MessageAPIV0: MessageAPI {
         self.httpClient = httpClient
     }
 
-    func broadcastProteusMessage(message: any ProteusMessage) async throws -> (Payload.MessageSendingStatus, ZMTransportResponse) {
+    func broadcastProteusMessage(message encryptedMessage: Data) async throws -> (Payload.MessageSendingStatus, ZMTransportResponse) {
         let path = "/broadcast/otr/messages"
 
-        // FIXME: [WPB-5499] move encryption out of the API - [jacob]
```
typfel, line 66: 🙌

wire-ios-request-strategy/Sources/Message Sending/ProteusMessagePayloadBuilder.swift
typfel: > added a function `resetAllUserClientsFailedSessions` called before the message is sent

I don't think you should do this after each message sent, as John explained a client will have `failedToEstablishSession = true` when we get a broken prekey when trying to establish a session with the client. Resetting this flag will not repair the session.

It only makes sense to reset this flag after you've successfully established session with the client. For example after tapping "reset session" and this time you get a valid prekey.

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
```diff
@@ -163,10 +190,27 @@ public final class MessageSender: MessageSenderInterface {
...
+            await context.perform {
+                if message.shouldExpire {
+                    message.setExpirationDate()
+                    self.context.saveOrRollback()
+                }
+            }
+
+            // 3) send it via API
+            await messageInfo.resetAllUserClientsFailedSessions(in: context)
+            let (messageStatus, response) = try await apiProvider.messageAPI(apiVersion: apiVersion).sendProteusMessage(message: messageData, conversationID: conversationID, expirationDate: nil)
```
typfel: Shouldn't you pass the `expirationDate` parameter?

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
```diff
@@ -163,10 +190,27 @@ public final class MessageSender: MessageSenderInterface {
...
+            // set expiration so request can be expired later
+            await context.perform {
+                if message.shouldExpire {
+                    message.setExpirationDate()
+                    self.context.saveOrRollback()
+                }
+            }
+
+            // 3) send it via API
+            await messageInfo.resetAllUserClientsFailedSessions(in: context)
```
typfel: I think you should remove this, see answer in thread.

typfel (changes requested):

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
netbe: thread which one?

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
netbe: nice catch

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
netbe: removed

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
netbe: done

wire-ios-request-strategy/Sources/Message Sending/ProteusMessagePayloadBuilder.swift
netbe: ok removed the code

wire-ios-request-strategy/Sources/Message Sending/MessageInfoExtractor.swift
```diff
@@ -0,0 +1,200 @@
...
+                guard let sessionID = $0.proteusSessionID,
+                      $0.remoteIdentifier != selfClientID else {
+                    // skips self client session
+                    WireLogger.proteus.warn("skips cliend id: \(String(describing: $0.remoteIdentifier)), proteusSession id: \(String(describing: $0.proteusSessionID))")
+                    return nil
+                }
+
+                guard !$0.failedToEstablishSession else {
+                    let data = ZMFailedToCreateEncryptedMessagePayloadString.data(using: .utf8)!
+                    WireLogger.proteus.error("Failed to encrypt payload: session is not established with client: \(String(describing: $0.remoteIdentifier))")
```
typfel:
```suggestion
                    WireLogger.proteus.error("Failed to encrypt payload: session couldn't be established with client: \(String(describing: $0.remoteIdentifier))")
```

It's not that the session is not established, it's that we tried to establish it and we failed.

typfel (approved):

PR #1956, opened 2024-09-23
fix: Biometrics app lock - WPB-11014

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11014" title="WPB-11014" target="_blank">BugWPB-11014</a>  [iOS] : When  biometry (passcode) protection enabled - Have to close and restart to use Wire
  </td></table>
  <br />
 

### Issue

When a user has password `Lock With Passcode` enabled, and the `AppLock` shows when returning from the background, we filter out calls to initiate authentication if the application state is not [UIApplicationStateActive](https://developer.apple.com/documentation/uikit/uiapplicationstate/uiapplicationstateactive?language=objc). This was a fix for another bug where users may be presented with the authentication UI twice, once caused by the appearance of the app lock and once in response to a `UIApplication.willEnterForegroundNotification` notification - at least this I my assumption based on the code and code comments. Unfortunately the bugfix PR and tickets are not available as this was an old code fix. 

However, in my tests the initiate authentication action due to viewDidAppear is filtered out as the application state at that point on my device / os version is [UIApplicationStateInactive](https://developer.apple.com/documentation/uikit/uiapplicationstate/uiapplicationstateinactive?language=objc) (which is still a foreground state). 

My assumption is that this bug might be iOS version dependent. Therefore with this PR I change how things work a little:

- We only considerer the first `viewDidAppear` when showing the authentication due to `viewDidAppear`.
- We begin observing `UIApplication.willEnterForegroundNotification` after  `viewDidAppear`.

I hope with these changes that we always show the authentication when required, but don't show it twice.

One thing I was not sure about is where to add the logic for this new behavior - is this view following Viper? I haven't really worked with Viper before. Is my implementation OK or should the logic be moved to the presenter or interactor?

### Testing

**Please try both on devices with `Face ID` and `Touch ID` and with multiple `iOS` versions.**

#### Test Case 1

1. Enable `Lock With Passcode` in Account > Settings > Options
3. Wait over 1 minute
4. Bring the app to the foreground -> `Face/Touch ID` is shown
5. Authenticate with `Face/Touch ID`

#### Test Case 2

Try to break things related to  `Face/Touch ID`

---

### Checklist

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Presenter.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Presenter.swift
@@ -69,11 +69,8 @@ extension AppLockModule.Presenter: AppLockPresenterViewInterface {
         switch event {
         // In iOS 14, it was found that 'viewDidAppear' may be invoked even when the app is in the background.
         // To prevent re-authentication when the app is in the background, there is the 'requireForegroundApp' parameter.
-        case .viewDidAppear, .unlockButtonTapped:
-            interactor.executeRequest(.initiateAuthentication(requireForegroundApp: true))
-
-        case .applicationWillEnterForeground:
-            interactor.executeRequest(.initiateAuthentication(requireForegroundApp: false))
+        case .viewDidFirstAppear, .unlockButtonTapped, .applicationWillEnterForeground:
+            interactor.executeRequest(.initiateAuthentication)
 
         case .passcodeSetupCompleted, .customPasscodeVerified:
             interactor.executeRequest(.openAppLock)
--- a/wire-ios/Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Presenter.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Presenter.swift
@@ -67,12 +67,13 @@ extension AppLockModule.Presenter: AppLockPresenterViewInterface {
 
     func processEvent(_ event: AppLockModule.Event) {
         switch event {
-        // In iOS 14, it was found that 'viewDidAppear' may be invoked even when the app is in the background. To prevent re-authentication when the app is in the background, there is the 'requireActiveApp' parameter.
+        // In iOS 14, it was found that 'viewDidAppear' may be invoked even when the app is in the background.
+        // To prevent re-authentication when the app is in the background, there is the 'requireForegroundApp' parameter.
         case .viewDidAppear, .unlockButtonTapped:
-            interactor.executeRequest(.initiateAuthentication(requireActiveApp: true))
+            interactor.executeRequest(.initiateAuthentication(requireForegroundApp: true))
 
         case .applicationWillEnterForeground:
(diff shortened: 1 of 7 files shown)


johnxnguyen (commented):
Looks fine to me, but I'm wondering if we could find a more direct approach. Do you think this could work:
- Use an Actor to keep track of whether authentication has been initiated
- Every time the app enters the foreground, only initiate authentication if it hasn't been and it is needed
- After authentication, set the state on the actor

samwyndham:
> Looks fine to me, but I'm wondering if we could find a more direct approach. Do you think this could work:
> 
> * Use an Actor to keep track of whether authentication has been initiated
> * Every time the app enters the foreground, only initiate authentication if it hasn't been and it is needed
> * After authentication, set the state on the actor

@johnxnguyen  I agree this is a better way to go. I suggest I don't implement it now. I'll will get this fix QA-ed and if it fails for some reason change approach

wire-ios/Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Presenter.swift
```diff
@@ -67,12 +67,10 @@ extension AppLockModule.Presenter: AppLockPresenterViewInterface {
 
     func processEvent(_ event: AppLockModule.Event) {
         switch event {
-        // In iOS 14, it was found that 'viewDidAppear' may be invoked even when the app is in the background. To prevent re-authentication when the app is in the background, there is the 'requireActiveApp' parameter.
-        case .viewDidAppear, .unlockButtonTapped:
-            interactor.executeRequest(.initiateAuthentication(requireActiveApp: true))
-
-        case .applicationWillEnterForeground:
-            interactor.executeRequest(.initiateAuthentication(requireActiveApp: false))
+        // In iOS 14, it was found that 'viewDidAppear' may be invoked even when the app is in the background.
```
netbe, line 70: I guess as we don't support iOS14 it's not relevant anymore, were you able to confirm it's fixed on newest iOS versions?

netbe (approved):

johnxnguyen (approved):

KaterinaWire (approved):

wire-ios/Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Presenter.swift
samwyndham: My assumption based on when the comment was written is that it became broken in iOS 14. I have tested my fix on iOS 18 device and iOS 17.5 simulator. I _think_ my fix would also keep working on iOS 14. However, I think it needs a good round of QA

samwyndham: I will test myself on other sim versions (iOS 15 etc) soon, before merging any cherrypick into the release branch

samwyndham: I have now tested on a range of simulators down to iOS 15.2 without issue. iOS 15.0 simulator doesn't work at all on my system but this is not related to the app.

PR #1957, opened 2024-09-23
feat: Add share profile page screen - WPB-10542

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10542" title="WPB-10542" target="_blank">StoryWPB-10542</a>  [iOS] Share Profile page
  </td></table>
  <br />
 

### Issue
This PR introduces a QR code screen (but **without** 2 tabs):
Screenshot 2024-09-25 at 09 41 02

### Testing

---

### Checklist

---


wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/QR Code/Models/UserQRCodeViewModel.swift
```diff
@@ -0,0 +1,78 @@
...
+            borderColor.setFill()
+            context.cgContext.fill(borderRect)
+
+            overlayImage.draw(in: CGRect(x: xPosition, y: yPosition, width: overlaySize.width, height: overlaySize.height))
+        }
+
+        return combinedImage
+    }
+
+}
```
jullianm: maybe it makes sense to move that extension to WireFoundation in Utilities folder, we have a UIImage extension

wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/QR Code/Views/QRCodeView.swift
```diff
@@ -0,0 +1,114 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import SwiftUI
+
+struct QRCodeView: View {
+
+    // MARK: - Properties
+
+    @ObservedObject var viewModel: UserQRCodeViewModel
```
jullianm, line 25: shouldn't it be a @StateObject here ?

jullianm (approved):
looks good, just let a few minor comments

wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/ProfileHeaderViewController.swift
```diff
@@ -248,6 +249,8 @@ final class ProfileHeaderViewController: UIViewController {
             teamObserver = TeamChangeInfo.add(observer: self, for: team)
         }
         view.backgroundColor = .clear
+
+        qrCodeButton.addTarget(self, action: #selector(qrCodeButtonTapped), for: .touchUpInside)
     }
```
agisilaos: We're trying to move away from the selector pattern. What we can use instead is a UIAction. 
```suggestion
        let qrCodeAction = UIAction { _ in
    self.qrCodeButtonTapped()
}
qrCodeButton.addAction(qrCodeAction, for: .touchUpInside)
    }
```

As a result, we can remove the @objc annotation from the `qrCodeButtonTapped` method.

wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/QR Code/Models/UserQRCodeViewModel.swift
agisilaos: +1

wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/QR Code/Views/QRCodeView.swift
agisilaos: @KaterinaWire can tell us but I think In this case, **@ObservedObject** is likely the correct choice for viewModel: **UserQRCodeViewModel**. Here's why:

The **QRCodeView** is probably not the owner of this view model. It's likely created and passed in from a parent view or coordinator.

**@ObservedObject** is typically used when the object is created elsewhere and passed into the view, which seems to be the case here.

However, if **QRCodeView** was actually responsible for creating and owning the **UserQRCodeViewModel**, then **@StateObject** would be more appropriate

agisilaos (commented):
Left some comments

wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/QR Code/Views/QRCodeView.swift
KaterinaWire: @jullianm @agisilaos  As far as I understand the difference between `ObservedObject` and `StateObject`, we should use `ObservedObject` since we are injecting `viewModel` and we don't need to save viewModel's state since it is being kept at higher level.
There's an example here: https://developer.apple.com/documentation/swiftui/observedobject

jullianm: Right, makes sense since it is injected from a higher level, I didn't see that. all good then!

wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/QR Code/Models/UserQRCodeViewModel.swift
KaterinaWire: done

agisilaos (approved):
Approved. Tests are failing though, please take a look before merging! 👍

PR #1958, opened 2024-09-24
refactor: process user update event - WPB-10200

jullianm:
WPB-10200

### Key points

Specifically, this PR is about porting the existing implementation of the `UserUpdate` event.

### Testing

- a user is properly updated locally after the event is processed

### Checklist

---


WireDomain/Sources/WireDomain/Repositories/User/UserRepository.swift
```diff
@@ -89,8 +132,75 @@ public final class UserRepository: UserRepositoryProtocol {
...
+        try await context.perform { [self] in
+
+            let user = fetchOrCreateUser(
+                with: event.id
+            )
+
+            if isFederationEnabled {
+                user.remoteIdentifier = event.userID.uuid
+                user.domain = event.userID.domain
+            }
```
netbe: question: isn't it only the domain that we should set ?
```suggestion
            user.remoteIdentifier = event.userID.uuid
            if isFederationEnabled {
               user.domain = event.userID.domain
            }
```

WireDomain/Tests/WireDomainTests/Repositories/UserRepositoryTests.swift
```diff
@@ -44,7 +44,8 @@ class UserRepositoryTests: XCTestCase {
         usersAPI = MockUsersAPI()
         sut = UserRepository(
             context: context,
-            usersAPI: usersAPI
+            usersAPI: usersAPI,
+            isFederationEnabled: true
```
netbe: do we need an extra test for isFederationEnabled: false?

netbe (commented):
Added some comments to check

WireDomain/Sources/WireDomain/Repositories/User/UserRepository.swift
jullianm: yes, fixed it

WireDomain/Tests/WireDomainTests/Repositories/UserRepositoryTests.swift
jullianm: done

WireDomain/Sources/WireDomain/Repositories/User/UserRepository.swift
```diff
@@ -89,8 +132,74 @@ public final class UserRepository: UserRepositoryProtocol {
...
+            if let accentColor = event.accentColorID {
+                user.accentColorValue = Int16(accentColor)
+            }
+
+            let assetKeys: Set<String> = [
+                ZMUser.previewProfileAssetIdentifierKey,
+                ZMUser.completeProfileAssetIdentifierKey
+            ]
+
+            /// Do not update assets if user has local modifications.
```
johnxnguyen: Hmm... I wonder why 🤔

WireDomain/Sources/WireDomain/Repositories/User/UserRepository.swift
```diff
@@ -89,8 +132,74 @@ public final class UserRepository: UserRepositoryProtocol {
...
+
+            user.isPendingMetadataRefresh = false
+
+            try context.save()
+        }
+    }
+
+    // MARK: - Private
+
     private func persistUser(from user: WireAPI.User) {
```
johnxnguyen, line 517: There seems to be some overlap with this method and the `updateUser` method. Maybe not now, but do you think it's worth at some point to find a solution that minimizes this duplication? I suppose it could be achieved with an intermediate `UpdateInfo` struct. Maybe something to consider if we introduce a proper data layer.

WireDomain/Tests/WireDomainTests/Repositories/UserRepositoryTests.swift
```diff
@@ -149,6 +204,18 @@ class UserRepositoryTests: XCTestCase {
...
+            id: userID,
+            userID: UserID(uuid: userID, domain: domain),
+            accentColorID: nil,
+            name: "username",
+            handle: "test",
+            email: "test@wire.com",
+            isSSOIDDeleted: nil,
+            assets: nil,
+            supportedProtocols: [.proteus, .mls]
+        )
```
johnxnguyen: I think it may be good to also have some more varying data, with some more `nil`, so see that only new changes are persisted and old values are preserved in case of `nil` values.

WireDomain/Sources/WireDomain/Repositories/User/UserRepository.swift
jullianm: Not sure why either, the documentation on `modifiedKeys` indicates:  `/// Keys that have been modified since the last save and not been synced and reset`. So, my assumption would be if user has local changes to its assets we don't want to update them and keep these changes until they're synced.

WireDomain/Sources/WireDomain/Repositories/User/UserRepository.swift
jullianm: Yes, I did some refactoring to avoid code duplication in this [ticket](https://github.com/wireapp/wire-ios/pull/1999) using an intermediate `updateUserMetadata(..)` func. So I'll be able to reuse it. 
I'll just mark a comment for now on this one.

jullianm: done in https://github.com/wireapp/wire-ios/pull/1958/commits/411d1ab958b6f0aa0735b39903db173c84783820

WireDomain/Tests/WireDomainTests/Repositories/UserRepositoryTests.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/1958/commits/411d1ab958b6f0aa0735b39903db173c84783820

jullianm:
@netbe @johnxnguyen  commit https://github.com/wireapp/wire-ios/pull/1958/commits/1919147f848f2ee5da5ab2265d3f17db1b116ff0 removes the QualifiedID object from the UserUpdate event. 
Following our discussion with backend and Android team, we only get a user update event for users belonging to our own domain (events don't go across federation), so I removed some unnecessary code.

WireDomain/Sources/WireDomain/Repositories/User/UserRepository.swift
johnxnguyen: Maybe worth adding that in the comment, even if it's just a guess.

johnxnguyen (approved):

netbe (approved):

WireDomain/Sources/WireDomain/Repositories/User/UserRepository.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/1958/commits/7457f6e25024c3c1f13930565644358273aa5112

PR #1969, opened 2024-09-25
fix: broken linking in SyncEngine for Countly - no ticket

netbe:
### Issue

Fixes the last feat commit with rebuilding the pbxproject.

### Testing

* run tests

---

### Checklist

---


agisilaos (approved):

WireAPI/Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.10
+// swift-tools-version: 6.0
```
samwyndham, line 1: Is setting `6.0` an important part of the fix? I think in many cases this will break migration to Xcode 16 - have you tested this?

netbe: not sure if it was needed, i reverted the change, I am checking if the CI can build first like this. to my understanding that comes from develop, no?

netbe: at the end do we need separate package files ?

netbe:
* [ ] checking if https://github.com/wireapp/wire-ios/actions/runs/11035872107 builds

WireAPI/Package.swift
samwyndham: I assume it is coming from develop, much of it from my PR - https://github.com/wireapp/wire-ios/pull/1869/files which was largely changing `6.0` to ` 5.10` and deleting the `Package@swift-5.10.swift` files

PR #1976, opened 2024-09-26
refactor: execute requests via `NetworkService` - WPB-10448

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10448" title="WPB-10448" target="_blank">TaskWPB-10448</a>  [iOS] Renew access token if needed
  </td></table>
  <br />
 

### Issue

There is a `APIService` and a `PushChannelService`, both of which need to make network requests via `URLSession`. This PR introduces `NetworkService` which factors out very basic network access from both services. As a result, the services become a bit leaner and can operate on a slightly higher level of abstraction since they no longer need to work with `URLSession` directly.

Key points:
- `NetworkService` is responsible for resolving against a base url and validating the response
- `APIService` is now only responsible for attaching an access token
- `PushChannelService` is now only responsible for attaching an access token and returning a `PushChannel`.

---

### Checklist

---


WireAPI/Sources/WireAPI/Network/NetworkService/NetworkService.swift
```diff
@@ -0,0 +1,148 @@
...
+        self.baseURL = baseURL
+    }
+
+    deinit {
+        urlSession?.invalidateAndCancel()
+    }
+
+    func configure(with urlSession: URLSession) {
+        self.urlSession = urlSession
+    }
```
jullianm, line 37: is there a particular reason we use a dedicated method to initialize the urlSession, could we not use the `init(baseURL: URL, session: URLSession)` directly ? we could avoid dealing with an optional

WireAPI/Tests/WireAPITests/Network/NetworkService/NetworkServiceTests.swift
```diff
@@ -0,0 +1,109 @@
...
+        try await super.setUp()
+        backendURL = try XCTUnwrap(URL(string: "https://www.example.com"))
+        sut = NetworkService(baseURL: backendURL)
+        sut.configure(with: .mockURLSession())
+    }
+
+    override func tearDown() async throws {
+        backendURL = nil
+        sut = nil
+        try await super.tearDown()
```
jullianm, line 40: unless there's a specific reason, I think we can call it before settings the properties to nil

WireAPI/Sources/WireAPI/Network/NetworkService/NetworkService.swift
```diff
@@ -0,0 +1,148 @@
...
+        request.url = URL(
+            string: url.absoluteString,
+            relativeTo: baseURL
+        )
+
+        let (data, response) = try await urlSession.data(for: request)
+
+        guard let httpURLResponse = response as? HTTPURLResponse else {
+            throw NetworkServiceError.notAHTTPURLResponse
+        }
```
jullianm, line 58: I don't know if you planned to handle it somewhere else but what do you think of handling errors more explicitly at the NetworkService level ?  Just an example here.

```
func mapResponse(response: (data: Data, response: URLResponse)) throws -> Data {
    guard let httpResponse = response.response as? HTTPURLResponse else {
        throw NetworkServiceError.notAHTTPURLResponse
    }
    
    switch httpResponse.statusCode {
    case 200..<300:
        return response.data
    case 400:
        throw NetworkServiceError.badRequest
    case 401:
        throw NetworkServiceError.unauthorized
    case 403:
        throw NetworkServiceError.forbidden
    case 404:
        throw NetworkServiceError.notFound
    default:
        throw NetworkServiceError.http(httpResponse: httpResponse, data: response.data)
    }
}
```

WireAPI/Tests/WireAPITests/Network/NetworkService/NetworkServiceTests.swift
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    // MARK: - Execute request
+
+    func testItDoesNotExecuteAnInvalidRequest() async throws {
```
jullianm: Not a request for a change but just so you know I started applying this naming convention [here](https://wearezeta.atlassian.net/wiki/spaces/IOS/pages/1434124328/Conventions) in many of my PRs (even if it's just a proposal for now). Maybe we could align on this, if so we could rename the test  : `testExecuteRequest_It_Throws_An_Invalid_Request_Error`

jullianm (approved):
looks good, approving with some minor comments. much cleaner to have this NetworkService factored out 👍

WireAPI/Tests/WireAPITests/Network/NetworkService/NetworkServiceTests.swift
netbe: agree let's find one and stick to it, no strong opinion on it

WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
```diff
@@ -49,10 +49,8 @@ public protocol APIServiceProtocol {
 
 public final class APIService: APIServiceProtocol {
```
netbe, line 50: so are these 3 classes within a new module of WireAPI (Network)?
that would make sense to separate from actual endpoints

WireAPI/Sources/WireAPI/Network/NetworkService/NetworkService.swift
```diff
@@ -0,0 +1,148 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+final class NetworkService: NSObject {
```
netbe, line 21: why do we need NSObject?

WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
netbe: these 3 would start to replace WireTransport right?

netbe (commented):
Just adding some questions to fully understand it

WireAPI/Sources/WireAPI/Network/NetworkService/NetworkService.swift
johnxnguyen: Yes, because the `NetworkService` is the `URLSession`s delegate which can only be set when initializing the session. This means we must initialize the network service first without the session, then the session (passing in the service as the delegate), then link it back to the service.

WireAPI/Tests/WireAPITests/Network/NetworkService/NetworkServiceTests.swift
johnxnguyen: Calling `super.tearDown()` first?

WireAPI/Sources/WireAPI/Network/NetworkService/NetworkService.swift
johnxnguyen: Yes, for general errors that apply to any endpoint (such as authentication or rate limiting errors) then we can handle them here, but we have another ticket for that.

WireAPI/Tests/WireAPITests/Network/NetworkService/NetworkServiceTests.swift
johnxnguyen: ah yep, I forgot already :)

WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
johnxnguyen: yes that's a good idea, we can create a `WireNetwork` target inside the package, but I do that in another PR. And yes, this is the basis for replacing `WireTransport`.

WireAPI/Sources/WireAPI/Network/NetworkService/NetworkService.swift
johnxnguyen: Because it is a delegate of `URLSessionDelegate`

netbe (approved):
good to go

WireAPI/Tests/WireAPITests/Network/NetworkService/NetworkServiceTests.swift
johnxnguyen: done 59fc0edfab22f8e279a0ca8d305a48a501773914

PR #1994, opened 2024-10-02
refactor: process user connection event - WPB-10191

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10191" title="WPB-10191" target="_blank">TaskWPB-10191</a>  [iOS] Process UserConnectionEvent
  </td></table>
  <br />
 

### Key points

Specifically, this PR is about porting the existing implementation of the `UserConnection` event.

When we receive an update connection event, the connection is updated locally.
if the connection is accepted, the processors tries to resolve the 1:1 conversation between a self user and another user depending on their common protocol (whether Proteus, MLS or no common protocol).

When the protocol is MLS, the processor tries to establish a new MLS group or join an existing group (depending on the epoch).

### Testing

- 1:1 conversation are properly resolved between a self user and another user (MLS, Proteus or no common protocol) by the processor
- connection is properly updated locally
- MLS 1:1 conversation is properly fetched from API
- MLS 1:1 conversation is properly pulled and stored locally.

### Checklist

---


WireDomain/Sources/WireDomain/Event Processing/UserEventProcessor/UserConnectionEventProcessor.swift
```diff
@@ -32,9 +33,248 @@ protocol UserConnectionEventProcessorProtocol {
...
+    let conversationsRepository: any ConversationRepositoryProtocol
+    let mlsService: any MLSServiceInterface
+
+    func processEvent(_ event: UserConnectionEvent) async throws {
+        let connection = event.connection
+
+        try await connectionsRepository.updateConnection(connection)
+
+        if connection.status == .accepted {
+            try await resolveOneOnOneConversation(for: connection.receiverQualifiedID)
```
jullianm: I wanted to encapsulate all that resolving 1:1 conversation logic into a `ResolveOneOnOneConversationUseCase` but I wasn't sure whether it made sense. IMO, it makes sense to create use cases and expose them publicly if it's going to be used by WireDomain clients (not sure if that's the case here). 

Looking at `slow sync` implementation it looks like we need to resolve all 1:1 conversations. Maybe it would make sense to create a dedicated 1:1 conversation resolver UseCase that we could reuse here.

jullianm: Following up on my last comment, once this [ticket](https://wearezeta.atlassian.net/browse/WPB-10727) is merged, we'll be able to reuse `OneOnOneResolverUseCase` (that will be available from the mentioned ticket) here to avoid code duplication.

johnxnguyen: When I think about a `use case`, I think about some public business logic defined in the domain layer intended for consumption from the view layer. In other words, the public interface to the domain layer.

Internally within the domain layer, we could still use the word use case but I think then it risks to confuse people about what a use case is (public business logic or reusable code?). 

I think the name `OneOnOneResolver` sufficiently expresses what it does.

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV5.swift
```diff
@@ -38,6 +38,26 @@ class ConversationsAPIV5: ConversationsAPIV4 {
             .success(code: .ok, type: QualifiedConversationListV5.self) // Change in v5
             .parse(response)
     }
+
+    override func getMLSOneToOneConversation(
+        userID: String,
+        in domain: String
+    ) async throws -> Conversation {
+        let resourcePath = "\(pathPrefix)/conversations/one2one/\(domain)/\(userID)"
```
johnxnguyen, line 50: It might be good to check that the domain and useriD are not empty, otherwise we may get a 404 which would be confusing.

WireDomain/Sources/WireDomain/Event Processing/UserEventProcessor/UserConnectionEventProcessor.swift
```diff
@@ -32,9 +33,248 @@ protocol UserConnectionEventProcessorProtocol {
 
 struct UserConnectionEventProcessor: UserConnectionEventProcessorProtocol {
 
-    func processEvent(_: UserConnectionEvent) async throws {
-        // TODO: [WPB-10191]
-        assertionFailure("not implemented yet")
+    enum UserConnectionEventError: Error {
```
johnxnguyen: Perhaps we can keep the name short and simply call it `Error`?

netbe (approved):

johnxnguyen (approved):
Nice work

WireDomain/Sources/WireDomain/Event Processing/UserEventProcessor/UserConnectionEventProcessor.swift
jullianm: That was my first approach indeed: `enum Error: Swift.Error` but then I updated to Xcode16 and now the compiler doesn't like it. It thinks it's a recursive enum.

WireDomain/Sources/WireDomain/Event Processing/UserEventProcessor/UserConnectionEventProcessor.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/1994/commits/dfdde494afe71e0aab769732462b5b730c85f426

WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV5.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/1994/commits/dfdde494afe71e0aab769732462b5b730c85f426

WireDomain/Sources/WireDomain/Event Processing/UserEventProcessor/UserConnectionEventProcessor.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/1994/commits/dfdde494afe71e0aab769732462b5b730c85f426

PR #2004, opened 2024-10-07
fix: backup restore events not tracked - WPB-11308

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11308" title="WPB-11308" target="_blank">BugWPB-11308</a>  [iOS] Backup Restore Succeeded event isn't being logged in Countly
  </td></table>
  <br />
 

### Issue

Backup restore (success and failure) events are not tracked. The events were being sent through the (optional) active user session, but at that point in time there doesn't exist an active user session. The solution is to pass the result of the backup import along to the point where an active user session exists and send the events via the analytics service:

- `BackupRestoreController` passes the result to its delegate (`AuthenticationCoordinator`)
- `AuthenticationCoordinator` enqueues the action `completeBackupStep(didSucceed: Bool?)`
- `AuthenticationCoordinator` executes the action, telling the `UnauthenticatedSession` to report the result. 
- `UnauthenticatedSession` will `upgradeToAuthenticatedSession`, then inform its delegate (`SessionManager`)
- `SessionManager` will create and activate `ZMUserSession`, then send an event to the user session via `userSession.trackAnalyticsEvent(.backupRestored)`.
- If `ZMUserSession` has been configured to send analytics events, it will forward the event to `analyticsEventTracker`.
- Otherwise it add events to `pendingAnalyticsEvents`, and when the `analyticsEventTracker` is set it will forward the pending events.

### Testing

- Log in, restore a backup successfully, enable analytics, check the event is tracked
- Log in, fail to restore a backup (insert wrong password), continue without backup, enable analytics, check the event is tracked

---

### Checklist

---


agisilaos (approved):

wire-ios/Wire-iOS/Sources/AnalytitcsServiceConfigurationBuilder.swift
```diff
@@ -23,14 +23,13 @@ struct AnalyticsServiceConfigurationBuilder {
...
         guard
-            let secretKey = Bundle.countlyAppKey,
             let countlyURL = BackendEnvironment.shared.countlyURL
         else {
             return nil
         }
 
         return AnalyticsServiceConfiguration(
-            secretKey: secretKey,
+            secretKey: "37cc418a6fc07403b0a918252133c15b95486883",
```
netbe: to revert

wire-ios/Wire-iOS/Sources/AnalytitcsServiceConfigurationBuilder.swift
```diff
@@ -23,14 +23,13 @@ struct AnalyticsServiceConfigurationBuilder {
 
     func build() -> AnalyticsServiceConfiguration? {
         guard
-            let secretKey = Bundle.countlyAppKey,
```
netbe, line 26: should stay here, with !secretKey.isEmpty

to test prefer override in debug.xcconfig

netbe (changes requested):
looks good implementation wise, some debug to clean up

wire-ios/Wire-iOS/Sources/AnalytitcsServiceConfigurationBuilder.swift
johnxnguyen: done

wire-ios/Wire-iOS/Sources/AnalytitcsServiceConfigurationBuilder.swift
johnxnguyen: done

netbe (approved):

PR #2005, opened 2024-10-07
fix: accent color picker presentation - WPB-10141

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10141" title="WPB-10141" target="_blank">StoryWPB-10141</a>  [iOS][iPad] Sidebar - Settings&Support
  </td></table>
  <br />
 

### Issue

Currently the accent color picker view contains a `NavigationView` which is similar to a `UINavigationController`. 
This view is pushed onto a navigation controller and results in weird behavior on iPad with split view controller.

### Testing

Open the account settings and change the profile color.

---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPicker.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPicker.swift
@@ -16,12 +16,15 @@
 // along with this program. If not, see http://www.gnu.org/licenses/.
 //
 
-import Inject
 import SwiftUI
 import WireCommonComponents
 import WireDesign
 import WireFoundation
 import WireSyncEngine
+
+#if canImport(Inject)
+import Inject
+#endif
 
 struct AccentColorPicker: View {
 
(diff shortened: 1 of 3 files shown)


wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPicker.swift
```diff
@@ -16,49 +16,50 @@
...
 //
 
-import Inject
 import SwiftUI
 import WireCommonComponents
 import WireDesign
 import WireFoundation
 import WireSyncEngine
 
+#if canImport(Inject)
```
agisilaos: This is not needed.

agisilaos (changes requested):
There is no need for those import of statements. Inject is only part of the debug builds by default.

jullianm (approved):

jullianm:
@caldrian there was a comment regarding why we use a NavigationView pointing out that title would be too small without the NavigationView. Doesn't removing the NavigationView reintroduce that issue ? otherwise looks good to me 👍

wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPicker.swift
```diff
@@ -16,49 +16,50 @@
...
     @ObserveInjection var inject
+#endif
 
     let onColorSelect: ((AccentColor) -> Void)?
 
     var body: some View {
-        // The NavigationView is needed here to properly set the font size for the navigation bar title.
-        // Currently, there's no other way to set the font size of the navigation bar title to the desired value.
-        // Without the NavigationView, the title would be too small.
-        NavigationView {
```
caldrian, line 40: @agisilaos removing this is fine, right?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/AccentColorPicker.swift
caldrian: It might not be needed for Inject to be not included in Release builds, however, it still allows us to compile this code without having Inject linked.

agisilaos (approved):

PR #2006, opened 2024-10-07
fix: duplicate 1-1 conversations - WPB-10260

netbe:
### Issue

**Note:** this PR is based on #2017, so preferably start with #2017 first.

Due to some race condition, an user could end up with duplicate proteus team 1-1 conversation. This PR aims to fix this, by showing only one **same on all platforms** 1-1 conversation.

* Fix one to one migration that was introduced in adding the last rule to choose the 1-1 conversation (sorted qualified id)
* Add coredata migration post action to latest db migration to reset the 1-1 conversation if it was wrongly set before AND merge all messages into the conversation to prepare for MLS migration.
* Remove PatchApplicator (unused)

### Testing

* Use production account where duplicate 1-1 conversation exist
* run the migration
* See messages have been merged into one conversation

---

### Checklist


wire-ios-data-model/Source/ManagedObjectContext/Migration/113-114/OneOnOneConversationMigrationAction.swift
```diff
@@ -97,6 +97,10 @@ final class OneOnOneConversationMigrationAction: CoreDataMigrationAction {
             sameParticipant
         ])
 
+        //  4. sort by their fully qualified conversation ID in ascending oder, and use the first one.
+        // primary_key is basically the qualified id
+        request.sortDescriptors = [NSSortDescriptor(key: "primaryKey", ascending: true)]
```
johnxnguyen, line 102: Is the reason why, given two team one on ones, client A might show one and client B might show the other?

wire-ios-data-model/Source/ManagedObjectContext/Migration/118-119/FixDuplicateOneOnOneConversationsAction.swift
```diff
@@ -0,0 +1,128 @@
...
+
+    private func conversationsPerUser(from conversations: [ZMConversation]) -> [ZMUser: [ZMConversation]] {
+        var fakeOnOnOnePerUser = [ZMUser: [ZMConversation]]()
+        for conversation in conversations {
+            if let otherUser = conversation.localParticipantsExcludingSelf.first {
+                if fakeOnOnOnePerUser[otherUser] == nil {
+                    fakeOnOnOnePerUser[otherUser] = [conversation]
+                } else {
+                    fakeOnOnOnePerUser[otherUser]?.append(conversation)
+                }
```
johnxnguyen: I think you can achieve this with:

```swift
fakeOneOnOnePerUser[otherUser, default: [])?.append(conversation)
```

wire-ios-data-model/Source/ManagedObjectContext/Migration/118-119/FixDuplicateOneOnOneConversationsAction.swift
```diff
@@ -0,0 +1,128 @@
...
+final class FixDuplicateOneOnOneConversationsAction: CoreDataMigrationAction {
+
+    override func execute(in context: NSManagedObjectContext) throws {
+
+        // 1) fetch all conversations to fix
+        guard let fake1On1Conversations: [ZMConversation] = try? fakeOneOnOneConversations(context: context), !fake1On1Conversations.isEmpty else {
+            return
+        }
+
+        let fakeOneOnOnePerUser = conversationsPerUser(from: fake1On1Conversations)
```
johnxnguyen: Elsewhere in our code we often name dictionaries as "valueByKey", so in this case `conversationsByUser`.

wire-ios-data-model/Source/ManagedObjectContext/Migration/118-119/FixDuplicateOneOnOneConversationsAction.swift
```diff
@@ -0,0 +1,128 @@
...
+        // get the draft message if available and none present
+        if theOneOnOneConversation.draftMessageData == nil {
+            theOneOnOneConversation.draftMessageData = otherConversation.draftMessageData
+        }
+        if theOneOnOneConversation.draftMessageNonce == nil {
+            theOneOnOneConversation.draftMessageNonce = otherConversation.draftMessageNonce
+        }
+    }
+
+    private func fakeOneOnOneConversations(context: NSManagedObjectContext) throws -> [ZMConversation] {
```
johnxnguyen, line 86: Does it make sense to define once and reuse it?

wire-ios-data-model/Source/Model/Conversation/ZMConversation+MLS.swift
```diff
@@ -305,6 +305,29 @@ public extension ZMConversation {
 
         return try context.fetch(request)
     }
+
+    static func fetchAllFakeOneOnOneGroupConversations(
```
johnxnguyen, line 309: Maybe rename the file since fake one on ones only concern proteus.

wire-ios-data-model/Source/ManagedObjectContext/Migration/113-114/OneOnOneConversationMigrationAction.swift
netbe: what do you mean? can you give an example?

wire-ios-data-model/Source/ManagedObjectContext/Migration/118-119/FixDuplicateOneOnOneConversationsAction.swift
```diff
@@ -0,0 +1,128 @@
...
+                }
+
+            }
+        }
+        return fakeOnOnOnePerUser
+    }
+
+    private func moveMessages(from otherConversation: ZMConversation,
+                              to theOneOnOneConversation: ZMConversation) {
+        otherConversation.allMessages.forEach { message in
```
samwyndham, line 66: Optional: I don't think the specifics of the content in the method are tested... eg. visible vs hidden, draft properties. Consider testing this if it is not too much trouble.

samwyndham (approved):
Looks good!

wire-ios-data-model/Source/ManagedObjectContext/Migration/118-119/FixDuplicateOneOnOneConversationsAction.swift
netbe: not sure to follow, it breaks down the length of initial function

wire-ios-data-model/Source/ManagedObjectContext/Migration/118-119/FixDuplicateOneOnOneConversationsAction.swift
netbe: done

wire-ios-data-model/Source/ManagedObjectContext/Migration/118-119/FixDuplicateOneOnOneConversationsAction.swift
netbe: done

wire-ios-data-model/Source/ManagedObjectContext/Migration/113-114/OneOnOneConversationMigrationAction.swift
johnxnguyen: I understood that part oft he bug is that if there are duplicate one on ones, then it's not always the same one that is shown across different self clients.  I was asking if the cause was because the sorting of duplicated one on one instances was undefined.

wire-ios-data-model/Source/ManagedObjectContext/Migration/118-119/FixDuplicateOneOnOneConversationsAction.swift
johnxnguyen: I think I was referring to if we need to or should factor out the predicates that define a fake one on one, so we don't have multiple implementations of it (I can't remember now if we do or not.

wire-ios-data-model/Source/ManagedObjectContext/Migration/113-114/OneOnOneConversationMigrationAction.swift
netbe: yes exactly, that's the part that was missing

wire-ios-data-model/Source/ManagedObjectContext/Migration/118-119/FixDuplicateOneOnOneConversationsAction.swift
netbe: we kind of do inside the [migrationAction](https://github.com/wireapp/wire-ios/blob/4e7f7d2a795a49646dacf82b8c8257a729eeefde/wire-ios-data-model/Source/ManagedObjectContext/Migration/113-114/OneOnOneConversationMigrationAction.swift#L85)  but it slightly different.

Here I select all OneOnOneConversations and on the OneOnOneConversationMigrationAction we select all oneOnOneConversations with a given user.

For now I would keep this similar implementation as I don't see benefits or how to factor this out

johnxnguyen (approved):

PR #2008, opened 2024-10-09
fix: cannot start conversation - WPB-11454

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11454" title="WPB-11454" target="_blank">BugWPB-11454</a>  [iOS] : Users cannot start conversation with users with in Team 
  </td></table>
  <br />
 

### Issue

Since supported-protocols were introduced in 3.112.4, we need to know the supported-protocol for the self user and other user for 1-1 conversations. This is either calculated during slowsync or if MLS is enabled during the resolve 1-1 conversations.

There is a bug where supported-protocols are not set for the self user that was logged in before 3.112.4. Indeed if these users update to latest version they don't perform a slowsync (only performed for new client) so they would end up with no supported protocols.

When trying to open a 1-1 conversation in their team, the app would try to resolve the common protocols but endup finding no common protocols and therefore fail with an error.

Solution:

Add a method to calculate the supported-protocols if none exist at the end of the quicksync (so when user put the app on foreground)

### Testing

1. have a build from 3.111.9
2. login
3. update to latest build
4. tap on a new team member to start a 1-1

---

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests.swift
+++ b/wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests.swift
@@ -484,7 +484,6 @@ final class ZMUserSessionTests: ZMUserSessionTestsBase {
         XCTAssertEqual(mockRecurringActionService.performActionsIfNeeded_Invocations.count, 1)
 
         XCTAssertEqual(getFeatureConfigsActionHandler.performedActions.count, 1)
-        XCTAssertEqual(pushSupportedProtocolsActionHandler.performedActions.count, 1)
     }
     
     func test_didFinishQuickSync_CalculateSupportedProtocolsIfNoProtocols() {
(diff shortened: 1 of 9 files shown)


caldrian (approved):

wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests.swift
```diff
@@ -304,11 +304,7 @@ final class ZMUserSessionTests: ZMUserSessionTestsBase {
             results: [.success(())],
             context: syncMOC.notificationContext
         )
-        let pushSupportedProtocolsActionHandler = MockActionHandler<PushSupportedProtocolsAction>(
```
netbe, line 307: this does not happen anymore because the fix will set supported protocols so during the resolve one on one conv, there is no protocols change therefore no push of protocols

agisilaos (approved):

PR #2021, opened 2024-10-11
fix: Crash navigating to notification options - WPB-11062

samwyndham:
### Issue

When building using iOS 18 SDK on iOS 18 devices there is a crash when navigating to the notifications options screen. This is because we are dequeuing the footer from inside `ConversationNotificationOptionsViewController.collectionView(_:layout:referenceSizeForFooterInSection:)` which we should not be doing. 

The fix is to create a separate sizing footer view that we only use for generating the reference size.

### Testing

**This must be tested using an Xcode 16 (iOS 18SDK) build on a iOS 18 device.**

1. Navigate to any group conversation
2. Go to the details (by clicking the title)
3. Tap Notifications
4. There should be no crash and the footer is displayed correctly

---

### Checklist


agisilaos (approved):

caldrian (approved):
nice work!

PR #2028, opened 2024-10-14
refactor: process conversation access update event - WPB-10164

jullianm:
WPB-10164

### Key points

Specifically, this PR is about porting the existing implementation of the `ConversationAccessUpdate` event.

### Testing

- conversation is properly fetched locally
- conversation roles and modes are properly updated locally
- conversation roles are properly updated based on legacy role

### Checklist

---


WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationAccessUpdateEventProcessor.swift
```diff
@@ -26,15 +27,48 @@ protocol ConversationAccessUpdateEventProcessorProtocol {
...
+
+        let accessRoles = if let legacyAccessRole = event.legacyAccessRole {
+            getAccessRoles(from: legacyAccessRole)
+        } else {
+            event.accessRoles
+        }
+
+        await context.perform {
+            localConversation.accessModeStrings = event.accessModes.map(\.rawValue)
+            localConversation.accessRoleStringsV2 = accessRoles.map(\.rawValue)
```
netbe: context.save() ?

johnxnguyen: Shouldn't this be part of the repository?

johnxnguyen: I think saves should happen on a higher level. For instance:

- Get batch of pending events
- Process batch of events (no saves in-between)
- After successfully processing all events in batch, one single save, then delete those pending events (they are consumed now

We need to take care that any committed work isn't repeated in case of any interruption to the event processing.

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationAccessUpdateEventProcessor.swift
```diff
@@ -26,15 +27,48 @@ protocol ConversationAccessUpdateEventProcessorProtocol {
...
+
+    func processEvent(_ event: ConversationAccessUpdateEvent) async {
+        let conversationID = event.conversationID
+
+        let localConversation = await repository.fetchOrCreateConversation(
+            with: conversationID.uuid,
+            domain: conversationID.domain
+        )
+
+        let accessRoles = if let legacyAccessRole = event.legacyAccessRole {
```
johnxnguyen, line 46: Hmm... do we even need to handle the legacy roles anymore? The newer roles are always present.

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationAccessUpdateEventProcessor.swift
jullianm: added a local store dedicated method for that specific object update in https://github.com/wireapp/wire-ios/pull/2028/commits/289eab5fa5f60942c24c2962717ac97c029e8a5c

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationAccessUpdateEventProcessor.swift
jullianm: it's actually an optional at the event decoder level, we default it to empty set if this is nil. in versioned API `access_role_v2`  is removed in v3 and renamed to `access_role` but since we're dealing with an event here (not versioned then) we assume we receive either one of these two properties. I did some changes to be more explicit in https://github.com/wireapp/wire-ios/pull/2028/commits/75042b6e13efa52d32e7103943e04a74c65a1af6 and https://github.com/wireapp/wire-ios/pull/2028/commits/d00fd5f66021960357311b2b55337b538169fb94

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationAccessUpdateEventProcessor.swift
jullianm: Just sharing my thoughts regarding these topics here:

**1-** Saving on a higher level would imply to delegate the responsibility of persisting CoreData changes to another layer/object (whatever that layer/object is) whereas IMO it should be the sole responsibility of the local stores (which will encapsulate all of our storage operations and should be responsible of actually saving these changes to the persistent store when needed _or when asked for it_). 

--> what we want to avoid is having context saving spread everywhere in codebase (we might end up with funny behaviors if somewhere we forget to save the changes). that's more error prone.

--> we want to be our different layers to be well-isolated and decoupled, for instance I don't think we want our `SyncManager` to access a CoreData context directly but go through an intermediate object for instance a `LocalStoreManager` that owns multiple local stores and can be responsible of handling "global decisions" on the persistent store (e.g `didFinishSlowSync()` `didFinishQuickSync()` ) and saving the context or rolling back latest commits.

**2-** Do we want to wait for all events to be processed to save the changes (I understand we might want to to so for performance reason) but that means if we have 8 events to process and 7 are successfully processed but the last one is not, we will discard all committed changes and redo the work all over again for 8 events even if only one failed. 

Also, even if we consider the events we receive as a batch, they don't seem interdependent to me. So we could consider saving each one of them right after they're being processed, so if we have to process a batch of 4 events, we'll process 1. then 2. then 3. then 4. and if the 2. fails we could still process 3. and 4. since they don't depend on 2.). Then, saving the context after each event is processed could make sense (I think).

Also, we want the UI to  _always_ reflect the latest database "state" so it could be risky to discard a batch of 100 events all at once.

It's true though that it comes at a cost, performance wise this is not ideal.. especially if we have a lot of events to process.

WireAPI/Sources/WireAPI/Models/UpdateEvent/ConversationEvent/ConversationAccessUpdateEvent.swift
```diff
@@ -35,11 +35,22 @@ public struct ConversationAccessUpdateEvent: Equatable, Codable {
...
 
     /// The new legacy access role.
 
     public let legacyAccessRole: ConversationAccessRoleLegacy?
 
+    public init(conversationID: ConversationID,
+                senderID: UserID,
+                accessModes: Set<ConversationAccessMode>,
+                accessRoles: Set<ConversationAccessRole>?,
+                legacyAccessRole: ConversationAccessRoleLegacy?) {
```
johnxnguyen: To remain consistent we should move the parameters on new lines.

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationAccessUpdateEventProcessor.swift
johnxnguyen: ok thanks for clarifying. Perhaps a short comment is needed here

johnxnguyen (approved):

WireAPI/Sources/WireAPI/Models/UpdateEvent/ConversationEvent/ConversationAccessUpdateEvent.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2028/commits/ead3b74b6aa5d7980d97061a124c80f8ac346e25

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationAccessUpdateEventProcessor.swift
```diff
@@ -26,15 +26,49 @@ protocol ConversationAccessUpdateEventProcessorProtocol {
...
+            getAccessRoles(from: legacyAccessRole)
+        } else {
+            event.accessRoles ?? [.teamMember, .nonTeamMember, .service]
+        }
+
+        await localStore.updateAccesses(
```
netbe, line 52: for now I would still add a save within the updateAccess method

netbe (changes requested):
for me it's missing the save of the context for the updateAccess

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationAccessUpdateEventProcessor.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2028/commits/dbcbbbc31b3d2c9fc17491832226eefaaf98e7f6

netbe (approved):

PR #2031, opened 2024-10-15
fix: Track call related events - WPB-11333

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11333" title="WPB-11333" target="_blank">BugWPB-11333</a>  [iOS] "Call initiated" event not being tracked
  </td></table>
  <br />
 

Also fixes: [WPB-11334](https://wearezeta.atlassian.net/browse/WPB-11334)

## Overview

This PR introduces improvements to our call analytics tracking system within the `ZMUserSession` class. It focuses on capturing and reporting key events during call initiation and joining, with specific attention to video call status and conversation types.

## Key Changes

- Added a `CallInfo` struct to store video toggle information for ongoing calls
- Implemented call state tracking in `ZMUserSession` extension conforming to `WireCallCenterCallStateObserver`
- Added methods to track call initialization and joining events
- Introduced a mapping function to convert `ZMConversationType` to `ConversationType` for analytics purposes

## Implementation Details

- `callInfos` dictionary in `ZMUserSession` stores call information indexed by conversation ID
- Call state changes are observed and relevant analytics events are triggered
- Video call status is tracked and reported in analytics events
- Conversation types are mapped to appropriate analytics categories

---

### Checklist

---

[WPB-11334]: https://wearezeta.atlassian.net/browse/WPB-11334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+WireCallStateObserver.swift
```diff
@@ -0,0 +1,72 @@
...
+        caller: any UserType,
+        timestamp: Date?,
+        previousCallState: CallState?
+    ) {
+        guard let conversationId = conversation.remoteIdentifier else { return }
+
+        switch callState {
+        case .outgoing:
+            let isVideo = conversation.voiceChannel?.isVideoCall ?? false
+            callInfos[conversationId] = CallInfo(toggledVideo: isVideo)
```
johnxnguyen: Why do we need to create and store call info?

netbe: @agisilaos I guess you can skip storing on relaying on the callInfos, no?

caldrian (approved):
Code-wise it looks okay. I'm lacking knowledge about the analytics epic and its requirements.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+WireCallStateObserver.swift
agisilaos: Fixed: @netbe @johnxnguyen

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+WireCallStateObserver.swift
```diff
@@ -0,0 +1,68 @@
...
+        caller: any UserType,
+        timestamp: Date?,
+        previousCallState: CallState?
+    ) {
+
+        switch (callState, previousCallState) {
+        case (.outgoing, _):
+            let isVideoCall = isVideoCall(for: conversation)
+            trackCallInitialized(isVideo: isVideoCall, conversationType: conversation.conversationType)
+        case (.answered(_), .incoming(video: let isVideoCall, shouldRing: _, degraded: _)):
```
johnxnguyen: Neither of these call states represent the self user actually being in the call. I think you need `.established`.

wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallQualityController/CallQualityController.swift
```diff
@@ -113,12 +113,12 @@ class CallQualityController: NSObject {
         let callDuration = callEndDate.timeIntervalSince(callStartDate)
 
         guard callDuration >= miminumSignificantCallDuration else {
-            Analytics.shared.tagCallQualityReview(.notDisplayed(reason: .callTooShort, duration: Int(callDuration)))
+            //Analytics.shared.tagCallQualityReview(.notDisplayed(reason: .callTooShort, duration: Int(callDuration)))
```
johnxnguyen: To delete?

johnxnguyen (changes requested):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+WireCallStateObserver.swift
johnxnguyen: I misread the switch statement, I didn't realize you're comparing previous and current call states. This code assumes that `incoming` always precedes `answered`, which might not be guaranteed.

What is the intention of comparing call states? To avoid repeated event tracking for a single call?

wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallQualityController/CallQualityController.swift
agisilaos: Done!

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+WireCallStateObserver.swift
agisilaos: We thought it as a solution with @netbe so we can get the value for `isVideoCall` for the call joined event. Otherwise the value would be set to always false.

johnxnguyen: Can you read of from the voice channel, just like the outgoing case?

netbe: so actually yes you can but it only covers the oneonone. I retested and if we do a group call the isVideo will be false all the time, the info should be available in the call participants though

netbe (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+WireCallStateObserver.swift
agisilaos: After discussing with @netbe we reverted it to the original implementation.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+WireCallStateObserver.swift
```diff
@@ -0,0 +1,72 @@
...
+        previousCallState: CallState?
+    ) {
+        guard let conversationId = conversation.remoteIdentifier else { return }
+
+        let isVideoCall = isVideoCall(for: conversation)
+
+        switch callState {
+        case .outgoing:
+            trackCallInitialized(isVideo: isVideoCall, conversationType: conversation.conversationType)
+        case .answered:
```
johnxnguyen, line 41: If the previous state is the same as the current state, it may be possible that you track events more than once. You could avoid this by adding a guard.

johnxnguyen (approved):

PR #2032, opened 2024-10-15
fix: user is not asked to enable analytics after login - WPB-11332

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11332" title="WPB-11332" target="_blank">BugWPB-11332</a>  [iOS] Analytics user consent is not requested after login
  </td></table>
  <br />
 

### Issue

After logging in, the user should be asked if they wish to enable analytics. This functionality was removed recently (a regression). In this PR, we reintroduce the request, which is now based on the presence of the user preference:

- Previously the preference `disableAnalyticsSharing` was given a default value of `true`. Now we don't set a default, and instead consider a `nil` value as both `true` (analytics is disabled) and that the user hasn't yet made a choice.
- After login, if the user hasn't yet made a choice, then they will be prompted to enable analytics.

### Testing

After a clean install (with an analytics capable build) and login, you should be prompted. Further logins should not trigger a prompt.

---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS Tests/ExtensionSettingsTests.swift
+++ b/wire-ios/Wire-iOS Tests/ExtensionSettingsTests.swift
@@ -38,7 +38,7 @@ final class ExtensionSettingsTests: XCTestCase {
     }
 
     func testThatItHandlesAnalyticsPreferenceChange() {
-        XCTAssertEqual(settings.disableAnalyticsSharing, true)
+        XCTAssertNil(settings.disableAnalyticsSharing)
 
         settings.disableAnalyticsSharing = false
         XCTAssertEqual(settings.disableAnalyticsSharing, false)
--- a/wire-ios/Wire-iOS Tests/ExtensionSettingsTests.swift
+++ b/wire-ios/Wire-iOS Tests/ExtensionSettingsTests.swift
@@ -37,18 +37,14 @@ final class ExtensionSettingsTests: XCTestCase {
         super.tearDown()
     }
 
-    func testThatItDisablesAnalyticsReportByDefault() {
-        XCTAssertTrue(settings.disableAnalyticsSharing)
-    }
-
     func testThatItHandlesAnalyticsPreferenceChange() {
-        XCTAssertTrue(settings.disableAnalyticsSharing)
+        XCTAssertEqual(settings.disableAnalyticsSharing, true)
 
         settings.disableAnalyticsSharing = false
-        XCTAssertFalse(settings.disableAnalyticsSharing)
+        XCTAssertEqual(settings.disableAnalyticsSharing, false)
 
         settings.disableAnalyticsSharing = true
-        XCTAssertTrue(settings.disableAnalyticsSharing)
+        XCTAssertEqual(settings.disableAnalyticsSharing, true)
     }
 
     func testThatItEnablesLinkPreviewsByDefault() {
--- a/wire-ios/Wire-iOS/Sources/AnalytitcsServiceConfigurationBuilder.swift
+++ b/wire-ios/Wire-iOS/Sources/AnalytitcsServiceConfigurationBuilder.swift
@@ -33,7 +33,7 @@ struct AnalyticsServiceConfigurationBuilder {
         return AnalyticsServiceConfiguration(
             secretKey: secretKey,
             serverHost: countlyURL,
-            didUserGiveTrackingConsent: !ExtensionSettings.shared.disableAnalyticsSharing
+            didUserGiveTrackingConsent: !(ExtensionSettings.shared.disableAnalyticsSharing ?? true)
         )
     }
 
--- a/wire-ios/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
+++ b/wire-ios/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
@@ -22,8 +22,12 @@ import WireSyncEngine
 import WireUtilities
 
 protocol TrackingInterface {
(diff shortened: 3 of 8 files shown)


wire-ios/Wire-iOS/Sources/AnalytitcsServiceConfigurationBuilder.swift
```diff
@@ -33,7 +33,7 @@ struct AnalyticsServiceConfigurationBuilder {
         return AnalyticsServiceConfiguration(
             secretKey: secretKey,
             serverHost: countlyURL,
-            didUserGiveTrackingConsent: !ExtensionSettings.shared.disableAnalyticsSharing
+            didUserGiveTrackingConsent: !(ExtensionSettings.shared.disableAnalyticsSharing ?? true)
```
netbe, line 36: maybe we should avoid to rely on the fallback value, for example making the isAnalyticsDisabled static this would be possible

```suggestion
            didUserGiveTrackingConsent: !TrackingManager.isAnalyticsDisabled
```

wire-ios/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
```diff
@@ -236,6 +236,17 @@ final class ZClientViewController: UIViewController {
...
     }
 
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        Task {
+            do {
+                try await trackingManager?.firstTimeRequestToEnableAnalytics()
+            } catch {
+                WireLogger.analytics.error("failed to first time enable analytics: \(error)")
+            }
```
netbe, line 250: suggestion move to own function

netbe (approved):
looks good just added a comment

wire-ios/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
johnxnguyen: done 6c8319105cd2a191a7e9c139b403a72b48082c2a

wire-ios/Wire-iOS/Sources/AnalytitcsServiceConfigurationBuilder.swift
johnxnguyen: I'd like to avoid a static accessor, Swift concurrency won't like it. There's some strange dependency relationships here that we need to solve (but I would defer it for later): TrackingManager and the ExtensionSettings should really go to WireDomain where they can be used more directly.

wire-ios/Wire-iOS Tests/ExtensionSettingsTests.swift
```diff
@@ -37,18 +37,14 @@ final class ExtensionSettingsTests: XCTestCase {
...
-        XCTAssertTrue(settings.disableAnalyticsSharing)
+        XCTAssertNil(settings.disableAnalyticsSharing)
 
         settings.disableAnalyticsSharing = false
-        XCTAssertFalse(settings.disableAnalyticsSharing)
+        XCTAssertEqual(settings.disableAnalyticsSharing, false)
 
         settings.disableAnalyticsSharing = true
-        XCTAssertTrue(settings.disableAnalyticsSharing)
+        XCTAssertEqual(settings.disableAnalyticsSharing, true)
```
samwyndham, line 47: Why did you change this? Just a style change?

wire-ios/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
```diff
@@ -226,20 +230,37 @@ final class SettingsPropertyFactory {
...
 
             let setAction: SetAction = { [unowned self] _, value, resultHandler in
-                if var tracking = self.tracking {
-                    switch value {
-                    case .number(let number):
-                        tracking.disableAnalyticsSharing(isDisabled: number.boolValue, resultHandler: resultHandler)
-                    default:
-                        throw SettingsPropertyError.WrongValue("Incorrect type \(value) for key \(propertyName)")
+                guard let tracking else {
+                    return
```
samwyndham, line 241: Is this a valid state where `tracking` is nil?

samwyndham (approved):
Looks good. Added a couple of questions

wire-ios/Wire-iOS Tests/ExtensionSettingsTests.swift
johnxnguyen: Because it became optional

wire-ios/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
johnxnguyen: It's hard to say. The tracking manager is passed around as optional in a lot of places, my guess it because for testing a nil value is passed in. I would like to replace the settings system with something much much simpler, which is why I didn't worry too much about making this code nice.

wire-ios/Wire-iOS Tests/ExtensionSettingsTests.swift
samwyndham: 🤦 of course!

wire-ios/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
samwyndham: 👍 Optionally consider logging an error if it is not there.

PR #1605, opened 2024-06-20
feat: add search bar to conversation list - WPB-7301

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7301" title="WPB-7301" target="_blank">StoryWPB-7301</a>  [iOS] Search: "Conversations" - Add search bar to top of conversation list
  </td></table>
  <br />
 

### Issue

This PR adds a search bar to the top of the conversation list.
Conversations are filtered by participant's and group conversation's names.

---

### Checklist

- [x] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [x] Description is filled and free of optional paragraphs.
- [x] Adds/updates automated tests.

PR #2075, opened 2024-10-22
chore: conversation filter use case - WPB-7301

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7301" title="WPB-7301" target="_blank">StoryWPB-7301</a>  [iOS] Search: "Conversations" - Add search bar to top of conversation list
  </td></table>
  <br />
 

### Issue

This PR adds a use case and protocols as a preparation to implement conversation searching/filtering in the UI.

---

### Checklist

---


wire-ios-data-model/Source/ConversationSearch/SearchConversationsUseCase.swift
```diff
@@ -0,0 +1,71 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireUtilities
+
+/// This use case takes grouped conversations as input and returns only non-empty groups of conversations
+/// where the provided search text matches.
+///
+/// The search text is compared to the name of conversations as well as the names of the conversation's participants.
+public struct SearchConversationsUseCase<ConversationContainer>: SearchConversationsUseCaseProtocol
```
johnxnguyen: The name `SearchConversationsUseCase` might lead some (including myself) to think it's about searching within conversations (i.e their content). Perhaps a less ambiguous name could be  `FilterConversationsUseCase`.

wire-ios-data-model/Source/ConversationSearch/SearchConversationsUseCase.swift
```diff
@@ -0,0 +1,71 @@
...
+where ConversationContainer: MutableConversationContainer {
+
+    private let conversationContainers: [ConversationContainer]
+
+    public init(conversationContainers: [ConversationContainer]) {
+        self.conversationContainers = conversationContainers
+    }
+
+    public func invoke(searchText: String) -> [ConversationContainer] {
```
johnxnguyen:
```suggestion
    public func invoke(query: String) -> [ConversationContainer] {
```

wire-ios-data-model/Source/ConversationSearch/MutableConversationContainer.swift
```diff
@@ -0,0 +1,27 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+/// A mutable container of conversations used by the `ConversationSearchUseCase`.
+/// This can be interpreted as a section in a table where a row is a conversation.
+public protocol MutableConversationContainer {
```
johnxnguyen, line 21: What do you expect the conforming type to be? Just an array?

wire-ios-data-model/Source/ConversationSearch/SearchConversationsUseCase.swift
caldrian: Thank you, I've changed it.

wire-ios-data-model/Source/ConversationSearch/MutableConversationContainer.swift
caldrian: It can be anything as long as it satisfies the requirements (`conversations` getter and `removeConversation(at:)` function). 

In the navigation overhaul branch the type `ConversationListViewModel.Section` conforms to this protocol.

wire-ios-data-model/Source/ConversationSearch/FilterConversationsUseCase Protocol.swift
```diff
@@ -0,0 +1,25 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+/// A use case which returns a collection of grouped conversations which match the provided search text.
+public protocol FilterConversationsUseCaseProtocol {
+    associatedtype ConversationContainer: MutableConversationContainer
+
+    /// Returns conversations which match the `searchText`.
```
johnxnguyen:
```suggestion
    /// Returns conversations which match the `query`.
```

wire-ios-data-model/Source/ConversationSearch/MutableConversationContainer.swift
johnxnguyen: Could you please clarify why we need this abstraction? I think I'm missing some context.

wire-ios-data-model/Source/ConversationSearch/FilterConversationsUseCase.swift
```diff
@@ -0,0 +1,71 @@
...
+                where (participant.searchableName.normalizedForSearch() as String).lowercased().contains(query) {
+                    continue conversationLoop
+                }
+
+                // no match, remove conversation from results
+                conversationContainers[containerIndex].removeConversation(at: conversationIndex)
+            }
+        }
+
+        return conversationContainers
```
johnxnguyen: At first I found this a bit hard to read, but I as I look at it more it makes sense. I'm just wondering if there's a way we can express it in simpler terms, something like:

```swift
return conversationContainers.filterConversations { conversation in
  if conversation.searchableName.contains(query) {
    return true
  }

  if conversation.searchableParticipants.contains { participant in
    participant.searchableName.contains(query)
  } {
    return true
  }
}
```

wire-ios-data-model/Source/ConversationSearch/FilterConversationsUseCase.swift
```diff
@@ -0,0 +1,71 @@
...
+        // iterate through all containers and all conversations and remove those who don't match the query
+        var conversationContainers = conversationContainers
+        for containerIndex in conversationContainers.indices {
+        conversationLoop:
+            for conversationIndex in conversationContainers[containerIndex].conversations.indices.reversed() {
+
+                let conversation = conversationContainers[containerIndex].conversations[conversationIndex]
+
+                // check if conversation name matches
+                if (conversation.searchableName.normalizedForSearch() as String).lowercased().contains(query) {
```
johnxnguyen: `searchableName` suggests to me that it's already normalized and lowercased and thus not necessary to do it at the call site.

wire-ios-data-model/Source/ConversationSearch/SearchableConversation.swift
```diff
@@ -0,0 +1,25 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+/// A protocol which represents a conversation that can be searched for.
+public protocol SearchableConversation {
+    associatedtype SearchableParticipant: SearchableConversationParticipant
+
+    var searchableName: String { get }
+    var searchableParticipants: [SearchableParticipant] { get }
+}
```
johnxnguyen: Is the purpose to conceal managed objects from the view layer?

caldrian: Yes, to make the search algorithm independent from the CoreData entities.

wire-ios-data-model/Source/ConversationSearch/MutableConversationContainer.swift
caldrian: Take a look into `ConversationListViewModel.Section`. It is used as a conversation-grouping model for the conversation list, but has the `fileprivate` access modifier and is not implemented in a clean way, in my opinion.

So I saw three options:
1. Implementing the filter algorithm (`FilterConversationsUseCase`) in Wire-iOS on top of the UI code using the actual types declared there (`ConversationListViewModel` etc.): I actually don't consider it an option.
2. Abstracting the search algorithm using simple collection types (like Array): Then unfortunately a double mapping is required, first from the existing data structure (`ConversationListViewModel.Section`) to an array and after the filtering, mapping the filtered array back to the custom data structure.
3. Declaring the generic `MutableConversationContainer` which allows for getting the current state and removing (= filtering out) items. This is the cleanest approach in my opinion.

caldrian: This is the original PR, btw:
https://github.com/wireapp/wire-ios/pull/1605

wire-ios-data-model/Source/ConversationSearch/FilterConversationsUseCase.swift
caldrian: Good point, thank you.
Still, I would like the normalization to be a part of the use case. Do you have a suggestion for a better property name?

wire-ios-data-model/Source/ConversationSearch/FilterConversationsUseCase.swift
caldrian: I have tried to rewrite the code but failed to find a more compact or easier to read solution, so I just improved the comments.

wire-ios-data-model/Source/ConversationSearch/FilterConversationsUseCase.swift
caldrian: I now removed the prefix "searchable".

wire-ios-mono.xcworkspace/xcshareddata/swiftpm/Package.resolved
samwyndham, line 1: Were changes in this file intentionally committed?

wire-ios-data-model/Source/ConversationSearch/FilterConversationsUseCase.swift
```diff
@@ -0,0 +1,74 @@
...
+
+        if query.isEmpty {
+            return conversationContainers
+        }
+
+        // Iterate through the grouped conversations and remove the conversations which don't match the query.
+        // Empty containers (conversation groups) will be kept in the result.
+        var containers = conversationContainers
+        for containerIndex in containers.indices {
+        conversationLoop:
```
samwyndham, line 49: Wow I don't think I've seen this API used in practice. Took me a while to work out what was happening!

wire-ios-data-model/Source/ConversationSearch/FilterConversationsUseCase.swift
```diff
@@ -0,0 +1,74 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireUtilities
+
+/// This use case takes grouped conversations as input and returns the same groups with their conversations
+/// filtered by the search text `query`.
+///
+/// The search text is compared to the name of conversations as well as the names of the conversation's participants.
+public struct FilterConversationsUseCase<ConversationContainer>: FilterConversationsUseCaseProtocol
```
samwyndham, line 25: Just sharing a thought - not a request for change... 

Part of me wonders if this should be done at the core data layer (if that is even possible). I guess by doing it here we can have more flexibility.

wire-ios-data-model/Tests/ConversationSearch/FilterConversationsUseCaseTests.swift
```diff
@@ -0,0 +1,205 @@
...
+        XCTAssertEqual(filtered, conversationContainers)
+    }
+
+    func testSingleCharacterSearchTextReturnsNoConversations() {
+
+        // Given
+        sut = .init(conversationContainers: conversationContainers)
+
+        // When
+        let filtered = sut.invoke(query: "y")
```
samwyndham, line 50: I don't think the data contains a "y". If you change this to an "a" I think you test will fail right? Maybe I misunderstand the purpose of this test. I thought it means that any single character search will always return no conversations

samwyndham (approved):

wire-ios-data-model/Source/ConversationSearch/FilterConversationsUseCase.swift
caldrian: Good question, I don't know. I have little experience with search in general. The way of searching is also not flexible, it just calls String.contains.
The other point why I did this isolated was because the code for showing the conversation list is really hard to maintain. I hope we can refactor the whole feature (conversation list including search) eventually.

wire-ios-data-model/Tests/ConversationSearch/FilterConversationsUseCaseTests.swift
caldrian: The name was misleading. The test is about not matching anything.
I changed the name.

wire-ios-data-model/Source/ConversationSearch/FilterConversationsUseCase.swift
caldrian: You mean the label? I need it for continuing the outer loop from within a nested loop.

wire-ios-mono.xcworkspace/xcshareddata/swiftpm/Package.resolved
caldrian: No, shall I create a separate PR for these?

wire-ios-data-model/Source/ConversationSearch/FilterConversationsUseCase.swift
samwyndham: Yes exactly. Nothing wrong with it. I just completely forgot about this feature from swift.

WireUI/Sources/WireAccountImageUI/AccountImageGenerator/AccountImageGenerator.swift
```diff
@@ -61,6 +61,7 @@ private final class InitialsRenderer: UIView {
 
     private func setupLabel(_ initials: String) {
         initialsLabel.font = .systemFont(ofSize: 13, weight: .light)
+        initialsLabel.textColor = .black
```
agisilaos: Setting it to black means when switching to dark mode the label's color would still be black. Is this actually the desired behaviour?

agisilaos (approved):

wire-ios-data-model/Source/ConversationSearch/MutableConversationContainer.swift
johnxnguyen: It seems that without refactoring also the view layer here we are limited in nice solutions and I agree that the one you have chosen is the cleanest option at the moment.

Ideally however I would imagine that the view model requests a simple grouped list of conversations via a use case, passing in an optional filter. I think the filter implementation shouldn't be constrained by the implementation of the view, but rather the view needs to work with a well designed interface provided to it.

For instance:

```swift
let conversationsByCategory: [ConversationCategory: [Conversation]] = usecase.invoke(filter: .query("foo"))
```

johnxnguyen (approved):

wire-ios-data-model/Source/ConversationSearch/MutableConversationContainer.swift
caldrian: Yes, this would be nice.
I vote for rewriting the conversation list in SwiftUI eventually.

WireUI/Sources/WireAccountImageUI/AccountImageGenerator/AccountImageGenerator.swift
caldrian: Good point, I will revert this and ask in the design team.

PR #2084, opened 2024-10-23
refactor: renew access token if needed - WPB-10448

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10448" title="WPB-10448" target="_blank">TaskWPB-10448</a>  [iOS] Renew access token if needed
  </td></table>
  <br />
 

### Issue

This PR includes changes to renew an access token within the `APIService` if one is needed and there is no existing token, or the existing token is expiring.

-  I've deleted `AuthenticationStorage` and related types, replacing it with `AuthenticationManager` and `CookieStorage`.
- `AuthenticationManager` is responsible for providing a valid access token.
- It does this by making a request to the backend and caching the result.
- It employs a non-reentrancy mechanism to avoid making more than one request to the backend. This means that if 100 tasks each request a valid access token but the current token is expiring, only one request is made to renew the token and all 100 tasks will wait for that token to return.
- `APIService` will attempt to make an authenticated request, and in case there is an authentication error, it will try only once more with a fresh token.

### Testing

- Unit test `APIService` for single retry.
- Move unit tests from `PersistentAuthenticationStorageTests` to `CookieStorageTests`
- Unit test `AuthenticationManager` for renewing and reentrancy logic

---

### Checklist

---


WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
```diff
@@ -99,14 +105,82 @@ public final class APIService: APIServiceProtocol {
         var request = request
 
         if requiringAccessToken {
-            guard let accessToken = authenticationStorage.fetchAccessToken() else {
-                throw APIServiceError.missingAccessToken
-            }
-
+            let accessToken = try await getAccessToken()
```
johnxnguyen: Existing behavior is that when we fail to get an access token because of an invalid cookie, we clear the cookie from the keychain which will log out the user. I think I need to add this behavior here, but perhaps in a separate PR.

WireAPI/Tests/WireAPITests/APIs/ConversationsAPI/ConversationsAPITests.swift
```diff
@@ -527,7 +527,7 @@ final class ConversationsAPITests: XCTestCase {
             for sut in suts {
                 taskGroup.addTask {
                     // Then
-                    await self.XCTAssertThrowsError(ConversationsAPIError.unsupportedEndpointForAPIVersion) {
+                    await self.XCTAssertThrowsErrorAsync(ConversationsAPIError.unsupportedEndpointForAPIVersion) {
```
johnxnguyen, line 542: The complier was having a hard time distinguishing this async variant (we define it WireTestingPackage) and the existing non-async variant. Hence, I just gave it a different name.

WireAPI/Tests/WireAPITests/Network/APIService/__Snapshots__/APIServiceTests/testExecuteRequest_Requiring_Access_Token_But_Existing_Token_Is_Expiring.1.txt
```diff
@@ -0,0 +1,5 @@
+curl \
+	--request POST \
+	--header "Accept: application/json" \
+	--header "Authorization: Bearer an-expiring-access-token" \
```
johnxnguyen: Nice that now we can see the access token in the snapshot.

wire-ios-data-model/Tests/Authentication/EAR/EARServiceTests.swift
```diff
@@ -17,6 +17,7 @@
 //
 
 import LocalAuthentication
+import WireTestingPackage
```
johnxnguyen: This is because I moved the `UserDefaults.temporary()`  from `WireSystem` to `WireTestingPackage`.

wire-ios-testing/Source/Public/ZMTBaseTest.m
```diff
@@ -16,7 +16,8 @@
 // along with this program. If not, see http://www.gnu.org/licenses/.
 //
 
-@import WireSystem;
+//@import WireSystem;
```
johnxnguyen:
```suggestion
```

WireAPI/Sources/WireAPI/Authentication/HTTPCodec.swift
```diff
@@ -0,0 +1,60 @@
...
+        archiver.encode(properties, forKey: "properties")
+        archiver.finishEncoding()
+
+        return archiver.encodedData
+    }
+
+    static func decodeData(_ data: Data) throws -> [HTTPCookie] {
+        let unarchiver: NSKeyedUnarchiver
+        do {
+            unarchiver = try NSKeyedUnarchiver(forReadingFrom: data)
```
jullianm, line 43: I think we can use a more "Swifty" API here, using the `PropertyListDecoder`, similar to decoding API objects.

```
let decoder = PropertyListDecoder()
let model = try decoder.decode(HTTPCookie.self, from: data) 

struct HTTPCookie: Decodable {
    ///
}
```

WireAPI/Sources/WireAPI/Authentication/InMemoryAuthenticationStorage.swift
```diff
@@ -31,12 +31,12 @@ final class InMemoryAuthenticationStorage: AuthenticationStorage {
         accessToken
     }
 
-    func storeCookieData(_ cookieData: Data?) {
-        self.cookieData = cookieData
+    func storeCookies(_ cookies: [HTTPCookie]) async throws {
```
jullianm: is async throws needed ?

WireAPI/Sources/WireAPI/Authentication/InMemoryAuthenticationStorage.swift
```diff
@@ -31,12 +31,12 @@ final class InMemoryAuthenticationStorage: AuthenticationStorage {
...
+        self.cookies = cookies
     }
 
-    func fetchCookieData() -> Data? {
-        cookieData
+    func fetchCookies() async throws -> [HTTPCookie] {
```
jullianm: async throws not needed here, no ?

WireAPI/Sources/WireAPI/Authentication/PersistentAuthenticationStorage.swift
```diff
@@ -0,0 +1,192 @@
...
+    ) {
+        self.userID = userID
+        self.sharedUserDefaults = sharedUserDefaults
+        self.keychain = keychain
+    }
+
+    // MARK: - Access token
+
+    public func storeAccessToken(_ accessToken: AccessToken) async {
+        self.accessToken = accessToken
```
jullianm: I guess we need to mark this method async because it is required by the actor, right ?

WireAPI/Sources/WireAPI/Authentication/PersistentAuthenticationStorage.swift
```diff
@@ -0,0 +1,192 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireFoundation
+
+public actor PersistentAuthenticationStorage: AuthenticationStorage {
```
jullianm: Nice!

WireAPI/Sources/WireAPI/Authentication/PersistentAuthenticationStorage.swift
```diff
@@ -0,0 +1,192 @@
...
+        kSecAttrService: "Wire: Credentials for wire.com",
+        kSecAttrAccount: userID.uuidString,
+        kSecClass: kSecClassGenericPassword
+    ]
+
+    private lazy var fetchQuery: [CFString: Any] = {
+        var result = baseQuery
+        result[kSecReturnData] = true
+        return result
+    }()
```
jullianm: these lazy var can be moved to the top

WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
```diff
@@ -99,14 +105,82 @@ public final class APIService: APIServiceProtocol {
...
+    let expiresIn: Int
+
+    enum CodingKeys: String, CodingKey {
+
+        case user
+        case accessToken = "access_token"
+        case tokenType = "token_type"
+        case expiresIn = "expires_in"
+
+    }
```
jullianm: we can remove the `CodingKeys` code by injecting a decoder with a `decoder.keyDecodingStrategy = .convertFromSnakeCase` into the ResponseParser init.

WireFoundation/Sources/WireFoundation/SDKAbstractions/Keychain.swift
```diff
@@ -0,0 +1,78 @@
...
+
+    public func fetchItem(
+        query: [CFString: Any],
+        result: UnsafeMutablePointer<CFTypeRef?>?
+    ) -> OSStatus {
+        SecItemCopyMatching(
+            query as CFDictionary,
+            result
+        )
+    }
```
jullianm, line 87: Not a request for a change in this PR but maybe we could later introduce a small layer of abstraction so the client (`PersistentAuthenticationStorage`) does not have to deal with keychain logic and types too directly, for instance when fetching an item.

```
enum Error: Swift.Error {
    case statusFailure(OSStatus)
    case failedToCastItem
}

public func fetchItem<T>(
    query: [CFString: Any]
) -> Result<T, Error> {
    
    var result: CFTypeRef?

    let status = SecItemCopyMatching(
          query as CFDictionary,
          result
     )

     switch status {
     case errSecSuccess:
         guard let value = result as? T else {
             return .failure(.failedToCastItem)
         }

         return .success(value)

     default:
        return .failure(.statusFailure(status))
    }

}
```

jullianm (commented):
Nice work, left a few comments

WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
```diff
@@ -99,14 +105,82 @@ public final class APIService: APIServiceProtocol {
...
+            request.setAccessToken(lastKnownAccessToken)
+        }
+
+        let (data, response) = try await networkService.executeRequest(request)
+
+        return try ResponseParser()
+            .success(code: .ok, type: AccessTokenPayload.self)
+            .failure(code: .forbidden, label: "invalid-credentials", error: APIServiceError.invalidCredentials)
+            .parse(code: response.statusCode, data: data)
+    }
```
samwyndham: I think we need to be really careful here to handle race conditions. I _assume_ each time we get a new access token a previous token is invalidated and as far as I know these methods can be called many times in a short period of time. I feel we need some mechanism of queuing or something similar

samwyndham (commented):
Just leaving a quick comment as I have limited time today. I will take a closer look tomorrow.

WireAPI/Sources/WireAPI/Network/URLRequestBuilder/URLRequestBuilder.swift
```diff
@@ -38,6 +38,26 @@ struct URLRequestBuilder {
...
+        name: String,
+        value: String
+    ) -> Self {
+        withCopy {
+            let queryItem = URLQueryItem(
+                name: name,
+                value: value
+            )
+
+            if #available(iOS 16, *) {
```
samwyndham: Now that we dropped iOS 15 this check shouldn't be necessary

WireFoundation/Sources/WireFoundation/SDKAbstractions/Keychain.swift
samwyndham: I also think it would be nice to get rid of the `C` types in any additional abstraction we create.

WireAPI/Tests/WireAPITests/Authentication/PersistentAuthenticationStorageTests.swift
```diff
@@ -0,0 +1,312 @@
...
+        )
+    }
+
+    override func tearDown() {
+        sut = nil
+        userDefaults = nil
+        keychain = nil
+    }
+
+    // MARK: - Helpers
```
samwyndham: Not a request for change. 

Personally I would put helpers at the end. Something else to decide for our conventions.

WireAPI/Sources/WireAPI/Authentication/PersistentAuthenticationStorage.swift
samwyndham: My assumption is that it is like that that to implement the `AuthenticationStorage` protocol. However, _I think_ it is not necessary and it will still conform to the protocol if `async` is removed.

WireAPI/Sources/WireAPI/Authentication/PersistentAuthenticationStorage.swift
```diff
@@ -0,0 +1,192 @@
...
+        self.keychain = keychain
+    }
+
+    // MARK: - Access token
+
+    public func storeAccessToken(_ accessToken: AccessToken) async {
+        self.accessToken = accessToken
+    }
+
+    public func fetchAccessToken() async -> AccessToken? {
```
samwyndham: Like @jullianm implied please check that these `async` is needed here an in other places. I was under the impression that an actor can conform to the async protocol definitions without it.

WireAPI/Sources/WireAPI/Authentication/PersistentAuthenticationStorage.swift
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    // MARK: - Cookie
+
+    public func storeCookies(_ cookies: [HTTPCookie]) async throws {
+        let cookieData = try HTTPCookieCodec.encodeCookies(cookies)
+        try await storeCookieData(cookieData)
+    }
+
+    public func fetchCookies() async throws -> [HTTPCookie] {
```
samwyndham: If the app is deleted and reinstalled and there is old cookie data from the previous install which **hopefully** cannot be decrypted, does this method throw an error?

If yes, perhaps we should consider if this is really an error. Alternatively it should be documented that this kind of error maybe expected as a doc comment so the consumer knows to relax.

WireAPI/Sources/WireAPI/Authentication/HTTPCodec.swift
johnxnguyen: Interesting, I'll look into it. My goal here was to keep the existing implementation since we already have encoded cookie data in the keychain that we need to decode.

WireAPI/Sources/WireAPI/Authentication/InMemoryAuthenticationStorage.swift
johnxnguyen: not here, but it's required by the protocol.

WireAPI/Sources/WireAPI/Authentication/InMemoryAuthenticationStorage.swift
johnxnguyen: not here, but it's required by the protocol.

WireAPI/Sources/WireAPI/Authentication/PersistentAuthenticationStorage.swift
johnxnguyen: It must be async because of the actor, and as a result the protocol method must be marked as async too otherwise there won't be conformance.

WireAPI/Sources/WireAPI/Authentication/PersistentAuthenticationStorage.swift
johnxnguyen: I don't know, I feel like they should stay closer to the code it's most relevant to. What do you think @samwyndham ?

WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
johnxnguyen: The backend doesn't guarantee that they consistently a single naming convention which is why we haven't used the key decoding strategy before. Do you think it's fine to use it here and there at the cost of being inconsistent? @samwyndham

WireFoundation/Sources/WireFoundation/SDKAbstractions/Keychain.swift
johnxnguyen: Yep I agree it would be nice, but indeed for another PR.

WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
johnxnguyen: What's a concrete scenario?

- Execute request A with token A, sent
- Execute request B with token A, A is expiring, request new token B, invalidating A.
- Request B reaches server, authentication error response is returned.
- Request B is sent with token B, succeeds

So the possibility is that one or more requests in flight will fail because a new token was created before they reached the server?

johnxnguyen: And the queue mechanism you're referring to would be to prevent reentry?

WireAPI/Sources/WireAPI/Authentication/PersistentAuthenticationStorage.swift
johnxnguyen: If I remove `async` it won't compile because of actor isolation. And if the protocol methods are not async but he actor methods are, then the actor won't conform to the protocol.

WireAPI/Sources/WireAPI/Authentication/PersistentAuthenticationStorage.swift
johnxnguyen: Hmm.. actually I think it with the AES256 cryptor it won't fail if you pass in a mismatching cipher text and decryption key, it just "decrypts" it to some junk. However, that would then lead to a decoding error when we try to convert `Data` to `HTTPCookie`.

I'm not sure though that we can be able to determine if such a decoding error is because of an old cookie from a previous installation. If it were to happen, then I think the appropriate response is to try get a new cookie.

Indeed it's worth noting in a doc comment that the keychain might contain old cookie data from a previous installations.

johnxnguyen: done 65b4f58aa1a23e8f1aa3142d6981f4fac153a204

WireAPI/Tests/WireAPITests/Authentication/PersistentAuthenticationStorageTests.swift
johnxnguyen: done 6642aeee56a6d64d891fc9f05df044b50db6fbb2

WireAPI/Sources/WireAPI/Network/URLRequestBuilder/URLRequestBuilder.swift
johnxnguyen: done 04a5b95490d9850bb4f15790b497da8a49881eb5

WireAPI/Sources/WireAPI/Authentication/PersistentAuthenticationStorage.swift
samwyndham: @johnxnguyen Have you actually tried this. For me the following compiles which makes me think the actor is conforming to the protocol despite missing the `async` call:

```swift
protocol Animal {
    func makeNoise() async -> String
}

actor Dog: Animal {
    func makeNoise() -> String {
        "Woof!"
    }
}
```

WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
samwyndham: > So the possibility is that one or more requests in flight will fail because a new token was created before they reached the server?

Yes exactly. And my worry is mainly how will the app behave in this situation, especially in a situation where the app makes many requests in a short period of time:

- No/flaky internet
- 5 requests are created
- Internet returns
- Token is expiring
- 5 requests for new tokens are made each succeeding but only the last token remains valid.
- After each token request succeeds it's original request is sent
- Perhaps 4/5 requests fail.

I assume if a request fails for token expiry a new request will be made for a new token. So perhaps we can get into some funny loop between as few a two requests where each keeps requesting a new token and expiring the others token.

I expect that usually things will resolve themselves but I think this situation is hard to reason about. For me it would be much simpler to have some kind of queuing where only one access token refresh is happening at a time. At least this will add simplicity to the users of `getAccessToken()` at the expense of complexity in its implementation.

> And the queue mechanism you're referring to would be to prevent reentry?

Essentially yes. 

~Perhaps a simple way to kind of achieve this is to use a specific session for these requests and limit that session to 1 concurrent request.~ Actually this won't help

samwyndham: If we are to address this situation it would probably make sense to extract this code into a new object.

samwyndham: ~Perhaps a simple way to kind of achieve this is to use a specific session for these requests and limit that session to 1 concurrent request.~

No this won't help

samwyndham: In a greenfield project what I imagine is that we have a specific object for this probably an `actor` that takes care of providing an non expired access token - it can both fetch from storage or update from network. Internally it would probably makes use of a DispatchQueue as I'm not sure how to achieve this async/await

```swift
actor AccessTokenRepository {

    /// Returns a cached access token if valid. If expired, updates the cached token from the server.
    ///
    /// Blocks if a token is currently being refreshed.
    func accessToken() async throws -> String {}
}
```



```

File: WireAPI/Sources/WireAPI/Authentication/InMemoryAuthenticationStorage.swift (hunk 7 above)
Comment (jullianm) [reply WireAPI/Sources/WireAPI/Authentication/InMemoryAuthenticationStorage.swift]: Ok I didn't see that `PersistentAuthenticationStorage` was also using it. 👍

File: WireAPI/Sources/WireAPI/Authentication/PersistentAuthenticationStorage.swift (hunk 11 above)
Comment (samwyndham) [reply WireAPI/Sources/WireAPI/Authentication/PersistentAuthenticationStorage.swift]: I don't really mind either way. I would say it is probably something we should align on and document in the conventions document. Until that is done I don't think it is necessary to make changes.

File: WireAPI/Sources/WireAPI/Network/APIService/APIService.swift (hunk 12 above)
Comment (samwyndham) [reply WireAPI/Sources/WireAPI/Network/APIService/APIService.swift]: I would say if individual endpoints follow a consistent convention then I would favor using a key decoding strategy in those cases over manually defining a key. But I think for each endpoint it should remain consistent... e.g. it is better not to mix a key decoding strategy and custom key names in a single endpoint.

File: WireAPI/Sources/WireAPI/Authentication/AuthenticationManager.swift [hunk 19]
```diff
@@ -0,0 +1,156 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireFoundation
+
+// sourcery: AutoMockable
+protocol AuthenticationManagerProtocol {
+
+    func getValidAccessToken() async throws -> AccessToken
```
samwyndham, line 25: I would add doc comments to clarify the difference between these two methods

WireAPI/Sources/WireAPI/Authentication/AuthenticationManager.swift
```diff
@@ -0,0 +1,156 @@
...
+        self.cookieStorage = cookieStorage
+        self.networkService = networkService
+    }
+
+    func getValidAccessToken() async throws -> AccessToken {
+        switch currentToken {
+        case .renewing(let task):
+            return try await task.value
+
+        case .cached(let accessToken) where !accessToken.isExpiring:
```
samwyndham: OK so the access token is only kept in memory? Is that correct?

WireAPI/Sources/WireAPI/Authentication/AuthenticationManager.swift
```diff
@@ -0,0 +1,156 @@
...
+            }
+
+            let (data, response) = try await networkService.executeRequest(request)
+
+            let accessToken = try ResponseParser()
+                .success(code: .ok, type: AccessTokenPayload.self)
+                .failure(code: .forbidden, label: "invalid-credentials", error: APIServiceError.invalidCredentials)
+                .parse(code: response.statusCode, data: data)
+
+            currentToken = .cached(accessToken)
```
samwyndham: For symmetry I think this should be set by the caller (refreshAccessToken()). And if this method throws current token can be set to `nil` which I expect is the correct behavior.

As it stands I think there is a bug:

If this exits due to failure, then we will be in the renewing case and any future call to getValidAccessToken() will await indefinitely (I think).

samwyndham (commented):
I like the approach! There maybe a bug. Please see my comments

WireAPI/Sources/WireAPI/Authentication/AuthenticationManager.swift
johnxnguyen: Yep will do

WireAPI/Sources/WireAPI/Authentication/AuthenticationManager.swift
johnxnguyen: Correct

WireAPI/Sources/WireAPI/Authentication/AuthenticationManager.swift
johnxnguyen: ok, I think I fixed this in 6295319907b5548a8d8526676f84e19ea2d0cdf1

WireAPI/Sources/WireAPI/Authentication/AuthenticationManager.swift
johnxnguyen: done 6295319907b5548a8d8526676f84e19ea2d0cdf1

WireAPI/Tests/WireAPITests/Authentication/CookieStorageTests.swift
```diff
@@ -0,0 +1,288 @@
...
+//
+
+import WireFoundation
+import WireTestingPackage
+import XCTest
+
+@testable import WireAPI
+@testable import WireFoundationSupport
+
+final class CookieStorageTests: XCTestCase {
```
johnxnguyen, line 26: This was moved from the now deleted `PersistentAuthenticationStorageTests`.

WireAPI/Sources/WireAPI/Authentication/CookieStorage.swift
```diff
@@ -0,0 +1,171 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireFoundation
+
+// sourcery: AutoMockable
+protocol CookieStorageProtocol: Sendable {
```
johnxnguyen, line 23: All of this was basically taken from `PersistentAuthenticationStorage`

WireFoundation/Sources/WireFoundation/SDKAbstractions/Keychain.swift
johnxnguyen: In the end I needed to do this in this PR because the `CookieStorage` is an actor and needs to be sendable, and for that we need sendable arguments which Keychain API doesn't have.

WireFoundation/Sources/WireFoundationSupport/Sourcery/AutoMockable.manual.swift
```diff
@@ -0,0 +1,146 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+public import WireFoundation
+
+public actor MockKeychainProtocol: KeychainProtocol {
```
johnxnguyen, line 22: sorcery can't mock actors yet, so I had to mock this manually 😢

WireAPI/Sources/WireAPI/Authentication/AuthenticationManager.swift
```diff
@@ -29,56 +29,92 @@ protocol AuthenticationManagerProtocol {
...
+    ///
+    /// - Returns: A valid (non-expired) access token.
 
     func getValidAccessToken() async throws -> AccessToken {
         switch currentToken {
         case .renewing(let task):
+            // A new token will come soon, wait
             return try await task.value
 
         case .cached(let accessToken) where !accessToken.isExpiring:
```
samwyndham: I don't think the `.isExpiring` case is tested right? I think it is also worth testing

WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
johnxnguyen: ok with the latest changes I think I've addressed this.

samwyndham: Looking good!

samwyndham (approved):
Nice work!

WireAPI/Sources/WireAPI/Authentication/AuthenticationManager.swift
johnxnguyen: yes it's tested in `testGetValidAccessToken_CacheHitButItIsExpiring`

WireAPI/Sources/WireAPI/Authentication/PersistentAuthenticationStorage.swift
johnxnguyen: moved to the top

WireAPI/Sources/WireAPI/Network/APIService/APIService.swift
johnxnguyen: done in 66142d9ee6949189756190fe034c187e3c5889ff

WireAPI/Tests/WireAPITests/Authentication/AuthenticationManagerTests.swift
```diff
@@ -0,0 +1,225 @@
...
+
+        // When we get a valid token the first time
+        let accessToken = try await sut.getValidAccessToken()
+
+        // Then a request was made to get a new access token.
+        try XCTAssertCount(receivedRequests, count: 1)
+        let snapshotter = HTTPRequestSnapshotHelper()
+        await snapshotter.verifyRequest(request: receivedRequests[0])
+
+        // Then we got back a vaild access token.
```
jullianm:
```suggestion
        // Then we got back a valid access token.
```

WireAPI/Tests/WireAPITests/Authentication/AuthenticationManagerTests.swift
```diff
@@ -0,0 +1,225 @@
...
+        var receivedRequests = [URLRequest]()
+        URLProtocolMock.mockHandler = {
+            receivedRequests.append($0)
+            return try $0.mockResponse(
+                statusCode: .ok,
+                jsonResourceName: "PostAccessSuccessResponse200"
+            )
+        }
+
+        // When mutliple tasks all want an access token.
```
jullianm:
```suggestion
        // When multiple tasks all want an access token.
```

WireAPI/Tests/WireAPITests/Authentication/AuthenticationManagerTests.swift
```diff
@@ -0,0 +1,225 @@
...
+        await withThrowingTaskGroup(of: Void.self) { group in
+            for _ in 1 ... 10 {
+                group.addTask { [sut] in
+                    guard let sut else { return }
+                    let accessToken = try await sut.getValidAccessToken()
+
+                    // Then each task go back a vaild access token.
```
jullianm:
```suggestion
                    // Then each task go back a valid access token.
```

WireAPI/Tests/WireAPITests/Network/APIService/APIServiceTests.swift
```diff
@@ -26,30 +26,30 @@ final class APIServiceTests: XCTestCase {
 
     var sut: APIService!
     var backendURL: URL!
-    var authenticationStorage: InMemoryAuthenticationStorage!
+    var authenticationManager: MockAuthenticationManagerProtocol!
 
     override func setUp() async throws {
         try await super.setUp()
```
jullianm:
```suggestion
```

WireAPI/Tests/WireAPITests/Network/APIService/APIServiceTests.swift
```diff
@@ -26,30 +26,30 @@ final class APIServiceTests: XCTestCase {
...
+            authenticationManager: authenticationManager
         )
     }
 
     override func tearDown() async throws {
         backendURL = nil
-        authenticationStorage = nil
+        authenticationManager = nil
         sut = nil
         try await super.tearDown()
```
jullianm:
```suggestion
```

jullianm (approved):
Nice work!

PR #2089, opened 2024-10-24
fix: Add Clear Button to Group Name Input Field While Editing - WPB-11822

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11822" title="WPB-11822" target="_blank">BugWPB-11822</a>  [iOS] While typing the group name 'x' button(clear text) is missing in the input text field 
  </td></table>
  <br />
 

## Description 

This PR addresses the issue where the clear (X) button was missing in the group name input field while typing. Additionally, it enhances the user experience by introducing a custom clear button with padding for better spacing and a cleaner look.

## Changes:

- **Custom Clear Button**: Replaced the default system-provided clear button with a custom UIButton using the “Clear” icon. The icon has been styled using SemanticColors.Icon.ForegroundDefaultBlack to maintain consistency with the app’s design.
- **Clear Button Padding**: Added padding to the custom clear button to improve the UI by creating space between the button and the edge of the text field.
- **TextField Events Using UIAction**: Updated the text field event handling to use UIAction instead of the #selector pattern for more modern and concise event handling.
- **Clear Button Visibility Logic**: Implemented a function (updateClearButtonVisibility) to control when the clear button should appear, based on whether the text field is empty or contains text.

## Design

Screenshot 2024-10-24 at 11 11 57

### Checklist

---


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/WireTextField.swift
```diff
@@ -96,18 +97,58 @@ class WireTextField: UITextField {
...
+
+        // Create a container view for the button with padding
+        // 16 (button) + 20 (padding)
+        let containerView = UIView(frame: CGRect(x: 0, y: 0, width: 36, height: 16))
+        clearButton.frame = CGRect(x: 0, y: 0, width: 16, height: 16)
+        containerView.addSubview(clearButton)
+
+        self.rightView = containerView
+        // Start with clear button hidden
+        self.rightViewMode = .never
```
samwyndham, line 129: Can't this be `.always` - the [docs](https://developer.apple.com/documentation/uikit/uitextfield/viewmode/always) say that it will display when the textview has text which I expect is what we want.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/WireTextField.swift
```diff
@@ -96,18 +97,58 @@ class WireTextField: UITextField {
...
+        }), for: .editingDidEnd)
+
+        addAction(UIAction(handler: { [weak self] _ in
+            self?.updateClearButtonVisibility()
+        }), for: .editingChanged)
+    }
+
+    // MARK: - Custom Clear Button
+
+    private func setupCustomClearButton() {
```
samwyndham, line 115: Personally I'm slightly against customizing this as it is more code that we have to maintain. Just my subjective thoughts on the topic.

samwyndham (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/WireTextField.swift
```diff
@@ -96,18 +97,58 @@ class WireTextField: UITextField {
...
+        clearButton.tintColor = SemanticColors.Icon.foregroundDefaultBlack
+
+        let clearAction = UIAction { [weak self] _ in
+            self?.text = ""
+            self?.sendActions(for: .editingChanged)
+        }
+        clearButton.addAction(clearAction, for: .touchUpInside)
+
+        // Create a container view for the button with padding
+        // 16 (button) + 20 (padding)
```
samwyndham: I **think** the expected way of setting this padding may be overriding [rightViewRect(forBounds:)](https://developer.apple.com/documentation/uikit/uitextfield/1619638-rightviewrect). However I've never tried this API myself

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/WireTextField.swift
agisilaos: if we set it to always it's going to show before we even type some text that's why I set it to never initially.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/WireTextField.swift
agisilaos: 100% agree the problem here is that Apple's x button isn't accessible in terms of colors for light and dark mode and that's why we need to use a custom one.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/WireTextField.swift
agisilaos: Agree. I prefer using [rightViewRect(forBounds:)](https://developer.apple.com/documentation/uikit/uitextfield/1619638-rightviewrect) instead. Done: https://github.com/wireapp/wire-ios/commit/acd3db02f0714f945a6ea436d6d0f78563fa9316

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/WireTextField.swift
samwyndham: But that is not correct according to the docs. Are you sure?

> The overlay view is always displayed `if` the text field contains text.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/WireTextField.swift
samwyndham: OK good reason!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/WireTextField.swift
```diff
@@ -96,18 +97,60 @@ class WireTextField: UITextField {
...
+        addAction(UIAction(handler: { [weak self] _ in
+            self?.updateClearButtonVisibility()
+        }), for: .editingChanged)
+    }
+
+    // MARK: - Custom Clear Button
+
+    private func setupCustomClearButton() {
+        let clearButton = UIButton(type: .custom)
+        let clearImage = UIImage(named: "Clear")?.withRenderingMode(.alwaysTemplate)
```
caldrian: Can't we use SF Symbols? I thought the design team actually wants us to use SF Symbols.
`x.circle.fill`

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/WireTextField.swift
```diff
@@ -67,6 +67,7 @@ class WireTextField: UITextField {
         setupPadding()
         setupTextFieldProperties()
         setupTextFieldEvents()
+        setupCustomClearButton()
```
caldrian, line 70: Are you sure we want custom behavior?
Didn't the designer say that we want native behavior and in doubt the mocks are wrong?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/WireTextField.swift
```diff
@@ -96,18 +97,60 @@ class WireTextField: UITextField {
     }
 
     private func setupTextFieldEvents() {
-        addTarget(self, action: #selector(textFieldDidStartEditing), for: .editingDidBegin)
-        addTarget(self, action: #selector(textFieldDidFinishEditing), for: .editingDidEnd)
```
caldrian, line 100: question: why do we replace the target/action method by action closures?

caldrian (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/WireTextField.swift
agisilaos: We're trying to get rid of the #selector pattern, that's why we're removing it and replacing it with more modern APIs

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/WireTextField.swift
agisilaos: The native clear button from Apple breaks accessibility in terms of color and contrast. That's why we needed to introduce a custom one.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/WireTextField.swift
agisilaos: `x.circle.fill` is not the same as the designs. In this case I imported the SVG. I need to get some clarification, on this. For now I will leave it as is and I will come back to you with a resolution on that.

PR #2091, opened 2024-10-24
refactor: process conversation member update event - WPB-10170

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10170" title="WPB-10170" target="_blank">TaskWPB-10170</a>  [iOS] Process ConversationMemberUpdateEvent
  </td></table>
  <br />
 

… UTs - WPB-10170

### Key points

Specifically, this PR is about porting the existing implementation of the `ConversationMemberUpdate` event.

### Testing

- event is correctly processed with correct method invocations
- participant is correctly added/removed from conversation
- isSelfUser returns the correct flag

### Checklist

---


WireDomain/Sources/WireDomain/OneOnOneResolver.swift
```diff
@@ -121,7 +121,7 @@ struct OneOnOneResolver: OneOnOneResolverProtocol {
 
         /// Sync the user MLS conversation from backend.
         let mlsGroupID = try await conversationsRepository.pullMLSOneToOneConversation(
-            userID: userID.uuid.uuidString,
+            for: userID.uuid.uuidString,
```
johnxnguyen: Isn't `userID` better in this case since the argument is `String`?

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationRepository.swift
```diff
@@ -37,10 +37,21 @@ public protocol ConversationRepositoryProtocol {
...
     ) async throws -> String
 
+    /// Fetches or creates a conversation locally.
+    /// - Parameters:
+    ///     - id: The conversation ID.
+    ///     - domain: The conversation domain if any.
+    /// - Returns: The `ZMConversation` found or created locally.
+
+    func fetchOrCreateConversation(
+        with id: UUID,
```
johnxnguyen: I think `id` here is perhaps more descriptive from the outside. From https://www.swift.org/documentation/api-design-guidelines/#naming:

> Compensate for weak type information to clarify a parameter’s role. Especially when a parameter type is NSObject, Any, AnyObject, or a fundamental type such as Int or String, type information and context at the point of use may not fully convey intent. In this example, the declaration may be clear, but the use site is vague.

```swift
grid.add(self, for: graphics) // vague
grid.addObserver(self, forKeyPath: graphics) // clear
```

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationMemberUpdateEventProcessor.swift
```diff
@@ -32,9 +33,48 @@ protocol ConversationMemberUpdateEventProcessorProtocol {
...
+            return
+        }
+
+        await conversationRepository.addParticipantToConversation(
+            conversationID: conversationID.uuid,
+            conversationDomain: conversationID.domain,
+            participantID: senderID.uuid,
+            participantDomain: senderID.domain,
+            participantRole: role
+        )
```
johnxnguyen: So the update for other users is only their role? This method suggests that they are new members (I would expect then the participant add event) that need adding, but I'm feel like we should be simply updating their existing role with a new one.

netbe (approved):
Looks good, nothing more to add than John's comments

WireDomain/Sources/WireDomain/OneOnOneResolver.swift
jullianm: fixed in https://github.com/wireapp/wire-ios/pull/2091/commits/cbebb8686ef60bc47af686531cbc8732218095fe thx

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationRepository.swift
jullianm: fixed in https://github.com/wireapp/wire-ios/pull/2091/commits/cbebb8686ef60bc47af686531cbc8732218095fe

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationMemberUpdateEventProcessor.swift
jullianm: It updates its role if the user is already part of the conversation or adds the user to the conversation if not. The legacy method `func addParticipantAndUpdateConversationState(user: ZMUser, role: Role? = nil)` automatically handles both logic. This is a bit confusing, I agree, however in legacy codebase this is how this event is currently processed.

I rename some methods and added comments in https://github.com/wireapp/wire-ios/pull/2091/commits/53edd4d6253f4c3bac2a4973f06d518d1a476786 to make it more clear.

johnxnguyen (approved):

PR #2096, opened 2024-10-28
refactor: create message repository - WPB-11839

jullianm:
WPB-11839

### Key points

This PR introduces a `MessageRepository` and a `MessageLocalStore` to isolate storage operations related to ZMMessage, typically adding a specific message to a given conversation.

### Testing

- different types of messages are correctly added to a given conversation

### Checklist

---


WireDomain/Sources/WireDomain/Repositories/Message/MessageRepository.swift
```diff
@@ -0,0 +1,58 @@
...
+        localStore: any MessageLocalStoreProtocol
+    ) {
+        self.localStore = localStore
+    }
+
+    public func addMessageToConversation(
+        messageType: MessageType,
+        conversationID: UUID,
+        conversationDomain: String?
+    ) async {
```
jullianm, line 51: I try to keep API simple  by avoiding passing around WireDataModel objects in repo methods. I think (ideally) only the storage layer (here the MessageLocalStore) should be dealing with these objects. Basically, we don't want API or Storage objects to leak. in practice, some our existing refactored code is not following this rule so at some point I would like to go over our existing repo and revisit them. what do you think ?

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
```diff
@@ -0,0 +1,383 @@
...
+    ) async
+
+}
+
+public final class MessageLocalStore: MessageLocalStoreProtocol {
+
+    // MARK: - Properties
+
+    let context: NSManagedObjectContext
+    let conversationLocalStore: any ConversationLocalStoreProtocol
```
jullianm, line 39: Here we interact with the conversation local store directly, to fetch a given conversation, I figured it's fine since these local local stores operate on the same level (storage layer), so I'd say we can allow them to do so. What do you think ?

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
```diff
@@ -0,0 +1,383 @@
...
+                sender: creator,
+                users: localParticipants,
+                timestamp: date
+            )
+
+            await context.perform {
+                newConversationMessage.text = userDefinedName
+            }
+
+            await context.perform {
```
netbe: might be good to merge this into one perform

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
```diff
@@ -0,0 +1,383 @@
...
+            systemMessage.systemMessageType = messageType
+            systemMessage.sender = sender
+            systemMessage.users = users ?? Set()
+            systemMessage.addedUsers = addedUsers
+            systemMessage.clients = clients ?? Set()
+            systemMessage.serverTimestamp = timestamp
+
+            if let duration {
+                systemMessage.duration = duration
+            }
```
netbe, line 341:
```suggestion
            systemMessage.duration = duration
```

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
```diff
@@ -0,0 +1,383 @@
...
+
+        let systemMessages = await createSystemMessages(
+            from: messageType,
+            conversation: conversation
+        )
+
+        await addSystemMessages(
+            systemMessages,
+            to: conversation
+        )
```
netbe, line 71: I wonder if we would be better doing once await context.perform and pass the context to each private function

WireDomain/Sources/WireDomain/Repositories/Message/MessageRepository.swift
netbe: yes it makes sense

netbe (approved):
looks good, left some comments

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
johnxnguyen: My gut feeling is that it's better to have fewer, but larger, calls to context.perform, rather than switching back and forth frequently. For now it's fine as it it, but certainly a topic we should discuss further.

johnxnguyen (approved):

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2096/commits/02656a2febcc09c6ddf02148e308b6b0b7565b60

PR #2097, opened 2024-10-28
fix: Placeholder copy of the SearchBar based on the filter that it's applied - WPB-11889

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11889" title="WPB-11889" target="_blank">BugWPB-11889</a>  [iOS] Search field placeholder text doesn’t change text based on filter in Conversation List
  </td></table>
  <br />
 

With this PR we update the placeholder of the search bar based on the filter that it's been applied in the conversation list. 

## Design

Screenshot 2024-10-28 at 15 53 45

Screenshot 2024-10-28 at 15 53 41

Screenshot 2024-10-28 at 15 53 33

---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
@@ -81,6 +81,19 @@ final class ConversationListViewController: UIViewController {
         }
     }
 
+    private var searchPlaceholderText: String {
+        switch listContentController.listViewModel.selectedFilter {
+        case .none:
+            return L10n.Localizable.ConversationList.SearchBar.placeholder
+        case .favorites:
+            return L10n.Localizable.ConversationList.SearchBar.favoritesPlaceholder
+        case .groups:
+            return L10n.Localizable.ConversationList.SearchBar.groupsPlaceholder
+        case .oneOnOne:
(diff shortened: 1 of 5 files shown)


wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
```diff
@@ -398,6 +411,8 @@ final class ConversationListViewController: UIViewController {
         searchController.searchBar.isTranslucent = false
         searchController.searchResultsUpdater = self
 
+        searchController.searchBar.placeholder = searchPlaceholderText
```
samwyndham: It would be nice for this to be tested. One way to achieve this without too much change is to extract this logic into the following at test that without the need to instantiate the SUT. Of course you could also do snapshot tests I imagine.

```swift
private static func makeSearchController(filter: ConversationFilter?) -> UISearchController {
  // ...
}
```

samwyndham (approved):

netbe (approved):

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
agisilaos: Done!

PR #2100, opened 2024-10-29
fix:  Image selector anchor - WPB-11605

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11605" title="WPB-11605" target="_blank">BugWPB-11605</a>  [iOS] Anchor profile image selector popover
  </td></table>
  <br />
 

### Issue

Currently on the navigation overhaul epic branch the popover `UIImagePickerController` is not properly anchored as it doesn't have a suitable `sourceView` set.

This PR:

- Removes an unreferenced protocol, `SettingsExternalScreenCellDescriptorType`, allowing more flexibility when creating `SettingsAppearanceCellDescriptor`, and thus allowing the `sender` of the settings screen to be forwarded to the `UIImagePickerController`s picker.
- Updates `ImagePickerManager.showActionSheet(on:completion:)` to accept a `popoverSourceView` which gets set on the popover controller.
- Uses a popover controller when displaying `UIImagePickerController` on any iPad, not just regular size classes as is necessary for apples API.
- Does some general clean up to simplify code and remove optional instance variables

### Testing

0. Please test on a **real** iPad.
1. Navigate to `Setting`
2. Tap `Profile Picture`
3. Tap `Choose from Library`
4. Photo picker should be well placed
5. Navigate to profile
6. Tap on your profile picker
7. Tap `Choose from Library`
8. Photo picker should be well placed

Also try rotating the device to check things are working

#### NOTE:

After selecting a photo there is a know issue that is not fixed in this PR: https://wearezeta.atlassian.net/browse/WPB-11929

### Checklist


wire-ios/Wire-iOS/Sources/Managers/Image/ImagePickerManager.swift
```diff
@@ -96,23 +107,13 @@ class ImagePickerManager: NSObject {
             imagePickerController.cameraDevice = .front
             imagePickerController.modalTransitionStyle = .coverVertical
         case .photoLibrary, .savedPhotosAlbum:
-            if viewController.isIPadRegular() {
+            if viewController.isIPad() {
```
samwyndham: `isIPadRegular` isn't sufficient. We need to use a popover on any iPad size

wire-ios/Wire-iOS/Sources/Managers/Image/ImagePickerManager.swift
```diff
@@ -76,11 +86,12 @@ class ImagePickerManager: NSObject {
...
 
-    private func getImage(fromSourceType sourceType: UIImagePickerController.SourceType) {
-        guard UIImagePickerController.isSourceTypeAvailable(sourceType),
-              let viewController else {
-                  return
-              }
+    private func getImage(
+        fromSourceType sourceType: UIImagePickerController.SourceType,
+        viewController: UIViewController,
+        popoverSourceView: UIView
```
caldrian: optional: We also have a type `PopoverPresentationControllerConfiguration`. Eventually we should be consistent  and either use `PopoverPresentationControllerConfiguration` or delete it, in my opinion.

wire-ios/Wire-iOS/Sources/Managers/Image/ImagePickerManager.swift
caldrian: I think correct would be to check if the split view controller's layout is collapsed, which is true for iPhones and for iPads when the window is resized so that the horizontal size class is compact.

caldrian (approved):

wire-ios/Wire-iOS/Sources/Managers/Image/ImagePickerManager.swift
agisilaos: Is this type really needed @samwyndham? If not let's start moving away from it.

wire-ios/Wire-iOS/Sources/Managers/Image/ImagePickerManager.swift
agisilaos: Agreed with @caldrian here. Is it possible to be done in this PR @samwyndham?

samwyndham: But won't that mean that it is possible to present this on iPad without using a popover? Won't that crash?

agisilaos (commented):
Left some comments and questions regarding a `PopoverPresentationControllerConfiguration` and how to check if it's iPad by checking if `SplitViewController` layout is collapsed.

wire-ios/Wire-iOS/Sources/Managers/Image/ImagePickerManager.swift
samwyndham: @agisilaos You mean `ImagePickerManager`? I agree it is not that good but I don't think it is worth refactoring this right now compared to other priorities

samwyndham: @caldrian I'll update this to use `PopoverPresentationControllerConfiguration`

wire-ios/Wire-iOS/Sources/Managers/Image/ImagePickerManager.swift
samwyndham: I'll see if there is a iPad at work to test this with

caldrian: I tried if it still works with the full-screen requirement disabled
Screenshot 2024-11-11 at 16 22 35
and resized the window so that the layout collapsed.
It looks like it does. 👍

caldrian (approved):

wire-ios/Wire-iOS/Sources/Managers/Image/ImagePickerManager.swift
samwyndham: Using `PopoverPresentationControllerConfiguration` in c6a5c7bda430113f08e6cda75818b19e60766e12

wire-ios/Wire-iOS/Sources/Managers/Image/ImagePickerManager.swift
samwyndham: @caldrian You mean you tested there was no crash?

caldrian: No crash and the presentation seems to be correct.

samwyndham: I just remembered this about `UIModalPresentationStyle.popover`:

> In iOS 13 and later, for horizontally or vertically compact environments, this option behaves the same as [UIModalPresentationStyle.formSheet](doc://com.apple.documentation/documentation/uikit/uimodalpresentationstyle/formsheet).

Perhaps this is what you meant @caldrian?

So we get the behavior we want just by checking for iPad. No need to worry about size class also.

Simulator Screenshot - iPad (10th generation) - 2024-11-25 at 09 12 48

jullianm (approved):
I think it would be nice for UI related changes to see before and after in a short video or maybe just some screenshots (help fasten the review), otherwise LGTM 👍

PR #2103, opened 2024-10-30
feat: Hide the ability to toggle services in during MLS group creation - WPB-10778

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10778" title="WPB-10778" target="_blank">StoryWPB-10778</a>  [iOS] Disable Add Service interaction in MLS Conversation
  </td></table>
  <br />
 

### Issue

If the protocol of a conversation is MLS, there should be no interface to add services to the conversation.

**Note:** We have `ADD_SERVICE_DISABLED` flag for C builds, and it should remain.

**Create conversation:**
* Hide toggle 
* Hide from the header: `Conversation options: Guests: On, Services: On, Read re…`
* Hide/show when changing the protocol
Simulator Screen Recording - iPhone 16 - 2024-11-01 at 15 03 06

**Add participants:**
* Hide services tab
* Hide section: Services
Simulator Screen Recording - iPhone 16 - 2024-11-01 at 15 05 33

### Testing

- create MLS conversation -> no "Allow services" section;
- add participants to the MLS conversation -> no "Services" tab;

### Checklist

---


wire-ios/Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -305,6 +305,7 @@
 "conversation.create.guidance.toolong" = "Too many characters";
 "conversation.create.options.title" = "Conversation options";
 "conversation.create.options.subtitle" = "Guests: %@, Services: %@, Read receipts: %@";
+"conversation.create.options.subtitle.without_services" = "Guests: %@, Read receipts: %@";
```
KaterinaWire, line 308: **Note:** After merging `Navigation overhaul`, we don't need it because the conversation creation UI has been changed.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/Cells/ConversationCreateOptionsCell.swift
```diff
@@ -101,7 +101,9 @@ extension ConversationCreateOptionsCell: ConversationCreationValuesConfigurable
         let guests = values.allowGuests.localized.capitalized
         let services = values.allowServices.localized.capitalized
         let receipts = values.enableReceipts.localized.capitalized
-        status = L10n.Localizable.Conversation.Create.Options.subtitle(guests, services, receipts)
+        status = values.shouldIncludeServices
+        ? L10n.Localizable.Conversation.Create.Options.subtitle(guests, services, receipts)
+        : L10n.Localizable.Conversation.Create.Options.Subtitle.withoutServices(guests, receipts)
```
KaterinaWire, line 106: **Note:** After merging `Navigation overhaul`, we don't need it because the conversation creation UI has been changed.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
```diff
@@ -59,12 +61,14 @@ final class ConversationCreationController: UIViewController {
         return section
     }()
 
-    private lazy var optionsSections = [
-        guestsSection,
-        servicesSection,
-        receiptsSection,
-        shouldIncludeEncryptionProtocolSection ? encryptionProtocolSection : nil
-    ].compactMap { $0 }
+    private var optionsSections: [ConversationCreateSectionController] {
```
KaterinaWire, line 60: It's not a **lazy** anymore because we need to update this array if the protocol has changed.

wire-ios/Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -324,7 +324,6 @@
 "conversation.create.guidance.empty" = "At least 1 character";
 "conversation.create.guidance.toolong" = "Too many characters";
 "conversation.create.options.title" = "CONVERSATION OPTIONS";
-"conversation.create.options.subtitle" = "Guests: %@, Services: %@, Read receipts: %@";
```
KaterinaWire, line 337: This string is no longer used and has not been removed previously.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
```diff
@@ -113,26 +115,43 @@ final class ConversationCreationController: UIViewController {
...
             }
         }
         return section
     }()
 
+    private func reloadOptionsSections() {
+        guard let collectionView = collectionViewController.collectionView else { return }
+        updateSections()
+
+        let startIndex = 2
```
johnxnguyen: What's the significance of 2? It seems error prone to hardcode an index like this.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationValues.swift
```diff
@@ -65,15 +73,14 @@ final class ConversationCreationValues {
...
-        allowServices: Bool = true,
         enableReceipts: Bool = true,
         encryptionProtocol: Feature.MLS.Config.MessageProtocol,
         selfUser: UserType
     ) {
         self.name = name
         self.unfilteredParticipants = participants
         self.allowGuests = allowGuests
-        self.allowServices = allowServices
+        self.allowServices = !encryptionProtocol.isOne(of: .mls, .mixed)
```
johnxnguyen: can you use `shouldIncludeServices`?

wire-ios/Wire-iOS/Sources/UserInterface/Services/ServiceDetailViewController.swift
```diff
@@ -20,11 +20,13 @@ import UIKit
...
 import WireSyncEngine
 
-extension ConversationLike where Self: SwiftConversationLike {
+extension ConversationLike where Self: GroupDetailsConversationType {
     var botCanBeAdded: Bool {
         return conversationType != .oneOnOne &&
                teamType != nil &&
-               allowServices
+               allowServices &&
+               !messageProtocol.isOne(of: .mls, .mixed)
```
johnxnguyen: This duplication could perhaps be avoided with an extension on the enum?

```swift
messageProtocol.supportsBots
```

KaterinaWire: done. I added an extension for messageProtocol

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
KaterinaWire: `startIndex is 2 because we don't want to reload the first 2 sections (nameSection and errorSection) when the message protocol changes, as this will clear the conversation name.`
I've added this comment to the code

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationValues.swift
KaterinaWire: I can't use `shouldIncludeServices` in `init` method. I reverted my changes and set `allowServices` to true by default, but it will change in the **didSet** of `encryptionProtocol` property

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
johnxnguyen: I think if you can it's better to calculate the indices rather than hardcode them. Hardcoded indices are brittle and can lead to out of bound exceptions. Another option is to modify the name section to take the current 'values' like the other sections, this way even if the section is reloaded it preserves the current name rather than starting with an empty name.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationValues.swift
```diff
@@ -36,7 +36,15 @@ final class ConversationCreationValues {
...
     var enableReceipts: Bool
-    var encryptionProtocol: Feature.MLS.Config.MessageProtocol
+    var encryptionProtocol: Feature.MLS.Config.MessageProtocol {
+        didSet {
+            allowServices = shouldIncludeServices
+        }
+    }
+
+    var shouldIncludeServices: Bool {
+        return !encryptionProtocol.isOne(of: .mls, .mixed)
```
johnxnguyen: Can this use  `encryptionProtocol.supportsBots`?

KaterinaWire: We can't now, because there are 2 enums `MessageProtocol` and `Feature.MLS.MessageProtocol`. The current `supportsBots` computed property is for `MessageProtocol`, but `ConversationCreationValues` uses `Feature.MLS.MessageProtocol`. I don't know the reason why we introduced 2 types. @johnxnguyen Do you remember the purpose of creating a separate type for Feature MLS?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
KaterinaWire: I removed the hardcode value and changed the code. Please take a look

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationValues.swift
KaterinaWire: If we were using `Feature.MLS.Config.MessageProtocol`  just to decode the payload, it could be a separate type. But we are using `Feature.MLS.Config.MessageProtocol` in the conversation creation flow, which means we have 2 types with the same values. This looks like code duplication.
I changed the conversation creation flow and use `MessageProtocol` there. As a result, we can re-use `supportsBots` property.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
```diff
@@ -113,28 +115,45 @@ final class ConversationCreationController: UIViewController {
...
         return section
     }()
 
+    private func reloadOptionsSections() {
+        guard let collectionView = collectionViewController.collectionView else { return }
+        updateSections()
+
+        let excludedSectionIndex = collectionViewController.sections.startIndex
+        let endIndex = collectionView.numberOfSections
+        let sectionsToReload = IndexSet(integersIn: (excludedSectionIndex + 1)..<endIndex)
```
johnxnguyen, line 142: Another way to write this could be:

```
let allIndices = IndexSet(integersIn: 0..<collectionView.numberOfSections)
let indicesToReload = allIndices.dropFirst()
```

johnxnguyen (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
netbe: just to fully understand: what section we avoid to reload? @KaterinaWire

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
```diff
@@ -470,3 +492,16 @@ extension ConversationCreationController {
         return alert
     }
 }
+
+extension Feature.MLS.Config.MessageProtocol {
```
netbe, line 503:
```suggestion
private extension Feature.MLS.Config.MessageProtocol {
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
KaterinaWire: @netbe  **Conversation name.** 
Use case:
- I enter a conversation name
- I change the protocol
As a result, I want to keep the conversation name and not reload it. So it will be available **only** for testing, since the protocol change will be hidden from users. We can just ignore it and reload all sections, anyway it's just to improve the user experience of testing.

netbe: ok it's nice to think about our XP, maybe adding a comment here
```suggestion
        // ignoring the conversation name so we don't loose the info while testing
        let excludedSectionIndex = collectionViewController.sections.startIndex
        let endIndex = collectionView.numberOfSections
        let sectionsToReload = IndexSet(integersIn: (excludedSectionIndex + 1)..<endIndex)
```

netbe (approved):
Looks good, just small suggestions

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
KaterinaWire: The extension can't be private because we use it in `AddParticipantsViewController` too.

PR #2117, opened 2024-11-01
chore: Simplify UI code by removing `ShapeView` abstraction - WPB-12032

agisilaos:
This PR removes unnecessary view abstractions to simplify our codebase while maintaining all existing functionality.

## Changes:
- Remove unused `ShapeView` class and its generic base class
- Remove unused `DotView` implementation
- Replace `ShapeView` usage in BaseAccountView with standard `UIView`
- Update `PersonalAccountView` and `TeamAccountView` to use native layer properties for selection indicator shapes

## Technical Details:
- Migrated from custom layer hosting to direct `UIKit` layer properties
- Moved corner radius setup to `layoutSubviews()` for both `PersonalAccountView` and `TeamAccountView` as it's the proper place for layout-dependent properties and ensures correct behavior during size changes
- Maintained same visual appearance and behavior:
  - `PersonalAccountView`: Circular selection indicator using cornerRadius = bounds.width / 2
  - `TeamAccountView`: Rounded rectangle selection indicator using cornerRadius = 6

## Note on layoutSubviews:
Corner radius is a layout-dependent property that might need to be updated when the view's bounds change. By setting it in `layoutSubviews()`, we ensure it's always correct, even after rotation, size changes, or other layout updates. This is more robust than setting it once during initialization.

---

### Checklist

---


wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/AccountViews/TeamAccountView.swift
```diff
@@ -65,6 +55,11 @@ final class TeamAccountView: BaseAccountView {
         }
     }
 
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        selectionView.layer.cornerRadius = 6
```
caldrian, line 60: Isn't it sufficient to set this value once?

caldrian (approved):

johnxnguyen (approved):

wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/AccountViews/TeamAccountView.swift
agisilaos: I would say so. Didn't find any issues during testing.

PR #2124, opened 2024-11-04
fix: Multiple conversation reloads on dynamic font size changes - WPB-12077

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-12077" title="WPB-12077" target="_blank">BugWPB-12077</a>  [iOS] ConversationView Reloading Multiple Times on Dynamic Font Size Change
  </td></table>
  <br />
 

## Issue

When users change their device's font size settings, the `ConversationView` was being unnecessarily reloaded and pushed multiple times onto the navigation stack. This was caused by the `ContentSizeCategoryDidChange` notification triggering a full conversation reload through the mainCoordinator.

## Solution

- Removed the reloadCurrentConversation() function that was causing multiple view pushes. 
- The view now handles font size changes naturally through the iOS dynamic type system
- Eliminated unnecessary view reloading and navigation stack modifications

## Impact

- Improved app performance during font size changes
- Fixed navigation stack issues
- Better memory usage by preventing multiple view controller instances
- More stable user experience when changing accessibility settings

## BEFORE

Before - with bug

## AFTER

After - without bug

---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
@@ -173,8 +173,6 @@ final class ZClientViewController: UIViewController {
 
         NotificationCenter.default.post(name: NSNotification.Name.ZMUserSessionDidBecomeAvailable, object: nil)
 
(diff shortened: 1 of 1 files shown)


jullianm (approved):

caldrian (approved):
Nice, thanks!

PR #2125, opened 2024-11-05
feat: Create `FolderPicker` - WPB-11892

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11892" title="WPB-11892" target="_blank">TaskWPB-11892</a>  [iOS] Implement non empty state in `Folders Picker` view
  </td></table>
  <br />
 

### Issue

This PR:

- introduces a `FolderPicker` SwiftUI  component. 
- adds a new test target - `WireFolderPickerUI` - along with `Localizable` & `Accessibility` string catalogues.
- moves the `Close` icon to `WireDesign`
- adds a `CloseButton` SwiftUI component to `WireReusableUIComponents`

 It is not yet used in the main project. Still todo:

- [ ] Implement empty state
- [ ] Snapshot test (somehow). I ran into [this issue](https://github.com/pointfreeco/swift-snapshot-testing/discussions/781) attempting to snapshot test this view. I _think_ the only way to snapshot test this correctly will be within an app host 😢 

Screenshot 2024-11-05 at 08 36 23

### Testing

No tests added yet

---

### Checklist

---


WireUI/Sources/WireDesign/Colors/Color+WireAccentColor.swift
```diff
@@ -0,0 +1,26 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import SwiftUI
+import WireFoundation
+
+public extension Color {
+    init(_ wireAccentColor: WireAccentColor) {
+        self = WireAccentColorMapping().color(for: wireAccentColor)
```
samwyndham, line 24: @caldrian Any issue with doing this?

crowdin.yml
```diff
@@ -54,5 +54,17 @@ files: [
    "dest" : "/App/WireMainNavigationUI/Resources/Accessibility.xcstrings",
    "translation" : "%original_path%/%original_file_name%",
    "multilingual": true,
-  },
+ },
+ {
+  "source" : "/WireUI/Sources/WireFolderPickerUI/Resources/Localizable.xcstrings",
+  "dest" : "/App/WireFolderPickerUI/Resources/Localizable.xcstrings",
```
samwyndham, line 60: @caldrian Do you know if I have to manually create these items on crowdin or will they be auto created when the localization action runs?

WireUI/Package.swift
```diff
@@ -16,7 +16,8 @@ let package = Package(
         .library(name: "WireMainNavigationUI", targets: ["WireMainNavigationUI"]),
         .library(name: "WireReusableUIComponents", targets: ["WireReusableUIComponents"]),
         .library(name: "WireSettingsUI", targets: ["WireSettingsUI"]),
-        .library(name: "WireSidebarUI", targets: ["WireSidebarUI"])
+        .library(name: "WireSidebarUI", targets: ["WireSidebarUI"]),
+        .library(name: "WireFolderPickerUI", targets: ["WireFolderPickerUI"])
```
caldrian, line 20: Do you mind ordering it alphabetically? :)

crowdin.yml
caldrian: They will be created.

wire-ios/Wire-iOS/Sources/UserInterface/DeviceView/Views/E2EIdentityCertificateDetailsView.swift
```diff
@@ -51,7 +51,8 @@ struct E2EIdentityCertificateDetailsView: View {
                         didDismiss?()
                     },
                     label: {
-                        Image(.close)
+                        Image
+                            .close
```
caldrian, line 55: Is this minor difference really worth spoiling the `Image` namespace?
`Image(.close)` vs. `Image.close`
First, I'd consider the first variant more readable, everybody is familiar with the `init` taking an `ImageResource`.
Second, it doesn't scale well. Imagine having all images in the codebase as extension to `SwiftUI.Image`. A long list of assets mixed with the original API.

In my opinion the `ImageResource` is a great shortcut on its own.

WireUI/Sources/WireFolderPickerUI/FolderPicker.swift
```diff
@@ -0,0 +1,123 @@
...
+                    CloseButton(
+                        action: didTapClose,
+                        accessibilityLabel: String(
+                            localized: "folderPicker.close.label",
+                            table: "Accessibility",
+                            bundle: .module
+                        )
+                    )
+                }
+            }
```
caldrian, line 83: In the design the folder title is on the left side. Have you considered implementing it like this?
```suggestion
            }
            ToolbarItem(placement: .topBarLeading) {
                Text("folderPicker.title", tableName: "Localizable", bundle: .module)
                    .font(.title3.bold())
            }
```
Screenshot 2024-11-05 at 14 41 17

We would need to inject the correct text style though.

WireUI/Sources/WireFolderPickerUI/FolderPicker.swift
```diff
@@ -0,0 +1,123 @@
...
+        dismiss()
+    }
+
+}
+
+// MARK: - Previews
+
+@available(iOS 17.0, *)
+#Preview {
+    @Previewable @State var isPresented = false
```
caldrian, line 97:
```suggestion
    @Previewable @State var isPresented = true
```
... otherwise during editing you constantly have to tap "Show Picker"

WireUI/Sources/WireDesign/Icons/Image+ImageResource.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import SwiftUI
+
+public extension Image {
+    static let close = Image(.close)
+}
```
caldrian, line 23: optional: I'd delete this code, see the related comment.

caldrian (approved):
Nice, why isn't all our UI in SwiftUI?

wire-ios/Wire-iOS/Sources/UserInterface/DeviceView/Views/E2EIdentityCertificateDetailsView.swift
samwyndham: @caldrian I agree. This is just a work around for an Xcode annoyance.

The issue is that the `close` asset has moved to another target. Unfortunately we can't reference an `ImageResource` from another target because as far as I can tell an `ImageResource` is an **internal** type 🤯 to the target. e.g. Xcode is generating a new `ImageResource` type in each target.

This is super annoying but I can't find a better workaround.

WireUI/Sources/WireFolderPickerUI/FolderPicker.swift
samwyndham: I originally had this but when discussing with Wolfgang he said to keep it as whatever the default is so kept it in the center

WireUI/Package.swift
samwyndham: I will fix this in https://github.com/wireapp/wire-ios/pull/2127

WireUI/Sources/WireFolderPickerUI/FolderPicker.swift
samwyndham: I will fix this in https://github.com/wireapp/wire-ios/pull/2127

WireUI/Sources/WireFolderPickerUI/FolderPicker.swift
agisilaos: @samwyndham Then please ping him to update the design in Figma.

wire-ios/Wire-iOS/Sources/UserInterface/DeviceView/Views/E2EIdentityCertificateDetailsView.swift
caldrian: I see, what about creating a `CloseButton` view in `ReuisableUIComponents`?

WireUI/Sources/WireFolderPickerUI/FolderPicker.swift
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    public var body: some View {
+        List {
+            Picker("", selection: $selected) {
+                ForEach(options) { option in
+                    Text(option.title)
+                        .font(.textStyle(.body1))
+                        .lineLimit(1)
+                        .foregroundStyle(option.id == selected ? Color(accentColor) : .primaryText)
```
agisilaos, line 58: We don't have a color named `.primaryText`. We should avoid using new colors or Apple defined ones. Instead if the color doesn't exist in ColorTheme we should create it.

wire-ios/Wire-iOS/Sources/UserInterface/DeviceView/Views/E2EIdentityCertificateDetailsView.swift
samwyndham: The `CloseButton` itself already exists so I could probably update `E2EIdentityCertificateDetailsView.swift` to use it. So then the pattern would be for the close icon to be in the `ReuisableUIComponents` target, not in `WireDesign`. 

I think that should work in this case but I'm worried it is too rigid. It basically means we should never add image assets to `WireDesign` right?

samwyndham: Interestingly a public `ImageResource` has been added to iOS 17 but it still only generates internal assets. It does mean that something like this is possible from iOS 17. Still not great:

```swift
enum Icon {
    static let close = ImageResource.close
}

let image = Image(Icon.close)
```

agisilaos (approved):
@samwyndham Looking great! Thanks. Left one small comment regarding the color we should create in ColorTheme.

WireUI/Sources/WireFolderPickerUI/FolderPicker.swift
samwyndham: This isn't apples. This is from `Color+Named.swift` 🤔

wire-ios/Wire-iOS/Sources/UserInterface/DeviceView/Views/E2EIdentityCertificateDetailsView.swift
samwyndham: @caldrian I'll look into improving this in a separate PR. Maybe we can discuss the topic a bit further on Friday... Essentially where should image resources live

PR #2139, opened 2024-11-07
feat: QR code opens links directly to the app - WPB-12114

KaterinaWire:
### Issue
Based on Security review and approval [here](https://wearezeta.atlassian.net/wiki/spaces/PAD/pages/1512833026/2024-10-28+RFC+Trusted+QR+Codes?focusedCommentId=1535901745),  QR code should open links directly to the app rather than going to the web and then coming to the app.

### Implementation
- add profileDeepLink for UserType
- use it to generate QR code

**Note:** the text link to the self profile should remain the same.

### Testing

Scan the QR code -> redirect directly to the app.

### Checklist

---


jullianm (approved):

wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/QR Code/Models/UserQRCodeViewModel.swift
```diff
@@ -27,10 +27,11 @@ final class UserQRCodeViewModel: ObservableObject {
 
     init(
         profileLink: String,
+        profileDeepLink: String,
         handle: String
     ) {
         self.profileLink = profileLink
-        let qrCodeImage = QRCodeGenerator.generateQRCode(from: profileLink)
+        let qrCodeImage = QRCodeGenerator.generateQRCode(from: profileDeepLink)
```
agisilaos, line 34: What happens if someone scans the QR code and they don't have the app installed? A much better flow would be to transition to the App Store so they can download the app. I think it can be achieved by introducing a method like this one:

```swift
func openProfileDeepLink() {
        guard let appURL = URL(string: profileDeepLink),
              let appStoreURL = URL(string: "https://apps.apple.com/app/idYOUR_APP_ID") else { return }

        // Check if the app can open the deep link
        if UIApplication.shared.canOpenURL(appURL) {
            UIApplication.shared.open(appURL, options: [:], completionHandler: nil)
        } else {
            // Redirect to the App Store if the app is not installed
            UIApplication.shared.open(appStoreURL, options: [:], completionHandler: nil)
        }
    }
}
```

Then we can call this method through our viewModel wherever we want.

agisilaos (commented):
Left a question and suggestion regarding the absence of the wire app on user's phone and what should happen next.

wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/QR Code/Models/UserQRCodeViewModel.swift
KaterinaWire: This was discussed in the RFC: https://wearezeta.atlassian.net/wiki/spaces/PAD/pages/1512833026/2024-10-28+RFC+Trusted+QR+Codes?focusedCommentId=1535901745
Screenshot 2024-11-08 at 16 30 04

Android also doesn't navigate to Google store. 
And additional question is about Col builds, that should not navigate to the App Store

agisilaos (approved):

PR #2140, opened 2024-11-08
fix: transparent background on connect for iPad - WPB-12041

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-12041" title="WPB-12041" target="_blank">BugWPB-12041</a>  [iOS] Transparent background on the invite screen
  </td></table>
  <br />
 

### Issue

* fix background
* fix navigation flow on connect when accessing to contacts or asking for share permissions
* replace ZClientController.shared?.present with instance method `presentOverAll()` to avoid access to deprecated shared instance

### Testing

* login on iPad with personal account
* tap connect
* tap invite

* [x] check also flows on iPhone


caldrian (approved):

wire-ios/Wire-iOS/Sources/UserInterface/StartUI/StartUI/StartUIViewController+AddressBook.swift
```diff
@@ -38,7 +38,21 @@ extension StartUIViewController {
 extension StartUIViewController: ShareContactsViewControllerDelegate {
 
     func shareContactsViewControllerDidFinish(_ viewController: ShareContactsViewController) {
-        viewController.dismiss(animated: true)
+        guard !needsAddressBookPermission else {
+            // do nothing just dismiss
```
samwyndham: Just to check I assume this is being dismissed by something else in this case

wire-ios/Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController.swift
```diff
@@ -205,12 +204,20 @@ final class ContactsViewController: UIViewController {
         let willAppear = (beginFrame.minY - endFrame.minY) > 0
         let padding: CGFloat = 12
 
-        UIView.animate(withKeyboardNotification: notification, in: view, animations: { [weak self] keyboardFrame in
+        // Calculate the effective keyboard height
+        var keyboardHeight = view.frame.intersection(endFrame).height
+
+        // On iPad, add an estimated accessory height if applicable
+        if traitCollection.userInterfaceIdiom == .pad {
+            keyboardHeight += 44  // Approximate height of an input accessory view
```
samwyndham: I'm pretty sure there must be a way to do this without resorting to an estimate which may break. Have you looked into this much?

netbe: good point, but I haven't found any way to get it so far

wire-ios/Wire-iOS/Sources/UserInterface/StartUI/StartUI/StartUIViewController+AddressBook.swift
netbe: this guard else never happens, removed

wire-ios/Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController.swift
samwyndham: I guess what I expect is that the `keyboardWillChangeFrameNotification` would include the height of any accessory. Did you see that this is no happening in practice?

Otherwise have you considered:

- Inheriting from `UITableViewController` - I think this takes care of keyboard avoidance completely.
- Using [UIKeyboardLayoutGuide](https://developer.apple.com/documentation/uikit/uikeyboardlayoutguide) instead of responding to notifications.

samwyndham (approved):
Nice work! Approving but please first take a look if there is a better approach to keyboard avoidance that doesn't require estimates.

wire-ios/Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController.swift
netbe: used [UIKeyboardLayoutGuide](https://developer.apple.com/documentation/uikit/uikeyboardlayoutguide) @samwyndham thks for the suggestion

PR #2142, opened 2024-11-08
feat: Implement ConversationFolderCreationUseCase for Folder Creation - WPB-12171

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-12171" title="WPB-12171" target="_blank">TaskWPB-12171</a>  [iOS] Introduce ConversationFolderCreationUseCase for folder creation
  </td></table>
  <br />
 

This PR introduces the `CreateConversationFolderUseCase`, a use case responsible for creating conversation folders with specified names in the Core Data model. The use case is designed to streamline folder creation within the app, ensuring that folders are correctly instantiated with necessary attributes and unique identifiers. Additionally, a suite of unit tests `CreateConversationFolderUseCaseTests` is included to verify the correctness and reliability of this use case.

### CreateConversationFolderUseCaseProtocol:

- Added a protocol defining the folder creation use case with an invoke(with name:) method.
- The method accepts a String name parameter and returns an optional LabelType, which represents the created folder.

### CreateConversationFolderUseCase Implementation:
- Implemented the `CreateConversationFolderUseCase` struct, conforming to `CreateConversationFolderUseCaseProtocol`.
	- The invoke(with name:) method:
	- Creates a LabelType entity in the provided `NSManagedObjectContext` with a unique `UUID` as its `remoteIdentifier`.
	- Sets the name property to the provided folder name and the kind property to .folder.
	 - Returns the created folder `LabelType`, or nil if creation fails.

### Unit Tests in `CreateConversationFolderUseCaseTests`:
- Introduced `CreateConversationFolderUseCaseTests` to validate the functionality of the use case.

---

### Checklist


wire-ios-sync-engine/Source/Use cases/ConversationFolderCreationUseCase.swift
```diff
@@ -0,0 +1,61 @@
...
+                remoteIdentifier: UUID(),
+                create: true,
+                in: managedObjectContext,
+                created: &created
+            )
+            label?.name = name
+            label?.kind = .folder
+            return label
+
+        }
```
caldrian:
```suggestion
            return label
        }
```

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+UserSession.swift
```diff
@@ -360,6 +360,10 @@ extension ZMUserSession: UserSession {
         return SubmitCallQualitySurveyUseCase(analyticsEventTracker: analyticsEventTracker)
     }
 
+    public func makeConversationFolderCreationUseCase() -> ConversationFolderCreationUseCaseProtocol {
+        return ConversationFolderCreationUseCase(managedObjectContext: self.syncContext)
```
caldrian:
```suggestion
        return ConversationFolderCreationUseCase(managedObjectContext: syncContext)
```

caldrian (approved):

wire-ios-sync-engine/Source/Use cases/ConversationFolderCreationUseCase.swift
```diff
@@ -0,0 +1,49 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireDataModel
+
+public struct ConversationFolderCreationUseCase: ConversationFolderCreationUseCaseProtocol {
```
samwyndham: Maybe a topic for architecture forum but I think it would make sense to follow naming typically used on Android: See documentation: https://developer.android.com/topic/architecture/domain-layer#conventions

> verb in present tense + noun/what (optional) + UseCase.

Optional: Rename this to `CreateConversationFolderUserCase` or `CreateFolderUseCase`

wire-ios-sync-engine/Source/Use cases/ConversationFolderCreationUseCase.swift
```diff
@@ -0,0 +1,49 @@
...
+            let label = Label.fetchOrCreate(
+                remoteIdentifier: UUID(),
+                create: true,
+                in: managedObjectContext,
+                created: &created
+            )
+            label?.name = name
+            label?.kind = .folder
+            return label
+        }
```
samwyndham: I wonder if this use case should also take care of saving... What do you think @caldrian?

samwyndham (approved):
Nice work!

wire-ios-sync-engine/Source/Use cases/ConversationFolderCreationUseCase.swift
agisilaos: Done! Renamed it to `CreateConversationFolderUserCase`

wire-ios-sync-engine/Source/Use cases/ConversationFolderCreationUseCase.swift
caldrian: I think saving should be explicitly done outside.
What if a use case should run several times in a row and we only want to save if all runs were successful?

agisilaos: I think it was discussed in the past and for now that's happening outside of use case. We usually don't save inside the use cases. @samwyndham Maybe we can discuss it in our next Architecture Forum?

PR #2143, opened 2024-11-08
chore: Run `SwiftFormat` across project - WPB-11990

samwyndham:
### Issue

This PR:

- Updates our `.swiftformat` to include the rules we agreed on.
- Updates the `Format & Lint Swift Code` target to run `SwiftFormat` across our project.
- Runs `SwiftFormat` across our project

There were some issues I had to fix. Most of these were related to `TODO` comments without tickets, which became `TODO:` after formatting. In most cases I just added a `// swiftLint:ignore` directive.

There were also a few compile errors which I had resolve manually but nothing too bad.

**Note**

I had planned to run all tests against this branch but it seems that `develop` is currently broken. I recommend we fix it before merging this PR.

### Testing

Run tests

---

### Checklist


.swiftlint.yml
```diff
@@ -22,7 +22,6 @@ opt_in_rules:
   - self_binding
   - self_in_property_initialization
   - shorthand_optional_binding
-  - sorted_imports
```
caldrian, line 25: Is imports sorting done by SwiftFormat now?

scripts/Package.swift
```diff
@@ -51,8 +51,8 @@ let package = Package(
         ),
         .binaryTarget(
             name: "swiftformat",
-            url: "https://github.com/nicklockwood/SwiftFormat/releases/download/0.54.0/swiftformat.artifactbundle.zip",
-            checksum: "edf4ed2f1664ad621ae71031ff915e0c6ef80ad66e87ea0e5a10c3580a27a6dd"
+            url: "https://github.com/nicklockwood/SwiftFormat/releases/download/0.54.6/swiftformat.artifactbundle.zip",
```
caldrian: Could you please also update `.github/workflows/swiftformat.yml` if there is an image for that version available?

caldrian (approved):

.swiftlint.yml
samwyndham: Yes they kind of interfere with each other otherwise... `SwiftFormat` will place all `testable` after regular imports but it doesn't take care of adding a space between them. If there is no space between them `SwiftLint` will try to sort them alphabetically so `testable` and regular get mixed up. I couldn't find an easy way to fix this

netbe: maybe disable the swiftlint rule @samwyndham ?

samwyndham: @netbe What do you mean. You mean something else than deleting the rule?

scripts/Package.swift
samwyndham: I've changed this to `0.54.5` as this is the latest image available and updated `.github/workflows/swiftformat.yml`. Thanks for pointing that out!

agisilaos (approved):

.swiftlint.yml
caldrian: By removing this line from opt_in_rules it's already disabled.

netbe (approved):
🚀

PR #2153, opened 2024-11-13
fix: Datadog integration leads to build failures - WPB-11936

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11936" title="WPB-11936" target="_blank">TaskWPB-11936</a>  Re-enable CallQualityControllerTests
  </td></table>
  <br />
 

This reverts commit 465f8be78b5f587a9f8e562fa44cea1c6c13eeb2.

### Issue
This PR attempts to fix develop tests by:

1. reverting PR #2110: this is still an issue we will need to fix (ticket back to TODO)

2. separate WireDatadog and WireAnalytics: 

WireAnalytics is still in SynEngine:
WireDatadog is in WireCommonComponents
The `canImport` is removed entirely from WireCommonComponents so all datadog imports are within WireDatadog

3. disable code coverage for SyncEngine

### Testing

* [x] run syncEngine tests
* [x] run C build with security tests plans - [DD enabled](https://github.com/wireapp/wire-ios/actions/runs/11857283394)
* [x] run [test develop](https://github.com/wireapp/wire-ios/actions/runs/11857283394)


WireAnalytics/Package.swift
```diff
@@ -19,20 +22,16 @@ let package = Package(
...
             name: "WireDatadog",
-            dependencies: [
-                .product(name: "DatadogCore", package: "dd-sdk-ios"),
-                .product(name: "DatadogCrashReporting", package: "dd-sdk-ios"),
-                .product(name: "DatadogLogs", package: "dd-sdk-ios"),
-                .product(name: "DatadogRUM", package: "dd-sdk-ios"),
-                .product(name: "DatadogTrace", package: "dd-sdk-ios")
-            ],
+            dependencies: datadogDependencies(),
+            path: "Sources/WireDatadog",
```
samwyndham, line 33: Did you consider making the path change instead of the source files? It might be easier to maintain 🤔

As a side note, if we take the file route perhaps this line is not necessary as I guess that is the default path

wire-ios-sync-engine/Tests/TestsPlans/AllTests.xctestplan
```diff
@@ -9,6 +9,7 @@
     }
   ],
   "defaultOptions" : {
+    "codeCoverage" : false,
```
samwyndham, line 12: Did you confirm that this was necessary?

samwyndham (approved):
🥳

WireAnalytics/Package.swift
netbe: from the doc it was saying here lies the directory and on the other one just filenames. I didn't play with this

wire-ios-sync-engine/Tests/TestsPlans/AllTests.xctestplan
netbe: not completely sure, but I think it does not hurt to keep it disabled for now since we don't collect code coverage for now

caldrian (approved):

PR #2141, opened 2024-11-08
refactor: create local stores and UTS - WPB-12100

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-12100" title="WPB-12100" target="_blank">TaskWPB-12100</a>  [iOS] Create remaining local stores and add UTs for all stores
  </td></table>
  <br />
 

### Key points

This PR isolates tests for our storage layer (local stores) and revisits tests for all our existing repositories.
Missing local stores were also created so the storage logic could be moved out from related repositories.

BEFORE:
- Local stores were indirectly tested through our repositories

AFTER:
- Each local store is **directly** tested.
- Mock local stores are injected into repositories, repositories tests were revisited consequently.

In addition:
- Removes super.setUp() and super.tearDown() code from all tests
- Ensures local stores know nothing about API objects (using model mappers)
- Uses .mockID instances from WireTestingPackage
- Harmonizes tests naming

### Testing

- Add dedicated tests for all local stores 
- Revisit repositories tests 

### Checklist

- [x] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [x] Description is filled and free of optional paragraphs.
- [x] Adds/updates automated tests.

---

### UI accessibility checklist

_If your PR includes UI changes, please utilize this checklist:_
- [ ] Make sure you use the API for UI elements that support large fonts.
- [ ] All colors are taken from WireDesign.ColorTheme or constructed using WireDesign.BaseColorPalette.
- [ ] New UI elements have Accessibility strings for VoiceOver.

PR #2164, opened 2024-11-18
refactor: process conversation proteus/MLS message add event - WPB-10174

jullianm:
WPB-10174

### Key points

Specifically, this PR is about porting the existing implementation of the `ConversationMLSMessageAdd` and the `ConversationProteusMessageAdd` event.

A previous [PR](https://github.com/wireapp/wire-ios/pull/2120) was opened and is now closed since this PR contains both implementation (Proteus and MLS) - related tickets:
- [ConversationProteusAddEvent](https://wearezeta.atlassian.net/browse/WPB-10174)
- [ConversationMLSAddEvent](https://wearezeta.atlassian.net/browse/WPB-10172)

### Testing

- [x] MLS and Proteus messages are correctly decrypted (using dedicated decryptors)
- [x] MessageRepository `addMessage` for MLS/Proteus methods are correctly invoked.
- [x] A Proteus message is added to a conversation (MessageLocalStoreTests)
- [x] A Proteus with a large payload message (externalData) is added to a conversation (MessageLocalStoreTests)
- [x] A MLS message is added to a conversation (MessageLocalStoreTests)

### Checklist


WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Event decoding/Conversation/ConversationProteusMessageAddEventDecoder.swift
```diff
@@ -47,8 +47,8 @@ struct ConversationProteusMessageAddEventDecoder {
             conversationID: conversationID,
             senderID: senderID,
             timestamp: timestamp.date,
-            message: .ciphertext(payload.text),
-            externalData: payload.data.map { .ciphertext($0) },
+            message: .init(encryptedMessage: payload.text),
+            externalData: payload.data.map { .init(encryptedMessage: $0) },
```
jullianm, line 51: as discussed, removed the enum and used a struct instead

WireAPI/Sources/WireAPI/Models/UpdateEvent/ConversationEvent/ConversationMLSMessageAddEvent.swift
```diff
@@ -41,5 +56,14 @@ public struct ConversationMLSMessageAddEvent: Equatable, Codable {
...
     public let message: String
+    
+    /// The date the message was received.
+
+    public let timestamp: Date?
+    
+    /// The decrypted current message + decrypted buffered messages
+    /// along with the related sender client ID for each message.
+
+    public var decryptedMessages: [DecryptedMessage] = []
```
jullianm, line 67: similar to the Proteus add event with the message content property, this property will be filled in the decryptor when the message + buffered messages are decrypted

WireDomain/Sources/WireDomain/Event Decryption/MLSMessageDecryptor.swift
```diff
@@ -0,0 +1,179 @@
...
+        let decryptedMessages = await processMLSMessageDecryptionResults(
+            decryptionResults,
+            mlsConversation: mlsConversation,
+            senderID: eventData.senderID.uuid,
+            senderDomain: eventData.senderID.domain,
+            date: eventData.timestamp
+        )
+        
+        var decryptedEvent = eventData
+        decryptedEvent.decryptedMessages = decryptedMessages
```
jullianm, line 83: filled here

WireDomain/Sources/WireDomain/Event Decryption/UpdateEventDecryptor.swift
```diff
@@ -93,6 +110,24 @@ struct UpdateEventDecryptor: UpdateEventDecryptorProtocol {
...
+                
+            case .conversation(.mlsMessageAdd(let eventData)):
+                
+                WireLogger.updateEvent.info(
+                    "decrypting MLS event...",
+                    attributes: logAttributes
+                )
+
+                do {
+                    let decryptedEventData = try await mlsMessageDecryptor.decryptedEventData(from: eventData)
```
jullianm, line 122: similar to what we did for the proteusMessageAdd, we use the decryptor to decrypt the MLS messages from the event and return the same event filled with the decrypted messages (current one and buffered ones)

WireDomain/Sources/WireDomain/Repositories/UserClients/UserClientsRepository.swift
```diff
@@ -138,6 +165,72 @@ public struct UserClientsRepository: UserClientsRepositoryProtocol {
...
+    public func proteusSessionID(
+        for client: WireDataModel.UserClient
+    ) async -> ProteusSessionID? {
+        await context.perform {
+            client.proteusSessionID
+        }
+    }
+    
+    // swiftlint:disable:next todo_requires_jira_link
+    // TODO: Move to UserClientsLocalStore when related PR is merged
```
jullianm: the related [PR](https://github.com/wireapp/wire-ios/pull/2141)

WireDomain/Tests/WireDomainTests/LocalStores/MessageLocalStoreTests.swift
```diff
@@ -53,35 +56,196 @@ final class MessageLocalStoreTests: XCTestCase {
...
+        
+        let expectedMessageText = "Everything"
+        
+        await internalTest_assertConversationLastMessage(
+            expectedMessageText: expectedMessageText,
+            conversation: mlsConversation
+        )
+    }
+    
+    func testAddProteusMessage_It_Adds_Message_To_Conversation() async throws {
```
jullianm: testing a regular message is added to conversation

WireDomain/Tests/WireDomainTests/LocalStores/MessageLocalStoreTests.swift
```diff
@@ -53,35 +56,196 @@ final class MessageLocalStoreTests: XCTestCase {
...
+        
+        let expectedMessageText = "Everything"
+        
+        await internalTest_assertConversationLastMessage(
+            expectedMessageText: expectedMessageText,
+            conversation: groupConversation
+        )
+    }
+    
+    func testAddProteusMessage_It_Adds_Big_Payload_Message_To_Conversation() async throws {
```
jullianm: testing a large message payload (external) is added to the conversation

wire-ios-data-model/Source/Model/Message/ZMClientMessage+Composite.swift
```diff
@@ -46,7 +46,7 @@ extension ZMClientMessage: CompositeMessageData {
 
 // MARK: - ButtonStates Interface
 extension ZMClientMessage {
-    static func updateButtonStates(withConfirmation confirmation: ButtonActionConfirmation,
+    public static func updateButtonStates(withConfirmation confirmation: ButtonActionConfirmation,
```
jullianm: some methods were made public to facilitate the work in this PR, many others had to be factored out because they could not be reused (the ones that take a ZMUpdateEvent as parameter)

WireDomain/Sources/WireDomain/Event Decryption/MLSMessageDecryptor.swift
```diff
@@ -0,0 +1,168 @@
...
+    ) async throws -> ConversationMLSMessageAddEvent {
+        let conversationID = eventData.conversationID
+
+        guard let mlsConversation = await conversationLocalStore.fetchConversation(
+            id: conversationID.uuid,
+            domain: conversationID.domain
+        ) else {
+            WireLogger.mls.error(
+                "failed to add mls message: conversation not found in db"
+            )
```
netbe: if we throw I guess we can remove logging the error

WireDomain/Sources/WireDomain/Event Decryption/MLSMessageDecryptor.swift
```diff
@@ -0,0 +1,168 @@
...
+                    "successfully decrypted mls message but no result was returned"
+                )
+
+                return []
+            }
+
+            return results
+
+        } catch {
+            WireLogger.mls.warn(
```
netbe:
```suggestion
            WireLogger.mls.error(
```

WireAPI/Sources/WireAPI/Models/UpdateEvent/ConversationEvent/ConversationMLSMessageAddEvent.swift
```diff
@@ -42,4 +57,13 @@ public struct ConversationMLSMessageAddEvent: Equatable, Codable {
...
     public let message: String
 
+    /// The date the message was received.
+
+    public let timestamp: Date?
+
+    /// The decrypted current message + decrypted buffered messages
+    /// along with the related sender client ID for each message.
+
+    public var decryptedMessages: [DecryptedMessage] = []
```
caldrian, line 67: I find it a bit confusing to name the type `DecryptedMessage`, having a property `message` as well as another property named `decryptedMessages`.

IS it a message, 
HAS it A message or 
HAS it MANY messages?

caldrian (approved):
I only checked superficially, since I'm not very familiar with this topic. 👍

WireAPI/Sources/WireAPI/Models/Messaging/MessageContent.swift
```diff
@@ -21,14 +21,14 @@ import Foundation
...
 
     /// Encrypted message content.
 
-    case ciphertext(String)
+    public let encryptedMessage: String
 
     /// Unencrypted message content.
 
-    case plaintext(String)
+    public var decryptedMessage: String?
```
johnxnguyen, line 32: Is this sendable even with the `var`? I guess so because of copies would be made before mutating, but just want to check.

WireDomain/Sources/WireDomain/Event Decryption/MLSMessageDecryptor.swift
```diff
@@ -0,0 +1,168 @@
...
+            domain: conversationID.domain
+        ) else {
+            WireLogger.mls.error(
+                "failed to add mls message: conversation not found in db"
+            )
+
+            throw MLSMessageDecryptorError.conversationNotFound
+        }
+
+        guard let mlsGroupID = await conversationLocalStore.mlsGroupID(
```
johnxnguyen: Isn't this accessible via `mlsConversation`?

WireDomain/Sources/WireDomain/Event Decryption/MLSMessageDecryptor.swift
```diff
@@ -0,0 +1,168 @@
...
+            for: mlsConversation
+        ) else {
+            WireLogger.mls.error(
+                "failed to add mls message: missing MLS group ID"
+            )
+
+            throw MLSMessageDecryptorError.missingMLSGroupID
+        }
+
+        guard await conversationLocalStore.isConversationMLSReady(
```
johnxnguyen: Also this? It seems like we are switching to the local store too frequently when it could all be done in a single access.

WireDomain/Sources/WireDomain/Event Decryption/MLSMessageDecryptor.swift
```diff
@@ -0,0 +1,168 @@
...
+
+            case .proposal(let commitDelay):
+                let scheduledDate = (date ?? Date.now) + TimeInterval(commitDelay)
+
+                await conversationLocalStore.storeConversation(
+                    commitPendingProposalDate: scheduledDate,
+                    conversation: mlsConversation
+                )
+
+                mlsService.commitPendingProposalsIfNeeded()
```
johnxnguyen: It would be nice if the `MLSMessageDecryptor` only did decryption and not logic on the decrypted messages. What do you think about pulling this up to the event processor?

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationMLSMessageAddEventProcessor.swift
```diff
@@ -32,9 +32,27 @@ protocol ConversationMLSMessageAddEventProcessorProtocol {
...
+    let repository: any MessageRepositoryProtocol
+
+    func processEvent(_ event: ConversationMLSMessageAddEvent) async throws {
+        let message = event.message
+        let conversationID = event.conversationID
+        let senderID = event.senderID
+        let subconversation = event.subconversation
+        let date = event.timestamp
+
+        let decryptedMessages = event.decryptedMessages.map {
```
johnxnguyen: It may be good to do a sanity check that there are some decrypted messages.

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProteusMessageAddEventProcessor.swift
```diff
@@ -32,9 +32,33 @@ protocol ConversationProteusMessageAddEventProcessorProtocol {
...
+        let conversationID = event.conversationID
+        let messageContent = event.message
+        let messageExternalData = event.externalData
+        let messageSenderClientID = event.messageSenderClientID
+        let messageRecipientClientID = event.messageRecipientClientID
+        let timestamp = event.timestamp
+
+        // Message should be decrypted see `ProteusEventDecryptor`
+        guard let decryptedMessage = messageContent.decryptedMessage else {
+            return
```
johnxnguyen: Perhaps we can log or something to surface that we reached a bad state.

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
```diff
@@ -311,7 +335,7 @@ public final class ConversationLocalStore: ConversationLocalStoreProtocol {
 
     public init(
         context: NSManagedObjectContext,
-        mlsService: MLSServiceInterface,
+        mlsService: any MLSServiceInterface,
```
johnxnguyen: Is this being used?

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationRepository.swift
```diff
@@ -194,6 +195,7 @@ public final class ConversationRepository: ConversationRepositoryProtocol {
     private let conversationsLocalStore: any ConversationLocalStoreProtocol
     private let userRepository: any UserRepositoryProtocol
     private let teamRepository: any TeamRepositoryProtocol
+    private let userClientsRepository: any UserClientsRepositoryProtocol
```
johnxnguyen: Is this being used?

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
```diff
@@ -363,24 +490,625 @@ public final class MessageLocalStore: MessageLocalStoreProtocol {
...
+        _ message: GenericMessage,
+        content: GenericMessage.OneOf_Content,
+        in conversation: ZMConversation,
+        senderID: UUID,
+        senderDomain: String,
+        senderClientID: String?,
+        logAttributes: LogAttributes,
+        date: Date?
+    ) async {
+        await context.perform { [self] in
```
johnxnguyen: I wouldn't expect to see processing inside the store, rather only actual message insertions. Some of these message types are processed and discarded, which I feel like should be done in the event processor level. What do you think?

WireDomain/Sources/WireDomain/Repositories/Message/MessageRepository.swift
```diff
@@ -22,37 +22,131 @@ import WireDataModel
...
             conversationID: conversationID,
             conversationDomain: conversationDomain
         )
     }
+
+    public func addMessage(
+        _ messageType: MessageType
+    ) async {
+        switch messageType {
+        case .mls(let decryptedMessages, let conversationID, let conversationDomain, let senderID, let senderDomain, let date):
```
johnxnguyen: These could be packaged into a struct:

```swift
case .mls(let message):
```

WireDomain/Sources/WireDomain/Event Decryption/MLSMessageDecryptor.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2164/commits/b31805a437cf70da6bd6c7ebba537c3bc769a76e

WireDomain/Sources/WireDomain/Event Decryption/MLSMessageDecryptor.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2164/commits/b31805a437cf70da6bd6c7ebba537c3bc769a76e

WireDomain/Sources/WireDomain/Event Decryption/MLSMessageDecryptor.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2164/commits/c08ed17ebb783810d1b65e98d08654d10a6aaf59, revisited the current implementation

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
jullianm: done in https://github.com/wireapp/wire-ios/commit/c08ed17ebb783810d1b65e98d08654d10a6aaf59, revisited the current implementation

WireDomain/Sources/WireDomain/Repositories/Message/MessageRepository.swift
jullianm: no longer exists after revisiting the implementation (in done in https://github.com/wireapp/wire-ios/commit/c08ed17ebb783810d1b65e98d08654d10a6aaf59, revisited the current implementation)

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationRepository.swift
jullianm: no, removed in https://github.com/wireapp/wire-ios/pull/2164/commits/6b004359cdfeb2215764967ca051b578a1238a7a

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
jullianm: yes

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationMLSMessageAddEventProcessor.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2164/commits/a93bff7d790887461b464caa9c2f0d90678f17ff

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProteusMessageAddEventProcessor.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2164/commits/a93bff7d790887461b464caa9c2f0d90678f17ff

WireDomain/Sources/WireDomain/Event Decryption/MLSMessageDecryptor.swift
jullianm: Although I see your point, I'm not sure that makes sense in that specific case since we're iterating over the decryption results which is an enum with two tightly coupled cases `message` (the decrypted message) and `proposal` (the code here).  We need to decrypt the MLS message here but since it is an enum (I think) it makes more sense to deal with the other case at the same time. 
What I did though is grouping this logic into a single method from the local store since we already use the mls service in that store. https://github.com/wireapp/wire-ios/pull/2164/commits/c08ed17ebb783810d1b65e98d08654d10a6aaf59#diff-508a6af2dc08e794560553f4f59fd1d749c0a36db25bd806d3d25fcd7a619b70

WireDomain/Sources/WireDomain/Event Decryption/MLSMessageDecryptor.swift
jullianm: yes, but it needs to be retrieved from a context (else we'll have a CoreData thread violation) and only our local stores can access the context.

WireDomain/Sources/WireDomain/Event Decryption/MLSMessageDecryptor.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2164/commits/6dee7ff5535a7025a7c8448af1f611d02a6254f2

WireAPI/Sources/WireAPI/Models/Messaging/MessageContent.swift
jullianm: yes

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationMLSMessageAddEventProcessor.swift
```diff
@@ -32,9 +34,133 @@ protocol ConversationMLSMessageAddEventProcessorProtocol {
...
+        senderID: UserID,
+        date: Date?
+    ) async {
+        guard let conversation = await conversationLocalStore.fetchConversation(
+            id: conversationID.uuid,
+            domain: conversationID.domain
+        ) else {
+            return WireLogger.proteus.error(
+                "failed to add MLS message: conversation not found in db"
+            )
```
johnxnguyen: I'm wondering if thIs should actually throw an error. If there's no conversation, then it seems like a pretty serious error.

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationMLSMessageAddEventProcessor.swift
```diff
@@ -32,9 +34,133 @@ protocol ConversationMLSMessageAddEventProcessorProtocol {
...
+            genericMessage: genericMessage,
+            date: date ?? .now
+        )
+
+        await conversationLocalStore.addParticipant(
+            participantID: senderID.uuid,
+            participantDomain: senderID.domain,
+            in: conversation,
+            date: date?.addingTimeInterval(-0.01) ?? .now
+        )
```
johnxnguyen: I wonder why this is needed... can we receive messages from someone who we don't think is a participant yet?

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProtobufMessageProcessor.swift
```diff
@@ -0,0 +1,174 @@
...
+        WireLogger.eventProcessing.debug("Processing message", attributes: logAttributes)
+
+        // Message content types: https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/20545866/Messages
+        switch content {
+        case .lastRead:
+
+            await messageLocalStore.updateLastReadMessageTimestamp(
+                message.lastRead,
+                in: conversation
+            )
```
johnxnguyen: This modifies the conversation, so I would expect to see this method in the conversation local store:

```swift
// Something like...
await conversationLocalStore.updateLastReadTimestamp(
  conversationID: message.lastRead.conversationID,
  lastReadTimestamp: message.lastRead.timestamp
)
```

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProtobufMessageProcessor.swift
```diff
@@ -0,0 +1,174 @@
...
+
+        case .cleared:
+
+            await messageLocalStore.updateClearedMessageTimestamp(
+                message.cleared,
+                in: conversation
+            )
```
johnxnguyen: Similar to the last read, this I think would be a conversation local store method.

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProtobufMessageProcessor.swift
```diff
@@ -0,0 +1,174 @@
...
+
+        case .calling, .availability:
+
+            // cases not handled
+            break
+
+        default:
+
+            await messageLocalStore.addTextMessage(
```
johnxnguyen: The default case doesn't necessarily mean it's a text message. This is especially true for any new message type that is added and falls into the default case. I know it's a lot of work, but it'd be good to also have an exhaustive switch where it's clear how each message type is handled. This means cleaving apart the various ways we handle different generic messages from inside the message local store.

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
```diff
@@ -17,18 +17,154 @@
...
+    /// - Parameters:
+    ///     - conversation: The conversation to add the message to.
+    ///     - senderID: The message sender id.
+    ///     - logAttributes: Attributes to add to the log.
+
+    func canAddMessage(
+        conversation: ZMConversation,
+        senderID: UUID,
+        logAttributes: LogAttributes
+    ) async -> Bool
```
johnxnguyen: Does this necessarily have to be part of the protocol? Could it not be an internal check that, when fails, results in an error beIng thrown?

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
```diff
@@ -71,10 +210,337 @@ public final class MessageLocalStore: MessageLocalStoreProtocol {
...
+            }
+
+            editedClientMessage.updateCategoryCache()
+            editedClientMessage.markAsSent()
+        }
+    }
+
     // MARK: - Private
 
+    private func shouldAddMessage(to conversation: ZMConversation, date: Date) -> Bool {
```
johnxnguyen: This seems to have some overlap with `canAddMessage`. Can they be combined?

johnxnguyen (commented):
It's quite hard to give a thorough and deep review considering how large the changes are. I know that most of the changes are porting over legacy code to new components, and I think that's a good first step. 

Once everything can start to run, I'd like to do a second pass through everything to see how we can polish and correct some of the legacy that has been brought over, some of which relate to the comments that I have left.

But overall, great job tackling this mammoth task!

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationMLSMessageAddEventProcessor.swift
jullianm: yes it makes sense, fixed in https://github.com/wireapp/wire-ios/pull/2164/commits/90da7addc7e950bb179e65ed21fe6583b88d9f4c

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationMLSMessageAddEventProcessor.swift
jullianm: it seems that if the sender is not part of the conversation, it means we're out of sync, in that case we need to add the participant to the conversation, I renamed the method and added some doc comment so it is more clear: https://github.com/wireapp/wire-ios/pull/2164/commits/90da7addc7e950bb179e65ed21fe6583b88d9f4c

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProtobufMessageProcessor.swift
jullianm: yes you're right, fixed in https://github.com/wireapp/wire-ios/pull/2164/commits/90da7addc7e950bb179e65ed21fe6583b88d9f4c

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProtobufMessageProcessor.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2164/commits/90da7addc7e950bb179e65ed21fe6583b88d9f4c thx

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationMLSMessageAddEventProcessor.swift
```diff
@@ -31,10 +33,138 @@ protocol ConversationMLSMessageAddEventProcessorProtocol {
...
+    ) async throws {
+        guard let conversation = await conversationLocalStore.fetchConversation(
+            id: conversationID.uuid,
+            domain: conversationID.domain
+        ) else {
+            throw Failure.mlsConversationNotFound
+        }
+
+        let logAttributes: LogAttributes = [
+            .messageType: "conversation.mls-message-add",
```
netbe, line 83: don't we have a constant (enum) to use here

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationMLSMessageAddEventProcessor.swift
```diff
@@ -31,10 +33,138 @@ protocol ConversationMLSMessageAddEventProcessorProtocol {
...
+            .conversationId: conversationID.uuid.safeForLoggingDescription
+        ]
+
+        // Ensure is self conversation, sender is self user and conversation is not read-only
+        guard await messageLocalStore.canAddMessage(
+            conversation: conversation,
+            senderID: senderID.uuid,
+            logAttributes: logAttributes
+        ) else {
+            return
```
netbe: should we throw? or log

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProteusMessageAddEventProcessor.swift
```diff
@@ -32,9 +33,150 @@ protocol ConversationProteusMessageAddEventProcessorProtocol {
...
+            id: conversationID.uuid,
+            domain: conversationID.domain
+        ) else {
+            return WireLogger.proteus.error(
+                "failed to add proteus message: conversation not found in db"
+            )
+        }
+
+        let logAttributes: LogAttributes = [
+            .messageType: "conversation.otr-message-add",
```
netbe, line 67: same here

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProteusMessageAddEventProcessor.swift
```diff
@@ -32,9 +33,150 @@ protocol ConversationProteusMessageAddEventProcessorProtocol {
...
+            .conversationId: conversationID.uuid.safeForLoggingDescription
+        ]
+
+        // Ensure is self conversation, sender is self user and conversation is not read-only
+        guard await messageLocalStore.canAddMessage(
+            conversation: conversation,
+            senderID: senderID.uuid,
+            logAttributes: logAttributes
+        ) else {
+            return
```
netbe: throw or log

netbe (commented):
Looks good, I left a couple of comments before approving

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProtobufMessageProcessor.swift
jullianm: I revisited the implementation in https://github.com/wireapp/wire-ios/pull/2164/commits/244f4d59aa504d7edb51a223ddcf618e4d691b75 so we're more transparent on what is done in here for each message content type. however there is definitely still some room for improvement in terms of refactoring specifically getting rid of all the generic message helper methods (that gives a lot of opaqueness) so we treat each content in a more straightforward way. This involves manipulating code that is still used by legacy code. I think it's worth creating a follow-up ticket on this quite large topic that we could treat later on. What do you think ? @johnxnguyen  @netbe

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
jullianm: I think so yes because we don't directly call the message local store right after, we're doing some work before that, and I think we don't want to this extra work if in the end we cannot add the message. that's why we make this check early on.

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
jullianm: removed in https://github.com/wireapp/wire-ios/commit/244f4d59aa504d7edb51a223ddcf618e4d691b75

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationMLSMessageAddEventProcessor.swift
jullianm: you mean in legacy code ? we could use `ZMUpdateEvent.eventTypeString(for: .conversationMLSMessageAdd)` part of WireTransport but I don't think we want to do so

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProteusMessageAddEventProcessor.swift
jullianm: we have `ZMUpdateEvent.eventTypeString(for: .conversationOtrMessageAdd)` from legacy code but I don't think we want to use it, it is legacy code from WireTransport

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationMLSMessageAddEventProcessor.swift
jullianm: yes there is already a log in the local store but I moved it up to the processor here, it makes more sense: https://github.com/wireapp/wire-ios/pull/2164/commits/0937b4263848015e220cf070900c62f272c4f7ca

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProteusMessageAddEventProcessor.swift
jullianm: yes there is already a log in the local store but I moved it up to the processor here, it makes more sense: https://github.com/wireapp/wire-ios/pull/2164/commits/0937b4263848015e220cf070900c62f272c4f7ca

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProtobufMessageProcessor.swift
johnxnguyen: Yes let's revise in a separate PR.

johnxnguyen (approved):

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationMLSMessageAddEventProcessor.swift
netbe: not a blocker for now, but I would expect to rely on a exhaustive list of values for messageType instead of any String. So porting an enum to either WireAPI or WireDomain

netbe (approved):

PR #2175, opened 2024-11-20
refactor: process conversation rename event - WPB-10177

jullianm:
WPB-10177

### Key points

Specifically, this PR is about porting the existing implementation of the `ConversationRename` event.

In addition: 

- Redundant message creation code was removed from the `ConversationRepository` now that we have a dedicated `MessageRepository`
- Implementation had to be consequently a bit revisited so we can make use of the `MessageRepository` where we need to.
- No longer necessary code was removed

### Testing

- [x] Testing dedicated conversation repository method is correctly invoked by the processor
- [x] Testing conversation name is correctly updated
- [x] Add use cases tests for message creation 
- [x] Fix existing tests

### Checklist

---


WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationRepository.swift
```diff
@@ -530,21 +582,41 @@ public final class ConversationRepository: ConversationRepositoryProtocol {
...
+        senderDomain: String?,
+        date: Date,
+        removedUsers: Set<UserID>,
+        reason: ConversationMemberLeaveReason
     ) async {
-        await conversationsLocalStore.addSystemMessage(
-            message,
-            to: conversation
+        var systemMessageType: MessageType = switch reason {
+        case .userDeleted, .left:
```
netbe:
```suggestion
        case .userDeleted, .userLeft:
```

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationRepository.swift
```diff
@@ -530,21 +582,41 @@ public final class ConversationRepository: ConversationRepositoryProtocol {
...
+            .teamMemberRemoved(
+                member: (senderID, senderDomain),
+                date: date
+            )
+        case .removed:
```
netbe:
```suggestion
        case .userRemoved:
```

netbe (approved):
LGTM, minor suggestions

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
```diff
@@ -398,10 +385,24 @@ public final class ConversationLocalStore: ConversationLocalStoreProtocol {
...
     ) async throws {
         typealias UserAndRole = (user: ZMUser, role: Role?)
 
+        let localConversation = await fetchConversation(
+            id: conversation.id,
+            domain: conversation.domain
+        )
+
+        guard let localConversation else {
+            return WireLogger.eventProcessing.error(
```
samwyndham, line 398: Personally I find this pattern (returning void by calling something that returns void) super confusing. If I look at `WireLogger.eventProcessing.error` without knowing about our WireLogger I assume we are actually throwing an error then see that there is no `throw`. Is this a common pattern? I've only seen it at Wire.

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationRepository.swift
```diff
@@ -170,14 +170,22 @@ public protocol ConversationRepositoryProtocol {
...
+    ///     - senderID: The user ID.
+    ///     - senderDomain: The user domain.
+    ///     - date: The date the conversation name was updated.
 
-    func addSystemMessage(
-        _ message: SystemMessage,
-        to conversation: ZMConversation
+    func updateConversationName(
+        newName: String,
+        conversationID: UUID,
```
samwyndham, line 184: Optional: We do have a lot of ID (UUID), domain (String?) pairs. They are almost a `QualifiedID` but not quite. I wonder if there is an improvement we can make here to save some boilerplate.

I also wonder if it is worth representing a conversation id+domain differently from a sender id+domain to prevent accidentally passing incorrect values.

I would consider introducing the following kinds of types:

```swift
struct ConversationID {
    id: UUID
    domain: String?
}
```

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
```diff
@@ -365,24 +440,41 @@ public final class MessageLocalStore: MessageLocalStoreProtocol {
...
     }
 
-    // swiftlint:disable:next todo_requires_jira_link
-    // TODO: Use UserLocalStore when related PRs are merged.
+    private func fetchUsers(
+        userIDs: [(id: UUID, domain: String?)]
+    ) async -> Set<ZMUser> {
+        await withTaskGroup(of: ZMUser.self) { taskGroup in
+            for userID in userIDs {
+                taskGroup.addTask { [self] in
```
samwyndham: Not a request for change but just a thought:

I think there is probably unnecessary overhead with awaiting, locking in core data etc here. To avoid this we I think we should have a `UserLocalStore.fetchOrCreateUsers(userIDs: [(id: UUID, domain: String?)]) -> [ZMUser]` and have everything done within a single perform.

When wrapping core data it is probably more performant that our methods accept arrays of values I guess.

samwyndham (approved):
Nice work!

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationRepository.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2175/commits/ceef40d35b39388b3a83ca4ef2aa175ba3b750db

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationRepository.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2175/commits/ceef40d35b39388b3a83ca4ef2aa175ba3b750db

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
jullianm: I don't think it's a common pattern, I use it to have a single line but it might be a bit confusing, might be interesting to add this to the convention discussion. I'll add it to the document thx

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationRepository.swift
jullianm: I totally agree, this is a discussion we had with John and there will be soon a dedicated ticket for that, specifically for handling id and non optional domain under a same QualifiedID object (all users should have a domain even it's the local one). As part of this future ticket,  I will also harmonize the rest

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
jullianm: Good point, done in https://github.com/wireapp/wire-ios/pull/2175/commits/ceef40d35b39388b3a83ca4ef2aa175ba3b750db thx!

PR #2180, opened 2024-11-21
feat: Change a key for the dependency cache - WPB-14445

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14445" title="WPB-14445" target="_blank">BugWPB-14445</a>  [iOS]  Change a key for dependencies cache
  </td></table>
  <br />
 

### Issue

We create a key for the cache when building the app as: `hash(Package.resolved)-ENABLE_DATADOG`
But for tests we can hardcode the value of `ENABLE_DATADOG`, and this will allow the cache to be reused.

### Testing

---

### Checklist

---


.github/workflows/_reusable_run_tests.yml
```diff
@@ -168,7 +168,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ${{ env.PACKAGES_DIR }}
-          key: ${{ runner.os }}-swiftpm-project-${{ hashFiles('**/*.xcworkspace/xcshareddata/swiftpm/Package.resolved') }}
+          key: ${{ runner.os }}-swiftpm-project-${{ hashFiles('**/*.xcworkspace/xcshareddata/swiftpm/Package.resolved') }}-true
```
caldrian: Looks good, however, I'm not sure if it should be `false` instead.

caldrian (approved):

.github/workflows/_reusable_run_tests.yml
KaterinaWire: I thought the default `ENABLE_DATADOG `value is true, but I think it shouldn't be a problem to use cache with DD.

netbe: for running tests I would say false also. ENABLE_DATADOG is not defined here

netbe (changes requested):
let's change it to false as it makes more sense, no DD for tests at all

netbe (approved):
looks good

PR #2194, opened 2024-11-25
refactor: process conversation typing event - WPB-10178

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10178" title="WPB-10178" target="_blank">TaskWPB-10178</a>  [iOS] Process ConversationTypingEvent
  </td></table>
  <br />
 

### Key points

Specifically, this PR is about porting the existing implementation of the `ConversationTyping` event.

The legacy code for processing that event was fully part of WireSyncEngine, some objects (`TypingUsers`) were moved from WireSyncEngine to WireDataModel for backward compatibility so when we plug our new quick sync mechanism UI don't break. (see comment for specifics)

### Testing

- [x] testing `ConversationTypingUsersTimeout` read/write methods (add, remove, contains..)
- [x] testing `ConversationTypingEventProcessor` repo and local store methods invocations. 
- [x] testing `ConversationRepository` a notification with the typing users data is correctly sent.

### Checklist

---


WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationTypingUsersTimeout.swift
```diff
@@ -0,0 +1,118 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireDataModel
+
+/// Keeping track of typing users timeouts
+final class ConversationTypingUsersTimeout: NSObject, ZMTimerClient {
```
jullianm: Please check this page [here](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/29065437/Use+case+Typing+indicator) for specifics regarding the business logic related to typing users timeouts.

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
```diff
@@ -461,6 +481,54 @@ public final class ConversationLocalStore: ConversationLocalStoreProtocol {
...
+            if let conversation = uiContext.object(with: conversationID) as? ZMConversation {
+                
+                let users = usersID.compactMap {
+                    uiContext.object(with: $0) as? ZMUser
+                }
+                
+                uiContext.typingUsers?.update(
+                    typingUsers: Set(users),
+                    in: conversation
+                )
```
jullianm, line 604: The underlying implementation (`typingUsers?.update..`) was not accessible from WireDomain (it was part of WireSyncEngine) so in this PR, `TypingUsers` object and the related NSManagedObjectContext extension were moved from WireSyncEngine to WireDataModel.

At the app level, the `ConversationInputBarViewController` relies on this code. Now that it is accessible in WireDataModel, we can reuse the same methods from WireDomain so when we plug our new quick sync mechanism, UI won't break (although it should probably be refactored as well eventually)

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationTypingEventProcessor.swift
```diff
@@ -26,15 +27,112 @@ protocol ConversationTypingEventProcessorProtocol {
...
-    func processEvent(_ event: ConversationTypingEvent) async throws
+    func processEvent(_ event: ConversationTypingEvent) async
 
 }
 
+public struct TypingUsersInfo {
+    let users: Set<NSManagedObjectID>
+    let conversationID: NSManagedObjectID
+}
+
```
johnxnguyen: I suggest to move this to a separate file and add doc comments.

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationTypingEventProcessor.swift
```diff
@@ -26,15 +27,112 @@ protocol ConversationTypingEventProcessorProtocol {
...
+            id: conversationID.uuid,
+            domain: conversationID.domain
+        )
+        
+        // Since we'll be manipulating managed object IDs in `ConversationTypingUsersTimeout`
+        // we need to make sure we have valid, consistent IDs for the user and conversation.
+        conversationLocalStore.obtainPermanentIDs(
+            user: user,
+            conversation: conversation
+        )
```
johnxnguyen: Is this something that can be internal to the stores?

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
johnxnguyen: This definitely needs refactoring... we shouldn't store anything on the managed object context. Do you think it's something we can do in this PR or it's too big?

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationTypingUsersTimeout.swift
```diff
@@ -0,0 +1,118 @@
...
+        return Set(keysToRemove.map(\.conversationObjectID))
+    }
+    
+    /// Updates next prune date and fire a timer at that specific date.
+    /// When timer fires, the provided callback will be called - see `timerDidFire`
+
+    func updateExpirationIfNeeded() {
+        guard
+            let date = firstTimeout,
+            date != nextPruneDate
```
netbe: why do we take the minimum timeout date here?

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationTypingUsersTimeout.swift
```diff
@@ -0,0 +1,118 @@
...
+        nextPruneDate = date
+    }
+    
+    
+    func timerDidFire(_ timer: ZMTimer!) {
+        guard timer === expirationTimer else {
+            return
+        }
+        
+        Task { [timerFiredCallback] in
```
netbe: do we need to hold on this task and cancel it if we're getting called again?

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
```diff
@@ -461,6 +481,54 @@ public final class ConversationLocalStore: ConversationLocalStoreProtocol {
...
         }
     }
+    
+    public func updateTypingUsers(
+        conversationID: NSManagedObjectID,
+        usersID: Set<NSManagedObjectID>
+    ) async {
+        
+        // Switching to main context
+        let uiContext = await context.perform { [context] in
```
netbe: it's not completely clear to me why this needs to be done on viewContext, any impact if we change the context?

wire-ios-data-model/Source/Model/Conversation/TypingUsers/NSManagedObjectContext+TypingUsers.swift
```diff
@@ -23,7 +23,7 @@ extension NSManagedObjectContext {
 
     private static let TypingUsersKey = "ZMTypingUsers"
 
-    var typingUsers: TypingUsers? {
+    public var typingUsers: TypingUsers? {
```
netbe, line 26: if TypingUsers just contain NSManagedObjectIDs I don't see why we force it to be on a specific context

netbe (commented):
@jullianm I left a couple of questions before approving

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
jullianm: I agree but I think it's too big and sensitive to be done in this PR. It means revisiting some legacy code at the app level to introduce a new way of communication between WireDomain and the app layer (which we haven't fully figured out yet) while still making sure legacy code don't break. It could lead to introducing bugs, worse breaking the feature since it still relies a lot on WireSyncEngine. 

I feel like we're still not at the stage yet to start revisiting legacy app layer code so it fits in with our new sync mechanism. 

A temporary solution that came to mind was to get rid of these intermediary objects and operations and directly observe the notification from the UI component (this is currently done in WireSyncEngine), after all it could be as straightforward as posting a notification and observing it (but again that means revisiting the app layer code)

What do you think ?

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationTypingEventProcessor.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2194/commits/2c63329f5e7ac219626fb9c1d9bec5983dc6d329 thx

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationTypingEventProcessor.swift
jullianm: unfortunately no, we need to call that method from the processor right before manipulating IDs in the `ConversationTypingUsersTimeout` object which is not owned by the store

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationTypingUsersTimeout.swift
jullianm: that's the closest timeout date in the dictionary entries, we ensure we always deal with the closest timeout date so the timer fires at the date and we don't end up with a user that seems to be always typing.

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationTypingUsersTimeout.swift
jullianm: if you mean holding a reference to that task and cancelling it when it is called again, I would say we don't need to do so since it is a short lived task designed to be called when the timer fires (so it could be called multiple times)

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
jullianm: typing users (ZMUser) managed objects are directly consumed by the UI component but at the local store level we don't need to use specific context indeed. 

Legacy code is not very straightforward, this could be simplified but as mentioned above this involves changing a little bit the app layer logic (specifically the `ConversationInputBarViewController`)
 
`view.typingUsers = conversation.typingUsers `
--> gets the typing users object nested in the main context user info dictionary -->  in that object it gets the users object ids  and fetch the ZMUsers from the main context 

I'm not sure why we need all that stuff, I could start a refactoring of this legacy code so it also fits in with our new quick sync mechanism but I'm not 100% sure we won't break anything by doing so and also I'm not sure it's in the scope of that PR. I don't mind doing it, just want your opinion before starting to work on that.

What do you think ? also mentioning @johnxnguyen  for the discussion

wire-ios-data-model/Source/Model/Conversation/TypingUsers/NSManagedObjectContext+TypingUsers.swift
jullianm: yes I agree, it doesn't make sense we don't need to use a specific context at the local store level

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
jullianm: also mentioning @netbe for the discussion

johnxnguyen: Makes sense, let's create a ticket though to come back to this.

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
johnxnguyen: We definitely need a way to send events to the UI, not just in this case but in others too (e.g when call events are processed we need to publish events to some observer in the UI).

Let's defer this for a separate PR, but as mentioned in another comment, please create a ticket so that we can revisit this.

johnxnguyen (approved):

netbe (approved):

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationLocalStore.swift
netbe: @jullianm create the ticket and we can tackle that later

wire-ios-data-model/Source/Model/Conversation/TypingUsers/NSManagedObjectContext+TypingUsers.swift
jullianm: fixed in https://github.com/wireapp/wire-ios/pull/2194/commits/75a2e7c46a506ea4b18825e2cfa19c33f8a5e470

PR #2204, opened 2024-11-25
fix: make WireLogger concurrency safe - WPB-11890

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11890" title="WPB-11890" target="_blank">TaskWPB-11890</a>  [iOS] Make `WireLogger` swift concurrency safe
  </td></table>

### Issue

This PR fixes some warnings regarding concurrency safety.

---

### Checklist

---


wire-ios-system/Source/Logging/SystemLogger.swift
```diff
@@ -90,5 +92,3 @@ struct SystemLogger: LoggerProtocol {
         }
     }
 }
-
-private var loggers: [String: OSLog] = [:]
```
caldrian, line 94: This array doesn't seem to be set or modified anywhere, so I deleted it.

wire-ios-system/Source/Logging/WireLogger.swift
```diff
@@ -16,12 +16,18 @@
...
+    public static func initialize(loggers: [any LoggerProtocol]) {
+        guard provider == nil else {
+            assertionFailure("WireLogger.initialize called more than once")
+            return
+        }
+
+        provider = AggregatedLogger(loggers: loggers)
+    }
+
+    private static nonisolated(unsafe) var provider: (any LoggerProtocol)!
```
caldrian: The `provider` value is now injected via `initialize(loggers:)`.
This func should only be called once.
Still, I don't consider it necessary to use any custom logic to prevent. So I put a simple `guard` + `assertionFailure` there.

Currently this code is called by `WireAnalytics.setup()` which is called in
`UIApplicationDelegate.willFinishLaunchingWithOptions` and in the two app extensions.

wire-ios-system/Source/Logging/WireLogger.swift
```diff
@@ -92,8 +98,4 @@ public struct WireLogger: LoggerProtocol {
     public static var logFiles: [URL] {
         provider.logFiles
     }
-
-    public static func addLogger(_ logger: LoggerProtocol) {
-        provider.addLogger(logger)
-    }
```
caldrian: By removing this the type can conform to `Sendable`.

wire-ios-system/Tests/ZMLogTests.swift
```diff
@@ -362,6 +363,7 @@ extension ZMLogTests {
         Thread.sleep(forTimeInterval: 0.2)
     }
 
+    @MainActor
```
caldrian, line 366: ... fixes warnings regarding `waitForExpectations`.

.github/pull_request_template.md
```diff
@@ -11,7 +11,6 @@ _Optional: add details about technical approach, solutions etc._
 
 _Optional: reference dependencies to other pull requests etc._
 
-
```
caldrian: I've been deleting this line for every PR, so I decided to just do it once and for all here.
Please let me know if you prefer this to be reverted or done in another PR.

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
```diff
@@ -16,5 +16,31 @@
...
+import WireSystem
+
 /// Namespace for analytics tools.
-public enum WireAnalytics {}
+public enum WireAnalytics {
+
+    private static var isSetUp = false
+
+    public static func setup() {
+        guard !isSetUp else {
```
samwyndham, line 34: I think we need to have some kind of lock here. I think we had issues without it in the notification service right @netbe ?

samwyndham (commented):
Nice work. Much better than my proposal I think to introduce a global actor. Just left a comment about `isSetUp` and threading.

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
netbe: yes the NSE can be reused multiple times, we should guarantee that the logger is added only once

netbe: so it should be thread safe too

netbe: I would keep the lock @caldrian unless there's something I am missing.

wire-ios-system/Source/Logging/CocoaLumberjackLogger.swift
```diff
@@ -20,46 +20,46 @@ import CocoaLumberjackSwift
 import Foundation
 
 /// Logger to write logs to fileSystem via CocoaLumberjack
-final class CocoaLumberjackLogger: LoggerProtocol {
+public final class CocoaLumberjackLogger: LoggerProtocol {
```
netbe: why do we need the classes to be public?

netbe (changes requested):
I think we should keep the lock to avoid multiple initialization of the logger

wire-ios-system/Source/Logging/CocoaLumberjackLogger.swift
caldrian: I haven't moved the file but just used it from WireCommonComponents.
I'll give it a try to move the file and revert this.

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
caldrian: Thanks, added a lock.

wire-ios-system/Source/Logging/WireLogger.swift
```diff
@@ -34,46 +40,46 @@ public struct WireLogger: LoggerProtocol {
     // MARK: - LoggerProtocol
 
     public var logFiles: [URL] {
-        Self.provider.logFiles
+        Self.provider?.logFiles ?? []
```
caldrian, line 43: I made `provider` optional, because for unit tests the `initialize` method is not called.
Do you agree with this approach or do you suggest something different?

netbe (approved):

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
```diff
@@ -16,5 +16,39 @@
...
+    private static var isSetUp = false
+
+    public static func setup() {
+        // Adding a lock here since some app extension might execute this setup in the same process as the main app.
+        // https://stackoverflow.com/a/62674277
+        isSetUpLock.lock()
+        defer { isSetUpLock.unlock() }
+
+        guard !isSetUp else {
+            assertionFailure("WireAnalytics.setup() called more than once")
```
samwyndham, line 35: I guess if https://stackoverflow.com/a/62674277 happens we might not want to assert here... It is normal behavior.

samwyndham (approved):

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
caldrian: True, let's change it if it happens during debugging.
For release builds this statement will have no effect.

samwyndham: FYI it happened to me in a debug build. I will remove the assertion.

PR #2207, opened 2024-11-26
fix: Crash unwrapping `MainCoordinator.splitViewController` - WPB-14510

samwyndham:
### Issue

This PR attempts to fix a crash when accessing `MainMainCoordinator.splitViewController`. I haven't managed to reproduce the crash but I have instead fixed a strong references cycle which I think is the underlying reason for the crash.

#### Details

`MainCoordinator.splitViewController` is an explicitly unwrapped weak property. `ZClientViewController` maintains a strong reference to both the `ZClientViewController` and `MainCoordinator`. When `ZClientViewController` is deallocated, `MainCoordinator` should also be deallocated. However when the `Lock With Passcode` option is enabled, and a user returns to the app and the password / faceID login app Lock Screen is shown, `ZClientViewController` is deallocated but it's `MainCoordinator` is **not** deallocated. This leaves `MainCoordinator.splitViewController` in a dangerous state. The following screenshot shows that after returning to the app twice we now have 3 `MainCoordinator` instances - there should be only one.

Screenshot 2024-11-26 at 14 15 38

This PR tackles the crash by fixing the underlying strong reference cycle. The reference cycle is:

```
MainCoordinator -> ConversationViewControllerBuilder -> SelfProfileViewControllerBuilder -> AnyMainCoordinator -> MainCoordinator
```

My fix removes the reference from `SelfProfileViewControllerBuilder` to `AnyMainCoordinator` and instead requires users of `SelfProfileViewControllerBuilder` to pass in a main coordinator reference when building.

### Testing

1. Enable  `Settings -> Options -> Lock With Passcode`
3. Wait more than 60 seconds
4. Return to the app
5. There should be no crash
6. Repeat from different screens

---

### Checklist

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/SelfProfileViewControllerBuilder.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/SelfProfileViewControllerBuilder.swift
@@ -26,7 +26,6 @@ final class SelfProfileViewControllerBuilder: SelfProfileViewControllerBuilderPr
     var selfUser: SettingsSelfUser
     var userRightInterfaceType: UserRightInterface.Type
     var userSession: UserSession
-    var mainCoordinator: AnyMainCoordinator!
     var accountSelector: AccountSelector?
     var trackingManager: TrackingManager?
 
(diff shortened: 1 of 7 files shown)


wire-ios/Wire-iOS/Sources/UserInterface/SelfProfile/SelfProfileViewControllerBuilder.swift
samwyndham, line 1: This is the main change

KaterinaWire (approved):

netbe (approved):

PR #2233, opened 2024-12-02
fix: crash calleventstatus - WPB-10449

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10449" title="WPB-10449" target="_blank">BugWPB-10449</a>  Crash on ZMTimer - CallEventStatus
  </td></table>
  <br />
 

### Issue

There is a crash (see ticket) involving CallEventStatus that happens from time to time. Looking at the usage of CallEventStatus, it appears we don't really need this at all. This PR cleans up unused code in order to resolve the crash.

### Testing

* [x] running app and check for regression
---

### Checklist


wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1037,12 +1037,13 @@ extension ZMUserSession: ZMSyncStateDelegate {
         WireLogger.updateEvent.info("process pending call events")
         Task {
             do {
+                // why not processing only the call events (should be stored here?)
```
netbe: question maybe to @johnxnguyen

wire-ios-sync-engine/Source/Notifications/VoIPPushManager.swift
```diff
@@ -182,7 +182,8 @@ public final class VoIPPushManager: NSObject, PKPushRegistryDelegate {
         payload: [AnyHashable: Any],
         completion: @escaping () -> Void
     ) {
-        Self.logger.trace("process voIP push, payload: \(payload)")
+        // TODO: check this is dead code right?
```
netbe: from my understanding this is never used, we always use the APNS code, is that right?

johnxnguyen: The device will only receive APNS pushes and wake the NSE, but the NSE can create a "fake voip push" when it detects a call event and pass it to the main app.

johnxnguyen: From the point of view of the app, it looks like a real VoIP push, but in reality it came from the NSE. So we need this code.

wire-ios-sync-engine/Source/Synchronization/ZMOperatonLoop+Background.swift
```diff
@@ -33,22 +33,6 @@ private enum PushNotificationType: String {
...
 public extension ZMOperationLoop {
 
-    @objc(fetchEventsFromPushChannelPayload:completionHandler:)
-    func fetchEvents(fromPushChannelPayload payload: [AnyHashable: Any], completionHandler: @escaping () -> Void) {
-        guard let nonce = messageNonce(fromPushChannelData: payload) else {
-            return completionHandler()
-        }
-
-        pushNotificationStatus.fetch(eventId: nonce, completionHandler: {
-            self.callEventStatus.waitForCallEventProcessingToComplete { [weak self] in
```
netbe, line 43: this is dead code where callEventStatus was used

wire-ios-sync-engine/Source/Synchronization/Strategies/CallingRequestStrategy.swift
```diff
@@ -313,10 +310,7 @@ public final class CallingRequestStrategy: AbstractRequestStrategy, ZMSingleRequ
             clientId: clientId
         )
 
-        callEventStatus.scheduledCallEventForProcessing()
-        callCenter?.processCallEvent(callEvent, completionHandler: { [weak self] in
-            self?.callEventStatus.finishedProcessingCallEvent()
```
netbe, line 318: as fara as I saw, there is no need anymore to use callEventStatus

wire-ios-sync-engine/Source/Notifications/VoIPPushManager.swift
```diff
@@ -20,33 +20,31 @@ import avs
 import CallKit
 import Foundation
 import PushKit
+import WireLogging
 
 public protocol VoIPPushManagerDelegate: AnyObject {
 
-    func processIncomingRealVoIPPush(payload: [AnyHashable: Any], completion: @escaping () -> Void)
```
netbe, line 26: this is pre iOS 15, when voip push was used

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
netbe: I guess this is an optim that could be done in another PR, to directly process callEvents instead of buffered ones. I don't see why we would need to process all events in a method `processPendingCallEvents`

samwyndham (commented):
Did a first pass and looks good but I guess we need to resolve @johnxnguyens points first and maybe add back some of the code that was removed?

wire-ios-sync-engine/Source/Notifications/VoIPPushManager.swift
netbe: @samwyndham you're referring to this?

samwyndham: Yes that is what I'm referring to

johnxnguyen: I was mistaken here which we clarified in a call a while ago.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: I'm not sure why it does this... but it does more than just process events, it also decrypts and stores first the buffered events that may have be received in the push channel. If you were going to optimize this, you need to decrypt and store the buffered events, and then process only the stored calling events.

johnxnguyen (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
netbe: added to a ticket for it

wire-ios-sync-engine/Source/Notifications/VoIPPushManager.swift
netbe: cc @samwyndham

samwyndham (approved):
Nice work 🛁!

PR #1707, opened 2024-07-16
feat: refactor CI tests - WPB-11406

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11406" title="WPB-11406" target="_blank">StoryWPB-11406</a>  Refactor CI tests workflow
  </td></table>
  <br />
 

### Issue

This PR allows to test all related frameworks when touching a framework on a PR. Previously it was not running tests of frameworks having the dependency.

It also moves the develop tests that are run when a PR is merged to be run nightly at 21:00.

* add testplans to all targets
* use fastlane to run tests
* create dependency graph in ruby
* run nightly all tests for release branches
* fastlane action to run all tests locally: `bundle exec fastlane select_frameworks`

#### TODO

* [x] add test plans for new frameworks introduced: WireAPI...
* [x] run tests for new frameworks introduced: WireAPI...
* [x] handle warnings report

### Testing

Running [PR](https://github.com/wireapp/wire-ios/pull/1913) as a test

---

### Checklist

- [ ] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [ ] Description is filled and free of optional paragraphs.
- [ ] Adds/updates automated tests.

PR #2235, opened 2024-12-03
feat: refactor CI tests - WPB-11406

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11406" title="WPB-11406" target="_blank">StoryWPB-11406</a>  Refactor CI tests workflow
  </td></table>
  <br />
 

### Issue

This is a cherry-pick of  #1707. It's a manual cherry-pick since it contains mostly workflows.

### DIFF
--- a/.github/workflows/test_pr_changes.yml
+++ b/.github/workflows/test_pr_changes.yml
@@ -26,41 +26,21 @@ jobs:
     runs-on: ubuntu-latest
     # Set job outputs to values from filter step
     outputs:
-      wire-ios: ${{ steps.filter.outputs.wire-ios == 'true' || steps.filter.outputs.carthage == 'true' }}
-      wire-ui: ${{ steps.filter.outputs.wire-ui }}
-      wire-ios-sync-engine: ${{ steps.filter.outputs.wire-ios-sync-engine }}
-      wire-ios-data-model: ${{ steps.filter.outputs.wire-ios-data-model }}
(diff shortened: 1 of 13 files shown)


caldrian (approved):

.github/workflows/test_pr_changes.yml
```diff
@@ -26,41 +26,21 @@ jobs:
...
-      wire-ios-testing: ${{ steps.filter.outputs.wire-ios-testing }}
-      wire-foundation: ${{ steps.filter.outputs.wire-foundation }}
-      scripts: ${{ steps.filter.outputs.scripts }}
+      folders: ${{ steps.filter.outputs.changes }}
 
     steps:
       - uses: actions/checkout@v4
         with:
-          lfs: 'true'
+          lfs: 'false' # as long as we use lfs only for snapshotTesting it should not be useful here
```
samwyndham, line 34: I don't understand this change. Don't we run snapshot tests using this workflow?

samwyndham (approved):

.github/workflows/test_pr_changes.yml
netbe: @samwyndham this part of the job only detect changes, so we don't need the references here. the _reusable_run_tests.yml will run them with lfs.

You could argue that we miss trigger tests if only references change, but i don't think it's a common use case, it will always involve some code change

PR #2237, opened 2024-12-03
feat: Certificate pinning and proxy mode in `WireAPI` - WPB-10450

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10450" title="WPB-10450" target="_blank">TaskWPB-10450</a>  [iOS] Support certificate pinning
  </td></table>
  <br />
 

### Issue

This PR adds a certificate pinning implementation to the `WireAPI` module. It is part of quick sync refactor. It doesn't attempt to use this new implementation in the app.

The code is essentially duplicated from `WireTransport` with some modifications. Main things to look at are:

- `ServerTrustValidator` basically combines the existing `ServerCertificateTrust.swift` with security key generation code from `TrustData.swift`. See my comments for specific changes.
- `PinnedKey.swift` will replace `TrustData.swift`. It is just a basic struct without code for reading the data from file, and generating security keys.

### Testing

Run the automated tests.

---

### Checklist


WireAPI/Sources/WireAPI/Network/NetworkService/NetworkService.swift
```diff
@@ -122,27 +124,23 @@ extension NetworkService: URLSessionDataDelegate {
     public func urlSession(
```
samwyndham, line 125: I've updated this to use the `async` version of the delegate method as it needs to use `async` code.

WireAPI/Sources/WireAPI/Network/NetworkService/NetworkService.swift
```diff
@@ -122,27 +124,23 @@ extension NetworkService: URLSessionDataDelegate {
...
+        URLAuthenticationChallenge
+    ) async -> (URLSession.AuthChallengeDisposition, URLCredential?) {
         let protectionSpace = challenge.protectionSpace
 
-        guard protectionSpace.authenticationMethod == NSURLAuthenticationMethodServerTrust else {
-            completionHandler(.performDefaultHandling, challenge.proposedCredential)
-            return
-        }
+        if protectionSpace.authenticationMethod == NSURLAuthenticationMethodServerTrust {
+            guard let trust = challenge.protectionSpace.serverTrust else { fatalError("Missing server trust") }
```
samwyndham: If `serverTrust` is nil this is an Apple breaking its contract. Currently we pass this from obj-c to swift and it is implicitly force unwrapped and not crashing. Therefore, I choosing to fatal in the migration.

WireAPI/Sources/WireAPI/Models/Backend/ServerTrustValidator.swift
```diff
@@ -0,0 +1,132 @@
...
+    /// - Parameter trust: The `SecTrust` of the server.
+    /// - Parameter host: The host of the server.
+    /// - Throws: An error if server certificate should not be trusted.
+    /// - Note: If no pinned keys are found for the `host`, the server certificate is trusted.
+
+    func validate(trust: SecTrust, host: String) async throws {
+        let matches = pinnedKeys.filter { $0.matches(host: host) }
+
+        // If no keys are pinned for `host`, we trust the server certificate
+        guard !matches.isEmpty else { return }
```
samwyndham: This is a change of behavior from  the `ServerCertificateTrust`. `ServerCertificateTrust` first verifies that a server certificate is valid before performing this check. However as we are only supposed to be validating certs from servers we control, I don't think this makes sense. Therefore here we first check that this is a server we care about before doing any validation.

WireAPI/Sources/WireAPI/Models/Backend/ServerTrustValidator.swift
```diff
@@ -0,0 +1,132 @@
...
+
+    // MARK: - Private
+
+    private static func verifyServerCertificateTrusted(_ serverTrust: SecTrust) async throws {
+        try await withCheckedThrowingContinuation { continuation in
+            // `SecTrustEvaluateAsyncWithError` requires the completion queue to be the same as the queue on which it
+            // is called.
+            let queue = DispatchQueue.global()
+            queue.async {
+                SecTrustEvaluateAsyncWithError(serverTrust, queue) { _, success, error in
```
samwyndham, line 66: I've chosen to use this API instead of the blocking version `SecTrustEvaluateWithError` that we are currently using. According to Apples docs, It is possible that  `SecTrustEvaluateWithError` makes a network request which I assume means we might unnecessarily block the `URLSessionDelegate` queue unnecessarily. Because of this change the public API of this class is now `async`.

WireAPI/Sources/WireAPI/Models/Backend/BackendEnvironment.swift
```diff
@@ -0,0 +1,47 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public struct BackendEnvironment {
```
johnxnguyen, line 23: It would be nice to have a doc comment here too.

WireAPI/Sources/WireAPI/Models/Backend/ServerTrustValidator.swift
```diff
@@ -0,0 +1,112 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+@preconcurrency import Security
+
+struct ServerTrustValidator: Sendable {
+
+    enum Error: Swift.Error {
```
johnxnguyen: My suggestion is to call the error `Failure` so we don't have to write `Swift.Error` everywhere in scope.

WireAPI/Sources/WireAPI/Models/Backend/ServerTrustValidator.swift
```diff
@@ -0,0 +1,112 @@
...
+
+    /// Verifies the server `trust` for the given `host`.
+    ///
+    /// - Parameter trust: The `SecTrust` of the server.
+    /// - Parameter host: The host of the server.
+    /// - Throws: An error if server certificate should not be trusted.
+    /// - Note: If no pinned keys are found for the `host`, the server certificate is trusted.
+
+    func validate(trust: SecTrust, host: String) async throws {
+        let matches = pinnedKeys.filter { $0.matches(host: host) }
```
johnxnguyen: Suggestion: `matchingKeys`

WireAPI/Sources/WireAPI/Models/Backend/ServerTrustValidator.swift
```diff
@@ -0,0 +1,112 @@
...
+        guard !matches.isEmpty else { return }
+
+        try await Self.verifyServerCertificateTrusted(trust)
+
+        let publicKey = try Self.publicKeyAssociatedWithServerTrust(trust)
+
+        let publicKeys = try matches.map { try Self.certificateKey(for: $0) }
+        guard publicKeys.contains(publicKey) else {
+            throw Error.noMatchingPublicKey
+        }
```
johnxnguyen: So we first verify the servers certificate, then get the public key from the certificate, then check that key is known locally?

WireAPI/Sources/WireAPI/Models/Backend/ServerTrustValidator.swift
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    // MARK: - Private
+
+    private static func verifyServerCertificateTrusted(_ serverTrust: SecTrust) async throws {
+        try await withCheckedThrowingContinuation { continuation in
+            // `SecTrustEvaluateAsyncWithError` requires the completion queue to be the same as the queue on which it
+            // is called.
+            let queue = DispatchQueue.global()
+            queue.async {
```
johnxnguyen, line 65: Is this fine to mix with async await?

WireAPI/Sources/WireAPI/Models/Backend/ServerTrustValidator.swift
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    private static func certificateKey(for value: PinnedKey) throws -> SecKey {
+        guard let certificate = SecCertificateCreateWithData(nil, value.key as CFData) else {
+            throw Error.cannotLoadCertificate
+        }
+
+        guard let publicKey = SecCertificateCopyKey(certificate) else {
+            throw Error.cannotRetrievePublicKey
+        }
```
johnxnguyen: I guess a difference with the existing implementation is that is currently done once, but here (I assume) it's done for every challenge. Or is it that there is only a single challenge when first contacting the server? So it practice it's not called often?

WireAPI/Sources/WireAPI/Network/NetworkService/NetworkService.swift
johnxnguyen: Maybe worth a comment here explaining the fatal error.

WireAPI/Sources/WireAPI/Models/Backend/ServerTrustValidator.swift
samwyndham: Yes that is correct

WireAPI/Sources/WireAPI/Models/Backend/ServerTrustValidator.swift
samwyndham: Yes it is as far as I know. What would not be fine is if we put in DispatchGroup/Semaphore to block the Tasks thread.

WireAPI/Sources/WireAPI/Models/Backend/ServerTrustValidator.swift
samwyndham: Correct it would be done for every challenge. But in practice (limited testing) I've only noticed a single challenge when first contacting the server. The alternatives I can think of are storing on the `PinnedKey` (not ideal due to breaking auto Sendable conformance), or caching it somewhere. I'm not sure it is worth it unless we notice a problem in practice.

samwyndham: I've thought about it a bit more. I'm actually going to move `SecKey` into `PinnedKey` and mark as @preconcurrency. This has the following advantages:

1. Fixes possible issues you mention
2. Allows for us to crash early on initialization if that is what we want - I think this is probably a good idea for the bundled data.
3. Potentially allows us to check this at compile time once we start using macros

WireAPI/Sources/WireAPI/Network/NetworkService/NetworkService.swift
```diff
@@ -122,27 +125,26 @@ extension NetworkService: URLSessionDataDelegate {
...
-            protectionSpace.serverTrust != nil,
-            true // TODO: [WPB-10450] support certificate pinning
-        else {
-            completionHandler(.cancelAuthenticationChallenge, nil)
-            return
+        if protectionSpace.authenticationMethod == NSURLAuthenticationMethodServerTrust {
+            guard let trust = challenge.protectionSpace.serverTrust else {
+                // If this is missing it is Apple breaking its API contract so crash.
+                fatalError("Missing server trust")
+            }
```
jullianm, line 137: I would chain the conditions since we **must** have a `serverTrust` value when the authentication method is NSURLAuthenticationMethodServerTrust. And fallback to the else statement if we don't (which should not happen unless Apple API breaks). Or we **actually** want the app to crash if their API breaks ?

```
if protectionSpace.authenticationMethod == NSURLAuthenticationMethodServerTrust, 
    let trust = challenge.protectionSpace.serverTrust { ... } else { ... }
```

WireAPI/Tests/WireAPITests/Backend/PinnedKeyTests.swift
```diff
@@ -0,0 +1,44 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+
+@testable import WireAPI
+
+final class PinnedKeyTests: XCTestCase {
+
+    func testMatches() throws {
```
jullianm: Shouldn't we use the test naming convention here ? we can also omit the throws

WireAPI/Sources/WireAPI/Models/Backend/ServerTrustValidator.swift
```diff
@@ -0,0 +1,112 @@
...
+
+        let publicKeys = try matchingKeys.map { try Self.certificateKey(for: $0) }
+        guard publicKeys.contains(publicKey) else {
+            throw Failure.noMatchingPublicKey
+        }
+    }
+
+    // MARK: - Private
+
+    private static func verifyServerCertificateTrusted(_ serverTrust: SecTrust) async throws {
```
jullianm, line 60: I see some methods are static inside the `ServerTrustValidator` do they need to be static ?

WireAPI/Tests/WireAPITests/Authentication/AuthenticationManagerTests.swift
```diff
@@ -30,7 +30,10 @@ final class AuthenticationManagerTests: XCTestCase {
     override func setUpWithError() throws {
         cookieStorage = MockCookieStorageProtocol()
         backendURL = try XCTUnwrap(URL(string: "https://www.example.com"))
-        let networkService = NetworkService(baseURL: backendURL)
+        let networkService = NetworkService(
+            baseURL: backendURL,
+            serverTrustValidator: ServerTrustValidator(pinnedKeys: [])
```
jullianm, line 35: Is there a way to test the ServerTrustValidator as well ? I'm assuming no, just wondering

jullianm (commented):
Looks good, left a few comment before approving

WireAPI/Tests/WireAPITests/Authentication/AuthenticationManagerTests.swift
samwyndham: Yes, I've just added a tests for that

samwyndham: I think I will also test `BackendService`

WireAPI/Sources/WireAPI/Models/Backend/ServerTrustValidator.swift
samwyndham: I personally prefer using `static` where possible because if I do that I know it is likely fairly independent and not relying on instance variables. I almost like to default to `static`. Do you know any cons to this?

WireAPI/Sources/WireAPI/Network/NetworkService/NetworkService.swift
samwyndham: Personally I think we want the app to crash because it is a security issue. If the method is `NSURLAuthenticationMethodServerTrust` we must validate - we shouldn't allow the request to continue. 

In our current implementation we would crash and this doesn't seem to be a problem so I'm just choosing to keep that behavior.

WireAPI/Tests/WireAPITests/Backend/PinnedKeyTests.swift
samwyndham: I will fix the `throws`. I can't think of a better name without modifying the test itself. Do you have any suggestions?

WireAPI/Sources/WireAPI/Network/NetworkService/NetworkService.swift
jullianm: makes sense 👍

WireAPI/Tests/WireAPITests/Backend/PinnedKeyTests.swift
jullianm: The test is about asserting the sut matches or not some host, so it could be something like `testHostMatches` or something similar

WireAPI/Sources/WireAPI/Models/Backend/ServerTrustValidator.swift
jullianm: I was afraid this could lead to some confusion since this method can be called on the type itself and it seems here we only call it from an instance method, not a big fan of mixing up instance and type methods when we don't need to tbh however I see your point and it makes sense too.

jullianm (approved):
looks good

WireAPI/Sources/WireAPI/Models/Backend/ServerTrustValidator.swift
samwyndham: I see. I won't change now but lets discuss in one of the Architecture forums.

WireAPI/Tests/WireAPITests/Backend/PinnedKeyTests.swift
samwyndham: 👍 I've renamed to `testMatchesHost` to better match the method name `matches(host: String)`. Thanks!

WireAPI/Sources/WireAPI/Models/Backend/PinnedKey.swift
```diff
@@ -0,0 +1,74 @@
...
+
+    public enum Host: Sendable {
+        case endsWith(String)
+        case equals(String)
+    }
+
+    let key: SecKey
+    let hosts: [Host]
+
+    public init(key: SecKey, hosts: [Host]) {
```
samwyndham, line 38: This is a temporary method for gluing things together under pressure of release. I need to think a bit more about how and where to map the data from `backend.bundle` but expect this method to be deleted in a follow up PR and instead favor throwing version.

wire-ios-transport/Source/Public/BackendEnvironment.swift
```diff
@@ -111,13 +111,16 @@ public extension EnvironmentType {
 
 public final class BackendEnvironment: NSObject {
     public let title: String
+    public let trustData: [TrustData]
```
samwyndham, line 114: In a follow up PR I plan to revert these changes. (see my early comment.)

wire-ios-transport/Source/Public/TrustData.swift
samwyndham, line 1: In a follow up PR I plan to revert these changes. (see my early comment.)

WireAPI/Tests/WireAPITests/Backend/ServerTrustValidatorTests.swift
```diff
@@ -0,0 +1,74 @@
...
+
+final class ServerTrustValidatorTests: XCTestCase {
+
+    override func setUpWithError() throws {
+        // Put setup code here. This method is called before the invocation of each test method in the class.
+    }
+
+    override func tearDownWithError() throws {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+    }
```
johnxnguyen: These should be deleted.

johnxnguyen (approved):
Looks good, nice work

WireAPI/Tests/WireAPITests/Backend/ServerTrustValidatorTests.swift
samwyndham: Done in [34a2d38](https://github.com/wireapp/wire-ios/pull/2237/commits/34a2d388c15645227d859804b6fb4a9a6574bf41)

samwyndham:
🤦 I accidentally merged `https://github.com/wireapp/wire-ios/pull/2260` into this instead of waiting for this PR to merge to develop.

PR #2244, opened 2024-12-04
feat: Calculate `defaultProtocol` for a user - WPB-14667

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14667" title="WPB-14667" target="_blank">BugWPB-14667</a>  [iOS] Default protocol if MLS is disabled
  </td></table>
  <br />
 

### Issue

We use `defaultProtocol` to create a conversation. If a team has a weird setup, it takes the wrong `defaultProtocol`.
Example:
- mls feature is **disabled**
- but the default protocol from the feature config is MLS 

In this example, the self client is not an MLS client, but is trying to create an MLS conversation and fails.

### Solution

If mls feature is disabled then the default protocol should be `proteus`. Otherwise, it should be taken from the feature config.

### Testing

Setup a team as:

```
"mls": {
        "config": {
            "allowedCipherSuites": [
                2
            ],
            "defaultCipherSuite": 2,
            "defaultProtocol": "mls",
            "protocolToggleUsers": [],
            "supportedProtocols": [
                "mls"
            ]
        },
        "lockStatus": "unlocked",
        "status": "disabled",
        "ttl": "unlimited"
    },

```
### Checklist

---


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
```diff
@@ -500,15 +500,20 @@ extension ConversationCreationController {
...
+        switch mlsFeature.config.defaultProtocol {
         case .proteus:
-            .proteus
+            return .proteus
         case .mls:
-            .mls
+            return .mls
         case .mixed:
-            .mixed
+            return .mixed
```
johnxnguyen: `.mixed` doesn't make sense as a default protocol. We had two separate message protocol types (one with mixed, one without) for the reason that one represents what the current protocol of a conversation is (which can be mixed) and the other represents what different protocols there are (without mixed).

The value should be either proteus or mls.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
```diff
@@ -500,15 +500,20 @@ extension ConversationCreationController {
...
-extension Feature.MLS.Config.MessageProtocol {
-    var toMessageProtocol: MessageProtocol {
-        switch self {
+extension UserSession {
+
+    var defaultProtocol: MessageProtocol {
+        guard mlsFeature.isEnabled else {
+            return .proteus
+        }
+        switch mlsFeature.config.defaultProtocol {
```
jullianm, line 505: I think here you can remove the switch and directly return `mlsFeature.config.defaultProtocol.toMessageProtocol`

jullianm (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
KaterinaWire: Unfortunately, I can't. 
mlsFeature.config.defaultProtocol is `Feature.MLS.Config.MessageProtocol `type. But I would like to have a result as `MessageProtocol` enum type.

jullianm: isn't that the case here ?

```
extension Feature.MLS.Config.MessageProtocol {
    var toMessageProtocol: MessageProtocol {
        switch self {
        case .proteus:
            .proteus
        case .mls:
            .mls
        case .mixed:
            .mixed
        }
    }
}
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
```diff
@@ -500,15 +500,20 @@ extension ConversationCreationController {
     }
 }
 
-extension Feature.MLS.Config.MessageProtocol {
-    var toMessageProtocol: MessageProtocol {
```
KaterinaWire, line 504: @jullianm I removed `toMessageProtocol`, and have only one `defaultProtocol` helper property

jullianm: Ok, sorry I thought this was legacy code

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
KaterinaWire: Changed it to Proteus

johnxnguyen (approved):

PR #2248, opened 2024-12-04
fix: new event decoding monitor - WPB-8700

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8700" title="WPB-8700" target="_blank">EpicWPB-8700</a>  [iOS] Rewrite quick sync
  </td></table>
  <br />
 

### Issue

The event decoding monitor (to catch decoding failures of new update events) is failing because the wrong event envelope is being decoded.

The solution is to decode the correct envelope, which requires making an internal payload public. We'll make it internal again after the monitor is removed.

### Testing

Receive events via the push channel and check data dog for decoding errors.

---

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/Synchronization/ZMOperationLoop+PushChannel.swift
+++ b/wire-ios-sync-engine/Source/Synchronization/ZMOperationLoop+PushChannel.swift
@@ -31,7 +31,7 @@ extension ZMOperationLoop: ZMPushChannelConsumer {
             // fix it. Once we're sure it works, we should remove this.
             do {
                 let decoder = JSONDecoder.defaultDecoder
-                _ = try decoder.decode(UpdateEventEnvelope.self, from: data)
+                _ = try decoder.decode(UpdateEventEnvelopeV0.self, from: data)
             } catch {
                 WireLogger.updateEvent.error("failed to decode 'UpdateEventEnvelope': \(error)")
             }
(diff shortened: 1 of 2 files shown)


jullianm (approved):

wire-ios-sync-engine/Source/Synchronization/ZMOperationLoop+PushChannel.swift
```diff
@@ -31,7 +31,7 @@ extension ZMOperationLoop: ZMPushChannelConsumer {
             // fix it. Once we're sure it works, we should remove this.
             do {
                 let decoder = JSONDecoder.defaultDecoder
-                _ = try decoder.decode(UpdateEventEnvelope.self, from: data)
+                _ = try decoder.decode(UpdateEventEnvelopeV0.self, from: data)
```
netbe, line 34: will this be enough in case of error, to know which event failed to parse and to fix the event parsing?

netbe (approved):
looks good, but left a question

wire-ios-sync-engine/Source/Synchronization/ZMOperationLoop+PushChannel.swift
johnxnguyen: yes, because internally there is a `UpdateEventDecodingProxy` that tells which event failed

PR #2249, opened 2024-12-04
refactor: replace http client with api service - WPB-14606

jullianm:
WPB-14606

### Key points

This PR replaces usage of `httpClient` by `APIService` in our multiple API components (builders and versioned APIs) from WireAPI package.

- ConnectionsAPI
- UsersAPI
- FeatureConfigAPI
- UserPropertyAPI
- ...

### Testing

- [x] Use APIServiceSnapshoptHelper instead of APIServiceHelper in all the API tests
- [x] Fix all tests accordingly
- [x] Ensure we also verify the generated request when asserting the response model.

### Checklist


WireAPI/Sources/WireAPI/APIs/FeatureConfigsAPI/FeatureConfigsAPIV1.swift
```diff
@@ -25,19 +25,28 @@ class FeatureConfigsAPIV1: FeatureConfigsAPIV0 {
     }
 
     override func getFeatureConfigs() async throws -> [FeatureConfig] {
-        let request = HTTPRequest(
-            path: "\(pathPrefix)/feature-configs",
-            method: .get
-        )
+        let components = URLComponents(string: resourcePath)
+
+        guard let url = components?.url else {
```
samwyndham: This is kind of repeated everywhere (excluding the error type). I think this should be de-duplicated. For example, by introducing a throwing way of getting the URL.

I also think the pattern of having different errors `ConnectionsAPIError.invalidURL` , `FeatureConfigsAPIError.invalidURL` that mean essentially the same thing can be a kind of code smell as it prevents generic handling of such errors.

WireAPI/Tests/WireAPITests/APIs/ConnectionsAPI/__Snapshots__/ConnectionsAPITests/testGetConnectionsRequest.request-0-v0.txt
```diff
@@ -1,4 +1,5 @@
-- HTTPRequest
-  - path: "/list-connections/"
-  - method: post
-  - body: {"size":500}
+curl \
+	--request POST \
+	--header "Content-Type: application/json" \
+	--data "{\"size\":500}" \
+	"/list-connections/"
```
samwyndham, line 5: This format is a choice right? I guess we don't want to keep changing it as otherwise we loose safety it provides. Are we moving to the `curl` style in general?

WireAPI/Tests/WireAPITests/APIs/FeatureConfigsAPI/FeatureConfigsAPITests.swift
```diff
@@ -437,11 +425,11 @@ extension FeatureConfigsAPITests {
...
-
-}
+// private extension APIVersion {
+//
+//    func buildAPI(client: any APIServiceProtocol) -> any FeatureConfigsAPI {
+//        let builder = FeatureConfigsAPIBuilder(apiService: apiService)
+//        return builder.makeAPI(for: self)
+//    }
+//
+// }
```
samwyndham: I guess this should be deleted

samwyndham (approved):
Nice work!

WireAPI/Sources/WireAPI/APIs/FeatureConfigsAPI/FeatureConfigsAPIV1.swift
johnxnguyen: There is a ticket to factor out common api errors and re-use them. As for the repetitive guard let, either we could add a convenience throwing method, or perhaps we can integrate it into the url request builder.

WireAPI/Tests/WireAPITests/APIs/ConnectionsAPI/__Snapshots__/ConnectionsAPITests/testGetConnectionsRequest.request-0-v0.txt
johnxnguyen: the curl style is coming from the snapshot library, and I think we should stick with that instead of the custom one we had before.

WireAPI/Tests/WireAPITests/APIs/FeatureConfigsAPI/FeatureConfigsAPITests.swift
```diff
@@ -139,60 +142,45 @@ final class FeatureConfigsAPITests: XCTestCase {
...
-                for try await result in taskGroup {
-                    // Then
-                    XCTAssertEqual(
-                        result,
-                        Scaffolding.featureConfigsV4
-                    )
-                }
-            }
+        // Then
+        try await apiSnapshotHelper.verifyRequest(for: supportedVersions, apiService: apiService) { sut in
```
johnxnguyen, line 154: These tests are adding more snapshots when asserting the response handling, is that intentional?

WireAPI/Tests/WireAPITests/APIs/ConnectionsAPI/__Snapshots__/ConnectionsAPITests/testGetConnectionsRequest.request-0-v0.txt
samwyndham: I think one thing that would have been nice process wise would be to have replaced these in a separate PR. That way they would have provided a guarantee during this refactor. 

I see that as a team we quite often update snapshot tests during refactors and in doing so lose some of the purpose of the tests.

WireAPI/Tests/WireAPITests/APIs/FeatureConfigsAPI/FeatureConfigsAPITests.swift
jullianm: yes, because we want to assert the model response but also ensure the request is correct for a given response. let's say the request changed for a specific version, without it we wouldn't know something changed, no?

WireAPI/Sources/WireAPI/APIs/FeatureConfigsAPI/FeatureConfigsAPIV1.swift
jullianm: makes sense, I'll fix that

WireAPI/Tests/WireAPITests/APIs/FeatureConfigsAPI/FeatureConfigsAPITests.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2249/commits/a4c9b508d43ee8d6c87dfdf0ad4952daa8f6ec78

WireAPI/Sources/WireAPI/APIs/FeatureConfigsAPI/FeatureConfigsAPIV1.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2249/commits/10b51780771639e68c672830917b4e4529c1eb43
We actually had already what we needed in URLRequestBuilder.

johnxnguyen (approved):

PR #2252, opened 2024-12-05
fix: CI tests - WPB-14822

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14822" title="WPB-14822" target="_blank">BugWPB-14822</a>  Nightly build failed
  </td></table>
  <br />
 

### Issue

- [x] Another issue with nightly tests is that when it runs in parallel, if one job (i.e. develop) fails the other (release branch) fails automatically.
- [x] Add a test target for WireCanvas (a schema without tests fails)
- [x] failure is not always reported
- [x] add command `bundle exec fastlane select_framework` to run only one scheme locally


samwyndham (approved):
Nice work!

caldrian (approved):

PR #2255, opened 2024-12-06
refactor: create WireAuthentication feature module - WPB-15768

johnxnguyen:
### Issue

This PR introduces a feature module to contain login and registration flows, in preparation for implementing the simplified login flow.

The module is comprised of 4 main targets:
- **WireAuthentication**: the public and root module that is used to assemble the module and obtain the root view(s). Depends on the api, ui, and logic targets and contains construction code only.
- **WireAuthenticationUI**:  Contains all view code (views + view models), and only depends on the api target. It's lightweight and quick to build.
- **WireAuthenticationLogic**: Contains all logic code (use case implementations) and depends on the api target plus any other logic targets it needs to implement use cases.
- **WireAuthenticationAPI**: Contains public api, such as use case protocols and models.

Some things to note:
- The ui target already contains some skeleton code that will need to be updated/implemented in future PRs. 
- The assembly target defines a basic dependency graph as a guide and uses `Needle` to resolve the graph.
- This code is not integrated at all in live code paths, it is intended as a starting point for the implementation of the new login flows.

---

### Checklist

---


WireAuthentication/WireAuthenticationAPI/Sources/Use cases/DetermineAuthenticationMethodUseCase.swift
```diff
@@ -0,0 +1,35 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public protocol DetermineAuthenticationMethodUseCaseProtocol {
+
+    @MainActor
+    func invoke(emailOrSSOCode: String) async -> AuthenticationMethod
```
johnxnguyen: I think it could be a good practice to always mark the invoke method as `@MainActor` because:

1. It creates a cleaner call side from the view model which should also be main actor isolated (i.e just create a `Task` and run the use case.
2. It makes it explicit to the implementation on which thread the use case is being called on and therefore the responsibility of the implantation to manage efficient concurrent execution.

WireAuthentication/WireAuthentication/Sources/WireAuthenticationAssembly.swift
```diff
@@ -0,0 +1,63 @@
...
+import WireAuthenticationAPI
+internal import WireAuthenticationUI
+internal import WireAuthenticationCore
+
+/// The entry point for the feature module.
+///
+/// To load the contents of the module, create an instance of
+/// the assembly, which you can then use to make required views.
+@MainActor
+public struct WireAuthenticationAssembly {
```
caldrian, line 26: Will there be more code than the assembly in the entry point target?
If not, isn't it a bit too much of an overhead to have a separate target?

johnxnguyen: I don't think there would be anything other than instantiating components, wiring them together and returning them. Where would be another place to put it if we don't have a separate target? It couldn't exist inside either the UI or Core targets because the assembly needs to combine components from both.

samwyndham (approved):
Looks good! I've only checked the rough structure as the rest will likely change but no issues for me merging this ASAP to start filling in the skeleton as none of this is currently in the production app.

KaterinaWire (approved):

PR #2257, opened 2024-12-06
feat: Hide "Invite More People" button for Personal users - WPB-14654

KaterinaWire:
### Issue

Invite More People" button for Personal users is a legacy function and we can remove it.

**Reasons:**
- Phone number login has been disabled (access to the address book is not relevant)
- This is a relic from the past
- Do not want the app to make users send message via phone number

### Testing

Login with a personal account -> click on "+" button
---

### Checklist

---


samwyndham (approved):
Nice! 🧹

wire-ios-sync-engine/Source/UserSession/Search/SearchRequest.swift
```diff
@@ -26,41 +26,37 @@ public struct SearchOptions: OptionSet {
...
     public static let contacts = SearchOptions(rawValue: 1 << 0)
 
-    /// Users found in your address book.
-
-    public static let addressBook = SearchOptions(rawValue: 1 << 1)
-
     /// Users which are a member of the same team as you.
 
-    public static let teamMembers = SearchOptions(rawValue: 1 << 2)
+    public static let teamMembers = SearchOptions(rawValue: 1 << 1)
```
netbe, line 31: is it normal that the option change here?

netbe (approved):
looks good just left a question

wire-ios-sync-engine/Source/UserSession/Search/SearchRequest.swift
samwyndham: I was thinking the same. My thought was that if it is not persisted anywhere it should be fine. But I didn't check the project that thoroughly

KaterinaWire: Yes, we don't persist it.
I can refactor it and create an Enum for `SearchOption`.

```
public enum SearchOption {
    case contacts
    case teamMembers
    case excludeNonActiveTeamMembers
    case excludeNonActivePartners
    case directory
    case conversations
    case services
    case federated
    case localResultsOnly
}
```
And we will pass around `searchOptions: [SearchOption]`
Maybe in the next PR, so it's more clear
@netbe and @samwyndham  what do you think?

netbe: I think it was good to check this was not persisted. For now let's keep it like this, I don't have a strong opinion on changing it to enum

samwyndham: I think it is good to keep it as it is. An `OptionSet` seems a good solution to it's purpose.

PR #2260, opened 2024-12-09
feat: Proxy mode in `WireApi` - WPB-10447

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10447" title="WPB-10447" target="_blank">TaskWPB-10447</a>  [iOS] Support proxy mode
  </td></table>
  <br />
 

### Issue

This PR:
 
- Adds proxy mode to `WireAPI`. 
- Additionally deletes `NativePushChannel` which is unused.

In our current implementation of `ProxySettings` it is possible to have an invalid state when generating the proxy dictionary - see in the following if `needsAuthentication` is true, but the username / password is nil, this is an invalid state:

```swift
    func socks5Settings(proxyUsername: String?, proxyPassword: String?) -> [AnyHashable: Any]? {
        // ... 

        if let username = proxyUsername, let password = proxyPassword, needsAuthentication {
            proxyDictionary[kCFStreamPropertySOCKSUser] = username
            proxyDictionary[kCFStreamPropertySOCKSPassword] = password
        }
        return proxyDictionary
    }
```

If we were to correctly handle this within `WireAPI` it would mean `Assembly.init` should be failing.

To avoid this situation I have made  `ProxySettings` an `enum` so that the invalid state cannot be represented -  we would need to handle it in before initializing `Assembly.init` which I think is the better place.

### Testing

Run tests
---

### Checklist


WireAPI/Sources/WireAPI/Models/Backend/ProxySettings.swift
```diff
@@ -0,0 +1,60 @@
...
+
+public enum ProxySettings {
+
+    /// Settings for an unauthenticated proxy.
+
+    case unauthenticated(host: String, port: Int)
+
+    /// Settings for an authenticated proxy.
+
+    case authenticated(host: String, port: Int, username: String, password: String)
```
johnxnguyen, line 31: Wonderful!

johnxnguyen (approved):
I really enjoyed reviewing this PR: simple, concise, elegant. 💯

jullianm (approved):

PR #2263, opened 2024-12-09
fix: tests run on CI - WPB-14822

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14822" title="WPB-14822" target="_blank">BugWPB-14822</a>  Nightly build failed
  </td></table>
  <br />
 

### Issue

* [x] WireCanvas tests target should be iOS 17 (from project)
* [x] run nightly build only on weekdays
* [x] fix upload artifacts collision when running parallel job
* [x]  ~run schemas in dependency order instead of random: so from bottom to top (WirFoundation ->... Wire-iOS)~


fastlane/Fastfile
```diff
@@ -269,12 +271,26 @@ platform :ios do
...
+                failed_schemas << scheme
+                UI.error("Failed to test #{scheme}: #{e.message}")
+            end            
+        end
+                    
+        if failed_targets.any?
+            UI.important("The following schemas failed: #{failed_targets.join(', ')}")
+        else
+            UI.success("All tests passed successfully!")
         end
```
netbe: this will run the next schema if the current one fails: not sure yet if that's a good idea. 

i.e.: WireFoundation fails (either build or tests) then WireDataModel (which depends on Foundation) will still run

caldrian (approved):

fastlane/Fastfile
```diff
@@ -465,26 +466,56 @@ platform :ios do
...
-            buildlog_path: build.build_path,
-            result_bundle: true,
-            output_directory: "#{build.artifact_path(with_filename: false)}/tests/#{scheme}",
-            output_types: "junit",
-            cloned_source_packages_path: ENV['PACKAGES_DIR'],
-            disable_package_automatic_updates: true,
-            skip_package_dependencies_resolution: ENV["HAS_PACKAGES_CACHE_HIT"],
-            number_of_retries: 2
-        )
+        begin
```
netbe, line 469: split in 2, so we fail on build, but continue testing next schemes on test failures

.github/workflows/_reusable_run_tests.yml
```diff
@@ -119,7 +118,7 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v4
         with:
-          name: xcresults
+          name: XCResults for ${{ env.SANITIZED_BRANCH }} (${{ github.run_id }})
```
samwyndham: `run_id` takes account of workflow run but not run attempt. Rerunning a job (for example if the first run attempt failed) will overwrite the this. I think we should **additionally** add the workflow run attempt to the name - `github.run_attempt`

.github/workflows/_reusable_run_tests.yml
```diff
@@ -220,15 +226,21 @@ jobs:
...
+            SANITIZED_BRANCH=$(echo "${{ inputs.branch }}" | sed 's/[^a-zA-Z0-9._-]/_/g')
+            echo "SANITIZED_BRANCH=$SANITIZED_BRANCH" >> $GITHUB_ENV
+
       - name: Download tests results
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
         continue-on-error: true
         with:
-          name: test-reports
+          name: Test reports for ${{ env.SANITIZED_BRANCH }} (${{ github.run_id }})
```
samwyndham: optional: Consider extracting to a local [composite action](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action)

WireAPI/Tests/WireAPITests/Helpers/HTTPRequestSnapshotHelper.swift
```diff
@@ -40,17 +40,20 @@ struct HTTPRequestSnapshotHelper {
...
-            as: .dump,
-            named: resourceName,
-            file: file,
-            testName: function,
-            line: line
-        )
-
-        if let errorMessage {
-            XCTFail(errorMessage, file: file, line: line)
+        let recordEnabled: SnapshotTestingConfiguration.Record? = ProcessInfo.processInfo.environment["CI"] == "true" ? .never : nil
```
samwyndham: 👍

samwyndham (approved):

.github/workflows/_reusable_run_tests.yml
netbe: fixed

.github/workflows/_reusable_run_tests.yml
netbe: yeah, I started to look into that, I'll rework it a bit in another PRs

PR #2276, opened 2024-12-11
crashfix: move typing users related objects back to WireSyncEngine - no ticket

jullianm:
no ticket

### Issue

A crash related to typing users was **most likely** introduced in one of the quick sync related PR (ticket WPB-10178 PR #2194) where some objects were moved from WireSyncEngine to WireDataModel so they could be accessed by WireDomain.

This PR moves these files back to WireSyncEngine, just like it was before.

### Checklist


netbe (commented):
@jullianm how can moving files be responsible for the crash? I don't get it

jullianm:
@netbe, tbh I'm not entirely sure that my recent changes are responsible for this crash but my assumption is that something went wrong at some level when moving these files out of WireSyncEngine in one of my recent PR since we didn't have that crash before.

jullianm:
Closing, the crash occurred before my recent changes so this PR is irrelevant. I will investigate and open a different PR for the crash fix.

PR #2279, opened 2024-12-11
fix: crash typing when timer fires - WPB-15029

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15029" title="WPB-15029" target="_blank">BugWPB-15029</a>  [iOS] Crash related to a typing user
  </td></table>
  <br />
 

### Crash fix

Screenshot 2024-12-11 at 15 52 32

This crash occurs when the typing users timer fires and the conversation object cannot be retrieved locally either from the context or the persistent store, with the usage of this method `syncContext.object(with: $0)`, as per Apple doc:

> When this method returns a fault, Core Data makes no attempts to verify the existence of the underlying object in the persistent store. If the object doesn’t exist when the context tries to the fetch the object’s values, the framework throws an exception.

**Proposed solution**

Use the variant `syncContext.existingObject(with: $0)` to explicitly throw an error when the object is missing so we avoid the app crash.

> unlike [object(with:)](doc://com.apple.documentation/documentation/coredata/nsmanagedobjectcontext/1506197-object), this method never returns a fault. If the object doesn’t exist in both the context and the persistent store, the method throws an error.

**Disclaimer**: I was not able to reproduce the crash however the crash message, Apple's doc and some online resources seem to be converging in this direction.

### Testing

_Describe how to test._

---

### Checklist

---


netbe (approved):
Thanks, let's see if that solves it

caldrian (approved):
👏

PR #2281, opened 2024-12-11
feat: Create an MLS client at the end of the quick sync, if needed - WPB-14987

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14987" title="WPB-14987" target="_blank">BugWPB-14987</a>  [iOS] no MLS client when MLS is enabled (on the backend or for a team)
  </td></table>
  <br />
 

### Issue

1. We started testing on the backend that doesn't have MLS public keys, so no MLS teams.
2. We are logged in as proteus clients.
3. Backend added MLS public keys.
4. MLS was enabled for a team with MLS as default protocol

As a result, MLS client wasn't created (even after quick sync)

### Solution

At the end of the quick sync:
- fetch mls public keys
- create mls client if needed

### Testing

Look at the steps from the Issue.

---

### Checklist

---


wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -925,10 +925,25 @@ extension ZMUserSession: ZMSyncStateDelegate {
...
+                return (selfClient, hasRegisteredMLSClient)
+            }
 
-            WaitingGroupTask(context: syncContext) { [self] in
+            // If we discover that
+            // the MLS feature is enabled, there are MLS public keys on the backend and there is no registered MLS client,
+            // we should create one.
+            let needsToRegisterMLSClient = BackendInfo.isMLSEnabled && mlsFeature.isEnabled && !hasRegisteredMLSClient
+            if let selfClient, needsToRegisterMLSClient {
+                await createMLSClient(client: selfClient)
```
johnxnguyen: try to avoid passing a managed object (the self client). It seems that all that is needed are some ids.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1014,6 +1028,31 @@ extension ZMUserSession: ZMSyncStateDelegate {
...
+        }
+    }
+
+    private func createMLSClient(client: UserClient) async {
+        let mlsClientID = await syncContext.perform {
+            return MLSClientID(userClient: client)
+        }
+        guard let mlsClientID else {
+            fatalError("Needs to register MLS client but can't retrieve qualified client ID")
+        }
```
johnxnguyen: If you pass in the ids you could use the non-fallible init and thus avoid this fatal error here.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1014,6 +1028,31 @@ extension ZMUserSession: ZMSyncStateDelegate {
...
+        do {
+            try await self.coreCryptoProvider.initialiseMLSWithBasicCredentials(mlsClientID: mlsClientID)
+        } catch {
+            WireLogger.mls.error("Failed to initialise mls client: \(error)")
```
johnxnguyen, line 1058: This would not be a nice state to be in.... we should have an mls client but we couldn't create one. What do you think would be the appropriate response here? thoughts @netbe ?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
KaterinaWire: Change it to `qualifiedClientID`

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
KaterinaWire: Yes, removed

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
KaterinaWire: I talked with Jacob, and his opinion:
`I would expect you to log that error and then try it again after the next quick sync.`

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -925,28 +925,45 @@ extension ZMUserSession: ZMSyncStateDelegate {
...
+                return (selfClient?.qualifiedClientID, hasRegisteredMLSClient)
+            }
+
+            if hasRegisteredMLSClient {
+                await performsMLSClientUpdates()
+            } else {
+                // If we discover that
+                // there are MLS public keys on the backend, the MLS feature is enabled and there is no registered MLS client,
+                // we should create one.
+                let needsToRegisterMLSClient = BackendInfo.isMLSEnabled && mlsFeature.isEnabled
```
netbe, line 962: question: I am not sure to fully understand the difference here between BackendInfo.isMLSEnabled && mlsFeature.isEnabled
when isMLSEnabled set ? and why we check both?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
netbe: great

netbe (commented):
@KaterinaWire Should we try to add a test for this behaviour? Also left a question to fully understand the change

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: BackendInfo.isMLSEnabled indicates if the backend supports mls (i.e it has public keys). `misFeature.isEnabled` indicates If mls is enabled for the team. The user can only use mls if it's enabled for both.

KaterinaWire: Yes, just additional info to what John said:

We can create MLS client if it's enabled on the backend side (`GET /mls/public-keys` contains an entry with the key removal) and `mls feature` config is enabled. These 2 conditions must be true for us to create an MLS client.

KaterinaWire:
> @KaterinaWire Should we try to add a test for this behaviour? Also left a question to fully understand the change

Tests are added

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
netbe: ok thanks make sense, not blocking for now, but what do you think about renaming BackendInfo.isMLSEnabled by BackendInfo.mlsSupported ?

netbe (approved):
All good

jullianm (approved):

PR #2285, opened 2024-12-12
fix: Some flaky tests - WPB-15021

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15021" title="WPB-15021" target="_blank">BugWPB-15021</a>  [iOS] Fix some flaky tests
  </td></table>
  <br />
 

### Issue

This PR:

- Fixes a couple of flaky tests in `MockTransportSessionTeamEventsTests` using the magical `waitForAllGroupsToBeEmpty`. These were tests that I could reproduce flakiness locally by repeating the tests suite 100 times.
- Sets the minimum timeout to 1 second for most uses of `waitForExpectations(timeout:)` & `await fulfillment(of:timeout:)`. 

#### Rational

Sub 1 second is often plenty of time locally but can be problematic when running tests on CI where resources can be extremely limited. In cases where we don't set `expectation.isInverted = true` the waiting will finish as soon as the expectation is met meaning that generally there will be no change in behavior setting longer timeouts.

Considering our full test suite now takes around 2 hours to run, a timeout can be very frustrating. Personally I don't see a reason not to set a minimum timeout of 5 seconds but I feel there may be some pushback so keeping it to 1 second. Looking at data dog, in the last month we have only had timeouts in cases of 0.5 sections so hopefully 1 second should be sufficient.

#### Note

I have left a few sub-second timeouts as they are in cases where I'm not confident that behavior isn't dependent on it.

### Testing

Run the automated tests

### Checklist

### DIFF
--- a/WireAPI/Tests/WireAPITests/Network/PushChannel/WebSocketTests.swift
+++ b/WireAPI/Tests/WireAPITests/Network/PushChannel/WebSocketTests.swift
@@ -80,13 +80,13 @@ final class WebSocketTests: XCTestCase {
         }
 
         // Wait for iteration to be in progress
-        await fulfillment(of: [didReceiveMessage], timeout: 0.5)
+        await fulfillment(of: [didReceiveMessage], timeout: 1)
 
         // When
         sut.close()
 
         // Then the stream finished successfully
-        await fulfillment(of: [didFinishIterating], timeout: 0.5)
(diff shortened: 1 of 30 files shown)


netbe (approved):
Let's see if that improve things

johnxnguyen (approved):

PR #2290, opened 2024-12-13
fix: other clients don't discover new team - WPB-15043

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15043" title="WPB-15043" target="_blank">BugWPB-15043</a>  [iOS] Other clients don't discover newly created team
  </td></table>
  <br />
 

### Issue

If the user has more than one device and upgrades their personal account to a team account, then the other devices don't know about the new team because they don't process the team creation event.

### Testing

1. Be logged in to a personal account on devices A and B.
2. Upgrade to team account on device A
3. Assert that device B discovers the team, by seeing the "manage team" button in the self profile.
---

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/Synchronization/Strategies/TeamDownloadRequestStrategy.swift
+++ b/wire-ios-sync-engine/Source/Synchronization/Strategies/TeamDownloadRequestStrategy.swift
@@ -16,6 +16,8 @@
 // along with this program. If not, see http://www.gnu.org/licenses/.
 //
 
+import WireLogging
+
 struct TeamListPayload: Decodable {
     let hasMore: Bool
     let teams: [TeamPayload]
@@ -31,17 +33,17 @@ struct TeamPayload: Decodable {
     let identifier: UUID
     let name: String
     let creator: UUID
-    let binding: Bool
     let icon: String
     let iconKey: String?
+    let splashScreen: String?
 
     private enum CodingKeys: String, CodingKey {
         case identifier = "id"
         case name
         case creator
-        case binding
         case icon
         case iconKey = "icon_key"
+        case splashScreen = "splash_screen"
     }
 
 }
@@ -67,10 +69,6 @@ extension TeamPayload {
         team.creator = ZMUser.fetchOrCreate(with: creator, domain: nil, in: managedObjectContext)
         team.pictureAssetId = icon
         team.pictureAssetKey = iconKey
-
-        if !binding {
-            managedObjectContext.delete(team)
-        }
     }
 
 }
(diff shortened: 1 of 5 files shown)


wire-ios-sync-engine/Source/Synchronization/Strategies/TeamDownloadRequestStrategy.swift
```diff
@@ -31,17 +33,17 @@ struct TeamPayload: Decodable {
...
-    let binding: Bool
     let icon: String
     let iconKey: String?
+    let splashScreen: String?
 
     private enum CodingKeys: String, CodingKey {
         case identifier = "id"
         case name
         case creator
-        case binding
```
johnxnguyen, line 42: According to swagger: 

binding is deprecated, and should be ignored. The non-binding teams API is not used (and will not be supported from API version V4 onwards), and binding will always be true.

wire-ios-sync-engine/Tests/Source/Synchronization/Strategies/TeamDownloadRequestStrategyTests.swift
```diff
@@ -207,11 +207,6 @@ final class TeamDownloadRequestStrategyTests: MessagingTest {
         }
 
         XCTAssert(waitForAllGroupsToBeEmpty(withTimeout: 0.2))
-
-        syncMOC.performGroupedAndWait {
-            // then
-            XCTAssertTrue(team == nil || team.isZombieObject)
-        }
```
johnxnguyen, line 214: Deleted because we ignore "binding".

caldrian (approved):
👏

jullianm (approved):

El-Fitz:
@johnxnguyen You seem to have many unrelated changes in this branch.

johnxnguyen:
> @johnxnguyen You seem to have many unrelated changes in this branch.

@El-Fitz thanks, all fixed now

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -499,6 +499,12 @@ public final class ZMUserSession: NSObject {
         restoreDebugCommandsState()
         configureRecurringActions()
 
+        // Proactively keep the self user in sync, which helps add resilience
+        // in cases where the self client may otherwise only have limited
+        // one time opportunities to discover important changes.
+        let selfUser = ZMUser.selfUser(in: managedObjectContext)
+        selfUser.needsToBeUpdatedFromBackend = true
```
johnxnguyen, line 506: Historically clients could only discover the team during a slow sync after log in, now that a user can be associated with a team after login we need to be more proactive about discovering the team, otherwise some clients may miss their opportunity to discover the team because, for example, they were an old client that doesn't process team creation events.

With this change, we fetch the self user  and  transitively the team on launch, which in most cases may be unnecessary, but is a small cost for extra protection.

PR #2322, opened 2024-12-20
feat(CI): report failed job - WPB-15166

netbe:
### Issue

* test_frameworks will fail if one of the frameworks tests fail
* add `test_only_frameworks` to run tests of only touched frameworks for **PR tests** - old behaviour
* **Merge queue** will run tests of frameworks and their dependency
* **Nightly Builds** will run all tests
* add some tests info to Test summary on Github

### Testing

- [x] Tests should fail this PR in the check:

Screenshot 2025-01-06 at 17 22 16

### DIFF
--- a/.github/workflows/_reusable_run_tests.yml
+++ b/.github/workflows/_reusable_run_tests.yml
@@ -111,15 +111,13 @@ jobs:
         if: ${{ inputs.test_dependencies || inputs.all }}
         run: |
           set -eo pipefail
-          bundle exec fastlane test_frameworks folders:${{ inputs.folders }} all:${{ inputs.all }}
-          exit ${PIPESTATUS[0]}
+          bundle exec fastlane test_frameworks folders:${{ inputs.folders }} all:${{ inputs.all }} || exit 1
 
       - name: "Test frameworks without dependencies" 
         if: ${{ inputs.test_dependencies == false }}
         run: |
           set -eo pipefail
-          bundle exec fastlane test_only_frameworks folders:${{ inputs.folders }}
-          exit ${PIPESTATUS[0]}
+          bundle exec fastlane test_only_frameworks folders:${{ inputs.folders }} || exit 1
 
       - name: Find all .xcresult files
         if: ${{ always() && github.event_name == 'pull_request' }}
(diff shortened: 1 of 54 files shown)


netbe:
Screenshot 2024-12-20 at 15 56 27
this is how the summary can look like

WireAPI/Tests/WireAPITests/APIs/AccountsAPI/AccountsAPITests.swift
```diff
@@ -36,7 +36,7 @@ final class AccountsAPITests: XCTestCase {
     }
 
     override func tearDown() {
-        apiSnapshotHelper = nil
+        apiSnapshotHelper = nil // this is a comment
```
KaterinaWire: a comment

.github/workflows/_reusable_run_tests.yml
```diff
@@ -105,9 +108,16 @@ jobs:
...
       - name: "Test all selected frameworks" 
-        id: set_frameworks
+        if: ${{ inputs.test_dependencies || inputs.all }}
         run: |
-          bundle exec fastlane test_frameworks folders:${{ inputs.folders }} all:${{ inputs.all }}
+          set -eo pipefail
+          bundle exec fastlane test_frameworks folders:${{ inputs.folders }} all:${{ inputs.all }} || exit 1
+
+      - name: "Test frameworks without dependencies" 
+        if: ${{ inputs.test_dependencies == false }}
```
samwyndham: Is it possible to have a situation where both of these steps are run in a single job - when `all` is `true` and `test_dependencies` is `false`. We don't want this right

WireAPI/Tests/WireAPITests/APIs/AccountsAPI/AccountsAPITests.swift
netbe: yes this is to trigger some tests to run

.github/workflows/_reusable_run_tests.yml
netbe: right, for now it's up to the caller (parent workflow) not to include both

samwyndham (approved):
Nice!

KaterinaWire (approved):

PR #2325, opened 2024-12-20
fix: Creating team membership & tests - WPB-15043

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15043" title="WPB-15043" target="_blank">BugWPB-15043</a>  [iOS] Other clients don't discover newly created team
  </td></table>
  <br />
 

### Issue

This PR:

- Fixes `-[ZMUserTests testThatItUpdatesTeamIdentifierOnExistingUser]`
- Updates implementation of `ZMUser.createOrDeleteMembershipIfBelongingToTeam()` fixing `-[ZMUserTests testThatItDoesNotCreateMembershipIfUserBelongsExternalTeamOnAnExistingUser]`
- Adds a few additional tests for `ZMUser.createOrDeleteMembershipIfBelongingToTeam()`.

@johnxnguyen determined the behavior of `ZMUser.createOrDeleteMembershipIfBelongingToTeam()` should be that:

- Calling this method on the self user will create its membership of the given team, creating the team if it doesn't exit.
- Calling this method on another user will only create its membership to the given team if it exists.

As we should only create a `Team` object that corresponds to the self users team, implicitly this means that:

- Calling this method on another user will only create its membership if the team is the **same** as the self users.

This should fix failing tests on `release/cycle-3.115`.

### Testing

Testing this is internal to the app by using the debugger, as there is no external way to verify that we don't create teams or memberships in the database for other teams.

---

### Checklist

### DIFF
--- a/wire-ios-data-model/Source/Model/User/ZMUser+Teams.swift
+++ b/wire-ios-data-model/Source/Model/User/ZMUser+Teams.swift
@@ -59,13 +59,14 @@ public extension ZMUser {
             return
         }
 
+        let team: Team
         if isSelfUser {
-            let team = Team.fetchOrCreate(with: teamIdentifier, in: managedObjectContext)
+            team = Team.fetchOrCreate(with: teamIdentifier, in: managedObjectContext)
             team.needsToBeUpdatedFromBackend = true
             team.needsToRedownloadMembers = true
-        }
-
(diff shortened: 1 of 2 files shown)


johnxnguyen (approved):

jullianm (approved):

johnxnguyen:
Merging despite critical flow check failing because it's known to fail when targeting non-develop.

PR #2333, opened 2025-01-02
fix: skip workflow actions for draft PRs - WPB-15214

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15214" title="WPB-15214" target="_blank">BugWPB-15214</a>  [iOS] - CI shouldn't run for draft PRs
  </td></table>
  <br />
 

### Issue

Currently the conditions about running workflow are evaluated incorrectly for draft-PRs.
This PR attempts to fix the conditions.

---

### Checklist

---


.github/workflows/test_pr_changes.yml
```diff
@@ -16,21 +16,15 @@ jobs:
...
-      ) && github.event.pull_request.draft == false
+    if: ${{ github.event.pull_request.draft != true }}
     runs-on: ubuntu-24.04
     # Set job outputs to values from filter step
     outputs:
       folders: ${{ steps.filter.outputs.changes }}
 
     steps:
       - uses: actions/checkout@v4
+        if: ${{ github.event_name == 'merge_group' }}
```
caldrian, line 27: Checkout is only required for merge_group, for pull_request the API is used: https://github.com/dorny/paths-filter?tab=readme-ov-file#supported-workflows

.github/workflows/build_pr.yml
```diff
@@ -2,7 +2,7 @@ name: Build for PR
 
 on:
   pull_request:
-    types: [ opened, synchronize, edited, ready_for_review ]
```
caldrian, line 5: I can't think of any reason why this workflow should run for the edited event.

.github/workflows/swiftformat.yml
```diff
@@ -2,9 +2,7 @@ name: SwiftFormat
 
 on:
   pull_request:
-    types: [opened, synchronize, edited, ready_for_review]
-  merge_group:
-    types: [checks_requested]
```
caldrian: There is no point in running SwfitFormat or SwiftLint in the merge queue.

.github/workflows/build_pr.yml
samwyndham: I think the only reason is to catch that the `base` of the PR changed. See [docs](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=edited#pull_request).

> The title or body of a pull request was edited, or the **base branch** of a pull request was changed.

I wish there was another event that was fired in this case.

samwyndham: How about replacing `edited` with `reopened`. At my last company we used `reopened`. It was useful because:

- it allowed force running CI jobs by closing/reopening the PR for those edge cases where things go wrong (no more need to push empty commits)
- it is low maintenance (vs using a specific label etc)

caldrian: > I think the only reason is to catch that the base of the PR changed.

I don't think we need to run the workflows again when changing the base branch. It would just test the very same commit again.

caldrian: About the suggestion to use the `reopen` event. Is it a workaround for the workflows not being started after the cherry-pick action?
What about creating a draft PR first, then make it ready for review with another API call?

samwyndham: > I don't think we need to run the workflows again when changing the base branch. It would just test the very same commit again.

I don't believe this is true. Unless we are doing something custom, the default behavior is to checkout the merge commit which first merges changes from base into the PR branch - https://github.com/actions/checkout/issues/426#issuecomment-764193404

However I'm still in favor of dropping "edited" due to the other issues it creates.

.github/workflows/swiftformat.yml
netbe: well, we still want SwiftFormat and Swiftlint to pass for every PR. I put it as required checks now.
If you remove it, the check will wait for ever. In a sense if it passed to go inside the queue it doesn't need to run again

I think we can keep the trigger, but skip the execution

```
if: github.event.name == "merge_queue"
```

.github/workflows/swiftformat.yml
```diff
@@ -14,20 +12,24 @@ concurrency:
...
+        with:
+          filters: |
+            swift:
+              - '**/*.swift'
+
       - uses: actions/checkout@v4
+        if: ${{ steps.changes.outputs.swift == 'true' }}
+
       - name: GitHub Action for SwiftFormat
+        if: ${{ steps.changes.outputs.swift == 'true' }}
```
netbe, line 34: nice

.github/workflows/swiftlint.yml
```diff
@@ -2,9 +2,7 @@ name: SwiftLint
 
 on:
   pull_request:
-    types: [opened, synchronize, edited, ready_for_review]
-  merge_group:
```
netbe, line 6: same here, regarding the merge_queue

.github/workflows/build_pr.yml
samwyndham: > About the suggestion to use the `reopen` event. Is it a workaround for the workflows not being started after the cherry-pick action? What about creating a draft PR first, then make it ready for review with another API call?

It's a work around for cherry-pick but also any other issues that occur due to edge cases. Obviously we should try to make our setup free of these issues, but in my experience there is always some edge case which will require manually running CI-jobs.

I also think it makes sense for it to be there anyway. It is one of the default types if we don't bother to add the types field: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request

.github/workflows/test_pr_changes.yml
netbe: did you test this? the doc example shows otherwise or I don't get it

netbe (changes requested):
The edit part looks good, but the merge_queue changes are going to be an issue if we remove them

.github/workflows/swiftformat.yml
caldrian: true, reverted, thanks

.github/workflows/swiftlint.yml
caldrian: reverted, thanks

.github/workflows/_reusable_run_tests.yml
```diff
@@ -247,8 +261,15 @@ jobs:
...
       - name: Sanitize branch name
         run: |
-            SANITIZED_BRANCH=$(echo "${{ inputs.branch }}" | sed 's/[^a-zA-Z0-9._-]/_/g')
-            echo "SANITIZED_BRANCH=$SANITIZED_BRANCH" >> $GITHUB_ENV
+          if [ -z "${{ inputs.branch }}" ]; then
+            BRANCH="${{ inputs.branch }}"
+          elif [ "${{ github.event_name }}" == "pull_request" ]; then
+            BRANCH="${{ github.head_ref }}"
+          else
+            BRANCH="${GITHUB_REF#refs/heads/}"
```
netbe, line 269: so this is the case for merge_queue?

caldrian: I added a check for merge_queue.

> so this is the case for merge_queue?

The `else` is for any other trigger.

.github/workflows/test_pr_changes.yml
caldrian: At least for the SwiftLint and SwiftFormat workflows in this PR it seems to work, but they use v3 of this action.
Screenshot 2025-01-08 at 12 56 50

.github/workflows/_reusable_run_tests.yml
```diff
@@ -48,8 +48,15 @@ jobs:
 
       - name: Sanitize branch name
         run: |
-            SANITIZED_BRANCH=$(echo "${{ inputs.branch }}" | sed 's/[^a-zA-Z0-9._-]/_/g')
-            echo "SANITIZED_BRANCH=$SANITIZED_BRANCH" >> $GITHUB_ENV
+          if [ -n "${{ inputs.branch }}" ]; then
```
samwyndham, line 51: I wonder if there is a way of using `git` although not sure how to deal with the situation where some kind of merge commit is checked out.

```
git rev-parse --abbrev-ref HEAD
```

Also it would be better to extract this code to a reusable GitHub action although this may be out of scope for this PR.

samwyndham (approved):
Nice work

.github/workflows/_reusable_run_tests.yml
netbe: that's a good point using github reusable action but non-blocking

netbe (approved):
looks good

PR #2334, opened 2025-01-02
fix: read receipts setting not synchronized after login - WPB-11738

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11738" title="WPB-11738" target="_blank">BugWPB-11738</a>  [iOS] Read receipt updates from web is not synchronized in iOS
  </td></table>
  <br />
 

### Issue

After a fresh login, the read receipts setting (under Settings > Account) is always false, even if it is true on other clients. Furthermore, when the setting is changed on other clients, the value is received on iOS but the UI doesn't refresh unless dismissing it and reloading it.

The first problem is due to not correctly parsing the response payload to fetch the initial value: the payload is simply a boolean integer which is not parsed as a json object and consequently we return early. The fix is to deserialize the json object with an additional option to allow root level values.

The second issue is due to not reacting to the read receipt changes in the settings view controler.

### Testing

1. Log in with user A on web. Enable read receipts
2. Login with user A on iOS. Assert the read receipt settings is enabled.
3. While the account settings is open on iOS, on web change the read receipt settings. Assert on iOS the toggle updates accordingly.

---

### Checklist

---

### DIFF
--- a/wire-ios-request-strategy/Sources/Request Strategies/User/UserPropertyRequestStrategy.swift
+++ b/wire-ios-request-strategy/Sources/Request Strategies/User/UserPropertyRequestStrategy.swift
@@ -333,7 +333,6 @@ extension UserPropertyRequestStrategy: ZMSingleRequestTranscoder {
             )
         default:
             logger.warn("unhandled user property (\(property.propertyName)) response: \(response.result)")
-            break
         }
     }
 }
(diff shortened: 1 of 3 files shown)


KaterinaWire (approved):

jullianm (approved):

johnxnguyen:
Merging because critical flows are known not to run yet against the release branch.

PR #2341, opened 2025-01-03
chore: update SwiftLint required pattern to current year, update all files header - no ticket

jullianm:
no ticket

### Issue

Linter raises an error related to the file header because it only accepts _2024_ in its required pattern.

**Solution**: Change the year to 2025 in the swiftlint.yml file and update all header files with 2025.

### Checklist


.swiftlint.yml
```diff
@@ -53,7 +53,7 @@ file_header:
   required_pattern: |
                     \/\/
                     \/\/ Wire
-                    \/\/ Copyright \(C\) 2024 Wire Swiss GmbH
+                    \/\/ Copyright \(C\) \d{4} Wire Swiss GmbH
```
caldrian: I think we should actually change 2024 to 2025 for all files.

KaterinaWire (approved):

.swiftlint.yml
jullianm: done in https://github.com/wireapp/wire-ios/pull/2341/commits/47d39ee726b9c8c75919004e84176a1880ed31b0 thx

caldrian (approved):

PR #2345, opened 2025-01-03
fix: indiviual to team migration disable dismiss and remove close button on last step - WPB-15044

El-Fitz:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15044" title="WPB-15044" target="_blank">BugWPB-15044</a>  Manage team button not visible after completing flow
  </td></table>
  <br />
 

### Issue

The profile view doesn't properly update when dismissing the individual to team flow's modal once the user has migrated to a team account.
Updating it properly is time-consuming, so instead we simply removed the close button and disabled pull to dismiss on the last screen.

### Testing

Create a new account (staging backend, api v7).
Go through the individual to team flow.
Once you have successfully migrated, on the last screen, the close button is gone and the modal can only be dismissed through one of the two bottom buttons.

---

### Checklist

### DIFF
--- a/WireUI/Sources/WireIndividualToTeamMigrationUI/IndividualToTeamMigrationViewController.swift
+++ b/WireUI/Sources/WireIndividualToTeamMigrationUI/IndividualToTeamMigrationViewController.swift
@@ -252,7 +252,7 @@ public class IndividualToTeamMigrationViewController: UIViewController {
                 onTransition: { @MainActor [weak self] in self?.transition(to: $0) }
             )
             childController.setViewControllers([vc], animated: true)
-            isModalInPresentation = false
+            isModalInPresentation = true
         case .toCompletionDismiss:
             analyticsFlowCompletionAction = nil
             actionCallback(.completionDismiss)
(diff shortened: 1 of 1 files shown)


WireUI/Sources/WireIndividualToTeamMigrationUI/IndividualToTeamMigrationViewController.swift
```diff
@@ -351,17 +351,21 @@ private func hostedView(
...
+    if case .completion = step {
+        vc.navigationItem.rightBarButtonItem = nil
+    } else {
+        vc.navigationItem.rightBarButtonItem = UIBarButtonItem.closeButton(
+            action: UIAction { _ in
+                switch step {
+                case .teamPlanSelection, .teamName, .confirmation:
+                    transitionCallback(.toCancellationAlert)
+                case .completion:
+                    transitionCallback(.toCompletionDismiss)
```
jullianm, line 363: suggestion(non-blocking): I think this code will never be called since it is already handled in the _if block_ right above. Maybe we could replace this line with a simple break and add some code comment to make it more explicit.

jullianm (approved):
LGTM, left one non-blocking comment

WireUI/Sources/WireIndividualToTeamMigrationUI/IndividualToTeamMigrationViewController.swift
samwyndham: suggestion(non-blocking): I think it's a perfect candidate for a `fatalError` 🥳

samwyndham (approved):

WireUI/Sources/WireIndividualToTeamMigrationUI/IndividualToTeamMigrationViewController.swift
El-Fitz: I’ll look into it this afternoon.

PR #2349, opened 2025-01-06
fix: Call notifications from self user - WPB-14925

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14925" title="WPB-14925" target="_blank">BugWPB-14925</a>  [iOS] User A is getting call on iOS, after a call is made from web by User A 
  </td></table>
  <br />
 

### Issue

Currently we do not ignore begin/end call notifications coming from self user. For example, given user A has an iOS and web client, if user A calls user B from web the web client, user A will receive call kit notifications on their iOS client.

The reason for this bug is since enabling federation, the `CallEventContent.callerID` returns `nil` as it is unable to parse the underlying identifies which used to be a `UUID` string but is now a `UUID` + `domain` string.

The fix is to return a `AVSIdentifier` instead of a `UUID` from `CallEventContent.callerID` and consume that correctly. This required moving `AVSIdentifier` to `WireRequestStrategy`.

### Testing

See steps to reproduce in ticket.

---

### Checklist


wire-ios-sync-engine/WireSyncEngine.xcodeproj/project.pbxproj
```diff
@@ -222,7 +222,6 @@
 		544F8FF31DDCD34600D1AB04 /* UserProfileUpdateNotifications.swift in Sources */ = {isa = PBXBuildFile; fileRef = 544F8FF21DDCD34600D1AB04 /* UserProfileUpdateNotifications.swift */; };
 		5458273E2541C3A9002B8F83 /* PresentationDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5458273D2541C3A9002B8F83 /* PresentationDelegate.swift */; };
 		5458AF841F7021B800E45977 /* PreLoginAuthenticationNotification.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5458AF831F7021B800E45977 /* PreLoginAuthenticationNotification.swift */; };
-		545FC3341A5B003A005EEA26 /* ObjectTranscoderTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 54F8D74919AB67B300146664 /* ObjectTranscoderTests.m */; };
```
samwyndham, line 225: This is an automatic change by Xcode

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
```diff
@@ -33,7 +33,6 @@ public enum WireAnalytics {
         defer { isSetUpLock.unlock() }
 
         guard !isSetUp else {
-            assertionFailure("WireAnalytics.setup() called more than once")
```
samwyndham, line 36: This assertion is invalid. See discussion: https://github.com/wireapp/wire-ios/pull/2204#discussion_r1862451987

wire-ios-request-strategy/Sources/Notifications/CallEventContentTests.swift
```diff
@@ -148,4 +148,28 @@ class CallEventContentTests: XCTestCase {
         }
     }
 
+    func testCallerID() {
+        // Given
+        let uuidString = "550e8400-e29b-41d4-a716-446655440000"
+        let testCases: [(callerUserID: String, callerID: AVSIdentifier?)] = [
```
netbe:
```suggestion
        let testCases: [(callerUserID: String, expectedCallUserID: AVSIdentifier?)] = [
```

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
netbe: hmm, ok so this should not be the case, the goal is not to setup multiple times in a process. During tests, we could avoid that in injecting something, I am curious to know more

jullianm (approved):
LGTM

netbe (commented):
made one small suggestion, the actual fix looks good, I'd like to discuss the removal of assertionFailure before approving this

netbe (approved):

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
samwyndham: As discussed this seems to be correct so resolving.

PR #2364, opened 2025-01-08
refactor: generate notification conversation member leave - WPB-11662

jullianm:
WPB-11662

### Key points

This PR is about generating notification content related to the member leave event populating the right data in the notification (title, body, sound, category..). 

A notification should be generated only if the member leave event concerns the self user.

### Testing

- [x] Notification for member leave is correctly populated when it concerns self user
- [x] When it doesn't concern self user, the flag `shouldBuildNotification` returns false.

---

### Checklist

### DIFF
--- a/WireDomain/Sources/WireDomain/Notifications/Builders/NewSystemMessageNotificationBuilder.swift
+++ b/WireDomain/Sources/WireDomain/Notifications/Builders/NewSystemMessageNotificationBuilder.swift
@@ -110,9 +110,18 @@ struct NewSystemMessageNotificationBuilder: NotificationBuilder {
         case .conversationDeleted:
             // TODO: [WPB-11658]
             break
-        case .messageTimerUpdate:
-            // TODO: [WPB-11663]
-            break
+        case let .messageTimerUpdate(timeoutValue):
+            var timeoutStrValue: String?
+
+            if let timeoutValue {
+                let timerInMilliseconds = Double(timeoutValue)
+                let timeoutValue = timerInMilliseconds / 1000
+                let timeout: MessageDestructionTimeoutValue = .init(rawValue: timeoutValue)
+
+                timeoutStrValue = timeout.displayString
+            }
+
+            return buildTimerUpdateNotification(timeout: timeoutStrValue)
         }
 
         return UNMutableNotificationContent()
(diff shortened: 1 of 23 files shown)


WireDomain/Sources/WireDomain/Notifications/Composers/NewSystemMessageNotificationBodyComposer.swift
```diff
@@ -0,0 +1,46 @@
...
+    let format: NotificationBody.SystemMessageBodyFormat
+
+    // TODO: [WPB-15153] - Localize strings
+    func make() -> String {
+        switch format {
+        case let .createdConversation(senderName):
+            // TODO: [WPB-11657]
+            ""
+        case let .removedYou(senderName):
+            senderName != nil ? "\(senderName!) removed you" : "Someone removed you"
```
netbe: localization of this string

WireDomain/Tests/WireDomainTests/Notifications/NewSystemMessageNotificationBuilderTests.swift
```diff
@@ -0,0 +1,304 @@
...
+        let isGroup = true
+        let isTeam = false
+
+        await setupMock(isGroup: isGroup, isTeam: isTeam, selfUserID: .mockID4)
+
+        let systemMessages: [NewSystemMessageNotificationBuilder.SystemMessage] = [
+            .memberLeave(removedUserIDs: [.mockID4]) // concerns self user
+        ]
+
+        for systemMessage in systemMessages {
```
netbe: why a for loop there's only one message

netbe (commented):

WireDomain/Sources/WireDomain/Notifications/Composers/NewSystemMessageNotificationBodyComposer.swift
jullianm: this will all be done in a separated task - see https://wearezeta.atlassian.net/browse/WPB-15153

WireDomain/Tests/WireDomainTests/Notifications/NewSystemMessageNotificationBuilderTests.swift
jullianm: There will be more messages with the PRs to follow.

WireDomain/Sources/WireDomain/Notifications/Builders/NewSystemMessageNotificationBuilder.swift
```diff
@@ -0,0 +1,194 @@
...
+        let selfUserID: UUID
+        let hidesNotificationContent: Bool
+        let systemMessage: SystemMessage
+    }
+
+    init(
+        systemMessage: SystemMessage,
+        conversationID: WireAPI.QualifiedID,
+        senderID: UserID
+    ) async {
```
caldrian: I find it a bit weird that the init needs to be async and already does many things.
What if you inject the builder somewhere but it is never used?

I usually use the init to just prepare everything for the actual build-invocations.

caldrian (approved):

PR #2365, opened 2025-01-08
refactor: make clientID optional in AuthenticationManager - WPB-10801

johnxnguyen:
<table>
<td>
  </td></table>
  <br />
 

### Issue

`AuthenticationManager` uses a non-optional client id to request an access access token, however this is an optional query parameter. This PR makes the client id optional to allow more flexibility (e.g use the authentication manager even before a self client exists).

### Testing

N/A
---

### Checklist

---


netbe (approved):
looks good, though I have 2 questions:
- in what case would you use /access without clientID?
- Once you do get the clientID, you'll need to recreate the actor, why not modifying it instead?

johnxnguyen:
> in what case would you use /access without clientID?

@netbe after login but before you've created and registered a self client.

> Once you do get the clientID, you'll need to recreate the actor, why not modifying it instead?

Yes we could modify it, but for the current purposes (integrating slow sync) we'll create the network stack and then discard it. Eventually when we keep it around, we'll want to update the client id.

KaterinaWire (approved):

caldrian (approved):

PR #2366, opened 2025-01-08
chore: individual to team UI snapshot & buttons tests - WPB-15281

El-Fitz:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15281" title="WPB-15281" target="_blank">TaskWPB-15281</a>  Individual to Teams Views & Components snapshot tests
  </td></table>
  <br />
 

### Issue

Individual to teams & the buttons didn't have snapshot UI tests.
They now do.

---

### Checklist


netbe (changes requested):
seems like the tests are failing

WireUI/Tests/WireDesignTests/Buttons/SecondaryButtonStyleUITests.swift
```diff
@@ -0,0 +1,75 @@
...
+    override func setUp() {
+        snapshotHelper = .init()
+            .withSnapshotDirectory(SnapshotTestReferenceImageDirectory)
+    }
+
+    override func tearDown() {
+        snapshotHelper = nil
+    }
+
+    @MainActor @ViewBuilder static var view: some View {
```
netbe, line 39: question:: is it necessary to annotate @MainActor here ?

also I would have expected the view to be created in the setUp(), but I have not seen any snapshot with SwiftUI yet, is that the general pattern to do.

WireUI/Tests/WireIndividualToTeamMigrationUITests/Components/CheckboxSnapshotTests.swift
```diff
@@ -0,0 +1,101 @@
...
+
+        Checkbox(isChecked: .constant(true), title: "Checkbox")
+            .frame(width: screenBounds.width, height: screenBounds.height)
+    }
+
+    @MainActor @ViewBuilder static var unchecked: some View {
+        let screenBounds = UIScreen.main.bounds
+
+        Checkbox(isChecked: .constant(false), title: "Checkbox")
+            .frame(width: screenBounds.width, height: screenBounds.height)
```
netbe, line 49: question: why do you use the full screen and not the intrinsic content size?

netbe (approved):
Looks good, just left some questions

WireUI/Tests/WireIndividualToTeamMigrationUITests/Components/CheckboxSnapshotTests.swift
El-Fitz: No specific reason. I just used the exact same bounds for each and every component.

caldrian (approved):

WireUI/Tests/WireDesignTests/Buttons/SecondaryButtonStyleUITests.swift
El-Fitz: > question:: is it necessary to annotate @mainactor here ?

Not necessarily for this view specifically; didn't check. But many SwiftUI static properties and methods are Main actor-isolated and cannot be used from a nonisolated context (eg `.circular`, from `.progressViewStyle`).

And since the view itself can only be instantiated on the Main actor (hence the @MainActor test methods)...

> also I would have expected the view to be created in the setUp()

I personally avoid optional references, which we would need to have if we created the view in the setUp method (like the snapshotHelper).

PR #2369, opened 2025-01-09
fix: use client id from legal hold info payload - WPB-10801

johnxnguyen:
<table>
<td>
  </td></table>
  <br />
 

### Issue

When fetching the legal hold info for the self user, the legal hold client id inside the response payload is ignored and instead the self client id is used instead, which is wrong (according to the existing implementation. In this PR, I fix the decoding of the response payload and remove the dependency on the self client id.

### Testing

N/A
---

### Checklist

---

### DIFF
--- a/WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
+++ b/WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
@@ -325,19 +325,30 @@ struct LegalHoldLastPrekeyV0: Decodable, ToAPIModelConvertible {
 
 struct TeamMemberLegalholdResponseV0: Decodable, ToAPIModelConvertible {
 
-    let lastPrekey: LegalHoldLastPrekeyV0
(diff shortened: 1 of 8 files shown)


WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
```diff
@@ -325,19 +325,30 @@ struct LegalHoldLastPrekeyV0: Decodable, ToAPIModelConvertible {
 
 struct TeamMemberLegalholdResponseV0: Decodable, ToAPIModelConvertible {
 
-    let lastPrekey: LegalHoldLastPrekeyV0
     let status: LegalholdStatusV0
+    let client: LegalholdClientV0?
+    let lastPrekey: LegalHoldLastPrekeyV0?
```
johnxnguyen, line 330: Swagger says that these two fields are optional.

caldrian (approved):
👍

jullianm (approved):

PR #2377, opened 2025-01-10
refactor: generate notification conversation member join - WPB-11661

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11661" title="WPB-11661" target="_blank">TaskWPB-11661</a>  [iOS] Generate UNNotificationContent for conversation member join
  </td></table>
  <br />
 

### Key points

This PR is about generating notification content related to the member join event populating the right data in the notification (title, body, sound, category..). 

A notification should be generated only if the member join event concerns the self user.

### Testing

- [x] Notification for member join is correctly populated when it concerns self user
- [x] When it doesn't concern self user, the flag `shouldBuildNotification` returns false.

---

### Checklist

### DIFF
--- a/WireDomain/Sources/WireDomain/Notifications/Builders/NewSystemMessageNotificationBuilder.swift
+++ b/WireDomain/Sources/WireDomain/Notifications/Builders/NewSystemMessageNotificationBuilder.swift
@@ -137,7 +137,7 @@ struct NewSystemMessageNotificationBuilder: NotificationBuilder {
 
         return content
     }
-    
+
     private func buildMemberJoinNotification() -> UNMutableNotificationContent {
         let content = UNMutableNotificationContent()
 
--- a/WireDomain/Sources/WireDomain/Notifications/Builders/NewSystemMessageNotificationBuilder.swift
+++ b/WireDomain/Sources/WireDomain/Notifications/Builders/NewSystemMessageNotificationBuilder.swift
@@ -90,8 +90,7 @@ struct NewSystemMessageNotificationBuilder: NotificationBuilder {
         case let .memberLeave(removedUserIDs):
             removedUserIDs.contains(context.selfUserID)
         case let .memberJoin(addedUserIDs):
-            // TODO: [WPB-11661]
-            true
+            addedUserIDs.contains(context.selfUserID)
         case .conversationCreated, .conversationDeleted, .messageTimerUpdate:
             true
         }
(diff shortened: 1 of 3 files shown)


jullianm:
this PR targets refactor/generate-notification-conversation-member-leave because the base implementation was done in this PR

WireDomain/Sources/WireDomain/Notifications/Composers/NewSystemMessageNotificationBodyComposer.swift
```diff
@@ -24,23 +24,22 @@ struct NewSystemMessageNotificationBodyComposer {
...
-            // TODO: [WPB-11657]
-            ""
         case let .removedYou(senderName):
             senderName != nil ? "\(senderName!) removed you" : "Someone removed you"
-        case let .setMessageTimer(senderName):
-            // TODO: [WPB-11663]
-            ""
         case let .addedYou(senderName):
-            // TODO: [WPB-11661]
+            senderName != nil ? "\(senderName!) added you" : "Someone added you"
```
netbe, line 33: what about localization?

WireDomain/Sources/WireDomain/Notifications/Builders/NewSystemMessageNotificationBuilder.swift
```diff
@@ -140,6 +138,26 @@ struct NewSystemMessageNotificationBuilder: NotificationBuilder {
...
+
+        let body = NotificationBody.newSystemMessage(
+            .addedYou(senderName: context.senderName)
+        )
+
+        content.body = body.make()
+        content.categoryIdentifier = makeCategory()
+        content.sound = makeSound()
+        content.userInfo = makeUserInfo()
+        content.threadIdentifier = context.conversationID.uuid.transportString()
```
netbe: is it worth to factor common logic into a function for common code with buildMemberLeaveNotification? or at least check after everything is implemented

netbe (changes requested):
I think localization should be taken care of

WireDomain/Sources/WireDomain/Notifications/Composers/NewSystemMessageNotificationBodyComposer.swift
jullianm: All localizations will be done in a separated PR: https://wearezeta.atlassian.net/browse/WPB-15153

WireDomain/Sources/WireDomain/Notifications/Builders/NewSystemMessageNotificationBuilder.swift
jullianm: Yes, that is was I was planning to do, once everything is merged into a single branch, before merging to develop, do a final review and possibly factor common logic to avoid code duplication.

jullianm: I think, following @johnxnguyen's suggestion, we will go with creating a builder per notification type, kind of following the pattern of the event processors. The changes will be done in source branch PR.

PR #2380, opened 2025-01-13
refactor: generate notification conversation timer update - WPB-11663

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11663" title="WPB-11663" target="_blank">TaskWPB-11663</a>  [iOS] Generate UNNotificationContent for conversation timer update
  </td></table>
  <br />
 

### Key points

This PR is about generating notification content related to timer update event (enabling/disabling self destructed messages) populating the right data in the notification (title, body, sound, category..). 

### Testing

- [x] Notification for timer update is correctly populated when the timer is enabled
- [x] Notification for timer update is correctly populated when the timer is disabled

---

### Checklist

### DIFF
--- a/WireDomain/Sources/WireDomain/Notifications/Builders/NewSystemMessageNotificationBuilder.swift
+++ b/WireDomain/Sources/WireDomain/Notifications/Builders/NewSystemMessageNotificationBuilder.swift
@@ -110,9 +110,18 @@ struct NewSystemMessageNotificationBuilder: NotificationBuilder {
         case .conversationDeleted:
             // TODO: [WPB-11658]
             break
-        case .messageTimerUpdate:
-            // TODO: [WPB-11663]
-            break
+        case let .messageTimerUpdate(timeoutValue):
+            var timeoutStrValue: String?
+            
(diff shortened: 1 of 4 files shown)


netbe (approved):
I didn't know we had a notification for this, lgtm;)

WireDomain/Sources/WireDomain/Notifications/Builders/NewSystemMessageNotificationBuilder.swift
johnxnguyen, line 1: **question:** I think I may have left this comment on another PR but just incase I didn't, what do you think about having separate builders for each notification type? Is there a reason to group them together in to one?

jullianm: I think your suggestion makes sense since this will be aligned with what we did with the event processors so we have some consistency in our new codebase.

jullianm:
merging this one into the source branch but reviews added here will be taken into consideration.

PR #2381, opened 2025-01-13
refactor: generate notification conversation deleted - WPB-11658

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11658" title="WPB-11658" target="_blank">TaskWPB-11658</a>  [iOS] Generate UNNotificationContent for deleted conversations
  </td></table>
  <br />
 

### Key points

This PR is about generating notification content related to the conversation deleted event populating the right data in the notification (title, body, sound, category..). 

### Testing

- [x] Notification for deleted conversation is correctly populated

---

### Checklist

### DIFF
--- a/WireDomain/Sources/WireDomain/Notifications/Builders/NewSystemMessageNotificationBuilder.swift
+++ b/WireDomain/Sources/WireDomain/Notifications/Builders/NewSystemMessageNotificationBuilder.swift
@@ -108,8 +108,7 @@ struct NewSystemMessageNotificationBuilder: NotificationBuilder {
             // TODO: [WPB-11661]
             break
         case .conversationDeleted:
-            // TODO: [WPB-11658]
-            break
+            return buildDeletedConversationNotification()
         case .messageTimerUpdate:
             // TODO: [WPB-11663]
             break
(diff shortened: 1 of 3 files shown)


netbe (approved):
LGTM;)

WireDomain/Sources/WireDomain/Notifications/Builders/NewSystemMessageNotificationBuilder.swift
```diff
@@ -140,6 +139,26 @@ struct NewSystemMessageNotificationBuilder: NotificationBuilder {
...
+
+        if let title = makeTitle() {
+            content.title = title
+        }
+
+        let body = NotificationBody.newSystemMessage(
+            .deletedGroup(senderName: context.senderName)
+        )
+
+        content.body = body.make()
```
johnxnguyen, line 225: **issue:** one of the things that I don't like in our legacy code is how the actual values for the notification content separated from the `content` object because it usually means that there is some work involved in understanding exactly how a notification is configured.

I know that in some cases there is some complexity in generating content such as the body, and that it should also be reused among various notifications types, but I believe that mostly applies to new message notifications. In this case, I think it would be better to have the actual values closer to the `content`, something like:

```swift
content.body = "A conversation was deleted"
content.categoryIdentifier = .nonActionable
content.sound = .click
```

What do you think?

jullianm: I would tend to agree and keep things simple when we can. I'm writing your review down and the related changes will be reported to the source PR (I will mention them to simplify the final review).

PR #2384, opened 2025-01-13
fix: calculate supported protocols - WPB-15297

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15297" title="WPB-15297" target="_blank">BugWPB-15297</a>  [iOS] Wrong Supported protocols
  </td></table>
  <br />
 

### Issue

During the calculation of the supported protocols, we look if all the self clients support MLS. The calculation depends on the MLS public key and the lastActive date fields of the client. Those two fields don't come from the `userClientAdd` event. Moreover, the upload of the MLS public keys is not done on login but a bit later for each client.

For all that reasons, we need to fetch the self clients to retrieve this information. This PR fetches the self clients before doing calculation of supported protocols.

* remove fix before MLS was enabled (calculateSupportedProtocolsIfNeeded)
* extracted code from UserClientsRepository to own object `PullSelfUserClients`
* add closure to build PullSelfUserClients to ResolveOneOnOneConversationsUseCase

### Testing

See steps in ticket

---

### Checklist


wire-ios-sync-engine/Source/Use cases/ResolveOneOnOneConversationsUseCase.swift
```diff
@@ -25,20 +27,22 @@ public protocol ResolveOneOnOneConversationsUseCaseProtocol {
 
 }
 
+typealias PullSelfUserClientsFactory = (NSManagedObjectContext) -> PullSelfUserClientsProtocol
```
netbe, line 30: we use a closure to provide the PullSelfUserClients object because it depends on apiService which won't be ready when creating the strategies

WireDomain/Sources/WireDomain/UseCases/PullSelfUserClients.swift
```diff
@@ -0,0 +1,90 @@
...
+//
+
+import CoreData
+import WireAPI
+
+// sourcery: AutoMockable
+public protocol PullSelfUserClientsProtocol {
+
+    func pullSelfClients() async throws
+}
```
johnxnguyen: We started to move protocols in their own files.

johnxnguyen (approved):

jullianm (approved):
LGTM

WireDomain/Sources/WireDomain/UseCases/PullSelfUserClients.swift
netbe: done

PR #2389, opened 2025-01-15
fix: notifications stop working when switching backend - WPB-15127

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15127" title="WPB-15127" target="_blank">BugWPB-15127</a>  [iOS] Notifications stop working
  </td></table>
  <br />
 

### Issue

When switching backend after logging out from an account and then log to a new account, the NSE will not work because the cookie is invalid.

This is due to the fact that the `environmentType` (prod, staging,...), although it is correctly saved to applicationGroup UserDefaults for the NSE to work correctly, it does not read from the `UserDefaults`. Therefore it falls back to production and invalidates the session.

Solution:

Read environmentType from applicationGroup in BackendEnvironment.shared.

### Testing

- [x] See steps in ticket to reproduce

Note: as the code is in a static variable it is hard to add a unit test.

---

### Checklist

### DIFF
--- a/wire-ios/WireCommonComponents/BackendEnvironment+Shared.swift
+++ b/wire-ios/WireCommonComponents/BackendEnvironment+Shared.swift
@@ -24,12 +24,11 @@ private let zmsLog = ZMSLog(tag: "backend-environment")
 public extension BackendEnvironment {
     static let backendSwitchNotification = Notification.Name("backendEnvironmentSwitchNotification")
     static var shared: BackendEnvironment = {
-        let environmentType: EnvironmentType
-        if let typeOverride = AutomationHelper.sharedHelper.backendEnvironmentTypeOverride() {
-            environmentType = EnvironmentType(stringValue: typeOverride)
+        let environmentType = if let typeOverride = AutomationHelper.sharedHelper.backendEnvironmentTypeOverride() {
+            EnvironmentType(stringValue: typeOverride)
         } else {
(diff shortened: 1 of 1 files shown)


wire-ios/WireCommonComponents/BackendEnvironment+Shared.swift
```diff
@@ -24,7 +24,8 @@ private let zmsLog = ZMSLog(tag: "backend-environment")
 public extension BackendEnvironment {
     static let backendSwitchNotification = Notification.Name("backendEnvironmentSwitchNotification")
     static var shared: BackendEnvironment = {
-        var environmentType: EnvironmentType?
+        // read from userDefaults first
+        var environmentType: EnvironmentType? = EnvironmentType(userDefaults: .applicationGroup)
```
samwyndham: I don't think environmentType needs to be optional after your change and you can do something like the following to remove the `var` too:

```swift
        let environmentType: EnvironmentType
        if let typeOverride = AutomationHelper.sharedHelper.backendEnvironmentTypeOverride() {
            environmentType = EnvironmentType(stringValue: typeOverride)
        } else {
            environmentType = EnvironmentType(userDefaults: .applicationGroup)
        }
```

samwyndham (approved):

caldrian (approved):

PR #2393, opened 2025-01-16
chore: move use case into WireDomainPkg - WPB-10347

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10347" title="WPB-10347" target="_blank">EpicWPB-10347</a>  B - Individuals creating a New Team
  </td></table>
 

### Issue

This PR deletes unused Swift package targets and moves the `IndividualToTeamMigrationUseCaseProtocol` into `WireDomainPkg`.

Screenshot 2025-01-16 at 17 31 10

---

### Checklist

---


netbe (commented):
question: why do we have unused targets in the first place?

caldrian:
> why do we have unused targets in the first place?

@netbe I had expected the WireDomain project to be a short-term temporary solution.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+IndividualToTeamMigration.swift
```diff
@@ -19,10 +19,13 @@
 import Foundation
 import WireAPI
 import WireDomain
-import WireDomainAPI
+import WireDomainPkg
```
jullianm, line 22: question(non-blocking): do we need that import since we already import the Xcode framework right above (line 21) ?

WireDomain/Package.swift
```diff
@@ -7,42 +7,16 @@ let package = Package(
     name: "WireDomainPackage",
     platforms: [.iOS(.v16), .macOS(.v12)],
     products: [
-        .library(name: "WireDomainAPI", targets: ["WireDomainAPI"]),
-        .library(name: "WireDomainPackage", targets: ["WireDomainPkg"]),
-        .library(name: "WireDomainPackageSupport", targets: ["WireDomainPkgSupport"])
```
jullianm, line 12: question(non-blocking): sanity check that UTs are still compiling in WireDomain, I'm assuming yes since this was moved at Xcode framework level. I'm talking specifically about the generated mocks by Sourcery.

jullianm (approved):
Looks good to me, left a couple of (non-blocking) comments 👍

WireDomain/Package.swift
caldrian: Since there is currently only a single protocol and nothing else in WireDomainPkg, this is fine, thanks.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+IndividualToTeamMigration.swift
caldrian: Yes, it is still required.

KaterinaWire (approved):

PR #2394, opened 2025-01-16
fix: Connection request team bug - WPB-15275

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15275" title="WPB-15275" target="_blank">BugWPB-15275</a>  [iOS] Cannot create 1:1 conversation with personal user that migrated to my team after cancelling connection request
  </td></table>
  <br />
 

### Issue

Currently when users have a connection request and those users both end up on the same team, the connection request still exists even though it is no longer needed. Furthermore if the connection request is cancelled the associated one on one conversation has its status set to invalid preventing a user from starting a conversation. See https://wearezeta.atlassian.net/browse/WPB-15275

What we want to happen when a personal user migrates to be a team user is:

1. Any existing connection requests disappear in the UX.
2. Users can have a one on one conversation with each other.

Unfortunately the backend doesn't seem to help much here (by for example automatically cancelling connections when migrating) so with this PR we attempt take some control of the situation. To do that we need to:

1. Automatically cancel / ignore  requests when a client detects a change in team. (cancel vs ignore depends on if the user sent or received the request). Unfortunately every connection has a related 1 on 1 conversation whose status will be set to `invalid` and the syncing of connections will attempt to reset the invalid 1 on 1 on the associated `oneOnOneConversation` property even if we have removed that connection. Therefore we attempt to:
2. Regularly set the `ZMUser.oneOnOneConversation` to `nil` using the `ConnectionValidator` where it determines an connection is no longer valid.
3. Rebuild this connection using the `OneOnOneResolver` which tries to ensure we link the correct 1 on 1 conversation (there can be multiple 😢)

As we know from the recent playlets, this solution is still not without bugs but hopefully we can improve on that with follow-up PRs.

### Testing

- user a creates a team account
- user b creates a personal account
- user a sends a connection request to user b
- user b does NOT answer the connection request
- user a invites user b to their team
- user b follows the team invitation flow to join user a’s team

The expected outcome should be:

- the connection request automatically disappears from both users devices
- user a can start a conversation with user b
- alternatively user b can start a conversation with user a

Essentially I want to exploratory test the above in different situations - eg. both users using the app at the same time, app in background etc..

### Checklist


wire-ios-data-model/Source/UseCases/PerformPostMembershipCleanUpUseCase.swift
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    /// Creates `ZMMemberships` for users which belong to `selfUsers` team but have no membership.
+    private func createMissingMemberships(selfUserTeamID: UUID) throws {
+        let fetchRequest = NSFetchRequest<ZMUser>(entityName: ZMUser.entityName())
+        fetchRequest.predicate = NSCompoundPredicate(
+            andPredicateWithSubpredicates: [
+                NSPredicate(format: "membership == nil"),
+                NSPredicate(format: "teamIdentifier_data != nil"),
```
netbe:
```suggestion
                NSPredicate(format: "teamIdentifier_data == %@", selfUserTeamID.data),
```
alternatively

wire-ios-request-strategy/Sources/Request Strategies/User/UserProfileRequestStrategy.swift
```diff
@@ -208,6 +208,15 @@ extension UserProfileRequestStrategy: ZMEventConsumer {
             for: user,
             authoritative: false
         )
+
+        if userProfile.updatedKeys.contains(.teamID) {
+            let useCase = PerformPostMembershipCleanUpUseCase(context: managedObjectContext, userID: user.objectID)
+            do {
+                try useCase.invoke()
+            } catch {
+                WireLogger.individualToTeamMigration.error("Error performing post membership cleanup")
```
netbe:
```suggestion
                WireLogger.individualToTeamMigration.error("Error performing post membership cleanup: \(error.localizedDescription)")
```

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -968,6 +968,13 @@ extension ZMUserSession: ZMSyncStateDelegate {
 
             await calculateSelfSupportedProtocolsIfNeeded()
             await resolveOneOnOneConversationsIfNeeded()
+
+            let useCase = PerformPostMembershipCleanUpUseCase(context: managedObjectContext, userID: nil)
+            do {
+                try await useCase.invoke()
+            } catch {
+                WireLogger.individualToTeamMigration.error("Error performing post membership cleanup")
+            }
```
netbe: still needed ?

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController/ViewModel/ConversationListViewModel.swift
```diff
@@ -232,6 +234,17 @@ final class ConversationListViewModel: NSObject {
 
     private func setupObservers() {
         conversationDirectoryToken = userSession?.conversationDirectory.addObserver(self)
+
+        guard let user = userSession?.selfUser as? ZMUser else { return }
+
+        user.publisher(for: \.teamIdentifier)
+            .receive(on: RunLoop.main)
+            .removeDuplicates()
```
netbe:
```suggestion
            .removeDuplicates()
            .receive(on: RunLoop.main)
```
can we swap the operators ?

netbe (commented):
Tested on the build, seems to work. Left some comments.

Just need some unit tests if possible and we're good to go 🚢

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
samwyndham: Yes, to add resilience to the system in case there is an issue handling the `user.update` event.

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController/ViewModel/ConversationListViewModel.swift
samwyndham: Good idea. Fixed!

wire-ios-request-strategy/Sources/Request Strategies/User/UserProfileRequestStrategy.swift
samwyndham: I was a bit afraid that we might log personal information here. Or is this safe?

wire-ios-data-model/Source/UseCases/PerformPostMembershipCleanUpUseCase.swift
samwyndham: I would definitely prefer to do the test here than further down but I'm not sure it is safe.

Is the data that we used to create these UUIDs always the same as what we get out from them?

samwyndham:
> Instead here I'm currently using KVO based publisher on a cast ZMUser object 🤮. In my minimal tests it works but I'm not sure if the self user ZMUser instance can change which may break things 🤔 . It needs some more thought

One situation I've checked is when we switch accounts. Thankfully things work fine here as the `ConversationListViewModel` is recreated and we re-subscribe to the user updates

wire-ios-data-model/Source/UseCases/PerformPostMembershipCleanUpUseCase.swift
samwyndham: This code has been deleted so resolving

wire-ios-data-model/Source/EntityValidation/ConnectionValidator.swift
```diff
@@ -0,0 +1,228 @@
...
+
+    /// Create a new `ConnectionValidator`.
+
+    public init(context: NSManagedObjectContext) {
+        self.context = context
+    }
+
+    /// Clean up all invalid connections.
+    ///
+    /// Invoking this method will search the local database for invalid
```
netbe, line 59:
```suggestion
    /// Invoking this method will search the local database for
```

wire-ios-data-model/Source/EntityValidation/ConnectionValidator.swift
```diff
@@ -0,0 +1,228 @@
...
+    /// invalid connections and reject them all. As a result, only valid
+    /// connections will remain in the database.
+    ///
+    /// A connection is considered invalid if it is pending (i.e not accepted,
+    /// not blocked) and between the self user and a fellow team member. This
+    /// can happen if the a pending connection exists between the self user
```
netbe:
```suggestion
    /// can happen if the pending connection exists between the self user
```

wire-ios-data-model/Source/EntityValidation/ConnectionValidator.swift
```diff
@@ -0,0 +1,228 @@
...
+            )
+        }
+
+        try await cleanUpState(for: searchResult)
+    }
+
+    private func cleanUpState(for searchResult: SearchResult) async throws {
+        // Cancel outgoing connections.
+        for connectionID in searchResult.connectionsToCancel {
+            try await updateConnectionStatus(
```
netbe: if one of the connection fails here, the others will not be updated

wire-ios-data-model/Source/EntityValidation/ConnectionValidator.swift
```diff
@@ -0,0 +1,228 @@
...
+                connectionID: connectionID,
+                newStatus: .cancelled,
+                context: context
+            )
+        }
+
+        // Ignore incoming connections.
+        for connectionID in searchResult.connectionsToIgnore {
+            try await updateConnectionStatus(
```
netbe: if one of the connection fails here, the others will not be updated

wire-ios-data-model/Source/EntityValidation/ConnectionValidator.swift
```diff
@@ -0,0 +1,228 @@
...
+                newStatus: .ignored,
+                context: context
+            )
+        }
+
+        // Invalidate and unlink the associated conversation.
+        try await context.perform { [context] in
+            let connections = try searchResult.invalidConnections.map {
+                try context.existingObject(with: $0) as! ZMConnection
+            }
```
netbe:
```suggestion
            let connections = searchResult.invalidConnections.compactMap {
                try? context.existingObject(with: $0) as? ZMConnection
            }
```

wire-ios-sync-engine/Source/Use cases/ResolveOneOnOneConversationsUseCase.swift
```diff
@@ -49,6 +49,7 @@ struct ResolveOneOnOneConversationsUseCase: ResolveOneOnOneConversationsUseCaseP
             }
         }
 
+        // Do we need to remove this check?
```
netbe: it would mean you cleanup the connections more often, on every quick sync.

netbe: only when mls is enabled, so it won't make an impact, let's keep the check for now

netbe (approved):
Approving, but left some comments to address

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneSource.swift
```diff
@@ -0,0 +1,185 @@
...
+
+        return .all(of: [
+            isOneOnOne,
+            isMLS,
+            hasTwoParticipants,
+            hasParticipant(user: otherUser)
+        ])
+    }
+
+    static func fakeProteusTeamOneOnOne(selfUser: ZMUser, otherUser: ZMUser) -> NSPredicate {
```
jullianm, line 127: question(non-blocking): Do we call it fake because it has no user defined name or for another reason ? maybe it's worth adding some code doc to describe why we have a `fake` case from the `OneOnOneType`. The other cases are self explanatory, this one is little bit less obvious (I think).

jullianm (approved):
LGTM, nice work dealing with these bugs!

wire-ios-data-model/Source/MLS/OneOnOne/OneOnOneSource.swift
samwyndham: Agree with the need for documentation. I will add some in a follow up PR where I add tests.

The reason for the `fake` term is that a conversation has a conversation type:

```objc
typedef NS_CLOSED_ENUM(int16_t, ZMConversationType) {
    ZMConversationTypeInvalid = 0,
    ZMConversationTypeSelf,
    ZMConversationTypeOneOnOne,
    ZMConversationTypeGroup,
    ZMConversationTypeConnection, // Incoming & outgoing connection request
};
```

If a one on one is based on a connection the conversation type will be  `OneOnOne` but if it is a team user it will be `Group` and we use a bunch of heuristics to determine whether it is actually a `OneOnOne` 😢 

`ZMConversation.conversationType` is actually a computed property because of this:

```objc
- (ZMConversationType)conversationType
{
    ZMConversationType conversationType = [self internalConversationType];
    
    // Exception: the group conversation is considered a 1-1 if:
    // 1. Belongs to the self team.
    // 2. Has no name given.
    // 3. Conversation has only one other participant.

    // Performance note: localParticipantsExcludingSelf will enumerate over all
    // local participant roles, so check its count first to avoid unncessary iterations.

    if (conversationType == ZMConversationTypeGroup &&
        self.team != nil &&
        self.userDefinedName.length == 0 &&
        self.localParticipantRoles.count == 2 &&
        self.localParticipantsExcludingSelf.count == 1)
    {
        conversationType = ZMConversationTypeOneOnOne;
    }
    
    return conversationType;
}
```

wire-ios-request-strategy/Sources/Request Strategies/User/UserProfileRequestStrategy.swift
samwyndham: Resolving this as the code has changed.

wire-ios-data-model/Source/EntityValidation/ConnectionValidator.swift
samwyndham: Fixed in [dd07fec](https://github.com/wireapp/wire-ios/pull/2394/commits/dd07fec6d4b462849e985a678363e385243a1320)

wire-ios-data-model/Source/EntityValidation/ConnectionValidator.swift
samwyndham: [dd07fec](https://github.com/wireapp/wire-ios/pull/2394/commits/dd07fec6d4b462849e985a678363e385243a1320)

wire-ios-data-model/Source/EntityValidation/ConnectionValidator.swift
samwyndham: Improved in https://github.com/wireapp/wire-ios/pull/2394/commits/84419fb98d1c0e634405c7ca8597aeed762ac7af. Please take a look

wire-ios-data-model/Source/EntityValidation/ConnectionValidator.swift
samwyndham: Improved in https://github.com/wireapp/wire-ios/pull/2394/commits/84419fb98d1c0e634405c7ca8597aeed762ac7af. Please take a look

wire-ios-data-model/Source/EntityValidation/ConnectionValidator.swift
samwyndham: Improved in https://github.com/wireapp/wire-ios/pull/2394/commits/84419fb98d1c0e634405c7ca8597aeed762ac7af. Please take a look

netbe (approved):

PR #2412, opened 2025-01-22
feat: Implement 2FA for new authentication flow - WPB-15226

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15226" title="WPB-15226" target="_blank">TaskWPB-15226</a>  [iOS] Implement new authentication flow views - 2FA
  </td></table>
  <br />
 

### Issue

This PR introduces the 2FA view for the new login flow:

Screenshot 2025-01-21 at 17 33 57

### Testing

---

### Checklist

---


WireUI/Sources/WireAuthenticationUI/Views/VerificationCodeView.swift
```diff
@@ -0,0 +1,123 @@
...
+                .multilineTextAlignment(.center)
+
+            Text(L10n.VerificationCode.message(receiver))
+                .wireTextStyle(.body1)
+                .multilineTextAlignment(.center)
+                .foregroundStyle(Color.primaryText)
+
+            HStack(spacing: 10) {
+                ForEach(0 ..< 6, id: \.self) { index in
+                    TextField("", text: $code[index])
```
jullianm: suggestion: Can't we do something like this here`TextField("", value: $code[index], format: .number)` and turn the code array into an array of integer ?
 
This API should automatically reject any input that is not a number.

WireUI/Sources/WireAuthenticationUI/Views/VerificationCodeView.swift
```diff
@@ -0,0 +1,123 @@
...
+                            RoundedRectangle(cornerRadius: 8)
+                                .stroke(
+                                    focusedIndex == index ? Color.primaryButtonBackground : Color.secondaryButtonBorder,
+                                    lineWidth: 1
+                                )
+                        )
+                        .multilineTextAlignment(.center)
+                        .font(.textStyle(.h2))
+                        .foregroundColor(.primary)
+                        .keyboardType(.numberPad)
```
jullianm: note: iPad users would still get the full keyboard, also (I think) it doesn't prevent a user to paste a non-numeric input. See suggestion above.

jullianm (commented):
LGTM, left a couple of comments before approving

WireUI/Sources/WireAuthenticationUI/Views/VerificationCodeView.swift
KaterinaWire: When using `TextField("", value: $code[index], format: .number)`, pasting letters doesn't trigger any validation, because:

- SwiftUI ignores non-numeric input automatically due to format: .number.
- `onChange` doesn't get called if the pasted content contains non-numeric characters.

I've updated `handleInput` method to accept only digits. And the verification code will be passed to the backend endpoint as a string anyway.

jullianm (approved):

WireUI/Sources/WireAuthenticationUI/Views/VerificationCodeView.swift
```diff
@@ -0,0 +1,165 @@
...
+    private let receiver: String
+    private let onConfirm: ([String]) -> Void
+    private let onResend: () -> Void
+
+    private var isConfirmButtonDisabled: Bool {
+        code.contains { $0.isEmpty }
+    }
+
+    package init(
+        initialCode: [String] = Array(repeating: "", count: 6),
```
El-Fitz: I would haver the count as a `static let` property of a private `Constants` enum nested in the struct.
This way it can easily be found if it ever needs changing.

WireUI/Sources/WireAuthenticationUI/Views/VerificationCodeView.swift
```diff
@@ -0,0 +1,165 @@
...
+        .cornerRadius(16)
+        .overlay(
+            RoundedRectangle(cornerRadius: 16)
+                .stroke(ColorTheme.Backgrounds.surface.color, lineWidth: 1)
+        )
+    }
+
+    private var verificationCodeView: some View {
+        HStack(spacing: 10) {
+            ForEach(0 ..< 6, id: \.self) { index in
```
El-Fitz: And this 6 would come from the same static property.

El-Fitz (changes requested):

WireUI/Sources/WireAuthenticationUI/Views/VerificationCodeView.swift
KaterinaWire: done

WireUI/Sources/WireAuthenticationUI/Views/VerificationCodeView.swift
KaterinaWire: done

El-Fitz (approved):

PR #2416, opened 2025-01-23
fix: MLS add message decryption error - WPB-15646

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15646" title="WPB-15646" target="_blank">BugWPB-15646</a>  [iOS] Decryption error wrong epoch
  </td></table>
  <br />
 

### Issue

When processing a MLS add message event, decryption operations can result in empty decrypted events (i.e a proposal) however, while the decryption operation is successful, we omit to store the last event ID in database. 

As a consequence, when fetching the stored events from the database again, we use a wrong event ID and try to process an already decrypted event which results in a CoreCrypto wrong epoch error specifically because  the epoch value was incremented but the message is still on a previous epoch.

**Solution**
1. Properly propagating the errors back to the EventDecoder
2. When MLS decryption successful, we ensure the last event ID is updated even if the decrypted events are empty.
3. When MLS decryption fails, we explicitly catch the errors and we do not update the last event ID locally.

### Testing
- [x] When MLS decryption fails, we do not update the last event ID locally
- [x] When MLS decryption succeed with empty decrypted events, we do update the last event ID locally

### Checklist

### DIFF
--- a/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
+++ b/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
@@ -20,7 +20,7 @@ import Foundation
 import WireLogging
 
 extension EventDecoder {
-    
+
     enum Failure: Error {
         case mlsDecodingError
         case missingMLSGroupID
--- a/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
+++ b/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
@@ -22,7 +22,6 @@ import WireLogging
 extension EventDecoder {
 
     enum Failure: Error {
-        case mlsDecodingError
         case missingMLSGroupID
     }
 
@@ -70,10 +69,10 @@ extension EventDecoder {
         }
 
         let decoder = EventPayloadDecoder()
-        guard let payload = try? decoder
-            .decode(Payload.UpdateConversationMLSMessageAdd.self, from: updateEvent.payload) else {
-            throw Failure.mlsDecodingError
-        }
+        let payload = try decoder.decode(
+            Payload.UpdateConversationMLSMessageAdd.self,
+            from: updateEvent.payload
+        )
 
         var conversation: ZMConversation?
         let groupID: MLSGroupID? = await context.perform {
--- a/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
+++ b/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
@@ -56,7 +56,7 @@ extension EventDecoder {
     func decryptMlsMessage(
         from updateEvent: ZMUpdateEvent,
         context: NSManagedObjectContext
-    ) async -> [ZMUpdateEvent] {
+    ) async throws -> [ZMUpdateEvent] {
         WireLogger.mls.info("decrypting mls message")
 
         guard let decryptionService = await context.perform({ context.mlsDecryptionService }) else {
@@ -136,7 +136,7 @@ extension EventDecoder {
 
         } catch {
             WireLogger.mls.warn("failed to decrypt mls message: \(String(describing: error))")
-            return []
+            throw error
         }
     }
 
--- a/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
+++ b/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
@@ -20,6 +20,11 @@ import Foundation
 import WireLogging
 
 extension EventDecoder {
+    
+    enum Failure: Error {
+        case mlsDecodingError
+        case missingMLSGroupID
+    }
 
     func processWelcomeMessage(
         from updateEvent: ZMUpdateEvent,
@@ -67,8 +72,7 @@ extension EventDecoder {
         let decoder = EventPayloadDecoder()
         guard let payload = try? decoder
             .decode(Payload.UpdateConversationMLSMessageAdd.self, from: updateEvent.payload) else {
-            WireLogger.mls.error("failed to decrypt mls message: invalid update event payload")
-            return []
+            throw Failure.mlsDecodingError
         }
 
         var conversation: ZMConversation?
@@ -92,8 +96,7 @@ extension EventDecoder {
         }
 
         guard let groupID else {
-            WireLogger.mls.error("failed to decrypt mls message: missing MLS group ID")
-            return []
+            throw Failure.missingMLSGroupID
         }
 
         let results = try await decryptionService.decrypt(
--- a/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
+++ b/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
@@ -96,48 +96,43 @@ extension EventDecoder {
             return []
         }
 
-        do {
-            let results = try await decryptionService.decrypt(
-                message: payload.data,
-                for: groupID,
-                subconversationType: payload.subconversationType
-            )
+        let results = try await decryptionService.decrypt(
+            message: payload.data,
+            for: groupID,
+            subconversationType: payload.subconversationType
+        )
 
-            if results.isEmpty {
-                WireLogger.mls.info("successfully decrypted mls message but no result was returned")
-                return []
-            }
+        if results.isEmpty {
+            WireLogger.mls.info("successfully decrypted mls message but no result was returned")
+            return []
+        }
 
-            var events = [ZMUpdateEvent]()
-            for result in results {
+        var events = [ZMUpdateEvent]()
+        for result in results {
 
-                switch result {
-                case let .message(decryptedData, senderClientID):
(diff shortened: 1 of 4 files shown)


wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
```diff
@@ -136,7 +136,7 @@ extension EventDecoder {
 
         } catch {
             WireLogger.mls.warn("failed to decrypt mls message: \(String(describing: error))")
-            return []
+            throw error
```
netbe: if we throw here, maybe we remove the log and log the error at higher level if neeeded

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -198,30 +198,46 @@ extension EventDecoder {
...
+                publicKeys: publicKeys,
+                proteusService: proteusService
+            )
 
-        guard !decryptedEvents.isEmpty else {
-            return []
-        }
+            // MLS message decryption operations, even successful, could result in empty decrypted events.
+            // Adding a condition to skip the guard for that specific usecase.
+            guard !decryptedEvents.isEmpty || event.type == .conversationMLSMessageAdd else {
```
netbe, line 210: if we throw we never reached this point, so condition is useless

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -198,30 +198,46 @@ extension EventDecoder {
...
+                }
             }
-        }
 
-        return decryptedEvents
+            return decryptedEvents
+
+        } catch {
+            // MLS message decryption failed
+            return []
```
netbe: log error here

netbe (changes requested):
left some comments, + could you add tests for it
2) update new code if needed

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder.swift
jullianm: If we reach this point we do have successfully decrypted events but these could be empty for the`conversationMLSMessageAdd` event and, in that case, we still want to skip the guard hence the added condition.
So, I'm not sure I see why it is useless or am I missing something ?

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
jullianm: https://github.com/wireapp/wire-ios/pull/2416/commits/8509b4191d9f28ac65cc38221cfe916f30799000

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder.swift
jullianm: https://github.com/wireapp/wire-ios/pull/2416/commits/8509b4191d9f28ac65cc38221cfe916f30799000

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder.swift
jullianm: I'll add some unit tests.

jullianm: added some UTs.
I also added some missing throwing errors.
the whole point is to distinguish a decryption error from a decryption success with empty decrypted events which was not the case before because we were always returning an empty array.
https://github.com/wireapp/wire-ios/pull/2416/commits/ec9628801d959fd75467da55b82c4b3dbbcc8961

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -276,8 +309,8 @@ extension EventDecoder {
                     decryptedEvents.append(event)
 
                 case .conversationMLSMessageAdd:
-                    let events = await decryptMlsMessage(from: event, context: syncMOC)
-                    decryptedEvents.append(contentsOf: events)
+                    let events = try? await decryptMlsMessage(from: event, context: syncMOC)
```
netbe, line 307: are we in deadcode ?

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
```diff
@@ -67,8 +72,7 @@ extension EventDecoder {
         let decoder = EventPayloadDecoder()
         guard let payload = try? decoder
             .decode(Payload.UpdateConversationMLSMessageAdd.self, from: updateEvent.payload) else {
-            WireLogger.mls.error("failed to decrypt mls message: invalid update event payload")
-            return []
+            throw Failure.mlsDecodingError
```
netbe: non-blocking: what about let it throw then like:

```suggestion
            let payload = try decoder
            .decode(Payload.UpdateConversationMLSMessageAdd.self, from: updateEvent.payload)
```

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder.swift
netbe: I might have read too quickly 😅, makes sense

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoderTest.swift
```diff
@@ -508,6 +508,126 @@ extension EventDecoderTest {
...
+            XCTAssert(error == .failedToDecryptMessage)
+            XCTAssertEqual(lastEventIDRepository.storeLastEventID_Invocations.count, 0)
+        }
+
+    }
+
+    func test_MLSEventDecryptionStoresLastEventIdIfDecryptionSuccessWithEmptyResults() async throws {
+
+        // Given
+        var proteusViaCoreCrypto = DeveloperFlag.proteusViaCoreCrypto
```
netbe: I wonder if we still need this by default this is enabled now.

netbe (approved):
looks good, so new code in UpdateEventDecryptor or MLSMessageDecryptor didn't have the issue, right?

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
samwyndham: Yeah I agree, especially because it likely contains helpful information.

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoderTest.swift
```diff
@@ -508,6 +508,126 @@ extension EventDecoderTest {
...
+        }
+
+        // When
+        let decryptedEvents = try await sut.decryptAndStoreEvents([mlsEvent])
+
+        XCTAssert(waitForAllGroupsToBeEmpty(withTimeout: 0.5))
+
+        // Then
+        XCTAssertEqual(decryptedEvents.isEmpty, true) // no decrypted events
+        XCTAssertEqual(
```
samwyndham, line 588: non-blocking: Assert that the event id is correct (if it doesn't take too much code)

samwyndham (approved):
Nice work!

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder.swift
jullianm: The whole block is invoked when DeveloperFlag.proteusViaCoreCrypto.isOn flag is false.
Since PROTEUS_BY_CORECRYPTO_ENABLED is set to true by default and we only seem to set it to false for testing, I would assume this is dead code. 
We still use this flag a lot across the project so there's probably a lot of things to clean up here, should we keep this work for a separated PR though ?

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoderTest.swift
jullianm: yeah but we still have unit tests for the legacy mechanism using key store which sets the flag to false.
These could probably be cleaned up too.

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2416/commits/07c5bb080f856ea73eafcec4c09a8402d8bf55c4

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoderTest.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2416/commits/07c5bb080f856ea73eafcec4c09a8402d8bf55c4

jullianm:
@netbe good point. I spotted an issue in new code, fixed in https://github.com/wireapp/wire-ios/pull/2416/commits/64912d18bbc70919ae02fb9029f746c5c78ac25c
thanks

PR #2436, opened 2025-01-27
fix: crash create app router - WPB-15683

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15683" title="WPB-15683" target="_blank">BugWPB-15683</a>  [iOS] Crash on createAppRootRouter
  </td></table>
  <br />
 

### Issue

We have a crash when trying to create the AppRouter when called by `applicationProtectedDataDidBecomeAvailable`. 

Screenshot 2025-01-27 at 16 41 59

This seems to be caused because the mainWindow is nil and is declared as implicitly unwrapped optional.

### Testing

As of now I haven't been able to reproduce, so I figure out it could be good to add logs and guard for a nil value

---

### Checklist


jullianm (approved):

wire-ios/Wire-iOS/Sources/AppDelegate.swift
```diff
@@ -342,11 +346,16 @@ private extension AppDelegate {
...
-
+        
         guard let sessionManager = createSessionManager() else {
             fatalError("sessionManager is not created")
         }
-
+        
+        guard mainWindow != nil else {
+            WireLogger.appDelegate.critical("no mainWindow this should not be possible at this point")
+            return
```
caldrian, line 357: What about also adding `assertionFailure` so that we will be aware of the issue when debugging?

caldrian (approved):

samwyndham:
Un-assigning myself as the PR has already been reviewed twice. Feel free to add me back if needed.

PR #2445, opened 2025-01-28
fix: rename system message - WPB-14824

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14824" title="WPB-14824" target="_blank">BugWPB-14824</a>  [iOS] No system message if a group name is changed
  </td></table>
  <br />
 

### Issue

When the self user renames a conversation, no system message is shown in his/her conversation.

The reason is that by the time we process the event response of renaming the group, the modifiedKey of the conversation (userDefined) is not present, therefore we don't add the system message.

In the ConversationRequestStrategy the `processEvent` method was calling an async method in a Task that finished after we reset the modifiedKeys in the `ZMUpstreamModifiedObjectSync`.

Solution:

Call the processRenameConverationEvent as a synchronous method so modifiedKeys are reset after the processing is finished.

### Testing

1. be admin of a group on iOS device
2. rename the group
3. check that the system message is inserted
---

### Checklist

### DIFF
--- a/wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationEventProcessor.swift
+++ b/wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationEventProcessor.swift
@@ -67,7 +67,8 @@ public class ConversationEventProcessor: NSObject, ConversationEventProcessorPro
             do {
                 let payload = try eventPayloadDecoder.decode(
                     Payload.ConversationEvent<Payload.UpdateConversationName>.self,
-                    from: event.payload)
+                    from: event.payload
+                )
 
                 processor.processPayload(
                     payload,
@@ -75,7 +76,8 @@ public class ConversationEventProcessor: NSObject, ConversationEventProcessorPro
                     in: context
                 )
             } catch {
-                WireLogger.eventProcessing.error("error processing UpdateConversationName: \(error.localizedDescription)")
+                WireLogger.eventProcessing
+                    .error("error processing UpdateConversationName: \(error.localizedDescription)")
             }
         }
     }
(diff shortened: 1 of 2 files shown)


wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationEventProcessor.swift
```diff
@@ -56,12 +56,23 @@ public class ConversationEventProcessor: NSObject, ConversationEventProcessorPro
...
+    /// - Parameter payload: payload containing the event
+    /// - Note: This method needs to be synchronous because it's used by a request Strategy
+    /// This can be removed once ConversationRequestStrategy is removed
+    func processConversationRenamePayload(_ payload: ZMTransportData) {
         // here's no uuid is needed since we process it directly it's just convenience to get the payload
         if let event = ZMUpdateEvent(fromEventStreamPayload: payload, uuid: nil) {
-            Task {
-                await processConversationEvents([event])
-            }
+            guard let payload = try? eventPayloadDecoder.decode(
```
samwyndham: **question:** Is this failing expected behavior or an error?

netbe: I am not sure, I would say error, I simply copy/pasted the logic we had. I can add a log at least. There's no way as far as I know to retry or report the error to the user from the request strategy

samwyndham (approved):
Nice work!

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationEventProcessor.swift
samwyndham: Yeah I was thinking just to log the error... nothing with UX change.

netbe: done

KaterinaWire (approved):

caldrian (approved):

PR #976, opened 2024-02-12
fix: duplicate participant added system messages on MLS group creation WPB-6382

typfel:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6382" title="WPB-6382" target="_blank">BugWPB-6382</a>  [iOS] MLS Group creation: After creating group with federated contacts, "You have added" messages is duplicated
  </td></table>
  <br />
 

----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [ ] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [ ] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [ ] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

After creating an MLS group, multiple participant added system messages would some times get inserted.

### Causes

MLS groups are created in two steps: first an empty group conversation is created and then the participants are added with an MLS commit. This creates several update events which the creator will process twice if there's no additional activity after the group has been created.

1. Create empty MLS group which  generates `.new-conversation` event
2. Insert participants with commit which generated `member-join` event
3. No further activity
4. App re-connects are process pending events 
5. Processing `new-conversation` event will remove the participants again since the conversation is empty
6. Processing `member-join` event will add the participants again with an additional system message.

### Solutions

Don't process `new-conversation` event if the conversation already exists locally.

#### How to Test

1. Create MLS group with user A
2. Put app into background
3. Put app into foreground

This will cause the events to be processed twice and trigger the scenario described above.

### Attachments (Optional)

Screenshot 2024-02-12 at 12 12 08

----
#### PR Post Submission Checklist for internal contributors (Optional)

 - [ ] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----
#### PR Post Merge Checklist for internal contributors

 - [ ] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #2454, opened 2025-01-29
fix: MLS group not visible and duplicated system messages - WPB-15681

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15681" title="WPB-15681" target="_blank">BugWPB-15681</a>  [General] MLS group created on Web (edge) does not show up on own iOS and invited member's Web clients
  </td></table>
  <br />
 

### Issue

**Issue 1**: When creating a MLS group with Web prod / edge and iOS app is in background, entering the foreground again will not show the MLS group. Reason is that we try to process the `conversation-create` event but we early exit since the conversation already exists locally.

It happens to work when app is in the foreground because usually `conversation-create` arrives before `conversation.mls-welcome`.

--> **Solution 1**: Removing the early exit when a conversation already exists so we can actually proceed the event.

**Issue 2**: The early exit was introduced in the first place to fix another issue related to duplicated system message when a participant is added to a conversation. See 'Causes' section in https://github.com/wireapp/wire-ios/pull/976 which provides some details regarding the reason of these duplicated messages. 

--> **Solution 2**: Don't remove participants when a `create-conversation` event is received, so when the app returns to the foreground and events are processed again, `member-join` event will be processed but it will not display a new system message since the participant was already added.

Note: Maybe it's not needed at all to call the `updateMembers` method for a `create-conversation` event since this is handled by the `member-join` or `member-leave` events however I'm not 100% confident this will not raise other bugs.

### Testing

For issue 1 : 
- Put iOS app in background
- Create MLS group conversation on Web prod / edge
- Put the iOS app in foreground, MLS group should appear

For issue 2: 
- on iOS app, create a MLS group and add a participant
- go to background
- go to foreground, there should be a single message `You added ..." 

---

### Checklist


johnxnguyen (commented):
It looks like your PR has unrelated changes.

jullianm:
because I branched out first from 3.116 and changed the target branch to 3.117 afterwards.
the unrelated commit was from this PR : https://github.com/wireapp/wire-ios/pull/2390
I reverted it. thanks @johnxnguyen

wire-ios-request-strategy/Tests/Sources/Payloads/Processing/ConversationEventPayloadProcessorTests.swift
```diff
@@ -60,56 +60,6 @@ final class ConversationEventPayloadProcessorTests: MessagingTestBase {
 
     // MARK: - Process NewConversation Event
 
-    func testProcessPayload_NewConversation_IgnoredWhenConversationAlreadyExists() async throws {
```
jullianm, line 63: no longer needed because we removed the early exit code

wire-ios-request-strategy/Tests/Sources/Payloads/Processing/ConversationEventPayloadProcessorTests.swift
```diff
@@ -60,56 +60,6 @@ final class ConversationEventPayloadProcessorTests: MessagingTestBase {
...
-        // When
-        await sut.processPayload(eventPayload, in: syncMOC)
-
-        // Then
-        await syncMOC.perform {
-            XCTAssertEqual(self.groupConversation.userDefinedName, initialName)
-        }
-    }
-
-    func testProcessPayload_NewConversation_IgnoredWhenConversationIDIsMissing() async throws {
```
jullianm, line 90: no longer needed because we removed the related early exit code

netbe (approved):
LGTM;) I guess we don't have to report these changes to new code, @jullianm  ?

jullianm:
@netbe good catch. Applied fixes in new code in https://github.com/wireapp/wire-ios/pull/2454/commits/c97e6ea686670c10ceb345adc82fcc9626ec074f
thanks.

wire-ios-request-strategy/Sources/Payloads/Processing/ConversationEventPayloadProcessor.swift
```diff
@@ -528,7 +517,7 @@ struct ConversationEventPayloadProcessor {
             conversation.isPendingInitialFetch = false
             updateAttributes(from: payload, for: conversation, context: context)
             updateMetadata(from: payload, for: conversation, context: context)
-            updateMembers(from: payload, for: conversation, context: context)
+            updateMembers(from: payload, for: conversation, shouldRemoveParticipants: false, context: context)
```
johnxnguyen, line 520: Just so I understand, we don't remove any participants for conversation creation events, but we do when processing member join and member leave events?

jullianm: yes, we no longer remove participants for conversation creation events (to fix the issue mentioned in PR description) but there's still the code that adds participants (in `updateMembers` method)
member join event adds participants and member leave removes them.
I'm not 100% sure it is needed at all to call `updateMembers` method here but I'm afraid if we don't  it'll raise other bugs.

johnxnguyen (approved):

PR #2461, opened 2025-01-30
fix: request loop - WPB-15570

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15570" title="WPB-15570" target="_blank">BugWPB-15570</a>  [iOS] Request loop on commit bundle with canary team
  </td></table>
  <br />
 

### Issue

**Context**: The app is in a weird state where it seems to keep syncing over and over again.

**Reasons:** When we try to update the key material, the following error 
`CommitError.failedToSendCommit(recovery: .retryAfterQuickSync, cause: .mlsCommitMissingReferences)` is thrown from `sendCommitBundle` method in the CommitSender. 
Then this error is propagated back to the MLSService inside the `retryOnCommitFailure` method which is recursive for certain error types.

**Issue**: Since this function is recursive, if we keep failing with certain error types (which what happened based on the logs), we end up with an infinite loop.

**Solution**: Break the infinite loop by providing an exit after a few attempts.

Note: I also handled another potential infinite loop issue on another error case (the external commit one)

### Testing

- [x] Simulate an infinite loop by keep throwing the same error, ensure we break from the loop at some point.

### Checklist


netbe (approved):
LGTM;)

KaterinaWire (approved):

samwyndham:
Removed myself as there are already 2 approvals

PR #2462, opened 2025-02-03
fix: request loop sending proteus message - WPB-15862

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15862" title="WPB-15862" target="_blank">BugWPB-15862</a>  Request loop issue
  </td></table>
  <br />
 

### Issue

Also related to this ticket: https://wearezeta.atlassian.net/browse/WPB-5454

Context: A request loop sometimes occur when attempting to send a Proteus message

Causes: When the network response is of type `tryAgainLater` and message is not expired, we fall into a case where we're trying to send the proteus message again using the `sendMessage` method recursively which ends up in a request loop when it keeps failing. It is specifically caused by returning an empty array instead of explicitly throwing an error.

Solution: Don't return an empty array and throw the error explicitly, then catch any MessageSendError occurring and attempt to send message again a few times, then break the loop by throwing the actual error so we don't end up with an infinite loop.

### Testing

- [x] Keep throwing an error should break the loop and throws the actual error.

### Checklist


KaterinaWire (approved):

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
```diff
@@ -27,6 +27,7 @@ public enum MessageSendError: Error, Equatable {
     case unresolvedApiVersion
     case messageExpired
     case missingProteusService
+    case failed
```
KaterinaWire: suggestion: to give a more meaningful name to the error

netbe: maxRetryReached ?

netbe (approved):
LGTM;)

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
jullianm: fixed in https://github.com/wireapp/wire-ios/pull/2462/commits/6241f03874ea6861418e0285b2aaf7fef252a746, difficult to qualify specifically the type of error for that specific use case so we simply add an associated value that propagates back the associated error.

PR #2467, opened 2025-02-03
fix: sharing assets on MLS conversations fails - WPB-15797

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15797" title="WPB-15797" target="_blank">BugWPB-15797</a>  [iOS] Can't share screenshot via share extension with mls enabled
  </td></table>
  <br />
 

### Issue

**Context**: When trying to share a photo / video on a MLS 1:1 or group conversation, it fails.

**Causes**: MLSService is not setup and injected from the share extension so when trying to share photo/ video, it will invoke `sendMessage` method from `MessageSender` and, at some point, attempt to access the MLSService to no avail and it will throw the `missingMlsService` error.

**Solution**: Setup and inject the MLSService from the share extension.

### Testing

Tested on Canary MLS.
- App can be closed or simply running in background (tested in both cases)
- Pick an asset (photo/video), share it on a 1:1 or group on MLS canary.
- Open the app again, asset should be correctly sent.

BEFORE: Sharing a photo/video would end up with an error (with a pop up and a failure message on the conversation)
AFTER: The shared asset will be sent and it will be properly show up in the MLS 1:1 or group conversation.

### Checklist


wire-ios-share-engine/Sources/SharingSession.swift
```diff
@@ -440,6 +459,17 @@ public final class SharingSession {
             mlsActionExecutor: mlsActionExecutor
         )
 
+        let mlsService = MLSService(
+            context: coreDataStack.syncContext,
+            notificationContext: coreDataStack.syncContext.notificationContext,
+            coreCryptoProvider: coreCryptoProvider,
+            conversationEventProcessor: ConversationEventProcessor(context: coreDataStack.syncContext),
+            featureRepository: FeatureRepository(context: coreDataStack.syncContext),
+            userDefaults: .standard,
```
netbe, line 468: Not related directly to this PR, but it looks weird that we use .standard. 

Looking at the code, it's used to store the date of  keyPackageQueriedTime:
```
 userDefaults.set(Date(), forKey: .keyPackageQueriedTime)
 ```
 
 To me it makes sense that this value is common to all processes (NSE, SharingExt, mainApp

```suggestion
            userDefaults: .shared(),
```
or .applicationGroup

Note: .shared() and .applicationGroup seems to be the same code but not defined at the same level.

wire-ios-share-engine/Sources/SharingSession.swift
```diff
@@ -147,6 +151,14 @@ final class ApplicationStatusDirectory: ApplicationStatus {
 
 }
 
+/// Required by `MLSService` initializer.
+/// No need to fill in the methods as we don't sync resources in the share engine.
+struct SyncStatus: SyncStatusProtocol {
+    func performQuickSync() async {}
+    func resyncResources() {}
+    func forceSlowSync() {}
+}
```
netbe, line 160: true, it looks like we only use performQuickSync in some case to handle error handling.
1) are we 100% sure the error handling case can not happen in Share Ext?
2) maybe better to replace the need for SyncStatusProtocol by just a closure or a smaller protocol that expose quickSync

netbe (approved):
Fix looks good, though there seems to be another issue in mlsService with the userDefaults

wire-ios-share-engine/Sources/SharingSession.swift
jullianm: Are you sure ? Because looking at the MLSService initialization in `ZMUserSessionBuilder` part of WireSyncEngine, we use the .standard UserDefaults, so it makes sense to use the same, no ?

wire-ios-share-engine/Sources/SharingSession.swift
```diff
@@ -147,6 +151,14 @@ final class ApplicationStatusDirectory: ApplicationStatus {
 
 }
 
+/// Required by `MLSService` initializer.
+/// No need to fill in the methods as we don't sync resources in the share engine.
+struct SyncStatus: SyncStatusProtocol {
```
samwyndham, line 156: **thought:** `SyncStatusProtocol` is confusingly named in my opinion - I wonder if we can come up with something better (obviously this is out of scope for this PR)

wire-ios-share-engine/Sources/SharingSession.swift
samwyndham: I like @netbe `2` suggestion.

wire-ios-share-engine/Sources/SharingSession.swift
samwyndham: I think this should be investigated. My understanding is that `.standard` will be a different store in the main app and the share extension. I assume it is important that the same store is used.

samwyndham (commented):
Nice work! I think the user default comments should be investigated before approving.

wire-ios-share-engine/Sources/SharingSession.swift
jullianm: 1. It can definitely happen, in that case it will throw the error and the sharing will end up failing (pop up + failed message in conversation as usual) 

-> in MLSService code, if an error occurs, it will **not** perform a quick sync since we defined empty methods but still perform the rest of the code.
The main reason these quick sync methods are empty is to stick with existing behavior in SharingSession - we don't resync resources (legacy method `requestResyncResources` in SharingSession is empty as well)

2. Maybe I'm missing your point but the MLSService is expecting an object conforming to SyncStatusProtocol
Do you mean these methods should not be empty ?

wire-ios-share-engine/Sources/SharingSession.swift
jullianm: I agree, the protocol name is confusing and does not reflect well what the object is actually doing.

jullianm: Just to make sure there's no confusion though, this is an existing protocol from legacy codebase and is what is expected by the MLSService.

wire-ios-share-engine/Sources/SharingSession.swift
jullianm: Update: I think you're right, we cannot use the .standard to share data between a host app and its app extension (as also mentioned by Sam)

wire-ios-share-engine/Sources/SharingSession.swift
jullianm: As discussed with @netbe, this will be handled in another ticket, probably https://wearezeta.atlassian.net/browse/WPB-15956, resolving the discussion

wire-ios-share-engine/Sources/SharingSession.swift
jullianm: as discussed with @netbe, we can keep the .standard one as this value doesn't need to be shared across targets
@samwyndham

samwyndham (approved):

PR #2473, opened 2025-02-04
feat: Get domain registration from backend - WPB-15916

KaterinaWire:
### Issue

This PR includes:
- add API v8 support,
- implement new [`POST /get-domain-registration` endpoint](https://staging-nginz-https.zinfra.io/v8/api/swagger-ui/#/default/get-domain-registration)

### Testing

---

### Checklist

---


WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/AuthenticationAPIV0.swift
```diff
@@ -0,0 +1,35 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+class AuthenticationAPIV0: AuthenticationAPI, VersionedAPI {
```
samwyndham, line 21: **question:** Considering this endpoint is only valid from `v8` do we to create all these previous versions? Couldn't we just throw an `.unsupportedEndpointForAPIVersion` error in the `AuthenticationAPIBuilder` for any api version less than `v8`?

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/Responses/DomainRegistrationConfigurationV8.swift
```diff
@@ -0,0 +1,44 @@
...
+        case backendUrl = "backend_url"
+        case domainRedirect = "domain_redirect"
+        case dueToExistingAccount = "due_to_existing_account"
+        case ssoCode = "sso_code"
+    }
+
+    public let backendUrl: String?
+    public let domainRedirect: DomainRedirect
+    public let dueToExistingAccount: Bool?
+    public let ssoCode: String?
```
samwyndham: **issue:** Looking at the swagger docs I see the following. Do you know if the swagger docs are correct?

Screenshot 2025-02-04 at 16 05 48

As far as I know the red asterisk means those fields are non optional. Therefore I think `backendUrl` and `ssoCode` should be non optional.

Furthermore I think `ssoCode` would be better as type `UUID` so that we catch any invalid types early.

samwyndham (commented):
Nice work. I have a few questions before approving.

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/AuthenticationAPIV0.swift
KaterinaWire: `AuthenticationAPI` will have more endpoints that will be moved to the WireAPI, and they are available in earlier versions of the API.

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/Responses/DomainRegistrationConfigurationV8.swift
KaterinaWire: You are right about the red asterisk, but it's wrong in Swagger. There is a discussion in the group chat about this (hopefully it will be fixed). In short:
```
sso_code will be null or not preset, if domain_redirect is not set to sso
same goes for backend_url and the domain_redirect value backend
```

I agree about the `ssoCode` type. I will change it, but not in the `DomainRegistrationConfigurationV8` (this is a struct to parse the response). I will change it in the `DomainRegistrationConfiguration`, that will be public and available for use cases.

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/AuthenticationAPIV0.swift
samwyndham: Got it. Thanks

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/Responses/DomainRegistrationConfigurationV8.swift
samwyndham: Thank you for the information.

Regarding `ssoCode` I think it is better done in `DomainRegistrationConfigurationV8` considering `UUID` is decodable for this main reason:

Given that initialization of a UUID from string is failable the following code will now need to be throwing or optional:

```diff
-    func toAPIModel() -> DomainRegistrationConfiguration {
+    func toAPIModel() -> DomainRegistrationConfiguration? {
        DomainRegistrationConfiguration(
            backendUrl: backendUrl,
            domainRedirect: domainRedirect,
            dueToExistingAccount: dueToExistingAccount,
            ssoCode: ssoCode.map { UUID(uuidString: $0) }
        )
    }
```

Furthermore in the code I wrote above this is actually a bug because if there is a `ssoCode` but it is not convertible to a UUI, the correct action is for `toAPIModel()` to return `nil`.

If instead `DomainRegistrationConfigurationV8` parses a UUID directly things are simplified and we remove the optionality.

WireAPI/Sources/WireAPI/Models/Authorization/DomainRegistrationConfiguration.swift
```diff
@@ -36,6 +36,26 @@ public struct DomainRegistrationConfiguration: Equatable, Sendable {
...
+            self.backendUrl = URL(string: backendUrlStr)
+        } else {
+            self.backendUrl = nil
+        }
+        self.domainRedirect = domainRedirect
+        self.dueToExistingAccount = dueToExistingAccount
+        if let ssoCodeStr {
+            self.ssoCode = UUID(uuidString: ssoCodeStr)
+        } else {
+            self.ssoCode = nil
```
samwyndham: I guess this is what I'm referring to in my other comment. If there is a `ssoCodeStr` but it can't be converted to a `UUID` in my opinion `DomainRegistrationConfiguration` is now in an invalid state. 

The same applies to `backendUrlStr` I think

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/Responses/DomainRegistrationConfigurationV8.swift
samwyndham: @KaterinaWire and I discussed this and we will keep the parsing as is to allow the error to propagate to the UI in case we need to surface it to the user. Essentially `DomainRegistrationConfiguration` will be allowed to hold an invalid state that we can then determine & surface in the use case. @KaterinaWire will also add some error logging in that case to surface the issue.

WireAPI/Sources/WireAPI/Models/Authorization/DomainRegistrationConfiguration.swift
samwyndham: Discussed with @KaterinaWire . See my other comment.

samwyndham (approved):
Nice work!

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/Responses/DomainRegistrationConfigurationV8.swift
KaterinaWire: @samwyndham I've added logs

WireAPI/Sources/WireAPI/Models/Authorization/DomainRedirect.swift
```diff
@@ -0,0 +1,26 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+public enum DomainRedirect: String, Decodable, Sendable {
```
johnxnguyen: I would rather have a decodable data type close to the place of decoding, so in this case a `DomainRedirectV8` enum that is private or internal, and having the public model not be decodable. 

The reason is that this model is intended for external consumption and by using it to decode network responses we give it dual responsibility. This creates a great risk that a change to the model for the purpose of external consumption breaks the decoding for the purpose of representing network models.

WireAPI/Sources/WireAPI/Models/Authorization/DomainRegistrationConfiguration.swift
```diff
@@ -0,0 +1,70 @@
...
+import Foundation
+import WireLogging
+
+/// The domain redirect configuration.
+
+public struct DomainRegistrationConfiguration: Equatable, Sendable {
+
+    /// The `URL` of the backend.
+
+    public let backendUrl: URL?
```
johnxnguyen:
```suggestion
    /// The `URL` of the on-prem backend.

    public let backendURL: URL?
```

WireAPI/Sources/WireAPI/Models/Authorization/DomainRegistrationConfiguration.swift
```diff
@@ -0,0 +1,70 @@
...
+
+    /// The configuration value that explains the appropriate login/registration flow.
+
+    public let domainRedirect: DomainRedirect
+
+    /// Whether the email is already in use on the cloud.
+
+    public let dueToExistingAccount: Bool?
```
johnxnguyen:
```suggestion
    public let isCloudAccountAlreadyRegistered: Bool?
```

WireAPI/Tests/WireAPITests/APIs/AuthenticationAPI/Resources/GetDomainRegistrationErrorResponse_InvalidDomain.json
johnxnguyen, line 1: Suggestion: include the api version in the name

johnxnguyen (changes requested):

WireAPI/Tests/WireAPITests/APIs/AuthenticationAPI/Resources/GetDomainRegistrationErrorResponse_InvalidDomain.json
KaterinaWire: Yes, sure. Renamed

WireAPI/Sources/WireAPI/Models/Authorization/DomainRedirect.swift
KaterinaWire: Yes, I agree. I added `DomainRedirectV8` as a decodable internal data type and use it to parse the response. `DomainRedirect` is public and non-decodable

WireAPI/Sources/WireAPI/Models/Authorization/DomainRegistrationConfiguration.swift
```diff
@@ -0,0 +1,70 @@
...
+
+    /// SSO code.
+
+    public let ssoCode: UUID?
+
+    public init(
+        backendUrlStr: String?,
+        domainRedirect: DomainRedirect,
+        dueToExistingAccount: Bool?,
+        ssoCodeStr: String?
```
johnxnguyen:
```suggestion
        backendURLString: String?,
        domainRedirect: DomainRedirect,
        isCloudAccountAlreadyRegistered: Bool?,
        ssoCodeString: String?
```

johnxnguyen (approved):

PR #2485, opened 2025-02-05
fix: unable to persist fetched users - WPB-10801

johnxnguyen:
<table>
<td>
  </td></table>
  <br />
 

### Issue
When fetching users from remote and storing them via `UserLocalStore`, the user would be inserted in the db but with no properties set.

### Cause
When inserting in the db, if the user is deleted on remote, we mark the user as deleted locally and return early (i.e not processing the rest of the payload). The check to understand if the user is deleted is wrong though... it's asserting that an optional boolean is `false`, which means a `nil` value fails the check and we think the user is deleted.

### Solution
Remove optionality and fallback to `false` in the case of `nil` value.

### Testing
Code paths aren't live yet, so can't be tested in app at the moment.

---

### Checklist

---


samwyndham (approved):
Nice bug hunting!

netbe (approved):
Ah Optional<Bool> again 🧹

PR #2488, opened 2025-02-05
fix: replace mlsDecryptionService by mlsService to recover from wrong epoch error - WPB-15573

jullianm:
WPB-15573

### Issue

**Context**

Some MLS messages are missing due to wrong epoch error.

Screenshot 2025-02-05 at 14 39 03

**Probable cause**

 Group epoch and client's local epoch may diverged, in this case no message will be decrypted and will result in a CC `WrongEpoch` error. In iOS codebase, we use the `MLSDecryptionService` which throws the error however we don't apply any recovery strategy: we simply log the error and return an empty update events array.

**Solution**

As mentioned in engineering specs [here](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/678035678/Use+case+recover+from+epoch+getting+out+of+sync#Step-2%3A-Insert-missed-message-system-message), a recovery strategy should be applied when a `WrongEpoch` error is thrown. 
This recovery strategy already exists in `MLSService` but is not currently used. 
Consequently, this PR simply replaces usage of `MLSDecryptionService` by `MLSService` in the `EventDecoder`.
When a `WrongEpoch` error is thrown, it will now try to recover.

**Disclaimer**

 I was not able to reproduce the issue yet so the issue might still be occurring however this PR introduces compliance with the engineering specs. 

### Testing

_Describe how to test._

---

### Checklist

---

### DIFF
--- a/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
+++ b/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
@@ -123,12 +123,11 @@ extension EventDecoder {
 
             case let .proposal(commitDelay):
                 let scheduledDate = (updateEvent.timestamp ?? Date()) + TimeInterval(commitDelay)
-                let mlsService = await context.perform {
+                await context.perform {
                     conversation?.commitPendingProposalDate = scheduledDate
-                    return context.mlsService
                 }
 
-                if let mlsService, updateEvent.source == .webSocket {
+                if updateEvent.source == .webSocket {
                     mlsService.commitPendingProposalsIfNeeded()
                 }
             }
--- a/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
+++ b/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
@@ -63,8 +63,8 @@ extension EventDecoder {
     ) async throws -> [ZMUpdateEvent] {
         WireLogger.mls.info("decrypting mls message")
 
-        guard let decryptionService = await context.perform({ context.mlsDecryptionService }) else {
-            WireLogger.mls.critical("failed to decrypt mls message: mlsDecyptionService is missing")
+        guard let mlsService = await context.perform({ context.mlsService }) else {
+            WireLogger.mls.critical("failed to decrypt mls message: mlsService is missing")
             fatalError("failed to decrypt mls message: mlsService is missing")
         }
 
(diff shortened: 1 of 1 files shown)


wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
```diff
@@ -63,8 +63,8 @@ extension EventDecoder {
     ) async throws -> [ZMUpdateEvent] {
         WireLogger.mls.info("decrypting mls message")
 
-        guard let decryptionService = await context.perform({ context.mlsDecryptionService }) else {
-            WireLogger.mls.critical("failed to decrypt mls message: mlsDecyptionService is missing")
+        guard let mlsService = await context.perform({ context.mlsService }) else {
```
netbe, line 66: **issue:** we get the mlsService from context but as far as I know we don't create mlsService in NSE ~or SharingExtension~ (actually we do 😅), so this will fail.

The quickest would be indeed to replace mlsService here AND inject mlsService in both extension.

I wonder how it is done for new NSE for instance: do we use mlsDecryptionService or mlsService?

Follow up: if we do use mlsService in all processes instead of mlsDecryptionService:
Should we remove mlsDecryptionService from extensions?

netbe (changes requested):
As far I can see we should still need to inject mlsService in NSE

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder+MLS.swift
johnxnguyen: interesting... I only see`context.mlsDecryptionService` st in the sharing session and notification session... so where does the value come from in the main app? Actually, I thought that in main app the value was already the full mls service.

johnxnguyen: Actually it does use the `mlsService` as a fallback: https://github.com/wireapp/wire-ios/blob/bad485567bff915c449d934553691b660dbdb214/wire-ios-data-model/Source/ManagedObjectContext/NSManagedObjectContext%2BMLSService.swift#L40

Which means that the recovery strategy should be employed. Do we not see in the logs that we attempt to recover? Probably something is going wrong in the recovery.

jullianm: @johnxnguyen you're right and we actually do see the log message related to the recovery after wrong epoch, I missed that. Closing this PR

jullianm:
Closing the PR as the changes in this PR doesn't fix anything since we already use the MLSService under the hood.

PR #2493, opened 2025-02-06
refactor: pull resources - WPB-10801

johnxnguyen:
<table>
<td>
  </td></table>
  <br />
 

### Issue

A large part of the initial sync is to pull resources stored on the backend. This includes the self user metadata, self team metadata (if any), metadata of known conversations and users, and so on. In this PR, we add a pull sync that fetches and stores all the relevant data for the self user.

### Testing

Code paths aren't live yet so this can not be tested at the moment.

---

### Checklist

---


WireDomain/Sources/WireDomain/Synchronization/PullResourcesSync.swift
```diff
@@ -0,0 +1,189 @@
...
+    private let pullSelfUserSync: any PullSelfUserSyncProtocol
+    private let pullSelfTeamSync: any PullSelfTeamSyncProtocol
+    private let pullSelfTeamRolesSync: any PullSelfTeamRolesSyncProtocol
+    private let pullSelfTeamMembersSync: any PullSelfTeamMembersSyncProtocol
+    private let pullSelfLegalholdInfoSync: any PullSelfLegalholdInfoSyncProtocol
+    private let pullUserConnectionsSync: any PullUserConnectionsSyncProtocol
+    private let pullAllConversationsSync: any PullAllConversationsSyncProtocol
+    private let pullKnownUsersSync: any PullKnownUsersSyncProtocol
+    private let pullConversationLabelsSync: any PullConversationLabelsSyncProtocol
+    private let pullAllFeatureConfigsSync: any PullAllFeatureConfigsSyncProtocol
```
samwyndham: **thought**: Out of scope but I wonder if we are creating too many protocols considering we only really need two - one that takes a team id and the other that doesn't.

I guess the main benefits are:

- It prevents us from passing in the wrong object.
- It allows the different objects to diverge.

For me the cons are that it requires more boilerplate and prevents some code re-use.

samwyndham (approved):
Nice work!

WireDomain/Sources/WireDomain/Synchronization/PullResourcesSync.swift
```diff
@@ -0,0 +1,189 @@
...
+
+            try await pullUserConnections()
+            try await pullAllConversations()
+
+            // Pulling known users must happen after we've discovered
+            // user ids from user connections and conversations.
+            try await pullKnownUsers()
+
+            try await pullConversationLabels()
+            try await pullFeatureConfigs()
```
samwyndham: **thought:** Are we making good use of the parallelism during networking? Can any of these pulls be done in parallel?

WireDomain/Sources/WireDomain/Synchronization/Protocols/PullSelfTeamMembersSyncProtocol.swift
```diff
@@ -18,8 +18,16 @@
 
 import Foundation
 
+// sourcery: AutoMockable
+/// An object to keep the local self team members up to date
+/// with the remote self team members.
 protocol PullSelfTeamMembersSyncProtocol {
 
+    /// Fetch the team members0from remote, then create or update
```
netbe:
```suggestion
    /// Fetch the team members from remote, then create or update
```

WireLogging/Sources/WireLogging/WireLogger+MessageTime.swift
```diff
@@ -0,0 +1,31 @@
...
+import Foundation
+
+public extension WireLogger {
+
+    func measureTime(label: String, block: () async throws -> Void) async throws {
+        debug("starting \(label)")
+        let start = Date.now
+        try await block()
+        let durationInSeconds = start.timeIntervalSinceNow.magnitude
+        debug("completed \(label) in \(String(format: "%.2f", durationInSeconds)) seconds")
```
netbe, line 28: **suggestion:** use log attributes to have the time exposed for datadog. See related PR: https://github.com/wireapp/wire-ios/pull/2373/files#diff-45d87eea84fc927ef7946561b5b63ae9895be938b19f5d80f93bbc15e48ea6d9
Can be done in a later PR

netbe (approved):
Looks good, one suggestion left and a typo to fix

WireDomain/Sources/WireDomain/Synchronization/PullResourcesSync.swift
johnxnguyen: Do you mean to have one protocol that wraps all the non team related pulls, and one protocol that wraps all the team related pulls?

In which way to do you see it prevents code re-use?

WireDomain/Sources/WireDomain/Synchronization/PullResourcesSync.swift
johnxnguyen: Good question! I think yes some of these could be done in parallel.

johnxnguyen: @samwyndham I created a ticket to investigate this further: https://wearezeta.atlassian.net/browse/WPB-15999

WireLogging/Sources/WireLogging/WireLogger+MessageTime.swift
johnxnguyen: ok, let's revisit the logging together to ensure we have the highest quality logs

johnxnguyen: https://wearezeta.atlassian.net/browse/WPB-16000

WireDomain/Sources/WireDomain/Synchronization/PullResourcesSync.swift
samwyndham: Yes exactly. Essentially in this case most protocols have the same API which in someway seems redundant. The most obvious code reuse example is the mocks - we would only need two types of mocks here (not really an issue as these are generated anyway). In this file we could also generalize the private methods - we could still do this though if these protocols inherited from some base protocols so I guess that isn't a valid argument. Anyway I'm just thinking out loud here. I don't necessarily think it is a good idea

johnxnguyen: Certainly we could reduce the protocols in this sync, but I would still keep individual sync objects (the ones conforming) because the point of splitting them out of the larger repositories was to increase re-use and depend on only what you need.

PR #2497, opened 2025-02-07
feat: begin connecting up detect auth UI - WPB-15839

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15839" title="WPB-15839" target="_blank">TaskWPB-15839</a>  [iOS] Connect Identification View with logic stubs
  </td></table>
  <br />
 

### Issue

This PR is a bit of a mix of changes to **begin** connecting up the detect authentication UI so that we can start running the feature while we develop it. It includes the following changes:

(I'm sorry this PR is so chaotic - I will try to make my next ones more targeted. At least the size isn't as bad as it seems as many lines are xcode project etc)

- Adds a new Xcode project `WirePreviewApps` and app (WireAuthenticationApp) so we can build & run the feature quickly independently of the `wire-ios` project. This might be a temporary project that we can delete after development. I will discuss in developer channel as I expect it to be controversial but I'm going to push that we have it for the short term.
- Merges Thomas' `AuthenticationIdentityInputView` with Johns `DetermineAuthMethodView` keeping the name `DetermineAuthMethodView`.
- Changes how we present the content to make use of a sheet instead of overlay as this seems to be easier to get things working (I discussed this in the design channel).
- Moves the scroll view to `DetermineAuthMethodView` rather than it's presenting container. This pattern means that every view will be able to calculate and define it's content display height by setting a custom height [detent](https://developer.apple.com/documentation/swiftui/presentationdetent/height(_:)). For now I'm setting `[.medium, .large]` so we don't need to think about this. We can improve it when the feature if functionally complete.
- Adds a missing error message feature to the view.
- Adds a temporary spinner to the next button which we can switch out when we have implemented our own custom version.
- Converts a few `@ObservedObject`s to `@StateObject`. I _believe_ if we use `@ObservedObject` we should be retaining these outside of the views, otherwise these may be recreated when views redraw loosing their state.
- Moves state to `DetermineAuthMethodViewModel` so that we can better unit test this. `DetermineAuthMethodViewModel` is still in a very unfinished state so doesn't yet have any tests.
- Fixes `BackgroundView` - currently the view expands beyond the bounds of the screen which can cause some side affects. The fix ensures that it is cropped to the screen bounds.
- Fixes the wire logo incorrect color in dark mode.

### Testing

Run unit tests

---

### Checklist


WireAuthentication/Sources/WireAuthenticationUI/MockDependencies.swift
```diff
@@ -33,12 +33,31 @@ final class MockDependencies {
...
+    }
+
 }
 
 extension MockDependencies: DetermineAuthMethodUseCaseProtocol {
 
-    func invoke(emailOrSSOCode: String) async -> AuthenticationMethod {
-        .login(email: emailOrSSOCode)
+    func invoke(emailOrSSOCode: String) async throws -> AuthenticationMethod {
+        try await Task.sleep(for: .seconds(3))
```
KaterinaWire, line 58: **question:** is it for loader?

WireAuthentication/Sources/WireAuthenticationUI/Views/BackgroundView.swift
```diff
@@ -18,15 +18,18 @@
...
-            .background(.black)
+    public var body: some View {
+        GeometryReader { geometry in
+            Image(ImageResource(name: "background", bundle: .module))
+                .resizable()
+                .scaledToFill()
+                .frame(width: geometry.size.width, height: geometry.size.height)
+                .clipped()
+                .background(.black)
+        }.ignoresSafeArea()
```
KaterinaWire:
```suggestion
        }
        .ignoresSafeArea()
```

WireAuthentication/Sources/WireAuthenticationUI/Views/DetermineAuthMethod/DetermineAuthMethodView.swift
```diff
@@ -17,6 +17,9 @@
 //
 
 import SwiftUI
+import WireDesign
+import WireFoundation
```
KaterinaWire: **suggestion:** I think you don't need to import `Foundation`

KaterinaWire:
@samwyndham Looks good, I just left some suggestions. I saw the screenshots with an error message, but the border is not red, is this intentional? Is there a ticket for error handling?

WireAuthentication/Sources/WireAuthenticationUI/MockDependencies.swift
samwyndham: Yes, this is to check that the loader is working in previews. It will remain in the mock but will be removed from the real use case - i.e. It won't be in production code.

samwyndham:
> @samwyndham Looks good, I just left some suggestions. I saw the screenshots with an error message, but the border is not red, is this intentional? Is there a ticket for error handling?

There is currently no ticket. I will create it and add a todo comment in the code.

WireAuthentication/Sources/WireAuthenticationUI/Views/DetermineAuthMethod/DetermineAuthMethodViewModel.swift
```diff
@@ -27,35 +27,63 @@ public final class DetermineAuthMethodViewModel: ObservableObject {
...
-    func isValidEmailOrSSOCode(_ emailOrSSOCode: String) -> Bool {
-        !emailOrSSOCode.isEmpty
-    }
+    func submitEmailOrSSOCode() {
+        isLoading = true
 
-    func submitEmailOrSSOCode(_ emailOrSSOCode: String) {
-        Task { [router] in
-            let method = await self.determineAuthMethod.invoke(
+        Task { [self] in
```
johnxnguyen, line 55: I suppose if you want to make testing easier we should make the enclosing method `async`, in that case though since the view model is `MainActor`, we should probably create a detached task to inside this method to ensure we don't go off the main thread.

johnxnguyen (approved):

WireAuthentication/Sources/WireAuthenticationUI/Views/DetermineAuthMethod/DetermineAuthMethodViewModel.swift
samwyndham: Good point. I'll take a look at improving this in a follow up PR

WireAuthentication/Sources/WireAuthenticationUI/Views/DetermineAuthMethod/DetermineAuthMethodView.swift
samwyndham: Fixed in https://github.com/wireapp/wire-ios/pull/2497/commits/fa473898ddcb49862a9f46e9561dbcd5c8376b1d

WireAuthentication/Sources/WireAuthenticationUI/Views/BackgroundView.swift
samwyndham: Fixed in https://github.com/wireapp/wire-ios/pull/2497/commits/fa473898ddcb49862a9f46e9561dbcd5c8376b1d

samwyndham:
> > @samwyndham Looks good, I just left some suggestions. I saw the screenshots with an error message, but the border is not red, is this intentional? Is there a ticket for error handling?
> 
> There is currently no ticket. I will create it and add a todo comment in the code.

@KaterinaWire Added a TODO in https://github.com/wireapp/wire-ios/pull/2497/commits/b6469fd9c77c2cf9b7b4298b06303ca3fa90e2af

KaterinaWire (approved):

PR #2499, opened 2025-02-10
feat: parse email or SSO input - WPB-15915

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15915" title="WPB-15915" target="_blank">TaskWPB-15915</a>  [iOS] Parse authentication input
  </td></table>
  <br />
 

### Issue

This PR introduces a `ValidateEmailOrSSOCodeUseCase` and updates the `DetermineAuthMethodViewModel` to use it to enable/disable the next button. It will also be used later in `DetermineAuthMethodUseCase`. 

Validation logic is done in `SSOCodeValidator` & `EmailValidator`. I put the logic there because, at least in the case of `EmailValidator` we may want to use this elsewhere in our app.

#### Discussion of architecture

For me this highlighted an anti-pattern in our current package set-up. Ideally I didn't want to create a `ValidateEmailOrSSOCodeUseCase` as it requires a lot of boilerplate.

- Unnecessary protocol creation in `WireAuthenticationAPI`.
- Mocking of `ValidateEmailOrSSOCodeUseCase` whose implementation is around the same lines of code as the real version and may lead to confusion in previews (e.g. why the button is enabled with invalid input).
- Testing of `ValidateEmailOrSSOCodeUseCase` which tests very little.

I just wanted to use `SSOCodeValidator` & `EmailValidator` directly.

The problem is that this validation logic needs to be used in both `DetermineAuthMethodViewModel` which lives in `WireAuthenticationUI` and `DetermineAuthMethodUseCase` which lives in `WireAuthenticationLogic`.

I think we should consider changing our package structure to either:

- Define a `WireAuthenticationUtilities` package that can be used from both `WireAuthenticationLogic` and `WireAuthenticationUI`.
- Move our view models to `WireAuthenticationLogic` - this will mean that `WireAuthenticationAPI` contains view model protocols rather than use case protocols. The side affect here is that we will be mocking view models which means that when working on previews we would only want to consider appearance, not logic. (also note that using @Published with protocols takes a bit more code)

Anyway we can discuss this elsewhere, but I just want to share that I don't think what I'm doing in this PR is great.

### Testing

Run automated tests

---

### Checklist


WireAuthentication/Sources/WireAuthenticationAPI/Use cases/ValidateEmailOrSSOCodeUseCase.swift
```diff
@@ -0,0 +1,39 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+package protocol ValidateEmailOrSSOCodeUseCaseProtocol: Sendable {
+
+    @MainActor
```
johnxnguyen: question: what's the reasoning to mark this `@MainActor`?

WireAuthentication/Sources/WireAuthenticationLogic/Utilities/EmailValidator.swift
```diff
@@ -0,0 +1,31 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+struct EmailValidator {
```
johnxnguyen: suggestion: make this an enum so we don't need to initialize an instance.

WireAuthentication/Sources/WireAuthenticationLogic/Utilities/SSOCodeValidator.swift
```diff
@@ -16,20 +16,18 @@
 // along with this program. If not, see http://www.gnu.org/licenses/.
 //
 
-import XCTest
-@testable import WireAuthenticationLogic
+import Foundation
 
-final class WireAuthenticationLogicTests: XCTestCase {
+struct SSOCodeValidator {
```
johnxnguyen: suggestion: make this an enum so we don't need to create an instance.

johnxnguyen (commented):
I suggest to bring the architecture topic to the next forum to discuss. Personally, I feel that all logic fired from a view model should be in the form of a use case, but for the rest you have very valid points and I think we need to make some achitectural adjustments.

WireAuthentication/Tests/WireAuthenticationLogicTests/Utilities/EmailValidatorTests.swift
```diff
@@ -0,0 +1,66 @@
...
+    override func tearDown() {
+        sut = nil
+    }
+
+    func testIsValid_withValidEmails() {
+        // given
+        let testCases: [String] = [
+            "MARIA@EXAMPLE.COM",
+            "maria@example.com",
+            "maria@sub-team22.example.com"
```
KaterinaWire: **suggestion:** we can also dd `firstName.secondName@example.com`

WireAuthentication/Sources/WireAuthenticationUI/Views/DetermineAuthMethod/DetermineAuthMethodViewModel.swift
```diff
@@ -83,7 +86,12 @@ public final class DetermineAuthMethodViewModel: ObservableObject {
     // MARK: - Private
 
     private func isValidEmailOrSSOCode() -> Bool {
-        !emailOrSSOCode.isEmpty
+        do {
+            _ = try validateEmailOrSSOCode.invoke(input: emailOrSSOCode.trimmingCharacters(in: .whitespaces))
```
KaterinaWire, line 90: **question:** Can't we write `return try validateEmailOrSSOCode.invoke(input: emailOrSSOCode.trimmingCharacters(in: .whitespaces))` ?

samwyndham: Not in this case because the use case returns a `ValidatedEmailOrSSOCode` not a `Bool`.

WireAuthentication/Sources/WireAuthenticationLogic/Utilities/EmailValidator.swift
samwyndham: Implemented in https://github.com/wireapp/wire-ios/pull/2499/commits/a2201f8df8c1fa1dbb9bae1484f33e90e38c6fb1

WireAuthentication/Sources/WireAuthenticationLogic/Utilities/SSOCodeValidator.swift
samwyndham: Implemented in https://github.com/wireapp/wire-ios/pull/2499/commits/a2201f8df8c1fa1dbb9bae1484f33e90e38c6fb1

WireAuthentication/Sources/WireAuthenticationAPI/Use cases/ValidateEmailOrSSOCodeUseCase.swift
samwyndham: I was actually using it to silence a warning when generating a mock version but have removed this in https://github.com/wireapp/wire-ios/pull/2499/commits/16e7031596017e93eb7b30e1e132515ea30a036b as I agree it is makes no sense for it to be `@MainActor`

WireAuthentication/Tests/WireAuthenticationLogicTests/Utilities/EmailValidatorTests.swift
samwyndham: Implemented in https://github.com/wireapp/wire-ios/pull/2499/commits/a4a82dd74de33937f28b72d585fd175e42d6f876

johnxnguyen (approved):
💯

KaterinaWire (approved):

PR #2507, opened 2025-02-10
fix: perform quick sync in wrong epoch error recovery logic - WPB-15573

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15573" title="WPB-15573" target="_blank">BugWPB-15573</a>  [iOS] Missing messages in canary team
  </td></table>
  <br />
 

### Issue

**Context**: We're sometimes missing MLS messages in a conversation, in app we get the following system message "You haven't used this device for a while. Some messages may not appear here"

**Causes**: Conversation local and remote epochs diverged resulting in a `WrongEpoch` error. 

In this case, we apply a recovery strategy to 1. sync the conversation with backend, 2. checks if we're out of sync (by comparing local and remote epochs) and if we are 3. display the system message. 

However, in this recovery strategy, we omit do fetch possible missed events (typically `conversation.mls-message-add` events) which could be the reason why local epoch count is behind the remote one.

**Solution**: Call `performQuickSync` to fetch and process possible missed events to ensure we're on the latest state.

Recovery strategy steps mentioned in the specs [here](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/678035678/Use+case+recover+from+epoch+getting+out+of+sync)

### Testing

To be tested during MLS playtests 

### Checklist

### DIFF
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -1314,6 +1314,10 @@ public final class MLSService: MLSServiceInterface {
         do {
             logger.info("repairing out of sync conversation... (\(groupID.safeForLoggingDescription))")
 
+            // In case of `WrongEpoch` error, local and remote epochs have diverged so we may have missed events.
+            // This ensures we're on the latest state.
+            await syncStatus.performQuickSync()
+
             guard let conversationInfo = fetchConversationInfo(
                 with: groupID,
                 in: context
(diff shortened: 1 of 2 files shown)


johnxnguyen (approved):

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1314,6 +1314,10 @@ public final class MLSService: MLSServiceInterface {
         do {
             logger.info("repairing out of sync conversation... (\(groupID.safeForLoggingDescription))")
 
+            // In case of `WrongEpoch` error, local and remote epochs have diverged so we may have missed events.
+            // This ensures we're on the latest state.
+            await syncStatus.performQuickSync()
```
netbe, line 1319: **thought:** I wonder what happens if we get wrongEpoch in NSE we won't be able to recover from this.

netbe (approved):
🤞

wire-ios-data-model/Source/MLS/MLSService.swift
jullianm: No, we won't as we currently use the MLSDecryptionService in the NSE which returns an empty array in case of a decryption failure. In that case, we don't update the last event ID locally so when we're back in the foreground we would get a wrong epoch error again and then the recovery strategy applies.  
I guess we could revisit and inject (in another PR?) the MLSService from the NSE so the recovery strategy is not delegated to the main app.

PR #2521, opened 2025-02-12
chore: implement `DetermineAuthMethodUseCase` - WPB-15918

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15918" title="WPB-15918" target="_blank">TaskWPB-15918</a>   [iOS] Implement use case to determine auth method
  </td></table>
  <br />
 

### Issue

This PR implements `DetermineAuthMethodUseCase` as defined in the related ticket. There is one significant change: In the case of an existing cloud user account with the same domain as an on prem backend, we don't apply any routing as defined in the ticket. Instead we throw an error which can be interpreted to by the caller.  This PR also changed the parsing of the email or SSO code. In the case of email we also want the `domain` from the email and in the case of sso code we parse it to `UUID` instead of the `wire-` prefixed string as this is what is needed / returned from the backend.

### Testing

Run unit tests

---

### Checklist


WireAPI/Sources/WireAPI/Models/Authorization/DomainInfo.swift
```diff
@@ -26,4 +26,8 @@ public struct DomainInfo: Equatable, Sendable {
 
     public let configurationURL: URL
 
+    public init(configurationURL: URL) {
```
samwyndham, line 29: To allow instantiation when mocking

WireAuthentication/Package.swift
```diff
@@ -17,7 +17,6 @@ let package = Package(
     ],
     dependencies: [
         .package(name: "WireAPI", path: "../WireAPI"),
-        .package(name: "WireDomainPackage", path: "../WireDomain"),
```
samwyndham, line 20: This package is not used currently

WireAuthentication/Sources/WireAuthenticationAPI/Use cases/ValidateEmailOrSSOCodeUseCase.swift
```diff
@@ -26,8 +26,8 @@ package protocol ValidateEmailOrSSOCodeUseCaseProtocol: Sendable {
 
 package enum ValidatedEmailOrSSOCode: Equatable {
 
-    case email(String)
-    case ssoCode(String)
+    case email(email: String, domain: String)
```
samwyndham, line 29: We need the domain for fallback on older API versions

WireAuthentication/Sources/WireAuthenticationAPI/Use cases/DetermineAuthMethodUseCase.swift
```diff
@@ -21,15 +21,44 @@ import Foundation
...
+
+    case loginViaSSO(code: UUID)
+
+    /// On-prem login, either via email or SSO
+
     case onPremLogin(email: String, backendConfig: URL)
 
 }
+
+public enum DetermineAuthMethodUseCaseFailure: Error, Equatable {
```
samwyndham, line 47: I just added a few cases that I could think of. I expect these to expand / change when we come to implementing the caller. Basically I'm trying to think of those that will be needed by the view model to determine behavior.

WireAuthentication/Sources/WireAuthenticationLogic/Utilities/SSOCodeValidator.swift
```diff
@@ -20,14 +20,15 @@ import Foundation
 
 enum SSOCodeValidator {
 
-    /// Returns `true` if `ssoCode` is a valid SSO code otherwise `false`.
+    /// Returns an SSO `UUID` if `ssoCode` is a valid SSO code otherwise `nil`.
     ///
     /// A valid SSO code is a UUID prefixed by `wire-`.
 
-    static func isValid(ssoCode: String) -> Bool {
+    static func isValid(ssoCode: String) -> UUID? {
```
samwyndham: The backend needs the `UUID` not the `wire-` prefix.

WireAuthentication/Sources/WireAuthentication/Components/RootComponent.swift
```diff
@@ -18,10 +18,17 @@
...
+import WireAPI
 internal import WireAuthenticationUI
 
 class RootComponent: BootstrapComponent {
 
+    public let authenticationAPI: AuthenticationAPI
+
+    init(authenticationAPI: AuthenticationAPI) {
+        self.authenticationAPI = authenticationAPI
+    }
```
johnxnguyen, line 30: Does this need to be injected? Can't we create it on demand?

WireAuthentication/Sources/WireAuthenticationLogic/DetermineAuthMethodUseCase.swift
```diff
@@ -17,17 +17,93 @@
...
+    private let validateEmailOrSSOCode: ValidateEmailOrSSOCodeUseCase
+    private let authenticationAPI: AuthenticationAPI
+
+    package init(validateEmailOrSSOCode: ValidateEmailOrSSOCodeUseCase, authenticationAPI: AuthenticationAPI) {
+        self.validateEmailOrSSOCode = validateEmailOrSSOCode
+        self.authenticationAPI = authenticationAPI
+    }
+
+    @MainActor
+    package func invoke(
```
johnxnguyen: This will occupy the main thread. I think a better way may be to not mark it main actor, and at the call side invoke the use case in a detached task. What do you think?

johnxnguyen (commented):
looks good, just a question about main actor.

WireAuthentication/Sources/WireAuthenticationLogic/DetermineAuthMethodUseCase.swift
```diff
@@ -17,17 +17,93 @@
...
+    private func determineAuthMethod(email: String, domain: String) async throws -> AuthenticationMethod {
+        let configuration: DomainRegistrationConfiguration
+        do {
+            configuration = try await authenticationAPI.getDomainRegistration(forEmail: email)
+        } catch AuthenticationAPIError.unsupportedEndpointForAPIVersion {
+            // Fallback if the API doesn't support the getDomainRegistration endpoint
+
+            do {
+                let onPremConfig = try await authenticationAPI.getOnPremConfigURL(forDomain: domain)
+                return .onPremLogin(email: email, backendConfig: onPremConfig.configurationURL)
```
KaterinaWire, line 78: @johnxnguyen @samwyndham Do we want to use it for clients who are still using API below v8? If so, from a UI perspective, it will show a redirect UI screen. It might be confusing for our old clients to see the redirect backend screen every time. What do you think?

WireAuthentication/Sources/WireAuthenticationLogic/Utilities/SSOCodeValidator.swift
KaterinaWire: **suggestion:** I would change the name of the method if it doesn't return Bool.

WireAuthentication/Sources/WireAuthenticationLogic/DetermineAuthMethodUseCase.swift
samwyndham: Agree. Fixed in https://github.com/wireapp/wire-ios/pull/2521/commits/7f0cac97bf6d4ec9045cad9e504cdd5b3578c9b5

WireAuthentication/Sources/WireAuthentication/Components/RootComponent.swift
samwyndham: @johnxnguyen Yeah I guess so. Is it OK to revisit in a later PR to better understand the networking Stack?

WireAuthentication/Sources/WireAuthenticationLogic/Utilities/SSOCodeValidator.swift
samwyndham: Fixed int [15724f1](https://github.com/wireapp/wire-ios/pull/2521/commits/15724f1c65c76bb5f26aa033508b58b972e4ea38)

WireAuthentication/Sources/WireAuthenticationLogic/DetermineAuthMethodUseCase.swift
samwyndham: @KaterinaWire I still don't understand all of this that well but won't it only show the redirect screen in cases where they have on prem backend, and therefore isn't that the correct behavior? Or you mean it will show it again once the user has already been redirected?

KaterinaWire: let's keep it for now and will test later

KaterinaWire (approved):

WireAuthentication/Sources/WireAuthentication/Components/RootComponent.swift
johnxnguyen: Yes of course

johnxnguyen (approved):

PR #2526, opened 2025-02-12
fix: Proteus messages not being sent anymore - WPB-16067

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16067" title="WPB-16067" target="_blank">BugWPB-16067</a>  [iOS] Unable to send any Proteus messages with 3.118
  </td></table>
  <br />
 

### Issue

**Context**: Proteus messages are not sent anymore on 3.118

**Cause**: we use the legacy Cryptobox instead of CC because of proteusViaCoreCrypto returning false: plist we use to retrieve the flag value (set to true) was removed in previous PR.

**Solution**: add the config value back to plist

### Testing

Proteus messages sending should behave normally

### Checklist


wire-ios-utilities/WireUtilities-Info.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>ProteusByCoreCryptoEnabled</key>
+	<string>$(PROTEUS_BY_CORECRYPTO_ENABLED)</string>
```
netbe, line 6: suggestion: can't we put it in the Xcode project file directly?

wire-ios-utilities/WireUtilities.xcodeproj/project.pbxproj
```diff
@@ -538,6 +541,7 @@
 			buildSettings = {
 				CODE_SIGN_IDENTITY = "";
 				GENERATE_INFOPLIST_FILE = YES;
```
caldrian, line 543: You haven't reverted `GENERATE_INFOPLIST_FILE`, will it still work?

caldrian (approved):

wire-ios-utilities/WireUtilities-Info.plist
jullianm: you mean not relying on the .xcconfig and do something like this `PROTEUS_BY_CORECRYPTO_ENABLED = YES;` directly in the project file ?

wire-ios-utilities/WireUtilities.xcodeproj/project.pbxproj
jullianm: it works in the sense that it retrieves properly the value from the plist yes

wire-ios-utilities/WireUtilities-Info.plist
netbe: `ProteusByCoreCryptoEnabled = $(PROTEUS_BY_CORECRYPTO_ENABLED)` within the Xcode project so not relying on a real info.plist.

netbe: ok nevermind it's what Xcode does

netbe (approved):
Looks good, I understood that Xcode creates the info.plist for additional keys

PR #2527, opened 2025-02-12
refactor: integrate new initial sync - WPB-10801

johnxnguyen:
<table>
<td>
  </td></table>
  <br />
 

### Issue
In this PR we integrate the new initial sync by:
- initializing and injecting `SyncAgent` into the `ZMUserSession`
- replace calls to the `SyncStatus` with calls to `SyncAgent`

If the developer flag `newInitialSync` is off (the default), then the `SyncAgent` will execute legacy code paths via the `SyncStatus`. If it is on, it will run the new initial sync, then continue with the legacy incremental sync.

Integrating the new incremental sync will follow in a subsequent PR.

### Testing
To check the legacy code paths are still in tact, install the app fresh and log in and use the app as normal. To test the new code paths, enabled the flag via the developer menu.

Issues to look for include infinite sync, getting stuck when logging in or registering a client (including deleting a client).

---

### Checklist

---


wire-ios-request-strategy/Sources/Protocols/ApplicationStatus.swift
```diff
@@ -20,11 +20,10 @@ import Foundation
 
 @objc(ZMApplicationStatus)
 public protocol ApplicationStatus: AnyObject {
+
     var synchronizationState: SynchronizationState { get }
     var operationState: OperationState { get }
     var clientRegistrationDelegate: ClientRegistrationDelegate { get }
     var requestCancellation: ZMRequestCancellation { get }
 
-    func requestResyncResources()
```
johnxnguyen, line 28: This was dead.

wire-ios-sync-engine/Source/SessionManager/SessionFactories.swift
johnxnguyen, line 1: Changes here are for injecting dependencies necessary to create the `SyncAgent`.

wire-ios-sync-engine/Source/Synchronization/ApplicationStatusDirectory.swift
```diff
@@ -108,12 +108,4 @@ public final class ApplicationStatusDirectory: NSObject, ApplicationStatus {
...
     }
 
-    public func requestResyncResources() {
-        syncStatus.resyncResources()
-    }
-
-    public func requestQuickSync() {
-        syncStatus.forceQuickSync()
-    }
-
```
johnxnguyen, line 118: This code was dead

wire-ios-sync-engine/Source/UserSession/SyncStatus.swift
```diff
@@ -118,26 +118,32 @@ public class SyncStatus: NSObject, SyncStatusProtocol, SyncProgress {
...
 
     public func forceSlowSync() {
-        // Refetch user settings.
-        ZMUser.selfUser(in: managedObjectContext).needsPropertiesUpdate = true
-        // Reset the status.
-        currentSyncPhase = SyncPhase.fetchingLastUpdateEventID
-        RequestAvailableNotification.notifyNewRequestsAvailable(nil)
-        log("slow sync")
-        syncStateDelegate?.didStartSlowSync()
+        managedObjectContext.perform { [weak self] in
```
johnxnguyen: Previously, these code paths were always executed on the sync context (which is what `managedObjectContext` is). Now these methods may be called from async methods, so we need to ensure we switch to the right queue otherwise we'll get some core data errors.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -770,12 +813,6 @@ public final class ZMUserSession: NSObject {
         }
     }
 
-    // MARK: Network
-
-    public func requestResyncResources() {
-        applicationStatusDirectory.requestResyncResources()
-    }
-
```
johnxnguyen, line 778: This was dead code

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -934,66 +995,72 @@ extension ZMUserSession: ZMSyncStateDelegate {
...
 
-        let selfClient = ZMUser.selfUser(in: syncContext).selfClient()
-
-        if selfClient?.hasRegisteredMLSClient == true {
-            Task {
-                do {
-                    try await mlsService.repairOutOfSyncConversations()
-                } catch {
-                    WireLogger.mls.error("Repairing out of sync conversations failed: \(error)")
+        syncContext.perform { [weak self] in
```
johnxnguyen, line 1033: Before we (bravely) assumed that this method would be invoked on the sync context, now it's not always true, so best to switch to it explicitly.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -934,66 +995,72 @@ extension ZMUserSession: ZMSyncStateDelegate {
...
             self?.isPerformingSync = true
             self?.updateNetworkState()
         }
     }
 
-    public func didFinishQuickSync() {
-        WireLogger.sync.debug("did finish quick sync")
-        processEvents()
+    private func didFinishIncrementalSync() {
+        syncContext.perform { [weak self] in
```
johnxnguyen: Changes here are to switch to the sync context.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1183,17 +1250,12 @@ extension ZMUserSession: ZMClientRegistrationStatusDelegate {
...
         }
 
-        if userClient.hasRegisteredMLSClient {
-            // Before the client was registered as an MLS client,
-            // They wouldn't have been able to migrate any conversations from Proteus to MLS.
-            // So we perform a slow sync to sync the conversations. This will ensure that
-            // the message protocol of each conversation is up-to-date.
-            // The client will then join any MLS groups they haven't joined yet.
-            syncStatus.forceSlowSync()
-        }
```
johnxnguyen, line 1193: I deleted this because now we always trigger an initial sync (a few lines below)

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1057,9 +1057,9 @@ public final class SessionManager: NSObject, SessionManagerType {
         let context = userSession.syncContext
         context.perform {
             if context.readMigrationNeedsSlowSyncFlag() {
-                userSession.syncStatus.forceSlowSync()
+                userSession.triggerInitialSync()
             } else if context.readMigrationNeedsSyncResourcesFlag() {
-                userSession.syncStatus.resyncResources()
+                userSession.triggerResourceSync()
```
netbe: **nitpic:** wouldn't it be?
```suggestion
                userSession.triggerResourcesSync()
```

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -524,8 +525,18 @@ public final class ZMUserSession: NSObject {
...
         }
 
-        if let clientId = selfUserClient?.safeRemoteIdentifier.safeForLoggingDescription {
-            WireLogger.authentication.addTag(.selfClientId, value: clientId)
+        if let selfUserClient {
+            WireLogger.authentication.addTag(.selfClientId, value: selfUserClient.safeRemoteIdentifier.safeForLoggingDescription)
+
+            Task {
+                do {
+                    // Only sync if there is a self client, otherwise we'll perform a initial sync
```
netbe:
```suggestion
                    // Only sync if there is a self client, otherwise we'll perform an initial sync
```

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -747,6 +758,38 @@ public final class ZMUserSession: NSObject {
...
+        Task {
+            do {
+                try await syncAgent.performInitialSync()
+            } catch {
+                WireLogger.sync.error("failed to perform initial sync: \(String(describing: error))")
+            }
+        }
+    }
+
+    public func triggerResourceSync() {
```
netbe:
```suggestion
    public func triggerResourcesSync() {
```

wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DebugActions/DeveloperDebugActionsViewModel.swift
```diff
@@ -42,9 +42,9 @@ final class DeveloperDebugActionsViewModel: ObservableObject {
     private func setupButtons() {
         buttons = [
             .init(title: "Send debug logs", action: sendDebugLogs),
-            .init(title: "Perform quick sync", action: performQuickSync),
-            .init(title: "Resync resources", action: resyncResources),
-            .init(title: "Break next quick sync", action: breakNextQuickSync),
+            .init(title: "Trigger incremental sync", action: triggerIncrementalSync),
+            .init(title: "Trigger resource sync", action: triggerResourceSync),
```
netbe:
```suggestion
            .init(title: "Trigger resources sync", action: triggerResourceSync),
```

WireDomain/Sources/WireDomain/Providers/InitialSyncBuilder.swift
johnxnguyen, line 1: I realized that I was missing a step in the pull resources sync to fetch the mls status, changes here are adding it.

WireDomain/Sources/WireDomain/Synchronization/PullResourcesSync.swift
johnxnguyen, line 1: Here we add the missing step I mentioned earlier.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -905,19 +942,43 @@ extension ZMUserSession: UpdateEventProcessor {
...
+    }
+
+    func syncAgentDidStartLegacyIncrementalSync(_ syncAgent: SyncAgent) {
+        didStartIncrementalSync()
+    }
 
-    public func didStartSlowSync() {
+    func syncAgentDidFinishLegacyIncrementalSync(_ syncAgent: SyncAgent) {
+        didFinishIncrementalSync()
+    }
```
netbe: **thought:** might be good to make these methods async as we rework later the codebase we will be able to  take advantage of it

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -934,66 +995,72 @@ extension ZMUserSession: ZMSyncStateDelegate {
...
     }
 
-    public func didFinishQuickSync() {
-        WireLogger.sync.debug("did finish quick sync")
-        processEvents()
+    private func didFinishIncrementalSync() {
+        syncContext.perform { [weak self] in
+            guard let self else { return }
+            WireLogger.sync.debug("did finish incremental sync")
+            processEvents()
```
netbe, line 1061: **issue:** here we process buffered events in legacy code, isn't the new sync handling those?

**suggestion:** separate the implementation of didFinishIncrementalSync in distinct methods for legacy and new code. That way we can also really wait for the async function in the WaitingGroupTask

netbe (commented):
looks good, adding some thoughts and concerns about processing buffered events in legacy

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: Not yet, this PR is only replacing the slow sync. I was planning to add the new quick sync in a next PR, but I could also add it here

netbe: as long as we don't forget, it's fine in the other PR. I'll approve as it is now

netbe (approved):
LGTM;)

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: done

WireDomain/Sources/WireDomain/Synchronization/PullResourcesSync.swift
```diff
@@ -32,6 +32,7 @@ struct PullResourcesSync: PullResourcesSyncProtocol {
     private let pullKnownUsersSync: any PullKnownUsersSyncProtocol
     private let pullConversationLabelsSync: any PullConversationLabelsSyncProtocol
     private let pullAllFeatureConfigsSync: any PullAllFeatureConfigsSyncProtocol
+    private let pullMLSStatusSync: any PullMLSStatusSyncProtocol
```
johnxnguyen, line 35: Added a missing step.

wire-ios-sync-engine/Source/UserSession/SyncStatus.swift
```diff
@@ -118,26 +118,33 @@ public class SyncStatus: NSObject, SyncStatusProtocol, SyncProgress {
...
 
     public func forceSlowSync() {
-        // Refetch user settings.
-        ZMUser.selfUser(in: managedObjectContext).needsPropertiesUpdate = true
-        // Reset the status.
-        currentSyncPhase = SyncPhase.fetchingLastUpdateEventID
-        RequestAvailableNotification.notifyNewRequestsAvailable(nil)
-        log("slow sync")
-        syncStateDelegate?.didStartSlowSync()
+        managedObjectContext.performAndWait { [weak self] in
```
johnxnguyen, line 121: Calling these on the context's thread since they may be invoked from a task.

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
```diff
@@ -0,0 +1,305 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireAPI
+import WireDataModel
+
+public final class ClientSessionComponent {
```
johnxnguyen, line 23: I want to start having a more structured dependency graph. When the incremental sync arrives, it'll require constructing a lot of various event processors so putting it all in the user session or yet another builder feel like it's contributing to the mess rather than trying to clean it up.

The idea here is to apply the concept of "components" (borrowed from the needle DI framework) to represent nodes in a dependency graph that act as scopes. One of the issues with `ZMUserSession` is that not only acts as a giant scope for all user related objects, but it fails to prevent some ugly state. For instance, the existence of a `ZMUserSession` only guarantees that the user has authenticated, but it does not guarantee the existence of a valid self user client. This is problematic because we should only run initial and incremental sync if there is a valid user client, which means we have to deal with optionals.

To solve this, I decided to make a split:
1. `UserSessionComponent` is the scope where a user is authenticated but may not yet have a client
2. `ClientSessionComponent` is the scope where the user has a registered client

Any object that depends on an authenticated user but not on a self client would belong to the `UserSessionComponent`. This could be for example objects that are used to register the self client. On the other hand, objects that depend on the self client (e.g `InitialSync`) belong to the `ClientSessionComponent`.

I haven't used Needle here, but it would be fairly simple to add it in order to minimize the boilerplate code.

caldrian: question: Couldn't `ClientSessionComponent` depend on `UserSessionComponent`?

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -33,7 +33,7 @@ final class SyncAgent: NSObject {
     private let initialSyncBuilder: any InitialSyncBuilderProtocol
     private let legacySyncStatus: any SyncStatusProtocol
 
-    private var hasPerformedInitialSync: Bool {
+    var hasPerformedInitialSync: Bool {
```
caldrian: suggestion: rename `hasCompletedInitialSync`
Not sure if it this property might be called during ongoing sync, but naming it this way makes it clearer (at least to me) that sync has been completed.

caldrian (approved):

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
netbe: praise: ok, this is a good idea !

> ClientSessionComponent is the scope where the user has a registered client

question: i guess mls or proteus we don't care here, right?
regarding Needle would that simplify the setup right now?

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
```diff
@@ -0,0 +1,305 @@
...
+    private lazy var backendConfigLocalStore: some BackendConfigLocalStoreProtocol = BackendConfigLocalStore(
+        sharedUserDefaults: sharedUserDefaults
+    )
+
+    private lazy var conversationLabelsLocalStore: some ConversationLabelsLocalStore = ConversationLabelsLocalStore(
+        context: syncContext
+    )
+
+    private lazy var conversationLocalStore: some ConversationLocalStoreProtocol = ConversationLocalStore(
+        context: syncContext,
```
netbe, line 127: **thought:** revisit these contexts once this is shipped, there could be a performance boost to have on the fly context

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
```diff
@@ -0,0 +1,305 @@
...
+        userDefaults: sharedUserDefaults
+    )
+
+    // MARK: - Pull syncs
+
+    private lazy var pullAllConversationsSync: some PullAllConversationsSyncProtocol = PullAllConversationsSync(
+        localDomain: localDomain,
+        isFederationEnabled: BackendInfo.isFederationEnabled,
+        isMLSEnabled: BackendInfo.isMLSEnabled,
```
netbe, line 172: question: so if the backendInfo change, I guess we tore down the clientSessionComponent

WireDomain/Sources/WireDomain/Providers/InitialSyncBuilderProtocol.swift
```diff
@@ -26,6 +26,14 @@ public protocol InitialSyncBuilderProtocol {
...
-    func build() throws -> Sync
+    func buildInitialSync() throws -> Sync
+
+}
+
+extension ClientSessionComponent: InitialSyncBuilderProtocol {
+
+    public func buildInitialSync() throws -> any InitialSyncProtocol {
+        initialSync
+    }
```
netbe: suggestion: move to own file ClientSessionComponent+InitialSyncBuilderProtocol.swift

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -934,66 +1030,73 @@ extension ZMUserSession: ZMSyncStateDelegate {
...
-                } catch {
-                    WireLogger.mls.error("Repairing out of sync conversations failed: \(error)")
+        syncContext.perform { [weak self] in
+            guard let self else { return }
+            let selfClient = ZMUser.selfUser(in: syncContext).selfClient()
+
+            if selfClient?.hasRegisteredMLSClient == true {
+                Task {
+                    do {
+                        try await self.mlsService.repairOutOfSyncConversations()
```
netbe, line 1040: **thought:** @jullianm might be good to look at this method as it will endup doing a retryOnCommitFailure if it fails to join

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -747,6 +793,38 @@ public final class ZMUserSession: NSObject {
         }
     }
 
+    // MARK: - Trigger syncing
+
+    public func triggerInitialSync() {
+        WaitingGroupTask(context: syncContext) { [syncAgent] in
```
netbe: **question:** why is the WaitingGroupTask needed here?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1183,17 +1286,15 @@ extension ZMUserSession: ZMClientRegistrationStatusDelegate {
...
-
         let clientId = userClient.safeRemoteIdentifier.safeForLoggingDescription
         WireLogger.authentication.addTag(.selfClientId, value: clientId)
+
+        // The client was just registered and still needs to perform the
+        // initial sync.
+        if let selfClientID = userClient.remoteIdentifier {
+            setUpSyncAgent(clientID: selfClientID)
+            triggerInitialSync()
+        }
```
netbe: **thought:** I wonder if we can factor out this code in a function that called also in init

netbe (approved):
Good changes with the Components, I left some more comments just ping me when I need to look again

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: I'll leave it as it is for now and change it when needed.

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
johnxnguyen: > question: Couldn't `ClientSessionComponent` depend on `UserSessionComponent`?

@caldrian if we use Needle, there is indeed a parent child relationship but we also wouldn't need to pass dependencies along explicitly. I intentionally did the most simplest approach here that would be most compatible with Needle, but didn't want to integrate Needle in this PR. It's also probably worth a last discussion before committing to it.

johnxnguyen: > praise: ok, this is a good idea !
> 
> > ClientSessionComponent is the scope where the user has a registered client
> 
> question: i guess mls or proteus we don't care here, right? regarding Needle would that simplify the setup right now?

@netbe for now I've just cared about the sync stuff and inject in everything else. But yes eventually we should move everything from `ZMUserSession` into these components. So CoreCrypto and relevant classes would be part of `UserClientSession` or some child of it.

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
johnxnguyen: done

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
johnxnguyen: Definitely. I also think the stores need an overhaul... they mostly wrap existing calls and there's a lot of switching back and forth between a context and an async scope.

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
johnxnguyen: Yeah, at the moment it's static, so probably we should rebuild the session. But I also think that we should only resolve backed information once during the lifetime of the app to simplify things. What do you think?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: It was to help tests, but actually I don't think it's necessary. Removing

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: We already do, kind of. The `setUpSyncAgent` method is used in the `setUp` method. Or did you mean something else?

WireDomain/Sources/WireDomain/Providers/InitialSyncBuilderProtocol.swift
johnxnguyen: done 16617699e92bb879265fbee8d82f6f71c2421e2d

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: done 00e55f23b22b81052bf28b649e07e9f1baacd768

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnxnguyen: done

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: done

wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DebugActions/DeveloperDebugActionsViewModel.swift
johnxnguyen: done

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -524,11 +542,39 @@ public final class ZMUserSession: NSObject {
...
+
+                Task {
+                    do {
+                        try await syncAgent?.performSyncIfNeeded()
+                    } catch {
+                        WireLogger.sync.error("failed to perform sync on session setup: \(String(describing: error))")
+                    }
+                }
+            }
         }
```
netbe: @johnxnguyen I mean here is basically the same as:

```

        // The client was just registered and still needs to perform the
        // initial sync.
        if let selfClientID = userClient.remoteIdentifier {
            setUpSyncAgent(clientID: selfClientID)
            triggerInitialSync()
        }
```

netbe (commented):
all good, just adding a last point

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
netbe: see https://github.com/wireapp/wire-ios/pull/2527#discussion_r1963674322

samwyndham:
Un-assigning myself as this PR is already approved. Feel free to add be back

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: ok, done 1c353474f4f619110c6450508ac31182bda0167c

PR #2534, opened 2025-02-13
fix: proteus send message error missing logic - WPB-16116

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16116" title="WPB-16116" target="_blank">BugWPB-16116</a>  [iOS] sent messages from iOS never appear on any other client
  </td></table>
  <br />
 

### Issue

**Context**: first Proteus message on a fresh install doesn't show up

**Cause**: Some logic was missing in this [PR](https://github.com/wireapp/wire-ios/pull/2462/files) to fix a request loop issue

When getting a network error, we'll try handle the proteus failure which either returns a set of missing clients qualified IDs or throws a specific `MessageSendError`.
-> if it throws `MessageSendError`, we'll retry a couple of times and then throws the error (done in [PR](https://github.com/wireapp/wire-ios/pull/2462/files))
-> If we do get the set of client qualified IDs, we do nothing, we don't try to actually resend the message.

**Solution**:  add missing code to send the message again

### Testing

Fresh install the app
Try to send a first message in a group
Message should show up

### Checklist

---

### DIFF
--- a/wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
+++ b/wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
@@ -173,7 +173,7 @@ public final class MessageSender: MessageSenderInterface {
                 .broadcastProteusMessage(message: messageData)
             await handleProteusSuccess(message: message, messageSendingStatus: messageStatus, response: response)
         } catch let networkError as NetworkError {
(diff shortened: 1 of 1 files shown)


netbe (approved):
Looks good, I would suggest to also add a regression test if possible

caldrian (approved):
👏

PR #2547, opened 2025-02-17
feat: Generate SSO link and get default SSO code

KaterinaWire:
### Issue

1. Add **Generate SSO link** (`/sso/initiate-login/\(ssoCode)`), [Swagger](https://staging-nginz-https.zinfra.io/v8/api/swagger-ui/#/default/auth-req-precheck)

Generating SSO link contains 2 steps:
- validated a company login token -> send a request and parse the error response 
- If the previous step was successful, generate the SSO URL for authentication screen on the Web view

For better testing and logic separation, I created `validateLoginToken` and `buildSSOLink` methods, which should be combined in one of the use cases in the next PRs.

2. Add **Get default SSO code"(`/sso/settings`)**, [Swagger](https://staging-nginz-https.zinfra.io/v8/api/swagger-ui/#/default/sso-settings)

### Checklist

---


WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/AuthenticationAPI.swift
```diff
@@ -46,4 +46,27 @@ public protocol AuthenticationAPI {
...
 
+    /// Generate the link to the SSO authentication screen
+    ///
+    /// - Parameters:
+    ///   - baseURL: Backend URL.
+    ///   - ssoCode: SSO code
+    ///   - callbackScheme: The URL scheme that where the callback will be provided.
+    /// - Returns: URL to the SSO authentication screen
+
+    func buildSSOLink(baseURL: URL, ssoCode: UUID, callbackScheme: String) async throws -> URL
```
johnxnguyen: suggestion: move this outside of the API and into its own type. The purpose of an API is to provide a level of abstraction around network communication with the server via its REST api. In my opinion, building a URL that will be used in a browser is out of place here.

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/AuthenticationAPIV0.swift
```diff
@@ -139,4 +139,91 @@ class AuthenticationAPIV0: AuthenticationAPI, VersionedAPI {
...
+
+        components.queryItems = [
+            URLQueryItem(name: URLQueryItem.Key.errorLabel, value: URLQueryItem.Template.errorLabel),
+            URLQueryItem(name: URLQueryItem.Key.validationToken, value: token.transportString())
+        ]
+
+        return components.url!.absoluteString
+    }
+
+    func validateLoginToken(baseURL: URL, ssoCode: UUID) async throws {
```
johnxnguyen: question: Do we need the baseURL here? It's resolved in the `NetworkService` inside the `APIService`.

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/AuthenticationAPIV0.swift
```diff
@@ -139,4 +139,91 @@ class AuthenticationAPIV0: AuthenticationAPI, VersionedAPI {
...
+            .resolvingAgainst(baseURL: baseURL)
+            .build()
+
+        let (_, response) = try await apiService.executeRequest(
+            request,
+            requiringAccessToken: false
+        )
+
+        return try ResponseParser()
+            .success(code: 200)
```
johnxnguyen: suggestion: continue to use `.ok` variant to be consistent.

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/Requests/URLQueryItem+Keys.swift
```diff
@@ -0,0 +1,51 @@
...
+        static let validationToken = "validation_token"
+    }
+
+    enum Template {
+        static let cookie = "$cookie"
+        static let userIdentifier = "$userid"
+        static let errorLabel = "$label"
+    }
+
+}
```
johnxnguyen: suggestion: make these private extension in the same file that they're used, or given them a new namespace so we avoid to crowd `URL` and `URLQueryItem` namespaces with feature specific values.

WireAPI/Sources/WireAPI/Components/ResponseParser.swift
```diff
@@ -115,3 +115,18 @@ struct ResponseParser<Success> {
     }
 
 }
+
+extension ResponseParser {
```
johnxnguyen: suggestion: make this a private extension in the file it's used so we avoid adding a feature specific value to a general namespace.

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/AuthenticationAPI.swift
```diff
@@ -46,4 +46,27 @@ public protocol AuthenticationAPI {
...
+    ///
+    /// - Parameters:
+    ///   - baseURL: The backend to validate SSO code against.
+    ///   - ssoCode: Company login token (SSO code).
+
+    func validateLoginToken(baseURL: URL, ssoCode: UUID) async throws
+
+    /// Get the default SSO code associated with the backend
+
+    func getSSOCode() async throws -> UUID?
```
samwyndham, line 59: question: Does it make sense that the result is optional or do you think there could be an error in the Swagger documentation?

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/AuthenticationAPIV0.swift
```diff
@@ -139,4 +139,91 @@ class AuthenticationAPIV0: AuthenticationAPI, VersionedAPI {
...
+            .build().url
+
+        guard let url else {
+            throw AuthenticationAPIError.SSOLoginError.invalidSSOCode
+        }
+
+        return url
+    }
+
+    private func makeSuccessCallbackString(using token: UUID, callbackScheme: String) -> String {
```
samwyndham: suggestion: Make this and the next method static to avoid implicitly capturing `self`. In practice it is no problem here but I think it is a good practice to follow.

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/AuthenticationAPIV0.swift
KaterinaWire: done

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/AuthenticationAPI.swift
KaterinaWire: This endpoint is already implemented in our code, UUID is optional there (but I don't know why).I also found a test:
```
func testThat200ResponseWithoutDefaultSSOCodeIsProcessedAsValid() {
        let payload: [String: Any] = [:]

        checkThat(
            statusCode: 200,
            isProcessedAs: .success(SSOSettings(ssoCode: nil)),
            payload: payload as ZMTransportData
        )
    }
```
It looks like, that ssoCode can be nil

WireAPI/Sources/WireAPI/Components/ResponseParser.swift
KaterinaWire: done

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/Requests/URLQueryItem+Keys.swift
KaterinaWire: done

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/AuthenticationAPIV0.swift
KaterinaWire: Could you please explain why should we follow it?
It will look like:
```
static private func makeSuccessCallbackString(using token: UUID, callbackScheme: String) -> String {...}
static private func makeFailureCallbackString(using token: UUID, callbackScheme: String) -> String {...}
```
and usage:
```
let successCallback = AuthenticationAPIV0.makeSuccessCallbackString(using: ssoCode, callbackScheme: callbackScheme)
        let errorCallback = AuthenticationAPIV0.makeFailureCallbackString(using: ssoCode, callbackScheme: callbackScheme)
```

samwyndham: Sure. When we do the following further up in the code we implicitly capture `self` (considering self is a reference type).

```swift
.withQueryItem(name: URLQueryItem.Key.successRedirect, value: successCallback)
```

As this capture is short lived it doesn't really matter so please feel free to ignore my comment and resolve this conversation. But in other situations it might lead to a strong reference cycle: Eg. The following might lead to a memory leak which could be avoided if `Foo` was a struct or `onCompletion` was `static`.

```swift
class Operation {
    let completion: (Int) -> Void

    init(completion: @escaping (Int) -> Void) {
        self.completion = completion
    }
}

class Foo {
    var operation: Operation?

    func configure() {
        operation = Operation(completion: onComplete(value:))
    }

    func onComplete(value: Int) {}
}
```

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/AuthenticationAPI.swift
KaterinaWire: I moved it to the `LoginViaSSOViewModel`. Let me know what you think. If you have a suggestion for a better place, I can implement it in the next PRs

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/AuthenticationAPIV0.swift
KaterinaWire: Thank you!
I would like to keep it like this, if you don't have a strong objection. I also moved this method to the LoginViaSSOViewModel (based on John's suggestion)

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/AuthenticationAPIV0.swift
KaterinaWire: Looks like we don't ned it. I will verify 100% when test the complete flow (SSO login). 
I removed it

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/AuthenticationAPIV0.swift
```diff
@@ -133,4 +133,49 @@ class AuthenticationAPIV0: AuthenticationAPI, VersionedAPI {
...
+        var copy = self
+        copy.parseBlocks.append { code, _ in
+            if let error = AuthenticationAPIError.SSOLoginError(responseCode: code) {
+                throw error
+            }
+            return nil
+        }
+        return copy
+    }
+
```
johnxnguyen: suggestion: seeing as this all we need to do is throw an optional error from a code, I would actually suggest to avoid this extension and simply initialize the error inline, i.e don't use the response parser:

```swift
let (_, response) = try await networkService.executeRequest(request)
if let error = AuthenticationAPIError.SSOLoginError(responseCode: response.statusCode) {
  throw error
}
```

Why? Because is simpler and avoids additional boilerplate that won't be reused elsewhere.

WireAPI/Sources/WireAPI/Components/ResponseParser.swift
```diff
@@ -30,10 +30,10 @@ struct ResponseParser<Success> {
         case noParseResult
     }
 
-    private typealias ParseBlock = (Int, Data?) throws -> Success?
+    typealias ParseBlock = (Int, Data?) throws -> Success?
 
     private let decoder: JSONDecoder
-    private var parseBlocks: [ParseBlock]
+    var parseBlocks: [ParseBlock]
```
johnxnguyen: If you follow my previous suggestion about not using the response parser, this could go back to private, which I think creates a better api to the parser.

WireAuthentication/Sources/WireAuthenticationUI/Views/Login/LoginViaSSO/LoginViaSSOViewModel.swift
```diff
@@ -22,4 +22,107 @@ import SwiftUI
...
+package final class LoginViaSSOViewModel: ObservableObject {
+
+    /// Generate the link to the SSO authentication screen
+    ///
+    /// - Parameters:
+    ///   - baseURL: Backend URL.
+    ///   - ssoCode: SSO code
+    ///   - callbackScheme: The URL scheme that where the callback will be provided.
+    /// - Returns: URL to the SSO authentication screen
+    func buildSSOLink(baseURL: URL, ssoCode: UUID, callbackScheme: String) async throws -> URL {
```
johnxnguyen, line 34: question: how do you envision this to be called? When the view appears? And where would the `ssoCode` and `callbackScheme` come from? The view model itself?

I'm wondering if it would be good to put this inside something like `SSOLinkGenerator` that can be invoked from here.

johnxnguyen (approved):
Looks good, I think there could still be some improvements but fine to see them come in later PRs if you plan to do them.

WireAPI/Sources/WireAPI/APIs/AuthenticationAPI/AuthenticationAPIV0.swift
KaterinaWire: applied

WireAPI/Sources/WireAPI/Components/ResponseParser.swift
KaterinaWire: changed

samwyndham (approved):

PR #2551, opened 2025-02-18
feat: implement `LoginViaEmailUseCase` - WPB-15923

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15923" title="WPB-15923" target="_blank">TaskWPB-15923</a>  [iOS] Implement use case to log in via email
  </td></table>
  <br />
 

### Issue

This PR implements `LoginViaEmailUseCase` which has very little functionality - it calls the underlying network API, and maps its error responses to `LoginViaEmailUseCaseFailure`.

One interesting thing is that `LoginViaEmailUseCase` needs to return an `AuthToken` which is defined in `WireAPI`. As we don't want to import `WireAPI` into `WireAuthenticationAPI` I have chosen to  use an associated type. This will be type erased by the `LoginViaEmailViewModel` so not to need generics everywhere. This is easy to do because we don't actually need to read the contents of `AuthToken`, we just need to pass it around.

### Testing

Run automated tests.

---

### Checklist

### DIFF
--- a/WireAuthentication/Sources/WireAuthenticationAPI/Use cases/LoginViaEmailUseCase.swift
+++ b/WireAuthentication/Sources/WireAuthenticationAPI/Use cases/LoginViaEmailUseCase.swift
@@ -30,7 +30,7 @@ public protocol LoginViaEmailUseCaseProtocol {
 
 }
 
-public enum LoginViaEmailUseCaseFailure: Error {
+public enum LoginViaEmailUseCaseFailure: Error, Equatable {
 
     case invalidCredentials
     case twoFactorAuthenticationRequired
--- a/WireAuthentication/Sources/WireAuthenticationAPI/Use cases/LoginViaEmailUseCase.swift
+++ b/WireAuthentication/Sources/WireAuthenticationAPI/Use cases/LoginViaEmailUseCase.swift
@@ -18,11 +18,12 @@
 
 import Foundation
 
-public protocol LoginViaEmailUseCaseProtocol: Sendable {
+public protocol LoginViaEmailUseCaseProtocol {
 
     func invoke(
         email: String,
-        password: String
+        password: String,
+        verificationCode: String?
     ) async throws(LoginViaEmailUseCaseFailure)
 
 }
@@ -30,9 +31,11 @@ public protocol LoginViaEmailUseCaseProtocol: Sendable {
 public enum LoginViaEmailUseCaseFailure: Error {
 
     case invalidCredentials
-    case verificationCodeRequired
+    case twoFactorAuthenticationRequired
+    case twoFactorAuthenticationFailed
     case accountPendingActivation
     case accountSuspended
-    case other(message: String)
+    case noInternet
+    case other
 
 }
--- a/WireAuthentication/Sources/WireAuthenticationAPI/Use cases/LoginViaEmailUseCase.swift
+++ b/WireAuthentication/Sources/WireAuthenticationAPI/Use cases/LoginViaEmailUseCase.swift
@@ -20,11 +20,13 @@ import Foundation
 
 public protocol LoginViaEmailUseCaseProtocol {
 
+    associatedtype AccessToken
+
     func invoke(
         email: String,
         password: String,
         verificationCode: String?
-    ) async throws(LoginViaEmailUseCaseFailure)
+    ) async throws(LoginViaEmailUseCaseFailure) -> ([HTTPCookie], AccessToken)
 
 }
 
--- a/WireAuthentication/Sources/WireAuthenticationLogic/LoginViaEmailUseCase.swift
+++ b/WireAuthentication/Sources/WireAuthenticationLogic/LoginViaEmailUseCase.swift
@@ -32,7 +32,7 @@ public struct LoginViaEmailUseCase: LoginViaEmailUseCaseProtocol {
         email: String,
         password: String,
         verificationCode: String?
-    ) async throws(LoginViaEmailUseCaseFailure)  -> ([HTTPCookie], AccessToken) {
+    ) async throws(LoginViaEmailUseCaseFailure) -> ([HTTPCookie], AccessToken) {
         do {
             return try await authenticationAPI.login(
                 email: email,
--- a/WireAuthentication/Sources/WireAuthenticationLogic/LoginViaEmailUseCase.swift
+++ b/WireAuthentication/Sources/WireAuthenticationLogic/LoginViaEmailUseCase.swift
@@ -17,17 +17,54 @@
 //
 
 import Foundation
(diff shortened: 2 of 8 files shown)


WireAuthentication/Sources/WireAuthenticationAPI/Use cases/LoginViaEmailUseCase.swift
```diff
@@ -18,21 +18,26 @@
 
 import Foundation
 
-public protocol LoginViaEmailUseCaseProtocol: Sendable {
+public protocol LoginViaEmailUseCaseProtocol {
+
+    associatedtype AccessToken
```
johnxnguyen, line 23: questions: does this mean we'll have to type cast it back when we leave `WireAuthentication`?

WireAuthentication/Sources/WireAuthenticationUI/Views/Login/LoginViaEmail/LoginViaEmailViewModel.swift
```diff
@@ -61,9 +61,10 @@ package final class LoginViaEmailViewModel: ObservableObject {
     func submitPassword(_ password: String) {
         Task.detached {
             do {
-                try await self.loginViaEmailUseCase.invoke(
+                _ = try await self.loginViaEmailUseCase.invoke(
```
johnxnguyen, line 65: suggestion: if we already have a jira ticket for completing the login flow then perhaps we can add a todo here.

johnxnguyen (approved):

WireAuthentication/Sources/WireAuthenticationAPI/Use cases/LoginViaEmailUseCase.swift
samwyndham: No it doesn't need to. We can have an init which makes use of generics and does some type erasure in it to get things working. The downside is that it is a little more complex **to read** than other options:

- Create a duplicate type of `AccessToken` in `WireAuthenticationAPI`.
- Move `AccessToken` somewhere else and import this in `WireAuthenticationAPI`.

WireAuthentication/Sources/WireAuthenticationUI/Views/Login/LoginViaEmail/LoginViaEmailViewModel.swift
samwyndham: Done in https://github.com/wireapp/wire-ios/pull/2551/commits/5a8fc026c01422279d9733d981febb7d12143e86

WireAuthentication/Sources/WireAuthenticationLogic/LoginViaEmailUseCase.swift
```diff
@@ -17,17 +17,54 @@
...
-    ) async throws(LoginViaEmailUseCaseFailure) {}
+        password: String,
+        verificationCode: String?
+    ) async throws(LoginViaEmailUseCaseFailure) -> ([HTTPCookie], AccessToken) {
+        do {
+            return try await authenticationAPI.login(
+                email: email,
+                password: password,
+                verificationCode: verificationCode,
+                label: nil
```
KaterinaWire, line 41: **question:** Not sure about nil. In the current code we pass `UUID()`. Could you take a look at the current implementation?

WireAuthentication/Sources/WireAuthenticationUI/Views/Login/LoginViaEmail/LoginViaEmailViewModel.swift
```diff
@@ -61,12 +61,14 @@ package final class LoginViaEmailViewModel: ObservableObject {
     func submitPassword(_ password: String) {
         Task.detached {
             do {
-                try await self.loginViaEmailUseCase.invoke(
+                // TODO: [WPB-15924] Handle happy path
+                _ = try await self.loginViaEmailUseCase.invoke(
                     email: self.email,
-                    password: password
+                    password: password,
+                    verificationCode: ""
```
KaterinaWire, line 68: **question:** Will have another viewModel where `verificationCode` is not empty?

WireAuthentication/Sources/WireAuthenticationAPI/Use cases/LoginViaEmailUseCase.swift
```diff
@@ -18,21 +18,26 @@
...
 
     case invalidCredentials
-    case verificationCodeRequired
+    case twoFactorAuthenticationRequired
+    case twoFactorAuthenticationFailed
     case accountPendingActivation
     case accountSuspended
-    case other(message: String)
+    case noInternet
+    case other
```
KaterinaWire, line 41: **suggestion:** what do you think about ` .unknown`

samwyndham: I have no strong opinion here. I would have gone with `.unknown` at first but then I thought that @johnxnguyen's `.other` name also makes sense because in many cases we can know the error we just don't want to do anything special with it 🤔. I can update it.

WireAuthentication/Sources/WireAuthenticationUI/Views/Login/LoginViaEmail/LoginViaEmailViewModel.swift
samwyndham: I expect we will use the same. This is really just stubbed code. I will fix it in https://wearezeta.atlassian.net/browse/WPB-15924

WireAuthentication/Sources/WireAuthenticationLogic/LoginViaEmailUseCase.swift
samwyndham: I asked in the SSO group chat. It looks like the backend doesn't really make use of it and it doesn't need to be sent. 

> By looking at backend code this is a cookie label, but I cannot find any usages. Also it is not a required field, so I think you can safely ignore it.

I suggest we keep it `nil` for now.

WireAuthentication/Sources/WireAuthenticationAPI/Use cases/LoginViaEmailUseCase.swift
samwyndham: @KaterinaWire I will change it to `unknown` in my follow up where I invoke the use case.

KaterinaWire: I don't have a strong opinion, we just use `unknown` for many other errors.

KaterinaWire (approved):

PR #2554, opened 2025-02-18
fix: crash on conversation content - WPB-15979

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15979" title="WPB-15979" target="_blank">BugWPB-15979</a>  [iOS] Crashes with many unread messages
  </td></table>
  <br />
 

### Issue

Screenshot 2025-02-18 at 15 17 31

In the crash report associated to the ticket, we can see that the app is killed by the system while fetching messages of the conversation in CoreData while in background.

It's not directly clear if the exhaustion of resources we get is caused by just fetching a lot of messages or processing in another thread messages.

**Solution**

In case of the first one, we could remove the observation of changes in the current conversation when moving to background. 

* Remove fetchController delegate on applicationDidEnterBackground
* Reset fetchController delegate on applicationWillEnterForeground

### Testing

1. spam messages on mls account
2. going back and again in foreground


samwyndham (commented):
question: If the conversations are updated while in the background can things become out of sync or is the data reloaded when we enter the foreground again?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
```diff
@@ -181,6 +182,35 @@ final class ConversationTableViewDataSource: NSObject {
...
+    }
+
+    @objc
+    private func willEnterForeground(_ notification: NSNotification) {
+        fetchController?.delegate = self
+    }
+
+    @objc
+    private func didEnterBackground(_ notification: NSNotification) {
+        fetchController?.delegate = nil
```
johnxnguyen: suggestion: add some comments explaining why we do this.

johnxnguyen (commented):
looks good but I want to see the answer to @samwyndham 's question.

jullianm (approved):
Fix looks good for the background. I think I saw (but I'm not certain) in the attached crash files that sometimes it also occurred in the foreground, there might be a different fix to do if that's the case.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
netbe: done

netbe:
@samwyndham I changed the implementation to refetch when resuming to foreground, so no updates are missed

johnxnguyen (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
```diff
@@ -181,6 +182,45 @@ final class ConversationTableViewDataSource: NSObject {
...
+
+    @objc
+    private func willEnterForeground(_ notification: NSNotification) {
+        // resume updates in that were stop in background
+        fetchController?.delegate = self
+        let contentOffset = tableView.contentOffset
+          do {
+              try fetchController?.performFetch()
+              tableView.reloadData()
+              tableView.setContentOffset(contentOffset, animated: false)
```
samwyndham, line 214: thought: I'm surprised that this is needed - I assume you saw an issue with the content offset changing when calling reload data?

samwyndham (approved):
Looks good. I just added a non-blocking comment.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
netbe: checking this, actually I found another issue: scrolling back up, if the fetchController just performFetch it misses the offset 🫠

netbe:
closing as fix does not work 100%

PR #2558, opened 2025-02-19
feat: assemble network stack within authentication module - WPB-16179

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16179" title="WPB-16179" target="_blank">TaskWPB-16179</a>  [iOS] Set up network stack in Authentication module
  </td></table>
  <br />
 

### Issue

This PR updates `WireAuthentication` assemble it's network stack (e.g. `WireAuthenticationAPI`) internally rather than have those dependencies injected from outside. This is necessary as it may need to recreate the network stack - e.g. when using a proxy or on prem backend etc.

It also configures the `WireAuthenticationApp` to use a live backend (currently Anta). 

### Testing

Run the `WireAuthenticationApp`.

---

### Checklist


WireAuthentication/Sources/WireAuthentication/WireAuthenticationAssembly.swift
```diff
@@ -33,12 +33,16 @@ public struct WireAuthenticationAssembly {
...
+        defaultAPIVersion: APIVersion,
         accountsURL: URL,
         passwordValidator: any PasswordValidator
     ) -> some View {
         RootComponent(
-            authenticationAPI: authenticationAPI,
+            defaultBackendEnvironment: defaultBackendEnvironment,
+            defaultAPIVersion: defaultAPIVersion,
+            minTLSVersion: minTLSVersion,
             accountsURL: accountsURL, // this is temp
```
samwyndham, line 46: question: @KaterinaWire Are you sure `accountsURL` should be temp?

WirePreviewApps/WireAuthenticationApp/Configuration.swift
```diff
@@ -0,0 +1,69 @@
...
+
+struct Configuration {
+
+    let defaultBackendEnvironment: BackendEnvironment
+    let minTLSVersion: TLSVersion
+    let defaultAPIVersion: APIVersion
+    let accountsURL: URL
+    let passwordValidator: any PasswordValidator
+
+    static let live = Configuration(
```
samwyndham, line 32: The idea here is that we can create a bunch of configurations with different mock behaviors while developing.

WireAuthentication/Sources/WireAuthentication/Components/DetermineAuthMethodComponent.swift
```diff
@@ -37,9 +39,16 @@ class DetermineAuthMethodComponent: Component<DetermineAuthMethodComponentDepend
     }
 
     private var determineAuthMethodUseCase: some DetermineAuthMethodUseCaseProtocol {
-        DetermineAuthMethodUseCase(
+        let authenticationAPI = AuthenticationAPIBuilder(
+            networkService: NetworkService.make(
+                backendEnvironment: dependency.defaultBackendEnvironment,
+                minTLSVersion: dependency.minTLSVersion
+            )
+        ).makeAPI(for: dependency.defaultAPIVersion)
```
johnxnguyen: suggestion: while it's probably not needed now, you may want to expose a `networkService` property that can fulfill dependencies in child components for their api clients.

johnxnguyen (approved):

WireAuthentication/Sources/WireAuthentication/Components/DetermineAuthMethodComponent.swift
samwyndham: Yeah, that's a good point. I think I'll leave it until it is needed to have a better understanding of how the pieces come together, considering that sometimes this should be a new network service and sometimes not.

WireAuthentication/Sources/WireAuthentication/WireAuthenticationAssembly.swift
KaterinaWire: `accountsURL` should be a part of the `BackendEnvironment` (it's not yet, but we need to add same as in the Wire-iOS)

KaterinaWire: @samwyndham  You can keep it, and I will fix it in the next PR

KaterinaWire (approved):

PR #2563, opened 2025-02-19
fix: infinite loop and code optimization - WPB-16115

jullianm:
WPB-16115

### Issue

**Context**: When logging out from an MLS account, the account is not deleted because the dispatchGroup notify method still has ongoing operations, specifically `commitPendingProposals` related work which fills in the dispatch group abnormally. Also we have sometimes an infinite loop when trying to recover after failing to commit pending proposals.

**Causes**: First reason is that some pending proposals are planned for the future and we do a `Task.sleep` to wait for that date to commit the proposal, which blocks the groups.
Another reason is that when we retry on a commit failure, we perform a quick sync which (when finishing) triggers the commit pending proposals method again which fills the group again and so on which ends up in a loop.

**Proposed solutions**: first, use dedicated Tasks to not block the group so we can leave it in a reasonable time and second, when we try to recover from a commit error and performing a quick sync, do not try to commit pending proposals again as we're already in the process of recovering these commits. 

This PR is tightly related to this [PR](https://github.com/wireapp/wire-ios/pull/2572) which solely focused on fixing the logout issue while this current PR introduces various code optimization related to the logic of committing pending proposals as well as fixing the loop.

### Testing

No infinite syncing occurs on MLS account

### Checklist

[WPB-16115]: https://wearezeta.atlassian.net/browse/WPB-16115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wire-ios-sync-engine/Source/UserSession/SyncStatus.swift
```diff
@@ -140,7 +141,9 @@ public class SyncStatus: NSObject, SyncStatusProtocol, SyncProgress {
         syncStateDelegate?.didStartSlowSync()
     }
 
-    public func performQuickSync() async {
+    public func performQuickSync(isRecovering: Bool) async {
+        self.isRecovering = isRecovering
```
netbe: issue: if we try to recover, but you go on background foreground (quicksync triggered), it will remove the flag, no?

jullianm:
Converting it back to draft, I've tested again after merging some new work done in 3.118 and now it fails again.. @netbe this won't make it for play test I need to see what's wrong

wire-ios-sync-engine/Source/UserSession/SyncStatus.swift
netbe: maybe better to have one method for recovering where the bool is used only

wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
```diff
@@ -97,7 +97,7 @@ actor EventProcessor: UpdateEventProcessor {
 
             guard !DeveloperFlag.ignoreIncomingEvents.isOn else { return }
 
-            let publicKeys = try? self.earService.fetchPublicKeys()
+            let publicKeys = try? await self.earService.fetchPublicKeys()
```
netbe, line 100: question: something changed here?

jullianm: we were not using await: simple compilation warning but this will be an error in Swift 6

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1757,18 +1769,18 @@ public final class MLSService: MLSServiceInterface {
                             logger.info("commit scheduled in the past, committing...")
                             try await commitPendingProposals(in: groupID)
                         } else {
-                            logger.info("commit scheduled in the future, waiting...")
+                            logger.info("TEST10: commit scheduled in the future, waiting...")
```
netbe: suggestion: might be good to check for task cancellation in the taskGroup and maybe within commitPendingProposals

wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
netbe: but I checked and fetchPublicKeys is not async

-> @netbe, yes this method is not async but still get a compilation warning maybe because the closure this is called from is marked as async

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -984,8 +984,10 @@ extension ZMUserSession: ZMSyncStateDelegate {
                 WireLogger.mls.warn("`qualifiedClientID` is missing for selfClient")
             }
 
-            if mlsFeature.isEnabled {
-                mlsService.commitPendingProposalsIfNeeded()
+            if !isRecovering, mlsFeature.isEnabled {
+                Task {
+                    mlsService.commitPendingProposalsIfNeeded()
+                }
```
netbe: issue: does not need the task here, or make the commitPendingProposalsIfNeeded async

wire-ios-sync-engine/Tests/Source/Synchronization/SynchronizationMocks.swift
```diff
@@ -230,7 +230,7 @@ public class MockSyncStateDelegate: NSObject, ZMSyncStateDelegate {
         didCallStartQuickSync = true
     }
 
-    public func didFinishQuickSync() {
+    public func didFinishQuickSync(isRecovering: Bool) {
```
netbe, line 233: todo: this can be reverted as original method did not change

wire-ios-sync-engine/Source/UserSession/SyncStatus.swift
jullianm: resolving this comment as you already pushed a commit for that

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1749,6 +1760,7 @@ public final class MLSService: MLSServiceInterface {
 
         // Committing proposals for each group is independent and should not wait for
         // each other.
+        logger.info("TEST10: Entering task group")
```
KaterinaWire: **question:** Should "TEST10" logs be merged?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -955,7 +955,7 @@ extension ZMUserSession: ZMSyncStateDelegate {
         }
     }
 
-    public func didFinishQuickSync() {
+    public func didFinishQuickSync(isRecovering: Bool) {
```
KaterinaWire, line 958: **suggestion:** I would add a comment explaining what `isRecovering` is and why we have it. This will help better understand the code in the future and not break it.

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1726,12 +1726,23 @@ public final class MLSService: MLSServiceInterface {
...
+    private var task: Task<Void, Never>?
+    private var lastExecutionTime = Date.distantPast
+    private let throttleInterval: TimeInterval = 2.0 // 2 seconds throttle
+
     public func commitPendingProposalsIfNeeded() {
-        guard let context else {
-            return
+        let now = Date()
+
+        guard now.timeIntervalSince(lastExecutionTime) > throttleInterval else {
```
KaterinaWire, line 1735: **suggestion:** maybe you can use `.now`

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -102,7 +102,7 @@ extension EventDecoder {
             StoredUpdateEvent.highestIndex(self.eventMOC)
         }
 
-        guard proteusProvider.canPerform else {
+        guard proteusProvider.canPerform, !events.isEmpty else {
```
netbe: suggestion: why is this change needed? add another guard as log is misleading.

netbe (changes requested):
See the comments

wire-ios-data-model/Source/MLS/MLSService.swift
jullianm: no, I removed them, thanks

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
jullianm: done thanks

wire-ios-data-model/Source/MLS/MLSService.swift
jullianm: done

wire-ios-data-model/Source/MLS/MLSService.swift
jullianm: it's already done somewhere (I don't remember where exactly though) because we get a CancellationError() thrown in the logs after adding the task cancellation code.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
jullianm: if I mark the `commitPendingProposalsIfNeeded` as async and we use await here, the Task cancellation mechanism within the method will be pointless. 

The reason we used a dedicated Task here is for optimization purpose since this is part of a WaitingGroupTask and we don't want to block the dispatch group because we're doing a Task.sleep when committing "future" pending proposals which could block the group from leaving in a reasonable time which is what this optimization is about. 

Although, we do now have a flag to not call this method again if we're under recovery. 
What do you think ? I guess it's safe to use async/await indeed here since we fixed the logout issue in a different PR and we introduced a flag here to avoid calling that method again if we're under recovery.

wire-ios-sync-engine/Tests/Source/Synchronization/SynchronizationMocks.swift
jullianm: we do need it, no ? because `recoverWithQuickSync` sets the flag to true and when quick sync finishes we call `syncStateDelegate?.didFinishQuickSync(isRecovering: isRecovering)` which avoids calling `commitPendingProposalsIfNeeded` if this flag is true.

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder.swift
jullianm: Not needed, I removed it, thanks

KaterinaWire (approved):

wire-ios-sync-engine/Tests/Source/Synchronization/SynchronizationMocks.swift
netbe: sorry I meant to remove this comment.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
netbe: I think we should remove the task :)

netbe (approved):
LGTM;)

PR #2572, opened 2025-02-20
fix: MLS account not being deleted when logging out - WPB-15774

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15774" title="WPB-15774" target="_blank">BugWPB-15774</a>  [iOS] Logging out consistently doesn't work (Canary)
  </td></table>
  <br />
 

### Issue

Context: When logging out from MLS account (userA) and logging in with other account (userB) and logging out again it will results in weird behavior (crashes, session corrupted popup or login screen showing expired session message..)
-> app behaves inconsistently when logging out from MLS account.

Causes: When logging out from MLS account, we're deleting everything related to the active session except the account itself because some completion block is not called whereas it should be - whether it is a MLS or Proteus account (works fine if we only test with Proteus accounts).

Solution: Ensure the code responsible to delete account data is properly called when logging out from a MLS account by removing usage of dispatch group.

### Testing

Log in with MLS account
Log out
Log in with other account (MLS or Proteus)
Log out
..
-> app should behave consistently when logging in / out from MLS or Proteus accounts (no corrupted session pop up, no crash, no expired session message on login screen..)

### Checklist

### DIFF
--- a/wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
+++ b/wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
@@ -889,17 +889,15 @@ public final class SessionManager: NSObject, SessionManagerType {
 
         requireInternal(activeUserSession.userId == account.userIdentifier, "User session and account are different")
 
-        delegate?.sessionManagerWillLogout(error: error, userSessionCanBeTornDown: { [dispatchGroup] in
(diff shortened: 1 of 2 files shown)


netbe (approved):
🚢

caldrian (approved):

PR #2583, opened 2025-02-24
fix: catch and suppress buffer errors inside transaction - WPB-16231

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16231" title="WPB-16231" target="_blank">BugWPB-16231</a>  [iOS] Catch and suppress buffer errors inside transaction
  </td></table>
  <br />
 

### Issue

When CoreCrypto errors are thrown in non transactional methods, the CoreCryptoContext is not saved, and so buffered messages are not saved. Errors:
* BufferedFutureMessage 
* Other("Incoming message is a commit for which we have not yet received all the proposals. Buffering until all proposals have arrived.")

Use transactional API when decrypting messages.

### Testing

* Updated the unit tests

---

### Checklist


wire-ios-data-model/Source/MLS/MLSActionExecutor.swift
```diff
@@ -318,7 +325,25 @@ public actor MLSActionExecutor: MLSActionExecutorProtocol {
...
-                try await $0.decryptMessage(conversationId: groupID.data, payload: message)
+                let result = try await $0.transaction { context in
+
+                    do {
+                        return try await context.decryptMessage(conversationId: groupID.data, payload: message)
+                    } catch let CoreCryptoError.Mls(error) {
+                        switch error {
+                        case .BufferedFutureMessage, .Other(coreCryptoCommitForMissingProposalError):
+                            // ignore error so transaction is saved and message is saved too.
+                            return nil
```
samwyndham, line 336: question: In this case we will return a `nil` result which means that we will throw a `Failure.bufferedDecryptedMessage` right? Why not throw it here instead?

wire-ios-data-model/Source/Core Crypto/TransactionExecutor.swift
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    func execute(context: CoreCryptoContext) async throws {
+        result = try await block(context)
+    }
+}
+
+extension CoreCryptoProtocol {
+    /// note: here we could return Result instead of Result? but to solve [WPB-16231]
+    /// this is the only way right now, once CC don't throw these errors we can adapt this
```
samwyndham, line 39: question: Can it be the job of whatever instantiates `TransactionExecutor` to determine if it is an optional result?

For example, we could do this if we want an optional result. But maybe I'm missing something:

```swift
let executer = TransactionExecutor<Optional<DecryptedMessage>> {
    // ... 
}
```

wire-ios-data-model/Source/MLS/MLSActionExecutor.swift
```diff
@@ -114,8 +114,15 @@ public protocol MLSActionExecutorProtocol {
 
 /// An actor responsible for performing commits on MLS groups and decrypting messages in a non-reentrant manner.
 
+let coreCryptoCommitForMissingProposalError =
```
samwyndham, line 117: question: This is a little brittle. Will core crypto define a specific error in the future?

samwyndham (commented):
Nice work. Left some comments for clarification

johnxnguyen (approved):
looks good

wire-ios-data-model/Source/MLS/MLSActionExecutor.swift
netbe: because here we're in the coreCrypto transaction, if we throw, the corecryptoContext won't be saved and buffered messages will be discarded. **This is specifically what we want to avoid and fix**.
Then we throw an error outside so we return the empty decryptedmessage

wire-ios-data-model/Source/MLS/MLSActionExecutor.swift
netbe: Not in 3.X.0 version for sure, 4.X version probably

wire-ios-data-model/Source/Core Crypto/TransactionExecutor.swift
netbe: just for the record, here would be the long term solution to have:


```
import WireCoreCrypto

/// Wrapper to execute CoreCrypto transactions
final class TransactionExecutor<Result>: CoreCryptoCommand {
    let block: (_ context: CoreCryptoContext) async throws -> Result
    var result: Result?

    init(
        _ block: @escaping (_ context: CoreCryptoContext) async throws -> Result
    ) {
        self.block = block
    }

    func execute(context: CoreCryptoContext) async throws {
        result = try await block(context)
    }
}

extension CoreCryptoProtocol {
    /// note: here we could return Result instead of Result? but to solve [WPB-16231]
    /// this is the only way right now, once CC don't throw these errors we can adapt this
    func transaction<Result>(
        _ block: @escaping (_ context: CoreCryptoContext) async throws -> Result?
    ) async throws -> Result {
        let transactionExecutor = TransactionExecutor<Optional<Result>>(block)
        try await transaction(command: transactionExecutor)
        return transactionExecutor.result!!
    }
    
    func transaction<Result>(
        _ block: @escaping (_ context: CoreCryptoContext) async throws -> Result
    ) async throws -> Result {
        let transactionExecutor = TransactionExecutor<Result>(block)
        try await transaction(command: transactionExecutor)
        return transactionExecutor.result!
    }

}
```


but since CC will not emit this errors in the future, no need to add this for now.

samwyndham (approved):

PR #2251, opened 2024-12-05
feat: simulcast support - WPB-11481

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11481" title="WPB-11481" target="_blank">StoryWPB-11481</a>  [iOS] Simulcast Support & Automatic Quality Adjustments
  </td></table>
  <br />
 

### Issue

This PR introduces simulcast support. Basically, it requests:
- high quality for 1-1 calls
- low quality in grid layout
- high quality when tile is maximized

Solution:

- avs bumped
- use AVSClientVideoStream instead of AVSClient for wcall_request_video_streams to include quality field.

### Testing

1. make a call with multiple users
2. turn video on
3. maximise a tile

---

### Checklist

- [x] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [x] Description is filled and free of optional paragraphs.
- [x] Adds/updates automated tests.

PR #2586, opened 2025-02-24
chore: enable simulcast - WPB-16137

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16137" title="WPB-16137" target="_blank">TaskWPB-16137</a>  [iOS] Update simulcast for hotfix
  </td></table>
  <br />
 

### Issue

This reverts removing simulcast.
Also move AVS to SPM

Note: this code has already been reviewed #2251.


WireAVS/Package.swift
```diff
@@ -0,0 +1,24 @@
+// swift-tools-version: 6.0
+
+import Foundation
+import PackageDescription
+
+// Temporary local package replacing integration of CoreCrypto via Carthage
+let package = Package(
```
jullianm, line 7: question(non-blocking): why is this temporary ?

jullianm (approved):
LGTM

WireAVS/Package.swift
netbe: because at some point the avs team should host this file, though there's no clear date or how they will do it

wire-ios-sync-engine/Source/Calling/AVS/AVSVideoStreams.swift
```diff
@@ -18,12 +18,50 @@
 
 import Foundation
 
-public struct AVSVideoStreams: Codable, Equatable {
+public struct AVSVideoStreams: Encodable, Equatable {
```
samwyndham, line 21: praise: only generating what we need

samwyndham (approved):
LGTM. Hopefully  we won't need to disabled it again 🤞

PR #2590, opened 2025-02-25
fix: MLS service UT - WPB-16280

jullianm:
WPB-16280

### Issue

Fix UT failing in MLS service with error "stall on main thread". Not exactly sure what happened as I was able to get the test passed multiple times so I simplified the UT, removing the expectations, now the UT runs much quicker.

### Checklist


samwyndham (approved):
Looks good. Is this still a WIP?

netbe (approved):
looks good can be ported to 3.119.0 instead

dmitrysimkin (approved):

PR #2591, opened 2025-02-25
refactor: make components factories - WPB-15944

johnxnguyen:
<table>
<td>
  </td></table>
  <br />
 

### Issue
In order to support injecting dynamic runtime dependencies into components, we need to create a component when we need to build its view. For instance, if we need to build the view for on-prem login, we may want to inject in the on-prem backend config into the component so it can configure a new network stack.

Currently, we create components much earlier than necessary because we inject them (as builders) into the parent view that can navigate to the components views. This PR introduces the following changes:

Introduce `Factory` objects:
- A parent component A that has children components B and C is now the `Factory` of B and C views.
- The view of component A now accepts a `factory` instance which is the component A.
- Since A navigates to B and C views, it uses the factory to produce these views via their corresponding components.

Other changes include:
- Simplify naming: each component has one `view` and `viewModel`.
- Organize component contents: view objects at the top, followed by private and public dependencies, finally children
- Rename component files for consistency.

### Testing
N/A

---

### Checklist

---


KaterinaWire (approved):

samwyndham (approved):

samwyndham:
🚢

PR #2592, opened 2025-02-25
fix: don't remove mls if previously supported mls - WPB-16253

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16253" title="WPB-16253" target="_blank">TaskWPB-16253</a>  [iOS] Prevent removal of MLS from supported protocols if it was previously supported
  </td></table>
  <br />
 

### Issue

If a user supports MLS but at some point adds an old Proteus client, this will break his 1:1 conversations because the supported protocols will result back to proteus (the allMLSClients condition would not be satisfied).

Solution:

Prevent from removing MLS if the user supported mls previously.

### Testing

Note: this will be tested tomorrow for the playtest

---

### Checklist

---

### DIFF
--- a/WireDomain/Sources/WireDomain/UseCases/PushSupportedProtocolsUseCase.swift
+++ b/WireDomain/Sources/WireDomain/UseCases/PushSupportedProtocolsUseCase.swift
@@ -39,7 +39,8 @@ public struct PushSupportedProtocolsUseCase: PushSupportedProtocolsUseCaseProtoc
     let featureConfigRepository: any FeatureConfigRepositoryProtocol
     let pushSupportedProtocolsSync: any PushSupportedProtocolsSyncProtocol
     let userClientsLocalStore: any UserClientsLocalStoreProtocol
-
+    let userLocalStore: any UserLocalStoreProtocol
+    
     private let logger = WireLogger(tag: "supported-protocols")
 
     public func invoke() async throws {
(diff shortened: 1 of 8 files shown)


caldrian (approved):

WireDomain/Sources/WireDomain/UseCases/PushSupportedProtocolsUseCase.swift
```diff
@@ -65,6 +67,11 @@ public struct PushSupportedProtocolsUseCase: PushSupportedProtocolsUseCaseProtoc
             result.insert(.proteus)
         }
 
+        // SelfUser supports mls (other client) at the moment, so we should not remove it
+        if currentSelfUserSupportedProtocols.contains(.mls) {
+            result.insert(.mls)
+        }
+
         /// We support mls if the backend does and all MLS clients are ready.
         if remoteProtocols.contains(.mls), allClientsMLSReady {
```
jullianm, line 76: suggestion: add the condition here to avoid duplicating the insertion code

jullianm (approved):
LGTM

WireDomain/Sources/WireDomain/UseCases/PushSupportedProtocolsUseCase.swift
netbe: @jullianm do you mean the `result.insert(.mls)` ? if yes, it's the same pattern as other cases below.

PR #2605, opened 2025-02-26
chore: move strings to subdirectory - WPB-16303

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16303" title="WPB-16303" target="_blank">BugWPB-16303</a>  [iOS] Unsupported string files from Crowdin
  </td></table>
  <br />
 

### Issue

Many localized string files imported by Crowdin are not used. Some even lead to the AppStore Connect submission being rejected.

This PR 
- moves all string files into a subdirectory `Resources/Localization`,
- deletes unused files
- configurs the crowdin action to only download required files.

---

### Checklist

---


.github/workflows/crowdin.yml
```diff
@@ -30,6 +30,7 @@ jobs:
 
           upload_sources: true
           download_translations: true
+          download_translations_args: "-l ar -l da -l de -l es-ES -l et -l fi -l fr -l it -l ja -l lt -l nl -l pl -l pt-BR -l ru -l sl -l tr -l uk -l zh-CN -l zh-TW"
           upload_translations: false
```
samwyndham, line 34: **thought (non-blocking):** I wonder if we can simplify this workflow and remove the trimming of localizations from string catalogues with this change? Maybe something to do in the future.

.gitignore
```diff
@@ -110,3 +110,29 @@ wire-ios-automation/ios/backendConnections.json
...
+#!**/Resources/Localization/lt.lproj/
+#!**/Resources/Localization/nl.lproj/
+#!**/Resources/Localization/pl.lproj/
+#!**/Resources/Localization/pt-BR.lproj/
+#!**/Resources/Localization/ru.lproj/
+#!**/Resources/Localization/sl.lproj/
+#!**/Resources/Localization/tr.lproj/
+#!**/Resources/Localization/uk.lproj/
+#!**/Resources/Localization/zh-Hans.lproj/
+#!**/Resources/Localization/zh-Hant.lproj/
```
samwyndham: **suggestion (non-blocking):** Personally I think we should remove this as I expect it will stay around as a comment for a long time otherwise and lose it's context.

samwyndham (approved):
Nice work. Thanks for fixing this!

.gitignore
caldrian: will do

.github/workflows/crowdin.yml
caldrian: It would be nice, however, it still doesn't seem to work properly for string catalogs.
Looking at the [result of the action after my changes](https://github.com/wireapp/wire-ios/pull/2609/commits/71ae90218ccb1c185e430c1c5c10c759b59b87c4) we can see, that the string catalog still has all translations.
I guess it is because string catalogs are not natively supported by crowdin, but rather some kind of plugin.

samwyndham: Thanks for checking this!

El-Fitz (approved):
So from now on we should always put new string files in `Resources/Localization/`?

netbe (approved):
looks good, one conflict to resolve to merge it to develop though

caldrian:
> So from now on we should always put new string files in Resources/Localization/?

Since François did this for Wire-iOS I think we should do it everywhere, yes. @El-Fitz

PR #2621, opened 2025-03-03
feat: Implement "email already in use" error alert - WPB-15936

KaterinaWire:
### Issue

If the backend detects that the email is already registered to a cloud account, the backend acts as if the domain was not registered (returning` { domain_registration: none, due_to_existing_account: true }`), allowing the client to continue the log in via password even if the domain was registered.

**Note:**  To ensure that the existence of the account in the Wire cloud is not publicly visible, we should display an alert after the user enters the credentials. Since we move the user to the "No History" screen after a successful response, the alert can be displayed there. There are no strict requirements from the design. @samwyndham @johnxnguyen Let me know if you have any objections.

### Needs releases with:
https://github.com/wireapp/wire-ios-build-assets/pull/148

From the user's perspective, we should display a warning:

Screenshot 2025-03-03 at 09 39 38

### Checklist

---


dmitrysimkin (commented):
👍

WireAuthentication/Sources/WireAuthentication/Components/DetermineAuthMethodComponent.swift
```diff
@@ -95,10 +95,16 @@ class DetermineAuthMethodComponent: Component<DetermineAuthMethodComponentDepend
 
 extension DetermineAuthMethodComponent: DetermineAuthMethodView.Factory {
 
-    func loginViaEmailView(email: String, canCreateAccount: Bool) -> LoginViaEmailView {
+    @MainActor
+    func loginViaEmailView(
+        email: String,
+        canCreateAccount: Bool,
+        emailConflictWithCloudAccount: Bool
```
johnxnguyen: suggestion for here and elsewhere: `didDetectDomainConflict`

WireAuthentication/Sources/WireAuthenticationLogic/DetermineAuthMethodUseCase.swift
```diff
@@ -83,13 +83,15 @@ package struct DetermineAuthMethodUseCase: DetermineAuthMethodUseCaseProtocol {
 
         switch configuration.domainRedirect {
         case .none where configuration.isCloudAccountAlreadyRegistered == true:
-            throw DetermineAuthMethodUseCaseFailure.onPremNotPossible(recovery: .loginViaEmail(email: email))
+            // The email domain has been claimed by an on-prem backend,
```
johnxnguyen, line 86: praise: thanks for this comment

johnxnguyen (approved):

WireAuthentication/Sources/WireAuthenticationUI/Views/Login/NoHistory/NoHistoryView.swift
```diff
@@ -52,19 +57,55 @@ package struct NoHistoryView: View {
...
+                Button(L10n.Authentication.Error.howToDeleteAccount, action: {
+                    viewModel.howToDeleteAccount()
+                })
+                Button(L10n.Authentication.Error.confirm, action: {})
+            }
+        )
+        .onAppear {
+            if viewModel.emailConflictWithCloudAccount {
+                viewModel.alert = .cloudAccountAlreadyRegistered
+            }
```
samwyndham: **suggestion:** In my opinion we should put as much logic as possible into the view model to ensure testability. Rather then this approach you could add a method on appear to the view model and call it here:

```swift
// Inside view model

func onAppear() {
    if viewModel.emailConflictWithCloudAccount {
        alert = .cloudAccountAlreadyRegistered
    }
}
```

samwyndham (approved):
Nice work!

PR #2625, opened 2025-03-04
feat: use new header - WPB-16013

dmitrysimkin:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16013" title="WPB-16013" target="_blank">StoryWPB-16013</a>  [iOS] Header Improvement
  </td></table>
  <br />
 

### Issue

- Use newly created `ConversationTitleView` in `ConversationViewController`.
- Added separate use case since `GetUserAccountImageSourceUseCase` is only for self (logged in) user but not for other participants
- Added user changed observation to update avatar and name changes

### Checklist

---


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/UseCases/GetParticipantImageSourceUseCase.swift
```diff
@@ -0,0 +1,50 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireAccountImageUI
+import WireDataModel
+import WireSyncEngine
+
+class GetParticipantImageSourceUseCase: GetParticipantImageSourceUseCaseProtocol {
```
dmitrysimkin, line 23: I added it near `CovnersationViewController` where it's used but let me know if there is better place

WireUI/Sources/WireConversationUI/ConverationTitleView.swift
```diff
@@ -68,19 +72,52 @@ public class ConversationTitleView: UIView {
...
+    }
+
+    private func updateAvatar(source: AccountImageSource, animated: Bool) {
+        let updateBlock = {
+            self.accountImageView.source = source
+        }
+
+        let animated = true
+
+        if animated, canAnimate {
```
netbe:
```suggestion
        if canAnimate {
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/UseCases/GetParticipantImageSourceUseCase.swift
```diff
@@ -0,0 +1,50 @@
...
+
+    private let userSession: UserSession
+
+    init(userSession: UserSession) {
+        self.userSession = userSession
+    }
+
+    @MainActor
+    func invoke(user: UserType?) async -> WireAccountImageUI.AccountImageSource {
```
samwyndham: **question:** Is a nil `user` some kind of error or is an expected state?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/UseCases/GetParticipantImageSourceUseCase.swift
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        return await withCheckedContinuation { continuation in
+            user.fetchProfileImage(
+                session: userSession,
+                imageCache: UIImage.defaultUserImageCache,
+                sizeLimit: 32,
+                isDesaturated: false
+            ) { image, _ in
+                if let image {
```
samwyndham, line 40: **issue:** I assume we need to handle the case where there is no image, otherwise this method may not return. Perhaps this use case could also throw or return some default?

WireUI/Sources/WireConversationUI/ConverationTitleView.swift
samwyndham, line 1: **nitpick:** Typo in file name

samwyndham (commented):
Nice work! I think there is one issue (see comment) to be fixed then I will take another look

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/UseCases/GetParticipantImageSourceUseCase.swift
dmitrysimkin: This is what I done initially but then there is a glitch in updating avatar. From websocket there are several events come right after image changed, so when first event comes this request always comes with no image so on UI it looks like "old image -> no image (just initials) -> new image (after other events came). Hence avatar is updated not smoothly. 
Same logic is in `UserImageView`, not handling if no image. Hope that make sense @samwyndham 
```
user.fetchProfileImage(
            session: userSession,
            imageCache: UIImage.defaultUserImageCache,
            sizeLimit: size.rawValue,
            isDesaturated: desaturate
        ) { [weak self] image, cacheHit in
            // Don't set image if nil or if user has changed during fetch
            guard let image, user.isEqual(self?.user) else { return }
            self?.setAvatar(.image(image), user: user, animated: !cacheHit)
        }
```

WireUI/Sources/WireConversationUI/ConverationTitleView.swift
dmitrysimkin: well spotted!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/UseCases/GetParticipantImageSourceUseCase.swift
dmitrysimkin: This is because of `conversation.firstActiveParticipantOtherThanSelf` that I pass in which is optional. Not sure how to get rid of optionality here, if you have an idea that would be nice @samwyndham

WireUI/Sources/WireConversationUI/ConversationTitleView.swift
```diff
@@ -68,19 +72,52 @@ public class ConversationTitleView: UIView {
...
+        nameLabel.text = source.title
+        subtitleLabel.text = source.subtitle
+    }
+
+    private func updateAvatar(source: AccountImageSource, animated: Bool) {
+        let updateBlock = {
+            self.accountImageView.source = source
+        }
+
+        let animated = true
```
netbe: **todo:** can be remove

netbe (approved):
Looks good, small todo left

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/UseCases/GetParticipantImageSourceUseCase.swift
samwyndham: I see. Personally I think the caller should do the nil check and:

1. Create the `WireAccountImageUI.AccountImageSource.text("")`  itself.
2. Log an error that the userType is unexpectedly nil.

I don't think this is the responsibility of the use case because it shouldn't really know how it is going to be used. Otherwise all use cases could follow the pattern of accepting nil parameters.

Perhaps this is just a matter of opinion though so feel free to close this comment without changes if you wish.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/UseCases/GetParticipantImageSourceUseCase.swift
samwyndham: I see. There is an issue with this implementation though. See the [documentation](https://developer.apple.com/documentation/swift/withcheckedcontinuation(isolation:function:_:))

> You must invoke the continuation’s resume method exactly once.
Missing to invoke it (eventually) will cause the calling task to remain suspended indefinitely which will result in the task “hanging” as well as being leaked with no possibility to destroy it.

So this use case should either throw or return an optional `WireAccountImageUI.AccountImageSource`

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/UseCases/GetParticipantImageSourceUseCase.swift
dmitrysimkin: Agree that it's just a matter of opinion, but anyway update with your suggestion

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/UseCases/GetParticipantImageSourceUseCase.swift
dmitrysimkin: Nice spot, fixed

samwyndham (approved):
Nice work! 🚢

PR #2629, opened 2025-03-04
feat: On-Prem user login with and without SSO - WPB-15944

KaterinaWire:
<table>
<td>
  </td></table>
  <br />
 

### Issue
This PR implements on-Prem user login. 
The flow starts with fetching backend configs -> displaying info -> attempting to fetch the default SSO code.
If SSO code exists, generate the SSO URL and open the web view. If the SSO code does not exist, show the on-prem login view.
Simulator Screen Recording - iPhone 14 - 2025-03-06 at 15 23 29

### Checklist

---


WireAuthentication/Sources/WireAuthentication/Components/LoginViaEmailOnPremComponent.swift
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @MainActor
+    private func viewModel(
+        email: String,
+        backendEnvironment: BackendEnvironmentInfo
+    ) -> LoginViaEmailOnPremViewModel {
+        LoginViaEmailOnPremViewModel(
+            router: dependency.router,
+            loginViaEmailUseCase: loginViaEmailUseCase,
```
samwyndham, line 71: **note:** As discussed we will need to make instantiating the use case lazy to allow for proxy but I can update this in [[iOS] Login via email - proxy support](https://wearezeta.atlassian.net/browse/WPB-16256)

WireAuthentication/Sources/WireAuthenticationAPI/Use cases/FetchBackendEnvironmentUseCaseProtocol.swift
```diff
@@ -18,18 +18,17 @@
...
-    case unauthenticated(host: String, port: Int)
+}
 
-    /// Settings for an authenticated proxy.
+public enum FetchBackendEnvironmentFailure: Error, Equatable {
 
-    case authenticated(host: String, port: Int, username: String, password: String)
+    case requestFailed
+    case invalidResponse
+    case unknown
```
samwyndham: issue: I guess we need some more info here. The case I'm thinking of is no internet for example. We could add an explicit no internet case or a `case urlError(URLError)` or parse it from `unknown(Error)`). Not sure what is best. I guess we should come up with something consistant.

WireAuthentication/Sources/WireAuthenticationAPI/Use cases/FetchDefaultSSOSettingsUseCaseProtocol.swift
```diff
@@ -18,14 +18,15 @@
...
-    let teamsURL: URL
-    let accountsURL: URL
-    let websiteURL: URL
+    func invoke() async throws(FetchDefaultSSOSettingsUseCaseFailure) -> UUID?
+
+}
+
+public enum FetchDefaultSSOSettingsUseCaseFailure: Error {
+
+    case networkFailure
```
samwyndham: question: In this case do we show the no internet error?

WireAuthentication/Sources/WireAuthenticationLogic/FetchBackendEnvironmentUseCase.swift
```diff
@@ -0,0 +1,58 @@
...
+            WireLogger.backend.info("Fetched custom configuration from \(url)")
+
+            return environmentResponse
+        } catch let decodingError as DecodingError {
+            WireLogger.backend.info("Error decoding response from \(url): \(decodingError)")
+            throw FetchBackendEnvironmentFailure.invalidResponse
+        } catch {
+            WireLogger.backend.error("Error fetching configuration from \(url): \(error)")
+            throw FetchBackendEnvironmentFailure.requestFailed
+        }
```
samwyndham: question: Can this can become the body of the `invoke(at configURL: URL`? Or is it extracted because we will be adding something else to this use case?

WireAuthentication/Sources/WireAuthenticationLogic/FetchDefaultSSOSettingsUseCase.swift
```diff
@@ -17,36 +17,25 @@
...
-    /// The proxy settings for the backend if any.
+    package init(authenticationAPI: AuthenticationAPI) {
+        self.authenticationAPI = authenticationAPI
+    }
 
-    public let proxySettings: LocalProxySettings?
+    package func invoke() async throws(FetchDefaultSSOSettingsUseCaseFailure) -> UUID? {
+        do {
+            return try await authenticationAPI.getSSOCode()
+        } catch FetchDefaultSSOSettingsUseCaseFailure.networkFailure {
```
samwyndham: issue: I don't think this case is possible... E.g. For the authentication API to throw a `FetchDefaultSSOSettingsUseCaseFailure`.

WireAuthentication/Sources/WireAuthenticationAPI/Use cases/FetchDefaultSSOSettingsUseCaseProtocol.swift
```diff
@@ -18,14 +18,15 @@
...
+public protocol FetchDefaultSSOSettingsUseCaseProtocol: Sendable {
 
-    let title: String
-    let backendURL: URL
-    let backendWSURL: URL
-    let blacklistURL: URL
-    let teamsURL: URL
-    let accountsURL: URL
-    let websiteURL: URL
+    func invoke() async throws(FetchDefaultSSOSettingsUseCaseFailure) -> UUID?
```
samwyndham, line 23: question: How does the app react when the UUID is nil? I'm wondering if this should be an error.

WireAuthentication/Sources/WireAuthenticationUI/Views/DetermineAuthMethod/DetermineAuthMethodViewModel.swift
```diff
@@ -143,8 +147,18 @@ package final class DetermineAuthMethodViewModel: ObservableObject {
...
+                        self.modalDestination = .switchBackend(email: email, environment: environmentInfo)
+                    }
+                } catch {
+                    await MainActor.run {
+                        self.alert = .onPremLoginNotPossible(recovery: .loginViaEmail(email: email))
+                    }
+                }
+            }
         }
     }
```
samwyndham: note: Out of scope for this PR but we should revisit this view model and do some clean up:

- We no longer need `didDismissAlert`
- `submitEmailOrSSOCode` has work that should be done with a Task.detached.
- this method should become async and blocking so that things can be tested. 

Lets fix this later.

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackendConfirmation/SwitchBackendConfirmationViewModel.swift
```diff
@@ -26,59 +28,70 @@ public class SwitchBackendConfirmationViewModel {
...
-
+    func confirm() async {
+        isLoading = true
+
+        let fetchDefaultSSOTask = Task.detached { [fetchDefaultSSOSettings] in
+            try await fetchDefaultSSOSettings.invoke()
+        }
+        do {
+            if let ssoCode = try await fetchDefaultSSOTask.value {
+                Task.detached {
```
samwyndham: issue: We want to avoid escaping the the async context because it makes testing `confirm` hard / not possible. To achieve that the result of `Task.detached` should be captured in a variable and then  we try await to get the value, like you are doing above. I expect that everything thing can be done in a single `Task.detached` within `confirm`. Something like the following:

```swift
let task = Task.detached { [fetchDefaultSSOSettings] in // We want to return optional URL
  guard let uuid = try await fetchDefaultSSOSettings.invoke() else {
    return nil  
  }

 return try await self.generateSSOLink(ssoCode: ssoCode)
}

do {
  let url = try await task.value
  if let url {
      // root to sso login
  else {
     // root to on prem login
  }
catch {

}

```

samwyndham (changes requested):
Nice work! I left a few questions/issues regarding error handling and a suggestion for improved testability.

WireAuthentication/Sources/WireAuthentication/Components/DetermineAuthMethodComponent.swift
```diff
@@ -47,6 +47,10 @@ class DetermineAuthMethodComponent: Component<DetermineAuthMethodComponentDepend
         )
     }
 
+    private var fetchBackendEnvironmentUseCase: some FetchBackendEnvironmentUseCaseProtocol {
```
johnxnguyen: suggestion: `FetchBackendConfigUseCaseProtocol`

WireAuthentication/Sources/WireAuthenticationAPI/Models/BackendEnvironmentInfo.swift
```diff
@@ -0,0 +1,141 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public struct BackendEnvironmentInfo: Decodable, Sendable, Hashable {
```
johnxnguyen: suggestion: `BackendConfig`

WireAuthentication/Sources/WireAuthenticationAPI/Models/BackendEnvironmentInfo.swift
```diff
@@ -0,0 +1,141 @@
...
+    ) {
+        self.title = title
+        self.endpoints = endpoints
+        self.proxySettings = proxySettings
+        self.pinnedKeys = pinnedKeys
+    }
+
+}
+
+public struct BackendURLs: Decodable, Sendable, Hashable {
```
johnxnguyen: suggestion: `Endpoints`

WireAuthentication/Sources/WireAuthenticationAPI/Models/BackendEnvironmentInfo.swift
```diff
@@ -0,0 +1,141 @@
...
+
+        guard let certificateKey = SecCertificateCopyKey(certificate) else {
+            throw DecodingError.dataCorruptedError(
+                forKey: CodingKeys.certificateKey,
+                in: container,
+                debugDescription: "Error extracting pinned key from certificate"
+            )
+        }
+        self.certificateKey = certificateKey
+        self.hosts = try container.decode([TrustData.Host].self, forKey: .hosts)
```
johnxnguyen: issue: this code is not necessary when configuring the network service as it is part of the `ServerTrustValidator`. I think we should simply decode the json payload.

WireAuthentication/Sources/WireAuthenticationAPI/Use cases/FetchBackendEnvironmentUseCaseProtocol.swift
```diff
@@ -18,18 +18,17 @@
 
 import Foundation
 
-// TODO: [WPB-16272] Remove duplication
+public protocol FetchBackendEnvironmentUseCaseProtocol: Sendable {
```
johnxnguyen: suggestion: `FetchBackendConfigUseCaseProtocol`.

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackendConfirmation/SwitchBackendConfirmationViewModel.swift
```diff
@@ -26,59 +28,70 @@ public class SwitchBackendConfirmationViewModel {
...
 
-    public enum Event {
-
-        case didCancel
-        case didConfirm
-
+    func confirm() async {
+        isLoading = true
+
+        let fetchDefaultSSOTask = Task.detached { [fetchDefaultSSOSettings] in
```
johnxnguyen: issue: shouldn't we switch backend after confirm? I would also then question if fetching the SSO link, and navigating to SSo or on prem login is the responsibility of `SwitchBackendConfirmationViewModel`. It would seem more appropriate that the `DetermineAuthMethodViewModel` handles this. @samwyndham what do you think?

johnxnguyen (commented):
Leaving some initial comments.

WireAuthentication/Sources/WireAuthentication/Components/DetermineAuthMethodComponent.swift
KaterinaWire: done

WireAuthentication/Sources/WireAuthenticationAPI/Models/BackendEnvironmentInfo.swift
KaterinaWire: done

WireAuthentication/Sources/WireAuthenticationAPI/Models/BackendEnvironmentInfo.swift
KaterinaWire: done

WireAuthentication/Sources/WireAuthenticationAPI/Use cases/FetchBackendEnvironmentUseCaseProtocol.swift
KaterinaWire: done

WireAuthentication/Sources/WireAuthenticationAPI/Models/BackendEnvironmentInfo.swift
KaterinaWire: I think here the issues with `certificateKey`. `SecKey` is not directly decodable using Decodable protocol. If you suggest simply decode the json, `certificateKeyData` can be stored as `Data`. However, I'm not sure who will be responsible for converting it to `SecKey`.

WireAuthentication/Sources/WireAuthenticationAPI/Use cases/FetchDefaultSSOSettingsUseCaseProtocol.swift
KaterinaWire: If default SSO code is nil, the user follow to the on-prem login with email

KaterinaWire: The flow:
- we receive the backend config from the new endpoint -> we check if new environment has a default SSO code:
if yes -> generate the link and navigate to the web page
if no -> login via email

WireAuthentication/Sources/WireAuthentication/Components/LoginViaEmailOnPremComponent.swift
KaterinaWire: yes, please. I will keep it like this

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackendConfirmation/SwitchBackendConfirmationViewModel.swift
samwyndham: @johnxnguyen Does `DetermineAuthMethodViewModel` just call the cloud endpoints or should it also call on prem endpoints? My assumption was that it just calls cloud in which case I think it makes more sense to switch backend here or after here passing the backend info to the following component. But if it also calls backend endpoint then I guess we should do something different 🤔

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackendConfirmation/SwitchBackendConfirmationViewModel.swift
KaterinaWire: Updated

WireAuthentication/Sources/WireAuthenticationAPI/Use cases/FetchBackendEnvironmentUseCaseProtocol.swift
KaterinaWire: Removed unknown error

WireAuthentication/Sources/WireAuthenticationAPI/Use cases/FetchDefaultSSOSettingsUseCaseProtocol.swift
KaterinaWire: renamed to "noInternet"

WireAuthentication/Sources/WireAuthenticationLogic/FetchBackendEnvironmentUseCase.swift
KaterinaWire: Yes, I moved it to body of the `invoke(at configURL: URL`

WireAuthentication/Sources/WireAuthenticationLogic/FetchDefaultSSOSettingsUseCase.swift
KaterinaWire: Updated the code

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackendConfirmation/SwitchBackendConfirmationViewModel.swift
KaterinaWire: @johnxnguyen We will have an issue if we switch the backend and get the SSO code on `DetermineAuthMethodViewModel` and the user presses **Cancel** on `SwitchBackendConfirmationView`. We will need to implement something that detects this and switches the backend back

WireAuthentication/Sources/WireAuthenticationAPI/Models/BackendEnvironmentInfo.swift
johnxnguyen: Yes just store it as data. In `WireAPI`, the `ServerTrustValidator` will deal with the `SecKey`, and for `WireTransport` we can map it on when it leaves `WireAuthentication`.

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackendConfirmation/SwitchBackendConfirmationViewModel.swift
johnxnguyen: ok, makes sense then that `DetermineAuthMethod` only deals with cloud. I suppose it does make sense that `SwitchBackendConfirmation` deals with navigating to email or sso, and in the case that we switch backend via deep link.

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackendConfirmation/SwitchBackendConfirmationViewModel.swift
```diff
@@ -99,8 +98,9 @@ public class SwitchBackendConfirmationViewModel {
                 )
             }
         } catch {
-            // alert = .unknownError
+            WireLogger.authentication.error("Unexpected error while fetching default SSO code: \(error)")
```
samwyndham, line 105: I think we show an alert in the following cases:

- No internet
- Unknown error / general alert.

samwyndham (commented):
Nice work. I left a comment about showing alert which I think should probably be done as part of this PR

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackendConfirmation/SwitchBackendConfirmationViewModel.swift
samwyndham: To establish no internet both use cases will need to throw either a no internet error or something wrapping the underlying error that can be parsed in the view model. 

```swift
case noInternet
// or
case unknownError(Error)
// or
case urlError(URLError)
```

WireAuthentication/Sources/WireAuthenticationAPI/Models/BackendEnvironmentInfo.swift
KaterinaWire: Changed, certificateKey is `Data` now

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackendConfirmation/SwitchBackendConfirmationViewModel.swift
KaterinaWire: I've added one general error, let's see if need more later

samwyndham: OK, I guess we will certainly need to handle the no network error but we can do that after I suppose. Perhaps we should have a ticket to review all the error handling

samwyndham (approved):
Nice work! I think we need to clarify our error handling throughout this module. We can discuss this together with @johnxnguyen next time we have a catchup on some topic

WireAuthentication/Sources/WireAuthentication/Components/DetermineAuthMethodComponent.swift
```diff
@@ -112,4 +121,8 @@ extension DetermineAuthMethodComponent: DetermineAuthMethodView.Factory {
         loginViaSSOComponent.view(ssoURL: ssoURL)
     }
 
+    func switchBackendView(email: String, environment: BackendConfig) -> SwitchBackendConfirmationView {
+        switchBackendConfirmationComponent.view(email: email, environment: environment)
```
johnxnguyen: suggestion: rather than inject the environment here in the view constructor, inject it into the component, this way it becomes part of the dependency graph and can be further propagated if needed.

WireAuthentication/Sources/WireAuthentication/Components/SwitchBackendConfirmationComponent.swift
```diff
@@ -0,0 +1,112 @@
...
+            email: email,
+            fetchDefaultSSOSettings: fetchDefaultSSOSettings(environment: environment),
+            ssoLinkGenerator: ssoLinkGenerator(environment: environment),
+            environment: environment
+        )
+    }
+
+    // MARK: - Private dependencies
+
+    private func authenticationAPI(environment: BackendConfig) -> AuthenticationAPI {
```
johnxnguyen: suggestion: as mentioned in the previous comment, if you have `BackendConfig` as part of the component you could clean up these initializations by also adding other objects to the graph, including backend environment.

WireAuthentication/Sources/WireAuthentication/Components/SwitchBackendConfirmationComponent.swift
```diff
@@ -0,0 +1,112 @@
...
+                        } catch {
+                            WireLogger.authentication.error("Failed to create PinnedKey: \(error)")
+                            return nil
+                        }
+                    } ?? [],
+                    proxySettings: convertProxySettings(from: environment.proxySettings)
+                ),
+                minTLSVersion: dependency.minTLSVersion
+            )
+        ).makeAPI(for: dependency.defaultAPIVersion)
```
johnxnguyen, line 97: issue: `defaultAPIVersion` is removed in my PR and will need to be resolved. Probably it's better to merge your PR first.

WireAuthentication/Sources/WireAuthenticationAPI/Use cases/FetchBackendEnvironmentUseCaseProtocol.swift
```diff
@@ -18,18 +18,16 @@
 
 import Foundation
 
-// TODO: [WPB-16272] Remove duplication
+public protocol FetchBackendConfigUseCaseProtocol: Sendable {
 
-/// Proxy settings for communicating with a backend server.
+    @MainActor
```
johnxnguyen: suggestion: I think it's better to avoid making use cases main actor and instead dispatch to background tasks at the call site, but I think we can do that in another PR.

WireAuthentication/Sources/WireAuthenticationUI/Views/DetermineAuthMethod/DetermineAuthMethodViewModel.swift
```diff
@@ -134,8 +139,16 @@ package final class DetermineAuthMethodViewModel: ObservableObject {
             }
 
         case let .onPremLogin(email, backendConfig):
-            // TODO: [WPB-15944] Handle on-prem login
-            break
+            Task.detached {
+                do {
+                    let environmentInfo = try await self.fetchBackendConfig.invoke(at: backendConfig)
```
johnxnguyen: thought: even though you dispatch on a detached task, because the use case is marked `@MainActor` it will switch back to the main thread.

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackendConfirmation/SwitchBackendConfirmationViewModel.swift
johnxnguyen: I think we can take advantage of `LocalizedError` and throw them higher up the invocation chain, possibly even from the network service..

WireAuthentication/Tests/WireAuthenticationUITests/Resources/ReferenceImages/LoginViaEmailOnPremViewTests/testDynamicTypeVariantsWithProxySettings.large.png
johnxnguyen, line 1: question: Shouldn't there be proxy settings here?

johnxnguyen (approved):
There will be several conflicts with my PR so I suggest you merge first and I deal with the conflicts. 

All my comments are non blocking, they're things we can address separately.

The only thing that seems wrong are the snapshots with proxy credentials in large font scales.

Otherwise, looks good!

WireAuthentication/Sources/WireAuthentication/Components/DetermineAuthMethodComponent.swift
KaterinaWire: done

WireAuthentication/Sources/WireAuthentication/Components/SwitchBackendConfirmationComponent.swift
KaterinaWire: done

WireAuthentication/Sources/WireAuthenticationAPI/Use cases/FetchBackendEnvironmentUseCaseProtocol.swift
KaterinaWire: removed

WireAuthentication/Sources/WireAuthenticationUI/Views/DetermineAuthMethod/DetermineAuthMethodViewModel.swift
KaterinaWire: changed

WireAuthentication/Tests/WireAuthenticationUITests/Resources/ReferenceImages/LoginViaEmailOnPremViewTests/testDynamicTypeVariantsWithProxySettings.large.png
KaterinaWire: yes, fixed

PR #2632, opened 2025-03-04
feat: wire channels group icon - WPB-16356

El-Fitz:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16356" title="WPB-16356" target="_blank">TaskWPB-16356</a>  [iOS] Create new groups avatars
  </td></table>
  <br />
 

### Issue

Implements the new conversation groups icons.
Integration into the conversations list will be done in a later PR.
Also updates the module's organisation.

Merging should wait until @caldrian's PR is merged and I've made the ensuing necessary changes to this PR.

### Testing

See tests screens.

---

### Checklist


WireConversations/Sources/WireConversationsResources/Media.xcassets/channel-icons/channel-icon-gray.imageset/channel-avater-gray-dark.svg
netbe, line 1: **question:** why is the channel avatar here? isn't it just for groups
**nitpic:** typo in asset name

WireConversations/Sources/WireConversationsUI/WireConversationGroupIcon.swift
```diff
@@ -0,0 +1,38 @@
...
+package import WireConversationsResources
+
+package struct WireConversationGroupIcon: View {
+    let groupIcon: WireConversationGroupIconAsset
+
+    package init(
+        conversationID: String,
+        conversationIDMapper: any ConversationIDToPaletteMapper<WireConversationGroupIconAsset>
+    ) {
+        self.groupIcon = ._1
```
netbe: issue: I don't see how we use the conversationIDMapper here?

WireConversations/Sources/WireConversationsUIBindings/WireConversationGroupIconFactory.swift
```diff
@@ -16,4 +16,21 @@
...
+    private let mapper = ConversationIDToIconMapper()
+
+    public init() { }
+
+    @MainActor
+    public func create(conversationID: String) -> some View {
+        WireConversationGroupIcon(
+            conversationID: conversationID,
+            conversationIDMapper: mapper
+        )
```
netbe: suggestion:

```suggestion
        WireConversationGroupIcon(
            asset: mapper.palette(for: conversationID)
        )
```

netbe (commented):

WireConversations/Sources/WireConversationsUIBindings/WireConversationGroupIconFactory.swift
El-Fitz: That works too!

WireConversations/Sources/WireConversationsUI/WireConversationGroupIcon.swift
El-Fitz: Indeed. I had set this up before implementing the mapper and forgot to come back to it. Thanks!

WireConversations/Sources/WireConversationsResources/Media.xcassets/channel-icons/channel-icon-gray.imageset/channel-avater-gray-dark.svg
El-Fitz: ~~Typo?~~ Oh, avatar instead of icon

El-Fitz: I was on an asset adding spree and when I remembered it was just about groups I had already added the channel avatars. And redoing the image set is a hassle.

El-Fitz: removed

netbe (approved):
looks good let's 🚢

crowdin.yml
```diff
@@ -59,14 +59,14 @@ files: [
     "translation" : "/WireAuthentication/Sources/WireAuthenticationUI/Resources/Localization/%osx_code%/%original_file_name%",
   },
   {
-    "source" : "/WireConversations/Sources/WireConversationsUI/Resources/Localization/en.lproj/Accessibility.strings",
-    "dest" : "/App/WireConversationsUI/Resources/Accessibility.strings",
-    "translation" : "/WireConversations/Sources/WireConversationsUI/Resources/Localization/%osx_code%/%original_file_name%",
+    "source" : "/WireConversations/Sources/WireConversationsResources/Localization/en.lproj/Accessibility.strings",
+    "dest" : "/App/WireConversationsResources/Accessibility.strings",
```
caldrian, line 63: suggestion: I wonder if we should keep the current name to match the other translation sources.
This argument is used to build a virtual file tree:
Screenshot 2025-03-06 at 10 19 20

After we have strings in there I think we can't change it anymore without losing translations.

What about leaving the `dest` argument unchanged (keeping `/App/WireConversationsUI/Resources/{Accessibility,Localizable}.strings`) @netbe ?
Will there be more strings within the package `WireConversations`?

caldrian (approved):

WireConversations/Sources/WireConversationsResources/WireConversationGroupIcon.swift
```diff
@@ -0,0 +1,85 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+package import SwiftUI
+
+package enum WireConversationGroupIconAsset {
```
caldrian, line 21: question: Can't we rely on the enum generated by Xcode?

caldrian: This line works within the target as well.
I see that the generated resources have the `internal` access modifier, which brings me to question: why do we want to have a separate target just for assets?
Do they really have to be shared across more than one other target?
```
let image = UIImage(resource: .groupIcon1)
```

crowdin.yml
netbe: I didn't realise this was under a special target, not sure why @El-Fitz is best to answer.

As to keep the dest unchanged, I don't mind as long as it's consistent with other targets we have

El-Fitz: For the `WireConversationGroupIcon`, and later `WireConversationChannelIcon`, types, have no reason to be exposed publicly and thus don't belong in API, but  are both tied to the assets themselves and used across multiple targets.

At the same time, they need to be in the same target as the assets because of how `Bundle.module` works. 

A resources / assets target solves both these issues. And if we are to have a dedicated assets target, we might as well put the strings in there too.

WireConversations/Sources/WireConversationsResources/WireConversationGroupIcon.swift
El-Fitz: > question: Can't we rely on the enum generated by Xcode?

No, the order of items matters (edit: we want icons to be consistent across platforms, and the algorithm picking the icon for a given conversation relies on the icons' order).

> [...] why do we want to have a separate target just for assets? Do they really have to be shared across more than one other target?

The two types that depend on them do (edit: they are shared across `API`, `Implementation`, and `UI`).

samwyndham:
Removing myself as PR is already approved

PR #2644, opened 2025-03-06
feat: wire channels channels icons - WPB-16355

El-Fitz:
### Issue

Implements channels icons

### Testing

See ui tests snapshots

---

### Checklist


WireConversations/Sources/WireConversationsImplementation/Array+WireConversationChannelIconAsset.swift
```diff
@@ -0,0 +1,25 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireConversationsResources
+
+extension [WireConversationChannelIconAsset] {
+    static var all: [WireConversationChannelIconAsset] {
+        WireConversationChannelIconAsset.all
+    }
+}
```
netbe: question: why doing an extension on array?

WireConversations/Sources/WireConversationsResources/WireConversationChannelIcon.swift
```diff
@@ -0,0 +1,61 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+package import SwiftUI
+
+package enum WireConversationChannelIconAsset {
```
netbe: question: is there a reason not to use rawValue as UInt and CaseIterable?

netbe (commented):
added some questions

WireConversations/Sources/WireConversationsResources/WireConversationChannelIcon.swift
El-Fitz: We need to have the same ordering across all platforms.
Using `CaseIterable` and a `UInt` raw value, adding a new case at the wrong position would mess up the order and introduce incoherencies with a other platforms.

Additionally, the index isn't the "raw value" for the enum (that would be the colour's name). Using it that way would work, but mixing technical convenience and business logic would lead to more implicit and possibly unclear API.

WireConversations/Sources/WireConversationsImplementation/Array+WireConversationChannelIconAsset.swift
El-Fitz: It's just a convenience so we can simply use `.all` in `ConversationIDToChannelIconMapper`, ie `let palette: [WireConversationChannelIconAsset] = .all`, which I find clearer than `let palette: [WireConversationChannelIconAsset] = WireConversationChannelIconAsset.all`, and definitely more readable than `let palette: [WireConversationChannelIconAsset] = [...moderately large array]`.

Which is why the extension is in `WireConversationsImplementation`, and scoped to `internal`.

netbe: it reads nice, but add more extra code compared to the benefit.

```
let palette: [WireConversationChannelIconAsset] = WireConversationChannelIconAsset.all
```
or
`let palette = WireConversationChannelIconAsset.all`

samwyndham (approved):
LGTM. I also wondered about making `WireConversationChannelIconAsset` `CaseIterable` but I see it could be brittle.

WireConversations/Sources/WireConversationsImplementation/Array+WireConversationChannelIconAsset.swift
El-Fitz: Sure

netbe (approved):

PR #2647, opened 2025-03-07
fix: ensure answer/start AVS call is triggered after joining subgroup and before setting the epoch - WPB-16001

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16001" title="WPB-16001" target="_blank">BugWPB-16001</a>  [iOS] No audio when joining call on canary team
  </td></table>
  <br />
 

…setting epoch

### Issue

When answering a call for a MLS group, currently we follow these steps (as mentionned [here](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/616792280/Use+case+answer+a+call+incoming)):
1. wcall_answer
2. join sub-conv
3. wcall_set_epoch_info

Logic should from now on be:
1. join sub-conv
2. wcall_answer
3. wcall_set_epoch_info

Same logic was applied to starting a call (`wcall_start`) 

Reason why we should call wcall_answer after having joined the sub-conv is that SFT requests authorization of clients, and unfortunately it seems that the join sub-conv can take several seconds some times, so when the SFT asks for authorization, we don't know of any authorized clients (since set_epoch_info was not called yet) and then it takes until the next timeout fo authorize.

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
+++ b/wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
@@ -756,7 +756,7 @@ public extension WireCallCenterV3 {
                         parentQualifiedID: parentQualifiedID,
                         parentID: parentGroupID
                     )
-                    
+
                     try completion()
 
                     // Generate and set the conference information for the subgroup
--- a/wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
+++ b/wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
@@ -756,6 +756,8 @@ public extension WireCallCenterV3 {
                         parentQualifiedID: parentQualifiedID,
                         parentID: parentGroupID
                     )
+                    
+                    try completion()
 
                     // Generate and set the conference information for the subgroup
                     let initialConferenceInfo = try await mlsService.generateConferenceInfo(
@@ -767,8 +769,6 @@ public extension WireCallCenterV3 {
                         conversationId: conversationID,
                         info: initialConferenceInfo
                     )
-
-                    try completion()
 
                     // Set up a task to observe changes in the conference information
                     // and update AVS accordingly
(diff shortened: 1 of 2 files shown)


jullianm:
@netbe keeping it as a draft for now because changing the order of the steps creates a bug when answering a call on MLS group (no audio/video). calling `wcall_answer` right before `wcall_set_epoch_info` works fine but calling it right after creates the bug. we're discussing with Dusan to see it could be AVS issue. works fine when starting a call though.

netbe (approved):
Looks good.
2 questions:
1. Are other platforms going to make that change?
2. Do you need to update the current tests?

wire-ios-sync-engine/Tests/Source/Calling/WireCallCenterV3Tests.swift
```diff
@@ -963,9 +963,6 @@ final class WireCallCenterV3Tests: MessagingTest {
     }
 
     func testThatItAnswersACall_conference_mls() throws {
-        // TODO: [WPB-7346]: enable this (flaky) test again
-        throw XCTSkip()
-
```
jullianm, line 968: @netbe these UTs were previously skipped. This PR reintroduced them and ensure that `wcall_answer` and `wcall_start` are triggered after joining the subgroup.

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
```diff
@@ -707,7 +723,10 @@ public extension WireCallCenterV3 {
     /// See documentation:
     /// https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/692027483/Use+case+Join+conference+sub-conversation+MLS
 
-    private func setUpMLSConference(in conversation: ZMConversation) throws {
+    private func setUpMLSConference(
+        in conversation: ZMConversation,
+        completion: @escaping () throws -> Void
```
caldrian: A throwing completion closure looks wrong.
Looking at the changes below, this is not really a completion close, right?

jullianm: yes you're right, I'll rename it `avsCallHandler`

jullianm: fixed in https://github.com/wireapp/wire-ios/pull/2647/commits/eca6be6fbca394e10a09b764343ef729cec2785d

caldrian (approved):

PR #2651, opened 2025-03-10
fix: make writing databaseKey on context async, fix UTs - WPB-16374

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16374" title="WPB-16374" target="_blank">BugWPB-16374</a>  [iOS] cannot switch EAR on current internal builds
  </td></table>
  <br />
 

### Issue

**Context**: when enabling/disabling EAR, we're stuck on loading screen
**Reason**: when doing the database migration, we're writing the databaseKey from the context and waiting for the operation to complete, this never completes.
**Solution**: make the writing operation async

### Testing

Enable/disable EAR on the settings screen. App should not be stuck anymore.

### Checklist

### DIFF
--- a/wire-ios-data-model/Source/Authentication/EAR/EARService.swift
+++ b/wire-ios-data-model/Source/Authentication/EAR/EARService.swift
@@ -558,9 +558,9 @@ public class EARService: EARServiceInterface {
         }
     }
 
-    private func performInAllContexts(_ block: (NSManagedObjectContext) -> Void) {
+    private func performInAllContexts(_ block: @escaping (NSManagedObjectContext) -> Void) {
         for context in databaseContexts {
-            context.performAndWait {
+            context.perform {
                 block(context)
             }
         }
--- a/wire-ios-data-model/Tests/Authentication/EAR/EARServiceTests.swift
+++ b/wire-ios-data-model/Tests/Authentication/EAR/EARServiceTests.swift
@@ -243,6 +243,8 @@ final class EARServiceTests: ZMBaseManagedObjectTest, EARServiceDelegate {
         // Then EAR is enabled on the context
         XCTAssertTrue(uiMOC.encryptMessagesAtRest)
 
+        XCTAssertTrue(waitForAllGroupsToBeEmpty(withTimeout: 0.5))
+
         // Then all contexts have the database key.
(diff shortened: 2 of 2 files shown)


wire-ios-data-model/Source/Authentication/EAR/EARService.swift
```diff
@@ -558,9 +558,9 @@ public class EARService: EARServiceInterface {
         }
     }
 
-    private func performInAllContexts(_ block: (NSManagedObjectContext) -> Void) {
+    private func performInAllContexts(_ block: @escaping (NSManagedObjectContext) -> Void) {
         for context in databaseContexts {
-            context.performAndWait {
+            context.perform {
```
netbe, line 563: **thought:** I don't have everything in mind, but I think there was a good reason to have the performAndWait in the first place.

netbe (commented):
> Reason: when doing the database migration, we're writing the databaseKey from the context and waiting for the operation to complete, this never completes.

Why does it never complete?

wire-ios-data-model/Source/Authentication/EAR/EARService.swift
netbe: @johnxnguyen should double check this

johnxnguyen: To make it synchronous, i.e when the method returns all work is done.

johnxnguyen: I don't quite understand why this fixes the issue though? What change introduced the issue?

johnxnguyen (commented):
@jullianm is this PR still relevant?

wire-ios-data-model/Source/Authentication/EAR/EARService.swift
jullianm: @johnxnguyen  yes still relevant. so, the logout fix introduced that issue where the performAndWait gets stuck forever so UI wise we get an infinite loading screen when switching EAR. I need to look more into that because the fix is not ideal and I need to understand why it gets stuck. I'll look into it soon, let's keep it open

PR #2657, opened 2025-03-10
fix: duplicate message - WPB-11435

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11435" title="WPB-11435" target="_blank">BugWPB-11435</a>  [iOS] All messages in a 1:1 conversation duplicated in iOS
  </td></table>
  <br />
 

### Issue

On the legacy code, it happens message get processed multiple times.
It could be that in the following **recursive** method, because of a suspension point, the `fetchNextEventsBatch` fetch the same event before we actually delete it from the database and so reprocess it.
```
  private func process(
        with privateKeys: EARPrivateKeys?,
        _ consumeBlock: ConsumeBlock,
        firstCall: Bool,
        callEventsOnly: Bool
```

* Make EventDecoder an actor so it's always executed on the same queue.
* Safe guard the method with `isProcessing` so we don't process twice the same event.

### Testing

1. on a conversation, run web spam messages
2. on the receiver side, send messages

Actual:
You'll see duplicate messages

Fix:
You'll get no duplicate messages

---

### Checklist

---


johnxnguyen (approved):

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder.swift
```diff
@@ -365,14 +366,21 @@ extension EventDecoder {
         _ consumeBlock: ConsumeBlock,
         firstCall: Bool,
         callEventsOnly: Bool
-    ) async {
+    ) async {        
+        guard !isProcessing else {
+            WireLogger.updateEvent.debug("⚠️ process called before end of processing, abort")
+            return
```
jullianm, line 372: question(non-blocking): so if a second events batch wants to be processed while the first one is not finished yet, we discard it (early return). And I guess it will be processed again once the first is finished otherwise we'll lose that events batch, right ?

jullianm (approved):
LGTM, just one non blocking comment about events batch processing logic

wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoder.swift
jullianm: Or maybe this logic only concerns the duplicated events since the function is recursive. Just checking that we don't lose any events along the way.

netbe: to be honest I am not sure, this might be not completely correct

netbe:
closing as new sync should solve the error and this has not been tested properly

PR #2664, opened 2025-03-11
feat: Deep Link Handling - WPB-15947

KaterinaWire:
### Issue

This PR implements deep link handling:
- passing backend configuration URL via bridge to authentication module, 
- [UI](https://www.figma.com/design/sOAbcd6UVtlZ5ejK4BoEvX/iOS?node-id=3880-14022&t=2uPJTMfdp15aE3nG-0)

### Checklist

---


WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackend/SwitchBackendHandler.swift
```diff
@@ -0,0 +1,48 @@
...
+
+    @MainActor
+    package func invoke(_ backendConfigURL: URL) async {
+        do {
+            let useCase = factory.fetchBackendConfigUseCase()
+            let backendConfig = try await Task.detached {
+                try await useCase.invoke(at: backendConfigURL)
+            }.value
+            router.presentSheet(RootView.ModalDestination.switchBackend(backendConfig: backendConfig))
+        } catch {
```
samwyndham: **issue:** Do we need to inform the user that this failed?

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackend/SwitchBackendConfirmationViewModel.swift
```diff
@@ -104,7 +104,7 @@ package class SwitchBackendConfirmationViewModel: ObservableObject {
         if let proxySettings = backendConfig.proxySettings, proxySettings.needsAuthentication {
             router.presentSheet(
                 RootView.ModalDestination.onPremiseLogin(
-                    email: email,
+                    email: email ?? "",
```
samwyndham: **issue:** If email is nil here then the user can't progress right because the login requires that the email has already been input - e.g. in the view models it is a let and in the designs it is grayed out. I think there is some issue we need to resolve here.

WireAuthentication/Sources/WireAuthenticationUI/Components/OnPremHeader/OnPremHeaderViewModel.swift
```diff
@@ -0,0 +1,45 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import UIKit
+import WireAuthenticationAPI
+import WireReusableUIComponents
+
+@MainActor
+package final class OnPremHeaderViewModel: ObservableObject {
```
samwyndham: **suggestion:** In one of the architecture forums we discussed when views need a view model and the main argument was that in general reusable components shouldn't need one. In this case especially there is no logic in the view model and no updates to observe. I think you should delete this view model and do everything in the view.

WireAuthentication/Sources/WireAuthenticationUI/Components/OnPremHeader/OnPremHeaderView.swift
```diff
@@ -0,0 +1,54 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import SwiftUI
+import WireDesign
+
+struct OnPremHeaderView: View {
```
samwyndham: **praise:** Nice to extract this reusable component.

samwyndham (changes requested):
Nice work! I think there are a couple of issues to fix.

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackend/SwitchBackendConfirmationViewModel.swift
samwyndham: Perhaps this is treated as an error and we show an alert?

WireAuthentication/Sources/WireAuthenticationUI/Components/OnPremHeader/OnPremHeaderView.swift
```diff
@@ -0,0 +1,54 @@
...
+    @StateObject var viewModel: OnPremHeaderViewModel
```
johnxnguyen: suggestion: for me `OnPremHeaderView` is a reusable component, which means it shouldn't have a view model, but just the data it needs (backend info).

WireAuthentication/Sources/WireAuthenticationUI/Views/DetermineAuthMethod/DetermineAuthMethodViewModel.swift
```diff
@@ -55,11 +57,15 @@ package final class DetermineAuthMethodViewModel: ObservableObject {
     package init(
         router: any Router,
         factory: any Factory,
+        backendConfig: BackendConfig?,
```
johnxnguyen: suggestion: make this non-optional (we use the default config in case it's not on-prem), then add a bool flag to determine if it's on-prem or not so we know whether to show the header view.

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackend/SwitchBackendConfirmationViewModel.swift
johnxnguyen: question: could we also make this optional and handle the optionality in the view model?

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackend/SwitchBackendHandler.swift
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @MainActor
+    package func invoke(_ backendConfigURL: URL) async {
+        do {
+            let useCase = factory.fetchBackendConfigUseCase()
+            let backendConfig = try await Task.detached {
+                try await useCase.invoke(at: backendConfigURL)
+            }.value
+            router.presentSheet(RootView.ModalDestination.switchBackend(backendConfig: backendConfig))
```
johnxnguyen: question: what do you think about replacing the root rather than switching sheets?

WireAuthentication/Sources/WireAuthenticationUI/Components/OnPremHeader/OnPremHeaderViewModel.swift
KaterinaWire: Yes, sure, I missed that discussion. I removed  `OnPremHeaderViewModel`

WireAuthentication/Sources/WireAuthenticationUI/Components/OnPremHeader/OnPremHeaderView.swift
KaterinaWire: Yep, I removed  `OnPremHeaderViewModel`

WireAuthentication/Sources/WireAuthenticationUI/Views/DetermineAuthMethod/DetermineAuthMethodViewModel.swift
KaterinaWire: Fixed, non optional now

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackend/SwitchBackendConfirmationViewModel.swift
KaterinaWire: No, it's not an error. We are navigating to the login screen (LoginViaEmailOnPremView or DetermineAuthMethodView) from the deep link, so we don't have an email yet. The user should be switched to the new backend and be able to enter an email address and password.

KaterinaWire: https://www.figma.com/design/sOAbcd6UVtlZ5ejK4BoEvX/iOS?node-id=3880-14022&t=VvyqVS7poP5lqkYH-0

samwyndham: @johnxnguyen I think this is an invalid state as the view isn't designed to handle email input. I propose we show an error alert if email is nil here

KaterinaWire: In the `LoginViaEmailOnPremView` we have .`disabled(!viewModel.email.isEmpty)` for the `emailField`

KaterinaWire: @samwyndham  but if the user came from a deep link, the user has not entered an email yet. I think this is a valid case, no?

KaterinaWire: @johnxnguyen @samwyndham We are talking about the use case:
- user opens a deep link
- proxy is enabled
as far as I understand, we should always allow login only via email

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackend/SwitchBackendHandler.swift
KaterinaWire: I added general error

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackend/SwitchBackendHandler.swift
KaterinaWire: Switch backend view is a modal, not a part of the navigation stack:
Screenshot 2025-03-13 at 18 00 26

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackend/SwitchBackendConfirmationViewModel.swift
johnxnguyen: I think think if the email is present we treat it as a constant prefilled value (email field is disabled but filled), and if it's nil then the email field is enabled and user can provide their email.

WireAuthentication/Sources/WireAuthenticationAPI/WireAuthenticationBridge.swift
```diff
@@ -63,6 +63,7 @@ public final class WireAuthenticationBridge {
 
         case ssoAuthenticationSuccess(userID: UUID, cookies: [HTTPCookie])
         case ssoAutheticationFailure
+        case onSwitchBackend(configURL: URL)
```
johnxnguyen: Suggestion: backendSwitchRequested

WireAuthentication/Sources/WireAuthenticationUI/Views/DetermineAuthMethod/DetermineAuthMethodViewModel.swift
```diff
@@ -55,20 +57,52 @@ package final class DetermineAuthMethodViewModel: ObservableObject {
...
+
+                    await self.handleAuthenticationMethod(
+                        .onPremLogin(email: nil, backendConfig: configURL),
+                        backendMetadata: resolvedBackendMetadata
+                    )
+                }
+            default:
+                break
+            }
+        }
```
dmitrysimkin: Suggestion: too deep nested, extract to method and maybe guard?

dmitrysimkin (commented):
Looks nice, just minor code style comment

WireAuthentication/Sources/WireAuthentication/Components/LoginViaEmailComponent.swift
```diff
@@ -32,24 +32,31 @@ protocol LoginViaEmailComponentDependency: Dependency {
...
+    var minTLSVersion: TLSVersion { get }
     @MainActor var bridge: WireAuthenticationBridge { get }
 
 }
 
 class LoginViaEmailComponent: Component<LoginViaEmailComponentDependency> {
 
-    public let backendMetadata: WireAuthenticationAPI.BackendMetadata
+    private let environmentType: BackendEnvironmentType
+    private let backendConfig: BackendConfig
```
johnxnguyen, line 43: question: do these need to be injected? Can't they be obtained from the parent?

WireAuthentication/Sources/WireAuthentication/Components/LoginViaEmailComponent.swift
```diff
@@ -58,6 +65,15 @@ class LoginViaEmailComponent: Component<LoginViaEmailComponentDependency> {
...
     }
 
+    private var networkService: NetworkService {
+        shared {
+            NetworkService.make(
+                backendEnvironment: BackendEnvironment(backendConfig),
+                minTLSVersion: dependency.minTLSVersion
+            )
+        }
+    }
```
johnxnguyen, line 75: question: why can't we receive this from the parent?

WireAuthentication/Sources/WireAuthenticationUI/Utilities/Alert.swift
```diff
@@ -20,9 +20,9 @@ import Foundation
 
 /// Identifies an alert and provides it's title and message.
 
-struct Alert: Hashable, Identifiable, Sendable {
+public struct Alert: Hashable, Identifiable, Sendable {
```
johnxnguyen: question: why does this need to be public?

WireAuthentication/Sources/WireAuthenticationUI/Views/DetermineAuthMethod/DetermineAuthMethodViewModel.swift
```diff
@@ -55,20 +57,52 @@ package final class DetermineAuthMethodViewModel: ObservableObject {
...
+        self.backendMetadata = backendMetadata
         self.backendConfig = backendConfig
         self.emailOrSSOCode = emailOrSSOCode
         self.isLoading = isLoading
+
+        self.cancellable = bridge.inboundEvents.sink { event in
+            switch event {
+            case let .onSwitchBackend(configURL):
+                Task {
+                    let resolvedBackendMetadata: BackendMetadata
```
johnxnguyen: suggestion: I understand that we resolve the metadata here so we can invoke `handleAuthenticationMethod`, but we pass `.onPremLogin` which doesn't use the backendMetadata. We could avoid this if we factor out the handling of `.onPremLogin` into a private method and just invoke that directly. What do you think?

WireAuthentication/Sources/WireAuthentication/Components/DetermineAuthMethodComponent.swift
```diff
@@ -38,36 +37,62 @@ protocol DetermineAuthMethodComponentDependency: Dependency {
...
-    @MainActor private var viewModel: DetermineAuthMethodViewModel {
-        DetermineAuthMethodViewModel(
-            router: dependency.router,
-            factory: self,
-            environmentType: dependency.environmentType,
-            backendConfig: dependency.backendConfig
-        )
+    private let environmentType: BackendEnvironmentType
+    private let backendConfig: BackendConfig
+    private let backendMetadata: BackendMetadata?
```
johnxnguyen: question: I understand these are injected now to accommodate the possibility now that we have received a different backend config. I wonder though, if it makes sense to have a different component `DetermineAuthMethodOnPremComponent` so these changes could be isolated to the new component only. Having said that though, we could keep using the same view and view model. What do you think?

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackend/SwitchBackendConfirmationViewModel.swift
```diff
@@ -149,14 +149,24 @@ package class SwitchBackendConfirmationViewModel: ObservableObject {
                     modalDestination = .ssoLogin(url: ssoURL, backendEnvironment: backendEnvironment)
                     WireLogger.authentication.info("Fetching default SSO URL succeed")
                 } else {
-                    router.presentSheet(
-                        RootView.ModalDestination.onPremiseLogin(
-                            email: email,
-                            environmentType: environmentType,
-                            environment: backendConfig,
-                            backendMetadata: backendMetadata
+                    if let email {
```
johnxnguyen: suggestion: it may be nice to include this in enclosing `if`:

```swift
else if let email {
```

WireAuthentication/Sources/WireAuthentication/Components/LoginViaEmailComponent.swift
KaterinaWire: because `loginViaEmailComponent` is a child for DetermineAuthMethodComponent, but backendConfig and environmentType runtime parameters. 
I pass them `SwitchBackendConfirmationComponent` -> `DetermineAuthMethodComponent` -> `LoginViaEmailComponent`

WireAuthentication/Sources/WireAuthenticationAPI/WireAuthenticationBridge.swift
KaterinaWire: done

WireAuthentication/Sources/WireAuthenticationUI/Views/DetermineAuthMethod/DetermineAuthMethodViewModel.swift
KaterinaWire: improved

WireAuthentication/Sources/WireAuthenticationUI/Utilities/Alert.swift
KaterinaWire: No, it should be **package**. Fixed

WireAuthentication/Sources/WireAuthenticationUI/Views/DetermineAuthMethod/DetermineAuthMethodViewModel.swift
KaterinaWire: Yes, it makes sense not to introduce additional requests.
Improved

WireAuthentication/Sources/WireAuthentication/Components/DetermineAuthMethodComponent.swift
KaterinaWire: I've added `DetermineAuthMethodOnPremComponent`

WireAuthentication/Sources/WireAuthentication/Components/LoginViaEmailComponent.swift
johnxnguyen: The config and environment type are obtained at runtime in the switch backend confirmation component, so I would expect that they are then injected into the determine auth method component, where they can then be passed to children via needle. Or do I still misunderstand something?

WireAuthentication/Sources/WireAuthentication/Components/RootComponent.swift
```diff
@@ -75,7 +76,12 @@ class RootComponent: BootstrapComponent {
     }
 
     @MainActor private var viewModel: RootViewModel {
-        shared { RootViewModel() }
+        shared { RootViewModel(
+            modalDestination: .authFlow(
+                environmentType: environmentType,
+                backendConfig: backendConfig
+            )
```
johnxnguyen: question: do we still need to inject type and config for determine auth method on cloud now that we have a separate component for determine auth method on prem?

johnxnguyen (approved):

WireAuthentication/Sources/WireAuthenticationUI/Views/DetermineAuthMethod/DetermineAuthMethodViewModel.swift
```diff
@@ -55,20 +57,38 @@ package final class DetermineAuthMethodViewModel: ObservableObject {
...
+        self.backendMetadata = backendMetadata
         self.backendConfig = backendConfig
         self.emailOrSSOCode = emailOrSSOCode
         self.isLoading = isLoading
+
+        self.cancellable = bridge.inboundEvents.sink { event in
+            switch event {
+            case let .backendSwitchRequested(configURL):
+                Task {
+                    await self.handleOnPremLogin(email: nil, backendConfigURL: configURL)
```
dmitrysimkin: question: capturing self? I believe better to weakly self

dmitrysimkin (approved):

WireAuthentication/Sources/WireAuthenticationUI/Views/DetermineAuthMethod/DetermineAuthMethodViewModel.swift
KaterinaWire: fixed

WireAuthentication/Sources/WireAuthenticationUI/Views/SwitchBackend/SwitchBackendConfirmationViewModel.swift
KaterinaWire: fixed

WireAuthentication/Sources/WireAuthentication/Components/RootComponent.swift
KaterinaWire: No, we don't need. I still used old component.

WireAuthentication/Sources/WireAuthentication/Components/LoginViaEmailComponent.swift
KaterinaWire: we will improve it later

PR #2667, opened 2025-03-12
refactor: integrate wire authentication - WPB-16044

johnxnguyen:
### Issue
This PR enables the continuation of the authentication flow into the existing app so the user ends up on the conversation list. Changes include:
- Passing backend information into `WireAuthentication` when loading it.
- Passing all necessary information to complete the flow through `AuthenticationResult`.
- Add the necessary handlers and steps in `AuthenticationCoordinator` to finish the flow.

### Testing
- Enable `useWireAuthentication` feature flag and log in
- Assert that you arrive at the conversation list.

---

### Checklist

---


WireAuthentication/Sources/WireAuthentication/WireAuthenticationAssembly.swift
```diff
@@ -25,6 +25,13 @@ import WireReusableUIComponents
 internal import WireAuthenticationUI
 internal import WireAuthenticationLogic
 
+public typealias BackendEnvironmentType = WireAuthenticationAPI.BackendEnvironmentType
+public typealias BackendConfig = WireAuthenticationAPI.BackendConfig
+public typealias Endpoints = WireAuthenticationAPI.Endpoints
+public typealias ProxySettings = WireAuthenticationAPI.ProxySettings
+public typealias TrustData = WireAuthenticationAPI.TrustData
+public typealias BackendMetadata = WireAuthenticationAPI.BackendMetadata
```
samwyndham, line 33: **question:** Why was this necessary. Is it because of duplicate implementations in WireAuthenticationAPI & WireAPI?

WireAuthentication/Sources/WireAuthenticationUI/Views/Login/LoginViaEmail/LoginViaEmailViewModel.swift
```diff
@@ -86,13 +89,26 @@ package final class LoginViaEmailViewModel: ObservableObject {
...
         do {
             let (cookies, token) = try await loginTask.value
+
+            let emailCredentials = EmailCredentials(
+                email: email,
+                password: trimmedPassword,
+                verificationCode: nil
+            )
+
+            let authenticationResult = AuthenticationResult(
```
samwyndham, line 99: **praise:** Nice pass this around rather than it's components.

WireAuthentication/Sources/WireAuthenticationAPI/WireAuthenticationBridge.swift
```diff
@@ -22,23 +22,19 @@ import Foundation
...
 
-public struct WireAuthenticationBridge {
+public final class WireAuthenticationBridge {
 
     public let onFlowCompletion: (AuthenticationResult) -> Void
     private let onRegisterAccount: () -> Void
-    private let onSSOSuccess: (UUID, [HTTPCookie]) -> Void
-    private let onSSOFailure: () -> Void
+    package var onSSOSuccess: ((UUID, [HTTPCookie]) -> Void)?
+    package var onSSOFailure: (() -> Void)?
```
samwyndham, line 30: **praise:** Nice that this is kept within the package now

samwyndham (approved):
Great work!

WireAuthentication/Sources/WireAuthenticationUI/Resources/Localization/en.lproj/Localizable.strings
```diff
@@ -52,6 +52,7 @@
 "switch_backend_confirmation.teams_url" = "Teams URL:";
 "switch_backend_confirmation.accounts_url" = "Accounts URL:";
 "switch_backend_confirmation.website_url" = "Website URL:";
+"switch_backend_confirmation.countly_url" = "Countly URL:";
```
KaterinaWire, line 55: **question:** why do you think we should display county URL? 
1. Android doesn't display it.
2. On premise backend doesn't have county URL.

WireAuthentication/Sources/WireAuthentication/WireAuthenticationAssembly.swift
johnxnguyen: it was to avoid having to import `WireAuthenticationAPI` to access the types. I think `WireAuthentication` should be the only target we import when we need access to module.

WireAuthentication/Sources/WireAuthenticationUI/Resources/Localization/en.lproj/Localizable.strings
johnxnguyen: Is there a guarantee that on-prem backends don't have a Countly URL? Currently in practice it may be true but in it's not reflected in our type system. 

If it is present, then I think it should be shown for full transparency otherwise it's a risk that someone tampers with the config and it's not known to the end user.

Similarly, if the user sees it and it shouldn't be there (as is the case), then that informs them that something is wrong.

WireAuthentication/Sources/WireAuthentication/WireAuthenticationAssembly.swift
samwyndham: Oh I see. And this works? Good solution then

WireAuthentication/Sources/WireAuthenticationUI/Resources/Localization/en.lproj/Localizable.strings
KaterinaWire: I feel it should be a product decision because we are building a different implementation from Android. And Astrid will ask us what is the source (design or ticket) of this string (for translation context)

WireAuthentication/Sources/WireAuthentication/Components/RootComponent.swift
```diff
@@ -101,25 +102,42 @@ class RootComponent: BootstrapComponent {
...
-    var loginViaSSOComponent: LoginViaSSOComponent {
-        LoginViaSSOComponent()
+    func loginViaSSOComponent(
+        ssoURL: URL,
+        backendEnvironment: WireAuthenticationBackendEnvironment
+    ) -> LoginViaSSOComponent {
+        LoginViaSSOComponent(
+            parent: self,
+            ssoURL: ssoURL,
+            backendEnvironment: backendEnvironment
```
KaterinaWire, line 139: **suggestion:**  maybe it's not a part of this PR, but I think we don't need `backendEnvironment`.  `LoginViaSSOView` only displays the SafariBrowser with the URL.

WireAuthentication/Sources/WireAuthentication/WireAuthenticationAssembly.swift
johnxnguyen: yep!

WireAuthentication/Sources/WireAuthenticationUI/Resources/Localization/en.lproj/Localizable.strings
johnxnguyen: Good point, I'll raise it in the team

WireAuthentication/Sources/WireAuthentication/Components/RootComponent.swift
KaterinaWire: I got the changes, we setup `onSSOSuccess` and `onSSOFailure` handlers in the LoginViaSSOComponent, so we need these dependencies

johnxnguyen: We need it because it must pass it along to the no history view and then via the authentication result. Unfortunately because several components don't relate to each other we can't pass them via needle, instead they must be passed through destinations.

KaterinaWire (approved):

PR #2681, opened 2025-03-17
feat: text message collapsable - WPB-16431

dmitrysimkin:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16431" title="WPB-16431" target="_blank">TaskWPB-16431</a>  [iOS] Create collapsing cell for text
  </td></table>
  <br />
 

### Issue

Added functionality to collapse text messages

### Testing

_Describe how to test._

---

### Checklist

---


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
```diff
@@ -146,13 +156,23 @@ final class ConversationMessageSectionController: NSObject, ZMMessageObserver {
...
-        let messageSupportsCollapsing = message.isFile || message.isAudio || message.isVideo || message
-            .isLocation || message.isImage
-        return message.isSentBySelfUser && messageSupportsCollapsing
+
+        if message.isText {
+            guard let textMessage = message.textMessageData?.messageText else {
+                return false
+            }
+
+            let margins = UIView().conversationHorizontalMargins
```
netbe: question: is it something that existed?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
```diff
@@ -167,6 +167,17 @@ extension ConversationContentViewController {
...
+
+    override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
+        super.viewWillTransition(to: size, with: coordinator)
+
+        coordinator.animate(alongsideTransition: nil) { _ in
+            self.dataSource.resetSectionControllers()
+            self.dataSource.reloadSections(newSections: self.dataSource.calculateSections())
+            self.tableView.reloadData()
+        }
+    }
```
netbe: question: when does this happen? and what are we doing exactly?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationCollapsedFileMessageCell.swift
```diff
@@ -97,28 +113,43 @@ final class ConversationCollapsedFileMessageCell: UIView, ConversationMessageCel
...
-        wholeViewTapButton.removeTarget(nil, action: nil, for: .allEvents)
+        if object.message.isText {
+            typeIcon.isHidden = true
+            if let textMessageData = object.message.textMessageData {
+                messageTextView.attributedText = NSAttributedString
+                    .format(
+                        message: textMessageData,
+                        isObfuscated: object.message.isObfuscated
+                    )
+            }
```
netbe, line 125: question: so we use this cell for text also? is it only one cell for all collapsing messages ? If so we should rename it

netbe (commented):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
dmitrysimkin: This is to update iPad app when you hide or show sidebar. Content area is different so we need to redraw, anyway it looks not ideal on iPad but at least it updates. I tried to improve but it was not updating easily and I had not time to fix it properly

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationCollapsedFileMessageCell.swift
dmitrysimkin: True, I'll rename it in last PRs for the feature

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
dmitrysimkin: yeah, there was `conversationHorizontalMargins` for UIView already, where for example we get margins for messages. But since `ConversationMessageSectionController` is not `UIView` I just call it on empty view

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationCollapsedFileMessageCell.swift
dmitrysimkin: It's in b3a2c804bbd0eb1c52a4967dc28ebc8562b2fc15 @netbe

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageActionController.swift
```diff
@@ -138,12 +144,13 @@ final class ConversationMessageActionController {
         case .collapse:
             guard let isCollapsed,
                   collapseOwnMessagesEnabled,
-                  !isCollapsed else {
+                  !isCollapsed,
+                  isCollapsedWasUpdated else {
```
johnxnguyen, line 148: question: why is `isCollapsedWasUpdated` necessary?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
johnxnguyen: suggestion: decouple the margins from `UIView`, e.g by moving it to an enum namespace.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
```diff
@@ -146,13 +156,23 @@ final class ConversationMessageSectionController: NSObject, ZMMessageObserver {
...
+
+            return willTextExceedOneLine(text: textMessage, availableWidth: contentWidth - margins.right - margins.left)
+        } else {
+            let messageSupportsCollapsing = message.isFile || message.isAudio || message.isVideo || message
```
johnxnguyen: suggestion: add a computed var as an extension to message so we don't have to repeat this expression.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
```diff
@@ -581,3 +609,25 @@ extension ConversationMessageSectionController: UserObserving {
...
     }
 }
+
+extension ConversationMessageSectionController {
+
+    func willTextExceedOneLine(text: String, availableWidth: CGFloat) -> Bool {
+
+        let textSize = CGSize(width: availableWidth, height: CGFloat.greatestFiniteMagnitude)
+
+        let font = UIFont.normalLightFont
```
johnxnguyen, line 619: issue: this seems brittle, if the font that is used to render the cell changes, then we'll also have to change it here, right?

wire-ios/Wire-iOS Tests/ConversationMessageCell/MessageActionsViewControllerTests.swift
```diff
@@ -166,16 +166,25 @@ final class MessageActionsViewControllerTests: XCTestCase {
...
-        let (actionController, sut) = makeSut(
+        var (actionController, _) = makeSut(
             message: message,
             sender: selfUser,
             isCollapsed: false,
             selfUserId: selfUser.remoteIdentifier
         )
         message.senderUser = selfUser
 
+        actionController.isCollapsed?.toggle()
```
netbe, line 179: question: why call it twice?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
netbe: todo: leave a comment to explain this in code, if you think there could be a better solution you can also create a follow up ticket.

```suggestion
        super.viewWillTransition(to: size, with: coordinator)
        // TODO: [WPB-XXX] this is to update the cells on iPad when sidebar 
        // is hidden or shown. This is a quick fix for now. To improve later
        coordinator.animate(alongsideTransition: nil) { _ in
            self.dataSource.resetSectionControllers()
            self.dataSource.reloadSections(newSections: self.dataSource.calculateSections())
            self.tableView.reloadData()
        }
    }
```

netbe (commented):
left some comments following the replies

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
dmitrysimkin: yes, but alternative as I can see is to have `messageTextView` instantiated and configured so we can get size, which is heavier approach and question how to pass data about layout from view (which not yet rendered) at time section controller checks which cell it needs. happy to hear better suggestions if there are 
Also there are snapshot tests so if we forgot to update font to calculate with when message font changes, so we should be pretty safe, wdyt @johnxnguyen ?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageActionController.swift
dmitrysimkin: This is for not long text messages when collapsing enabled. initially `isCollapsed = false` for shorter messages and in order not to show `Collapse` menu item I needed to filter this out, instead of calculating width again I added this flag which makes sure that only collapsable messages can be collapsed (in this case long text)

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
dmitrysimkin: Decoupled to `HorizontalMargins`

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
dmitrysimkin: Improved

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
dmitrysimkin: Improved

wire-ios/Wire-iOS Tests/ConversationMessageCell/MessageActionsViewControllerTests.swift
dmitrysimkin: you can collapse back if it was expanded first, bc of `isCollapsedWasUpdated` John asked about as well

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
johnxnguyen: Would it be possible that from here we can ask the view that would render the message if it would exceed multiple lines? It could be a static function, a kind of "preview" provided by the view responsible for rending the text with the correct font. Otherwise I see your point so as long as changes would be caught in tests then I think it's ok.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageActionController.swift
johnxnguyen: so do I understand correctly:

- a text message has `isCollapsed` set to `false`
- we check the message, if it exceeds more than one line we set `isCollapsed` to false, which also sets `isCollapsedWasUpdated` to `true`
- here when checking if the message supports collapsing we make sure only uncollapsed messages that have been explicitly set (i.e multi-line text) will have the collapse option available

Is that correct?

If so, could we perhaps rename `isCollsedWasUpdated` to something like `isMultilineTextMessage`?

dmitrysimkin: 1. yes
2. not exactly, this is `ConversationMessageActionController`, responsible for available actions that can be done with message.  `ConversationMessageActionController`lives same long as `ConversationMessageSectionController`. 
If message was collapsed and then taped -> section sets `isCollapsed = false` for action controller and `isCollapsedWasUpdated = true`. 
If message is short then it would never be expanded, and we need to show `Collapse` action only when not collapsed and collapse is possible, and `isCollsedWasUpdated` serves as easy flag to identify that, without any double calculations or decrease in performance
Overall communication between layers: views, sections, actions is very tricky
Thus, I don't think `isMultilineTextMessage` makes sense. 
wdyt @johnxnguyen ?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
dmitrysimkin: Ideally yes but under time pressure and complexity of a code that makes attributed string, it would be too heavy to call that just to calculate width. Plus we safer with snapshot tests. Considering this suggest to improve that later, wdyt @johnxnguyen  ?

wire-ios/Wire-iOS Tests/ConversationMessageCell/MessageActionsViewControllerTests.swift
netbe: let's add a comment to it then

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
netbe: if for later please left a TODO with jira ticket

netbe (approved):
approving counting on ongoing comments with John to be resolved

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
dmitrysimkin: created ticket and added in code https://wearezeta.atlassian.net/browse/WPB-16627

wire-ios/Wire-iOS Tests/ConversationMessageCell/MessageActionsViewControllerTests.swift
dmitrysimkin: added

johnxnguyen (approved):

PR #2696, opened 2025-03-18
fix: Fetch backend config info - WPB-15944

KaterinaWire:
<table>
<td>
  </td></table>
  <br />
 

### Issue

There is a bug for Path 5: **On-Prem User login without SSO**.
We were expecting to get from the GET "/get-domain-registration":
```
"domain_redirect" = backend
"backend_url": **deep link**
```
and fetch backend config from the deep link.
But it's wrong and [here's ](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/150930346/Use+case+connect+to+a+custom+backend+via+domain+redirect) what we should expect.

In short, instead of a deep link we get
{
  "config_json_url": "https://something/config.json",
  "webapp_welcome_url": "https://something.com/"
}

### Testing

Test Path 5 for new login

---

### Checklist

---


WireAuthentication/Sources/WireAuthenticationLogic/DetermineAuthMethodUseCase.swift
```diff
@@ -93,11 +93,35 @@ package struct DetermineAuthMethodUseCase: DetermineAuthMethodUseCaseProtocol {
...
+            guard let configURL = configuration.backendURL else {
                 throw AuthenticationAPIError.invalidResponse
             }
+            let backendURL = try await fetchBackendConfig(from: configURL)
+
             return .onPremLogin(email: email, backendConfig: backendURL)
         }
     }
 
+    private func fetchBackendConfig(from backendURL: URL) async throws -> URL {
```
johnxnguyen: suggestion: `fetchBackendConfigURL`

WireAuthentication/Sources/WireAuthenticationLogic/DetermineAuthMethodUseCase.swift
```diff
@@ -93,11 +93,35 @@ package struct DetermineAuthMethodUseCase: DetermineAuthMethodUseCaseProtocol {
...
+        let (data, _) = try await URLSession.shared.data(from: backendURL)
+
+        let decoder = JSONDecoder()
+        decoder.keyDecodingStrategy = .convertFromSnakeCase
```
johnxnguyen: question: is this needed if we provide explicit coding keys?

WireAuthentication/Sources/WireAuthenticationLogic/DetermineAuthMethodUseCase.swift
```diff
@@ -93,11 +93,35 @@ package struct DetermineAuthMethodUseCase: DetermineAuthMethodUseCaseProtocol {
...
                 throw AuthenticationAPIError.invalidResponse
             }
+            let backendURL = try await fetchBackendConfig(from: configURL)
+
             return .onPremLogin(email: email, backendConfig: backendURL)
         }
     }
 
+    private func fetchBackendConfig(from backendURL: URL) async throws -> URL {
+        let (data, _) = try await URLSession.shared.data(from: backendURL)
```
johnxnguyen: question: how do you intend to test this? I think you may need to inject some object into the use case so you can mock it.

johnxnguyen (approved):

WireAuthentication/Sources/WireAuthenticationLogic/DetermineAuthMethodUseCase.swift
netbe: question: we don't need extra setup of URLSession here?

netbe (approved):
Looks good

WireAuthentication/Sources/WireAuthenticationLogic/DetermineAuthMethodUseCase.swift
KaterinaWire: fixed

WireAuthentication/Sources/WireAuthenticationLogic/DetermineAuthMethodUseCase.swift
KaterinaWire: I injected URLSessionHTTPClient

WireAuthentication/Sources/WireAuthenticationLogic/DetermineAuthMethodUseCase.swift
KaterinaWire: No, we don't need because I added `CodingKeys`

WireAuthentication/Sources/WireAuthenticationLogic/DetermineAuthMethodUseCase.swift
KaterinaWire: No, we don't need

KaterinaWire: We will receive a link (example  https://wire-webapp-s3-test.s3.eu-central-1.amazonaws.com/bella.json ) and need to fetch and parse the result.

WireAuthentication/Sources/WireAuthenticationAPI/HTTPClientProtocol.swift
```diff
@@ -0,0 +1,23 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+package protocol HTTPClientProtocol: Sendable {
```
johnxnguyen: question: does this need to be in WireAuthenticationAPI? If it's only used by the use case then you could move it there.

WireAuthentication/Sources/WireAuthenticationLogic/DetermineAuthMethodUseCase.swift
```diff
@@ -24,13 +24,16 @@ package struct DetermineAuthMethodUseCase: DetermineAuthMethodUseCaseProtocol {
 
     private let validateEmailOrSSOCode: any ValidateEmailOrSSOCodeUseCaseProtocol
     private let authenticationAPI: AuthenticationAPI
+    private let httpClient: HTTPClientProtocol
 
     package init(
         validateEmailOrSSOCode: any ValidateEmailOrSSOCodeUseCaseProtocol,
-        authenticationAPI: AuthenticationAPI
+        authenticationAPI: AuthenticationAPI,
+        httpClient: HTTPClientProtocol
```
johnxnguyen: suggestion: not blocking, but another way you can achieve this is to inject a `URLSession`, then when you want to test it you can inject a session that is configured to use a mock url protocol. Take a look at the `NetworkServiceTests` to see how a mock session is used.

johnxnguyen (approved):

WireAuthentication/Sources/WireAuthenticationAPI/HTTPClientProtocol.swift
KaterinaWire: but I also need it in Tests

KaterinaWire: removed

WireAuthentication/Sources/WireAuthenticationLogic/DetermineAuthMethodUseCase.swift
KaterinaWire: Done

netbe (approved):
🚢

PR #2699, opened 2025-03-19
fix: only renew access token if user client is self client - WPB-16494

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16494" title="WPB-16494" target="_blank">BugWPB-16494</a>  [iOS/Web]: Session expiration + Not able to send message + no notification since logging into web chrome
  </td></table>
  <br />
 

### Issue

**Context**: On MLS account, when adding a user client (for example a web client), killing the iOS app and launching it again sometimes ends up with session expired on login screen. App is in a broken state if we try to log in from this point on. This seems to be an edge case because the bug is not always happening.

**Causes**: Sometimes (but not always) a request to renew the access token is triggered using the added user client ID (the web one for instance) and it ends up failing with a 403, wiping token and cookie data.

**Solution**: When falling into that edge case we want to avoid trigger a request to get the access token using the added user client ID so we add a condition to check that the added user client ID matches the self client ID (that request works fine). 
Note: on a fresh install and registering the iOS device (self client), the /access endpoint with the specified self client ID will be triggered (as it was before).   

### Testing

1. Add a user client 
2. Kill the app
3. Launch again 

App should behave normally. No login screen with expired session message, and able to send and receive MLS messages normally.


wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -837,10 +837,17 @@ public final class ZMUserSession: NSObject {
     // MARK: Access Token
 
     private func renewAccessTokenIfNeeded(for userClient: WireDataModel.UserClient) {
+        let selfUserClientID = managedObjectContext.performAndWait {
+            selfUserClient?.remoteIdentifier
+        }
+
+        let isSelfClient = selfUserClientID == userClient.remoteIdentifier
+
```
netbe: question: as this for no doubt fix the issue, I wonder how another userClient of self gets here. can we just avoid registering clients that were created by other devices.

wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests+AccessToken.swift
```diff
@@ -23,14 +23,27 @@ import WireTransport
...
+        }
+        XCTAssertTrue(waitForAllGroupsToBeEmpty(withTimeout: 0.5))
+    }
+
+    override func tearDown() {
+        super.tearDown()
+        selfClient = nil
+    }
 
     func test_itRenewsAccessTokenAfterClientRegistration_StartingFromApiV3() {
```
netbe: todo: add test that we don't renew token for other clients

netbe (commented):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
jullianm: this happens within `UserClientRequestStrategy` in `updateUpdatedObject` method where we sometimes we fall into the last `if else` block and `clientRegistrationStatus?.didRegisterMLSClient(userClient)` is called for another client of self.

I haven't investigated further on why this registration is even happening, it is clearly inconsistent and might be related to `UserClient.needsToUploadMLSPublicKeysKey`

wire-ios-sync-engine/Tests/Source/UserSession/ZMUserSessionTests+AccessToken.swift
jullianm: fixed in https://github.com/wireapp/wire-ios/pull/2699/commits/a5a61f7b6dc0e76ff21ce1e371d54811b0bb2b7a thanks

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
jullianm: I can investigate more on that @netbe (maybe on a separated non-blocking ticket related to MLS?), just wanted to have this quick fix for MLS playtest

jullianm: or maybe we can move that condition before calling `clientRegistrationStatus?.didRegisterMLSClient(userClient)` to avoid calling that method and avoid doing the logic in there (including access token renewal) if condition is not met

jullianm: updated the fix in https://github.com/wireapp/wire-ios/pull/2699/commits/3c3b0f874900774150cefa771412bf9f4f4b0a14, I moved it higher in the call stack so we don't trigger didRegisterMLSClient when client is not self

johnxnguyen (approved):
Nice catch! ⚾️

netbe (approved):
looks good

PR #2720, opened 2025-03-21
chore: Collapse final fixes - WPB-16351

dmitrysimkin:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16351" title="WPB-16351" target="_blank">StoryWPB-16351</a>  Collapse Self Message
  </td></table>
  <br />
 

### Issue

To fix:
- improved safety of calculation size for an image resize by checking divide by zero

- on iPad when changing "Collapse own messages" setting it does not update view of chat history that was before change
 
Originally I wanted to make publisher so chat screen, which remains in memory underneath settings screen, will be updated silently off screen. I've implemented publisher from `PrivateUserDefaults` but then I found that on iPad when chat screen is off screen it sets `contentWidth` for chat table view as like 1032 (full windows size, incorrect) instead of 500 or smth, not sure why. So I changed approach to update table view on when view is shown (on screen) 
Also since publisher for user default is already written and might be used later so I kept it, let me know if you don't want to keep it

- Updated copies from Astrid

### Testing

_Describe how to test._

---

### Checklist

---


WireFoundation/Sources/WireFoundation/Utilities/PrivateUserDefaults.swift
```diff
@@ -43,6 +44,17 @@ public final class PrivateUserDefaults<Key: DefaultsKey> {
 
     private func scopeKey(_ key: Key) -> String {
         "\(Self.scopePrefix(userID: userID))\(key.rawValue)"
+    }    
+}
+
+extension PrivateUserDefaults {
+    public func boolPublisher(for key: Key) -> AnyPublisher<Bool, Never> {
```
caldrian: question: since it's not used, shouldn't we remove it?

dmitrysimkin: That's what I wrote about in description, I'd like to keep it as it's already written and might be used later. Not sure what team think about it

caldrian: We usually try hard to delete dead code.

caldrian (approved):

WireFoundation/Sources/WireFoundation/Utilities/PrivateUserDefaults.swift
dmitrysimkin: For me it's not dead code, it has future potential. I spent time writing it and it might be useful in the future

dmitrysimkin: Removed

netbe (approved):
looks good to me

samwyndham:
Removing myself as this PR is approved

PR #2749, opened 2025-03-26
feat: track alert confirmation - WPB-15936

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15936" title="WPB-15936" target="_blank">TaskWPB-15936</a>  [iOS] Path 3b: Implement "email already in use" error alert
  </td></table>
  <br />
 

### Issue
Screenshot 2025-03-26 at 07 50 53

This alert has 2 buttons that open web links. When the user opens one of them and returns to the app, there is no alert and therefore no option to select the second link.

### Solution

Check if the user has pressed the confirm button and only then stop showing the alert.

### Checklist

---

### DIFF
--- a/WireAuthentication/Sources/WireAuthenticationUI/Views/Login/NoHistory/NoHistoryView.swift
+++ b/WireAuthentication/Sources/WireAuthenticationUI/Views/Login/NoHistory/NoHistoryView.swift
@@ -88,8 +88,10 @@ package struct NoHistoryView: View {
         .onAppear {
             viewModel.onAppear()
         }
-        .onReceive(NotificationCenter.default.publisher(
-            for: UIApplication.willEnterForegroundNotification)
+        .onReceive(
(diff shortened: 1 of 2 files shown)


samwyndham (approved):
Nice fix - I never thought about how to do this. Out of interest did you also try the alternative approach of setting again the alert property when selecting the how to delete account or how to change email options?

johnxnguyen (approved):

johnxnguyen:
it may be good to add a small comment explaining these changes

KaterinaWire:
> Nice fix - I never thought about how to do this. Out of interest did you also try the alternative approach of setting again the alert property when selecting the how to delete account or how to change email options?

@samwyndham Yes, I've tried. There are some issues. For example: 
- user presses OK -> alert disappears
- move app to background -> to foreground -> result: alert becomes visible again

I would keep the extra property for more resistance.

PR #2758, opened 2025-03-26
fix: MLSthumbprint not displayed after registering self MLS client - WPB-16465

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16465" title="WPB-16465" target="_blank">BugWPB-16465</a>  [iOS/Web] When a user has both iOS & Web before migrating to MLS, after MLS is enabled, iOS does not show MLS until the web client is refreshed
  </td></table>
  <br />
 

### Issue

**Context**: When MLS is enabled for a team, it registers MLS client however in devices screen MLS thumbprint doesn't show up as if the MLS client wasn't registered. MLS thumbprint only shows up when another client (web) is added. Looking at the codebase, when fetching CC `getDevicesIdentities` on that screen, it throws CC error "Couldn't find conversation". 

**Causes**: When MLS team gets enabled, we're registering the self MLS client and performing pending joins, trying to join via external commit which ends up throwing due to API failure (missing group info) because MLS conversation has not been established yet. 

**Solution**: When performing pending joins, if the epoch is 0 it's our responsibility to establish the group (i.e creating the CC conversation) and marks the conversation as MLS .ready so when we fetch CC device identities on devices screen it will find the client and shows the MLS thumbprint.

### Testing

- Create team on Jenkins
- Enable MLS for team on Jenkins (after script finishes, this will trigger an event and registers the MLSClient in the app)
- Navigate to the devices screen, MLS thumprint should now show up


wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1250,18 +1250,46 @@ public final class MLSService: MLSServiceInterface {
...
+                        return
+                    }
+
                     do {
-                        try await self.joinByExternalCommit(groupID: pendingGroup)
+                        let epoch = await context.perform { pendingGroup.epoch }
+                        let conversationExists = try await self.conversationExists(
+                            groupID: mlsGroupID
+                        )
+                        let shouldEstablishGroup = epoch == 0 && !conversationExists
```
typfel: Let's limit this to conversations of type `self`. We should not establishing `group` or `oneOnOne` conversations here that's handled else where.

typfel (changes requested):

netbe:
@jullianm retarget develop for 3.121

wire-ios-data-model/Source/MLS/MLSService.swift
jullianm: fix in https://github.com/wireapp/wire-ios/pull/2758/commits/e22a64c9ec214a6a620da9bc7615a2675634d2fd thanks

typfel (approved):

PR #2759, opened 2025-03-26
feat: implement API add channel permission - WPB-16708

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16708" title="WPB-16708" target="_blank">TaskWPB-16708</a>  [iOS] Implement API to add permission
  </td></table>
  <br />
 

### Issue

- Implements API endpoint `addChannelPermission` (API version >= v8)

### Testing

- Adds UTs


WireAPI/Sources/WireAPI/APIs/ConversationsAPI/ConversationsAPIV8.swift
```diff
@@ -72,10 +72,54 @@ final class ConversationsAPIV8: ConversationsAPIV7 {
...
+            .withBody(body, contentType: .json)
+            .build()
+
+        let (data, response) = try await apiService.executeRequest(
+            request,
+            requiringAccessToken: true
+        )
+
+        // TODO: [WPB-16708] When Swagger doc is updated by backend, ensure failure cases are correct.
+        // https://staging-nginz-https.zinfra.io/v8/api/swagger-ui/
```
jullianm, line 96: Swagger is not updated with this endpoint yet, however specs here [here](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/1761378435/Add+new+conversation+member+permission) indicates two specific error cases. 
I've also added common conversation error cases. 
We'll need to double check when Swagger doc is updated.

samwyndham (approved):
Looks good. There is a duplicate `currentDateProvider` to fix but other than that looks good.

WireAPI/Tests/WireAPITests/APIs/ConversationsAPI/ConversationsAPITests.swift
```diff
@@ -1220,6 +1252,151 @@ final class ConversationsAPITests: XCTestCase {
...
+
+        for sut in suts {
+            await XCTAssertThrowsErrorAsync(ConversationsAPIError.notAChannel) {
+                try await sut.addChannelPermission(
+                    conversationID: Scaffolding.conversationID.uuidString,
+                    conversationDomain: Scaffolding.domain,
+                    permission: .admins
+                )
+            }
+        }
```
dmitrysimkin, line 1451: suggestion: add assert on number of suts, if somehow number of sut is zero then we'd have false positive UT, here and other similar places in file

dmitrysimkin (approved):
👍  just one non blocking suggestion

WireAPI/Tests/WireAPITests/APIs/ConversationsAPI/ConversationsAPITests.swift
jullianm: done thx

PR #2763, opened 2025-03-27
feat: view channels in conversation list - WPB-16383

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16383" title="WPB-16383" target="_blank">StoryWPB-16383</a>  [iOS] Display channels within conversation list
  </td></table>
  <br />
 

### Issue

This PR shows channels in the conversation list which includes:

- Creating a new messaging database version - 2.123
- Updating ZMConversation to have an additional `groupType` property which returns `ConversationGroupType`
- Updating models that fetch conversations
- Ensuring that data is set on the ZMConversation for both legacy and WireAPI paths
- Updating `ConversationGroupAvatarView` and some of it's dependencies to conditionally show the channel icon

| iPhone | iPad |
| --- | --- |
| Simulator Screenshot - iPhone 14 - TESTS - 2025-03-26 at 16 40 01 | Simulator Screenshot - iPad (10th generation) - 2025-03-27 at 09 35 24 |

### Testing

It is hard to test at them moment as we can't create channels in the app but I can give you some credentials which already have a channel and group created.

0. Switch to Anta and set preferred API of v8
1. Log into an account with both groups and channels - ask me for credentials
2. Navigate to the conversation list
3. Not the difference in icon between a channel and a group.

---

### Checklist


dmitrysimkin (approved):
Looks good to me 👍

wire-ios-data-model/Source/Model/Conversation/ZMConversation+ConversationGroupType.swift
```diff
@@ -0,0 +1,64 @@
...
+
+            // Default to `group` type if not set.
+            guard let value = groupTypeValue?.intValue else { return .group }
+
+            return ConversationGroupType(rawValue: value)
+        }
+        set {
+            groupTypeValue = newValue.map { NSNumber(value: $0.rawValue) }
+        }
+    }
```
samwyndham, line 56: I am conflicted if this `optional ConversationGroupType` is the best way to go because it can't be used from obj-c.

I do think we want to allow the concept of nil because it makes no sense if this returns for example a default of `.group` in cases where this is a one on one conversation.

But to make this compatible with obj-c we could do something like this:

```swift
enum ConversationGroupType: Int {
    case none
    case group
    case conversation
}
```

We could then use this directly as the manage property:

```swift
@NSManaged private var groupType: ConversationGroupType
```

@netbe @El-Fitz what do you think?

wire-ios-data-model/Source/Model/Conversation/ZMConversation+ConversationGroupType.swift
```diff
@@ -0,0 +1,64 @@
...
+
+}
+
+public protocol HasConversationGroupType {
```
samwyndham, line 60: I was unable to extend the `ConversationLike` protocol because this property is not obj-c compatible

jullianm (approved):
LGTM 👍

PR #2764, opened 2025-03-27
refactor: support proxy mode - WPB-16256

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16256" title="WPB-16256" target="_blank">TaskWPB-16256</a>  [iOS] Login via email - proxy support
  </td></table>
  <br />
 

### Issue
This is a feature branch for supporting proxy mode in WireAuthentication. It contains only PR merge commits that have already been reviewed. For details about the  changes, check the PRs mentioned in the commit history.

---

### Checklist

---


dmitrysimkin (approved):

WireAuthentication/Sources/WireAuthentication/Components/LoginViaEmailComponent.swift
```diff
@@ -27,127 +27,143 @@ import WireReusableUIComponents
...
+                                    email: email,
+                                    backendEnvironment
+                                )
+                            )
+                        }
+                    } catch {
+                        await MainActor.run {
+                            dependency.router.presentAlert(for: error)
+                        }
+                    }
```
jullianm, line 93: question(non-blocking): should Component be responsible of this kind of logic ? I thought this would be more the role of the view model whereas the component here would only inject the necessary objects to the view model to perform that logic. 

I'm concerned that components start doing any kind of logic (like navigation logic, business logic or performing side effects) and they gradually become more than just objects providers.

In that sense, I would expect to see that logic performed by the view model.

jullianm (approved):
LGTM, nice work! just one non-blocking question about components

WireAuthentication/Sources/WireAuthentication/Components/LoginViaEmailComponent.swift
johnxnguyen: yes you're right, we should be injecting the bridge into the view model so this logic can go in the view model. I'll make a note to correct it for later.

PR #2766, opened 2025-03-27
fix: slow OneOnOne resolving - WPB-16735

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16735" title="WPB-16735" target="_blank">BugWPB-16735</a>  [iOS] MLS Bug - Infinite sync and slowness after enabling MLS
  </td></table>
  <br />
 

### Issue

When resolving one on one conversations, the query takes up to 80 sec to execute. This blocks the syncContext from running and the end user ends up with infinite sync.

Solution:

Breaking the query into 3 separate queries, resolve the issue.

### Testing

Tested on Jacob's device


jullianm (approved):
LGTM

caldrian (approved):

PR #2771, opened 2025-03-28
feat: process add conversation permission (private channel) event - WPB-16712

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16712" title="WPB-16712" target="_blank">TaskWPB-16712</a>  [iOS] Process new add permission event (private channels)
  </td></table>
  <br />
 

### Issue

- Process new "conversation.add-permission-update" event (when a private channel permission changes) in legacy code
- Port changes to new WireDomain code
- Store the "add_permission" value locally on slow sync (legacy and new code)
- Add UTs

### Testing

_Describe how to test._

---


WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift
```diff
@@ -224,6 +224,15 @@ public final class ConversationLocalStore: ConversationLocalStoreProtocol {
         }
     }
 
+    public func storeConversation(
```
samwyndham, line 227: suggestion: When I look at the API I expect this is about storing a new conversation, not mutating an existing conversation. I suggest a small name change:

```swift
public func updatePermission(_ permission: String, in conversation: ZMConversation) async
```

samwyndham (commented):
issue (maybe): Isn't this PR also missing code where we set the permission not through an update event... eg. after slow sync we need to set the permission. Or is this handled in another PR

WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift
jullianm: I aligned the naming with other APIs, I remember this was a suggestion I applied in an old PR. Maybe this could be a discussion for an architecture forum.

jullianm:
@samwyndham Not sure if it's in the scope of this ticket nor if it's handled by another ticket however just to be safe, I added it this logic in my last commits.

samwyndham (approved):
Nice work!

dmitrysimkin (approved):
Looks good 👍

PR #2762, opened 2025-03-26
feat: wire channels update create group screen for channels - WPB-16507

El-Fitz:
### Issue

Creates a new form for creating channels

### Testing

Will be opened on selecting "channel" in the conversation creation view.
---

### Checklist

- [x] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [x] Description is filled and free of optional paragraphs.
- [x] Adds/updates automated tests.

---

### UI accessibility checklist

_If your PR includes UI changes, please utilize this checklist:_
- [x] Make sure you use the API for UI elements that support large fonts.
- [x] All colors are taken from WireDesign.ColorTheme or constructed using WireDesign.BaseColorPalette.
- [x] New UI elements have Accessibility strings for VoiceOver.

PR #2774, opened 2025-03-28
feat: access channels - WPB-16711

dmitrysimkin:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16711" title="WPB-16711" target="_blank">TaskWPB-16711</a>  [iOS] Adapt Details screen for channels
  </td></table>
  <br />
 

### Issue

Add Channel Access level screen, UI plus integration of `groupType`, `addPermisson`. 
Part to actually hit API to change  access level is not yet implemented, will be in next PR

### Testing

_Describe how to test._

---

### Checklist

---


WireConversations/Sources/WireConversationsUI/ChannelAccess/ChannelAccessHostingController.swift
```diff
@@ -0,0 +1,82 @@
...
+        super.viewDidLoad()
+
+        title = L10n.Localizable.ChannelAccessLevel.navigationTitle
+        view.backgroundColor = SemanticColors.View.backgroundDefault
+
+        navigationItem.rightBarButtonItem = UIBarButtonItem(
+            barButtonSystemItem: .close,
+            target: self,
+            action: #selector(didTapClose)
+        )
```
dmitrysimkin, line 57: That has to be done here so swift UI has nice integration in UIKit world and we show navigation title and bar button properly as in design

WireConversations/Tests/WireConversationsUITests/Resources/ReferenceImages/ChannelAccessViewSnapshotTests/testPrivateAccessLevel.light.png
samwyndham, line 1: thought: Considering that these render incorrectly (in tests) I'm not sure we should bother adding snapshot tests for them as it can make things confusing. If we keep them, consider making a refactoring Jira ticket where we move the tests to another test target that uses a hosting controller.

WireConversations/Sources/WireConversationsImplementation/ChannelAccessLevel/ChannelAccessUseCase.swift
```diff
@@ -0,0 +1,52 @@
...
+// sourcery: AutoMockable
+package protocol ChannelAccessUseCaseProtocol {
+    var settings: ChannelAccessSettings { get }
+    func updateAccessLevel(to level: ChannelAccessLevel)
+    func updateParticipantPermission(to permission: ChannelAccessLevelPermission)
+}
+
+package class ChannelAccessUseCase: ChannelAccessUseCaseProtocol {
+
+    package var settings: ChannelAccessSettings
```
samwyndham: suggestion: Make `private(set)`

WireConversations/Sources/WireConversationsUI/ChannelAccess/ChannelAccessViewModel.swift
```diff
@@ -0,0 +1,70 @@
...
+            applyAccessLevel(level)
+        }
+    }
+
+    func confirmPrivateAccessChange() {
+        applyAccessLevel(.private)
+    }
+
+    private func applyAccessLevel(_ level: ChannelAccessLevel) {
+        useCase.updateAccessLevel(to: level)
```
samwyndham: question/issue: I don't quite understand the point of the use case. As far as I understand it updates it's internal `settings` value but that value is never read. Am I misunderstanding something?

samwyndham (commented):
Hey @dmitrysimkin - I added a comment for clarification. Once that is answered re-request a review from me and I will continue reviewing.

WireConversations/Tests/WireConversationsUITests/Resources/ReferenceImages/ChannelAccessViewSnapshotTests/testPrivateAccessLevel.light.png
dmitrysimkin: I know it's not perfect, but I'd rather have them still, visuals are not so critical, separator and corners rounding a bit off yes, but at least we have other UI. Btw, in Thomas's PR [https://github.com/wireapp/wire-ios/pull/2762/files#diff-5669edf24789c652f2bafe25b59184e17eb501400b322b5a077f83dc1c876702](url) we have the same. Let's keep them now and if we see issues we can remove them pretty easily, wdyt @samwyndham ?

WireConversations/Sources/WireConversationsUI/ChannelAccess/ChannelAccessViewModel.swift
dmitrysimkin: it's used in     `func applyAccessLevel(_ level: ChannelAccessLevel) async `. Idea of use case is that it contains the logic and business rules, when we update access level or permissions, and VM uses useCase's settings for UI representation. Would that be clearer now @samwyndham ?

WireConversations/Sources/WireConversationsImplementation/ChannelAccessLevel/ChannelAccessUseCase.swift
dmitrysimkin: As I described below, it's used in VM

samwyndham: But this property is only _read_ from the VM right? The VM doesn't directly write to it. Therefore it can be `private(set)` which still allows reading. The advantage is that we then enforce that it is only updated methods such as `applyAccessLevel`

WireConversations/Sources/WireConversationsUI/ChannelAccess/ChannelAccessViewModel.swift
samwyndham: Oh yeah, somehow I missed it was being read.

WireConversations/Sources/WireConversationsImplementation/ChannelAccessLevel/ChannelAccessUseCase.swift
```diff
@@ -0,0 +1,67 @@
...
+            // TODO: [WPB-16860] https://wearezeta.atlassian.net/browse/WPB-16860
+            accessLevel: .private,
+//            accessLevel: permission == nil ? .public : .private,
+            participantPermission: permission ?? .adminsAndMembers
+        )
+        self.settings = settings
+        self.repository = repository
+    }
+
+    public func updateAccessLevel(to level: ChannelAccessLevel) async throws {
```
samwyndham:
```suggestion
    public func updateAccessLevel(to level: ChannelAccessLevel) async throws -> ChannelAccessSettings {
```
suggestion: Return the updated settings so the the compiler warns us if we are forgetting to make use of the new settings value.

WireConversations/Sources/WireConversationsUI/ChannelAccess/ChannelAccessView.swift
```diff
@@ -0,0 +1,175 @@
...
+package import SwiftUI
+import WireConversationsAPI
+import WireConversationsImplementation
+import WireConversationsImplementationSupport
+import WireDesign
+
+package struct ChannelAccessView: View {
+
+    @ObservedObject var viewModel: ChannelAccessViewModel
+    @Environment(\.dismiss) private var dismiss
```
samwyndham: question: Is this needed? I don't see it being called

WireConversations/Sources/WireConversationsUI/ChannelAccess/ChannelAccessViewModel.swift
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    func selectParticipantPermission(_ permission: ChannelAccessLevelPermission) async {
+        await applyParticipantPermission(permission)
+    }
+
+    func applyParticipantPermission(_ permission: ChannelAccessLevelPermission) async {
+        isLoading = true
+        try? await useCase.updateParticipantPermission(to: permission)
+        settings.participantPermission = permission
```
samwyndham: question: Should this be `settings = useCase.settings`? Otherwise we miss out on any of the internal logic that `updateParticipantPermission` might perform and it is confusing because it uses a different pattern from `applyAccessLevel`

WireConversations/Sources/WireConversationsUI/ChannelAccess/ChannelAccessViewModel.swift
```diff
@@ -0,0 +1,80 @@
...
+        try? await useCase.updateAccessLevel(to: level)
+        settings = useCase.settings
+        isLoading = false
+    }
+
+    func selectParticipantPermission(_ permission: ChannelAccessLevelPermission) async {
+        await applyParticipantPermission(permission)
+    }
+
+    func applyParticipantPermission(_ permission: ChannelAccessLevelPermission) async {
```
samwyndham: suggestion: If possible make this one private

samwyndham (approved):
Nice work! 
Personally I find it a little confusing having the both the view model and use case having a settings property. When reading the code it makes me wonder what is the source of truth. You could consider making the vm return the settings from the use case and calling `objectWillChange.send()` on the view model after calling methods on the use case:

```swift
package class ChannelAccessViewModel: ObservableObject {

    var settings: ChannelAccessSettings {
        useCase.settings
    }

    // ...

    func applyAccessLevel(_ level: ChannelAccessLevel) async {
        isLoading = true
        try? await useCase.updateAccessLevel(to: level)
        objectWillChange.send() //<-----------
        isLoading = false
    }
}
```

dmitrysimkin:
> Nice work! Personally I find it a little confusing having the both the view model and use case having a settings property. When reading the code it makes me wonder what is the source of truth. You could consider making the vm return the settings from the use case and calling `objectWillChange.send()` on the view model after calling methods on the use case:
> 
> ```swift
> package class ChannelAccessViewModel: ObservableObject {
> 
>     var settings: ChannelAccessSettings {
>         useCase.settings
>     }
> 
>     // ...
> 
>     func applyAccessLevel(_ level: ChannelAccessLevel) async {
>         isLoading = true
>         try? await useCase.updateAccessLevel(to: level)
>         objectWillChange.send() //<-----------
>         isLoading = false
>     }
> }
> ```

for me they serve different roles, in use case it's source of truth, in VM it's for reactive way to update UI. in my implementation or you in your suggestion we would have 2 manual calls to set VM's settings or `objectWillChanged.send()`. For me both approaches are pretty similar

WireConversations/Sources/WireConversationsUI/ChannelAccess/ChannelAccessView.swift
dmitrysimkin: well spotted, not needed anymore

WireConversations/Sources/WireConversationsUI/ChannelAccess/ChannelAccessViewModel.swift
dmitrysimkin: yeah, updated, thanks

WireConversations/Sources/WireConversationsAPI/ChannelAccessModel.swift
```diff
@@ -0,0 +1,43 @@
...
+import Foundation
+
+public enum ChannelAccessLevel {
+    case `public`
+    case `private`
+}
+
+public enum ChannelAccessLevelPermission {
+    case admins
+    case adminsAndMembers
```
jullianm: suggestion: When calling the add permission endpoint, we use the `ChannelPermission`  enum with two cases : admins or everyone. Maybe we should be consistent with that naming here. What do you think ?

wire-ios-data-model/Source/Model/Conversation/ZMConversation+PrivateChannelPermission.swift
```diff
@@ -18,21 +18,54 @@
...
+        case everybodyPermissionKey:
+            .everybody
+        default:
+            nil
+        }
+    }
+}
+
+private let adminsPermissionKey = "admins"
+private let everybodyPermissionKey = "everybody"
```
jullianm: shouldn't it be`everyone` ?

wire-ios-data-model/Source/Model/Conversation/ZMConversation+PrivateChannelPermission.swift
```diff
@@ -18,21 +18,54 @@
...
-
-    public var channelPermission: String? {
+    public var accessLevelPermission: ChannelAccessLevelPermission? {
         get {
-            privateChannelPermissionValue
+            guard conversationType == .group else { return nil }
+
+            switch privateChannelPermissionValue {
+            case adminsPermissionKey: return .admins
+            case everybodyPermissionKey: return .everybody
```
jullianm: question: did these values change ? I think the values are "admins" or "everyone"

wire-ios/Wire-iOS/Sources/ChannelAccessRepository.swift
```diff
@@ -0,0 +1,72 @@
...
+//
+
+import WireAPI
+import WireConversationsAPI
+import WireConversationsImplementation
+import WireDomain
+import WireSyncEngine
+import WireTransport
+
+class ChannelAccessRepository: ChannelAccessRepositoryProtocol {
```
jullianm, line 26: question: Is there a particular reason why this repository is at the app level ? I think it should be part of WireDomain with all the other repositories (all our new code is there)

wire-ios/Wire-iOS/Sources/ChannelAccessRepository.swift
```diff
@@ -0,0 +1,72 @@
...
+
+    private let conversationID: String
+    private let conversationDomain: String
+    private let session: ZMUserSession
+
+    init(
+        conversationID: String,
+        conversationDomain: String,
+        session: ZMUserSession
```
jullianm, line 35: suggestion: after moving this repository to WireDomain, you can directly inject API service and the context / ConversationsLocalStore. Note: ZMUserSession will not be reachable from WireDomain.

wire-ios/Wire-iOS/Sources/ChannelAccessRepository.swift
```diff
@@ -0,0 +1,72 @@
...
+        self.session = session
+    }
+
+    func updateParticipantPermission(to permission: WireConversationsAPI.ChannelAccessLevelPermission) async throws {
+
+        guard let backendInfoApiVersion = BackendInfo.apiVersion,
+              let apiVersion = WireAPI.APIVersion(rawValue: UInt(backendInfoApiVersion.rawValue)),
+              let apiService = session.apiService else { return }
+
+        let context = session.syncContext
```
jullianm: question: is it needed ?

jullianm (commented):
looks good overall, left a couple of comments before approving

wire-ios/Wire-iOS/Sources/ChannelAccessRepository.swift
jullianm: I guess yes you're going to need it to update locally the database.

wire-ios/Wire-iOS/Sources/ChannelAccessRepository.swift
dmitrysimkin: This is what I'd like to discuss tomorrow at Acrh forum, I just didn't want to add dependencies to 
`WireAPI, WireDomain, WireSyncEngine, WireTransport`, to me it's too much details for a feature to know. I'll address this based on solution we'd have

wire-ios/Wire-iOS/Sources/ChannelAccessRepository.swift
dmitrysimkin: Do we really need that? I can see there is event `add-permissions ` comes and is saved in database, do we need two saves for one action?

WireConversations/Sources/WireConversationsAPI/ChannelAccessModel.swift
dmitrysimkin: Agree, when I started I didin't know about tech details so I used values form UI, updated

wire-ios-data-model/Source/Model/Conversation/ZMConversation+PrivateChannelPermission.swift
dmitrysimkin: updated

wire-ios-data-model/Source/Model/Conversation/ZMConversation+PrivateChannelPermission.swift
dmitrysimkin: as my comment above, agree and updated

wire-ios/Wire-iOS/Sources/ChannelAccessRepository.swift
jullianm: Yes, I don't think we need that, we receive the event after changing the permission which already performs that logic.

wire-ios/Wire-iOS/Sources/ChannelAccessRepository.swift
jullianm: yes, let's discuss this tomorrow in architecture forum 👍

jullianm (approved):
Nice work, approving but let's discuss the repository topic tomorrow in architecture forum

El-Fitz (commented):
I'm late to the party, but I've noticed the `WireConversationsImplementation` is both imported in the UI target and publicly exposed. We should discuss this in the architecture forum.

PR #2784, opened 2025-04-01
refactor: process team create event - WPB-15052

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15052" title="WPB-15052" target="_blank">TaskWPB-15052</a>  [iOS] Process team.create event in new event processor
  </td></table>
  <br />
 

### Issue

Previously we didn’t handle `team.create` event so it was never added to the new sync refactoring. 
But now with individual to team migration, we do need to handle it. 
This PR ports the changes from legacy code to the new code.

### Testing
 
N/A

### Checklist

---


WireDomain/Sources/WireDomain/Event Processing/TeamEventProcessor/TeamCreateEventProcessor.swift
```diff
@@ -0,0 +1,41 @@
...
+        await repository.createOrUpdateTeam(
+            identifier: identifier,
+            name: name,
+            creator: creator,
+            icon: icon,
+            iconKey: iconKey
+        )
+    }
+
+}
```
samwyndham, line 41: thought (non-blocking): It is interesting that we have this type that doesn't really do anything (only that it forwards some data) and it has a test that doesn't really test anything (only that the data is forwarded). Is our code becoming too decoupled?

WireDomain/Tests/WireDomainTests/Event Processing/TeamEventProcessor/TeamCreateEventProcessorTests.swift
```diff
@@ -0,0 +1,69 @@
...
+
+        teamRepository.createOrUpdateTeamIdentifierNameCreatorIconIconKey_MockMethod = { _, _, _, _, _ in }
+
+        // When
+
+        try await sut.processEvent(Scaffolding.event)
+
+        // Then
+
+        XCTAssertEqual(teamRepository.createOrUpdateTeamIdentifierNameCreatorIconIconKey_Invocations.count, 1)
```
samwyndham, line 56: suggestion: assert on the resulting value to ensure all properties are mapped correctly which is the more likely bug I think thank forgetting to call the method.

samwyndham (approved):
Looks good. Just left some subjective opinions.

WireAPI/Sources/WireAPI/Models/UpdateEvent/TeamEvent/TeamCreateEvent.swift
```diff
@@ -0,0 +1,65 @@
...
+
+    public let creator: UUID
+
+    /// The team icon.
+
+    public let icon: String
+
+    /// The team icon key.
+
+    public let iconKey: String?
```
dmitrysimkin, line 43: question: What is the difference between icon which is string and iconKey which also string?

dmitrysimkin (approved):
👍

netbe (approved):
LGTM;)

WireAPI/Sources/WireAPI/Models/UpdateEvent/TeamEvent/TeamCreateEvent.swift
jullianm: I think, without being 100% sure that  it's related to asset encryption, from what I see in doc:
```The icon_key field is ignored. The asset is not encrypted.```
So I'm not even sure this is used.

WireDomain/Tests/WireDomainTests/Event Processing/TeamEventProcessor/TeamCreateEventProcessorTests.swift
jullianm: There's no returning value from `createOrUpdateTeam` so we simply test that the call is properly forwarded to the repo

WireDomain/Sources/WireDomain/Event Processing/TeamEventProcessor/TeamCreateEventProcessor.swift
jullianm: To me it's one of the clean architecture bottleneck (or at least something I find funny sometimes) where you can end up having types that only forward data to another object (here to the repository) when there's not much else to do. 

And here we start to experience that with many objects like processors / repositories / sync objects (which are also owned by repo) / local stores which are sometimes don't bring any additional benefits.

So, maybe the architecture could be simplified a bit: removing the repository layer to only keep sync objects as (I think) they're kind of doing what the repo does (fetching data from backend using the API objects and storing them locally using the stores)

PR #2799, opened 2025-04-03
feat: improve channels data model - WPB-16907

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16907" title="WPB-16907" target="_blank">TaskWPB-16907</a>  [iOS] Improve channels core data model properties
  </td></table>
  <br />
 

### Issue

This PR makes some changes to channel related properties of `ZMConversation` as discussed in the architecture forum. These changes simplify the code and, according to apple docs, can improve performance. Specifically it:

- introduces a `none` field to `ConversationGroupType`.
- It stores `ConversationGroupType` directly in the DB without the need of a custom getter/setter
- It renames the core data property `groupTypeValue` to `groupType`
- It introduces `ConversationLike.isChannel`

### UPDATE

This PR now also simplifies the `privateChannelPermission` property:

- Renames `privateChannelPermissionValue` to `privateChannelPermission`
- Makes it an Int16 instead of String. This is because a. it is potentially more performant and b. string is adding no benefits as we are already parsing the "string" value before setting it on the DB
- Removes the need for custom getters / setters.
- Adds an `unset` value to represent the unset state.

There was some different terminology being used to represent this property - privateChannelPermission vs accessLevelPermission. I kept `privateChannelPermission` as this property only relates to private channels.

**To simply processes, this PR should not be merged until the other current channel related work is complete.**

### Testing

Test creating and filtering channels is working as expected.

---

### Checklist


dmitrysimkin (approved):
Looks good 👍

johnxnguyen (approved):

PR #2809, opened 2025-04-07
fix: use fetchRequest instead of in memory predicate - WPB-17064

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17064" title="WPB-17064" target="_blank">BugWPB-17064</a>  [iOS] hang on conversation list
  </td></table>
  <br />
 

### Issue

When resuming the app, the conversation lists are recreated and conversations filtered and sort again. This is done in memory and causes a hang when applying the filtering (touching conversationType).

Use fetchRequests for all filters instead.

Note: Eventually, we could replace ConversationList and ConversationListSnapshot with just a fetchrequestController that would be applied a predicate each time.

### Testing

1. enable hang detection
2. put app in background
3. put app in foreground
Observe  no hang is produced

---

### Checklist

---


dmitrysimkin (approved):
Looks good 👍 
Only concern, is it covered by any test, implicitly or explicitly?

samwyndham (approved):
I'm actually very surprised this is an improvement. Any idea what specifically was the bottleneck before?

netbe:
> Looks good 👍 Only concern, is it covered by any test, implicitly or explicitly?

@dmitrysimkin yes actually I forgot to update the tests

KaterinaWire (approved):

PR #2823, opened 2025-04-07
fix: unable to pull some conversations - WPB-17015

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17015" title="WPB-17015" target="_blank">BugWPB-17015</a>  [iOS] Failed to pull conversation
  </td></table>
  <br />
 

### Issue
When performing the new initial sync, there were decoding errors when fetching all conversations. Some conversations had an `epoch_timestamp` value that had fractional milliseconds, and some had non-fractional milliseconds. The swagger docs say they should all have non-fractional milliseconds.

While backend should fix it on their side, it's safest for the iOS client to try to decode both time formats. This PR removes `UTCTimeMillis` (originally designated for fractional millisecond timestamps), and makes `UTCTime` (originally designated for non-fractional timestamps) to attempt decoding both formats.

---

### Checklist

---


dmitrysimkin (approved):
Looks good 👍 
One question however, do we need to add test to ensure both formats of timestamp is parsed correctly? Would be nice to have if easy to add

johnxnguyen:
> Looks good 👍 One question however, do we need to add test to ensure both formats of timestamp is parsed correctly? Would be nice to have if easy to add

@dmitrysimkin currently there many tests that have both formats decoded into the same `UTCTime` type, so it is implicitly well covered, but I will add an additional test to be more explicit.

WireAPI/Sources/WireAPI/Components/UTCTime.swift
```diff
@@ -32,12 +31,18 @@ struct UTCTime: Codable {
...
+        // single time object.
         if let date = ISO8601DateFormatter.internetDateTime.date(from: string) {
             self.date = date
+        } else if let date = ISO8601DateFormatter.fractionalInternetDateTime.date(from: string) {
+            self.date = date
         } else {
             throw DecodingError.dataCorruptedError(
                 in: container,
-                debugDescription: "Expected date string to be ISO8601-formatted without fractional seconds"
+                debugDescription: "Expected date string to be ISO8601-formatted either with or without fractional seconds"
```
caldrian, line 45: note: When using `ISO8601FormatStyle` we don't need to throw any error manually, unless you wanted to throw a `DecodingError` explicitly.

```
// format styles instead of date formatter instances

let iso8601FormatStyle =
    Date.ISO8601FormatStyle()
let iso8601FormatStyleIncludingFractionalSeconds =
    Date.ISO8601FormatStyle(includingFractionalSeconds: true)

// succeeds
try iso8601FormatStyle.parse("2025-04-08T10:16:00Z")
// throws
try iso8601FormatStyleIncludingFractionalSeconds.parse("2025-04-08T10:16:00Z")

// throws
try iso8601FormatStyle.parse("2025-04-08T10:16:00.774Z")
// succeeds
try iso8601FormatStyleIncludingFractionalSeconds.parse("2025-04-08T10:16:00.774Z")
```

caldrian (approved):

WireAPI/Sources/WireAPI/Components/UTCTime.swift
johnxnguyen: Cool, I didn't know! I will keep it as it is now as I'd like to move on but I'll add a note to look at this.

PR #2825, opened 2025-04-07
fix: Request loop on /v7/search/contacts - WPB-16757

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16757" title="WPB-16757" target="_blank">BugWPB-16757</a>  [iOS] Request loop on /v7/search/contacts with long domain names
  </td></table>
  <br />
 

### Issue

We have a request loop on `/v7/search/contacts` because the user who is actually searching for a federated user is sending many identical requests, which causes the request loop message.
Why this happens:
- there so me restrictions for searching federated users, described here: https://wearezeta.atlassian.net/browse/WPB-14455.
-  we always replace the domain part for searched federated users.
As a result, the user enters the long part of the domain of the searching user, but in fact sends the same request over and over again.

### Solution

Do not replace the domain part .

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/UserSession/Search/SearchRequest.swift
+++ b/wire-ios-sync-engine/Source/UserSession/Search/SearchRequest.swift
@@ -112,9 +112,14 @@ public struct SearchRequest {
         searchOptions: SearchOptions,
         team: Team? = nil
     ) {
-        let (query, parsedDomain) = Self.parseQuery(query)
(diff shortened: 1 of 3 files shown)


johnxnguyen (approved):
Could we add a unit test for this change?

dmitrysimkin (approved):
👍

PR #2828, opened 2025-04-08
chore: don't show read receipt alert - WPB-16980

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16980" title="WPB-16980" target="_blank">BugWPB-16980</a>  [iOS] channels - error when enabling read receipts 
  </td></table>
  <br />
 

### Issue

On MLS read receipts are not currently enabled. As channels are only on MLS this affects them too. Currently on iOS when we attempt to toggle the read receipt setting, it fails, the toggle is reverted and an error message is shown. In this PR we just match web behavior - the toggle behavior is the same but we don't show the error alert.

https://github.com/user-attachments/assets/3a311bd8-6b3a-4d75-9120-2ae96c23a4d0

### Testing

1. Enter a channel where your are an admin
2. Navigate to the conversation details
3. Enable Read receipts
4. Assert that the toggle is re-disabled
5. Assert that no error message is shown

---

### Checklist


dmitrysimkin (approved):

jullianm (approved):

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/Sections/ReceiptOptionsSectionController.swift
```diff
@@ -98,7 +98,6 @@ final class ReceiptOptionsSectionController: GroupDetailsSectionController {
                 switch result {
                 case .failure:
                     cell.configure(with: conversation)
-                    self?.presentingViewController?.present(UIAlertController.checkYourConnection(), animated: true)
```
netbe, line 101: suggestion: leave a comment to explain why?, 
2) I wonder if it should be only for mls

netbe (approved):
left  a suggestion

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/Sections/ReceiptOptionsSectionController.swift
samwyndham: 👍 I've implemented your suggestion

samwyndham:
This PR now also fixes:

- The group icon during the create channel flow which currently has the wrong aspect ratio
- Snapshot tests
- Not running tests in WireConversations on CI

El-Fitz (approved):

PR #2838, opened 2025-04-08
feat: don't ask for password to delete sso clients - WPB-17020

KaterinaWire:
### Issue
If an sso user tries to log into an account with multiple devices and is prompted to remove one of the devices, an error is displayed after submitting the request with a password:

  ```
      - code : 403
      - label : WireRequestStrategy.Payload.ResponseFailure.Label.invalidCredentials
      - message : "Authentication failed"
      - data : nil
```

### Solution

In fact, we shouldn't be asking sso users for password, and this bug has probably been around for a long time. 
For example, we don't ask sso user for a password from the Devices list. I also confirmed this with other platforms.

There is also an old [Jira ticket ](https://wearezeta.atlassian.net/browse/WPB-13484)that is about omitting password for sso users. This ticket is not specific to this case, but it does prove the theory that we shouldn't even ask SSO users for their password.

### Testing

Details it the ticket

### Checklist

---


netbe (approved):
verified 🚢

suggestion: It might be good to cover with a test for the long term

johnxnguyen (approved):

PR #2839, opened 2025-04-08
fix: don't process duplicate update events - WPB-16165

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16165" title="WPB-16165" target="_blank">TaskWPB-16165</a>  [iOS] Prevent duplicate event processing
  </td></table>
  <br />
 

### Issue
It's possible to process events twice which happens in the following scenario: 
- an incremental sync is started, which opens the push channel, but all events are buffered
- the user receives a new event, so the backend sends the event as a live event through the push channel as well as add it to the pending event queue
- the incremental sync fetches pending events, including the one that was just added
- the pending events are all processed
- the push channel events are released and processed
- since the new event was received twice, it would be processed twice.

The solution is to de-duplicate live events by:
- keeping track of all envelope ids that were processed before the push channel is iterated
- when iterating the live events, skip any that have been previously processed

### Testing
Not easy to test, but if you want to try:
1. Amass several pending events (150 or so)
2. Open the app to initiate a sync, while still receiving a large number of events
3. Assert nothing appears twice.
4. Better yet, check datadog for the log `"live event already processed, skipping..."`

### Note
Any encrypted event wouldn't be processed twice anyway since we only decrypt them once and ignore them if they are decrypted a second time. However unencrypted events (e.g conversation add, member join, etc) don't have this benefit, though we probably already safely process them multiple times.

---

### Checklist

---


WireAPI/Sources/WireAPI/APIs/PushChannelAPI/PushChannelAPI.swift
```diff
@@ -24,6 +25,9 @@ public protocol PushChannelAPI {
     /// - Parameter clientID: The id of the self client.
     /// - Returns: A push channel.
 
-    func createPushChannel(clientID: String) async throws -> any PushChannelProtocol
+    func createPushChannel(clientID: String) async throws -> AnyPushChannel
 
 }
+
+// Workaround for automockable compiler error.
+public typealias AnyPushChannel = any PushChannelProtocol
```
dmitrysimkin, line 33: 👍

dmitrysimkin (approved):
👍

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -61,45 +66,76 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
+                            logger.debug(
+                                "processing live event: \(event.name)",
+                                attributes: [.eventEnvelopeID: envelope.id]
+                            )
                             try await processor.processEvent(event)
                         } catch {
-                            logger.error("failed to process live event: \(String(describing: error))")
+                            logger.error(
+                                "failed to process live event: \(String(describing: error))",
+                                attributes: [.eventEnvelopeID: envelope.id]
```
jullianm, line 120: question(non-blocking): I wonder what happens to failed events,  are they considered as processed and then deleted from the db?

jullianm (approved):

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
johnxnguyen: yes, we just skip them. But we have another ticket to further inspect this and decide an appropriate behavior.

PR #2841, opened 2025-04-08
fix: UI not up to date after event processing - WPB-16737

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16737" title="WPB-16737" target="_blank">BugWPB-16737</a>  [iOS] UI not updated when events are processed
  </td></table>
  <br />
 

### Issue
After completing an incremental sync, the UI doesn't show the latest changes and is out of date, e.g messages that were received aren't visible. Only if the app is brought back to the foreground does the UI update.

The cause is simply that the sync context, which is mutated during event processing, isn't saved after event processing, so the changes are not yet known to the UI.

The fix is simply to save the database after processing events.

### Testing
1. Launch app with new sync
2. Receive events while the app is in the foreground
3. Verify that those events are visible immediately.

---

### Checklist

---


WireDomain/Tests/WireDomainTests/Synchronization/IncrementalSyncTests.swift
```diff
@@ -156,6 +163,10 @@ final class IncrementalSyncTests: XCTestCase {
 
         // Then live events were deleted (duplicates skipped).
         XCTAssertEqual(store.deleteEventEnvelopeAtIndex_Invocations, [11, 12])
+
+        // Then the database was saved once after processing pending events
+        // and once after processing each live event.
+        XCTAssertEqual(databaseSaver.save_Invocations.count, 3)
```
dmitrysimkin, line 169: question: We have save operation for each event, right? I remember when we brainstormed issues with chat performance @netbe mentioned that we might need to optimise database so UI does not receive a lot of updates. If we save for each event it might cause UI to re-render too much. 

I do not know how new sync works just wanted to raise a discussion if needed

netbe: good point @dmitrysimkin. I think if something need to change is not how frequent saves are, but how frequent we notify the UI. See NotificationDispatcher

netbe (approved):
Simple fix :)

jullianm (approved):
LGTM

WireDomain/Tests/WireDomainTests/Synchronization/IncrementalSyncTests.swift
johnxnguyen: In this flow, there are two main places we save:
1. once per batch of accumulated events (a batch is 500, so for every 1000 accumulated events there are 2 saves)
2. once per live event, which occur in real time.

So in short, we already optimize the amount of saves where it matters most.

dmitrysimkin (approved):

PR #2844, opened 2025-04-08
fix: renew access token - WPB-16808

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16808" title="WPB-16808" target="_blank">BugWPB-16808</a>  [iOS] Access token is not upgraded on login with existing client
  </td></table>
  <br />
 

### Issue

**Context**:  on MLS, when the access token/ cookie is invalidated, it triggers a logout and if we try to login again it will fail to send messages, backend returning 400 "Header Z-Client is required" meaning the bearer token we pass in the request is not valid.

**Reasons**:  some MLS endpoints require an access token with the client ID whereas in codebase some nested methods to get the access token don't pass the client ID in the request. This use case happens specifically when the user is logged out unexpectedly because on logging in again  it will try to get a new access token without the client ID. 

_Note_: It works fine on initial sync because we get an access token using the client ID after registering the self client.

**Solutions**:  Make sure that if a self client is stored locally, we use its ID to get the access token in the legacy methods.
--> Inject the selfClientID to ZMTransportSession at initialization. 

### Testing

in DEBUG ACTIONS panel, select new action "CLEAR ACCESS TOKEN & COOKIE (forces logout)" then login again:

Before -> MLS messages could not be sent
Now -> MLS Messages are successfully sent

### Checklist

---


wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DebugActions/DeveloperDebugActionsViewModel.swift
```diff
@@ -71,10 +76,27 @@ final class DeveloperDebugActionsViewModel: ObservableObject {
...
+        let accessTokenHandler = userSession?.transportSession.accessTokenHandler
+
+        let responseFailure = ZMTransportResponse(
+            payload: nil,
+            httpStatus: 400,
+            transportSessionError: nil,
+            apiVersion: APIVersion.v0.rawValue
+        )
+
+        accessTokenHandler?.processAccessTokenResponse(responseFailure)
```
jullianm, line 96: to help debug: it will clear access token and cookie and triggers a logout

wire-ios-transport/Source/TransportSession/ZMTransportSession.m
```diff
@@ -612,7 +618,7 @@ @implementation ZMTransportSession (RequestScheduler)
 
 - (void)sendAccessTokenRequest;
 {
-    [self.accessTokenHandler sendAccessTokenRequestWithURLSession:self.sessionsDirectory.foregroundSession];
+    [self.accessTokenHandler sendAccessTokenRequestWithURLSession:self.sessionsDirectory.foregroundSession clientID:_selfClientID];
```
jullianm, line 621: this fixes the issue. the legacy methods that were calling `sendAccessTokenRequest` (from what I saw when debugging) are `processBackoffItem` and `prepareForSuspendedState`, we'll now use the selfClientID if we have one

wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DebugActions/DeveloperDebugActionsViewModel.swift
```diff
@@ -71,10 +76,27 @@ final class DeveloperDebugActionsViewModel: ObservableObject {
...
+        onDismiss?()
```
netbe, line 97: question: what does onDismiss do?

netbe (commented):
question: is there anything to do on new stack WireAPI?

wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DebugActions/DeveloperDebugActionsViewModel.swift
jullianm: Debug panel would not appear again after triggered once because there's a flag turned to false when dismissing, I needed to call this to make sure debug panel would show up again

jullianm: (to make sure the flag is correctly turned off) so it can be presented again

jullianm:
> question: is there anything to do on new stack WireAPI?

@netbe no, nothing to do, in `AuthenticationManager` we already pass the selfClientID (optional) and use it if we have one

netbe (approved):
LGTM;)

wire-ios-notification-engine/Sources/NotificationSession.swift
```diff
@@ -168,6 +168,10 @@ public final class NotificationSession {
 
         let credentials = environment.proxy.flatMap { ProxyCredentials.retrieve(for: $0) }
 
+        let selfClientID = coreDataStack.syncContext.performAndWait {
+            ZMUser.selfUser(in: coreDataStack.syncContext).selfClient()?.remoteIdentifier
+        }
+
```
dmitrysimkin, line 174: 2 questions here
1. there is `accountIdentifier` already passed in init, it's not the same? What is the difference 
2. `performAndWait` is sync method, my concern is that core data fetch might take some time and maybe it's worth to fetch `selfUser.remoteIdentifier` on caller side and asynchronously? Just to avoid maybe any concurrency issues, I do not know on what thread this init would be called but wanted just to ask if it's relevant

Same in other similar places in 2 more files

netbe: 1. probably the same you're right, @jullianm can you double check
2. that is called on a background thread

jullianm: 1. the account ID is related to the user account (he can have several accounts  and switch from one to another) while the user clients are the devices he's using for that account. So a user can have multiple accounts and for each account, multiple devices. Here's some documentation that can help : https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/15958580/Client+device

2. We use performAndWait (sync) because we need that value to initialize the ZMTransportSession which is done synchronously. This is basically the place where we setup all the dependencies. I agree this is not ideal but to access that value from the nested legacy methods that triggers the access token request (and fixes the bug), I didn't find a better solution.

dmitrysimkin: 1. Ok, makes sense but a bit confusing (C'est la vie)
2. From what I see `NotificationSession` is recreated in 
```
override func didReceive(
        _ request: UNNotificationRequest,
        withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void
```

Is it possible to fetch self user id async there and pass `remoteIdentifier` which is just UUID down the chain? 
We already have quite some issues with performance and concurrency, if we make work a bit more async it could be small win. 
Also I do not know much that part of code yet and what are deadlines are so it's completely yours call @jullianm , just as an idea

dmitrysimkin (approved):

PR #2873, opened 2025-04-10
chore: a few preparations before integrating multi-platform lib - WPB-14598

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14598" title="WPB-14598" target="_blank">TaskWPB-14598</a>  [iOS] Add shared backup library as a dependency
  </td></table>
  <br />
 

### Issue

This PR
- renames the current sourcery stencil "AutoMockable" -> "LegacyAutoMockable"
- adds an updated stenicl
- renames target "WireDomainPkg" to "WireDomainPackage"
- updates the WireDomain package to Swift 6
- prevents a build error with Xcode 16.3

### Testing

Automatic tests should pass.

---

### Checklist

---


WireDomain/Package.swift
```diff
@@ -7,24 +7,44 @@ let package = Package(
...
+        .package(path: "../WireFoundation"),
+        .package(path: "../WireLogging"),
+        .package(path: "../WirePlugins"),
     ],
     targets: [
         .target(
-            name: "WireDomainPkg",
-            path: "./Sources/WireDomain",
-            sources: ["./UseCases/Protocols"]
-        )
```
caldrian, line 20: Two targets shared the same directory tree: "WireDomain" (Xcode project) and "WireDomainPkg" (in Swift package).
Now the the files of "WireDomainPackage" are moved to their own subtree.

WireDomain/Sources/WireDomainPackageSupport/Sourcery/sourcery.yml
```diff
@@ -0,0 +1,8 @@
+sources:
+- ${PACKAGE_ROOT_DIR}/Sources/WireDomainPackage
+templates:
+- ${TARGET_DIR}/Sourcery/AutoMockable.stencil
```
caldrian, line 4: WireDomainPackage is currently the only target using the new stencil.

WirePlugins/Plugins/SourceryPlugin/Stencils/LegacyAutoMockable.stencil
```diff
@@ -0,0 +1,236 @@
+//
```
caldrian, line 1: This file has actually just been renamed.

WirePlugins/Plugins/SourceryPlugin/Stencils/AutoMockable.stencil
```diff
@@ -1,4 +1,3 @@
-
```
caldrian, line 1: This is actually the new file, downloaded from: https://github.com/krzysztofzablocki/Sourcery/blob/master/Templates/Templates/AutoMockable.stencil

I added our header and the public and package imports.

wire-ios/Wire-iOS/Resources/Configuration/Version.xcconfig
```diff
@@ -17,3 +17,4 @@
 //
 
 WIRE_SHORT_VERSION = 3.122.0
+MAJOR_VERSION = 0
```
caldrian: This fixes an error with Xcode 16.3.
Since we define 
```
DYLIB_COMPATIBILITY_VERSION = $(MAJOR_VERSION).0
```
but WireCommonComponents didn't have `MAJOR_VERSION` set, the resulting value for `DYLIB_COMPATIBILITY_VERSION` was ".0" which is not a correct version string (compile error).

caldrian: @netbe please let me know if you have objections to this change.

netbe: so it's 0.0?

caldrian: Exactly.

samwyndham (approved):
Nice work!

WirePlugins/Plugins/SourceryPlugin/Stencils/AutoMockable.stencil
netbe: question: why do we need both files ?

WireDomain/Package.swift
netbe: question: why do we need the WireDomainPackage and Project again? It always confuse me. I wonder if we went to quickly on moving to SPM and have redundancy here

WirePlugins/Plugins/SourceryPlugin/Stencils/AutoMockable.stencil
caldrian: Because applying the new template might result in hundreds of files changed and this way we can incrementally switch to the new template.

netbe (commented):
left a couple questions before approving

WireDomain/Package.swift
caldrian: Because some code in WireDomain depends on WireSyncEngine and the other Xcode projects, while some code of WireDomain is already used by Swift packages (WireUI etc.).

We might have adopted Swift Packages too early, I agree.
But there are also some benefits of swift packages we wouldn't want to miss now, like the `package` access modifier.

netbe (approved):
good to merge

WireDomain/Sources/WireDomain/UseCases/IndividualToTeamMigrationUseCase.swift
```diff
@@ -19,7 +19,7 @@
 import Foundation
 import WireAPI
 import WireDataModel
-import WireDomainPkg
+import WireDomainPackage
```
jullianm, line 22: Nice 👍

WireDomain/Tests/WireDomainPackageTests/PlaceholderTests.swift
```diff
@@ -0,0 +1,31 @@
...
+@testable import WireDomainPackage
+
+struct PlaceholderTests {
+
+    @Test
+    func testNothing() async throws {
+        #expect(true)
+    }
+
+}
```
jullianm, line 31: can't we remove this whole file ?

jullianm (approved):
👍

WireDomain/Tests/WireDomainPackageTests/PlaceholderTests.swift
caldrian: A target usually needs at least one source file.
In another [PR](https://github.com/wireapp/wire-ios/pull/2889) it will be removed soon.

PR #2900, opened 2025-04-15
feat: Broken error handling on message decryption - WPB-17193

KaterinaWire:
### Issue

MLS conversation are not repaired as expected when they go out of sync to the broken state (the epoch is old). User can't receive messages.
We have a way to repair the conversation, but it waits for **DecryptionError.wrongEpoch** during the message decryption.
There’s general issue with error handling when decrypting MLS messages.
```
decryptMessage(message: Message) {
  // The WrongEpoch will fail the transaction and and be re-thrown as a ClientError
  coreCrypto.transaction { context ->
      // Can for example throw WrongEpoch
      context.decryptMessage(message)
  }
}
```
Any error which is not handled inside the transaction will cancel the transaction and new `ClientError` will be thrown instead.
As a result, we don't repair the conversation.

### Solution

The inner error needs to be captured and re-thrown when the transaction fails.

**Note: this solution won't bring back messages that were sent while conversation was broken, and the first message sent will repair the conversation but will be lost!!!**

### Testing

Break the conversation and have an old epoch, try to receive messages.

### Checklist

---

### DIFF
--- a/wire-ios-data-model/Source/MLS/MLSActionExecutor.swift
+++ b/wire-ios-data-model/Source/MLS/MLSActionExecutor.swift
@@ -325,26 +325,32 @@ public actor MLSActionExecutor: MLSActionExecutorProtocol {
     public func decryptMessage(_ message: Data, in groupID: MLSGroupID) async throws -> DecryptedMessage {
         try await performNonReentrant(groupID: groupID) {
             try await coreCrypto.perform {
-                let result: DecryptedMessage? = try await $0.transaction { context in
-
-                    do {
-                        return try await context.decryptMessage(conversationId: groupID.data, payload: message)
-                    } catch let CoreCryptoError.Mls(error) {
-                        switch error {
-                        case .BufferedFutureMessage, .Other(coreCryptoCommitForMissingProposalError):
-                            // ignore error so transaction is saved and message is saved too.
-                            return nil
-                        default:
+                var capturedError: Error?
+                do {
+                    let result: DecryptedMessage? = try await $0.transaction { context in
+                        do {
+                            return try await context.decryptMessage(conversationId: groupID.data, payload: message)
+                        } catch let CoreCryptoError.Mls(error) {
+                            switch error {
+                            case .BufferedFutureMessage, .Other(coreCryptoCommitForMissingProposalError):
+                                // ignore error so transaction is saved and message is saved too.
+                                return nil
+                            default:
+                                capturedError = CoreCryptoError.Mls(error)
+                                throw error
+                            }
+                        } catch {
+                            capturedError = error
                             throw error
                         }
-                    } catch {
-                        throw error
                     }
-                }
-                if let result {
-                    return result
-                } else {
-                    throw Failure.bufferedDecryptedMessage
+                    if let result {
+                        return result
+                    } else {
+                        throw Failure.bufferedDecryptedMessage
+                    }
+                } catch {
+                    throw capturedError ?? error
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


wire-ios-data-model/Source/MLS/MLSActionExecutor.swift
```diff
@@ -325,26 +325,32 @@ public actor MLSActionExecutor: MLSActionExecutorProtocol {
...
+                        do {
+                            return try await context.decryptMessage(conversationId: groupID.data, payload: message)
+                        } catch let CoreCryptoError.Mls(error) {
+                            switch error {
+                            case .BufferedFutureMessage, .Other(coreCryptoCommitForMissingProposalError):
+                                // ignore error so transaction is saved and message is saved too.
+                                return nil
+                            default:
+                                capturedError = CoreCryptoError.Mls(error)
+                                throw error
```
netbe: question: if we throw here then don't we loose what has been done in the transaction?

KaterinaWire: It doesn't matter, because it's inside `transaction`.
If `decryptMessage` throws any error, we'll try to throw captureError on line 353.

netbe: but we rethrow inside the transaction should we instead throw outside not to cancel the transaction

netbe: sorry I got it now. But I think we can remove several throw levels in having the previous throw wrapped again:

```
  public func decryptMessage(_ message: Data, in groupID: MLSGroupID) async throws -> DecryptedMessage {
        try await performNonReentrant(groupID: groupID) {
            try await coreCrypto.perform {
                let result: DecryptedMessage? = try await $0.transaction { context in

                    do {
                        return try await context.decryptMessage(conversationId: groupID.data, payload: message)
                    } catch let CoreCryptoError.Mls(error) {
                        switch error {
                        case .BufferedFutureMessage, .Other(coreCryptoCommitForMissingProposalError):
                            // ignore error so transaction is saved and message is saved too.
                            return nil
                        default:
                            throw CoreCryptoError.Mls(error)
                        }
                    } catch {
                        throw error
                    }
                }
                if let result {
                    return result
                } else {
                    throw Failure.bufferedDecryptedMessage
                }
            }
        }
    }
    ```
    Or did i miss something? @typfel

Comment (johnxnguyen) [reply wire-ios-data-model/Source/MLS/MLSActionExecutor.swift]: question: why is there a do catch inside another do catch? it's quite hard to follow this method, I wonder if it can be simplified.

File: wire-ios-data-model/Source/MLS/MLSService.swift [hunk 2]
```diff
@@ -1697,7 +1697,9 @@ public final class MLSService: MLSServiceInterface {
                 subconversationType: subconversationType
             )
         } catch DecryptionError.wrongEpoch {
-            await fetchAndRepairGroupIfPossible(with: groupID)
+            Task.detached { [self] in
+                await fetchAndRepairGroupIfPossible(with: groupID)
+            }
```
johnxnguyen: suggestion: leave a comment here explaining why waiting on this will cause a deadlock.

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1697,7 +1697,9 @@ public final class MLSService: MLSServiceInterface {
                 subconversationType: subconversationType
             )
         } catch DecryptionError.wrongEpoch {
-            await fetchAndRepairGroupIfPossible(with: groupID)
+            Task.detached { [self] in
+                await fetchAndRepairGroupIfPossible(with: groupID)
```
netbe, line 1705: suggestion: please add a comment why this is detached, 
question: also wonder what caller do on this? apparently just log.
So how does it retry to decrypt it?

netbe (commented):
left some suggestion before approving

wire-ios-data-model/Source/MLS/MLSService.swift
KaterinaWire: added comment

wire-ios-data-model/Source/MLS/MLSService.swift
KaterinaWire: added comment

netbe (approved):
Looks good

jullianm (approved):
LGTM

PR #2902, opened 2025-04-15
fix: Unable to send MLS messages - WPB-16922

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16922" title="WPB-16922" target="_blank">BugWPB-16922</a>  [iOS] Unable to send messages 1:1 or calls
  </td></table>
  <br />
 

### Issue

If the conversation in the broken state (the epoch is old), we receives from the backend:
```
{
    code = 409;
    label = "mls-stale-message";
    message = "The conversation epoch in a message is too old";
}
```
When this error occurs, we should try to repair the conversation and send the message again.

### Testing

Break the conversation and have an old epoch, try to send messages.

### Checklist

---

### DIFF
--- a/wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
+++ b/wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
@@ -438,8 +438,6 @@ public final class MessageSender: MessageSenderInterface {
             retryCount += 1
 
             try await operation()
-        } catch {
-            throw error
         }
     }
 
--- a/wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
+++ b/wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
@@ -413,7 +413,8 @@ public final class MessageSender: MessageSenderInterface {
                 try await handleMLSStaleMessageError(
                     groupID: groupID,
                     mlsService: mlsService,
-                    operation: operation)
+                    operation: operation
+                )
             default:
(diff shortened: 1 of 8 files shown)


netbe (approved):
Looks good, would be good to add a test to cover this

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
```diff
@@ -376,29 +376,70 @@ public final class MessageSender: MessageSenderInterface {
...
-        await context.perform {
-            // handle 201 case failed_to_send
-            // https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/556564601/Use+case+sending+a+message+MLS
-            self.mlsPayloadProcessor.updateFailedRecipients(from: payload, for: message)
-            message.delivered(with: response)
+            retryCount += 1
+
+            try await operation()
+        } catch {
+            throw error
```
johnxnguyen: suggestion: no need to catch if all it does is throws the error... if you you commit these two lines it will have the same effect.

johnxnguyen (approved):

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
KaterinaWire: done

typfel (approved):

PR #2903, opened 2025-04-15
fix: pulling read receipt mode when disabled - WPB-16352

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16352" title="WPB-16352" target="_blank">BugWPB-16352</a>  [iOS] Failed to pull self user settings
  </td></table>
  <br />
 

### Issue

New sync currently can fails to pull self user settings when the send read receipts option is disabled. This is because in this case the service returns a 404 rather than a `0` (e.g. false) response.

This PR updates `UserPropertiesAPIV0.getProperty(forKey:)` to catch `UserPropertiesAPIError.propertyNotFound`  errors and return `false` for both fetching `UserPropertiesAPIV0.areTypingIndicatorsEnabled` & `UserPropertiesAPIV0.areReadReceiptsEnabled`.

To do this this PR also updates the behavior of `ResponseParser.failure(code:label:error:)` which was throwing a JSON decoding error due to unexpected body in the 404 case instead of the expected `UserPropertiesAPIError.propertyNotFound` error. The new behavior:

- if no label is given it only cares about matching the status code
- if a label is given it requires that both the status code and label match. If this doesn't happen it doesn't match but also does not throw.

### Testing

_Describe how to test._

1. Fresh install build with new initial sync enabled
2. On web, ensure your account has "Read receipts" setting off
3. Log in
4. Log in should complete
5. Repeat with "Typing indicator" off 

### Checklist

### DIFF
--- a/WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
+++ b/WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
@@ -48,8 +48,8 @@ class TeamsAPIV0: TeamsAPI, VersionedAPI {
 
         return try ResponseParser()
             .success(code: .ok, type: TeamResponseV0.self)
-            .failure(code: .notFound, error: TeamsAPIError.invalidTeamID)
             .failure(code: .notFound, label: "no-team", error: TeamsAPIError.teamNotFound)
+            .failure(code: .notFound, error: TeamsAPIError.invalidTeamID)
             .parse(code: response.statusCode, data: data)
     }
 
@@ -119,8 +119,8 @@ class TeamsAPIV0: TeamsAPI, VersionedAPI {
 
         return try ResponseParser()
             .success(code: .ok, type: TeamMemberLegalholdResponseV0.self)
-            .failure(code: .notFound, error: TeamsAPIError.invalidRequest)
             .failure(code: .notFound, label: "no-team-member", error: TeamsAPIError.teamMemberNotFound)
+            .failure(code: .notFound, error: TeamsAPIError.invalidRequest)
             .parse(code: response.statusCode, data: data)
     }
 
--- a/WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
+++ b/WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
@@ -48,8 +48,8 @@ class TeamsAPIV0: TeamsAPI, VersionedAPI {
 
         return try ResponseParser()
             .success(code: .ok, type: TeamResponseV0.self)
-            .failure(code: .notFound, label: "no-team", error: TeamsAPIError.teamNotFound)
             .failure(code: .notFound, error: TeamsAPIError.invalidTeamID)
+            .failure(code: .notFound, label: "no-team", error: TeamsAPIError.teamNotFound)
             .parse(code: response.statusCode, data: data)
     }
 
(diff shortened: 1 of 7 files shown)


johnxnguyen (commented):
For me this is a good change, but I think it would also be good (for debugging purposes) to know what the response is in case of no match. Perhaps we can log the response and additionally try to decode the data to a string (of the content is json).

jullianm:
@samwyndham changes looks good 👍

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
```diff
@@ -48,8 +48,8 @@ class TeamsAPIV0: TeamsAPI, VersionedAPI {
 
         return try ResponseParser()
             .success(code: .ok, type: TeamResponseV0.self)
-            .failure(code: .notFound, error: TeamsAPIError.invalidTeamID)
             .failure(code: .notFound, label: "no-team", error: TeamsAPIError.teamNotFound)
+            .failure(code: .notFound, error: TeamsAPIError.invalidTeamID)
```
samwyndham: Note that with the behavior change it is important that when there are multiple matchers for a given status code, the labelled versions appear first. Do you think it is worth changing the internal workings of `ResponseParser` to improve this?

WireAPI/Tests/WireAPITests/APIs/ConversationsAPI/ConversationsAPITests.swift
```diff
@@ -295,11 +294,10 @@ final class ConversationsAPITests: XCTestCase {
         // then
         do {
             _ = try await api.getConversations(for: [])
-        } catch let error as FailureResponse {
-            XCTAssertEqual(error.code, 400)
-            XCTAssertEqual(error.label, "invalid body")
```
samwyndham, line 300: Note that the error type has changed which seems more correct now.

WireAPI/Tests/WireAPITests/APIs/UserPropertiesAPI/UserPropertiesAPITests.swift
```diff
@@ -91,7 +91,7 @@ final class UserPropertiesAPITests: XCTestCase {
         // Then
         try await apiSnapshotHelper.verifyRequest(for: [.v0], apiService: apiService) { sut in
             // When
-            let result = try await sut.areReadReceiptsEnabled
```
samwyndham, line 94: This was incorrect and should have been testing `areTypingIndicatorsEnabled`. See also the snapshot data update

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
jullianm: I would assume it's worth it because when implementing a new endpoint or adjusting an existing one we might forget this behavior and this can lead to confusion. 

Maybe we could add multiple parseBlocks in the `ResponseParser`:
`private var successParseBlocks: [ParseBlock]` 
`private var labeledFailureParseBlocks: [ParseBlock]` 
`private var unlabeledFailureParseBlocks: [ParseBlock]` 

and when executing the parsing with `func parse(code: Int, data: Data?)` we start by the successParseBlocks, then the `labeledFailureParseBlocks` one and finally the unlabeled ones.

So we don't have to worry of the order when consuming that parser.
But that looks a bit overkill, there's probably a simpler way to do it.

johnxnguyen (approved):

jullianm (approved):

WireAPI/Sources/WireAPI/APIs/TeamsAPI/TeamsAPIV0.swift
samwyndham: I've done something to _improve_ things in https://github.com/wireapp/wire-ios/pull/2903/commits/f3f3aa0630d2c9d6f6c2d680214e503b95199ba9

PR #2918, opened 2025-04-18
feat: adapt Websocket for consumable notifications - WPB-17225

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17225" title="WPB-17225" target="_blank">TaskWPB-17225</a>  [iOS] Adapt webSocket
  </td></table>
  <br />
 

### Issue

This covers the sync part while the app is in foreground with consumable notifications.

* Creates a `IncrementalSyncV2` depending on the consumable-notifications capabilities which will be used under a `LiveSyncProtocol`. It basically create and open a webSocket where we listen to events and process them.
* Adapts Websocket to send data to server
* Creates new `PushChannelV2` to acknowledge events

**Note** ~also that to hide the syncbar we assume that after X seconds without receiving new events we're caught up~.
**EDIT**: the backend will put a new notification type in the stream `message_count`, this will allow us to know what we're uptoDate (equivalent of IncrementalSync, then we're livesyncing).
 The `IncrementalSyncV2` will then inform the `SyncAgent` via the `LiveSyncDelegate`

Other:
* Disabled extra tags in console logs

### Testing

Use `lich` environment where notifications.missed event occur every 5minutes meaning the client was offline for too long and needs to perform slow sync


WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Responses/UpdateEventEnvelopeV0.swift
```diff
@@ -30,7 +30,8 @@ public struct UpdateEventEnvelopeV0: Decodable, ToAPIModelConvertible {
         UpdateEventEnvelope(
             id: id,
             events: (payload ?? []).map(\.updateEvent),
-            isTransient: transient ?? false
+            isTransient: transient ?? false,
+            deliveryTag: nil
```
netbe, line 34: for old enveloppes there is no acknowledgement so no deliveryTag needed

WireAPI/Sources/WireAPI/Assembly.swift
```diff
@@ -60,22 +60,6 @@ public final class Assembly {
         return service
     }()
 
-    private lazy var pushChannelService: some PushChannelServiceProtocol = PushChannelService(
```
netbe, line 63: unused code

WireAPI/Tests/WireAPITests/Network/PushChannel/NewPushChannelTests.swift
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        // Then envelopes are received
+        try XCTAssertCount(receivedEnvelopes, count: 3)
+        XCTAssertEqual(receivedEnvelopes[0], .event(Scaffolding.envelope1))
+        XCTAssertEqual(receivedEnvelopes[1], .event(Scaffolding.envelope2))
+        XCTAssertEqual(receivedEnvelopes[2], .event(Scaffolding.envelope3))
+    }
+
+    func testOpen_MissedNotificationsEvent() async throws {
```
netbe: this is new compared to other PushChannel.

WireAPI/Tests/WireAPITests/Network/PushChannel/NewPushChannelTests.swift
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    override func tearDown() async throws {
+        request = nil
+        webSocket = nil
+        sut = nil
+        try await super.tearDown()
+    }
+
+    func testOpen_DecodeEventSuccessfully() async throws {
```
netbe: same test but adapted than other PushChannel.

WireDomain/Sources/WireDomain/Synchronization/NewIncrementalSync.swift
```diff
@@ -0,0 +1,214 @@
...
+        let task: Task<Void, Error> = Task { @Sendable [
+            logger,
+            decryptor,
+            store,
+            processor,
+            databaseSaver,
+            pushChannel,
+            delegate,
+            syncStateSubject
+        ] in
```
netbe: some Sendable warnings here, not sure how to resolve them, will cleanup in a following PR

WireLogging/Sources/WireLogging/SystemLogger.swift
```diff
@@ -90,12 +90,12 @@ public class SystemLogger: LoggerProtocol {
         }
 
         var finalMessage = "\(message.logDescription)\(attributesDescription(from: mergedAttributes))"
-
-        if !tags.isEmpty {
-            let extraInfo = tags.map { key, value in "[\(key.rawValue):\(value)]" }.joined()
-            finalMessage += extraInfo
-        }
-
+        #if !DEBUG
```
netbe, line 93: skipping tags to show in console locally

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -205,8 +197,15 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
                 try await incrementalSyncTaskManager.performIfNeeded { [weak self] in
                     guard let self else { return }
                     delegate?.syncAgentDidStartIncrementalSync(self)
-                    incrementalSyncToken = try await incrementalSyncProvider.provideIncrementalSync().perform()
-                    delegate?.syncAgentDidFinishIncrementalSync(self, isRecovering: false)
+
+                    if liveSync {
+                        incrementalSyncToken = try await incrementalSyncProvider.provideLiveSync(delegate: self)
+                            .perform(acknowledgeFullSync: shouldAcknowledgeFullSync)
```
netbe: in case of liveSync the incrementalSync will call the SyncAgent via LiveSyncDelegate when we're up to date and therefore hide the sync bar

**thought:** I guess at some point we could rely on the publisher of sync state instead of the delegates to interact with the sync bar

wire-ios-sync-engine/Source/Synchronization/ZMOperationLoop.m
```diff
@@ -89,7 +89,11 @@ - (instancetype)initWithTransportSession:(id<TransportSessionType>)transportSess
         // this is needed to avoid loading from syncMOC on the main queue
         [moc performGroupedBlock:^{
             [self.transportSession configurePushChannelWithConsumer:self groupQueue:moc];
-            [self.transportSession.pushChannel setKeepOpen:operationStatus.operationState == SyncEngineOperationStateForeground];
+            if (isSyncV2Enabled) {
```
netbe, line 92: avoid unecessary on/off legacy websocket

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -567,21 +568,18 @@ public final class ZMUserSession: NSObject {
...
 
-        coreCryptoProvider.registerMlsTransport(clientSessionComponent.mlsTransport)
-
-        let incrementalSyncProvider: IncrementalSyncProvider = if !asyncStreamEnabled {
-            clientSessionComponent
-        } else {
-            // TODO: [WPB-17225] replace syncProvider here
-            clientSessionComponent
+        if asyncStreamEnabled {
+            journal[.skipPullingLastNotificationID] = true
```
netbe: in asyncSteam, clients cannot get /notifications/last

WireAPI/Sources/WireAPI/APIs/PushChannelAPI/NewPushChannelAPI.swift
```diff
@@ -0,0 +1,33 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+// sourcery: AutoMockable
+/// An API access object for endpoints concerning the push channel.
+public protocol NewPushChannelAPI {
```
johnxnguyen: question: why a new api and not use the existing one?

WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Responses/UpdateEventEnvelopeV8.swift
```diff
@@ -0,0 +1,29 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+struct UpdateEventEnvelopeV8: Decodable {
+    enum CodingKeys: String, CodingKey {
+        case id
+        case payload
+    }
```
johnxnguyen: suggestion: the coding keys aren't necessary, perhaps we could delete them.

WireAPI/Sources/WireAPI/Models/UpdateEvent/UpdateEventEnvelope.swift
```diff
@@ -37,6 +37,8 @@ public struct UpdateEventEnvelope: Equatable, Sendable {
 
     public let isTransient: Bool
 
+    public let deliveryTag: UInt64?
```
johnxnguyen, line 42: suggestion: add a doc comment.

WireAPI/Sources/WireAPI/Network/PushChannel/AcknowledgmentType.swift
```diff
@@ -0,0 +1,23 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+import Foundation
+
+enum AcknowledgmentType: String, Encodable {
+    case fullSync = "ack_full_sync"
+    case ack
```
johnxnguyen: suggestion: `singleEvent`

WireAPI/Sources/WireAPI/Network/PushChannel/EventAcknowledgmentNotification.swift
```diff
@@ -0,0 +1,39 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+import Foundation
+
+/// Payload sent via NewPushChannel for acknowledgment of event in async stream
+struct EventAcknowledgmentNotification: Encodable {
```
johnxnguyen: suggestion: `SingleEventAcknowledgment`

WireAPI/Sources/WireAPI/Network/PushChannel/FullSyncAcknowledgmentNotification.swift
```diff
@@ -0,0 +1,23 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+import Foundation
+
+/// Payload of acknowledgment of initial sync for async stream
+struct FullSyncAcknowledgmentNotification: Encodable {
```
johnxnguyen: suggestion: `FullSyncAcknowledgment`

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
```diff
@@ -0,0 +1,253 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireFoundation
+import WireLogging
+
+/// PushChannel using new async stream
+public final class NewPushChannel: NewPushChannelProtocol {
```
johnxnguyen: question: do you intend to update the name later to drop `New`?

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
```diff
@@ -0,0 +1,253 @@
...
+    private var timeoutTask: Task<Void, any Error>?
+
+    private let channelState = ChannelState()
+
+    private var keepAliveTask: Task<Void, any Error>?
+    private let keepAliveInterval: TimeInterval
+
+    private var (stream, continuation) = AsyncThrowingStream<Element, any Error>.makeStream()
+
+    /// Initial PushChannel with Async Stream capabitilites
```
johnxnguyen: question: did you mean `Initialize`?

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
```diff
@@ -0,0 +1,253 @@
...
+    /// - Parameters:
+    ///   - webSocket: webSocket to use
+    ///   - keepAliveInterval: interval for sending ping and keep webSocket open
+    ///   - timeout: interval after we consider we're up to date with events
+    public init(
+        webSocket: any WebSocketProtocol,
+        keepAliveInterval: TimeInterval = 5,
```
johnxnguyen: question: 5 seconds is very frequent compared to the current 30 second interval. Is there a reason we need to ping so much?

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
```diff
@@ -0,0 +1,253 @@
...
+        timeout: TimeInterval = 1
```
johnxnguyen: suggestion: `timeout` could be misconstrued to mean the push channel will closed after the timeout is reached. I suggest to call it `upToDateThreshold`, or even `silenceInterval` to suggest it's the number of seconds of inactivity to trigger the `upToDate` element.

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
```diff
@@ -0,0 +1,253 @@
...
+                return
+            }
+            continuation.finish()
+        }
+
+        // The server will drop the connection (possibly silently)
+        // if the client doesn’t send a ping message every so often.
+        setUpKeepAliveTask()
+
+        return stream
```
johnxnguyen: question: where is `stream` declared? I also can't find the `continuation` in the task... is some code missing?

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
```diff
@@ -0,0 +1,253 @@
...
+
+        await webSocket.close()
+        tearDownTimeoutTask()
+        tearDownKeepAliveTask()
+    }
+
+    func receiveMessage(_ message: URLSessionWebSocketTask.Message) async throws -> Element {
+        timeoutTask?.cancel()
+        await channelState.receivedMessage()
+        await channelState.startProcessing()
```
johnxnguyen: suggestion: for transparency it may be nice to pull these up a level just before this method is invoked.

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
```diff
@@ -0,0 +1,253 @@
...
+    private func tearDownTimeoutTask() {
+        guard let timeoutTask else { return }
+        WireLogger.pushChannel.debug("tearing down timeoutTask")
+        timeoutTask.cancel()
+        self.timeoutTask = nil
+    }
+
+    // MARK: - Acknowledgement
+
+    public func ackEvent(deliveryTag: UInt64, multiple: Bool = false) async throws {
```
johnxnguyen: suggestion: `acknowledgeEvent`

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
```diff
@@ -0,0 +1,253 @@
...
+        WireLogger.pushChannel.debug("ackEvent \(deliveryTag)", attributes: .pushChannelV3)
+        let acknowledgement = EventAcknowledgmentNotification(
+            deliveryTag: deliveryTag,
+            multiple: multiple
+        )
+        let data = try JSONEncoder().encode(acknowledgement)
```
johnxnguyen: suggestion: perhaps it's worth initializing the encoder once and reusing it.

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
```diff
@@ -0,0 +1,253 @@
...
+        try await write(data: data)
+    }
+
+    public func ackFullSync() async throws {
```
johnxnguyen: suggestion: `acknowledgeFullSync`

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
```diff
@@ -0,0 +1,253 @@
...
+
+    func timeSinceLastMessage() -> TimeInterval {
+        Date().timeIntervalSince(lastMessageUpdate)
+    }
+
+    func caughtUp() {
+        catchingUp = false
+    }
+
+    func wait(timeout: TimeInterval) -> Bool {
```
johnxnguyen: suggestion: perhaps we should inject the `silenceInterval` into `ChanelState`, then rename this method: `didExceedSilenceInterval`.

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
```diff
@@ -0,0 +1,253 @@
...
+        timeSinceLastMessage() > timeout && !isProcessing
```
johnxnguyen: question: what is the significance of of `!isProcessing`?

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
```diff
@@ -0,0 +1,253 @@
...
+                    try await Task.sleep(nanoseconds: 100_000_000)
+                    if await channelState.wait(timeout: timeout) {
+                        if await channelState.catchingUp {
+                            WireLogger.pushChannel.debug("caught up")
+                            await channelState.caughtUp()
+                            continuation.yield(.upToDate)
+                            break
+                        }
+                    }
+                }
```
johnxnguyen: question: I just want to confirm my understanding is correct:

- Each time after we have received a message in the stream we create a new task
- The task will poll every second
- For each poll, it checks if the time since the that last event was received exceeds the timeout
- If the threshold has been reached, and in the meantime no new message has started to be processed and we haven't already send the `upToDate` element, then we send the `upToDateElement`.

My questions are:
- Is the `if await channelState.catchingUp` check necessary because there may be another task created and we don't want to risk sending the `upToDate` message a second time from an different task?
- Why do we create a new task for each message and not have a single task running?
- Why do we poll every second, instead of something like:

```swift
do {
  try await Task.sleep(nanoseconds: timeout.nanoseconds)
  // If we get this far without being cancelled then the timeout has been reached.
  // ...
} catch {
  // ..
}
```

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelService.swift
```diff
@@ -21,13 +21,18 @@ import Foundation
...
     /// - Parameter request: A request for a web socket connection.
     /// - Returns: A push channel.
 
     func createPushChannel(_ request: URLRequest) async throws -> any PushChannelProtocol
 
+    /// Create a new push channel (v2). Async notifications
+    ///
+    /// - Parameter request: A request for a web socket connection.
+    /// - Returns: A push channel.
+    func createNewPushChannel(_ request: URLRequest) async throws -> any NewPushChannelProtocol
```
johnxnguyen: question: do we intend to rename this once the old push channel is removed? Maybe in the meantime we can refer to it as v2 also in the method and type names.

WireAPI/Sources/WireAPI/Network/PushChannel/WebSocket.swift
```diff
@@ -98,4 +99,14 @@ public actor WebSocket: WebSocketProtocol {
         }
     }
 
+    public func write(data: Data) async throws {
+        connection.resume()
```
johnxnguyen: question: do we resume to handle the case we try to write data before the web socket is open? Wouldn't that be a mistake?

WireAPI/Sources/WireAPI/Network/PushChannel/WebSocketNotification.swift
```diff
@@ -0,0 +1,79 @@
...
+            return UpdateEventEnvelope(
+                id: UUID(),
+                events: [],
+                isTransient: false
+            )
+        }
+        return UpdateEventEnvelope(
+            id: event.id,
+            events: event.payload.map(\.updateEvent),
+            isTransient: false,
```
johnxnguyen: question: is `isTransient` irrelevant now?

WireAPI/Sources/WireAPI/Network/PushChannel/WebSocketNotification.swift
```diff
@@ -0,0 +1,79 @@
...
+        self.type = type
+        self.data = data
+    }
+}
+
+extension WebSocketNotification: ToAPIModelConvertible {
+
+    func toAPIModel() -> UpdateEventEnvelope {
+        guard let event = data?.event  else {
+            assertionFailure("don't call toAPIModel() when type is `notificationsMissed`")
```
johnxnguyen: suggestion: rather than open up this invalid code path, you could move this code to the callsite, or add this extension to the NotificationData type.

WireDomain/Sources/WireDomain/Synchronization/Protocols/IncrementalSyncProtocol.swift
```diff
@@ -24,9 +24,10 @@ import Foundation
 public protocol IncrementalSyncProtocol {
 
     /// Perform the incremental sync.
-    ///
+    /// - Parameter acknowledgeFullSync: In case of New incremental sync, this will inform backend a full sync has been
+    /// done
     /// - Returns: A token to retain to keep the push channel open.
 
-    func perform() async throws -> IncrementalSync.Token
+    func perform(acknowledgeFullSync: Bool) async throws -> IncrementalSync.Token
```
johnxnguyen: question: I noticed that the `NewIncrementalSync` conforms to `LiveSyncProtocol`, so it would seem that these changes are not necessary, right?

WireDomain/Sources/WireDomain/Synchronization/NewIncrementalSync.swift
```diff
@@ -0,0 +1,240 @@
...
+        self.syncStateSubject = syncStateSubject
+    }
+
+    public func perform(acknowledgeFullSync: Bool) async throws -> IncrementalSync.Token {
+        logger.debug("performing live sync v3")
+        let pushChannel = try await pushChannelAPI.createPushChannel(clientID: selfClientID)
+
+        if acknowledgeFullSync {
+            try await pushChannel.ackFullSync()
+        }
```
johnxnguyen: question: I would expect to open the push channel, then send the acknowledgment. If the open the push channel and an acknowledgment is needed, doesn't that mean that there would be no events received until the acknowledgment is sent?

WireDomain/Sources/WireDomain/Synchronization/NewIncrementalSync.swift
```diff
@@ -0,0 +1,240 @@
...
+        syncStateSubject.send(.liveSyncing)
+
+        do {
+            for try await element in liveEventStream {
+                logger.debug("received live event envelope v3")
+                switch element {
+                case .upToDate:
+                    logger.debug("upToDate event v3")
+                    syncStateSubject.send(.idle)
+                    delegate?.didFinishSync(sync: self)
```
johnxnguyen: suggestion: I think `didFinishSync` can be misconstrued as the whole sync finishing. Perhaps we should call this `didCatchUp` or `isUpToDate`.

WireDomain/Sources/WireDomain/Synchronization/NewIncrementalSync.swift
```diff
@@ -0,0 +1,240 @@
...
+
+                case .missedEvents:
+                    logger.debug("missedEvents event v3")
+                    await delegate?.didMissedEvents(sync: self)
+                    // TODO: [WPB-17609] insert potential gap message here with messageLocalStore
+                    try await pushChannel.ackFullSync()
```
johnxnguyen: question: I'm confused, why do we acknowledge the full sync before opening the push channel and then also here? I would expect that we must tear down the push channel, perform the initial sync, and then open the push channel again. In that case, we should only report that we're missing events.

WireDomain/Sources/WireDomain/Synchronization/NewIncrementalSync.swift
```diff
@@ -0,0 +1,240 @@
...
+
+                case let .event(envelope):
+                    do {
+                        var envelope = envelope
+                        envelope.events = try await decryptEnvelope(envelope)
+
+                        let index = try await storeEnvelope(envelope)
+
+                        // Bump the last event id so we don't refetch it.
+                        // there's no events marked as transcient anymore
```
johnxnguyen:
```suggestion
                        // there's no events marked as transient anymore
```

WireDomain/Sources/WireDomain/Synchronization/NewIncrementalSync.swift
```diff
@@ -0,0 +1,240 @@
...
+                        store.storeLastEventID(id: envelope.id)
```
johnxnguyen: question: what purpose does the last event id have now? I thought that this is rendered obsolete now that we acknowledge events.

WireDomain/Sources/WireDomain/Synchronization/NewIncrementalSync.swift
```diff
@@ -0,0 +1,240 @@
...
+                logger.debug(
+                    "ack event envelope v3",
+                    attributes: [.eventEnvelopeID: envelope.id]
+                )
+                try await pushChannel.ackEvent(deliveryTag: deliveryTag, multiple: false)
+            }
+        } catch {
+            logger.error(
+                "failed to ack live event envelope v3: \(String(describing: error))",
+                attributes: [.eventEnvelopeID: envelope.id]
```
johnxnguyen: question: seeing as we catch all errors one level higher and just move to the next event anyway, shouldn't we throw here too so that it's clear that this is a fallible method?

WireDomain/Sources/WireDomain/Synchronization/NewIncrementalSync.swift
```diff
@@ -0,0 +1,240 @@
...
+                logger.debug(
+                    "processing live event: \(event.name)",
+                    attributes: [.eventEnvelopeID: envelope.id]
+                )
+                try await processor.processEvent(event)
+            } catch {
+                logger.error(
+                    "failed to process live event: \(String(describing: error))",
+                    attributes: [.eventEnvelopeID: envelope.id]
+                )
```
johnxnguyen: question: seeing as we catch all errors one level higher and just move to the next event anyway, shouldn't we throw here too so that it's clear that this is a fallible method?

WireDomain/Sources/WireDomain/Synchronization/NewIncrementalSync.swift
```diff
@@ -0,0 +1,240 @@
...
+            logger.debug(
+                "deleting live event envelope",
+                attributes: [.eventEnvelopeID: envelope.id]
+            )
+            try await store.deleteEventEnvelope(atIndex: index)
+        } catch {
+            logger.error(
+                "failed to delete live event envelope v3: \(String(describing: error))",
+                attributes: [.eventEnvelopeID: envelope.id]
+            )
```
johnxnguyen: question: seeing as we catch all errors one level higher and just move to the next event anyway, shouldn't we throw here too so that it's clear that this is a fallible method?

WireDomain/Sources/WireDomain/Synchronization/NewIncrementalSync.swift
```diff
@@ -0,0 +1,240 @@
...
+        }
+    }
+
+    private func save() async {
+        do {
+            // Save.
+            try await databaseSaver.save()
+        } catch {
+            logger.error("failed to save database v3: \(String(describing: error))")
```
johnxnguyen: question: seeing as we catch all errors one level higher and just move to the next event anyway, shouldn't we throw here too so that it's clear that this is a fallible method?

WireLogging/Sources/WireLogging/SystemLogger.swift
johnxnguyen: thank you!

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
johnxnguyen: Yes my intention was that the sync state would be used as the data source for the sync bar.

WireDomain/Sources/WireDomain/Utilities/Journal/JournalKey.swift
```diff
@@ -53,6 +53,13 @@ public extension JournalKey where Value == Bool {
         defaultValue: false
     )
 
+    /// Whether client should skip calling /notifications/last
+
+    static let skipPullingLastNotificationID = Self(
+        "skipPullingLastNotificationID",
+        defaultValue: false
+    )
```
johnxnguyen: question: in the sync agent this is used to determine if we should use the new incremental sync. Perhaps we should call this `isSyncV3Enabled` or something similar.

WireAPI/Sources/WireAPI/APIs/PushChannelAPI/NewPushChannelAPI.swift
netbe: because it does not return the same protocol: NewPushChannelProtocol

WireAPI/Sources/WireAPI/Network/PushChannel/EventAcknowledgmentNotification.swift
netbe: hmm this could be used to ack multiple events if we wish to:
for now we do:
1
ack
2
ack
3
ack
4
ack

but for example you receive :
1
2
3
4
ack 4, will inform backend that you receive 1 to 4 with the multiple attribute

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: no, maybe a better naming :), so I guess we have to live with the two systems for a while, since this one will require apiV8, so on prem users might not run this at first.
Note: we can allow PushChannel to be used for apiVersion 0 and later on and remove the old legacy code path

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: yep

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: i thought we pinged every 5 seconds, I'll use the same interval as we use right now

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: upToDateThreshold sounds good

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
```diff
@@ -0,0 +1,253 @@
...
+
+    private let timeout: TimeInterval
+    private var timeoutTask: Task<Void, any Error>?
+
+    private let channelState = ChannelState()
+
+    private var keepAliveTask: Task<Void, any Error>?
+    private let keepAliveInterval: TimeInterval
+
+    private var (stream, continuation) = AsyncThrowingStream<Element, any Error>.makeStream()
```
netbe: @johnxnguyen here is the stream created with its continuation

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: https://github.com/wireapp/wire-ios/pull/2918/files/7e68369d77339a738b9cc4329d3ba81968b19770#r2123437093

WireAPI/Sources/WireAPI/Network/PushChannel/WebSocket.swift
netbe: yes that was my intention, you can't write on websocket if it's not opened. At first this was a guarantee since i closed the websocket when having the full_sync event. After testing, it was better to acknowledge the event during the websocket is opened in order not to receive the event twice.

So I guess we can remove thre resume as it's not necessary

WireAPI/Sources/WireAPI/Network/PushChannel/WebSocketNotification.swift
netbe: exactly, backend does not give this info.

Note that typing events are still transient: "You won't receive them unless you're connected to the websocket when they're sent".

WireDomain/Sources/WireDomain/Synchronization/Protocols/IncrementalSyncProtocol.swift
netbe: good catch, must be a leftover

WireDomain/Sources/WireDomain/Utilities/Journal/JournalKey.swift
netbe: yes we could simplify the logic with `isSyncV3Enabled` as to migrate to this new system I would need this flag.

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: done

WireAPI/Sources/WireAPI/Network/PushChannel/AcknowledgmentType.swift
netbe: it could be multiple event, would `event` make more sense?

WireAPI/Sources/WireAPI/Network/PushChannel/FullSyncAcknowledgmentNotification.swift
netbe: done

WireAPI/Sources/WireAPI/Network/PushChannel/EventAcknowledgmentNotification.swift
netbe: dropped the Notification suffix

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: done

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: renamed

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: done

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: done

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: done

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: processing the event, referring to the schema here: https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/1741258776/Incremental+synchronization#Mobile-client-considerations-regarding-WebSocket

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: your understanding is correct. maybe better to get on a paring session for this

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelService.swift
netbe: yes I hesitate with the name. Should we say v2 or v3 if we consider legacy code as v1?
PushChannel stays as v1
NewPushChannel becomes PushChannelV2

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: simplified, the version as suggested

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: Renamed to PushChannelV2

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: this is gone, we don't need this channelState nor the task. The backend will send a new notification type giving the number of remaining messages in the queue to retrieve before being **alive**

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: not relevant anymore

WireAPI/Sources/WireAPI/Network/PushChannel/NewPushChannel.swift
netbe: this is gone, we don't this. The backend will send a new notification type giving the number of remaining messages in the queue to retrieve before being **alive**

WireAPI/Sources/WireAPI/Network/PushChannel/WebSocketNotification.swift
netbe: updated

WireDomain/Sources/WireDomain/Synchronization/NewIncrementalSync.swift
netbe: exactly, nothing is sent until it is acknowledged

WireDomain/Sources/WireDomain/Synchronization/NewIncrementalSync.swift
netbe: renamed to isUpToDate

WireDomain/Sources/WireDomain/Synchronization/NewIncrementalSync.swift
netbe: ah sorry, previously I was closing the pushchannel and doing the initialsync and acknowledge before opening again the pushchannel. This was not the best as we were getting again a notifications.missed when opening the pushchannel.

So i switch to this: keep the websocket open, do the initial sync acknowledge

I cleaned up the other ack

WireDomain/Sources/WireDomain/Synchronization/NewIncrementalSync.swift
netbe: good point, I removed this

WireAPI/Sources/WireAPI/APIs/PushChannelAPI/PushChannelV2API.swift
```diff
@@ -30,4 +30,4 @@ public protocol NewPushChannelAPI {
 }
 
 // Workaround for automockable compiler error.
-public typealias AnyNewPushChannel = any NewPushChannelProtocol
+public typealias AnyNewPushChannel = any PushChannelV2Protocol
```
johnxnguyen: nitpick: `AnyPushChannelV2`

WireAPI/Sources/WireAPI/Network/PushChannel/AcknowledgmentType.swift
johnxnguyen: yes

WireAPI/Sources/WireAPI/Network/PushChannel/MessageCountAcknowledgment.swift
```diff
@@ -0,0 +1,23 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+import Foundation
+
+/// Payload of acknowledgment of messageCount for async stream
+struct MessageCountAcknowledgment: Encodable {
```
johnxnguyen, line 21: question: what's the acknowledgement for?

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelService.swift
```diff
@@ -32,7 +32,7 @@ public protocol PushChannelServiceProtocol {
     ///
     /// - Parameter request: A request for a web socket connection.
     /// - Returns: A push channel.
-    func createNewPushChannel(_ request: URLRequest) async throws -> any NewPushChannelProtocol
+    func createNewPushChannel(_ request: URLRequest) async throws -> any PushChannelV2Protocol
```
johnxnguyen: nitpick: `createPushChannelV2`

WireAPI/Sources/WireAPI/Network/PushChannel/WebSocketNotification.swift
```diff
@@ -36,31 +46,70 @@ struct WebSocketNotification: Decodable {
...
+            UpdateEventEnvelope(
+                id: eventData.event.id,
+                events: eventData.event.payload.map(\.updateEvent),
+                isTransient: false,
+                deliveryTag: eventData.deliveryTag
             )
+        case let .messageCount(data):
+            fatalError()
+        case .none:
+            fatalError()
```
johnxnguyen: question: shall we find a way to avoid this? Why does the `WebSocketNotification` need to conform to `ToAPIModelConvertible`? Can the event data conform?

WireAPI/Sources/WireAPI/Network/PushChannel/WebSocketNotification.swift
```diff
@@ -22,10 +22,20 @@ struct WebSocketNotification: Decodable {
...
 
-    struct NotificationData: Decodable {
+    enum DataType: Decodable {
+        case event(EventNotificationData)
+        case messageCount(MessageCountData)
+    }
+
+    struct MessageCountData: Decodable {
+        var count: Int
+    }
```
johnxnguyen, line 36: question: so backend sends a count of pending events now?

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelV2.swift
```diff
@@ -104,34 +90,41 @@ public final class NewPushChannel: NewPushChannelProtocol {
...
-                "received web socket data, decoding..., \(String(data: data, encoding: .utf8))",
-                attributes: .pushChannelV3
-            )
+            WireLogger.pushChannel.debug("received web socket data, decoding...", attributes: .pushChannelV2)
             let envelope = try decoder.decode(WebSocketNotification.self, from: data)
-            if envelope.type == .event {
+
+            switch envelope.type {
+            case .event:
+                elementsToSync += 1
```
johnxnguyen: question: should this be called `numberOfReceivedEvents`?

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelV2.swift
```diff
@@ -104,34 +90,41 @@ public final class NewPushChannel: NewPushChannelProtocol {
...
                 return Element.event(envelope.toAPIModel())
-            } else {
+            case .messagesCount:
+                WireLogger.pushChannel.info(
+                    "\(envelope.messageCount) until we're up to date",
+                    attributes: .pushChannelV2
+                )
+                elementsUntilUpToDate = envelope.messageCount
```
johnxnguyen: suggestion: `numberOfEventsUntilUpToDate` or `pendingEventCount` or `remainingEventCount`

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelV2.swift
```diff
@@ -144,57 +137,27 @@ public final class NewPushChannel: NewPushChannelProtocol {
...
-                        continuation.yield(.upToDate)
-                        break
-                    }
-                }
-            } catch {
-                WireLogger.pushChannel.warn("upToDateTask was cancelled")
-                tearDownUpToDateTask()
-            }
-        }
-    }
```
johnxnguyen: I'm glad we could avoid all these tasks

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelV2.swift
```diff
@@ -204,63 +167,23 @@ public final class NewPushChannel: NewPushChannelProtocol {
...
-        WireLogger.pushChannel.debug("acknowledgeFullSync", attributes: .pushChannelV3)
+        WireLogger.pushChannel.debug("acknowledgeFullSync", attributes: .pushChannelV2)
         let acknowledgement = FullSyncAcknowledgment()
         let data = try encoder.encode(acknowledgement)
         try await write(data: data)
     }
 
+    public func acknowledgeMessageCount() async throws {
+        WireLogger.pushChannel.debug("acknowledgeMessageCount", attributes: .pushChannelV2)
+        let acknowledgement = MessageCountAcknowledgment()
```
johnxnguyen, line 145: question: is there no need to say which message count we're acknowledging?

WireDomain/Tests/WireDomainTests/Synchronization/IncrementalSyncTests.swift
```diff
@@ -117,10 +117,6 @@ final class IncrementalSyncTests: XCTestCase {
         store.deleteEventEnvelopeAtIndex_MockMethod = { _ in }
 
         // Live events are decrypted.
-//        decryptor.decryptEventsIn_MockMethod = { EventDecryptorResult(
-//            events: $0.events,
-//            brokenMLSGroupIDs: [Scaffolding.mlsGroupID]
-//        ) }
```
johnxnguyen, line 132: question: to delete?

WireAPI/Sources/WireAPI/Network/PushChannel/MessageCountAcknowledgment.swift
netbe: there's a new notification type message_count which gives the count of messages in the queue. this acknowledgement is for this `message_count`

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelV2.swift
netbe: no need, no it works as ackFullsync

WireAPI/Sources/WireAPI/APIs/PushChannelAPI/PushChannelV2API.swift
netbe: done 9102f920e895cf5d9f7b19e6eac9c23f530cc837

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelService.swift
netbe: done

WireAPI/Sources/WireAPI/Network/PushChannel/WebSocketNotification.swift
netbe: So I tried to move and have this:

```
  func toAPIModel() -> PushChannelV2.Element {
        switch (type, data) {
        case (.event, let .event(eventData)):
            return .event(
                UpdateEventEnvelope(
                    id: eventData.event.id,
                    events: eventData.event.payload.map(\.updateEvent),
                    isTransient: false,
                    deliveryTag: eventData.deliveryTag
                )
            )
        case (.notificationsMissed, .none):
            return .missedEvents
        case (.messagesCount, let .messageCount(data)):
            return .syncing(eventsCount: data.count)
        default:
            fatalError("unexpected data \(String(describing: data)), for type: \(type)")
        }
    }
    ```
    
    but because the returned data is only for 2 of the types we still end up with an impossible case
    moreover I'll still need to switch over the element to check how many elements are left until we're caught up

File: WireDomain/Tests/WireDomainTests/Synchronization/IncrementalSyncTests.swift (hunk 53 above)
Comment (netbe) [reply WireDomain/Tests/WireDomainTests/Synchronization/IncrementalSyncTests.swift:132]: it was commented out so I removed it

File: WireAPI/Sources/WireAPI/Network/PushChannel/AcknowledgmentType.swift (hunk 13 above)
Comment (netbe) [reply WireAPI/Sources/WireAPI/Network/PushChannel/AcknowledgmentType.swift]: done

File: WireAPI/Sources/WireAPI/Network/PushChannel/WebSocketNotification.swift (hunk 48 above)
Comment (netbe) [reply WireAPI/Sources/WireAPI/Network/PushChannel/WebSocketNotification.swift:36]: right, it's the number of events in the queue, so the number of events you'll need to processed to be up to date and live

File: WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift [hunk 54]
```diff
@@ -0,0 +1,338 @@
...
+
+    public func perform() async throws -> IncrementalSync.Token {
+        logger.debug("performing live sync", attributes: .syncAttributes(initialSync: false))
+        let pushChannel = try await pushChannelAPI.createPushChannel(clientID: selfClientID)
+
+        logger.debug("opening new push channel", attributes: .syncAttributes(initialSync: false))
+        syncStateSubject.send(.incrementalSyncing(.openPushChannel))
+        let liveEventStream = try await pushChannel.open()
+
+        let processedEnvelopeIDs = try await processPendingEvents(pushChannel: pushChannel)
```
netbe: @johnxnguyen here I added the missing pending events from the NSE

wire-ios-sync-engine/Source/Synchronization/IncrementalSyncObserver.swift
```diff
@@ -55,7 +55,8 @@ final class IncrementalSyncObserver: IncrementalSyncObserverProtocol {
                 .receive(on: decryptionQueue)
                 .sink { [weak self] syncState in
                     switch syncState {
-                    case .incrementalSyncing(.pullPendingEvents):
+                    case .incrementalSyncing(.pullPendingEvents),
+                         .incrementalSyncing(.receivingLiveEvents):
```
netbe, line 59: note: this is to differentiate IncrementalSyncV2 doing quickSync so we should not be allowed to send messages.

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelV2.swift
```diff
@@ -0,0 +1,197 @@
...
+//
+
+import Foundation
+import WireFoundation
+import WireLogging
+
+/// PushChannel using new async stream
+public final class PushChannelV2: PushChannelV2Protocol {
+
+    public enum Element: Equatable {
```
jullianm, line 26: suggestion: could be a more descriptive like `public enum State`?

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelV2.swift
```diff
@@ -0,0 +1,197 @@
...
+    private func tearDownKeepAliveTask() {
+        guard let keepAliveTask else { return }
+        WireLogger.pushChannel.debug("tearing down keep alive task", attributes: .pushChannelV2)
+        keepAliveTask.cancel()
+        self.keepAliveTask = nil
+    }
+
+    // MARK: - Acknowledgement
+
+    public func acknowledgeEvent(deliveryTag: UInt64, multiple: Bool = false) async throws {
```
jullianm, line 126: suggestion: API could be simplified a bit with a single public API acknowledge that takes an enum as parameter

```
public func acknowledge(acknowledgementType: AcknowlegmentType) {
    switch acknowlegmentType {
        case let .event(deliveryTag, multiple):
            {...}
        case .fullSync:
            {...}
        case .messageCount:
           {...}
    }
}
```

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
```diff
@@ -137,7 +140,11 @@ public final class ClientSessionComponent {
 
     private lazy var pushChannelAPI = PushChannelAPIBuilder(
         pushChannelService: pushChannelService
-    ).makeAPI()
+    ).makeAPI(for: apiVersion)
+
+    private lazy var newPushChannelAPI = PushChannelV2APIBuilder(
```
jullianm:
```suggestion
    private lazy var pushChannelV2API = PushChannelV2APIBuilder(
```

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
```diff
@@ -374,6 +381,17 @@ public final class ClientSessionComponent {
         journal: journal
     )
 
+    public lazy var newIncrementalSync = IncrementalSyncV2(
```
jullianm:
```suggestion
    public lazy var incrementalSyncV2 = IncrementalSyncV2(
```

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -0,0 +1,338 @@
...
+                        }
+
+                        var envelope = envelope
+                        envelope.events = try await decryptEnvelope(envelope)
+
+                        let index = try await storeEnvelope(envelope)
+
+                        await acknowledgeEnvelope(envelope, through: pushChannel)
+
+                        await processEnvelope(envelope)
```
jullianm, line 191: praise: decrypt, store, acknowledge, process, delete. very clear and nice!

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -0,0 +1,338 @@
...
+                switch element {
+                case .upToDate:
+                    logger.debug("upToDate event", attributes: .syncAttributes(initialSync: false))
+                    syncStateSubject.send(.liveSyncing(.ongoing))
+                    delegate?.isUpToDate(sync: self)
+                case .missedEvents:
+                    logger.debug("missedEvents event", attributes: .syncAttributes(initialSync: false))
+                    await delegate?.didMissedEvents(sync: self)
+                    // TODO: [WPB-17609] insert potential gap message here with messageLocalStore
+                    try await pushChannel.acknowledgeFullSync()
```
jullianm, line 167: question: if we missed events we'll trigger another (full) sync initial + incremental so should we acknowledge the full sync to the backend here if we missed events? Just wonder what impact it has on backend side

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -0,0 +1,338 @@
...
+            }
+
+        } catch {
+            // if we end up here, the pushChannel is closed
+            logger.warn(
+                "live event stream encountered error: \(String(describing: error))",
+                attributes: .syncAttributes(initialSync: false)
+            )
+            syncStateSubject.send(.liveSyncing(.finished))
+            delegate?.didFail(sync: self, error: error)
```
jullianm, line 213: question: I wonder if we could use our `syncStateSubject` here to propagate any error (or any updates for that matter) instead of relying on a delegate

WireDomain/Sources/WireDomain/Synchronization/SyncState.swift
```diff
@@ -55,6 +55,8 @@ public enum SyncState: Equatable {
         case openPushChannel
         case pullPendingEvents
         case processPendingEvents
+        case receivingLiveEvents // in async stream notifications, we don't pullPendingEvents, the pushChannel is open
+        // and events until we're up to date come from the websocket.
```
jullianm:
```suggestion
        // and events are received until we're up to date come from the websocket.
```

wire-ios-sync-engine/Source/Synchronization/IncrementalSyncObserver.swift
jullianm: sanity check that the incrementalSyncV2 update the state at some point to make sure this unblocks the message sender

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -231,6 +231,25 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
         }
     }
 
+    private func performInitialSyncV2() async throws {
```
jullianm, line 248: question: why is it called V2 ? Initial Sync remains unchanged no ?

jullianm (commented):
Nice work 👍  left a couple of suggestions/questions

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelV2.swift
netbe: it's not a State, I can add a description, it's objects delivered by the pushChannel

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
netbe: good catch

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: not sure to understand, here

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: in a second time, I guess we could rely only on the syncStateSubject. but for errors, they need to be in context (i think), ie:
`.livesSyncing(.finished(error))`

`private let syncStateSubject: CurrentValueSubject<SyncState, Never>`

maybe transform to 

`private let syncStateSubject: CurrentValueSubject<SyncState, SyncStateError>`

with SyncStateError being something like:
```
struct SyncStateError: Error {
   var state: SyncState
   var error: Error
}
```

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
netbe: it's to differentiate from legacyInitialSync:
```
if isSyncV2Enabled {
            try await performInitialSyncV2()
```

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
jullianm: if the push channel delivers "missedEvents", do we have to acknowledge full sync ? it is a bit confusing to me that we acknowledge something that hasn't technically taken place but maybe that's backend requirement

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
jullianm: yes, we could revisit this a little bit in second time because right now what is confusing to me is that the syncStateSubject only handles side effects and we also have delegate and catching thrown error from methods. I feel like we could all streamline this a little bit.
 
Maybe we could extend `syncStateSubject` responsibilities to handle any kind of states the sync could be in and rely on the observation of that subject to perform actions / handle side effects / errors from a single place.

(not for this PR though, probably when the sync V2 is stable)

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelV2.swift
```diff
@@ -0,0 +1,197 @@
...
+            guard let self else { return }
+            do {
+                for try await message in sourceStream {
+
+                    let result = try receiveMessage(message)
+
+                    continuation.yield(result)
+                    if numberOfReceivedEvents == remainingEventCount {
+                        continuation.yield(.upToDate)
+                    }
```
johnxnguyen, line 74: question: I wonder if we should be a bit more relaxed with this condition to account for the possibility of overshooting the number of events... `numberOfReceivedEvents >= remainingEventCount`

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -0,0 +1,338 @@
...
+        do {
+            logger.debug("processing stored update events", attributes: .syncAttributes(initialSync: false))
+            syncStateSubject.send(.incrementalSyncing(.processPendingEvents))
+            return try await processStoredEvents()
+        } catch {
+            logger.debug(
+                "incremental sync interrupted, tearing down...",
+                attributes: .syncAttributes(initialSync: false)
+            )
+            await pushChannel.close()
```
johnxnguyen: suggestion: seems strange to pass in the push channel only to use it for error handling, maybe catch one layer above and close the push channel there?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
johnxnguyen: The sync state was mostly intended as a way for the UI to know what's going on so it can inform the user, not for performing side effects. I think though we should take a broad look at the SyncAgent and friends to see how it can be simplified.

WireDomain/Sources/WireDomain/Synchronization/SyncState.swift
```diff
@@ -55,6 +55,8 @@ public enum SyncState: Equatable {
         case openPushChannel
         case pullPendingEvents
         case processPendingEvents
+        case receivingLiveEvents // in async stream notifications, we don't pullPendingEvents, the pushChannel is open
```
johnxnguyen, line 58: question: I wonder if it makes sense to have separate v1 and v2 incremental sync states so we don't confuse the two.

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
```diff
@@ -65,6 +65,7 @@ public final class ClientSessionComponent {
     private let proteusService: any ProteusServiceInterface
     private let coreCryptoProvider: any CoreCryptoProviderProtocol
 
+    public let asyncStreamEnabled: Bool
```
johnxnguyen, line 68: question: I thought we were going to call it `consumableNotifications`? `asyncStream` for me is too generic and easily confused with Swift Concurrency.

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelV2.swift
netbe: done

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelV2.swift
netbe: @johnxnguyen which case are you trying to catch here? backend says 100 remainingEventCount and we receive 101 numberOfReceivedEvents? if that's the case the sync bar would close one early.

WireDomain/Sources/WireDomain/Synchronization/SyncState.swift
netbe: yep, let's revise this in a followup PR

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
netbe: right this one goes away, in the next PR anyway, I can rename it there. https://github.com/wireapp/wire-ios/pull/3139

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelV2.swift
netbe: I am not sure if the benefit is obvious, as the public api would be just one method, the internal implementation would be a bit more complex:

For encoding the payload, the AcknowledgeType would become public + need to rework the enum to have associatedValue and manually implement encode function:
```
enum AcknowledgmentType: String, Encodable {
    case fullSync = "ack_full_sync"
    case messageCount = "ack_message_count"
    case event = "ack"
}
```

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: it has happened:

await delegate?.didMissedEvents(sync: self) // we perform the fullSync
 try await pushChannel.acknowledgeFullSync() // we acknowledge here

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: done

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
netbe: done

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelV2.swift
johnxnguyen: I know it shouldn't happen, but if it does, then we would only say we're up to date if we exactly match the count. I was thinking it would be more defensive to say we're up to date if we exceed the count.

johnxnguyen (approved):

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
jullianm: sounds good, let's discuss that later but there's definitely something to be careful about using the syncStateSubject.

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannelV2.swift
jullianm: let's keep it that way then,  no need to add complexity if the benefits are not that big, API is still pretty clear currently

jullianm (approved):
Approving 👍

PR #2919, opened 2025-04-18
fix: no new messages indicator - WPB-16349

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16349" title="WPB-16349" target="_blank">BugWPB-16349</a>  [iOS] New messages are not getting marked with new sync
  </td></table>
  <br />
 

### Issue

With new sync enabled, new unread messages do appear when performing the initial sync but when quick syncing (str: tapping on the conversation then back to the conversation list and trying to send a new message from another client), message will be not marked as unread in the list (nothing shows up) however it does show up when entering the conversation.

**Solution**: explicitly re-calculates the last unread messages after inserting the message.

Note: not sure why it does not work in new sync because implementation looks pretty similar than the one in legacy `ClientMessageRequestStrategy` - at least I haven't spotted any differences.

### Testing

- enable NEW SYNC
- tap on a conversation then go back to conversation list
- send message from another client
- new unread messages should show up in conversation list

Capture d’écran   2025-04-18 à 13 52 49

### Checklist

---


WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
```diff
@@ -406,6 +406,9 @@ public final class MessageLocalStore: MessageLocalStoreProtocol {
         clientMessage.unarchiveIfNeeded(conversation)
         clientMessage.updateCategoryCache()
         clientMessage.markAsSent()
+        // Re-calculates unread messages after insertion of the message.
+        // https://wearezeta.atlassian.net/browse/WPB-16349
+        conversation.calculateLastUnreadMessages()
```
netbe: question: we used to calculateLastUnreadMessages at the end of processing - https://github.com/wireapp/wire-ios/blob/05030fe82480a4bfd3f61b7311fe00f92a135ef6/wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift#L143

maybe that's what missing ?

jullianm: This one I think is to increase application badge number, I tried using it, just to be sure, and it doesn't fix our issue (nor does it update the app badge number but that's another topic)

dmitrysimkin (approved):

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
johnxnguyen: Maybe it also needs an explicit save to the context for those changes to propagate in the app.

johnxnguyen: Actually, we also call this at the end of processing stored events: https://github.com/wireapp/wire-ios/blob/36536aa6807544cdf4ecbd30ae1ee6a3171edce2/wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift#L265

So I suspect that that's what was missing. I would suggest though that instead of invoking this method after every message insertion, to do it once at the end of event processing.

johnxnguyen (changes requested):

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
jullianm: I did the requested changes in https://github.com/wireapp/wire-ios/pull/2919/commits/b50a807b55578c4174d7453b0aa4e8d3a7776f5f
thanks

johnxnguyen (approved):

PR #2921, opened 2025-04-21
chore: offload conversation calculation - WPB-17293

dmitrysimkin:
### Issue

Attempt to improve conversation screen performance issues
What was done:
- offload calculations of a sections to background thread. In order to do that I refetch messages in background thread, make calculations and re-set main thread `selfUser` and messages back to `section controller`
- Also made `sectionControllers` and `actionControllers` thread safe dictionary to synchronise access from 2 threads
- Added leading and trailing debounce for changes from DB, sometimes the are a lot of updates comes and we spam with them UI as well. Debounce limits amount of UI updates
- Removed `nonPersistenceObjectIdentifier` as It does not fit environment with 2 threads, it gives different IDs in different threads. 

### Testing

- All kind of message types
- See that all updates come and see in time (need to check because of usage of debounce)
- Long history with a lot of messages
- Live updates on hight load
- 

### Checklist

---


wire-ios-data-model/Source/Model/Message/ConversationMessage.swift
```diff
@@ -245,10 +242,6 @@ extension ZMMessage: ZMConversationMessage {
             .sortedAscendingPrependingNil(by: \.serverTimestamp)
     }
 
-    public var objectIdentifier: String {
-        nonpersistedObjectIdentifer
-    }
-
```
dmitrysimkin, line 251: Not used anymore

wire-ios/Wire-iOS/Sources/AppDelegate.swift
```diff
@@ -169,6 +169,9 @@ final class AppDelegate: UIResponder, UIApplicationDelegate {
 
         self.launchOptions = launchOptions ?? [:]
 
+        _ = NSAttributedString.paragraphStyle
+        _ = HorizontalMargins.windowWidth
```
dmitrysimkin: Initialise it here so it does not give main thread issues when used in background

wire-ios/Wire-iOS Tests/ConversationMessageCell/ConversationMessageSectionControllerTests.swift
```diff
@@ -64,13 +64,7 @@ final class ConversationMessageSectionControllerTests: XCTestCase {
...
 
         // GIVEN
-        let section = ConversationMessageSectionController(
-            message: MockMessage(),
-            context: context,
-            userSession: userSession,
-            useInvertedIndices: false,
-            contentWidth: 0
-        )
+        let section = makeSUT()
```
samwyndham, line 67: praise: Nice work cleaning up this boilerplate

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/GetUserByIdUseCase.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import CoreData
+import Foundation
+import WireDataModel
+
+protocol GetUserByIdUseCaseProtocol {
```
samwyndham: nitpick:  
```suggestion
protocol GetUserByIDUseCaseProtocol {
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/GetUserByIdUseCase.swift
```diff
@@ -0,0 +1,34 @@
...
+    func getUserById(id: Any, context: NSManagedObjectContext) -> (any UserType)?
```
samwyndham: nitpick:
```suggestion
    func getUserByID(id: Any, context: NSManagedObjectContext) -> (any UserType)?
```

WireFoundation/Sources/WireFoundation/LeadingTrailingDebouncer.swift
```diff
@@ -0,0 +1,81 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+public import Foundation
+
+public final class LeadingTrailingDebouncer<ID: Hashable> {
```
samwyndham, line 24: question: Did you consider wrapping apple's debounce publisher? https://developer.apple.com/documentation/combine/fail/debounce(for:scheduler:options:)

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
```diff
@@ -538,25 +666,32 @@ extension ConversationTableViewDataSource: ConversationMessageSectionControllerD
         _ controller: ConversationMessageSectionController,
         didRequestRefreshForMessage message: ZMConversationMessage
     ) {
-        reloadSections(newSections: postProcessedSections(calculateSections(updating: controller)))
+        debouncer.call(id: message.nonce!) { [weak self] in
+            guard let self else { return }
+            reloadSections(newSections: calculateSections(updating: controller))
+        }
```
samwyndham, line 721: question: Is this safe? If this is called multiple times will the last `controller` contains everything needed for the complete set of changes?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
```diff
@@ -781,3 +920,12 @@ extension ZMConversationMessage {
         textMessageData?.messageText
     }
 }
+
+extension ZMConversationMessage {
+
+    var isSentFromThisDevice: Bool {
```
samwyndham, line 989: thought (out of scope - no change necessary): `isSentFromThisDevice` this is not a very accurate naming for what it checks 🤔

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
```diff
@@ -117,34 +117,103 @@ final class ConversationTableViewDataSource: NSObject {
...
+        completion: @escaping ([Section]) -> Void
+    ) {
+        let messageIds = messages.map(\.objectID)
+        let selfUserOnMainThread = userSession.selfUser
+        let selfUserObjectID = selfUserOnMainThread.objectId
+        let firstUnreadMessageNonce = firstUnreadMessage?.nonce
+        let runId = UUID().uuidString
+
+        // Dispatching to background thread to offload sections calculation
+        userSession.performBackgroundTask { [weak self] backgroundContext in
```
samwyndham: issue: I _think_ using `performBackgroundTask` is the wrong API to use and can result in performance issue...

When we fetch objects into a context core data will access the SQLite DB and cache those object in that context. Therefore if we fetch those same objects again everything will be done in memory... core data won't need to access the DB.

As far as I know `performBackgroundTask` will create a new context each time it is called meaning that it can't rely on previously fetched objects.

The fix is to create a context for this specific case and always use that same context when calculating sections.

samwyndham (changes requested):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
dmitrysimkin: it dispatches work to one main thread and all happens there, no other thread involved. As I can see if it's called multiple times then it will be called one after another, so no issues should be. Am I missing something? Do you have any specific concerns @samwyndham ?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
dmitrysimkin: Interesting, really good comment. Thanks @samwyndham! I'll investigate

WireFoundation/Sources/WireFoundation/LeadingTrailingDebouncer.swift
dmitrysimkin: Apple's debounce is simpler, version I have is leading and trailing debounce, where if it's first call it will fire right away and with debounce other calls happens during time interval, and if there were debounced call it will fire once after time interval.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/GetUserByIdUseCase.swift
dmitrysimkin: Updated

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/GetUserByIdUseCase.swift
dmitrysimkin: Updated

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
dmitrysimkin: I made a change, instead of `performBackgroundTask` I introduced `viewBackgroundContext`. Thanks for the input, good one @samwyndham 👍

dmitrysimkin: Actually I got crash with self deleting messages when I did that. If you have saved background context it stores objects in memory but also you need to sync its data when there is new updates, like we do when `syncContext` saves we merge it to `viewContext`. And it adds overload on saving and keeping contexts in sync, which is bug prone and I tried and yet still even with syncyng syncContext with both `viewContext` and `viewBackgroundContext` it crashes. It does not crash only when I recreate `viewBackgroundContext`.

So I propose to keep that and see how it goes, right now I do not see issues with recreating `viewBackgroundContext` loading and scrolling looks ok, no delays

wire-ios-utilities/Source/Public/SafeSubscript.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+
+import Foundation
+
+public extension Collection {
+    /// Returns the element at the specified index if it is within bounds, otherwise nil.
+    subscript(safe index: Index) -> Element? {
+        indices.contains(index) ? self[index] : nil
+    }
+}
```
samwyndham: issue: Personally I think this as an anti-pattern and am against adding it to our utilities and is one of the things that the Swift team currently refuses to add to the standard library. See discussion in this thread: https://forums.swift.org/t/safe-random-access-collection-element/36509/6

I wouldn't mind it being a private helper in a particular file but if it is to be something global I think it should be discussed in the architecture forum.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
samwyndham: I see your point but I do feel that in large conversations it will cause obvious performance issues. Lets see what some of the other reviewers think

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
samwyndham: To clarify I don't mean any threading related issues. I just want to sanity check that logic wise, it is fine to discard previous updates if a new update comes through. I guess it is safe because the controller will have the latest message state right?

samwyndham (changes requested):

WireFoundation/Sources/WireFoundation/LeadingTrailingDebouncer.swift
samwyndham: OK I see. Interesting

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
dmitrysimkin: it's valid input, but our other code causes much more performance impact then this one :)
From other side when using new context each time helps not having a lot of objects in memory and no need to synchronize contexts when there are updates. Usually when you use messagers you do not scroll to much to too often, most of the time we just have tip of conversation history of like up to 100 messages which is easy to fetch. If we'd talk about like >1000 message each time then it would be bigger issue

wire-ios-utilities/Source/Public/SafeSubscript.swift
dmitrysimkin: We check index if within bounds everywhere, why this subscript is different from just another if? It's just syntax sugar. 
If you use it wise and not overuse just for simplicity I do not see a problem in it. 

Apple has their own responsibility to trend smth or not because they have users of swift of all skill levels. Here in Wire we all pretty experienced not to overuse it in bad way I think

WireFoundation/Sources/WireFoundation/LeadingTrailingDebouncer.swift
johnxnguyen, line 24: suggestion: add a doc comment explaining the debounce behavior.

wire-ios-utilities/Source/Public/SafeSubscript.swift
johnxnguyen: @samwyndham what are your reservations about it? From my understanding, adding a safe subscript is ignoring a larger problem which is not working with safe indices to begin with. @dmitrysimkin for what purpose do we need this exactly and why do we anticipate using invalid indices?

wire-ios/Wire-iOS/Sources/AppDelegate.swift
johnxnguyen: question: I wonder if there is a better place to put this.

wire-ios/Wire-iOS/Sources/Helpers/UITraitEnvironment+LayoutMargins.swift
```diff
@@ -38,9 +38,10 @@ struct HorizontalMargins {
         }
     }
 
+    static var windowWidth: CGFloat? = UIApplication.shared.delegate?.window??.frame.width ?? UIScreen.main.bounds.width
```
johnxnguyen: issue: won't this need to change if the window size changes? I'm thinking about landscape mode, or the various multitasking window sizes on iPad.

wire-ios/Wire-iOS/Sources/Helpers/ZMConversationMessage+Announcement.swift
```diff
@@ -34,7 +35,11 @@ extension ZMConversationMessage {
         if isKnock {
             return ConversationAnnouncement.Ping.description(senderName)
         } else if isText, let textMessageData {
-            let messageText = NSAttributedString.format(message: textMessageData, isObfuscated: isObfuscated)
+            let messageText = NSAttributedString.format(
```
johnxnguyen, line 38: question: I guess we need the formatted string to get the final plaintext version because of links and mentions and without markdown syntax, right?

wire-ios-data-model/Source/Model/Conversation/ConversationLike.swift
```diff
@@ -24,6 +24,10 @@ public typealias Conversation = ConversationLike & SwiftConversationLike
 // sourcery: AutoMockable
 @objc
 public protocol ConversationLike: AnyObject {
+
+    // Any as type eraser to hide NSManagedObjectID behind it
+    var objectId: Any { get }
```
johnxnguyen, line 29: question: why do we want to hide `NSManagedObjectID` if it's expected to be that? I would prefer transparency, and ideally I would also like to see these protocol abstractions of Core Data entities be removed, but that would be for another day.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationCollapsedMessageCell.swift
```diff
@@ -250,12 +257,19 @@ final class ConversationCollapsedMessageCellDescription: ConversationMessageCell
 
     typealias View = ConversationCollapsedMessageCell
 
-    let configuration: View.Configuration
+    var configuration: View.Configuration
```
johnxnguyen, line 260: question: do I understand correctly that the `configuration` of all the cells are now mutable because we are now reusing the cells and therefore need to change the configuration with new data?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
```diff
@@ -117,34 +118,120 @@ final class ConversationTableViewDataSource: NSObject {
...
+                let allMessages = messagesOnMainThread
+                for (messageObjectId, sectionController, context) in result {
+
+                    // saving calculations result in local cache
+                    self.sectionControllers.set(value: sectionController, for: messageObjectId)
+                    self.actionControllers.set(value: sectionController.actionController, for: messageObjectId)
+
+                    // Re-set messages from Main thread to section controller to not have crash with later interactions
+                    // with data
+                    if let mainThreadObject = allMessages.first(where: {
```
johnxnguyen: question: would it not be more efficient to use fetch the existing object rather than iterate through the array?

```swift
let mainThreadObject = ZMMessage.existingObject(id: sectionController.message.objectID, context: viewContext)
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
```diff
@@ -117,34 +118,120 @@ final class ConversationTableViewDataSource: NSObject {
...
+                try? backgroundContext.existingObject(with: objectID) as? ZMMessage
+            }
+
+            // sort if needed
+            messages = messages.sorted {
+                ($0.serverTimestamp ?? .distantPast) > ($1.serverTimestamp ?? .distantPast)
             }
 
-            return ArraySection(model: sectionIdentifier, elements: sectionController.tableViewCellDescriptions)
+            guard let selfUserOnBackgroundThread = getUserByIDUseCase.getUserByID(
```
johnxnguyen, line 138: question: why have a use case to fetch the user from a different context and not use a static helper like we do for other entities?

```
let selfUserOnBackgroundThread = ZMUser.existingObject(id: selfUserObjectID, context: backgroundContext)
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/GetUserByIdUseCase.swift
```diff
@@ -0,0 +1,34 @@
...
+protocol GetUserByIDUseCaseProtocol {
+    func getUserByID(id: Any, context: NSManagedObjectContext) -> (any UserType)?
+}
+
+class GetUserByIdUseCase: GetUserByIDUseCaseProtocol {
+    func getUserByID(id: Any, context: NSManagedObjectContext) -> (any UserType)? {
+        guard let managedObjectId = id as? NSManagedObjectID else {
+            return nil
+        }
+        return context.object(with: managedObjectId) as? UserType
```
johnxnguyen: question: why not `existingObject`?

wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack.swift
```diff
@@ -120,6 +121,8 @@ public class CoreDataStack: NSObject, CoreDataStackProtocol {
         messagesContainer.viewContext
     }
 
+    public var viewBackgroundContext: NSManagedObjectContext { messagesContainer.newBackgroundContext() }
```
johnxnguyen: suggestion: the name `viewBackgroundContext` is a little unclear, I would suggest to make this a function `newBackgroundContext()`.

wire-ios-utilities/Source/Public/SafeSubscript.swift
samwyndham: @johnxnguyen It is definitely a opinionated subject. These two comments from[ this thread](https://forums.swift.org/t/add-accessor-with-bounds-check-to-array/16871/57) express my opinion:
 
1. > You have an index into a collection from somewhere. This index has a meaning (maybe it's a particular element, or the beginning of a range, whatever).
Now you use the safe accessor, ostensibly because you aren't sure the index is valid anymore -- so the collection may have changed, I guess? If you aren't sure whether the collection hasn't changed in the meantime, how do you know that the safe accessor will not return you a different element (now at the original index) than the one you thought you were dealing with?
I fear this "safe" access pattern will lead to more code that is subtly wrong, but we'll notice it in fewer instance and thus have a harder time to fix it.

2. > The tl;dr here is that safe does not mean, "does not crash." It means "don't silently propagate invalid data", which can be accomplished here by either crashing or returning nil. As out of bounds indexing is a logic error, crashing is appropriate.

My fear is that it might be used too often and indicates that the programmer might not really understand what is happening in the code. However I'm not apposed to using it in a limited scope.

wire-ios/Wire-iOS/Sources/AppDelegate.swift
dmitrysimkin: I don't know, we need some place so it's not in the time when it's on background queue, also I can think it being part of app's initialisation, open to suggestions

wire-ios-utilities/Source/Public/SafeSubscript.swift
caldrian: **suggestion**: could we use a different label instead of `"safe"`?
I don't mind having this helper as long as it's just a getter, although generally I prefer to write plain Swift code.
```suggestion
public extension Collection {
    /// Returns the element at the specified index if it is within bounds, otherwise nil.
    subscript(ifExists index: Index) -> Element? {
        indices.contains(index) ? self[index] : nil
    }
}
```

dmitrysimkin: I call it to get `let lastVisibleMessage = dataSource.allMessages[safe: firstIndexPath.section]` which may be not existing anymore when it was deleted or for self-deleting messages. 
Messages indeed can be updated as we have fetch result controller which is updated when there is a change in DB, so the safe check will skip for this time and on next reload (which can happen with delay because of debauncer) it will be in synced

samwyndham: @dmitrysimkin I'm just afraid of silent failures and this convenience can be an easy way to ignore issues. However, maybe this is just personal opinion, maybe everyone else likes it. I see that Chris Lattner is in favor of it from the linked thread above 😄 Feel free to ignore this comment

wire-ios/Wire-iOS/Sources/Helpers/ZMConversationMessage+Announcement.swift
dmitrysimkin: yes

wire-ios-data-model/Source/Model/Conversation/ConversationLike.swift
dmitrysimkin: In prod code we need to pass objects by managed ID but there is no such thing in tests where we have only mocked objects, that's why I needed some abstraction over it

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationCollapsedMessageCell.swift
dmitrysimkin: Not reusing, but init of cells configuration happens on background thread, then when processing is finished we need to access some ZM objects on main thread, hence the set of configuration with ZM objects from main thread. 
Ideally would be that configuration is pure data without ZM but it's much to fix for now. Hope that makes sense

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
dmitrysimkin: Also because of tests, somewhere we use DB, somewhere we mock, so I needed a way to inject returned objects

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/GetUserByIdUseCase.swift
dmitrysimkin: no particular reason tbh, but agree `existingObject` is better fit 👍

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
dmitrysimkin: Yeah, controller is updated right away through CoreData

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
dmitrysimkin: Well spotted 👍 , it indeed faster to get message from db rater then filtering, improved

wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack.swift
dmitrysimkin: Renamed

wire-ios/Wire-iOS/Sources/Helpers/UITraitEnvironment+LayoutMargins.swift
dmitrysimkin: yeah, well spotted. anyway I don't need it here, I could pass `contentWidth` including margins from main thread

wire-ios-utilities/Source/Public/SafeSubscript.swift
dmitrysimkin: @samwyndham  I agree about ignoring issues, but I think here I have reasons to check if it's out of bounds. And we can try not to over use it.

@caldrian I've updated to `ifExists`

WireFoundation/Sources/WireFoundation/LeadingTrailingDebouncer.swift
dmitrysimkin: Added

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
samwyndham: 👍 OK I won't have this block this PR. I'll leave this point open so other reviews can comment if they want. However @dmitrysimkin feel free to resolve this PR when it comes to merging.

wire-ios-data-model/Source/Model/User/UserType.swift
```diff
@@ -24,6 +24,9 @@ public protocol UserType: NSObjectProtocol, UserConnections {
     /// The identifier which uniquely idenitifies the user in its domain
     var remoteIdentifier: UUID! { get }
 
+    /// Any as type eraser to hide NSManagedObjectID behind it
+    var objectId: Any { get }
```
samwyndham, line 28: question: What benefits do you see of erasing the type here? For me the negative is that we loose some compiler safety and have to perform some failable casting later on.

wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack.swift
```diff
@@ -120,6 +121,14 @@ public class CoreDataStack: NSObject, CoreDataStackProtocol {
         messagesContainer.viewContext
     }
 
+    public func newBackgroundContext() -> NSManagedObjectContext {
```
samwyndham, line 124: question: Is this for testing? If yes maybe add a doc comment somewhere.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
```diff
@@ -18,16 +18,20 @@
...
 extension String: Differentiable {}
+extension UUID: Differentiable {}
+
 extension AnyConversationMessageCellDescription: Differentiable {
 
     typealias DifferenceIdentifier = String
 
     var differenceIdentifier: String {
-        message!.objectIdentifier + String(describing: baseType)
+        String(describing: (message as! ZMMessage).objectID) + String(describing: baseType)
```
samwyndham, line 34: thought: This is potentially brittle because it relies on Apples description of objectID remaining something that is diffable. You could do something like `objectID.uriRepresentation().absoluteString` instead although I don't know about the performance implications.

samwyndham (approved):
Apologies @dmitrysimkin for the slow review of this. Well done for fighting with this screen. To be honest this screen deserves a complete rewrite as I believe the current way it is implemented is fundamentally flawed.

wire-ios-data-model/Source/Model/User/UserType.swift
dmitrysimkin: It was necessity, in some tests we use `ZMConversation` and in some mocked conversation which requires some abstraction over type

wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack.swift
dmitrysimkin: No, it's for actual background context to fetch and calculate sections

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
dmitrysimkin: I agree, but it was similar before, I had no time to benchmark that part as well. I addressed most crucial parts as I could see from instruments and logs. Think it's good for a start and later we can improve. 
I'll create follow up ticket with other possible imporvements I can see

johnxnguyen (approved):

PR #2923, opened 2025-04-22
refactor: migrate legacy events - WPB-15974

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15974" title="WPB-15974" target="_blank">TaskWPB-15974</a>  [iOS] Add migration of legacy events to new sync mechanism
  </td></table>
  <br />
 

### Issue
In this PR we add code to switch to the new sync and perform a migration of existing legacy events if needed. Changes include:
- Initialize new `UpdateEvent` from various legacy `ZMUpdateEvent` types. Not all event types are mapped, only those that can't be recovered via an initial sync (e.g message events)
- Create `UpdateEventMigrator` that is responsible for performing the migration
- Create `JournalStore` that is used to keep track of important pieces of information, in this case whether the new sync is enabled and whether an initial sync is required.
- After setting up the `CoreDataStack` and before setting up `ZMUserSession`, perform the enable the new sync if needed (with possible migration). The new sync is available to anyone using at least api version v8.
- Trigger an initial sync if one is required after the migration.
- Remove dependency on the `DeveloperFlag.newInitialSync`, replacing it with the journal entry for `SyncV2JournalEntry`.
- Use new NSE implementation if new sync is available, only allow it to run if new sync is enabled

### Testing
2. Be logged in on a previous version.
3. Be logged in on another device too, so you can later check the iOS is up to date.
4. Accumulate various update events while in the background.
5. Without having opened the app, upgrade to this version.
6. Open the app, let it load, and when the conversation list is visible, you should eventually see all your history and it should be up to date with your other clients.

---

### Checklist

---


WireDomain/Sources/WireDomain/Utilities/Journal/JournalStore.swift
```diff
@@ -0,0 +1,78 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// An object that facilitates access to the persistent
+/// journal entries for a particular user.
+
+public struct JournalStore {
```
johnxnguyen: This is intended as a general purpose tool to help in with interruptible flows. For instance, considering a particular migration, a single journal entry could keep track of several flags indicating if the migration is required, whether it is has begun, and whether it was completed.

wire-ios-data-model/Source/Model/ZMEventModel/StoredUpdateEvent/StoredUpdateEvent+Helpers.swift
```diff
@@ -20,35 +20,7 @@ import CoreData
 import Foundation
 import WireLogging
 
-@objc(StoredUpdateEvent)
```
johnxnguyen, line 23: This moved to its own file.

wire-ios-data-model/Source/Model/ZMEventModel/StoredUpdateEvent/StoredUpdateEvent+Helpers.swift
johnxnguyen, line 1: This file moved from request strategy to data model.

wire-ios-data-model/Source/Model/ZMEventModel/StoredUpdateEvent/StoredUpdateEvent.swift
johnxnguyen, line 1: This moved from request strategy to data model.

wire-ios-data-model/Source/Model/ZMEventModel/StoredUpdateEventEnvelope/StoredUpdateEventEnvelope.swift
johnxnguyen, line 1: This was moved to a new file.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1044,33 +1045,117 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+            onCompletion: { [weak self] coreDataStack in
                 guard let self else {
                     assertionFailure("expected 'self' to continue!")
                     return
                 }
 
-                let userSession = startBackgroundSession(
-                    for: account,
-                    with: coreDataStack
+                let journal = try! JournalStore(
```
johnxnguyen: I'm not sure what to do with these possible errors. It's hard to bubble errors up, but we also can't go forward without the journal.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1044,33 +1045,117 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+        }
+
+//        return completion()
+
+        // New sync is only available if the client supports v8.
+        guard apiVersion >= .v8 else {
+            return completion()
+        }
+
+        var journalEntry = try! journal.fetchEntry(SyncV2JournalEntry.self)
```
johnxnguyen: I also need to get rid of this try!

wire-ios-sync-engine/Source/SessionManager/UpdateEventMigratorDAO.swift
```diff
@@ -0,0 +1,231 @@
...
+    func save() async throws {
+        try await context.perform {
+            try context.save()
+        }
+    }
+
+}
+
+@available(iOS 17, *)
+actor UpdateEventMigratorDAO2: UpdateEventMigratorDAOProtocol {
```
johnxnguyen: I'm wondering if I should leave this here or remove it and use the other implementation instead.

WireDomain/Sources/WireDomain/Utilities/Journal/JournalEntry.swift
```diff
@@ -16,22 +16,20 @@
...
-import Foundation
+/// An object that can be recorded into the user's
+/// account journal. This can be used to keep track
+/// of information to help make decisions at a later
+/// date, such as whether some migration code needs
+/// to be run.
 
-/// Computes an hash to compare UpdateEvent and StoredUpdateEvent
-enum EventHasher {
+public protocol JournalEntry: Codable {
```
samwyndham: praise: I like the naming/concept of _journal_

WireDomain/Sources/WireDomain/Utilities/Journal/JournalStore.swift
samwyndham: question: Am I correct that you will be changing the implementation here or should I review it?

wire-ios-data-model/Source/Model/ZMEventModel/StoredUpdateEventEnvelope/StoredUpdateEventEnvelope+Helpers.swift
```diff
@@ -16,39 +16,45 @@
...
+    /// - Parameters:
+    ///   - data: JSON encoded data containing the envelope.
+    ///   - sortIndex: An index specifying the order of the envelope.
+    ///   - context: The context to insert into.
+    ///
+    /// - Returns: The newly inserted object.
 
-    @NSManaged public var data: Data
+    @discardableResult
+    static func insertNewObject(
```
samwyndham, line 33: suggestion: Possibly I missed it but I think it is worth adding some tests for these new methods on `StoredUpdateEventEnvelope`. Or create a follow up ticket if time doesn't permit

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
samwyndham: Yeah, I think the only options are 1. crash, as you are doing or 2. display some message other app state where the user is asked to upgrade / contact customer support. But `2` is not worth the effort unless we have something in place already.

wire-ios-sync-engine/Source/SessionManager/UpdateEventMigrator.swift
```diff
@@ -0,0 +1,408 @@
...
+                // Store the next event at this index.
+                currentIndex += 1
+            }
+
+            WireLogger.sync.debug("deleting batch of legacy events...")
+            await dao.deleteNextBatchOfLegacyEvents()
+        }
+
+        WireLogger.sync.debug("no more legacy events to migrate, saving...")
+        try await dao.save()
```
samwyndham: question: Do we want to move this save into the end of the while loop?

wire-ios-sync-engine/Source/SessionManager/UpdateEventMigratorDAO.swift
```diff
@@ -0,0 +1,230 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+@preconcurrency import CoreData
```
samwyndham, line 19: suggestion: Add tests for these classes. I wonder if you can test both implementations with the same tests, by having an array of SUTs

samwyndham (commented):
Nice work!  I've left a few comments. I _think_ you are still working on this PR right? Once done I will re-review

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnxnguyen: this is now irrelevant with the new `Journal`.

WireDomain/Sources/WireDomain/Notifications/VerifyUserSessionUseCase.swift
```diff
@@ -31,26 +31,34 @@ struct VerifyUserSessionUseCase {
...
         self.cookieStorage = cookieStorage
         self.coreData = coreData
     }
 
     /// Ensures user is properly authenticated.
 
     func invoke() async throws {
+        guard journal[.isSyncV2Enabled] else {
+            throw Failure.syncV2IsNotEnabled
+        }
```
johnxnguyen, line 60: This is to handle the scenario where the extension is started before the main app has had a chance to enable sync v2. In which case, we won't process any notifications.

WireDomain/Sources/WireDomain/Utilities/Journal/Journal.swift
```diff
@@ -0,0 +1,93 @@
...
+/// For instance, the journal can be used to keep track of various
+/// flags, such as whether a particular migration is required and/or
+/// completed.
+///
+/// The Journal users a scoped `UserDefaults` suite that is accessible
+/// within the app group.
+
+import WireFoundation
+
+public class Journal: JournalProtocol {
```
johnxnguyen, line 33: I've updated the implementation to use shared user defaults instead of the file system.

wire-ios-data-model/Source/Model/ZMEventModel/StoredUpdateEventEnvelope/StoredUpdateEventEnvelope+Helpers.swift
johnxnguyen: Yes there are some tests missing, I'll create follow up tickets.

wire-ios-sync-engine/Source/SessionManager/UpdateEventMigratorDAO.swift
johnxnguyen: I'll create some follow up tickets

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1044,33 +1045,106 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+                    try await migrator.migrateLegacyUpdateEvents()
+                    // Since we only migrate some events, we require an
+                    // initial sync to ensure we didn't miss updates.
+                    journal[.isInitialSyncRequired] = true
+                } else {
+                    WireLogger.sync.debug("no migration needed")
+                }
+
+                journal[.isSyncV2Enabled] = true
+                try journal.save()
```
jullianm: question: why do we need to explicitly save here ?

wire-ios-sync-engine/Source/Synchronization/ZMOperationLoop.h
```diff
@@ -47,7 +48,8 @@ extern NSString * const ZMPushChannelIsOpenKey;
                   pushNotificationStatus:(PushNotificationStatus *)pushNotificationStatus
                                    uiMOC:(NSManagedObjectContext *)uiMOC
                                  syncMOC:(NSManagedObjectContext *)syncMOC
-                  isDeveloperModeEnabled:(BOOL)isDeveloperModeEnabled;
+                  isDeveloperModeEnabled:(BOOL)isDeveloperModeEnabled
+                         isSyncV2Enabled:(BOOL)isSyncV2Enabled;
```
jullianm, line 52: question: why is this needed at this level ?

wire-ios/Wire Notification Service Extension/NotificationService.swift
```diff
@@ -44,13 +43,42 @@ final class NotificationService: UNNotificationServiceExtension {
...
+            notificationService = loadNotificationService()
+        }
+
+        if let notificationService {
+            notificationService.didReceive(
+                request,
+                withContentHandler: contentHandler
+            )
+        } else {
+            WireLogger.notifications.warn("no service loaded yet, suprressing notification")
```
jullianm: suggestion: I think we fall into this case (notification service is nil) only if the BackendInfo.apiVersion is nil so it might be nice to specify it in the log (or maybe add a dedicated log in the guard case statement in `loadNotificationService` )

wire-ios-transport/Source/Public/ZMTransportSession.h
```diff
@@ -87,7 +87,8 @@ extern NSString * const ZMTransportSessionReachabilityIsEnabled;
          applicationGroupIdentifier:(nullable NSString *)applicationGroupIdentifier
                  applicationVersion:(nonnull NSString *)applicationVersion
                       minTLSVersion:(nullable NSString *)minTLSVersion
-                       selfClientID: (nullable NSString *)selfClientID;
+                       selfClientID:(nullable NSString *)selfClientID
+                    isSyncV2Enabled:(bool)isSyncV2Enabled;
```
jullianm, line 91: question: why is it needed at this level ?

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1044,33 +1045,106 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+                        journal: journal
+                    )
+
+                    self.triggerMigrationsNeedsActionsIfNeeded(
+                        journal: journal,
+                        userSession: userSession
+                    )
+
+                    onCompletion(userSession)
+                }
```
jullianm, line 1087: suggestion: I think here we could avoid the `enableNewSyncIfNeeded` completion handler and make the code a little bit more readable: 

```
if shouldActivateNewSync() { // here we check api version and whether syncV2 is not already enabled
    try await activateNewSync() // here we perform the event migration to new sync and turn syncV2 flag 
}

// here goes the code from the completion handler above
```

what do you think ?

WireDomain/Sources/WireDomain/Utilities/Journal/JournalKey.swift
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+}
+
+public extension JournalKey where Value == Bool {
+
+    /// Whether new sync mechanism (initial sync, incremental
+    /// sync, live sync) is used.
+
+    static let isSyncV2Enabled = Self(
```
jullianm, line 44: thought: The V2 part could be a little bit misleading as it might imply there are (or could be) several versions of the sync (like we have for API versions). 
Maybe we could rename it `isNewSyncEnabled` but tbh I don't have a strong opinion on this and I'm fine with that naming.

jullianm (commented):
Nice work! left a couple of comments (questions mostly) before approving

wire-ios-sync-engine/Source/SessionManager/UpdateEventMigratorDAO.swift
jullianm: I guess we could keep it so when we move to iOS 17, we keep this one and clean up easily the rest

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnxnguyen: it's not absolutely necessary but I thought it'd be good to make sure we save so the NSE can know about the new value as soon as possible.

wire-ios-sync-engine/Source/Synchronization/ZMOperationLoop.h
johnxnguyen: I didn't want to inject the `Journal`, and we only need to know if the new sync is enabled anyway, so that we prevent trying to open the legacy push channel.

wire-ios-transport/Source/Public/ZMTransportSession.h
johnxnguyen: Passing the value all the way to the legacy push channel to make sure we don't open it.

WireDomain/Sources/WireDomain/Utilities/Journal/JournalKey.swift
johnxnguyen: Yeah I thought about that but "new" is only really new now and not in the future. Besides, we don't plan to keep both syncs around for a long time, so we'll delete this flag.

wire-ios/Wire Notification Service Extension/NotificationService.swift
johnxnguyen: done 66585cadf1a6ee99be6eab38217955544ab761a3

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnxnguyen: done f9114715d4f5f39c0ea0591606fac19472ba0c42

jullianm (approved):
Looks good, nice work!

WireDomain/Sources/WireDomain/Utilities/Journal/Journal.swift
```diff
@@ -0,0 +1,91 @@
...
+            storage.set(newValue, forKey: rawKey(for: key))
+        }
+    }
+
+    /// Save changes immediately.
+    ///
+    /// Note: saving happens automatically, so only call this
+    /// method if absolutely necessary.
+
+    public func save() throws {
```
samwyndham: According to apples doc comments this method is not necessary I believe:

```swift
    /**
     -synchronize is deprecated and will be marked with the API_DEPRECATED macro in a future release.
     
     -synchronize blocks the calling thread until all in-progress set operations have completed. This is no longer necessary. Replacements for previous uses of -synchronize depend on what the intent of calling synchronize was. If you synchronized...
     - ...before reading in order to fetch updated values: remove the synchronize call
     - ...after writing in order to notify another program to read: the other program can use KVO to observe the default without needing to notify
     - ...before exiting in a non-app (command line tool, agent, or daemon) process: call CFPreferencesAppSynchronize(kCFPreferencesCurrentApplication)
     - ...for any other reason: remove the synchronize call
     */
    open func synchronize() -> Bool
    ```

samwyndham (approved):
Nice work John!

WireDomain/Sources/WireDomain/Utilities/Journal/Journal.swift
johnxnguyen: ok then I will remove it

wire-ios-sync-engine/Source/SessionManager/UpdateEventMigrator.swift
johnxnguyen: yeah I think that makes sense, done: 83cf172547fc33eb8c835ab951acc82d995dcb07

PR #2925, opened 2025-04-22
fix: raw api error displayed to the user - WPB-17232

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17232" title="WPB-17232" target="_blank">BugWPB-17232</a>  [iOS] Rate limit error on manage devices screen is just the raw error
  </td></table>
  <br />
 

### Issue

**Context**: When removing a user client, a raw api error is displayed to the user, looking at the error it happens when too many requests were made to the server.

**Causes**:  When mapping the error `RemoveUserClientError` some errors are not mapped to `RemoveUserClientError` and propagated as is to to the UIViewController extension helper method `showAlert` where we display the `error.localizedDescription`.

**Solutions**: Ensure all errors, specifically the `TooManyRequests` one are caught and mapped to a `RemoveUserClientError` so we provide a readable message to the user.

image

### Testing

- Use the 2FA Login automated bund test to replicate easily (removing the wait before login). 

Otherwise:
- Have 7 devices on your account
- Before logging in, cause 4 password related requests in a 4-minute window
- Login with the 8th device
- Accept prompt to remove a device and choose one

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/Use cases/RemoveUserClientUseCase.swift
+++ b/wire-ios-sync-engine/Source/Use cases/RemoveUserClientUseCase.swift
@@ -96,12 +96,20 @@ class RemoveUserClientUseCase: RemoveUserClientUseCaseProtocol {
             case .invalidCredentials, .missingAuth, .badRequest:
                 throw RemoveUserClientError.invalidCredentials
 
+            case .unknown:
+                switch failureResponse.code {
+                case TooManyRequestsStatusCode:
+                    throw RemoveUserClientError.tooManyRequests
+                default:
+                    throw RemoveUserClientError.generic
+                }
+
             default:
-                throw failure
+                throw RemoveUserClientError.generic
             }
 
         default:
-            throw failure
+            throw RemoveUserClientError.generic
         }
     }
 }
@@ -111,5 +119,10 @@ public enum RemoveUserClientError: Error {
     case clientToDeleteNotFound
     case clientDoesNotExistLocally
     case invalidCredentials
+    case tooManyRequests
+
+    // other error types that are not explicitly handled
+    // this allows showing a readable error message to the user
+    case generic
 
 }
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/RemoveClients/RemoveClientsViewController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/RemoveClients/RemoveClientsViewController.swift
@@ -220,6 +220,11 @@ final class RemoveClientsViewController: UIViewController,
 
 extension RemoveUserClientError: LocalizedError {
     public var errorDescription: String? {
-        L10n.Localizable.General.failure
+        switch self {
+        case .tooManyRequests:
+            L10n.Localizable.Error.User.tooManyRequests
+        default:
+            L10n.Localizable.General.failure
+        }
     }
 }
--- a/wire-ios-sync-engine/Tests/Source/Use cases/RemoveUserClientUseCaseTests.swift
+++ b/wire-ios-sync-engine/Tests/Source/Use cases/RemoveUserClientUseCaseTests.swift
@@ -100,6 +100,29 @@ final class RemoveUserClientUseCaseTests: XCTestCase {
         }
     }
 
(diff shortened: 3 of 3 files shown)


wire-ios-sync-engine/Source/Use cases/RemoveUserClientUseCase.swift
```diff
@@ -96,12 +96,20 @@ class RemoveUserClientUseCase: RemoveUserClientUseCaseProtocol {
             case .invalidCredentials, .missingAuth, .badRequest:
                 throw RemoveUserClientError.invalidCredentials
 
+            case .unknown:
+                switch failureResponse.code {
+                case TooManyRequestsStatusCode:
+                    throw RemoveUserClientError.tooManyRequests
```
jullianm, line 102: this fixes the current bug by displaying a nice, readable message to the user

wire-ios-sync-engine/Source/Use cases/RemoveUserClientUseCase.swift
```diff
@@ -96,12 +96,20 @@ class RemoveUserClientUseCase: RemoveUserClientUseCaseProtocol {
...
+                }
+
             default:
-                throw failure
+                throw RemoveUserClientError.generic
             }
 
         default:
-            throw failure
+            throw RemoveUserClientError.generic
```
jullianm, line 112: these were also changed in case we have other error types (no bugs were raised for this but could have happened in the future)

wire-ios/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/RemoveClients/RemoveClientsViewController.swift
```diff
@@ -220,6 +220,11 @@ final class RemoveClientsViewController: UIViewController,
 
 extension RemoveUserClientError: LocalizedError {
     public var errorDescription: String? {
-        L10n.Localizable.General.failure
+        switch self {
+        case .tooManyRequests:
+            L10n.Localizable.Error.User.tooManyRequests
+        default:
+            L10n.Localizable.General.failure
+        }
```
jullianm, line 228: when displaying the error message using the localized description, it will get the values here (and not the raw api error)

samwyndham (approved):
LGTM!

wire-ios-sync-engine/Source/Use cases/RemoveUserClientUseCase.swift
```diff
@@ -111,5 +119,10 @@ public enum RemoveUserClientError: Error {
     case clientToDeleteNotFound
     case clientDoesNotExistLocally
     case invalidCredentials
+    case tooManyRequests
```
caldrian, line 122: Isn't this a generic error, not specific to the remove-user-client API?

jullianm: yes I think it's a generic one but it's not treated properly when displaying the error message to the user so I thought we'd handle it explicitly as part of the RemoveUserClientError definition with the proper localization string. I could always remove that use case and solely use the `generic` one but it will display "Something went wrong" without more details.

jullianm: maybe I could rename `generic` by `other` to handle all other error cases ?

caldrian (approved):

wire-ios-sync-engine/Source/Use cases/RemoveUserClientUseCase.swift
```diff
@@ -96,12 +96,20 @@ class RemoveUserClientUseCase: RemoveUserClientUseCaseProtocol {
             case .invalidCredentials, .missingAuth, .badRequest:
                 throw RemoveUserClientError.invalidCredentials
 
+            case .unknown:
+                switch failureResponse.code {
+                case TooManyRequestsStatusCode:
+                    throw RemoveUserClientError.tooManyRequests
+                default:
+                    throw RemoveUserClientError.generic
```
netbe, line 104: question: we're loosing the underlying error here, no? should we not keep it for debugging as associated value?

PR #2948, opened 2025-04-25
fix: sync status bar never goes away on error - WPB-17151

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17151" title="WPB-17151" target="_blank">BugWPB-17151</a>  [iOS] if sync fails the sync bar never goes away
  </td></table>
  <br />
 

### Issue

**Context**: When sync fails, sync status bar never goes away.

**Causes**: We don't do anything when either the initial or the incremental sync fails, just logging the error.

**Solution**:  Trigger an update of the sync status bar state when the sync fails.

### Testing

Only testable internally - on Xcode added specific debug code:

- Use existing debug action (like the incremental sync one)
- Add code to update the sync status bar (so it changes to .onlineSynchronizing) in this method
- In the same method, call syncAgent.resume()
- in one of the initial sync / incremental sync methods, force throwing an error after a couple of seconds (so we have time to see the sync status bar showing up)

After triggering the debug action, sync status bar should go away when sync fails

### Checklist

### DIFF
--- a/wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
+++ b/wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
@@ -70,7 +70,10 @@ final class SyncAgent: NSObject {
             do {
                 try await performSync()
             } catch {
-                WireLogger.sync.error("failed to perform sync: \(String(describing: error))")
+                delegate?.syncAgentDidFailSyncing(
+                    self,
+                    error: error
+                )
             }
         }
     }
@@ -112,10 +115,9 @@ final class SyncAgent: NSObject {
                 WireLogger.sync.error("failed to perform new initial sync: \(String(describing: error))")
                 throw error
             }
-
+            // Incremental sync automatically follows the slow sync.
             try await performIncrementalSync()
         } else {
-            // Incremental sync automatically follows the slow sync.
             legacySyncStatus.forceSlowSync()
         }
     }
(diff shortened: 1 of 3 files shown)


KaterinaWire (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1077,6 +1078,15 @@ extension ZMUserSession: SyncAgentDelegate {
         didFinishIncrementalSync(isRecovering: isRecovering)
     }
 
+    func syncAgentDidFailSyncing(_ syncAgent: SyncAgent, error: Error) {
```
KaterinaWire, line 1081: **question:** why do you need to pass `syncAgent`? looks like you don't use it

jullianm: to align with the other delegate methods (they don't use the syncAgent as well) but more generally I think it's a common pattern for delegate to pass (even if not used) the object it's called from

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -112,10 +115,9 @@ final class SyncAgent: NSObject {
                 WireLogger.sync.error("failed to perform new initial sync: \(String(describing: error))")
                 throw error
             }
-
+            // Incremental sync automatically follows the slow sync.
```
samwyndham: suggestion: Don't move the comment or improve it. I think it's purpose in the original position is to say that although we don't specify performing incremental sync, it happens explicitly after legacySyncStatus.forceSlowSync()

samwyndham (approved):
Nice work

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
jullianm: you're right, thanks

PR #2952, opened 2025-04-29
refactor: encoding & decoding of update events - WPB-15107

samwyndham:
### Issue

In new sync, the content of WireAPI event updates are stored as Data on the `StoredUpdateEventEnvelope` model. This data is encoded and decoded by conforming the `WireAPI` models to `Codable`. These models depend on other models such as `QualifiedID`. This setup is brittle because if anything changes on any of these models, the data may be unable to be decoded which may result in lost messages. This corruption could happen very simply, for example an innocent renaming of a property during a refactor, of if a model is extended to have other properties for a new API version.

This PR attempts to address this by introducing a specific swift package where this encoding & decoding takes place. If one of the wire API models changes compilation will be broken in this new package. This package should be covered by tests making in clear that we need to deal with this issue - these tests will be done in a follow up PR.

To achieve these goals, this PR:

- Introduces a new SPM target, `WireUpdateEventCoding` with a very narrow public API - just 2 methods.
- Removes `Codable` conformance from the `WireAPI` event update models.
- Adds missing public inits to the `WireAPI` event update models
- Duplicates all models necessary for representing the event updates including dependency models such as `QualifiedID` prefixed with `Storable`. 
- Adds mapping methods from `Storable` to `WireAPI` models.

#### Decisions

1. The `Storable` models make use of **automatic** `Codable` conformance. I investigated using custom implementations as we are doing in things like `ConversationDeleteEventDecoder` but found that more complex and harder to reason about.
2. These `Storable` models are intended as a kind of [anti corruption layer](https://learn.microsoft.com/en-us/azure/architecture/patterns/anti-corruption-layer). As such I decided not necessarily to care about the custom property key names defined in WireAPI models as these might change anyway with API version updates. For example:

```swift
// WireAPI
public enum ConversationAccessRole: String, Codable, Sendable {
    case teamMember = "team_member"
    case nonTeamMember = "non_team_member"
    case guest
    case service
}

// WireUpdateEventCoding
enum StorableConversationAccessRole: String, Equatable, Codable, Sendable {
    case teamMember
    case nonTeamMember
    case guest
    case service
}
```

### Testing

I'm not sure the best way to test this. @johnxnguyen any ideas?

---

### Checklist


WireAPI/Sources/WireAPI/Models/Conversation/Conversation.swift
```diff
@@ -126,4 +126,50 @@ public struct Conversation: Equatable, Codable, Sendable {
 
     }
 
+    public init(
```
samwyndham, line 129: For the public inits I just used those generated by Xcode, and didn't attempt to improve on them.

WireDomain/Sources/WireUpdateEventCoding/UpdateEventCoder.swift
```diff
@@ -0,0 +1,37 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+public import Foundation
+public import WireAPI
+
+public struct UpdateEventCoder {
```
jullianm: suggestion: rename it `StorableUpdateEventCoder` and add comment to emphasize that this coder is solely used to safely encode/decode update events to/from the db by bridging from/to API models

WireDomain/Sources/WireUpdateEventCoding/Utilities/Array+Utilities.swift
```diff
@@ -0,0 +1,25 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+extension Array where Element: Hashable {
+
+    func toSet() -> Set<Element> {
+        Set(self)
+    }
+
+}
```
jullianm, line 25: suggestion: probably worth moving this to WireFoundation/Utilities

jullianm (approved):
Nice work, much safer!

johnxnguyen (approved):
Fantastic!

WireDomain/Sources/WireUpdateEventCoding/Utilities/Array+Utilities.swift
samwyndham: Currently I'm not sure that this is a _good_ general utility to have. I'll keep it where it is for the time being.

WireDomain/Sources/WireUpdateEventCoding/UpdateEventCoder.swift
samwyndham: Updated. Thanks for the suggestion

PR #2958, opened 2025-04-30
fix: app sends large amount of last read messages - WPB-17439

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17439" title="WPB-17439" target="_blank">BugWPB-17439</a>  [iOS] Application sends a lot of MLS messages in a loop
  </td></table>
  <br />
 

### Issue
After enabling mls client registration on cloud, we observed some clients getting stuck sending a large amount of messages. We believe that the messages are all originating from the MLS self conversation, and are all likely to be "last read" messages (these tell your other clients about updated last read markers for various conversations).

We found that even without an mls client registered, the app would append a last read message to **both the proteus and mls self conversations.**. Without an mls client registered, these messages added to the mls self conversation would fail to send, and should have been expired and not retried. However, these unsent messages still exist in the database, and their number would accumulate. We observed in one case over 135,000 unsent messages. 😱

Each time the app is launched, it would have fetched all unsent messages needing to be sent (this excludes expired messages, which should be the case here but it could be that these last read messages weren't expired or became unexpired... not sure) and then try to send them. Once an mls client was registered, these could all proceed to be encrypted and sent to the backend, which is causing an immense load on the app and would take a long time to complete.

This immense load most likely degraded the performance of the app so much that it became unusable, and certainly would have blocked any new messages that the new user tried to send. However, given enough time, the message queue would eventually be cleared and the app could return to its normal state.

### Fix
The immediate remedy is to simply not send anything over the mls self conversation. This is fine because we still send successfully over the proteus self conversation. Additionally, we won't try use the mls self conversation if we don't have an mls client.

The longer term fix may be more complex which we need more time to understand, such as trying to know why the expired messages were retried automatically.

### Testing
Can only be verified artificially by inserting a large number of last read messages on the mls self conversation with no registered mls client, then register an mls client.

---

### Checklist

---

### DIFF
--- a/wire-ios-data-model/Source/Model/Conversation/ZMConversation+SelfConversation.swift
+++ b/wire-ios-data-model/Source/Model/Conversation/ZMConversation+SelfConversation.swift
@@ -123,7 +123,13 @@ public extension ZMConversation {
         _ content: MessageCapable,
         in context: NSManagedObjectContext
     ) throws -> ZMClientMessage? {
-        guard let selfConversation = ZMConversation.fetchSelfMLSConversation(in: context) else {
+        let selfClient = ZMUser.selfUser(in: context).selfClient()
+        let hasRegisteredMLSClient = selfClient?.hasRegisteredMLSClient ?? false
+
+        guard
+            hasRegisteredMLSClient,
+            let selfConversation = ZMConversation.fetchSelfMLSConversation(in: context)
+        else {
             return nil
         }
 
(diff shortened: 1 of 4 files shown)


netbe (approved):
looks great!

dmitrysimkin (approved):

samwyndham (approved):
Nice work!

KaterinaWire (approved):

PR #2960, opened 2025-05-02
feat: app version migration service - WPB-18482

johnxnguyen:
### Issue
Often times we have a need to perform some function when users update their app. For instance, it could be an action to remedy an issue, or it could be some underlying cleanup of stale data.

In the past we have had similar mechanisms (eg. sync engine hotfixes, data model patches), but they are now no longer operational because they rely on releases of framework versions which we stopped doing since we moved to a monorepo.

This PR contains a solution that uses the app semantic version to perform interruptible migrations. In a nutshell, it works by keeping track of a previous version, comparing it with the current version, filtering all the eligible migrations, and iteratively performing each migration.

This system would have ideally been introduced in the first release of the app as it would eliminate a challenging edge case: how are we able to determine if the user **updated** the app from a version before the service was introduced, to a version that contains the system (and potentially the first migration)? 

The only reliable way (I think) that we are able to determine this is by checking very early on in the app life cycle which accounts already exist (and therefore an update has occurred) and which are added (and therefore are consider completely new).

### Testing

There are no migrations as of yet.... perhaps I could add a developer menu cell to reveal the what the last version marked is...

---

### Checklist

---


WireDomain/Sources/WireDomain/AppVersionMigration/AppVersionMigrationService.swift
```diff
@@ -0,0 +1,108 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+class AppVersionMigrationService {
```
johnxnguyen: todo: add tests that the migrations are always run in sorted order

wire-ios/Wire-iOS/Sources/AppDelegate.swift
```diff
@@ -369,8 +369,14 @@ private extension AppDelegate {
...
     private func createSessionManager() throws -> SessionManager {
-        guard let appVersion = Bundle.main.infoDictionary?[kCFBundleVersionKey as String] as? String  else {
-            throw SessionManagerSetupError.missingAppVersion
+        let infoDictionary = Bundle.main.infoDictionary
+
+        guard let currentAppVersion = infoDictionary?["CFBundleShortVersionString"] as? String  else {
+            throw SessionManagerSetupError.missingCurrentAppVersion
+        }
+
+        guard let currentBuildVersion = infoDictionary?[kCFBundleVersionKey as String] as? String  else {
```
KaterinaWire, line 378: **question:** how do we use `currentBuildVersion` ?

WireDomain/Tests/WireDomainTests/AppVersionMigration/AppVersionMigrationServiceTests.swift
```diff
@@ -0,0 +1,249 @@
...
+        #expect(error == "Interrupted migration: 1.0.0")
+        #expect(migrationsPeformed == ["0.2.0"])
+        #expect(journal[.lastCompletedAppVersionMigration] == "0.2.0")
+
+        // When
+        migrationToInterrupt = nil
+        try await sut.performAppMigrations()
+
+        // Then
+        #expect(migrationsPeformed == ["0.2.0", "1.0.0", "1.2.3"])
```
KaterinaWire: **question:** if `journal[.lastCompletedAppVersionMigration]`  is `"0.2.0"`, why do we expect to have it the performed migrations?

johnxnguyen: `migrationsPerformed` is not cleared after each call to `performAppMigrations()`, its just a list of all the migrations that have been performed so far in the test.

KaterinaWire (approved):

WireDomain/Tests/WireDomainTests/AppVersionMigration/AppVersionMigrationServiceTests.swift
KaterinaWire: maybe we can clean up `migrationsPerformed` in `Then`, no?

johnxnguyen: done 09a10c4ea9fa0f7143eb81330ab43c8c1963a871

WireDomain/Sources/WireDomain/AppVersionMigration/AppVersionMigrationService.swift
```diff
@@ -0,0 +1,121 @@
...
+        self.journal = journal
+        self.currentVersion = currentVersion
+        self.allMigrations = allMigrations
+
+        assert(
+            allMigrations.allSatisfy { $0.version <= currentVersion },
+            "There should be no migration pinned to a future version"
+        )
+
+        assert(
```
netbe, line 49: question: would this crash on prod? and if yes, how does the stacktrace looks like? is assert the correct method

WireDomain/Sources/WireDomain/AppVersionMigration/AppVersionMigrationService.swift
```diff
@@ -0,0 +1,121 @@
...
+        // Get the last completed migration version.
+        let lastVersion = journal.lastCompletedAppVersionMigration ?? currentVersion
+
+        // Find eligible migrations.
+        var eligibleMigrations = allMigrations
+            .filter { $0.version > lastVersion }
+            .sorted { $0.version > $1.version }
+
+        // Perform each migration.
+        while let nextMigration = eligibleMigrations.popLast() {
```
netbe, line 65: **question:** just to make sure i got it right we'll get the eligibleMigrations like: [4.1, 4.0, 3.124...]
and we popLast so we start with the oldest one, right?

WireDomain/Sources/WireDomain/Notifications/Components/NotificationServiceExtensionFlow.swift
```diff
@@ -35,6 +37,11 @@ final class NotificationServiceExtensionFlow: BootstrapComponent {
         self.contentHandler = contentHandler
 
         let infoDictionary = Bundle.main.infoDictionary
+
+        guard let currentAppVersion = infoDictionary?["CFBundleShortVersionString"] as? String else {
+            throw Failure.missingAppGroupID
```
netbe: issue: don't you mean?
```suggestion
            throw Failure.missingCurrentAppVersion
```

netbe: question: so it means no notifications until user opened the app, right?

netbe: thought: I wonder if we should show a specific notification to the user. Please open the app (it applies also to DB migration)

WireDomain/Sources/WireDomain/AppVersionMigration/AppVersionMigrationService.swift
johnxnguyen: Yes, but the migrations are build time values so we should always catch it before it would ever get to prod. Not sure what the trace looks like... maybe we can see once we wire this thing up.

WireDomain/Tests/WireDomainTests/AppVersionMigration/SemanticVersionTests.swift
```diff
@@ -0,0 +1,49 @@
...
+
+import Foundation
+import Testing
+@testable import WireDomain
+
+struct SemanticVersionTests {
+
+    @Test("initializedWithSensibleDefaults", arguments: [
+        "0",
+        "0.0",
```
netbe: thought: in the past I think we add only two numbers 3.110 and the 0 for patch version was missing. So I wonder if it should be more flexible and return 3.110.0 instead of 0.0.0.

I don't expect to be an issue, as we blacklisted all the old apps, and I guess it would result to be the same, right? back to 0.0.0

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -400,15 +402,15 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+            appVersion: currentBuildVersion,
             environment: environment,
             proxyUsername: proxyCredentials?.username,
             proxyPassword: proxyCredentials?.password,
             reachability: reachability
         )
 
         let authenticatedSessionFactory = AuthenticatedSessionFactory(
-            appVersion: appVersion,
+            appVersion: currentBuildVersion,
```
netbe, line 413: question: it's used for http headers?

netbe (commented):
Looks good, there's just a small issue, the rest are comments to make sure I understood all of this

WireDomain/Sources/WireDomain/AppVersionMigration/AppVersionMigrationService.swift
johnxnguyen: Yes correct

WireDomain/Sources/WireDomain/Notifications/Components/NotificationServiceExtensionFlow.swift
johnxnguyen: yes fixed: 5ca2880a12e75daf30da66cfe08fd99cd85711fd

johnxnguyen: > question: so it means no notifications until user opened the app, right?

How so? The version comes from the bundle

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnxnguyen: I only found this: https://github.com/wireapp/wire-ios/blob/219459fee2fc879cafd2d74865a577cd4d95cbf6/wire-ios-transport/Source/TransportSession/Unauthenticated%20Session/UnauthenticatedTransportSession.swift#L92

WireDomain/Tests/WireDomainTests/AppVersionMigration/SemanticVersionTests.swift
johnxnguyen: actually it does already handle these cases. I've updated the test to include these cases too: a43d9e4c18317c6b82045325853ad52cbfb175d7

netbe (approved):
lgtm;)

PR #2963, opened 2025-05-05
fix: typing indicator not showing, revisit typing users implementation - WPB-17450

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17450" title="WPB-17450" target="_blank">BugWPB-17450</a>  [iOS] Typing indicator doesn't work with new sync
  </td></table>
  <br />
 

### Issue

**Context**: Typing indicator is not showing up when a user is typing in new sync mechanism

**Causes**: in our new code, specifically the local store, we're posting a notification passing the object (self which is the store) then the notification observer is not called because we early exit in `addUnboundedObserver` method from `NotificationInContext` class since the object we use to register the observer and the object we use to post the notification is not the same.

**Solution**: Move the UI update logic part out from the store and repository, create a handler to handle the UI update logic part once the typing users have been processed.

### Testing

UserA on iOS client, userB on another client, send message from userB, check that the typing indicator works properly (shows up and disappears) on userA when userB is typing

### Checklist

### DIFF
--- a/WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
+++ b/WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
@@ -22,6 +22,22 @@ import WireDataModel
 
 public final class ClientSessionComponent {
 
+    public struct ProcessorHandlers {
+        let onProcessedCallEvent: (CallEventInfo) -> Void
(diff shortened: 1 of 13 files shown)


johnxnguyen (approved):

dmitrysimkin (approved):
👍

PR #2964, opened 2025-05-05
fix: CI broken due to images update - WPB-17488

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17488" title="WPB-17488" target="_blank">TaskWPB-17488</a>  [iOS] broken CI due to Cirrus CI images update
  </td></table>
  <br />
 

### Issue

CI broke because Cirrus-CI removed Xcode 15.4 from the image which contained the iOS 17.5 runtime.

As a fix here, we updated to use iOS 18.0 simulator

* Updated snapshot library version
* Update all screenshots

Note: there is an issue in Wire iOS where we don't run the German tests anymore. I'll create a separate ticket for this.

### Testing

* [ ] CI should pass
* [ ] make sure no issue when checking out the branch with git-lfs images


samwyndham:
Closing this in favor of keeping https://github.com/wireapp/wire-ios/pull/2971/files

PR #2965, opened 2025-05-05
chore: review sync logging - WPB-16000

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16000" title="WPB-16000" target="_blank">TaskWPB-16000</a>  [iOS] Review logging of sync mechanism
  </td></table>
  <br />
 

### Issue

This PR improves logging in the new sync mechanism (V2, V3) to measure time each sync phase takes to complete with dedicated tags:
- sync_type: initial vs incremental
- sync_version: v1 (legacy), v2 (without async stream notifs), v3 (with async stream notifs)
- sync_phase: pulling self user, pulling conversations, pulling feature configs..
- duration

Similar logging was introduced for legacy sync (V1) so we can compare durations between sync versions.

Also added some new logs for the new NSE.

### NEW SYNC LOGS (V2)

Screenshot 2025-05-06 at 10 47 15

Screenshot 2025-05-06 at 10 47 32

Screenshot 2025-05-06 at 10 47 52

### LEGACY SYNC LOGS

Screenshot 2025-05-06 at 10 52 21

Screenshot 2025-05-06 at 10 52 39

Screenshot 2025-05-06 at 10 52 59

### Testing

_Describe how to test._

---

### Checklist


wire-ios-sync-engine/Source/Synchronization/Transcoders/ZMLastUpdateEventIDTranscoder.m
```diff
@@ -84,6 +84,8 @@ - (SyncPhase)isSyncing
 - (ZMTransportRequest *)nextRequestIfAllowedForAPIVersion:(APIVersion)apiVersion
 {
     if (self.isSyncing && !self.isDownloadingLastUpdateEventID) {
+        // Slow sync started, resetting tracked time values
+        [self.syncStatus resetSyncTimeTracker];
```
jullianm, line 88: the place we determine the initial sync phase (in SyncStatus) is not **always** the time we actually start to perform the fetch and this can end up in a wrong duration value.

Specifically, this edge case would only occur when we run into the case where we have too many devices: we determine the first slow sync phase and start the slow sync but the app is kind of in an idle state where the first phase is not being actually executed until we removed a device (issue was that the timer would've already started).

So we start the timer only when the first slow sync phase is actually executed to ensure an accurate sync duration in logs

wire-ios-sync-engine/Source/UserSession/SyncTimeTracker.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    func resetStartTime() {
+        phaseStartTime = .now
+    }
+
+    func reset() {
+        phaseStartTime = .now
+        phasesDurations = []
+    }
+}
```
jullianm, line 52: we use this object for legacy sync logging only, it's simpler with new sync code with the usage of async/await

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -50,116 +50,120 @@ public struct IncrementalSync: IncrementalSyncProtocol {
     }
 
     public func perform() async throws -> Token {
-        logger.debug("performing incremental sync")
-        let pushChannel = try await pushChannelAPI.createPushChannel(clientID: selfClientID)
+        try await logger.measureTime(
+            label: "new incremental sync",
+            attributes: .newSyncAttributes(initialSync: false)
+        ) {
+            let pushChannel = try await pushChannelAPI.createPushChannel(clientID: selfClientID)
```
dmitrysimkin: non blocking: so long method and too nested, might be worth splitting into smaller parts

WireLogging/Sources/WireLogging/WireLogger+MessageTime.swift
```diff
@@ -20,12 +20,22 @@ import Foundation
 
 public extension WireLogger {
 
-    func measureTime(label: String, block: () async throws -> Void) async throws {
-        debug("starting \(label)")
+    func measureTime<T>(
+        label: String,
+        attributes: LogAttributes = [:],
+        block: () async throws -> T
```
dmitrysimkin, line 26: question: Will it be run in production code? I mean is this kind of longs we emit no matter what app version\build?

dmitrysimkin (approved):
LGTM, left couple of minor questions

WireDomain/Sources/WireDomain/Extensions/LogAttributes/SyncLogAttributes.swift
```diff
@@ -0,0 +1,121 @@
...
+    ) -> Self {
+        [
+            .syncType: initialSync ? Constants.initial: Constants.incremental,
+            .syncVersion: Constants.v1,
+            .public: true
+        ]
+    }
+
+    static func legacySyncDidFinishAttributes(
+        duration: String,
```
netbe, line 57: question: does duration as a string work for datadog to do operation on it?

WireDomain/Sources/WireDomain/Extensions/LogAttributes/SyncLogAttributes.swift
```diff
@@ -0,0 +1,121 @@
...
+    ) -> Self {
+        [
+            .syncType: initialSync ? Constants.initial: Constants.incremental,
+            .syncVersion: newSyncVersion,
+            .public: true
+        ]
+    }
+
+    static func newSyncPhaseAttributes(
+        _ phase: String,
```
netbe, line 107: question: are newSync phase different from legacy phase?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
netbe: that's a valid point, I wonder if we could  capture self in the Task and just do something like:

```
openPushChannel
pullPendingEvents
processStoredEvents

let task = Task { @Sendable [self] 
 do 
   for envelope in liveEventStream {
     decrypt
     store
     process
     delete
     save
    ....
   }
 
}
```

Not for this PR though

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -50,116 +50,120 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
 
-        logger.debug("opening push channel")
-        let liveEventStream = try await pushChannel.open()
+            logger.debug("opening push channel")
```
netbe: suggestion: pass newSync attributes here v2 for all logger

WireDomain/Sources/WireDomain/Synchronization/PullResourcesSync.swift
```diff
@@ -68,145 +68,221 @@ struct PullResourcesSync: PullResourcesSyncProtocol {
...
     private func pullConversationLabels() async throws {
-        do {
-            logger.debug("pulling conversation labels")
-            try await pullConversationLabelsSync.pull()
-        } catch {
-            throw Failure(resourceName: "pull conversation labels", reason: error)
+        let phase = "pull conversation labels"
+
+        try await logger.measureTime(
+            label: "new sync phase: \(phase)",
```
netbe: issue: I don't think we need to repeat the phase in the label if it's passed in the attributes. I think we would search for "new sync phase" and display the attribute we want

WireLogging/Sources/WireLogging/WireLogger+MessageTime.swift
netbe: we only display info logs in prod

WireLogging/Sources/WireLogging/WireLogger+MessageTime.swift
```diff
@@ -20,12 +20,22 @@ import Foundation
...
+        info(startMessage, attributes: attributes)
         let start = Date.now
-        try await block()
+        let result = try await block()
         let durationInSeconds = start.timeIntervalSinceNow.magnitude
-        debug("completed \(label) in \(String(format: "%.2f", durationInSeconds)) seconds")
+        var updatedAttributes = attributes
+        let formattedDuration = String(format: "%.2f", durationInSeconds)
+        updatedAttributes[.duration] = formattedDuration
+        let completedMessage = "did complete \(label) in \(formattedDuration) seconds"
```
netbe: thought: same here, i wonder if we should add the formattedDuration in the message itself. what do you think @jullianm @dmitrysimkin @johnxnguyen ?

wire-ios-sync-engine/Source/UserSession/SyncStatus.swift
```diff
@@ -328,21 +340,91 @@ public extension SyncStatus {
...
+    }
+}
+
+// MARK: - Logging
+
+extension SyncStatus {
+    
+    /// Logs the initial / incremental sync start
+    private func logSyncStarted() {
+        let message = "did start \(isSlowSyncing ? "legacy initial sync" : "did start legacy incremental sync")"
```
netbe: suggestion: legacy info is in attributes, so no need to repeat it here
```suggestion
        let message = "\(isSlowSyncing ? "initial sync " : "incremental sync") started"
```

wire-ios-sync-engine/Source/UserSession/SyncStatus.swift
```diff
@@ -328,21 +340,91 @@ public extension SyncStatus {
...
+            )
+        )
+    }
+
+    /// Logs the initial or incremental sync completion with the related duration
+    private func logSyncCompleted() {
+        let syncTotalDuration = syncTimeTracker.totalSyncDuration()
+        let formattedDuration = String(format: "%.2f", syncTotalDuration)
+        
+        let message = "did complete \(isSlowSyncing ? "legacy initial sync" : "legacy incremental sync") in \(formattedDuration)"
```
netbe: suggestion: legacy info is in attributes, so no need to repeat it here
```suggestion
        let message = "\(isSlowSyncing ? "initial sync" : "incremental sync") completed"
```

wire-ios-sync-engine/Source/UserSession/SyncStatus.swift
```diff
@@ -328,21 +340,91 @@ public extension SyncStatus {
...
+            )
+        )
+    }
+
+    /// Logs the initial / incremental sync phase starting
+    private func logSyncPhaseStarted(
+        phase: SyncPhase
+    ) {
+        WireLogger.sync.info(
+            "did start legacy sync phase: \(phase.description)",
```
netbe:
```suggestion
            "starting sync phase: \(phase.description)",
```

wire-ios-sync-engine/Source/UserSession/SyncStatus.swift
```diff
@@ -328,21 +340,91 @@ public extension SyncStatus {
...
+        )
+    }
+
+    /// Logs the initial / incremental sync phase completion with the related duration
+    private func logSyncPhaseCompleted(
+        phase: SyncPhase,
+        duration: Double
+    ) {
+        let formattedDuration = String(format: "%.2f", duration)
+        let message = "did complete legacy sync phase: \(phase.description) in \(formattedDuration)"
```
netbe:
```suggestion
        let message = "completed sync phase: \(phase.description)"
```

netbe (changes requested):
overall looks good, but I think some information should be removed from log message as it is contained in the attributes and we log attributes anyway

WireLogging/Sources/WireLogging/WireLogger+MessageTime.swift
dmitrysimkin: Ok so we need these measures in Prod? I'm not sure how much\often this would be called, my concern is just not to slow down production app

WireLogging/Sources/WireLogging/WireLogger+MessageTime.swift
dmitrysimkin: I guess it depends on how often it's called and how heavy this formatting will get in total. is it or would it be called often?

samwyndham:
Removing myself as there are already 2 reviewers. Feel free to add me back

WireDomain/Sources/WireDomain/Synchronization/PullResourcesSync.swift
jullianm: That was my initial thought as well but then I thought this is nice to have this information too when we read the log messages, we don't have to look into the attributes

WireLogging/Sources/WireLogging/WireLogger+MessageTime.swift
jullianm: as mentioned above, I thought this would be nice to have log message with these informations in addition to the attributes, it reads better but that's a personal preference, I don't mind removing this if you think that's overkill

jullianm: @dmitrysimkin the formatting will take place anyways as we need to set the attributes. I think @netbe 's question was more whether we should add the formattedDuration in the message itself (in addition to the attributes)

WireDomain/Sources/WireDomain/Extensions/LogAttributes/SyncLogAttributes.swift
jullianm: no, naming is confusing, I'll rework that

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
jullianm: Yes, I think we could split in smaller functions so it's more clear and also capturing [self] is possible although I think we would have to make the `IncrementalSync` Sendable (which is fine) but also the objects it owns would have to be Sendable (we'll probably have many warnings otherwise and errors on Swift 6). For another PR though

WireDomain/Sources/WireDomain/Extensions/LogAttributes/SyncLogAttributes.swift
jullianm: https://github.com/wireapp/wire-ios/pull/2965/commits/c34cf5e6c6115b0517e4422837745e09f5ded47d I removed the prefix `new` which I think is not necessary

wire-ios-sync-engine/Source/UserSession/SyncStatus.swift
jullianm: https://github.com/wireapp/wire-ios/pull/2965/commits/c34cf5e6c6115b0517e4422837745e09f5ded47d

wire-ios-sync-engine/Source/UserSession/SyncStatus.swift
jullianm: https://github.com/wireapp/wire-ios/pull/2965/commits/c34cf5e6c6115b0517e4422837745e09f5ded47d

wire-ios-sync-engine/Source/UserSession/SyncStatus.swift
jullianm: https://github.com/wireapp/wire-ios/pull/2965/commits/c34cf5e6c6115b0517e4422837745e09f5ded47d

wire-ios-sync-engine/Source/UserSession/SyncStatus.swift
jullianm: https://github.com/wireapp/wire-ios/pull/2965/commits/c34cf5e6c6115b0517e4422837745e09f5ded47d

WireDomain/Sources/WireDomain/Synchronization/PullResourcesSync.swift
jullianm: removed this extra information in message as we already have it in attributes https://github.com/wireapp/wire-ios/pull/2965/commits/80b3bb9bcc8080fa89bd6f478506f52e6e6df061

WireLogging/Sources/WireLogging/WireLogger+MessageTime.swift
jullianm: removed this extra information in message as we already have it in attributes https://github.com/wireapp/wire-ios/pull/2965/commits/80b3bb9bcc8080fa89bd6f478506f52e6e6df061

WireDomain/Sources/WireDomain/Extensions/LogAttributes/SyncLogAttributes.swift
jullianm: Not sure

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
jullianm: https://github.com/wireapp/wire-ios/pull/2965/commits/9d244c2ecf42f53e9e3be0961a90f3c94db023cd

WireLogging/Sources/WireLogging/WireLogger+MessageTime.swift
netbe: we don't have the distinction between internal apps and prod apps at the moment. but ideally we would only need this in internal apps (datadog), i think it's fine for now

netbe (approved):
looks good

PR #2966, opened 2025-05-05
refactor: wait for decryption to finish before sending a message - WPB-16348

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16348" title="WPB-16348" target="_blank">TaskWPB-16348</a>  [iOS] Update logic to new sync code for sending
  </td></table>
  <br />
 

### Issue
Before sending a message (and therefore encrypting), we should ensure we've received and decrypted all incoming messages. The current code only waits for decryption in the legacy sync. This PR updates it to consider also the new sync, if it's enabled.

Changes include:
- Move `QuickSyncObserver` to `WireSyncEngine`, now it depends on `SyncAgent`.
- Introduce tracking of `SyncState` to identify which part of the sync we have reached
- Delay creation of request strategies that depending on `MessageSender`, which now has a dependency on `SyncAgent`. These strategies shouldn't exist before the is a registered client anyway.

### Testing
1. Be logged in, have app in the background, accumulate lots of messages.
2. Open app, sync starts, immediately try to send a message
3. Assert it doesn't send until the decryption is finished (not directly observable from the app, but will be a before the sync completes).

---

### Checklist

---


wire-ios-request-strategy/Sources/Request Strategies/Client Message/ClientMessageRequestStrategy.swift
```diff
@@ -42,7 +42,6 @@ public class ClientMessageRequestStrategy: NSObject, ZMContextChangeTrackerSourc
     public init(
         context: NSManagedObjectContext,
         localNotificationDispatcher: PushMessageHandler,
-        applicationStatus: ApplicationStatus,
```
johnxnguyen, line 45: dead code

wire-ios-sync-engine/Source/Synchronization/Strategies/CallingRequestStrategy.swift
```diff
@@ -81,13 +80,15 @@ public final class CallingRequestStrategy: AbstractRequestStrategy, ZMSingleRequ
...
 
         if let clientId = selfUser.selfClient()?.remoteIdentifier {
-            self.callCenter = WireCallCenterV3Factory.callCenter(
-                withUserId: selfUser.avsIdentifier,
-                clientId: clientId,
-                uiMOC: managedObjectContext.zm_userInterface,
-                flowManager: flowManager,
-                transport: self
-            )
+            managedObjectContext.zm_userInterface.performAndWait {
```
johnxnguyen, line 83: to prevent core data threading violations

wire-ios-sync-engine/Source/Synchronization/StrategyDirectory.swift
```diff
@@ -413,6 +354,84 @@ public class StrategyDirectory: NSObject, StrategyDirectoryProtocol {
         ]
     }
 
+    func makeRemainingStategies(
```
johnxnguyen: This is not very elegant but I would prefer to improve it in a separate PR.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -599,14 +599,33 @@ public final class ZMUserSession: NSObject {
...
         syncAgent.resume()
+
+        // Finish setting up the final strategies.
+        if
+            let strategyDirectory = strategyDirectory as? StrategyDirectory,
+            let localNotificationDispatcher {
+            let quickSyncObserver = QuickSyncObserver(
+                syncAgent: syncAgent,
+                notificationContext: notificationContext
+            )
```
jullianm, line 597: question: is there a slight chance that between the time `syncAgent.resume()` is called and `QuickSyncObserver.setupObservation` method is called, the legacy sync is already finished and the notification posted but the NotificationCenter publisher stream would not be triggered thus the `decryptionState` not changed and the message sender waits forever ?

I'm specifically asking about v1 because (I think) the NotificationCenter built-in publisher only listens to notifications from the point it is subscribed onward, it doesn't buffer or replay past notifications. 
Which is fine for V2 because the `CurrentValueSubject` would replay the last stored value (`.done`) if that edge case would happen.

jullianm (approved):
Looks good! Left a (non blocking) question

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: I'm not sure, but could move the resume call afterwards just in case.

wire-ios-data-model/Support/Sourcery/generated/AutoMockable.generated.swift
```diff
@@ -1,4 +1,4 @@
-// Generated using Sourcery 2.2.6 — https://github.com/krzysztofzablocki/Sourcery
+// Generated using Sourcery 2.2.4 — https://github.com/krzysztofzablocki/Sourcery
```
dmitrysimkin, line 1: I think we move mostly to 2.2.6 version?

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
```diff
@@ -45,27 +45,24 @@ public final class MessageSender: MessageSenderInterface {
 
     public init(
         apiProvider: APIProviderInterface,
-        clientRegistrationDelegate: ClientRegistrationDelegate,
         sessionEstablisher: SessionEstablisherInterface,
         messageDependencyResolver: MessageDependencyResolverInterface,
-        quickSyncObserver: QuickSyncObserverInterface,
-        context: NSManagedObjectContext
+        context: NSManagedObjectContext,
+        quickSyncObserver: QuickSyncObserverInterface
```
netbe: thought: maybe quickSyncObserver does not make a lot of sense as a name. Do we have an exact mapping:
InitialSync -> SlowSync
IncrementalSync -> QuickSync

at some point we should align these terms as they are used in documentation too

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
```diff
@@ -76,7 +73,7 @@ public final class MessageSender: MessageSenderInterface {
         let logAttributes = await logAttributesBuilder.logAttributes(message)
         WireLogger.messaging.debug("broadcast message", attributes: logAttributes)
 
-        await quickSyncObserver.waitForDecryptionOfEventsToFinish()
+        await quickSyncObserver.waitUntilCanSendMessage()
```
netbe: praise: clearer than before

wire-ios-sync-engine/Source/UserSession/SyncStatus.swift
```diff
@@ -61,6 +61,13 @@ public class SyncStatus: NSObject, SyncStatusProtocol, SyncProgress {
 
     var quickSyncContinuation: CheckedContinuation<Void, Never>?
 
+    public var isLive: Bool {
+        managedObjectContext.performAndWait {
+            guard !isSyncV2Enabled else { return false }
```
dmitrysimkin: question: can guard be outside of perform block?

wire-ios-share-engine/Sources/SharingSession.swift
```diff
@@ -158,6 +158,7 @@ struct SyncStatus: SyncStatusProtocol {
     func resyncResources() {}
     func forceSlowSync() {}
     func recoverWithQuickSync() async {}
+    var isLive: Bool = false
```
netbe, line 161: question: is this syncStatus useful here? as it was mainly for mlsService recovery actions but is now obsolete with the delegate

dmitrysimkin (approved):
LGTM, just couple of questions

wire-ios-data-model/Support/Sourcery/generated/AutoMockable.generated.swift
johnxnguyen: hmm I wonder where this version comes from because I have 2.2.6 locally... but I guess it's the plugin version not what I have installed on my machine.

wire-ios-sync-engine/Source/Synchronization/StrategyDirectory.swift
netbe: suggestion: add a comment to which kind of strategies we're talking about, maybe:
```suggestion
    func makeClientRelatedStrategies(
```

wire-ios-sync-engine/Source/UserSession/SyncStatus.swift
johnxnguyen: yes it can, done: 42d5ebb23a4f5e9236452de9e42bf99d8e8d0c81

wire-ios-sync-engine/Tests/Source/Synchronization/QuickSyncObserverTests.swift
```diff
@@ -0,0 +1,115 @@
...
+            )
+        }
+
+        // Then
+        let before = Date.now
+        await confirmation("sync is done within 500ms") { confirm in
+            // When
+            await sut.waitUntilCanSendMessage()
+            #expect(Date.now.timeIntervalSince(before) < 0.5, "sync duration > 500ms")
+            confirm()
```
netbe: question: where is confirm define?

wire-ios-sync-engine/Source/Synchronization/Strategies/CallingRequestStrategy.swift
netbe: question: and this is called from sync context or swift concurrency pool?

netbe (approved):
no blocker, but left some suggestions and questions to understand correctly

wire-ios-share-engine/Sources/SharingSession.swift
johnxnguyen: it's completely dead in the share extension, it does nothing, but for some reason has to be there because of architectural issues.

wire-ios-sync-engine/Tests/Source/Synchronization/QuickSyncObserverTests.swift
johnxnguyen: it's passed into the closure, part of the Confirmation api

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
johnxnguyen: I'll rename it to IncrementalSyncObserver

johnxnguyen: done 113cd93e7e6ee4100cfd98aaa87d3c4433b73b25

PR #2971, opened 2025-05-05
fix: CI broken due to images update - WPB-17488

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17488" title="WPB-17488" target="_blank">TaskWPB-17488</a>  [iOS] broken CI due to Cirrus CI images update
  </td></table>
  <br />
 

### Issue

CI broke because Cirrus-CI removed Xcode 15.4 from the image which contained the iOS 17.5 runtime.

As a fix here, we updated to use iOS 18.0 simulator

* Updated snapshot library version
* Update all screenshots

Now the annoying thing is why we needed to update the shapshots... Many of them render awfully on iOS 18.x simulator:

| iOS 17.5 | iOS 18.0 |
| --- | --- |
| old-available dark |  new-available dark |

This is due to this issue: https://github.com/pointfreeco/swift-snapshot-testing/issues/790

But I guess we don't want to be stuck on iOS 17.5 forever so we should update.

Note: there is an issue in Wire iOS where we don't run the German tests anymore. @netbe has create a separate ticket for this.

### Testing

* [ ] CI should pass
* [ ] make sure no issue when checking out the branch with git-lfs images


wire-ios-data-model/Source/Public/ZMUser.h
netbe, line 1: suggestion: this can be removed

netbe (approved):
let's see if this works

dmitrysimkin (approved):

wire-ios-data-model/Source/Public/ZMUser.h
samwyndham: Oh I didn't see this comment. I will remove this in a separate PR if the current test run passes otherwise push a fix.

PR #2972, opened 2025-05-06
refactor: send pings to keep push channel alive - WPB-17498

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17498" title="WPB-17498" target="_blank">BugWPB-17498</a>  [iOS] Push channel dies after some time
  </td></table>
  <br />
 

### Issue
The backend will close the push channel if the client doesn't periodically send a ping. This PR adds sending a ping every 30 seconds (same as the legacy push channel) to the new push channel implementation.

### Testing
1. Open the app and leave it open for a while (maybe 10 or 20 minutes) with no activity, but don't let it go in the background.
2. Assert after the long pause, try to receive a message and assert that you can receive the new message.

---

### Checklist

---


WireAPI/Sources/WireAPI/Network/PushChannel/PushChannel.swift
```diff
@@ -55,11 +68,51 @@ public final class PushChannel: PushChannelProtocol {
...
+                WireLogger.pushChannel.warn("keep alive task was cancelled")
+                tearDownKeepAliveTask()
+            }
+        }
+    }
+
+    private func sendKeepAlivePing() async {
+        do {
+            WireLogger.pushChannel.debug("sending keep alive ping")
+            try await webSocket.write(Data())
```
typfel: So you send an empty payload? Any reason why you don't send a real ping message

https://developer.apple.com/documentation/foundation/urlsessionwebsockettask/sendping(pongreceivehandler:)

dmitrysimkin (approved):
👍

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannel.swift
johnxnguyen: I was doing the same thing that we do in the legacy push channel: https://github.com/wireapp/wire-ios/blob/4f2610096998575a9bf041123104d31655c74385/wire-ios-transport/Source/PushChannel/StarscreamPushChannel.swift#L242

johnxnguyen: But I see now that actually does call the ping method you linked too, I'll update it

netbe: @johnxnguyen do you want to try @typfel suggestion or defer to later?

netbe (approved):
looks good, just wondering if suggestion makes sense to do it now or later, up to you

WireAPI/Sources/WireAPI/Network/PushChannel/PushChannel.swift
johnxnguyen: fixed 67c5e4d5dce9da97b8e6563da4be19d95c99043b

PR #2975, opened 2025-05-06
feat: prepare UI for message attachment carousel - WPB-17536

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17536" title="WPB-17536" target="_blank">TaskWPB-17536</a>  [iOS] Prepare for message attachment carousel
  </td></table>
  <br />
 

### Issue

With wire cells there will be a new message type that includes text and many files. When composing this type of message the attachments will be displayed in the message compose area in a carousel. Something like:

Screenshot 2025-05-06 at 17 13 24

This PR is just prep for adding the carousel which includes:

- Pinning the markdown (Aa) button to the top of the message area. (currently it is pinned to the bottom)
- Pinning the send button to the top of the message area. (currently it is pinned to the middle).
- Add a temporary placeholder view that will become the carousel and is visible when the `wireCells` feature flag is enabled. This placeholder is added to a new UIStackView that means we can easily hide/show it with animation when attachments are added / deleted.

There is also a small (accidental) fix for a slight vertical misalignment of the markdown button. This is what caused the need for the snapshot test updates.

See the new behavior in this video:

https://github.com/user-attachments/assets/12a7cb8c-5969-4d9e-bfad-0e707bc4f768

### Testing

**Please test on both iPhone & iPad**

1. With the `wireCells` feature flag disabled, play around with writing messaged ensuring that the behavior is the same as the current app accept that the markdown and send buttons stay pinned to the top of the message compose area.
2. Repeat the above with `wireCells` feature flag enabled. The behavior should be the same but there should be a large red rectangle placeholder where the attachments carousel will eventually be.

---

### Checklist


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController/ConversationInputBarViewController.swift
```diff
@@ -1130,11 +1127,8 @@ extension ConversationInputBarViewController: UIGestureRecognizerDelegate {
             ephemeralIndicatorButton.widthAnchor.constraint(equalToConstant: InputBar.rightIconSize),
             ephemeralIndicatorButton.heightAnchor.constraint(equalToConstant: InputBar.rightIconSize),
 
-            markdownButton.centerXAnchor.constraint(equalTo: markdownButton.superview!.centerXAnchor),
-            markdownButton.bottomAnchor.constraint(equalTo: markdownButton.superview!.bottomAnchor, constant: -14),
```
samwyndham, line 1134: This positioning is set in the `InputBar` and doesn't rely on superview magic.

dmitrysimkin (approved):
👍

KaterinaWire:
@samwyndham I would recommend testing it on an iPad too. We had problems with these buttons on the iPad in the past.

KaterinaWire (approved):

PR #2977, opened 2025-05-07
fix: logout when access token request fails - WPB-12035

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-12035" title="WPB-12035" target="_blank">TaskWPB-12035</a>  [iOS] Handle invalid credentials errors
  </td></table>
  <br />
 

### Issue

In legacy code, when access token request fails, we clean up token and cookie and we logout in `ZMAccessTokenHandler` class.
This PR introduces this behavior for new sync code, in `AuthenticationManager` class.

### Testing

In debug actions, hit "forces logout": it will mimic the behavior of failing to get the access token, triggers the completion handler and logout the user session

### Checklist


WireAPI/Sources/WireAPI/Assembly.swift
```diff
@@ -79,7 +79,8 @@ public final class Assembly {
     public lazy var authenticationManager: some AuthenticationManagerProtocol = AuthenticationManager(
         clientID: clientID,
         cookieStorage: cookieStorage,
-        networkService: apiNetworkService
+        networkService: apiNetworkService,
+        onAuthenticationFailure: {}
```
jullianm, line 83: this is used from the `AuthenticatedSessionFactory` to build the `ZMUserSession` which is not needed at this point (and could probably be removed after a bit of refactoring) since we initialize again an `AuthenticationManager` in `ClientSessionComponent` when setting up the sync.

WireAPI/Sources/WireAPI/Authentication/AuthenticationManager.swift
```diff
@@ -108,7 +112,14 @@ public actor AuthenticationManager: AuthenticationManagerProtocol {
...
         } catch {
+            WireLogger.authentication.error(
+                "Failed to renew access token with error: \(error.localizedDescription)"
+            )
+
+            try await cookieStorage.removeCookies()
             currentToken = nil
+            onAuthenticationFailure()
+
             throw error
```
jullianm, line 134: note: the `onAuthenticationFailure` handler will proceed to logout the user so throwing the error here (which is currently required by the API) is probably useless thus the whole class could probably revisited as well (in another PR because I don't think it's a blocker for now?) to remove the thrown errors from the public APIs.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -590,7 +591,25 @@ public final class ZMUserSession: NSObject {
         syncAgent.resume()
     }
 
-    func onProcessedTypingUsers(
+    // MARK: - Callbacks from WireDomain
+
+    @Sendable
+    public func onAuthenticationFailure() {
```
jullianm, line 621: had to make it public for the debug action to access it

dmitrysimkin (approved):
👍

WireAPI/Sources/WireAPI/Authentication/AuthenticationManager.swift
```diff
@@ -108,7 +112,14 @@ public actor AuthenticationManager: AuthenticationManagerProtocol {
             currentToken = .cached(newToken)
             return newToken
         } catch {
+            WireLogger.authentication.error(
+                "Failed to renew access token with error: \(error.localizedDescription)"
+            )
+
+            try await cookieStorage.removeCookies()
             currentToken = nil
+            onAuthenticationFailure()
```
samwyndham: issue: I think we need to somehow have more granular error handling here. For example, I assume an error will occur if the user has no internet - we don't want to remove the cookie and log out in that case right?

samwyndham (changes requested):

WireAPI/Sources/WireAPI/Authentication/AuthenticationManager.swift
johnxnguyen: Maybe if you catch the specific errors for invalid access tokens, you don't need to rethrow given they're handled by logging out the user.

wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DebugActions/DeveloperDebugActionsViewModel.swift
```diff
@@ -78,7 +80,7 @@ final class DeveloperDebugActionsViewModel: ObservableObject {
             .init(title: "Update MLS migration status", action: updateMLSMigrationStatus),
             .init(title: "Delete domains in the database", action: deleteDomains),
             .init(title: "Find Conversation with MLS Group", action: showSearchMLSConversations),
-            .init(title: "Clear access token & cookie (forces logout)", action: clearAccessTokenAndCookie)
+            .init(title: "Force logout", action: forceLogout)
```
johnxnguyen: suggestion: "Simulate access token failure"

WireAPI/Sources/WireAPI/Authentication/AuthenticationManager.swift
jullianm: yes, it makes sense. So in commit https://github.com/wireapp/wire-ios/pull/2977/commits/eb61e37fb6d30fd79acdce8d9b9bf35669901217 we're specifically catching the `invalidCredentials` error, removing the cookies and logging out because I don't think we can't recover from that particular error.  

for the other error types, we simply throw, I'm not sure if there are particular cases to handle where we should logout except the 403 one (that's the only one mentioned in Swagger)

wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DebugActions/DeveloperDebugActionsViewModel.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/2977/commits/eb61e37fb6d30fd79acdce8d9b9bf35669901217

samwyndham (approved):
LGTM. Nice work!

PR #2982, opened 2025-05-07
fix: channel icon - WPB-17556

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17556" title="WPB-17556" target="_blank">BugWPB-17556</a>  [iOS] fix channel icon color
  </td></table>
  <br />
 

### Issue

This fixes channel icon to be the same with other platform. `UUID().uuidstring` gives capitalized string, so it needs to be lowercased.


jullianm (approved):

dmitrysimkin (approved):
👍

PR #2991, opened 2025-05-12
fix: channel icon issue - WPB-17589

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17589" title="WPB-17589" target="_blank">BugWPB-17589</a>  [iOS] Self-created channel icon shows as group on client it is created by
  </td></table>
  <br />
 

### Issue

**Context**: After creating a channel, the conversation shows up in the list like it is a group conversation instead of a channel (it is noticeable looking at the icon). When performing an initial sync (or triggering resources sync in debug action) the conversations shows up normally, as a channel. 

**Causes**: Upon creation of the channel by the user, it is stored locally in the db (as a channel) then immediately followed by a `conversation.create` event that we're decoding **without** decoding the groupType property (which is an omission) so the groupType is wrongly mapped and the conversation updated locally as not being a channel anymore.

**Solution**: Decode the missing groupType property in the conversation create event decoder.

### Testing

channels should all properly show up as channels (with the right icon + the lock icon (since we only have private channels currently)

### Checklist

### DIFF
--- a/WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationCreateEventProcessor.swift
+++ b/WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationCreateEventProcessor.swift
@@ -25,7 +25,6 @@ struct ConversationCreateEventProcessor: ConversationCreateEventProcessorProtoco
     let repository: any ConversationRepositoryProtocol
 
     func processEvent(_ event: ConversationCreateEvent) async {
-        let conversationID = event.conversationID
         let conversation = event.conversation
         let timestamp = event.timestamp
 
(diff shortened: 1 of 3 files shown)


WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationCreateEventProcessor.swift
```diff
@@ -25,7 +25,6 @@ struct ConversationCreateEventProcessor: ConversationCreateEventProcessorProtoco
     let repository: any ConversationRepositoryProtocol
 
     func processEvent(_ event: ConversationCreateEvent) async {
-        let conversationID = event.conversationID
```
jullianm, line 28: unused code

samwyndham (approved):
Nice! Happy this was a small fix.  Am I correct that this bug only affected new sync?

dmitrysimkin (approved):
👍

PR #2994, opened 2025-05-12
fix: event processor is missing some event consumers - WPB-17587

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17587" title="WPB-17587" target="_blank">BugWPB-17587</a>  [iOS] Missing messages 3.123 to 3.124 - Sync refactor 3.124
  </td></table>
  <br />
 

### Issue
Sometimes the app is not able to receive messages, but is able to send messages. The reason is that we made a change in https://github.com/wireapp/wire-ios/pull/2966  that sets up some request strategies only when there is a registered client, which happens at a later time. The request strategies are also event consumers, so after creating the remaining strategies (i.e consumers) they are added to the strategy directory but the `EventProcessor` has already been created and is not updated. Therefore, it had only a partial list of event consumers. As a result, when some events we received, such as a new message event, there was no consumer available to process it and it looks like the message is missing.

The solution is to ensure the `EventProcessor` always has the full list of event consumers. Due to some limitations of the legacy architecture, I had to resort to non optimal solution of injecting in the strategy directory (which has the updated event consumers) into the event processor. I wanted to delay creation of the event processor, but it's needed early to create the operation loop. Perhaps this is something we can revisit.

### Testing
1. Log in
2. Receive messages
3. Assert that you see them.

---

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
+++ b/wire-ios-sync-engine/Source/Synchronization/EventProcessor.swift
@@ -32,10 +32,18 @@ actor EventProcessor: UpdateEventProcessor {
     private let eventProcessingTracker: EventProcessingTrackerProtocol
     private let earService: EARServiceInterface
     private var processingTask: Task<Void, Error>?
-    private let eventConsumers: [ZMEventConsumer]
-    private let eventAsyncConsumers: [ZMEventAsyncConsumer]
-
+    private let strategyDirectory: any StrategyDirectoryProtocol
     private let processedEventList = ProcessedEventList()
+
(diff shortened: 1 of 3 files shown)


netbe (approved):
sounds like a good solution for now, any tests need to be updated? Also I guess the EventProcessor will go away after rolling out the new sync right ? @johnxnguyen

johnxnguyen:
@netbe yeah I need to update some tests. The old event processor will go away after give everyone access to the new sync, not just v8 users.

jullianm (approved):

PR #2995, opened 2025-05-12
fix: update event migration missed member removed events - WPB-17588

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17588" title="WPB-17588" target="_blank">BugWPB-17588</a>  [iOS] Migration from legacy sync to new sync missed member removed events
  </td></table>
  <br />
 

### Issue
When updating to the new sync and there are legacy member leave events in the database, after migration the conversation still see the member that is no longer a participant. This is because we didn't migrate the member leave event (and was instead relying on the initial sync to update the conversation participant list, which doesn't work). The solution is to also migrate the legacy member leave event to the new member leave event.

### Testing
1. Be logged in on a previous version (e.g 3.123)
2. Be in a conversation
3. Put the app in the background
4. Another user should remove a participant in the conversation.
5. Update to 3.124
6. Assert that the participant has been removed.

---

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/SessionManager/UpdateEventMigrator.swift
+++ b/wire-ios-sync-engine/Source/SessionManager/UpdateEventMigrator.swift
@@ -71,7 +71,12 @@ struct UpdateEventMigrator {
                         localDomain: localDomain
                     )
                 } catch {
-                    WireLogger.sync.error("failed to map legacy event, skipping... reason: \(String(describing: error))")
+                    WireLogger.sync.error(
+                        """
+                        failed to map legacy event, skipping... \
+                        reason: \(String(describing: error))
+                        """
+                    )
                     continue
                 }
 
--- a/wire-ios-sync-engine/Source/SessionManager/UpdateEventMigrator.swift
+++ b/wire-ios-sync-engine/Source/SessionManager/UpdateEventMigrator.swift
@@ -64,10 +64,18 @@ struct UpdateEventMigrator {
 
             for legacyEvent in legacyEvents {
                 // Map it to the new event model.
-                guard let newUpdateEvent = UpdateEvent(
-                    legacyEvent: legacyEvent,
-                    localDomain: localDomain
-                ) else {
+                let newUpdateEvent: UpdateEvent?
+                do {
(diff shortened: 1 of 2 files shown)


wire-ios-sync-engine/Source/SessionManager/UpdateEventMigrator.swift
```diff
@@ -210,6 +220,62 @@ private extension UpdateEvent {
         )
     }
 
+    private static func coversationMemberLeaveEvent(
```
netbe:
```suggestion
    private static func conversationMemberLeaveEvent(
```

wire-ios-sync-engine/Source/SessionManager/UpdateEventMigrator.swift
```diff
@@ -210,6 +220,62 @@ private extension UpdateEvent {
...
+            let payload = try? decoder.decode(
+                Payload.ConversationEvent<Payload.UpdateConversationMemberLeave>.self,
+                from: event.payload
+            ),
+            let conversationID = payload.conversationID(localDomain: localDomain),
+            let senderID = payload.senderID(localDomain: localDomain),
+            let timestamp = payload.timestamp,
+            let leaveReason = payload.data.reason
+        else {
+            return nil
```
netbe: question: is there a way to know about the error at some level?

netbe (approved):
left some comments, would be nice to add some unit tests too, but that could be differ to later to speed things up for release

wire-ios-sync-engine/Source/SessionManager/UpdateEventMigrator.swift
johnxnguyen: We could catch a decoding error throw it up, or we could throw a `missingData` error for the rest?

johnxnguyen: done 698e84594df52360b3650adda0c2d03795aa99d3

jullianm (approved):
LGTM

PR #3014, opened 2025-05-14
feat: batch decryption - WPB-17614

typfel:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17614" title="WPB-17614" target="_blank">TaskWPB-17614</a>  [iOS] : Batch Decryption
  </td></table>
  <br />
 

### Issue

Decrypt events in batches when pulling events.

This accomplished by adding an optional `context` argument to all decryption functions. If omitted the decryption methods behave as before by opening a separate transaction for each single operation, but if it's provided the decryption operation will happen on an existing **transaction context**.

Currently only the `PullPendingUpdateEventsSync` use the `context` argument. We would also need to add batch decryption to the async notifications code path once it's available.

### Testing

App is still processing events without errors.

### Non-scientific benchmark

Running on a simulator I've collected roughly 500 text messages and measured how long it took for `PullPendingUpdateEventsSync` to decrypt them all.

#### using batch decryption
```
It took 967.583625 ms to decrypt 504 event(s)[process_name:Wire][account_id:9e98572***][self_client_id:29c8014***][process_id:70968]
```

#### without batch decryption
```
It took 10356.630792 ms to decrypt 504 event(s)[process_id:71884][self_client_id:29c8014***][process_name:Wire][account_id:9e98572***]
```

---

### Checklist

---


WireDomain/Sources/WireDomain/Synchronization/PullPendingUpdateEventsSync.swift
```diff
@@ -78,30 +79,46 @@ public struct PullPendingUpdateEventsSync: PullPendingUpdateEventsSyncProtocol {
...
-                try await store.persistEventEnvelope(
-                    decryptedEnvelope,
-                    index: currentIndex
-                )
+                var lastEnvelopeID: UUID?
+                var decryptedEnvelopes: [UpdateEventEnvelope] = []
 
-                events.append(contentsOf: decryptedEvents)
+                for envelope in envelopes {
+                    count += 1
```
caldrian, line 85: I don't see the `count` variable being read after this line. Should it be used to increment `currentIndex`?

typfel: It's read in the `log()` function scoped to this function. `currentIndex` doesn't need to updated since we are fetching it from the repository.

WireDomain/Sources/WireDomain/Synchronization/PullPendingUpdateEventsSync.swift
```diff
@@ -78,30 +79,46 @@ public struct PullPendingUpdateEventsSync: PullPendingUpdateEventsSyncProtocol {
                 )
             }
 
-            for envelope in envelopes {
-                count += 1
+            // We'll insert new events from this index.
+            var currentIndex = try await store.indexOfLastEventEnvelope() + 1
```
caldrian:
```suggestion
            let currentIndex = try await store.indexOfLastEventEnvelope() + 1
```

caldrian (approved):

KaterinaWire:
@typfel I think we should target 3.125 release branch

.github/workflows/test_develop.yml
```diff
@@ -14,7 +14,7 @@ jobs:
   trigger_tests:
     strategy:
       matrix:
-        branch: [develop, release/cycle-3.120, release/cycle-3.123, release/cycle-3.124]  # List other branches here
+        branch: [develop, release/cycle-3.120, release/cycle-3.123, release/cycle-3.124, release/cycle-3.125]  # List other branches here
```
netbe: issue: hmm weird this shows up here, maybe because it was develop before

typfel: Should it be removed then?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -92,7 +92,7 @@ public struct IncrementalSync: IncrementalSyncProtocol {
                             "decrypting live event envelope",
                             attributes: [.eventEnvelopeID: envelope.id]
                         )
-                        envelope.events = try await decryptor.decryptEvents(in: envelope)
+                        envelope.events = try await decryptor.decryptEvents(in: envelope, context: nil)
```
netbe, line 95: **question:** so here for live events we don't batch right? we could do the same as suggested for async notifs, but as it's going to be replaced I guess it's ok for now

wire-ios-data-model/Source/MLS/MLSActionExecutor.swift
```diff
@@ -203,13 +207,24 @@ public actor MLSActionExecutor: MLSActionExecutorProtocol {
...
+        if let context {
+            try await processWelcomeMessageInternal(message, context: context)
+        } else {
+            try await coreCrypto.perform { context in
+                try await self.processWelcomeMessageInternal(message, context: context)
+            }
         }
+    }
+
+    private func processWelcomeMessageInternal(
```
netbe, line 220: **nit-pic:**
```suggestion
    private func internalProcessWelcomeMessage(
```

netbe: **question:** actually isn't it possible to name it `processWelcomeMessage` ?

wire-ios-data-model/Source/MLS/MLSActionExecutor.swift
```diff
@@ -338,29 +353,39 @@ public actor MLSActionExecutor: MLSActionExecutorProtocol {
...
                 }
             }
         }
+    }
 
-        if let result {
-            return result
-        } else {
-            throw Failure.bufferedDecryptedMessage
+    private func decryptMessageInternal(
```
netbe, line 372:
```suggestion
    private func decryptMessage(
```

wire-ios/Wire-iOS/Resources/Configuration/Version.xcconfig
```diff
@@ -16,5 +16,5 @@
 // along with this program. If not, see http://www.gnu.org/licenses/.
 //
 
-WIRE_SHORT_VERSION = 3.125.0
+WIRE_SHORT_VERSION = 3.126.0
```
netbe: issue: this should be reverted

netbe (changes requested):
looks good, left some comments. Asking for changes since some app version is changed and will cause issue for 3.125 (I guess it's when changing from develop to release branch)

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
typfel: I guess we could in theory, but I guess the impact is less since you usually don't get a big burst of messages at once.

wire-ios-data-model/Source/MLS/MLSActionExecutor.swift
typfel: since the signature is different you mean?

typfel: They would have the same signature so no I don't think I can do it.

.github/workflows/test_develop.yml
typfel: 💥

wire-ios/Wire-iOS/Resources/Configuration/Version.xcconfig
typfel: 💥

netbe (approved):
looks good

PR #3015, opened 2025-05-14
fix: incremental sync failure - WPB-17645

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17645" title="WPB-17645" target="_blank">BugWPB-17645</a>  [iOS] Incrementalsync did not start after migration
  </td></table>
  <br />
 

### Issue

Introduces backoff retrier to retry sync a couple of times (with exponential backoff) when it fails
After reaching the max attempts, it will throw the error and stop retry syncing

### Testing

N/A

### Checklist

### DIFF
--- a/WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetrier.swift
+++ b/WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetrier.swift
@@ -92,10 +92,15 @@ public actor BackoffRetrier {
 
         let queue = DispatchQueue(label: "BackoffRetrier")
 
-        monitor.pathUpdateHandler = { [weak self] path in
-            if path.status == .satisfied {
+        monitor.pathUpdateHandler = { [weak self] newPath in
+            guard let self else { return }
+
+            let currentPath = monitor.currentPath
+            let didRetrieveConnection = currentPath.status != .satisfied && newPath.status == .satisfied
+
+            if didRetrieveConnection {
                 Task {
-                    await self?.reset()
+                    await reset()
                 }
             }
         }
--- a/WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetrier.swift
+++ b/WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetrier.swift
@@ -18,18 +18,18 @@
 
 
 import Foundation
-import UIKit
 import Network
+import UIKit
 
 public actor BackoffRetrier {
     public typealias SleepFunction = @Sendable (
         _ seconds: Double
     ) async throws -> Void
-    
+
     enum Failure: Error {
         case exceededMaxAttempts(latestError: any Error)
     }
-    
+
     private let policy: BackoffRetryPolicy
     private let monitor = NWPathMonitor()
     private let sleep: SleepFunction
@@ -43,7 +43,7 @@ public actor BackoffRetrier {
     ) {
         self.policy = policy
         self.sleep = sleep
-        
+
         setupObservers()
     }
 
@@ -51,7 +51,7 @@ public actor BackoffRetrier {
         monitor.cancel()
         NotificationCenter.default.removeObserver(self)
     }
-    
+
     public func retry<T: Sendable>(
         _ operation: @escaping () async throws -> T
     ) async throws -> T {
@@ -60,27 +60,27 @@ public actor BackoffRetrier {
             do {
                 return try await operation()
             } catch {
-                
+
                 attempt += 1
-                
+
                 guard attempt < policy.maxRetries else {
                     throw Failure.exceededMaxAttempts(latestError: error)
                 }
-                
+
                 // Exponential backoffs
                 var delay = policy.baseTime * Double(pow(policy.exponentMultiplier, Double(attempt)))
-                
+
                 if policy.jitter {
                     // Adds jitter (randomness)
-                    delay = Double.random(in: 0...min(policy.maxTime, delay))
+                    delay = Double.random(in: 0 ... min(policy.maxTime, delay))
                 }
-                
+
                 try await sleep(delay)
             }
         }
     }
 
-    nonisolated private func setupObservers() {
+    private nonisolated func setupObservers() {
         NotificationCenter.default.addObserver(
             forName: UIApplication.willEnterForegroundNotification,
             object: nil,
@@ -90,9 +90,9 @@ public actor BackoffRetrier {
                 await self?.reset()
             }
         }
-        
+
         let queue = DispatchQueue(label: "BackoffRetrier")
-        
+
         monitor.pathUpdateHandler = { [weak self] path in
             if path.status == .satisfied {
                 Task {
@@ -100,10 +100,10 @@ public actor BackoffRetrier {
                 }
             }
         }
-        
+
         monitor.start(queue: queue)
     }
-    
+
     private func reset() {
         attempt = 0
     }
--- a/WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetrier.swift
+++ b/WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetrier.swift
@@ -56,7 +56,12 @@ public actor BackoffRetrier {
 
         while true {
             do {
-                return try await operation()
+                // Temporary code to test retry mechanism, to be removed after playtest
+                if attempt < policy.maxRetries {
+                    throw NSError(domain: "wire.com", code: 400)
+                } else {
+                    return try await operation()
+                }
             } catch {
 
                 guard attempt < policy.maxRetries else {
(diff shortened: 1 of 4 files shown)


WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetrier.swift
```diff
@@ -0,0 +1,110 @@
...
+        _ seconds: Double
+    ) async throws -> Void
+    
+    enum Failure: Error {
+        case exceededMaxAttempts(latestError: any Error)
+    }
+    
+    private let policy: BackoffRetryPolicy
+    private let monitor = NWPathMonitor()
+    private let sleep: SleepFunction
```
jullianm, line 34: for testing

WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetrier.swift
```diff
@@ -0,0 +1,110 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+import Foundation
+import UIKit
+import Network
+
+public actor BackoffRetrier {
```
jullianm, line 23: had to make this actor

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -104,8 +105,12 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
 
     func resume() {
         Task {
+            let retrier = BackoffRetrier()
+
             do {
-                try await performSync()
+                try await retrier.retry { [self] in
```
jullianm, line 111: For now it's only used for the sync, I wonder if it should be more generally integrated in the NetworkService. I think the back off retry mechanism is used by the ZMTransportSession in legacy code

WireFoundation/Tests/WireFoundationTests/Utilities/BackoffRetrierTests.swift
```diff
@@ -0,0 +1,83 @@
...
+
+    override func setUp() {
+        sut = BackoffRetrier()
+    }
+
+    override func tearDown() {
+        sut = nil
+    }
+
+    // TODO: [WPB-17645] Assert sleep durations exponentially increased
```
jullianm: Added this comment because I started testing sleep durations make sure it exponentially increased and I had flaky tests probably because of isolation issue which I'm not yet sure how to fix hence the comment.

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
netbe: I would have said put it in networkService to be in sync with legacy code. But at least here it seems in sync with what Android does

netbe (approved):
LGTM;) as first this could work, could be moved to NetworkService in a second time - to be tested

WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetryPolicy.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    /// The maximum amount of time in seconds to wait before the next attempt
+    public let maxTime: TimeInterval
+
+    /// When calculating backoff time, the exponent is the number of attempts multiplied by
+    /// this value. Decrease this value to shorten the backoff time.
+    public let exponentMultiplier: Double
+
+    /// Whether to introduce randomness into the backoff
+    public let jitter: Bool
```
samwyndham, line 37: suggestion: Remove this as we are not using it in practice. If we need it in the future we can add it.

WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetrier.swift
```diff
@@ -0,0 +1,109 @@
...
+    ) {
+        self.policy = policy
+        self.sleep = sleep
+
+        setupObservers()
+    }
+
+    deinit {
+        monitor.cancel()
+        NotificationCenter.default.removeObserver(self)
```
samwyndham: suggestion: remove this line. I don't think it is necessary:

> If your app targets iOS 9.0 and later or macOS 10.11 and later, and you used [addObserver(_:selector:name:object:)](https://developer.apple.com/documentation/foundation/notificationcenter/addobserver(_:selector:name:object:)), you do not need to unregister the observer. If you forget or are unable to remove the observer, the system cleans up the next time it would have posted to it.

WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetrier.swift
```diff
@@ -0,0 +1,109 @@
...
+
+                // Exponential backoff
+                var delay = policy.baseTime * Double(pow(policy.exponentMultiplier, Double(attempt)))
+
+                if policy.jitter {
+                    // Adds jitter (randomness)
+                    delay = Double.random(in: 0 ... min(policy.maxTime, delay))
+                }
+
+                try await sleep(delay)
```
samwyndham, line 74: issue: I don't think this is the concurrency version but just the classic sleep which we don't want as it will sleep the thread instead of suspending the task. You probably meant: `Task.sleep(nanoseconds`

samwyndham (changes requested):
The approach looks good but I think there is bug - easy fix

WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetrier.swift
netbe: Isn’t it the closure pass init for testing ?

WireFoundation/Tests/WireFoundationTests/Utilities/BackoffRetrierTests.swift
jullianm: fixed this in https://github.com/wireapp/wire-ios/pull/3015/commits/01ac58bd56c1c39466c061b658c11d3e6d0adcb1,

WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetrier.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/3015/commits/99b50c37cd0cc3be1494baeaf1aaa9d6659ce75a

WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetrier.swift
jullianm: actually we use the Task.sleep version, the `sleep` is the naming of function we inject (to facilitate testing):

```
public init(
        policy: BackoffRetryPolicy = .init(),
        sleep: @escaping SleepFunction = { delay in
            try await Task.sleep(nanoseconds: UInt64(delay * 1_000_000_000))
        }
    )
```

WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetryPolicy.swift
jullianm: I would be in favor of keeping it with a default value to false, we have some logic in the BackoffRetrier, so if we need to use it someday we just have to set the flag to true here

WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetrier.swift
samwyndham: 🤦 Of course. Sorry I missed it.

samwyndham (approved):

dmitrysimkin (approved):
LGTM 👍

WireFoundation/Sources/WireFoundation/Utilities/BackoffRetry/BackoffRetrier.swift
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public func retry<T: Sendable>(
+        _ operation: @escaping () async throws -> T
+    ) async throws -> T {
+
+        while true {
+            do {
+                return try await operation()
+            } catch {
```
johnxnguyen, line 60: question: so we retry for any failure? Or is it supposed to be for rate limiting errors?

jullianm: Yes, currently this implementation retries on any thrown error that occurs during the sync so we get a chance to recover.. however I'd say this is a first (quick fix) step.. as we could use this retrier at the `NetworkService` level and only retry for specific errors (like rate limiting errors). 

I think on Android they only retry on syncing but in our legacy code base I think we use a similar mechanism `ZMExponentialBackoff` in `ZMTransportSession` though (I haven't look deeper). 

Do you want me to create a ticket to follow-up on this ?

johnxnguyen: yes please, a follow up ticket for backing off in the networkService after rate limiting would be a good next step.

PR #3038, opened 2025-05-20
fix: pull self user after pulling known users - WPB-17516

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17516" title="WPB-17516" target="_blank">BugWPB-17516</a>  [iOS] Personal users created without email and password
  </td></table>
  <br />
 

### Issue

**Context**: After doing a personal user registration, go to the settings, the user email doesn't show up. Issue happens on syncV2 only.

**Causes**: After registration is completed, we perform an initial sync by pulling ressources (syncV2) where first we `pullSelfUser`: backend sends back the user info including the email that we persist correctly in database. Once this phase is completed, we `pullKnownUsers`: backend sends back the known users which only includes one user at this point (the self one) however the response object doesn't contain the email value (null) and we persist that to the database ending up in a state where self user has no longer an email address.

**Solution**: When pulling resources, first `pullKnownUsers` then `pullSelfUser`, issue doesn't happen on syncV1 because we already do pull known users before pulling self user.

### Testing

Go through personal user registration, then go to settings, check the email, value should be present.

### Checklist

---

### DIFF
--- a/WireDomain/Sources/WireDomain/Synchronization/PullResourcesSync.swift
+++ b/WireDomain/Sources/WireDomain/Synchronization/PullResourcesSync.swift
@@ -69,6 +69,13 @@ struct PullResourcesSync: PullResourcesSyncProtocol {
 
     func pull() async throws {
         try await logger.measureTime(label: "pull resources") {
+            try await pullUserConnections()
+            try await pullAllConversations()
+
+            // Pulling known users must happen after we've discovered
+            // user ids from user connections and conversations.
+            try await pullKnownUsers()
+
             let teamID = try await pullSelfUser()
             try await pullSelfUserClients()
             try await pullSelfUserSettings()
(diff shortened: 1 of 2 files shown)


dmitrysimkin (approved):
Nice finding 👍  Approved 
also would be nice to add test to make sure order is correct in the future

johnxnguyen (commented):
question: I wonder if it would be more robust to not overwrite the email and password fields with nil values for the self user, which seems reasonable to expect that you can never clear your email or password, only change them). The current solution as it is depends on the temporal ordering of tasks which I feel is prone to breaking. What do you think?

jullianm:
@johnxnguyen personally I think it's better to keep it this level (a higher level in the hierarchy) since the order we pull resources already matter (i.e `pullKnownUsers()` must be done after we've discovered user ids from user connections and conversations) although I agree this is prone to breaking if we change the order, and in any cases we must enforce that the order matters. 

As a solution, I can add a UT, as @dmitrysimkin suggested, to enforce that the order must be respected in a way that if the order changes the UT will break, in addition I can add a comment on the method. (done in https://github.com/wireapp/wire-ios/pull/3038/commits/76fe100bbad4fef89880907c9ae55d3a8ac994d8) 

Note: I tried the other suggested approach before and I felt it was a bit weird to introduce such specific logic into the store `persistUser` method as it is currently pretty generic and should not be responsible of that in my opinion.

johnxnguyen (approved):

PR #3042, opened 2025-05-20
chore: update core crypto to 6.x - WPB-17651

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17651" title="WPB-17651" target="_blank">TaskWPB-17651</a>  [iOS] Adopt Core Crypto 6.x to fix out of sync MLS issues
  </td></table>
  <br />
 

### Issue

Update CoreCrypto to 6.0.1
https://github.com/wireapp/core-crypto/releases

In this PR I implemented the breaking changes:
```
Migration: before instantiating this version of core crypto for the first time, call migrateDatabaseKeyTypeToBytes()
with the appropiate arguments (old key and your new key) exactly once.
Then, instantiate core crypto with the new key.
```
We store key as Data and that's why we don't need to create or store another one. We just need to send an "oldKey" as a string, and a `newKey` as a Data.

### Checklist

---


wire-ios-data-model/Source/Core Crypto/CoreCryptoConfiguration.swift
```diff
@@ -47,32 +47,11 @@ public class CoreCryptoConfigProvider {
 
     // MARK: - Configuration
 
-    public func createFullConfiguration(
```
KaterinaWire, line 50: This method was used only in tests.

wire-ios-data-model/Source/Core Crypto/CoreCryptoConfiguration.swift
```diff
@@ -83,28 +62,20 @@ public class CoreCryptoConfigProvider {
...
-                key: key.base64EncodedString()
+                key: key
             )
         } catch {
             WireLogger.coreCrypto.error("Failed to get core crypto key \(String(describing: error))")
             throw ConfigurationSetupFailure.failedToGetCoreCryptoKey
         }
     }
 
-    public func clientID(of selfUser: ZMUser) throws -> String {
```
KaterinaWire, line 97: This method was used in `createFullConfiguration` and tests.

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -129,3 +166,35 @@ struct LegacyCoreCryptoKeychainItem: KeychainItemProtocol {
...
+            kSecClass: kSecClassGenericPassword,
+            kSecAttrService: keychainServiceName,
+            kSecAttrAccount: id,
+            kSecReturnData: true
+        ]
+    }
+
+    func setQuery(value: some Any) -> [CFString: Any] {
+        [
+            kSecClass: kSecClassGenericPassword,
```
KaterinaWire: The new key is `keychainServiceName` because we can use attributes as `kSecAttrComment` or `kSecAttrLabel` to provide more info about the key.

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -17,27 +17,64 @@
...
         do {
-            return try fetchCoreCryptoKey()
+            return try fetchCoreCryptoKeyV2()
         } catch {
             if createIfNeeded {
-                return try createCoreCryptoKey()
+                guard let oldKey = try? fetchCoreCryptoKey() else {
+                    return try createCoreCryptoKeyV2()
+                }
+                return try await migrateDatabaseKey(path: path, oldKey: oldKey)
```
johnxnguyen:
```suggestion
                let newKey = try createCoreCryptoKeyV2()
                guard let oldKey = try? fetchCoreCryptoKey() else {
                    return newKey
                }
                return try await migrateDatabaseKey(path: path, oldKey: oldKey, newKey: newKey)
```

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -17,27 +17,64 @@
...
+        let item = CoreCryptoKeychainItemV2()
+        WireLogger.coreCrypto.info("Core crypto key_v2 doesn't exist. Creating...")
+        let key = try KeychainManager.generateKey(numberOfBytes: 32)
+        WireLogger.coreCrypto.info("Created core crypto key_v2: \(key.base64String()). Storing...")
+        try KeychainManager.storeItem(item, value: key)
+        WireLogger.coreCrypto.info("Stored core crypto key_v2. Returning...")
+        return key
+    }
+
+    private func migrateDatabaseKey(path: String, oldKey: Data) async throws -> Data {
```
johnxnguyen: question: should we delete the old key?

KaterinaWire: I think yes, I don't have ideas why we can use it again.

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -17,27 +17,65 @@
...
-            return try fetchCoreCryptoKey()
+            return try fetchCoreCryptoKeyV2()
         } catch {
             if createIfNeeded {
-                return try createCoreCryptoKey()
+                let newKey = try createCoreCryptoKeyV2()
+                guard let oldKey = try? fetchCoreCryptoKey() else {
+                    return newKey
+                }
+                return try await migrateDatabaseKey(path: path, oldKey: oldKey, newKey: newKey)
```
johnxnguyen: issue: in the description it says

> Migration: before instantiating this version of core crypto for the first time, call migrateDatabaseKeyTypeToBytes()
with the appropiate arguments (old key and your new key) exactly once.
Then, instantiate core crypto with the new key.

Suppose you invoke this method `coreCryptoKey`:
1. There is no v2 key, so we enter the catch
2. We successfully create a new v2 key
3. We attempt to migrate but it fails
4. We relaunch the app, enter the method again
5. There is now a v2 key so we return it and use it to initialize core crypto v6

In this scenario, we initialized core crypto with the new key but without migrating it, which is a problem.

 Yet again we see another need for an interruptible migration, and possibly an app version migration (rather than relying on the presence of keys to know whether to do a migration or not). What do you think?

KaterinaWire: You're right, this is a valid case that the current implementation doesn't handle well. We shouldn't store the v2 key until the migration completes successfully.
Even if we trigger during an app version migration, we face the same risk unless the migration process is designed to be interruptible and resumable. Is an app version migration currently designed this way?

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -22,10 +22,18 @@ import WireSystem
...
+    public init(coreCryptoKeyMigrationManager: CoreCryptoKeyMigrationManagerProtocol?) {
+        self.coreCryptoKeyMigrationManager = coreCryptoKeyMigrationManager
+    }
+
+    public func coreCryptoKey(
+        createIfNeeded: Bool,
+        path: String
+    ) async throws -> Data {
         removeLegacyKeyIfNeeded()
+        try await migrateKeyIfNeeded(path: path)
```
typfel, line 36: Aren't we missing marking the migration as not necessary when we don't have an old key we are end up creating a new one?

KaterinaWire: Yes, good point. I implemented a fix.

CC throws an error: "SQL logic error".
Maybe it's better to improve the error description.

typfel (approved):

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
johnxnguyen: App version migrations should be interruptible, in fact all necessary multi-step flows should be interruptible. However, you don't necessarily need an app version migration (such as thing is only an experiment anyway), the important thing is to make it interruptible.

WireDomain/Sources/WireDomain/Notifications/Components/PullEventsStep.swift
```diff
@@ -102,13 +102,25 @@ extension PullEventsStep {
...
+    }
+
     var coreCryptoProvider: any CoreCryptoProviderProtocol {
         CoreCryptoProvider(
             selfUserID: selfUserID,
             sharedContainerURL: dependency.applicationContainer,
             accountDirectory: accountContainer,
             syncContext: dependency.coreData.syncContext,
             cryptoboxMigrationManager: CryptoboxMigrationManager(),
+            coreCryptoKeyMigrationManager: CoreCryptoKeyMigrationManager(journal: journal),
```
johnxnguyen, line 123: question: do we want to perform the migration in the NSE?

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -38,6 +46,20 @@ public class CoreCryptoKeyProvider {
         }
     }
 
+    private func migrateKeyIfNeeded(path: String) async throws {
+        WireLogger.coreCrypto.info("Migrating core crypto key...")
+
+        if let oldKey = try? fetchCoreCryptoKey() {
```
johnxnguyen, line 52: question: so if we have no key, then it means it's a fresh install? If so, perhaps a comment is necessary here.

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -38,6 +46,20 @@ public class CoreCryptoKeyProvider {
...
+            try await coreCryptoKeyMigrationManager?.performMigrationIfNeeded(
+                path: path,
+                oldKey: oldKey.base64EncodedString(),
+                newKey: oldKey
+            )
```
johnxnguyen, line 59: suggestion: I think a comment here about why we pass the old key as the new key is also needed here.

WireDomain/Sources/WireDomain/Core Crypto/CoreCryptoKeyMigrationManager.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    public var isMigrationNeeded: Bool {
+        journal[.isCoreCryptoKeyMigrationRequired]
+    }
+
+    public func performMigrationIfNeeded(path: String, oldKey: String, newKey: Data) async throws {
+        if isMigrationNeeded {
+            WireLogger.coreCrypto.info("Core crypto key migration is required")
+
+            try await migrateDatabaseKeyTypeToBytes(path: path, oldKey: oldKey, newKey: newKey)
```
johnxnguyen, line 39: question: where is this method implemented?

KaterinaWire: This is a CC method

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
KaterinaWire: Added

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
KaterinaWire: Yes.
Added comment

WireDomain/Sources/WireDomain/Notifications/Components/PullEventsStep.swift
KaterinaWire: I think yes. If the user gets an auto update of the app, it would be better to provide the ability to receive notifications. I've added to NSE and the Share extension (may be similar, but less rare, as notifications)

johnxnguyen (approved):

PR #3049, opened 2025-05-21
fix: push channel suspension - WPB-17754

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17754" title="WPB-17754" target="_blank">BugWPB-17754</a>  [iOS] Push channel may not be suspended correctly
  </td></table>
  <br />
 

### Issue

**Context**: going to the background, we `suspend` the sync by using an incremental sync token which closes the push channel. For some reason, it seems that in some cases the incremental sync token would be null at the point we suspend so we don't close the push channel and it's alive while in background. When going back to the foreground it's resumed with a failure saying there's already an incremental sync running.

**Solutions**: A couple of safety measures to avoid that issue
- Have a Task that wraps the sync operations (initial + incremental) and keep a reference to it so we can cancel it later.
- Cancel that task when hit suspend (we know that Task cannot be nil as we always create a new one when resuming the sync)
- In the incremental sync, use `try Task.checkCancellation()` to check whether a cancellation occurred (from the parent task) and then explicitly close the push channel.

In summary, we want to cover two use cases when suspending the sync :

1- A token has already been created 
-> cancel the parent Task and use the `incrementalSyncToken` to cancel the child task and close the push channel.

2- A token has not yet been created.
-> cancel the parent Task, explicitly catch any errors (including CancellationError) in the IncrementalSync and close the push channel right away.

### Testing

_Describe how to test._

---

### Checklist

---

### DIFF
--- a/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
+++ b/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
@@ -257,10 +257,6 @@ public struct IncrementalSync: IncrementalSyncProtocol {
         let task: Task<Void, Never>
         let closePushChannel: () async -> Void
 
-        public var isCancelled: Bool {
-            task.isCancelled
-        }
-
         public init(
             task: Task<Void, Never>,
             closePushChannel: @escaping () async -> Void
--- a/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
+++ b/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
@@ -53,150 +53,166 @@ public struct IncrementalSync: IncrementalSyncProtocol {
         self.syncStateSubject = syncStateSubject
     }
 
-    public func perform() async throws {
-        // Parent task was cancelled, immediately abort incremental sync operations.
-        try Task.checkCancellation()
+    public func perform() async throws -> Token {
+        var wasCanceled = false
         
-        logger.debug("performing incremental sync")
-        syncStateSubject.send(.incrementalSyncing(.createPushChannel))
-        let pushChannel = try await pushChannelAPI.createPushChannel(clientID: selfClientID)
+        return try await withTaskCancellationHandler {
+            logger.debug("performing incremental sync")
+            syncStateSubject.send(.incrementalSyncing(.createPushChannel))
+            let pushChannel = try await pushChannelAPI.createPushChannel(clientID: selfClientID)
 
-        logger.debug("opening push channel")
-        syncStateSubject.send(.incrementalSyncing(.openPushChannel))
+            logger.debug("opening push channel")
+            syncStateSubject.send(.incrementalSyncing(.openPushChannel))
+            
+            let liveEventStream = try await pushChannel.open()
 
-        let processedEnvelopeIDs: Set<UUID>
-        do {
-            logger.debug("pulling pending update events")
-            syncStateSubject.send(.incrementalSyncing(.pullPendingEvents))
-            try await updateEventsSync.pull()
-
-            logger.debug("processing stored update events")
-            syncStateSubject.send(.incrementalSyncing(.processPendingEvents))
-            processedEnvelopeIDs = try await processStoredEvents()
-        } catch {
-            logger.debug("incremental sync interrupted, tearing down...")
-            await pushChannel.close()
-            throw error
-        }
-        
-        // Processing stored events can take some time to complete so we ensure parent task has not been cancelled in the meantime to avoid creating the push channel.
-        try Task.checkCancellation()
-        
-        let liveEventStream = try await pushChannel.open()
-
-        let task = Task { @Sendable [logger, decryptor, store, processor, databaseSaver, syncStateSubject] in
-            logger.debug("handling live event stream")
-            syncStateSubject.send(.liveSyncing)
-
+            let processedEnvelopeIDs: Set<UUID>
             do {
-                for try await var envelope in liveEventStream {
-                    
-                    // If the parent task has been cancelled and push channel is already opened, we close it and abort processing live events.
-                    do {
-                        try Task.checkCancellation()
-                    } catch {
-                        logger.debug("incremental sync interrupted, tearing down...")
-                        await pushChannel.close()
-                    }
-                    
-                    logger.debug("received live event envelope")
+                logger.debug("pulling pending update events")
+                syncStateSubject.send(.incrementalSyncing(.pullPendingEvents))
+                try await updateEventsSync.pull()
 
-                    if processedEnvelopeIDs.contains(envelope.id) {
-                        logger.debug(
-                            "live event already processed, skipping...",
-                            attributes: [.eventEnvelopeID: envelope.id]
-                        )
-                        continue
-                    }
-
-                    do {
-                        // Decrypt.
-                        logger.debug(
-                            "decrypting live event envelope",
-                            attributes: [.eventEnvelopeID: envelope.id]
-                        )
-                        envelope.events = try await decryptor.decryptEvents(in: envelope)
-                    } catch {
-                        logger.error(
-                            "failed to decrypt live event envelope: \(String(describing: error))",
-                            attributes: [.eventEnvelopeID: envelope.id]
-                        )
-                        continue
-                    }
-
-                    let index: Int64
-                    do {
-                        // Store.
-                        logger.debug(
-                            "storing live event envelope",
-                            attributes: [.eventEnvelopeID: envelope.id]
-                        )
-                        index = try await store.indexOfLastEventEnvelope() + 1
-                        try await store.persistEventEnvelope(envelope, index: index)
-                    } catch {
-                        logger.error(
-                            "failed to store live event envelope: \(String(describing: error))",
-                            attributes: [.eventEnvelopeID: envelope.id]
-                        )
-                        continue
-                    }
-
-                    // Bump the last event id so we don't refech it.
-                    if !envelope.isTransient {
-                        logger.debug(
-                            "updating last event id",
-                            attributes: [.eventEnvelopeID: envelope.id]
-                        )
-                        store.storeLastEventID(id: envelope.id)
-                    }
-
-                    // Process.
-                    for event in envelope.events {
-                        do {
-                            logger.debug(
-                                "processing live event: \(event.name)",
-                                attributes: [.eventEnvelopeID: envelope.id]
-                            )
-                            try await processor.processEvent(event)
-                        } catch {
-                            logger.error(
-                                "failed to process live event: \(String(describing: error))",
-                                attributes: [.eventEnvelopeID: envelope.id]
-                            )
-                        }
-                    }
-
-                    do {
-                        // Delete.
-                        logger.debug(
-                            "deleting live event envelope",
-                            attributes: [.eventEnvelopeID: envelope.id]
-                        )
-                        try await store.deleteEventEnvelope(atIndex: index)
-                    } catch {
-                        logger.error(
-                            "failed to delete live event envelope: \(String(describing: error))",
-                            attributes: [.eventEnvelopeID: envelope.id]
-                        )
-                    }
-
-                    await store.calculateLastUnreadMessages()
-
-                    do {
-                        // Save.
-                        try await databaseSaver.save()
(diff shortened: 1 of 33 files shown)


WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -53,29 +53,52 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
+        } catch {
+            logger.debug("incremental sync interrupted, tearing down...")
+            await pushChannel.close()
+            throw error
+        }
+        
+        // Processing stored events can take some time to complete so we ensure parent task has not been cancelled in the meantime to avoid creating the push channel.
+        try Task.checkCancellation()
+        
+        let liveEventStream = try await pushChannel.open()
```
johnxnguyen: issue: we must open the push channel before pulling events, otherwise we risk to miss some events between the call to pull and before the push channel is opened.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -53,29 +53,52 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
             do {
                 for try await var envelope in liveEventStream {
+                    
+                    // If the parent task has been cancelled and push channel is already opened, we close it and abort processing live events.
+                    do {
+                        try Task.checkCancellation()
+                    } catch {
+                        logger.debug("incremental sync interrupted, tearing down...")
+                        await pushChannel.close()
+                    }
```
johnxnguyen: thought: if we've reached this far, then the normal suspension should have already closed the push channel and ended the stream. But I see now that the suspension no longer closes the push channel explicitly. The problem here is that we would only reach this cancellation check when a new live event arrives, and we don't know how long that will be. I think it's best to continue to explicitly close the push channel.

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -104,9 +104,9 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
     /// This method logs any errors and does not wait for the sync to finish.
 
     func resume() {
-        Task {
+        ongoingSyncTask = Task {
```
johnxnguyen, line 108: This task will complete after we've processed the pending events. What would happen if we cancel it when it's already finished? Are you sure that the live sync task spawned inside will be cancelled too?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
jullianm: makes sense, thanks for the feedback, I'll rework that

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -53,150 +53,168 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
+
+            return Token(task: liveEventTask, closePushChannel: {
+                await pushChannel.close()
+            })
+
+        } onCancel: {
+            // this cancellation handler is always and immediately invoked when the parent task (see `SyncAgent`) is
+            // canceled.
+            wasCanceled = true
+        }
```
johnxnguyen: If I understand correctly, checking `if wasCanceled` is no different to `try Task.checkCancellation` because it's only checking at a single point in time if the task has been canceled.

Wouldn't it be better to do the tear down whenever `onCancel` is called? I guess it means that you need to have another task layer:

```swift
func perform() async throws -> Token {
  let pushChannel = ...
  let liveEventStream = ...
  try await pullPendingEvents()
  try await processPendingEvents()
  
  return try await  withTaskCancellationHandler {
    return Token(
      task: Task {
        // Live sync goes here
      }, 
    closePushChannel: {
        // Maybe we don't even need this... if we call suspend
        // then the enclosing task should be cancelled and the
        // onCancel closure below should be called.
        await pushChannel.close()
    }
  } onCancel: {
    // Called implicitly when the task is cancelled
    await pushChannel.close()
  }
}


```

jullianm: Issue here is that the cancellation handler is scoped to the `withTaskCancellationHandler` duration and since we're immediately returning from it (with a Token), the handler is effectively done, and any cancellation that happens later won't go through the `onCancel` block.

However.. now I wonder whether we need that `withTaskCancellationHandler` at all, we could simply return the `Token` directly with its live events task setup inside directly. When we hit suspend in the parent task, we use that token to close the push channel already. 

Could the push channel not be closed properly ? I don't see how as we now handle any cancellation before returning that Token (and closing the push channel if that happens) and directly returning the Token if no cancellation occurred in the meantime, token that we use to close the push channel at the `SyncAgent` level when we hit suspend.

```
    return Token(
      task: Task {
        // Live sync goes here
      }, 
    closePushChannel: {
        await pushChannel.close()
    }
  ```

Comment (jullianm) [reply WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift]: As an extra safety measure, we could check if there's an existing token when resume() again and clean it up at this point

File: WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift [hunk 5]
```diff
@@ -54,131 +54,168 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
-        let pushChannel = try await pushChannelAPI.createPushChannel(clientID: selfClientID)
+        var wasCanceled = false
 
-        logger.debug("opening push channel")
-        syncStateSubject.send(.incrementalSyncing(.openPushChannel))
-        let liveEventStream = try await pushChannel.open()
+        return try await withTaskCancellationHandler {
+            logger.debug("performing incremental sync")
+            syncStateSubject.send(.incrementalSyncing(.createPushChannel))
+            let pushChannel = try await pushChannelAPI.createPushChannel(clientID: selfClientID)
```
netbe: suggestion: move content of withTaskCancellationHandler within own method to see only the cancellation logic

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -196,10 +218,6 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
 
     func performIncrementalSync() async throws {
         if isSyncV2Enabled {
-            guard incrementalSyncToken == nil else {
```
netbe, line 200: question: why is it removed?

netbe (commented):

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
johnxnguyen: Maybe we're overthinking this. Would it suffice for the `SyncAgent` to hold on to two possible tasks, one for the finite syncing (i.e initial sync, incremental sync) and one for the live syncing, so that when `suspend()` is called we just cancel both tasks?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -57,128 +57,146 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
+            logger.debug("pulling pending update events")
+            syncStateSubject.send(.incrementalSyncing(.pullPendingEvents))
+            try await updateEventsSync.pull()
+            
+            logger.debug("processing stored update events")
+            syncStateSubject.send(.incrementalSyncing(.processPendingEvents))
+            processedEnvelopeIDs = try await processStoredEvents()
+        } catch {
+            logger.debug("incremental sync interrupted, tearing down...")
+            await pushChannel.close()
```
jullianm, line 77: any errors occurring (including the cancellation from the parent sync task) will be caught and we close the push channel right away (`try Task.checkCancellation` is used when decrypting the events and processing the stored events)

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -104,7 +105,11 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
     /// This method logs any errors and does not wait for the sync to finish.
 
     func resume() {
-        Task {
+        ongoingSyncTask = Task {
```
jullianm, line 108: parent sync task that we cancel when suspending so the cancellation error is propagated inside the IncrementalSync (so we can close the push channel)

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
jullianm: withTaskCancellationHandler was removed as it is not necessary

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
jullianm: not needed because when we suspend we tear down everything so push channel is always closed and live event task is cancelled, token is nullified as well

netbe (approved):
LGTM;)

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -239,6 +257,10 @@ public struct IncrementalSync: IncrementalSyncProtocol {
         let task: Task<Void, Never>
         let closePushChannel: () async -> Void
 
+        public var isCancelled: Bool {
+            task.isCancelled
+        }
```
johnxnguyen: question: is this still needed?

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
johnxnguyen: suggestion: I don't see this task being nilled out anywhere... maybe it's not necessary but might be good to do.

johnxnguyen (approved):
Looks good but there are a few improvements that can be made:
- these changes only address the `performSync` method and not when invoking `performInitialSync`, `performResourceSync`, and `performIncrementalSync` directly.
- If the incremental sync fails we update the sync state subject with `.suspended`, but we don't do this for any of the other sync methods.

I think it may be worth having an `enqueueSync` method that wraps any of the syncs inside a task (to allow for suspension and catches any errors and updates the sync state). For instance:

```swift
    private func enqueueSync(_ sync: @escaping () async throws -> Void) async throws {
        // Cancel anything ongoing so we can start a new sync.
        ongoingSyncTask?.cancel()

        // Wrap work in a task.
        ongoingSyncTask = Task {
            try await sync()
        }

        do {
            // Wait for the work to finish.
            try await ongoingSyncTask?.value
        } catch {
            // Clean up
            syncStateSubject.send(.suspended)
            ongoingSyncTask = nil
        }
    }

    /// Perform an initial sync.

    func performInitialSync() async throws {
        try await enqueueSync { [self] in
            if isSyncV2Enabled {
                do {
                    delegate?.syncAgentDidStartInitialSync(self)
                    WireLogger.sync.debug("did start new initial sync")
                    try await initialSyncProvider.provideInitialSync().perform(skipPullingLastUpdateEventID: false)
                    WireLogger.sync.debug("did finish new initial sync")
                    journal[.isInitialSyncRequired] = false
                    delegate?.syncAgentDidFinishInitialSync(self)
                } catch {
                    WireLogger.sync.error("failed to perform new initial sync: \(String(describing: error))")
                    throw error
                }

                try await performIncrementalSync()
            } else {
                // Incremental sync automatically follows the slow sync.
                legacySyncStatus.forceSlowSync()
            }
        }
    }
}
```

The idea here being that all sync work must be enqueued, so regardless of which sync is performed, the suspension will not have problems.

I will approve these changes for now but I think we should think about these improvements as a followup. What do you think?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
jullianm: no, removed

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
jullianm: it is when we suspend

PR #3055, opened 2025-05-22
feat: repair 1:1 conversation - WPB-17469

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17469" title="WPB-17469" target="_blank">BugWPB-17469</a>  [iOS] Recovering from conversation being out of sync
  </td></table>
  <br />
 

### Issue

We have a 1:1 conversation that has an wrong epoch state. During Incremental Sync, events containing messages are skipped and users don't have history on their clients.
In this PR we collect broken (wrong epoch) conversations during Incremental Sync and store them in a Journal. `MLSGroupRepairAgent` observes when the app enters `.liveSyncing` state and tries to repair them.

### Checklist

---


WireDomain/Sources/WireDomain/Event Decryption/Protocols/UpdateEventDecryptorProtocol.swift
```diff
@@ -27,6 +27,13 @@ public protocol UpdateEventDecryptorProtocol {
...
 
-    func decryptEvents(in eventEnvelope: UpdateEventEnvelope) async throws -> [UpdateEvent]
+    func decryptEvents(in eventEnvelope: UpdateEventEnvelope) async throws -> EventDecryptorResult
+
+}
+
+public struct EventDecryptorResult {
+
+    let events: [UpdateEvent]
+    let brokenMLSGroupIDs: [String]
```
johnxnguyen: suggestion: use a `Set` rather than an `Array` to avoid duplicates.

WireDomain/Sources/WireDomain/Event Decryption/UpdateEventDecryptor.swift
```diff
@@ -110,7 +113,7 @@ struct UpdateEventDecryptor: UpdateEventDecryptorProtocol {
                     "decrypting MLS add message event...",
                     attributes: logAttributes
                 )
-
+                // brokenMLSGroupIDs.append("AAEAAGWE4vNlaUGzpxLgWY9P4k8Ad2lyZS5jb20=")
```
johnxnguyen: issue: to be deleted

WireDomain/Sources/WireDomain/Utilities/Journal/Journal.swift
```diff
@@ -77,4 +92,39 @@ public class Journal: JournalProtocol {
...
+        currentSet.insert(value)
+        self[key] = currentSet
+    }
+
+    /// Adds multiple values to a Set<String>.
+    ///
+    /// - Parameters:
+    ///   - values: A set of strings to insert.
+    ///   - key: The journal key associated with the Set<String>.
+    public func addValue(_ values: Set<String>, for key: JournalKey<Set<String>>) {
```
johnxnguyen: suggestion: `addValues`

WireDomain/Sources/WireDomain/Utilities/Journal/Journal.swift
```diff
@@ -77,4 +92,39 @@ public class Journal: JournalProtocol {
         "\(namespace).\(key.name)"
     }
 
+    /// Remove a single value from a Set<String> .
```
johnxnguyen: suggestion: it may be nice to move these to an extension specifically for `Set<String>` keys.

WireDomain/Tests/WireDomainTests/Synchronization/PullPendingUpdateEventsSyncTests.swift
```diff
@@ -22,21 +22,27 @@ import XCTest
 @testable import WireDomain
 @testable import WireDomainSupport
 
-final class PullPendingUpdateEventsSyncTests: XCTestCase {
+final class PullPendingUpdateEventsSyncTests: XCTestCase { // maybe here
```
johnxnguyen: question: to delete?

WireDomain/Sources/WireDomain/Synchronization/PullPendingUpdateEventsSync.swift
```diff
@@ -106,6 +113,8 @@ public struct PullPendingUpdateEventsSync: PullPendingUpdateEventsSyncProtocol {
             }
         }
 
+        journal.addValue(brokenMLSGroupIDs, for: .brokenMLSGroupIDs)
```
johnxnguyen: suggestion: it may be wise to add to the journal at the end of the batch since if we happen to crash in the next batch we wouldn't have lost the information we gained from the last batch. But if you do, remember that you would need to move the initialization of the set to the start of the batch so it gets cleared for each batch.

wire-ios-sync-engine/Source/Synchronization/MLSGroupRepairAgent.swift
```diff
@@ -0,0 +1,102 @@
...
+import WireDomain
+import WireFoundation
+import WireLogging
+import WireUtilities
+
+// sourcery: AutoMockable
+protocol MLSGroupRepairAgentProtocol {
+
+    var isSyncV2Enabled: Bool { get }
+    var syncStatePublisher: AnyPublisher<SyncState, Never> { get }
```
johnxnguyen: question: why are these needed in the protocol?

wire-ios-sync-engine/Source/Synchronization/MLSGroupRepairAgent.swift
```diff
@@ -0,0 +1,102 @@
...
+                }
+                return (groupID, mlsGroupID)
+            }
+            .forEach { groupID, mlsGroupID in
+                Task {
+                    await mlsService.fetchAndRepairGroup(with: mlsGroupID)
+                    journal.removeValue(groupID, for: .brokenMLSGroupIDs)
+                    WireLogger.sync.debug("Successfully repaired group: \(groupID)")
+                }
+            }
```
johnxnguyen: issue: while probably not likely, there may be many groups to repair. If we create a separate task for each of them then they may all fire quickly after the other and result in a burst of network activity. What do you think about repairing them serially?

WireDomain/Sources/WireDomain/Event Decryption/Protocols/UpdateEventDecryptorProtocol.swift
KaterinaWire: We store them as a set in the journal anyway. I thought both results would be here with the same type.

WireDomain/Sources/WireDomain/Event Decryption/UpdateEventDecryptor.swift
KaterinaWire: done

WireDomain/Tests/WireDomainTests/Synchronization/PullPendingUpdateEventsSyncTests.swift
KaterinaWire: removed

WireDomain/Sources/WireDomain/Utilities/Journal/Journal.swift
KaterinaWire: done

WireDomain/Sources/WireDomain/Event Decryption/Protocols/UpdateEventDecryptorProtocol.swift
KaterinaWire: changed

WireDomain/Sources/WireDomain/Utilities/Journal/Journal.swift
KaterinaWire: done

wire-ios-sync-engine/Source/Synchronization/MLSGroupRepairAgent.swift
KaterinaWire: yes, agree.

wire-ios-sync-engine/Source/Synchronization/MLSGroupRepairAgent.swift
KaterinaWire: no need, removed

WireDomain/Sources/WireDomain/Synchronization/PullPendingUpdateEventsSync.swift
KaterinaWire: updated

wire-ios-sync-engine/Source/Synchronization/MLSGroupRepairAgent.swift
```diff
@@ -0,0 +1,101 @@
...
+import WireUtilities
+
+// sourcery: AutoMockable
+protocol MLSGroupRepairAgentProtocol {
+
+    var isSyncV2Enabled: Bool { get }
+
+}
+
+final class MLSGroupRepairAgent: MLSGroupRepairAgentProtocol {
```
jullianm, line 34: suggestion: code looks pretty similar to the `IncrementalSyncObserver`, we could probably avoid some duplication here.
Looks like what we do in both is observing the sync state changes (using `SyncState`) and reacting to these changes with some specific actions. 
We could have a more unique and global object dedicated to the sync (like `SyncObserver`) that observes these changes and perform some actions consequently.

jullianm (commented):
Nice work, added a suggestion

wire-ios-sync-engine/Source/Synchronization/MLSGroupRepairAgent.swift
johnxnguyen: Indeed we should be careful to have many sync state observers separately so that we have better control on what work gets triggered and when, but I think it's a premature optimization right now. 

Having said that, I can imagine having a kind of `AsyncWorkDispatcher` that is responsible for finding some work to run in the background at the appropriate time.

wire-ios-sync-engine/Source/Synchronization/MLSGroupRepairAgent.swift
```diff
@@ -39,7 +38,7 @@ final class MLSGroupRepairAgent: MLSGroupRepairAgentProtocol {
     }
 
     private let syncStateSubject: CurrentValueSubject<SyncState, Never>
-    var syncStatePublisher: AnyPublisher<SyncState, Never> {
+    private var syncStatePublisher: AnyPublisher<SyncState, Never> {
```
johnxnguyen: suggestion: better to inject the `AnyPublisher` rather than the current value subject since we only want to receive events and not send them.

johnxnguyen (approved):

wire-ios-sync-engine/Source/Synchronization/MLSGroupRepairAgent.swift
KaterinaWire: @jullianm Could you link a ticket for these changes?

wire-ios-sync-engine/Source/Synchronization/MLSGroupRepairAgent.swift
KaterinaWire: done

jullianm (approved):
LGTM

PR #2916, opened 2025-04-18
chore: remove voip handling - WPB-17290

netbe:
### Issue

We don't use VOIP push anymore since iOS 15. Still we have some code that handles it. This PR cleans up what's not used.

### Testing

N/A

PR #3060, opened 2025-05-22
fix: CallKit crashes when in background - WPB-17848

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17848" title="WPB-17848" target="_blank">BugWPB-17848</a>  [iOS] App crashes in background when there are CallKit calls
  </td></table>
  <br />
 

### Issue

**Context**: When incoming calls arrive in the background and they're handled with CallKit, app crashes.

**Causes**: Some changes were introduced recently in this [PR](https://github.com/wireapp/wire-ios/pull/2916) where we replaced `pushRetry` and `reportNewIncomingCall` completionHandler versions of Apple's API with their async versions which appear to be broken as they're not called properly when waking up the app from the background. 

It ends up crashing as since iOS 13:
> if you fail to report a call to CallKit, the system will terminate your app.

..which is what we experienced when looking at the console.

**Solution**: As mentioned in the latest comments in this [SO](https://stackoverflow.com/questions/56788314/ios-13-killing-app-because-it-never-posted-an-incoming-call-to-the-system-after) post and after testing with both versions (async vs. completionHandler), the issue was reproducible with the async version and appeared to work properly with the completionHandler one.
Solution is to revert these changes to make sure we use the completionHandler API versions as before, thus properly report the call and avoid crashing.

### Testing

Tested internally on both async / completionHandler versions, issue was reproducible on the async version, not the completionHandler one.

### Checklist

### DIFF
--- a/wire-ios-sync-engine/Source/Calling/CallKitManager.swift
+++ b/wire-ios-sync-engine/Source/Calling/CallKitManager.swift
@@ -409,7 +409,7 @@ public class CallKitManager: NSObject, CallKitManagerInterface {
 
         // Don't use the async version, it's broken
         // It doesn't get executed when waking up the app from the background and ends up crashing
-        // See latest comments https://developer.apple.com/documentation/callkit/sending_end-to-end_encrypted_voip_calls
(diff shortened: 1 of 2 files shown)


johnxnguyen (approved):

dmitrysimkin (approved):
👍

PR #3066, opened 2025-05-23
fix: messages not being sent, waiting forever for decryption to finish - WPB-17866

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17866" title="WPB-17866" target="_blank">BugWPB-17866</a>  [iOS] Not able to send messages on mobile data
  </td></table>
  <br />
 

### Issue

**Context**: Messages are stuck forever in a sending state, no way to recover. Next messages can't be sent as well, stuck in same state.

We introduced in this [PR](https://github.com/wireapp/wire-ios/pull/2966) a mechanism that waits for pending events to be decrypted before proceeding to message sending.

Specifically in `MessageSender`, we await for `incrementalSyncObserver.waitUntilCanSendMessage()` and it only gets through when finishing decrypting the pending events and setting the `decryptionState` to done at the  `IncrementalSyncObserver` level.

**Causes**:  The decryption state might never gets updated to `.done` hence blocking the message sender to proceed the message which can lead to the bugs mentioned [here](https://wearezeta.atlassian.net/browse/WPB-17772) and [here](https://wearezeta.atlassian.net/browse/WPB-17866).

What happens after is that if we try to send a message again we'll be stuck in a loop in `try await messageDependencyResolver.waitForDependenciesToResolve(for: message)` where the dependencies never resolve specifically `dependentObjectNeedingUpdateBeforeProcessing` (code doc below) will never be nil, probably because the initial message was never really processed and it blocks the next messages.

`    
/// Other entities which has to complete an update before this entity can be processed, i.e. another message
/// needs to be sent first because it was scheduled for sending before this message.
`

**Solution**: Update the sync state whenever the sync is suspended or when an error occurs while pulling (and decrypting) pending events so we ensure the decryption state is changed to done so the `MessageSender` method is "unlocked" and the messages are sent or at least attempted to be sent.

### Testing

Tested internally
- Comment out the code where the sync state is updated so the MessageSender gets stuck and the message not sent.
- try to send message again and again
- messages shows up as in sending state just like the screenshots from the Jira tickets

Capture d’écran   2025-05-23 à 16 47 36

### Checklist

---

### DIFF
--- a/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
+++ b/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
@@ -63,8 +63,14 @@ public struct IncrementalSync: IncrementalSyncProtocol {
         let liveEventStream = try await pushChannel.open()
 
         logger.debug("pulling pending update events")
-        syncStateSubject.send(.incrementalSyncing(.pullPendingEvents))
-        try await updateEventsSync.pull()
+        syncStateSubject.send(.incrementalSyncing(.pullPendingEvents(.inProgress)))
+        do {
+            try await updateEventsSync.pull()
+        } catch {
+            logger.debug("pulling pending update events interrupted \(String(describing: error))")
+            syncStateSubject.send(.incrementalSyncing(.pullPendingEvents(.interrupted)))
+            throw error
+        }
 
         logger.debug("processing stored update events")
         syncStateSubject.send(.incrementalSyncing(.processPendingEvents))
--- a/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
+++ b/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
@@ -63,14 +63,8 @@ public struct IncrementalSync: IncrementalSyncProtocol {
         let liveEventStream = try await pushChannel.open()
 
         logger.debug("pulling pending update events")
-        syncStateSubject.send(.incrementalSyncing(.pullPendingEvents(.inProgress)))
-        do {
-            try await updateEventsSync.pull()
-        } catch {
-            logger.debug("pulling pending update events interrupted \(String(describing: error))")
-            syncStateSubject.send(.incrementalSyncing(.pullPendingEvents(.interrupted)))
-            throw error
-        }
+        syncStateSubject.send(.incrementalSyncing(.pullPendingEvents))
+        try await updateEventsSync.pull()
 
         logger.debug("processing stored update events")
         syncStateSubject.send(.incrementalSyncing(.processPendingEvents))
--- a/wire-ios-sync-engine/Source/Synchronization/IncrementalSyncObserver.swift
+++ b/wire-ios-sync-engine/Source/Synchronization/IncrementalSyncObserver.swift
@@ -55,9 +55,16 @@ final class IncrementalSyncObserver: IncrementalSyncObserverProtocol {
                 .receive(on: decryptionQueue)
                 .sink { [weak self] syncState in
                     switch syncState {
-                    case .incrementalSyncing(.pullPendingEvents):
-                        self?.decryptionState = .inProgress
+                    case .incrementalSyncing(.pullPendingEvents(let state)):
+                        switch state {
+                        case .inProgress:
+                            self?.decryptionState = .inProgress
+                        case .interrupted:
+                            self?.decryptionState = .done
+                        }
                     case .incrementalSyncing(.processPendingEvents), .liveSyncing:
+                        self?.decryptionState = .done
+                    case .suspended:
                         self?.decryptionState = .done
                     default:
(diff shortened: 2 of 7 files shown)


wire-ios-sync-engine/Source/Synchronization/IncrementalSyncObserver.swift
```diff
@@ -55,10 +55,17 @@ final class IncrementalSyncObserver: IncrementalSyncObserverProtocol {
...
                 .sink { [weak self] syncState in
                     switch syncState {
-                    case .incrementalSyncing(.pullPendingEvents):
-                        self?.decryptionState = .inProgress
+                    case let .incrementalSyncing(.pullPendingEvents(state)):
+                        switch state {
+                        case .inProgress:
+                            self?.decryptionState = .inProgress
+                        case .interrupted:
+                            self?.decryptionState = .done
```
samwyndham: thought (non-blocking): It is kind of funny that `interrupted` maps to `done`. I'm wondering if the current states we have make sense:

``` swift
enum DecryptionState {
    case notStarted
    case inProgress
    case done
}
```

In the code I don't think we ever check for the `notStarted` state so I wonder if there should be only 2 states:

```swift
enum DecryptionState {
    case idle
    case decrypting
}
```

samwyndham (approved):
Nice work!

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -63,8 +63,14 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
 
         logger.debug("pulling pending update events")
-        syncStateSubject.send(.incrementalSyncing(.pullPendingEvents))
-        try await updateEventsSync.pull()
+        syncStateSubject.send(.incrementalSyncing(.pullPendingEvents(.inProgress)))
+        do {
+            try await updateEventsSync.pull()
+        } catch {
+            logger.debug("pulling pending update events interrupted \(String(describing: error))")
+            syncStateSubject.send(.incrementalSyncing(.pullPendingEvents(.interrupted)))
```
johnxnguyen: suggestion: perhaps it's better to catch the error in the `SyncAgent` and update the sync state there, that way we don't have to catch every possible error inside.

wire-ios-sync-engine/Source/Synchronization/IncrementalSyncObserver.swift
johnxnguyen: I like this simplification.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
jullianm: updated in https://github.com/wireapp/wire-ios/pull/3066/commits/4dd0f41c48425ef50930316a8bcb96903555572d

wire-ios-sync-engine/Source/Synchronization/IncrementalSyncObserver.swift
jullianm: I think this will break the message sender logic which explicitly knows (through the states) it has to wait before proceeding with message sending. I think the cases we need to have is ones that prevents the message sender from sending a message and a one that gives the green light, which I think are currently well represented although it could be simplified a bit:

```
enum DecryptionState {
    case waiting
    case done
}

```

But let's keep it for another PR as this could be improved later on.

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -210,6 +211,7 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
                 }
             } catch {
                 WireLogger.sync.error("failed to perform new incremental sync: \(String(describing: error))")
+                syncStateSubject.send(.suspended)
```
netbe, line 214: **question:** is it really only suspended here? I wonder if the state should be .failure(error)

netbe (commented):
left a question and wonder if a unit test can be added

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
jullianm: probably the term `suspended` is a bit broad and less accurate than failure but I wanted to make it simple for the `IncrementalSyncObserver` to rely on the same case when it needs to update the decryptionState to done

johnxnguyen: I think 'suspended' here is synonymous with 'idle', indicating that no sync is currently active.

johnxnguyen (approved):

jullianm:
added some UTs in last commit @netbe , merging the PR

PR #3069, opened 2025-05-26
feat: add spinner - WPB-17596

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17596" title="WPB-17596" target="_blank">BugWPB-17596</a>  [iOS] Own devices not shown as MLS enabled while they actually are
  </td></table>
  <br />
 

### Issue

Sometimes it takes a while to get an MLS thumbprint, and from the user's perspective it's not clear that something is fetching. 
This PR introduces a spinner for the user's device list.

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientListViewController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientListViewController.swift
@@ -119,10 +119,8 @@ final class ClientListViewController: UIViewController,
             self.userObserverToken = UserChangeInfo.add(observer: self, for: user, in: session)
         }
 
-        if clientsList == nil {
-            if clients.isEmpty {
-                activityIndicator.start()
-            }
+        if clients.isEmpty {
+            activityIndicator.start()
             userSession?.fetchAllClients()
         }
     }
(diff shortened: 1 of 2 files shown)


jullianm (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientListViewController.swift
```diff
@@ -167,14 +165,19 @@ final class ClientListViewController: UIViewController,
...
 
         // Prevent more then one removalObserver in self and SettingsClientViewController
         removalObserver = nil
     }
 
-    private func dismissLoadingView() {
-        activityIndicator.stop()
+    private func dismissLoadingView() async {
+        let minimumDelay: TimeInterval = 0.2
+        let nanoseconds = UInt64(minimumDelay * 1_000_000_000)
```
samwyndham, line 175: question: Why do we need a delay?

samwyndham (commented):
question: I don't quite understand the description. It says that this PR introduces a spinner but it looks like there already was one. Was it not working?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientListViewController.swift
samwyndham: The reason for the delay is to avoid a blink in the common case where the client list info has already been fetched

samwyndham (approved):
Nice work

PR #3076, opened 2025-05-27
fix: push notification connection accepted placeholder - WPB-17903

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17903" title="WPB-17903" target="_blank">BugWPB-17903</a>  [iOS] Connection push notification show placeholder 
  </td></table>
  <br />
 

### Issue

**Context**: When a connection is accepted, push notification doesn't display the user we connected with.

**Causes**: We solely rely on the username value which might be null.

**Solution**: As in legacy code, rely on the sender name and fallback on the username value.

BEFORE:
Screenshot 2025-05-27 at 08 07 38

NOW: 
Screenshot 2025-05-27 at 08 07 48

### Testing

_Describe how to test._

---

### Checklist

---

### DIFF
--- a/WireDomain/Sources/WireDomain/Notifications/Builders/User/UserConnectionEventNotificationBuilder.swift
+++ b/WireDomain/Sources/WireDomain/Notifications/Builders/User/UserConnectionEventNotificationBuilder.swift
@@ -52,9 +52,20 @@ struct UserConnectionEventNotificationBuilder {
         let connectionStatus = isPendingConnection ? ConnectionStatus.pending : .accepted
         let selfUser = await context.getSelfUser()
         let selfUserID = await context.selfUserID(selfUser: selfUser)
+        // The receiver is the user we are connecting with
         let username: String? = if let receiverQualifiedID = connection.receiverQualifiedID {
             await context.username(
-                for: context.getUser(userID: receiverQualifiedID)
+                for: context.getUser(
+                    id: receiverQualifiedID.uuid,
+                    domain: receiverQualifiedID.domain
+                )
+            )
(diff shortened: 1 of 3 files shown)


WireDomain/Sources/WireDomain/Notifications/Builders/User/UserConnectionEventNotificationBuilder.swift
```diff
@@ -159,11 +166,11 @@ extension UserConnectionEventNotificationBuilder {
...
 
         func getSender(
-            senderID: UserID
+            senderID: UUID
         ) async -> ZMUser {
             await userLocalStore.fetchOrCreateUser(
-                id: senderID.uuid,
-                domain: senderID.domain
+                id: senderID,
+                domain: nil
```
netbe: question: should it be nil or Backend.domain here?

netbe (approved):
looks good, left a question though

WireDomain/Sources/WireDomain/Notifications/Builders/User/UserConnectionEventNotificationBuilder.swift
jullianm: thanks for pointing that out. I messed up the logic here and should have used the receivedQualifiedID which contains the id and the domain instead of the sender which is actually self so the push notification didn't make sense. 
https://github.com/wireapp/wire-ios/pull/3076/commits/2d34085ebb2d75da610201abb12ff68b3e6eeb6b

caldrian (approved):

dmitrysimkin (approved):

PR #3077, opened 2025-05-27
fix: decoding error client add event - WPB-17753

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17753" title="WPB-17753" target="_blank">BugWPB-17753</a>  [iOS] Decoding error for client event
  </td></table>
  <br />
 

### Issue

**Context**:  A decoding error sometimes occur while trying to decode the stored update event envelope.

**Causes**: Obvious cause is that the "time" value we're trying to decode should be optional when decoding.
As to why we ended up storing an envelope with an optional "time" value is less obvious since it looks like we've been working with a non optional until now. Could have been related to migration from legacy event but looks like we don't map that event either.

**Solution**: Quick win solution is to make the "time" value optional just like in legacy code (`Payload.UserClient` creationDate property) to avoid that decoding error.

### Testing

Couldn't reproduce

### Checklist

---

### DIFF
--- a/WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Event decoding/User/UserClientAddEventDecoder.swift
+++ b/WireAPI/Sources/WireAPI/APIs/UpdateEventsAPI/Event decoding/User/UserClientAddEventDecoder.swift
@@ -32,7 +32,7 @@ struct UserClientAddEventDecoder {
             client: SelfUserClient(
                 id: payload.id,
                 type: payload.type,
-                activationDate: payload.activationDate.date,
+                activationDate: payload.activationDate?.date,
                 label: payload.label,
                 model: payload.model,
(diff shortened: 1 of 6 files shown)


KaterinaWire:
@jullianm I see "ketNotFound" from the screenshot.  And there is `case creationDate = "time"` in the `Payload.UserClient `. Maybe key should be "time" instead of "activationDate"? What do you think?

dmitrysimkin (approved):

jullianm:
> @jullianm I see "ketNotFound" from the screenshot. And there is `case creationDate = "time"` in the `Payload.UserClient `. Maybe key should be "time" instead of "activationDate"? What do you think?

@KaterinaWire we don't see it the PR but the `activationDate` coding key is indeed "time"

KaterinaWire (approved):

PR #3096, opened 2025-05-28
feat: emit granular progress when uploading a file to wire cells - WPB-17765

samwyndham:
### Issue

Wire cells uses the swift AWS SDK for uploading files to wire cells. One of the requirements is to show a progress bar while files are uploading. Unfortunately this SDK doesn't yet support tracking the progress of an upload 🤯.

The workaround is to wrap the reading of file data. This gives us a good enough approximation of upload progress. To achieve this, this PR introduces `ObservableStream`. This wraps a `Smithy.Stream` (an SDK object) and exposes an swift AsyncStream for obtaining progress.

Additionally this PR does some refactoring and adds some tests - currently WireCells is missing a lot of tests so I'm trying to add them as I go.

Changes include:

- Deleting some unused protocols as these provide no use and are not necessary for tests.
- Moving some types that are defined in `WireCellsAPI` to `WireCellsImplementation` as they don't need to be used outside that module. Also downgrading their access modifiers where possible.
- Introducing and using `ObservableStream` as discussed above
- Refactoring of some APIs to make them return AsyncSequences instead of taking progress closures.
- Moving & renaming of the types that make up the `NodesAPI` (see following note)

#### `NodesAPI`

I have renamed the following:

- `WireCellsAWSClientImplementation` -> `AWSClient`
- `WireCellsNodesAPI` -> `RestAPI`
- `WireCellsNodesDataSource` -> `NodesAPI`

`NodesAPI` forms the API for interacting with wire cells. A _node_ is a file or folder. An implementation detail of this is that some operations are done via the AWS SDK (e.g. uploading & downloading of files), and other operations are done via a wire cells rest API (e.g. moving & deleting of files). This separation is an implementation detail and as such `NodesAPI` creates a stable interface that forwards operations to the `AWSClient` or `RestAPI`.

Regarding naming, I have dropped the `WireCells` prefix as these types won't be used outside the `WireCells` package. I chose `NodesAPI` as it matches the terminology used in the `WireAPI` module - e.g. `AccountsAPI` etc.

### Testing

This code is currently not manually testable without code changes. If you wish to test it let me know.

---

### Checklist


WireCells/Sources/WireCellsAPI/WireCellsAWSClient.swift
samwyndham, line 1: This protocol is not needed

WireCells/Sources/WireCellsAPI/WireCellsNodesAPIProtocol.swift
samwyndham, line 1: This protocol is not needed

WireCells/Sources/WireCellsAPI/WireCellsNodesRepository.swift
samwyndham, line 1: This has moved to the implementation file as it is not need across modules (only for assembly)

WireCells/Sources/WireCellsImplementation/NodesAPI/NodesAPI.swift
```diff
@@ -19,44 +19,81 @@
...
 package import WireCellsAPI
 
-package final actor WireCellsNodesDataSource: WireCellsNodesRepository {
-    private let awsClient: any WireCellsAWSClient
-    private let cellsAPI: any WireCellsNodesAPIProtocol
+package enum NodesAPIError: Error {
+    case failedToCreateWriteStream
+}
+
+package protocol NodesAPIProtocol: Actor {
```
samwyndham, line 26: This protocol was moved from WireCells/Sources/WireCellsAPI/WireCellsNodesRepository.swift

dmitrysimkin: suggestion: There is a lot of methods in one protocol, i think it worth to split into smaller ones to comply with Interface Segregation principle

WireCells/Sources/WireCellsImplementation/NodesAPI/ObservableStream.swift
```diff
@@ -0,0 +1,97 @@
...
+
+import Foundation
+import Smithy
+
+/// Wraps a `Smithy.Stream` allowing its progress to be observed. This is a workaround to enable reading upload progress
+/// using the AWS SDK.
+
+final class ObservableStream: Smithy.Stream, Sendable {
+
+    private let wrapped: any Smithy.Stream
```
dmitrysimkin: minor: for a bit more readability
```suggestion
    private let wrappedStream: any Smithy.Stream
```

WireCells/Sources/WireCellsImplementation/WireCellsNodeUploadManager.swift
```diff
@@ -100,30 +101,24 @@ package final actor WireCellsNodeUploadManager: WireCellsNodeUploadManagerProtoc
...
-                        }
-                    }
-                )
+                for try await progress in upload {
+                    await self.updateUploadProgress(
+                        nodeID: node.id,
+                        uploaded: UInt64(progress),
+                        total: node.size ?? 1
+                    )
+                    continuation.yield(.uploading(progress: Float(progress) / Float(node.size ?? 1)))
```
dmitrysimkin: question: what ?? 1 means?

dmitrysimkin (commented):
Looks good, left couple of comments before approving

johnxnguyen (approved):

WireCells/Sources/WireCellsImplementation/WireCellsNodeUploadManager.swift
samwyndham: Here I was just moving around existing code. I think `1` was chosen to avoid dividing by 0. However, the caller of this class actually has a non optional size. I will update the code to pass this value through.

WireCells/Sources/WireCellsImplementation/NodesAPI/ObservableStream.swift
samwyndham: I've updated this in [954b36d](https://github.com/wireapp/wire-ios/pull/3096/commits/954b36d3d0abb80e021cb97272f80d303833bfa8)

WireCells/Sources/WireCellsImplementation/WireCellsNodeUploadManager.swift
samwyndham: Fixed in https://github.com/wireapp/wire-ios/pull/3096/commits/403a48d10e2d298e1ad4e33fb575f01be6136d3f

samwyndham:
>suggestion: There is a lot of methods in one protocol, i think it worth to split into smaller ones to comply with Interface Segregation principle

@dmitrysimkin Please let me think about this a bit more and adjust in future PRs. To me it is not clear yet what these different interfaces should be - ie. how the methods should be grouped.

dmitrysimkin:
> > suggestion: There is a lot of methods in one protocol, i think it worth to split into smaller ones to comply with Interface Segregation principle
> 
> @dmitrysimkin Please let me think about this a bit more and adjust in future PRs. To me it is not clear yet what these different interfaces should be - ie. how the methods should be grouped.

Ok

dmitrysimkin (approved):
👍

PR #3099, opened 2025-05-28
fix: app recovery update events api error - WPB-17609

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17609" title="WPB-17609" target="_blank">BugWPB-17609</a>  [iOS] App doesn't recover from 'UpdateEventsAPIError.notFound' 
  </td></table>
  <br />
 

### Issue

When pulling pending events during `IncrementalSync` and it fails with 400 or 404, it means we have missed (too many) events and app needs to recover by:
- triggering a full sync (initial + incremental)
- add a system message to let user know there is a potential message gap

Uploading ScreenRecording_05-28-2025 14-14-50_1.MP4…

### Testing

tested internally by throwing a 404 error and ensuring gap system message is added and full sync is performed.

### Checklist

### DIFF
--- a/WireDomain/Sources/WireDomain/Repositories/UpdateEvents/Protocols/UpdateEventsLocalStoreProtocol.swift
+++ b/WireDomain/Sources/WireDomain/Repositories/UpdateEvents/Protocols/UpdateEventsLocalStoreProtocol.swift
@@ -30,7 +30,7 @@ public protocol UpdateEventsLocalStoreProtocol {
     /// Stores last event ID.
     /// - parameter id: The last event ID to store.
 
-    func storeLastEventID(id: UUID)
+    func storeLastEventID(id: UUID?)
 
     /// Retrieves the index of the last event envelope.
     /// - returns: The last index event envelope.
--- a/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
+++ b/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
@@ -27,10 +27,12 @@ public struct IncrementalSync: IncrementalSyncProtocol {
     private let pushChannelAPI: any PushChannelAPI
     private let updateEventsSync: any PullPendingUpdateEventsSyncProtocol
     private let decryptor: any UpdateEventDecryptorProtocol
-    private let store: any UpdateEventsLocalStoreProtocol
+    private let updateEventsStore: any UpdateEventsLocalStoreProtocol
+    private let messageStore: any MessageLocalStoreProtocol
     private let processor: any UpdateEventProcessorProtocol
     private let databaseSaver: any DatabaseSaverProtocol
     private let syncStateSubject: CurrentValueSubject<SyncState, Never>
+    private let onMissedEvents: () -> Void
     private let logger = WireLogger.sync
 
     public init(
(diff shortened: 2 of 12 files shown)


WireDomain/Sources/WireDomain/Repositories/UpdateEvents/Protocols/UpdateEventsLocalStoreProtocol.swift
```diff
@@ -30,7 +30,7 @@ public protocol UpdateEventsLocalStoreProtocol {
     /// Stores last event ID.
     /// - parameter id: The last event ID to store.
 
-    func storeLastEventID(id: UUID)
+    func storeLastEventID(id: UUID?)
```
netbe, line 33: question: why is it optional?
thought(non-blocking): maybe we should get a resetLastEventID instead

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -75,7 +81,24 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
+            switch error {
+            case let apiError as UpdateEventsAPIError:
+                switch apiError {
+                case .notFound, .invalidParameters:
+                    logger.debug("too many missed events.. recovering")
+                    // nullifying the last event ID since we missed events and we want to
+                    // reset with a full sync (initial + incremental)
+                    updateEventsStore.storeLastEventID(id: nil)
+                    try await messageStore.addPotentialGapSystemMessage()
+                    onMissedEvents()
```
johnxnguyen, line 94: suggestion: I think it would be better that the incremental sync throws an error like `missingEvents`, which is handled by the `SyncAgent` that would then trigger the initial sync. For me the `IncrementalSync` should just try to do what it's ask to do and report any failures, where as the `SyncAgent` is responsible for starting and stopping various syncs. What do you think?

jullianm: that was my initial thought as well but then I figured in that case the SyncAgent would not only be responsible of high level sync operations (starting / stopping syncs) but also add potential gap system message and clean up the last event ID which I think is outside of its responsibilities because way too specific. 

In my opinion, we should report to the SyncAgent only for sync operations and keep the more specific logic closer to the object its related to  (i.e add potential system message, clean up the last event ID).

netbe: @johnxnguyen For what is worth, on async stream I don't throw an error (since it would close the websocket and avoid acknowledgement of event) instead I use a delegate to inform the SyncAgent to do the initialSync then I can continue on my flow.

netbe (approved):
Looks good to me, left a comment

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
johnxnguyen: I think the incremental can add the system message and then throw the error so the sync agent can perform the initial sync. Doing so should update the last event id already as a first step so there's no cleanup really necessary, right?

@netbe I would say that closing the push channel to perform the initial sync and then reopen the push channel is the behavior we want. Or is there any issue in doing this? Maybe one I can think of is that it may create a window of time where events are pushed to the NSE.

netbe:
@jullianm moving to 4.0.0

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
johnxnguyen: For me, the incremental sync can still clear the last event id and add the system message, but I think it should throw the error and let the `SyncAgent` handle it by triggering the appropriate sync.

netbe: so maybe better not to close the pushChannel and wait until initial sync is done

jullianm: thanks for the feedbacks, I'll close this PR and open a new one and do the changes from 4.0 as there's too much discrepancies between branches now.

PR #3100, opened 2025-05-28
fix: message sending stuck - WPB-17866

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17866" title="WPB-17866" target="_blank">BugWPB-17866</a>  [iOS] Not able to send messages on mobile data
  </td></table>
  <br />
 

### Issue

**Context**: While in foreground and switching Wifi off, sending a message gets stuck in sending without the ability to retry (button doesn't show up) and when turning the Wifi on again, it still gets stuck. It only gets resolved when going to the background and foreground again.

**Causes**: When switching Wifi off, the live events Task finishes and we update the sync state to .idle which turns the decryption state to `.notStarted` and the MessageSender gets stuck again waiting for the decryption state to be `.done`. 

**Solution**: Don't update the state in the live events Task. If we reach this point, we know the decryption was already done (current state) and the decryption state should remain to .done (done should always be the final state) to not block any message sending.

### Testing

- open the app, keep in foreground, send message it works, switch Wifi off, send message, retry button now shows up, switch Wifi on again, hit retry, message gets sent. 

### Checklist

---

### DIFF
--- a/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
+++ b/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
@@ -80,7 +80,7 @@ public struct IncrementalSync: IncrementalSyncProtocol {
 
         let liveEventTask = Task { @Sendable [self] in
             logger.debug("handling live event stream")
-            syncStateSubject.send(.liveSyncing)
+            syncStateSubject.send(.liveSyncing(.ongoing))
 
             await processLiveEvents(
                 liveEventStream: liveEventStream,
@@ -88,6 +88,7 @@ public struct IncrementalSync: IncrementalSyncProtocol {
             )
 
             logger.debug("live event stream did finish")
+            syncStateSubject.send(.liveSyncing(.terminated))
         }
 
         return Token(task: liveEventTask, closePushChannel: {
--- a/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
+++ b/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
@@ -88,7 +88,7 @@ public struct IncrementalSync: IncrementalSyncProtocol {
             )
 
             logger.debug("live event stream did finish")
-            syncStateSubject.send(.liveSyncing(.terminated))
+            syncStateSubject.send(.liveSyncing(.finished))
         }
 
         return Token(task: liveEventTask, closePushChannel: {
--- a/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
+++ b/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
@@ -88,7 +88,6 @@ public struct IncrementalSync: IncrementalSyncProtocol {
             )
 
             logger.debug("live event stream did finish")
-            syncStateSubject.send(.idle)
         }
 
         return Token(task: liveEventTask, closePushChannel: {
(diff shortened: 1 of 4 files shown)


dmitrysimkin (approved):
question: need to add test?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -88,7 +88,6 @@ public struct IncrementalSync: IncrementalSyncProtocol {
             )
 
             logger.debug("live event stream did finish")
-            syncStateSubject.send(.idle)
```
jullianm, line 91: removing it at this level which causes the issue

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -105,6 +105,8 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
     /// This method logs any errors and does not wait for the sync to finish.
 
     func resume() {
+        syncStateSubject.send(.idle)
```
jullianm, line 111: adding it here to make sure when we resume again and we perform the incremental sync message sender waits again for decryption to finished, it was missing in previous PR which could be the root cause of other recent issues.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
johnxnguyen: issue: if the stream finishes and we're still in the FG we should try to do the incremental sync again. Wouldn't that also fix getting stuck?

jullianm: so.. if I got this right, when trying to do the incremental sync again, either it will fail and we already catch the error so the decryption state will be resolved to done thus unlocking the message sender or it works and decryption state will also be resolved to done, in any cases message sender is unlocked.. it makes sense I'll do the changes thanks

jullianm: done in https://github.com/wireapp/wire-ios/pull/3100/commits/6fc83fe411340dbf7e8b6cae5bea2e6aff530e7c

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -80,15 +80,15 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
+            syncStateSubject.send(.liveSyncing(.ongoing))
 
             await processLiveEvents(
                 liveEventStream: liveEventStream,
                 processedEnvelopeIDs: processedEnvelopeIDs
             )
 
             logger.debug("live event stream did finish")
-            syncStateSubject.send(.idle)
+            syncStateSubject.send(.liveSyncing(.terminated))
```
netbe: question: do you mean here the sync finished because websocket was closed?

jullianm: for the specific issue we had (when switching Wifi off) the websocket is terminated due to the error 57 "Socket is not connected" but I figured we wanted to have a more global naming that the live sync was terminated (for whatever reason)

netbe (approved):
approving expecting .terminated to be renamed in .finished

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
jullianm: renamed to `.finished` in last commit

PR #3105, opened 2025-06-02
feat: update protobufs to include multipart message - WPB-17881

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17881" title="WPB-17881" target="_blank">TaskWPB-17881</a>  [iOS] Update data model to support message with attachments
  </td></table>
  <br />
 

### Issue

This PR updates our protofbuf library to `v1.52.0` so that we have access to the new `multipart` message type. This message type combines an optional message text with an array of file assets. It is the type that is used when sending files using Wire Cells.

#### Reviewing note

Most of the code changes are generated code. No need to check the `.pb.swift` files closely.

### Testing

Run the app and test message sending / receiving is working using a variety of message types - i.e nothing is broken by this PR. It is not currently possible to test the new multipart message type.

---

### Checklist

### DIFF
--- a/Cartfile.private
+++ b/Cartfile.private
@@ -1,2 +1,2 @@
 # used by Wire-Proto, this is not compiled but just checked out by Carthage.
-github "wireapp/generic-message-proto" "v1.51.0"
+github "wireapp/generic-message-proto" "v1.52.0"
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,5 +1,5 @@
 github "wireapp/ZipArchive" "v2.4.2"
 github "wireapp/cryptobox-ios" "v1.1.0_xcframework_arm64simulator"
-github "wireapp/generic-message-proto" "v1.51.0"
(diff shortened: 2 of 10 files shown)


Cartfile.private
```diff
@@ -1,2 +1,2 @@
 # used by Wire-Proto, this is not compiled but just checked out by Carthage.
-github "wireapp/generic-message-proto" "v1.51.0"
+github "wireapp/generic-message-proto" "v1.52.0"
```
netbe, line 2: suggestion: as mentioned in dev channel, I would suggest to remove from here and add this as submodule instead, which would be one step further to remove carthage altogether

netbe (approved):
looks good, I left a suggestion please have a look if that makes sense

jullianm (approved):

Cartfile.private
samwyndham: Merging this now for time saving but will address your comment in a follow up PR.

PR #3112, opened 2025-06-03
fix: use wrong user ID - WPB-18006

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18006" title="WPB-18006" target="_blank">BugWPB-18006</a>  [iOS] Updating the group admin toggle for participants from Web doesn't reflect on iOS
  </td></table>
  <br />
 

### Issue

**Context**: When turning a member into an admin (via the toggle) from Web, the changes aren't reflect on iOS, nothing happens.

**Causes**: We're not using the correct user ID, so the changes are targeting the user that initiated the changes whereas it should be targeting the user that is impacted by these changes.

**Solution**: Use the correct user ID.

### Testing

1. User B creates a group conversation from Web 
2. Adds another participant (user A on iOS)
3. Toggle on to make this participant an admin
4. on iOS, user A should be marked as admin in conversation details and toggling on/off on web should immediately be reflected on iOS 

### Checklist


netbe (approved):
looks good, could be nice to add a UT for it too

dmitrysimkin (approved):
question: need to be covered by new test?

👍

PR #3113, opened 2025-06-03
fix: participants added system message wrong position - WPB-18005

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18005" title="WPB-18005" target="_blank">BugWPB-18005</a>  [iOS] System message participants added at wrong position in group conversation
  </td></table>
  <br />
 

### Issue

**Context**: participants added system message appear to be sometimes wrongly positioned in the conversation list, specifically, one could add participant then conversation goes on and the system message appear after last messages.

**Causes**: We omitted to pass the timestamp for this specific system message which is most likely the root cause of this issue as it will take the Date.now as default value if we don't pass anything.

**Solution**: Pass the timestamp so the system message timestamp is correctly set.

### Testing

_Describe how to test._

---

### Checklist

---


netbe (approved):
LGTM;)

dmitrysimkin (approved):
👍

PR #3114, opened 2025-06-03
fix: weird transition conversation going offline + background to foreground - WPB-17982

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17982" title="WPB-17982" target="_blank">BugWPB-17982</a>  [iOS] App goes abruptly from conversation to conversation list 
  </td></table>
  <br />
 

### Issue

**Context**: When app is opened on a specific conversation, switch Wifi off then go to background, return to foreground. The app abruptly goes back to conversation list whereas it should remain on the conversation it was before.

**Causes**: The app transition router ends up in a weird state when showing the debug sync error alert multiple times and causes that abrupt transition to the conversation list. The sync error alert is intended for debug builds only however the problem is that where we check that developer flag is enabled in code should be higher in the hierarchy (basically where it's initially called) so there could be a risk that issue also happens in release builds.

**Solution**:  Move the developer flag check higher in the hierarchy so we don't start anything if it is on. This ensures we don't end up with that weird transition in production. 

### Testing

In debug builds: issue would still occur but that's fine, not really blocking anything, just an awkward transition.
In release builds: be on a conversation, switch Wifi off, go to background and back to foreground, still on the same conversation (it should not transition to the conversation list)

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
+++ b/wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
@@ -36,6 +36,24 @@ public extension Bundle {
     @objc var appGroupIdentifier: String? {
         bundleIdentifier.map { "group." + $0 }
     }
+    
+    static var developerModeEnabled: Bool {
+        Bundle.appMainBundle.infoForKey("EnableDeveloperMenu") == "1"
+    }
+    
+    private static var appMainBundle: Bundle {
+        let mainBundle: Bundle
+        let runningInExtension = Bundle.main.bundlePath.hasSuffix(".appex")
+        if runningInExtension {
+            let extensionBundleURL = Bundle.main.bundleURL
+            let mainAppBundleURL = extensionBundleURL.deletingLastPathComponent().deletingLastPathComponent()
+            guard let bundle = Bundle(url: mainAppBundleURL) else { fatalError("Failed to find main app bundle") }
+            mainBundle = bundle
+        } else {
+            mainBundle = .main
+        }
+        return mainBundle
+    }
 }
 
 @objc
--- a/wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
+++ b/wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
@@ -36,11 +36,11 @@ public extension Bundle {
     @objc var appGroupIdentifier: String? {
         bundleIdentifier.map { "group." + $0 }
     }
-    
+
     static var developerModeEnabled: Bool {
         Bundle.appMainBundle.infoForKey("EnableDeveloperMenu") == "1"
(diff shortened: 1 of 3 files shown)


wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1155,20 +1155,22 @@ extension ZMUserSession: SyncAgentDelegate {
     }
 
     func syncAgentDidFailSyncing(_ syncAgent: SyncAgent, error: any Error) {
-        let onRetry: () -> Void = { [weak self] in
-            self?.managedObjectContext.performGroupedBlock {
-                self?.isPerformingSync = true
-                self?.updateNetworkState()
+        if Bundle.developerModeEnabled { // Only show sync error alert for debugging
```
netbe, line 1158: question: is the error here logged in production build?

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -36,6 +36,24 @@ public extension Bundle {
...
+        if runningInExtension {
+            let extensionBundleURL = Bundle.main.bundleURL
+            let mainAppBundleURL = extensionBundleURL.deletingLastPathComponent().deletingLastPathComponent()
+            guard let bundle = Bundle(url: mainAppBundleURL) else { fatalError("Failed to find main app bundle") }
+            mainBundle = bundle
+        } else {
+            mainBundle = .main
+        }
+        return mainBundle
+    }
```
netbe, line 56: todo: to remove, it's not used, right?

netbe (commented):
the fix looks good, though some things seem not necessary (see comments)

dmitrysimkin (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
jullianm: yes, we don't see it here but right below we log it : `WireLogger.sync.error("failed to perform sync: \(String(describing: error))")`

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
jullianm: I added this code in this PR to be able to access the flag, it's only available at the app level

netbe: oh sorry I didn't see it was under Bundle extension

netbe (approved):
LGTM;)

PR #3120, opened 2025-06-04
fix: wrong system message user left - WPB-17642

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17642" title="WPB-17642" target="_blank">BugWPB-17642</a>  [iOS] Participant leaving incorrectly says removed
  </td></table>
  <br />
 

### Issue

**Context**: When user left a group conversation by themselves, other user see a system message "user was removed by the team", it should see the system message "user left"

**Causes**: We don't use the correct system message for the .userLeft case

**Solution**: Use .participantsRemoved case from ZMSystemMessage

### Testing

- Have a user A using iOS, and a User B on any
- Log in as both A and B and make sure they are in a group together
- Have B leave the group themselves
- Look at system message with A, it should be "User B left"

### Checklist

---


netbe (approved):
LGTM;) could add a UT for it too

samwyndham (approved):
LGTM

PR #3125, opened 2025-06-04
fix: extra dash missed call message - WPB-17974

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17974" title="WPB-17974" target="_blank">BugWPB-17974</a>  [iOS] Missed call indicator has an extra "-"
  </td></table>
  <br />
 

### Issue

**Context**: The missed call system message in a conversation is followed by a cell that only contains an extra dash "-" whenever we don't have a timestamp

**Solution**: unless I'm missing some context and there's a specific reason why we do that simply return an empty string instead of "-" if we don't have a timestamp

### Testing

_Describe how to test._

---

### Checklist

---


netbe (approved):
LGTM;)

dmitrysimkin (approved):

PR #3144, opened 2025-06-06
fix: read receipts not working - WPB-18016

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18016" title="WPB-18016" target="_blank">BugWPB-18016</a>  [iOS] : No read receipts
  </td></table>
  <br />
 

### Issue

Read receipts are not working in the new sync. This is because we are currently ignoring the message confirmation type update event, which is used to create `ZMMessageConfirmation` objects which is the source of our read receipt data. 

This PR fixes that so that the confirmation update events are no longer ignored.

### Testing

1. Post a message in a 1:1 or group proteus conversation (MLS groups don't have read receipts)
2. Have another user read that message.
3. Confirm that the read receipt (eye icon) for the message is visible.

---

### Checklist

### DIFF
--- a/wire-ios-data-model/Source/Model/Confirmation/ZMMessageConfirmation.swift
+++ b/wire-ios-data-model/Source/Model/Confirmation/ZMMessageConfirmation.swift
@@ -69,7 +69,6 @@ open class ZMMessageConfirmation: ZMManagedObject, ReadReceipt {
     ) -> [ZMMessageConfirmation] {
 
         guard
-            let managedObjectContext = conversation.managedObjectContext,
             let senderUUID = updateEvent.senderUUID,
             let timestamp = updateEvent.timestamp
         else {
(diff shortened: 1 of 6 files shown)


wire-ios-data-model/Source/Model/Confirmation/ZMMessageConfirmation.swift
```diff
@@ -68,17 +68,40 @@ open class ZMMessageConfirmation: ZMManagedObject, ReadReceipt {
...
+            confirmation,
+            conversation: conversation,
+            senderUUID: senderUUID,
+            senderDomain: updateEvent.senderDomain,
+            timestamp: timestamp
+        )
+    }
+
+    @discardableResult
+    public static func createMessageConfirmations(
```
samwyndham, line 88: I created a new `createMessageConfirmations` as the old one requires `ZMUpdateEvent` which new sync doesn't have. The old method then uses the new version.

caldrian (approved):

johnxnguyen (approved):

PR #3152, opened 2025-06-10
fix: app recovery update events error - WPB-17609

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17609" title="WPB-17609" target="_blank">BugWPB-17609</a>  [iOS] App doesn't recover from 'UpdateEventsAPIError.notFound' 
  </td></table>
  <br />
 

### Issue

follow up on  https://github.com/wireapp/wire-ios/pull/3099

When pulling pending events during IncrementalSync and it fails with 400 or 404, it means we have missed (too many) events and app needs to recover by:

triggering a full sync (initial + incremental)
add a system message to let user know there is a potential message gap

### Testing

tested internally by throwing a 404 error and ensuring gap system message is added and full sync is performed.

### Checklist

### DIFF
--- a/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
+++ b/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
@@ -22,8 +22,8 @@ import WireAPI
 import WireLogging
 
 public struct IncrementalSync: IncrementalSyncProtocol {
-    
-    enum Failure: Error {
+
+    public enum Failure: Error {
         case missedEvents
     }
 
@@ -49,8 +49,7 @@ public struct IncrementalSync: IncrementalSyncProtocol {
         processor: any UpdateEventProcessorProtocol,
         databaseSaver: any DatabaseSaverProtocol,
         syncStateSubject: CurrentValueSubject<SyncState, Never>,
-        journal: Journal,
-        onMissedEvents: @escaping () -> Void
+        journal: Journal
     ) {
         self.selfClientID = selfClientID
         self.pushChannelAPI = pushChannelAPI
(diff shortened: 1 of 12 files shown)


WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -76,9 +83,29 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
+            switch error {
+            case let apiError as UpdateEventsAPIError:
+                switch apiError {
+                case .notFound, .invalidParameters:
+                    // nullifying the last event ID since we missed events and we want to
+                    // reset with a full sync (initial + incremental)
+                    updateEventsStore.resetLastEventID()
+                    try await messageStore.addPotentialGapSystemMessage()
+                    await tearDown()
+                    throw Failure.missedEvents
```
jullianm, line 100: throwing the error

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -219,7 +219,14 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
...
                 syncStateSubject.send(.suspended)
-                throw error
+
+                if let incrementalSyncError = error as? IncrementalSync.Failure,
+                   case .missedEvents = incrementalSyncError {
+                    // recovering with a full sync
+                    resume()
+                } else {
+                    throw error
+                }
```
johnxnguyen: suggestion: catch the error specifically:

```swift
} catch IncrementalSync.Failure.missedEvents {
    // ....
}
```

Also... it would be clearer to trigger the initial sync explicitly `performInitialSync`, but I guess we want to call `resume` because of the task cancellation logic? It'd be nice if we could use the logic more generically with any of the sync methods.

johnxnguyen (approved):

netbe (approved):
LGTM;)

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
jullianm: added the suggestion in https://github.com/wireapp/wire-ios/pull/3152/commits/343055cba8ca4c8a6b3d39600864b1785a6a5f7c 

yes, we call `resume()` to reuse the existing logic that performs a full sync (initial + incremental) and is currently our main entry point for that purpose. Overall, I think we could improve these different flows a little bit by introducing something like this:

```
public var triggerSyncFlow: PassthroughSubject<SyncFlow>

enum SyncFlow {
    case initial
    case incremental
    case resources
    case full
}

```
.. where triggerSyncFlow would be our only entry point to perform the sync we want from inside or outside the class. Any of the flows could be run independently (with all the logic it needs in terms of cancellation, recovery etc)

PR #3165, opened 2025-06-11
fix: unread badges count - WPB-17699

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17699" title="WPB-17699" target="_blank">BugWPB-17699</a>  [iOS] Unread badge counts incorrectly
  </td></table>
  <br />
 

### Issue

**Context**: App badges count doesn't update properly when receiving a notification on a conversation. 

**Causes**: When increasing / decreasing the internal estimated unread count on a `ZMConversation` we omit to save the changes to the store.

**Solution**: Explicitly save changes ((if any) to the store using the `DatabaseSaver`.

### Testing

Have several conversations opened and send messages with user B
user A in same conversation with app in background will receive notifications and should see the app badges count increasing / decreasing properly depending on the notification (i.e badges count would increase)

### Checklist

### DIFF
--- a/WireDomain/Sources/WireDomain/Notifications/GenerateNotificationUseCase.swift
+++ b/WireDomain/Sources/WireDomain/Notifications/GenerateNotificationUseCase.swift
@@ -47,7 +47,7 @@ struct GenerateNotificationUseCase: GenerateNotificationUseCaseProtocol {
                 }
             }
         }
-        
+
         // Ensures unread conversations count is up-to-date.
         try await databaseSaver.save()
 
--- a/WireDomain/Sources/WireDomain/Notifications/GenerateNotificationUseCase.swift
+++ b/WireDomain/Sources/WireDomain/Notifications/GenerateNotificationUseCase.swift
@@ -33,6 +33,7 @@ struct GenerateNotificationUseCase: GenerateNotificationUseCaseProtocol {
 
     let conversationEventBuilder: any ConversationEventNotificationBuilderProtocol
     let userEventBuilder: any UserEventNotificationBuilderProtocol
+    let databaseSaver: any DatabaseSaverProtocol
     let eventID: UUID
 
     /// Processes the events stream.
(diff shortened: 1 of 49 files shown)


caldrian (approved):

WireDomain/Sources/WireDomain/Notifications/GenerateNotificationUseCase.swift
```diff
@@ -47,6 +48,9 @@ struct GenerateNotificationUseCase: GenerateNotificationUseCaseProtocol {
             }
         }
 
+        // Ensures unread conversations count is up-to-date.
+        try await databaseSaver.save()
```
netbe: question: where is unread conversations count calculated?

netbe (commented):
left a comment before approving

WireDomain/Sources/WireDomain/Notifications/GenerateNotificationUseCase.swift
jullianm: in `ShowNotificationUseCase` right before displaying the notification

```
    private func getNotificationBadge() async -> NSNumber {
        let unreadConversationCount = await Int(
            conversationLocalStore.unreadConversationCount()
        )

        selectedAccount.unreadConversationCount = unreadConversationCount
        let totalUnreadCount = accountManager.totalUnreadCount

        return NSNumber(value: totalUnreadCount)
    }
```

netbe: why don't we save there then?

jullianm: done in https://github.com/wireapp/wire-ios/pull/3165/commits/1ac38a44f21b498d70144b876151e02d832c8727 thx

netbe (approved):
LGTM;) 🚀

PR #3168, opened 2025-06-11
fix: missing sync bar - WPB-17717

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17717" title="WPB-17717" target="_blank">BugWPB-17717</a>  [iOS] Sync bar missing with new Sync
  </td></table>
  <br />
 

### Issue

**Context**:  Sync bar seems to be missing when the sync is not completely done

**Causes**: When the incremental sync is done, we may have many pending MLS groups to join (see `performPendingJoins()`)  which can take some time to complete however we hide the sync bar too early so it looks like the app is synced and ready to be used whereas we still see group conversations being added and ordered in the conversation list (as shown in the video attached to the ticket and also reproducible on my side)

**Solutions**: Ensure we hide the sync bar only after all these operations are completed (see `func didFinishIncrementalSync(isRecovering: Bool)` in `ZMUserSession`.

### Testing

Not easily reproducible but if you have many pending MLS groups to join, you wouldn't see any sync bar, looking like app is ready to be used but then gradually see group conversations being added to the list.

With the above fix, we do see the sync bar and it hides when all group conversations are there.

### Checklist

### DIFF
--- a/wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
+++ b/wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
@@ -1137,6 +1137,16 @@ extension ZMUserSession: SyncAgentDelegate {
         syncContext.performGroupedBlock { [weak self] in
             guard let self else { return }
             WireLogger.sync.debug("did finish incremental sync")
+
+            func showSyncBar(_ show: Bool) {
+                managedObjectContext.performGroupedBlock { [weak self] in
+                    self?.isPerformingSync = show
+                    self?.updateNetworkState()
+                }
+            }
+
+            showSyncBar(true)
+
             processLegacyEvents()
 
             NotificationInContext(
@@ -1146,7 +1156,7 @@ extension ZMUserSession: SyncAgentDelegate {
 
             guard !isRecovering else {
                 // in case of recovery, we don't need more
-                return
+                return showSyncBar(false)
             }
 
             WaitingGroupTask(context: syncContext) { [weak self] in
@@ -1179,6 +1189,8 @@ extension ZMUserSession: SyncAgentDelegate {
 
                 await calculateSelfSupportedProtocolsIfNeeded()
                 await resolveOneOnOneConversationsIfNeeded()
+
+                showSyncBar(false)
             }
 
             recurringActionService.performActionsIfNeeded()
@@ -1278,11 +1290,6 @@ extension ZMUserSession: SyncAgentDelegate {
     }
 
     func processLegacyEvents() {
-        managedObjectContext.performGroupedBlock { [weak self] in
-            self?.isPerformingSync = true
-            self?.updateNetworkState()
-        }
-
         let groups = syncContext.enterAllGroupsExceptSecondary()
         Task {
             var processingInterrupted = false
@@ -1302,10 +1309,6 @@ extension ZMUserSession: SyncAgentDelegate {
                 }
             }
 
-            await managedObjectContext.perform { [weak self] in
-                self?.isPerformingSync = isSyncing || processingInterrupted
-                self?.updateNetworkState()
-            }
             self.syncContext.leaveAllGroups(groups)
         }
     }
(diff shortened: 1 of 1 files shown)


wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1137,6 +1137,16 @@ extension ZMUserSession: SyncAgentDelegate {
...
             WireLogger.sync.debug("did finish incremental sync")
+
+            func showSyncBar(_ show: Bool) {
+                managedObjectContext.performGroupedBlock { [weak self] in
+                    self?.isPerformingSync = show
+                    self?.updateNetworkState()
+                }
+            }
+
+            showSyncBar(true)
```
netbe, line 1148: issue: should we show false instead, Sync bar represent the time when we sync so incremental sync. Not post actions we do after incremental sync is done

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1146,7 +1156,7 @@ extension ZMUserSession: SyncAgentDelegate {
 
             guard !isRecovering else {
                 // in case of recovery, we don't need more
-                return
+                return showSyncBar(false)
```
netbe, line 1157: issue: followup, I think this is not needed

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1137,6 +1137,16 @@ extension ZMUserSession: SyncAgentDelegate {
...
+            func showSyncBar(_ show: Bool) {
+                managedObjectContext.performGroupedBlock { [weak self] in
+                    self?.isPerformingSync = show
+                    self?.updateNetworkState()
+                }
+            }
+
+            showSyncBar(true)
+
             processLegacyEvents()
```
netbe: todo: we should skip processLegacyEvents for syncV2

KaterinaWire (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1278,11 +1290,6 @@ extension ZMUserSession: SyncAgentDelegate {
     }
 
     func processLegacyEvents() {
-        managedObjectContext.performGroupedBlock { [weak self] in
```
netbe, line 1281: suggestion: avoid executing processLegacyEvents when syncV2 here

This is called from other places.

netbe (commented):
As the fix might work, I think there's a better solution here, happy to discuss it

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
netbe: as discussed. add comment to follow up ticket, we only need to add sync bar as we don';t performPendingJoins during slowSync

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
jullianm: done

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
jullianm: done

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
jullianm: added comment thx

netbe (approved):
🚀

PR #3171, opened 2025-06-12
feat: send multipart attachment - WPB-18009

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18009" title="WPB-18009" target="_blank">TaskWPB-18009</a>  [iOS] Create and hook up skeleton AppendMultipartMessageUseCase
  </td></table>
  <br />
 

### Issue

Apologies for this large and somewhat unfocused PR!

This PR implements sending multipart messages - these are messages for wire cells that can contain optional text and an array of file attachments.

Additionally it fixes not passing the wire cell name to the use cases. The cell name is kind of like the root folder that the files belong too. Currently it is just the qualified id of the conversation.

To achieve the above, this PR:

- Updates `WireCellsDraft` with more fields required for sending.
- Updates `WireCellsNode` and similar models not to have a version ID.
- Introduced a WireCellsClearPublishedDraftsUseCaseProtocol which removes all publish drafts from the drafts repository.
- Fixes a bug where drafts repo doesn't emit it's latest values to observers.
- Adds a bunch of glue code.

This PR also closes https://wearezeta.atlassian.net/browse/WPB-17880

### Testing

1. Configure the app for using wire cells as outlined [here](https://wearezeta.atlassian.net/wiki/spaces/PAD/pages/1948680479/2025-06-12+Wire+Cells+iOS+configuration).
2. Go to a conversation.
3. Type some text
4. Add some files via the attachment option - other options (such as taking a photo) are not currently supported.
5. Send the message
6. A dummy message should appear in the conversation.

---

### Checklist

---


WireCells/Sources/WireCellsAPI/Model/WireCellsNode.swift
```diff
@@ -41,8 +41,7 @@ public struct WireCellsPublicLinkID: Codable, Equatable, Hashable, Sendable {
 }
 
 public struct WireCellsNode: Equatable, Identifiable, Sendable {
-    public let id: WireCellsNodeID
-
+    public let id: UUID
```
samwyndham, line 30: `WireCellsNodeID` contains a version id that is not needed after uploading a file and is generally not available when fetching nodes from the backend. Therefore id is now just a UUID and if we need a version ID this is passed separately.

WireCells/Sources/WireCellsImplementation/DraftsRepository.swift
```diff
@@ -102,7 +110,7 @@ package actor DraftsRepository: DraftsRepositoryProtocol {
         let continuationID = UUID()
         let (stream, continuation) = AsyncStream.makeStream(
             of: [WireCellsDraft].self,
-            bufferingPolicy: .bufferingOldest(0)
+            bufferingPolicy: .bufferingNewest(1)
```
samwyndham, line 115: I misunderstood buffering policies. The new buffering policy means that the stream only caches the latest 1 value which is what we want.

WireCells/Sources/WireCellsImplementation/DraftsRepository.swift
```diff
@@ -28,6 +28,7 @@ package protocol DraftsRepositoryProtocol: Actor {
     func add(assetURL: URL, assetSize: Int, cellName: String, fileName: String, fileType: UTType?) async
     func drafts(for cellName: String) -> AsyncStream<[WireCellsDraft]>
     func publishAll(for cellName: String) async throws
+    func clearPublished(for cellName: String)
```
dmitrysimkin: suggestion: not clear published what, `clearPublishedDrafts` is more precise, wdyt @samwyndham ?

wire-ios-data-model/Source/Model/Message/MultipartAttachment.swift
```diff
@@ -0,0 +1,128 @@
...
+                asset.contentType = contentType
+
+                // Only set if values are not nil to avoid protobufs setting nonsense defaults.
+                initialName.map { asset.initialName = $0 }
+                initialSize.map { asset.initialSize = Int64($0) }
+
+                if let initialMetadata {
+                    switch initialMetadata {
+                    case let .image(width, height):
+                        asset.initialMetaData = .image(
```
dmitrysimkin: minor: guard to avoid one level of nesting

wire-ios-sync-engine/Source/Use cases/AppendMultipartMessageUseCase.swift
```diff
@@ -0,0 +1,64 @@
...
+        replyingTo: ZMConversationMessage?,
+        in conversation: ZMConversation,
+        fetchLinkPreview: Bool,
+        attachments: [MultipartAttachment]
+    ) throws
+}
+
+public struct AppendMultipartMessageUseCase: AppendMultipartMessageUseCaseProtocol {
+
+    weak var analyticsEventTracker: (any AnalyticsEventTrackerProtocol)?
```
dmitrysimkin: question: why weak? I think life time of tracker should be at least same as use case, I'd expect not weak property for it

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+SendButton.swift
```diff
@@ -42,9 +45,28 @@ extension ConversationInputBarViewController {
             editingMessage = nil
             updateWritingState(animated: true)
         } else {
+            if !attachments.isEmpty {
+                do {
+                    try await wireCellsPublishDraftsUseCase.invoke()
+                    await wireCellsClearPublishedDraftsUseCase.invoke()
```
dmitrysimkin: question: do we need `wireCells...` prefix for use cases? I'd drop them

dmitrysimkin (commented):
Looks good overall, nice work! 
left couple of questions before approving

WireCells/Sources/WireCellsAPI/Model/WireCellsNode.swift
netbe: question: should this information be documented

WireCells/Sources/WireCellsAPI/Persistence/WireCellsMessageAttachmentDraftRepository.swift
```diff
@@ -17,7 +17,7 @@
 //
 
 import CellsSDK
-import Foundation
+public import Foundation
```
netbe, line 20: question: I wonder why this is a public import?

WireCells/Sources/WireCellsBindings/WireCellsAssembly.swift
```diff
@@ -46,4 +46,12 @@ public struct WireCellsAssembly {
         ObserveDraftsUseCase(cellName: cellName, draftRepository: Self.draftsRepository)
     }
 
+    public func makePublishDraftsUseCase(cellName: String) -> any WireCellsPublishDraftsUseCaseProtocol {
+        PublishDraftsUseCase(cellName: cellName, draftRepository: Self.draftsRepository)
```
netbe, line 50: question: what's a published draft? published where? and what does it contain?

netbe (commented):
left a few comments to fully understand it before approving

WireCells/Sources/WireCellsBindings/WireCellsAssembly.swift
samwyndham: When sending a file using wire cells we first upload it. Only after it is uploaded can we send the message. Initially it is in a draft state after upload. This means that it does not appear in searches, or when viewing all files in a conversation. When we tap send, we **publish** the draft. It is now visible in the conversation and searches etc. I will add some documentation to the PublishDraftsUseCaseProtocol explaining this

WireCells/Sources/WireCellsAPI/Persistence/WireCellsMessageAttachmentDraftRepository.swift
samwyndham: @netbe it is necessary because one of the public protocols now takes a UUID as a parameter - without the `public` there is a compile error. In an upcoming feature imports are _internal_ by default. In the `WireCells` module and others we have `InternalImportsByDefault` upcoming feature flag enabled. More info here: https://www.polpiella.dev/swift-6-import-access-level

wire-ios-sync-engine/Source/Use cases/AppendMultipartMessageUseCase.swift
samwyndham: I was following the pattern used in the other append message use cases, but you are right. It doesn't really makes sense. I think it might be there because how analytics is disabled - when that happens the tracker itself is expected to be deallocated. But this doesn't make sense because if analytics is enabled again this tracker will no longer exist.

So the issue is if I don't make it weak I _think_ it may cause a bug with disabling tracking. I don't want to fix this now but have added a comment in [d0bc403](https://github.com/wireapp/wire-ios/pull/3171/commits/d0bc40326bc36c766ee0002181621394ebe56ddc) to look at that when implementing tracking in this use case.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+SendButton.swift
samwyndham: Good idea. Fixed in [e936eb7](https://github.com/wireapp/wire-ios/pull/3171/commits/e936eb7bbe8b7fa3a836c48ec4b3095b38e63df6)

wire-ios-data-model/Source/Model/Message/MultipartAttachment.swift
samwyndham: Addressed in https://github.com/wireapp/wire-ios/pull/3171/commits/9d7c1b527831d0657f6558c745cd1fdad7067def

WireCells/Sources/WireCellsBindings/WireCellsAssembly.swift
samwyndham: I've added some doc comments to `WireCellsDraft` to help explain in https://github.com/wireapp/wire-ios/pull/3171/commits/45ca769008fb4d26838c554b5a17f39d53f71767

WireCells/Sources/WireCellsAPI/Model/WireCellsNode.swift
samwyndham: I've now completely deleted `WireCellsNodeID` in https://github.com/wireapp/wire-ios/pull/3171/commits/45ca769008fb4d26838c554b5a17f39d53f71767 as it doesn't model the problem well - one node can have many versions so WireCellsNodeID which was a node id + version id doesn't work. Instead I've added version id to WireCellsDraft and added a little docs there.

WireCells/Sources/WireCellsImplementation/DraftsRepository.swift
samwyndham: Updated in https://github.com/wireapp/wire-ios/pull/3171/commits/ca32297b00152114f1e40a55de8694feb87b68f0

dmitrysimkin (approved):

WireCells/Sources/WireCellsBindings/WireCellsAssembly.swift
netbe: great, another place also good to add document would be in docc, each package (should) have a md file to give an overview

netbe (approved):
LGTM;)

PR #3172, opened 2025-06-12
chore: Clean arch PoC

dmitrysimkin:
### Issue

_Please describe the issue._

### Testing

_Describe how to test._

---

### Checklist

---


WireAuthentication/Sources/WireAuthenticationData/AuthenticationAPIRepositoryAdapter.swift
```diff
@@ -0,0 +1,217 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireAPI
+import Foundation
+import WireAPI
+import WireAuthenticationDomain
+
+public final class AuthenticationAPIRepositoryAdapter: AuthenticationAPIRepository {
```
johnxnguyen, line 24: thought: I'm concerned about the increased boilerplate code that adapters will bring. In this particular case, I don't see much benefit to wrapping each method and mapping all models and errors to mirror types. 

I'm wondering if we can simply make the implementation type conform to the protocol without any changes... but of course to do this then any custom models would need to be known to both layers. In the case of an AccessToken, I think is acceptable, just as it would be for `QualifiedID`. Such foundational models could be defined in a `FoundationModel` package that can be imported freely.

```swift
// In FoundationModels
public struct AccessToken { ... }

// In domain layer
import FoundationModels
protocol AuthenticationRepository {

  func login(
    email: String, 
    password: String, 
    verificationCode: String?,
    label: String?
  ) async throws -> ([HTTPCookie), AccessToken)

}

// In data layer
extension WireAPI.AuthenticationAPI: WireAuthenticationDomain.AuthenticationRepository {}
```

I'm trying to spark a discussion here about where we draw the line architecture-wise for various rules. The way I see it, the benefits of theoretical architecture patterns change when you consider them in practical contexts, and depending on those contexts the costs and benefits may shift to more positive or more negative tendencies. It is of course a completely subjective topic.

johnxnguyen: And please do explain why this may be a bad approach... I want us to understand what are the pro's and con's to help us make an informed decision.

WireAuthentication/Sources/WireAuthenticationDomain/NetworkStack.swift
```diff
@@ -0,0 +1,88 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireFoundation
+import WireLogging
+
+package final class NetworkStack {
```
netbe, line 23: question: is it a duplication of NetworkClient in WireAPI?

PR #3193, opened 2025-06-16
fix: infinite loop - WPB-18213

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18213" title="WPB-18213" target="_blank">BugWPB-18213</a>  [iOS] Request loop on v8/conversations/list
  </td></table>
  <br />
 

### Issue

**Context**: An infinite loop on v8/conversations/list occurs on latest beta

**Causes**: During incremental sync and we're live syncing, the MLS repair agent is trying to repair the broken groups with an incremental sync which suspend and trigger again an incremental sync and in the case we're still not able to recover from that broken MLS group, the MLS repair agent will try to recover over and over again the same MLS group.. creating a loop

Another issue was that when going to the background and suspending the sync, we would also try to repair these groups

**Solutions**: Inject the MLSGroupRepairAgent in the IncrementalSync and repairConversations directly before live syncing + add an extra parameter to **not** perform the incremental sync in the MLSService.fetchAndRepairGroup(with:) method .

### Checklist


wire-ios-sync-engine/Source/Synchronization/MLSGroupRepairAgent.swift
```diff
@@ -95,4 +105,25 @@ final class MLSGroupRepairAgent: MLSGroupRepairAgentProtocol {
         }
     }
 
+    private func clearBrokenMLSGroups() {
+        let brokenGroupIDs = journal[.brokenMLSGroupIDs]
+
+        // Attempted to repair these MLS groups multiple times to no avail, clean them up to avoid an infinite loop
```
netbe: todo: fix comment

wire-ios-sync-engine/Source/Synchronization/MLSGroupRepairAgent.swift
```diff
@@ -37,8 +37,11 @@ final class MLSGroupRepairAgent: MLSGroupRepairAgentProtocol {
         journal[.isSyncV2Enabled]
     }
 
-    private let syncStatePublisher: AnyPublisher<SyncState, Never>
+    typealias AttemptCount = Int
 
+    private let syncStatePublisher: AnyPublisher<SyncState, Never>
+    private let maxAttemptCount = 4
```
netbe: **question**: so if a group goes to the 4th attempt we discard it right, could it be repaired later?

netbe (commented):
add some comments before approving

wire-ios-sync-engine/Source/Synchronization/MLSGroupRepairAgent.swift
jullianm: during a next incremental sync yes

wire-ios-sync-engine/Source/Synchronization/MLSGroupRepairAgent.swift
jullianm: what's to fix ?

wire-ios-sync-engine/Source/Synchronization/MLSGroupRepairAgent.swift
jullianm: during a next incremental sync, the broken group will show up again and we'll try again to recover with 4 attempts max.

wire-ios-sync-engine/Source/Synchronization/MLSGroupRepairAgent.swift
netbe: > Attempted to repair these MLS groups multiple times to no avail
is it cut of?
I don't really get the meaning

wire-ios-sync-engine/Source/Synchronization/MLSGroupRepairAgent.swift
netbe: I think we can lower the attemps.

netbe (approved):
LGTM;)

KaterinaWire (approved):

PR #3205, opened 2025-06-17
fix: connections requests not working - WPB-18159

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18159" title="WPB-18159" target="_blank">BugWPB-18159</a>  [iOS] NSE: connection requests not processed properly
  </td></table>
  <br />
 

### Issue

**Context**:  When receiving a push notification for a connection request, it's not processed correctly when going to to the foreground (request doesn't show up). Another issue is when ignoring a connection request from web client, app shows a notification saying connection was accepted.

**Causes**: Legacy logic for connections request processing was not correctly ported to new code (see `ConnectionRequestStrategy`). Also when receiving a push notification we were not properly discarding the connection statuses we don't want to display notifications for.

**Solutions**: Port legacy code to properly process the connection requests. Add validation criteria in NSE to only handle `.pending` or `.accepted` connection statuses.

### Testing

Have another user send a connection request
in iOS app, wait for the push notification, tap on it, when going to the foreground, it should correctly show up in the conversation list.
for second bug, same process but from a web client (same user as the iOS app) ignore the connection request
iOS app should not receive any notifications 

### Checklist

---


WireDomain/Sources/WireDomain/Event Processing/UserEventProcessor/UserConnectionEventProcessor.swift
```diff
@@ -18,19 +18,60 @@
...
 
+        guard let userID = event.connection.receiverQualifiedID?.toDomainModel() else { return }
+
         if connection.status == .accepted {
-            try await oneOnOneResolver.resolveAllOneOnOneConversations()
+            do {
+                // The client who accepts the connection resolves the conversation immediately.
+                // Other clients (from self and other user) resolve after a delay to avoid a race condition,
+                // but also to re-attempt resolution in case of failure.
+                try await Task.sleep(for: .seconds(oneOnOneResolutionDelay))
```
netbe: **thought:** hmm, so I wonder if this could be avoided, I am concerned of this adding automatically 3 seconds to the sync

WireDomain/Tests/WireDomainTests/Notifications/User/UserConnectionEventNotificationBuilderTests.swift
```diff
@@ -62,7 +62,8 @@ final class UserConnectionEventNotificationBuilderTests: XCTestCase {
             Scaffolding.userPendingConnectionEvent,
             Scaffolding.userAcceptedConnectionEvent,
             Scaffolding.userPendingConnectionEventNoUsername,
-            Scaffolding.userAcceptedConnectionEventNoUsername
+            Scaffolding.userAcceptedConnectionEventNoUsername,
+            Scaffolding.userIgnoredConnectionEvent // should be discarded
```
netbe, line 66: question: do you mean it's unused? If so let's remove it

WireDomain/Sources/WireDomain/Event Processing/UserEventProcessor/UserConnectionEventProcessor.swift
netbe: **issue:** looking at the legacy code, this is in a Task and the `func processEvents(_ events: [ZMUpdateEvent], liveEvents: Bool, prefetchResult: ZMFetchRequestBatchResult?) {` method where it's called is synchronous, so we don't block the rest of sync.

netbe (commented):
There might be an issue with delays the sync, see my comment

WireDomain/Sources/WireDomain/Event Processing/UserEventProcessor/UserConnectionEventProcessor.swift
johnxnguyen: I think it would be fine to trigger delayed resolving in a task so we don't block the sync, I also wonder if there's a better way to schedule resolving (probably out of scope.)

WireDomain/Tests/WireDomainTests/Notifications/User/UserConnectionEventNotificationBuilderTests.swift
johnxnguyen: I think he means that such an event should not generate the notification

johnxnguyen (commented):
looks good, but will wait for conversations to be resolved first.

WireDomain/Tests/WireDomainTests/Notifications/User/UserConnectionEventNotificationBuilderTests.swift
jullianm: yes

WireDomain/Sources/WireDomain/Event Processing/UserEventProcessor/UserConnectionEventProcessor.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/3205/commits/ccaa78d1b94727b87ffbe105422604b959e8dee5

netbe (approved):
LGTM;)

WireDomain/Sources/WireDomain/Event Processing/UserEventProcessor/UserConnectionEventProcessor.swift
```diff
@@ -18,19 +18,62 @@
...
     func processEvent(_ event: UserConnectionEvent) async throws {
         let connection = event.connection
 
         try await connectionsRepository.updateConnection(connection)
 
-        if connection.status == .accepted {
-            try await oneOnOneResolver.resolveAllOneOnOneConversations()
+        guard let userID = event.connection.receiverQualifiedID?.toDomainModel() else { return }
+
+        Task {
```
johnxnguyen: I was thinking that we would only wrap the handling of `connection.status == .accepted` in the task, but for the rest not. This would prevent blocking the sync for 3 seconds only. I also think we should add a comment why we have a separate task.

johnxnguyen (changes requested):

WireDomain/Sources/WireDomain/Event Processing/UserEventProcessor/UserConnectionEventProcessor.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/3205/commits/7938d3d0b649133537a646f25262ad6b9a1d5fb8

johnxnguyen (approved):

PR #3218, opened 2025-06-18
feat: NSE changes for consumable notifications - WPB-17229

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17229" title="WPB-17229" target="_blank">TaskWPB-17229</a>  [iOS] Integrate new async stream in NSE
  </td></table>
  <br />
 

### Issue

When consumable-notifications sync system is enabled, the NSE must open the websocket, decrypt and store the events.

### Testing

1. enable consumable-notifications developer-flag
2. login
3. put app in background
4. receive notifications


WireDomain/Sources/WireDomain/Notifications/SyncEventsUseCase.swift
```diff
@@ -0,0 +1,68 @@
...
+        case timedOut
+        case pendingEventsSyncFailed(Error)
+    }
+
+    let pendingEventsSync: PullPendingUpdateEventsSyncV2Protocol
+    let timeout: Duration
+
+    init(
+        pendingEventsSync: PullPendingUpdateEventsSyncV2Protocol,
+        timeout: Duration = .seconds(25)
```
netbe, line 38: we sync events (decrypt and store) for 25 seconds max, rest of 5 seconds of NSE lifetime, is for displaying the notifications. 

Note that if the sync reach the uptodate point we stop and show the notification

WireDomain/Sources/WireDomain/Notifications/Components/SyncEventsStep.swift
```diff
@@ -0,0 +1,445 @@
...
+    }
+
+    // MARK: - Children
+
+    var generateNotificationStep: GenerateNotificationStep {
+        GenerateNotificationStep(parent: self)
+    }
+}
+
+extension SyncEventsStep {
```
netbe, line 99: here's the same kind of dependency than VerifyUserStep, it should get replace/simplify by NetworkStack

WireDomain/Sources/WireDomain/Synchronization/PullPendingUpdateEventsV2Sync.swift
```diff
@@ -0,0 +1,201 @@
...
+
+                logger.debug(
+                    "received live element: \(element)",
+                    attributes: .syncAttributes(initialSync: false)
+                )
+                switch element {
+                case .upToDate:
+                    logger.debug("upToDate event", attributes: logAttributes)
+                    continuation.finish()
+                    break streamLoop
```
netbe, line 85: here we're upToDate so we close the stream we produce so notifications are generated + break from the forloop to close the pushchannel so NSE can finish and terminate after showing the notifications

WireDomain/Sources/WireDomain/Notifications/Components/VerifyUserStep.swift
```diff
@@ -80,10 +80,18 @@ final class VerifyUserStep: Component<VerifyUserDependency>, VerifyUserStepProto
...
+                selfClientID: selfClientID
+            ).pullEvents()
+
+        } else {
+            try await pullEventsStep(
+                selfUserID: selfUser.id,
+                selfClientID: selfClientID
+            ).pullEvents()
+        }
     }
```
jullianm, line 93: thought: I wonder if both objects should be renamed the same way and adding the suffix V2 or V3 like:
`SyncEventsStepV3`  (when consumable notification enabled)
`SyncEventsStepV2` (instead of `PullEventsStep` when disabled)

so we know right away they have the same purpose but we use one or the other depending on the "version"

WireDomain/Sources/WireDomain/Notifications/Components/SyncEventsStep.swift
```diff
@@ -0,0 +1,445 @@
...
+        case missingProxyCredentials
+        case apiVersionNotFound
+    }
+
+    private var selfUserID: UUID
+    private var selfClientID: String
+
+    init(
+        parent: any Scope,
+        selfUserID: UUID,
```
johnxnguyen: question: why inject it if the dependency provides it?

WireDomain/Sources/WireDomain/Notifications/Components/SyncEventsStep.swift
```diff
@@ -0,0 +1,445 @@
...
+        )
+
+        let useCase = SyncEventsUseCase(pendingEventsSync: pendingEventsSync)
+
+        do {
+            try await useCase.invoke()
+        } catch {
+            WireLogger.sync.warn(
+                "syncing events via websocket: \(error.localizedDescription)",
+                attributes: .syncAttributes(initialSync: false)
```
johnxnguyen, line 83: question: does it make sense to continue if we couldn't pull events?

WireDomain/Sources/WireDomain/Synchronization/PullPendingUpdateEventsV2Sync.swift
```diff
@@ -0,0 +1,201 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireCoreCrypto
+import WireDataModel
+import WireLogging
+import WireNetwork
+
+public struct PullPendingUpdateEventsSyncV2: PullPendingUpdateEventsSyncV2Protocol {
```
johnxnguyen, line 25: question: seems like a bit of duplication here and the incremental sync v2... do you think we can find a way to share common code? or if it makes sense?

wire-ios-data-model/Tests/Source/ManagedObjectContext/DatabaseMigrationTests.swift
```diff
@@ -138,6 +138,7 @@ final class DatabaseMigrationTests: DatabaseBaseTest {
         }
 
         try allVersions.forEach { version in
+            print("😀 \(version)")
```
johnxnguyen: question: to delete?

johnxnguyen (commented):
looks good, left some questions

WireDomain/Sources/WireDomain/Synchronization/PullPendingUpdateEventsV2Sync.swift
netbe: For me, it's similar duplication as IncrementalSyncV1  and IncrementalSyncV2. now it does not make sense as we don't have processing events in background. THen it could be that IncrementalSyncV2 uses PullPendingUpdateEventsSyncV2

WireDomain/Sources/WireDomain/Notifications/Components/SyncEventsStep.swift
netbe: yes, so here's there are 2 reasons, either timeout because it's been 25 seconds decrypting and storing events OR an issue during the sync.

In both cases, we end up with a stream of notifications that has not been shown, so we need to continue to show them

wire-ios-data-model/Tests/Source/ManagedObjectContext/DatabaseMigrationTests.swift
netbe: db2562eb5de21eae7f92be7572dca4a75e614f03

WireDomain/Sources/WireDomain/Notifications/Components/SyncEventsStep.swift
netbe: copy paste from PullEventsStep, and fixed here: ff999e445fa4bdb7ca19af1460f608a5f4453f5e

jullianm (approved):
Looks good, nice work

WireDomain/Sources/WireDomain/Notifications/Components/SyncEventsStep.swift
johnxnguyen: maybe worth a comment since it's not obvious

johnxnguyen (approved):

WireDomain/Sources/WireDomain/Notifications/Components/VerifyUserStep.swift
netbe: it's a good point but not sure about the naming right now, I'll come back to that later

PR #3222, opened 2025-06-19
fix: wrong conditional checking when initiating call - WPB-18292

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18292" title="WPB-18292" target="_blank">BugWPB-18292</a>  [iOS] Calling issue when on a group conv, someone joining, it initiates ringing by that person on the same group
  </td></table>
  <br />
 

### Issue

**Context**:
- A group conv call is ongoing
- Someone join the call
- NSE displays calling notification from that participant on the same group

**Causes**: `initiatesRinging` condition checking is wrong, we don't discard the notification is the call was already reporter. which means we receive the event for the user that joined the group conv, it checks the calling content (which it has) then checks whether it should initiate ringing and it does.

**Solution**: Fix `initiatesRinging` condition checking logic -> discard notification if the call was already reported

_legacy code_

```
        if callContent.initiatesRinging, !wasCallHandleReported {
            WireLogger.calling.info("should initiate ringing", attributes: .legacyNSE)
            return CallEventPayload(
                accountID: accountIdentifier.uuidString,
                conversationID: conversationID.uuidString,
                shouldRing: true,
                callerName: conversation.localizedCallerName(with: caller),
                hasVideo: callContent.isVideo
            )
        }
```

### Testing

- have group conv call 
- someone join the group conv
- it should not display any notification

### Checklist

---

### DIFF
--- a/WireDomain/Sources/WireDomain/Notifications/Builders/Conversation/ConversationCallingEventNotificationBuilder.swift
+++ b/WireDomain/Sources/WireDomain/Notifications/Builders/Conversation/ConversationCallingEventNotificationBuilder.swift
@@ -394,7 +394,7 @@ extension ConversationCallingEventNotificationBuilder {
             let knownCallHandles = userDefaults.object(forKey: Constants.knownCalls) as? [String] ?? []
             let wasCallHandleReported = knownCallHandles.contains(handle)
 
-            let initiatesRinging = callContent.isIncomingCall || wasCallHandleReported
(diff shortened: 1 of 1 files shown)


caldrian (approved):

netbe (approved):
great, thks for showing the legacy code too, could you add a regression test so we're covered @jullianm ?

jullianm:
@netbe yeah, I'll fix the UTs thx

PR #3228, opened 2025-06-20
chore: update CoreCrypto to 7.0.2 - WPB-18348

typfel:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18348" title="WPB-18348" target="_blank">TaskWPB-18348</a>  [iOS] Core - Update CC to 7.0.2
  </td></table>
  <br />
 

### Issue

Since core crypto https://github.com/wireapp/core-crypto/releases/tag/v7.0.1 we are allowed register the epoch observer before calling `mlsInit`, we can therefore stop trying to register the epoch observer in code paths which initialises MLS.

### Checklist

---


wire-ios-data-model/Source/Core Crypto/CoreCryptoProvider.swift
```diff
@@ -112,7 +112,6 @@ public actor CoreCryptoProvider: CoreCryptoProviderProtocol {
             )
             try await self.generateClientPublicKeys(with: context, credentialType: .basic)
         }
-        try await registerEpochObserverIfNecessary(with: coreCrypto)
```
netbe, line 115: removed because we don't need to do this before mlsInit

netbe (changes requested):
need to update the CoreCryptoContextProtocolExt

wire-ios-data-model/Source/Core Crypto/CoreCryptoProvider.swift
dmitrysimkin: It's now called only in MLSService.init
```
    public func startObservingEpochs() {
        Task {
            await coreCryptoProvider.registerEpochObserver(self)
        }
    }
```

dmitrysimkin (approved):

netbe (approved):
LGTM;)

PR #3232, opened 2025-06-20
fix: verification status for self clients being lost - WPB-17594

caldrian:
### Issue

When receiving updates for other clients of the self user we accidentally also set existing clients as untrusted.
This PR adds a missing condition.

### Testing

- Have more than one client for a user.
- After sync or restarting the app the previously verified devices must still be marked as verified.

---

### Checklist

---


netbe (commented):
Should we add a test to verify this?

+ I wonder why this targets 3.120? fixVersion says 4.1.0

netbe (approved):
Looks good to me

jullianm (approved):
Looks good

PR #3242, opened 2025-06-23
fix: crash when leaving group conv - WPB-18343

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18343" title="WPB-18343" target="_blank">BugWPB-18343</a>  [iOS] App crashing frequently while clear and leave group conversation
  </td></table>
  <br />
 

### Issue

**Context**: Sometimes a crash occurs when leaving a group conv (not reproducible very often)

**Causes**: We force crash the app when system message `senderUser` or `conversationLike` values are null.

**Solution (quick fix)**:  Add an assertionFailure (for debug) and return an empty array. 

Would probably require further investigation to see why some values are missing but since crash happens when leaving group conv it can be safe to assume that there's no risk in returning an empty array here to avoid the crash.

### Testing

_Describe how to test._

---

### Checklist

---


caldrian (approved):

netbe (approved):
good as current fix

PR #3139, opened 2025-06-05
feat: migrate sync system - WPB-17223

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17223" title="WPB-17223" target="_blank">TaskWPB-17223</a>  [iOS] Migrate from legacy incremental sync to new system
  </td></table>
  <br />
 

### Issue

This PR adds migration from sync system v2 to sync system v3 (consumable notifications). Migration consists of:
* register consumable-notifications capability
* run incremental sync V1 (slow sync) without opening websocket
* start new sync v3 (incremental sync - live sync)

Also it removes any naming inconsistency. AsyncStream is now ConsumableNotifications

### Testing

Upgrade scenario
1. login
2. turn on consumableNotifications Developer flag
3. restart the app
-> migration should run

New client scenario
1. turn on consumableNotifications Developer flag
2. login
-> migration should not run
---

### Checklist

- [ ] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [ ] Description is filled and free of optional paragraphs.
- [ ] Adds/updates automated tests.

---

### UI accessibility checklist

_If your PR includes UI changes, please utilize this checklist:_
- [ ] Make sure you use the API for UI elements that support large fonts.
- [ ] All colors are taken from WireDesign.ColorTheme or constructed using WireDesign.BaseColorPalette.
- [ ] New UI elements have Accessibility strings for VoiceOver.

PR #3251, opened 2025-06-24
feat: remove codable public implementation in WireNetwork - WPB-18279

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18279" title="WPB-18279" target="_blank">TaskWPB-18279</a>  [iOS] Remove Codable conformance in public models in WireAPI
  </td></table>
  <br />
 

### Issue

Following discussion https://github.com/wireapp/wire-ios/pull/3139#discussion_r2154340136, in order to make sure we don’t break something with WireAPI models, we need to remove all encodable / decodable conformance of all public models in WireAPI. 

* remove all Codable or Decodable, Encodable conformance on public models
* create internal version ...V0 with Codable or Decodable, Encodable 
* add new protocol ToNetworkConvertible to convert public models to encodable models
* replaced QualifiedID implemenetation from WireNetwork and create an alias on WireFoundation

### Testing

* run tests and run app

---

### Checklist

---


WireDomain/WireDomain Project.xcodeproj/project.pbxproj
```diff
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		017F67942C207AA000B6E02D /* WireDomainSupport.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 017F67822C207A3200B6E02D /* WireDomainSupport.framework */; };
+		01A9C1BD2E0A106D000617B4 /* WireFoundation in Frameworks */ = {isa = PBXBuildFile; productRef = 01A9C1BC2E0A106D000617B4 /* WireFoundation */; };
```
netbe, line 11: added import WireFoundation in WireDomainSupport to resolve:

```
Undefined symbol: type metadata accessor for WireFoundation.QualifiedID
```

WireNetwork/Sources/WireNetwork/Models/Conversation/ChannelPermission.swift
```diff
@@ -18,7 +18,32 @@
...
+extension ChannelPermission: ToNetworkConvertible {
+    func toNetworkModel() -> ChannelPermissionV8 {
+        switch self {
+        case .everyone:
+            .everyone
+        case .admins:
+            .admins
+        }
+    }
+}
```
dmitrysimkin, line 49: What is the difference between API and Network model? They all live WireNetwork module which previously was WireAPI module :) it's confusing

netbe: I would guess Network models is whatever is required to communicate to the server (Token, cookie,..)
API models are whatever is returned by the backend endpoints.

but I am not sure we need a separation of it. Also not in scope of this PR (we can take this point to the Architecture forum).

johnxnguyen: In WireNetwork you can find three main types of models:

1. Codable models for specific REST API version.
2. Models that are part of the abstracted REST API protocols. These are not versioned, and map to and from the codable versioned models.
3. Other models that aren't part of the REST API but are relevant to the network code, such as AccessToken, APIVersion, etc.

Perhaps we should make this distinction more clear in the way we organize the code. I was actually thinking that maybe having the network code in on target and the REST api code (protocols, versioned implementations, and models) in a separate REST api target.

dmitrysimkin: I think naming that would clear the difference is more benefecial, and then maybe split in targets if needed, to me that optional

dmitrysimkin (approved):
LGTM 👍

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationModelMappings.swift
```diff
@@ -39,6 +39,26 @@ extension WireNetwork.MLSCipherSuite {
             .MLS_256_DHKEMP384_AES256GCM_SHA384_P384
         }
     }
+
+    func toConfigModel() -> Feature.MLS.Config.MLSCipherSuite {
```
johnxnguyen, line 43: question: is this from REST to CoreData?

johnxnguyen (approved):
Nice work! If backend followed and declared a [json schema](https://json-schema.org), I imagine a lot of this code could be generated.

WireDomain/Sources/WireDomain/Repositories/Conversations/ConversationModelMappings.swift
netbe: yes

WireDomain/WireDomain Project.xcodeproj/project.pbxproj
netbe: @caldrian is that ok?

netbe:
some crashes happen, since for instance the public enum still expose the rawType and this is used on higher levels as strings

netbe:
@johnxnguyen there was some new changes since your last review. Basically some public enum had their rawType exposed and we used it in higher levels. I now hide those too, and added methods to convert on domain level. It's a bit more verbose but I guess it's the cost to pay here

johnxnguyen (approved):

PR #3263, opened 2025-06-24
feat: batch decryption for consumable notifications - WPB-17703

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17703" title="WPB-17703" target="_blank">TaskWPB-17703</a>  [iOS] Add batch decryption for async notifications
  </td></table>
  <br />
 

### Issue

This PR adds batch decryption to IncrementalSyncV2.

* Changed the way PushChannelV2 streams element: .event becomes .events([UpdateEventEnvelope]). It nows collects up to 100 events within 0,5 seconds.
* Update IncrementalSyncV2 to decrypt in batch events

### Testing

* see UTs
---

### Checklist

---


WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -247,28 +269,7 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
         }
     }
 
-    private func storeEnvelope(_ envelope: UpdateEventEnvelope) async throws -> Int64 {
```
netbe, line 250: we don't need to storeEnvelopes nor delete them, as if we fail to acknowledge an event backend will send it again via websocket

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -223,15 +209,51 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
...
+                await processEnvelope(envelope)
+
+                lastEnvelope = envelope
+            }
+        }
+
+        await updateEventsStore.calculateLastUnreadMessages()
+        await save()
+        if let lastEnvelope {
+            await acknowledgeUntilEnvelope(lastEnvelope, through: pushChannel)
```
netbe: we save after batch of events is decrypted and processed then we acknowledge all events

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
typfel: You don't store decrypted event anymore?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -171,41 +177,21 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
...
-                        await deleteEnvelope(envelope, at: index)
-                        await updateEventsStore.calculateLastUnreadMessages()
-
-                        await save()
+                    do {
+                        try await processEnvelopes(
+                            envelopes,
+                            pushChannel: pushChannel,
+                            processedEnvelopeIDs: processedEnvelopeIDs
+                        )
```
johnxnguyen: suggestion: pull acknowledgment out of `processEnvelopes` and call it directly after.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -223,15 +209,51 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
...
+                        "live event already processed, skipping...",
+                        attributes: .syncAttributes(initialSync: false) + [.eventEnvelopeID: envelope.id]
+                    )
+                    // TODO: [WPB-17947] handle duplicate events, reacknowledge and move on
+                    // will need to store the deliveryTag...
+                    continue
+                }
+
+                var envelope = envelope
+                envelope.events = try await decryptEnvelope(envelope, in: coreCryptoContext)
```
johnxnguyen: question: what happens if this throws an error, will the enclosing call to core crypto rethrow it?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -223,15 +209,51 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
 
     // MARK: - Event Steps
 
-    private func decryptEnvelope(_ envelope: UpdateEventEnvelope) async throws -> [UpdateEvent] {
+    private func processEnvelopes(
+        _ envelopes: [UpdateEventEnvelope],
+        pushChannel: PushChannelV2Protocol,
+        processedEnvelopeIDs: Set<UUID>
+    ) async throws {
```
johnxnguyen, line 215: question: if this throws then it we simply catch it and ignore it one level higher. But what does that mean? Will the push channel continue with the next batch of events? Does it make sense to continue processing the next batch if the previous batch failed? Perhaps it's possible that the first batch fails but the second doesn't, and as a result we acknowledge everything up until the last event of the second batch. This would mean we don't retry the first batch.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -278,7 +279,7 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
                     "ack event envelope",
                     attributes: [.eventEnvelopeID: envelope.id] + .syncAttributes(initialSync: false)
                 )
-                try await pushChannel.acknowledgeEvent(deliveryTag: deliveryTag, multiple: false)
+                try await pushChannel.acknowledgeEvent(deliveryTag: deliveryTag, multiple: true)
             }
         } catch {
             logger.error(
```
johnxnguyen, line 310: question: if we fail to acknowledge... is it fine to simply continue on anyway?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
johnxnguyen: I suppose we don't need to if we are able to receive the events, decrypt them again, and try to process them once more.

johnxnguyen (commented):
Leaving some initial questions

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: @johnxnguyen that was my point too, but @typfel mentions that if we crash on processing the events we won't be able to decrypt again the event as CC transaction would have changed the CC state.

I'll put it back for now to be safe

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: right with batch decryption we have to make sure we successfully decrypt first batch before acknowledging

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: if we fail, we'll get the events again when we reopen the websocket

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: 8265b4f5d1e4b14653da6ec3db6eafb25c9e4771

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: I assume yes, it will throw and the full transaction wi

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: here the only thing that throws is decryption, so maybe if we have duplicate events it's ok to ignore and go to next batch

netbe: so finally i adding a ack per event in case of error. so if one of the event fails to process, we still acknowledge the other ones. We should get the fail one again on next websocket connection.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -223,15 +223,74 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
...
+                    attributes: .syncAttributes(initialSync: false) + [.eventEnvelopeID: envelope.id]
+                )
+            }
+        }
+
+        // only delete successful processed envelopes
+        await deleteEnvelopes(at: envelopeIdsToDelete)
+
+        await updateEventsStore.calculateLastUnreadMessages()
+        await save()
```
netbe: for now we don't throw for this operations but we could revisit in WPB-10458

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -18,19 +18,27 @@
...
+import WireDataModel
 import WireLogging
 import WireNetwork
 
 /// IncrementalSync using new backend API consumable notifications sync system
 public struct IncrementalSyncV2: LiveSyncProtocol {
 
+    enum Failure: Error {
+        /// Contains all envelopes that were successfully processed
+        case uncompleteBatchProcessed(processedEnvelopes: [UpdateEventEnvelope])
```
johnxnguyen: typo:

```suggestion
        case incompleteBatchProcessed(processedEnvelopes: [UpdateEventEnvelope])
```

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -171,41 +183,29 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
...
-                            )
-                            // TODO: [WPB-17947] handle duplicate events, reacknowledge and move on
-                            // will need to store the deliveryTag...
-                            continue
+                        if let lastEnvelope = envelopes.last {
+                            await acknowledgeUntilEnvelope(lastEnvelope, through: pushChannel)
+                        }
+                    } catch let Failure.uncompleteBatchProcessed(processedEnvelopes) {
+                        for envelope in processedEnvelopes {
+                            await acknowledgeEnvelope(envelope, through: pushChannel)
```
johnxnguyen: question: so you can acknowledge events one by one and the non-acknowledged ones will be delivered again? I understood the acknowledgment to be a kind of marker that you can only push forward, so if you acknowledge x but not x - 1, the next event delivered would be x + 1. Did I understand wrong?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -223,15 +223,74 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
...
+                        await acknowledgeEnvelope(envelope, through: pushChannel)
+                    }
+                    continue
+                }
+
+                var envelope = envelope
+                envelope.events = try await decryptEnvelope(envelope, in: coreCryptoContext)
+
+                let index = try await storeEnvelope(envelope)
+                storedEnvelopes.append((envelope, index))
```
johnxnguyen, line 227: question: why don't we acknowledge the last event at the end after storing all events? We've stored them so there's no need to refetch them again right?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -223,15 +223,74 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
...
+            }
+        }
+
+        // store
```
johnxnguyen:
```suggestion
        // process
```

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -223,15 +223,74 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
...
+        var envelopeIdsToDelete = [Int64]()
+        for (envelope, index) in storedEnvelopes {
+            do {
+                try await processEnvelope(envelope)
+                envelopeIdsToDelete.append(index)
+            } catch {
+                logger.error(
+                    "Failed to process envelope: \(error)",
+                    attributes: .syncAttributes(initialSync: false) + [.eventEnvelopeID: envelope.id]
```
johnxnguyen: suggestion: I think we should have an assertion failure here... ideally we can process every event as skipping some that failed leads to undefined behavior. Therefore we should know about it asap.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
johnxnguyen: so then next time we would get decryption errors for duplicate messages? Perhaps we should catch duplicate message errors and acknowledge those duplicate events.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -278,45 +337,53 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
...
-                attributes: [.eventEnvelopeID: envelope.id] + .syncAttributes(initialSync: false)
+                "processing live event: \(event.name)",
+                attributes: [.eventEnvelopeID: envelope.id]
             )
-            try await updateEventsStore.deleteEventEnvelope(atIndex: index)
+            try await processor.processEvent(event)
+        }
+    }
+
+    private func deleteEnvelopes(at indexes: [Int64]) async {
```
johnxnguyen: typo: 
```suggestion
    private func deleteEnvelopes(at indices: [Int64]) async {
```

WireNetwork/Sources/WireNetwork/Network/PushChannel/PushChannelV2.swift
```diff
@@ -50,28 +58,60 @@ public final class PushChannelV2: PushChannelV2Protocol {
...
+                    let result = try receiveMessage(message)
+                    switch result {
+                    case let .event(event):
+                        batch.append(event)
+                        if batch.count == maxBatchEventsCount {
+
+                            if !batch.isEmpty {
+                                continuation.yield(.events(batch))
+                            }
+                            batch = []
```
johnxnguyen: question: is a `maxBatchEventsCount` of 0 valid?

wire-ios-data-model/Source/Model/ZMEventModel/StoredUpdateEventEnvelope/StoredUpdateEventEnvelope+Helpers.swift
```diff
@@ -77,4 +77,13 @@ public extension StoredUpdateEventEnvelope {
         return request
     }
 
+    static func fetchRequest(sortIndexes: [Int64]) -> NSFetchRequest<StoredUpdateEventEnvelope> {
```
johnxnguyen: typo:

```suggestion
    static func fetchRequest(sortIndices: [Int64]) -> NSFetchRequest<StoredUpdateEventEnvelope> {
```

johnxnguyen (commented):
Left a few more questions

WireNetwork/Sources/WireNetwork/Network/PushChannel/PushChannelV2.swift
netbe: not really. I'll adapt this to default to 1

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: with multiple:true yes this is true.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: yes that's what we do, the acknowledge true is now outside of this method

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: done

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: done

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: it's part of a second ticket yes WPB-17947

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: made a fix here: [224c0f57b9ac366904a51ebed622e053fc15399d](https://github.com/wireapp/wire-ios/pull/3263/commits/6f3319fc4cd9a839aa339392df1766e665874dd6) so we don't acknowledge failed events by mistake on next batch

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
typfel: I don't think you want to give up the whole batch if you get a decryption error? Then you'll close the connection and then you'll get the decryption error again and repeat...

I think the right the thing to do is to report the decryption error and move on.

WireNetwork/Sources/WireNetwork/Network/PushChannel/PushChannelV2.swift
johnxnguyen: Now that the max batch count is non zero, then `if !batch.isEmpty` will always be true so you can remove the condition and just send the batch.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
johnxnguyen: I'm still a bit confused by the error handling strategy, such as why do we only acknowledge successfully processed events? Aren't the failed events still in the event db? In which case, we can acknowledge that they are decrypted and stored.

johnxnguyen (commented):
Some more questions/comments

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
typfel: Ah wait I guess they are already caught and handled in `UpdateEventDecryptor.decryptEvents()`

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: we now ack after stored events

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: I changed the signature of decryptEnvelope as it's not throwing actually

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: As discussed this has been removed, and we always ack in batch after events are stored

WireNetwork/Sources/WireNetwork/Network/PushChannel/PushChannelV2.swift
netbe: discussed and updated

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: discussed and updated

johnxnguyen (approved):

typfel (approved):

PR #3266, opened 2025-06-25
fix: hanging IncrementalSyncObserverTests tests - WPB-18424

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18424" title="WPB-18424" target="_blank">TaskWPB-18424</a>  [iOS] Hanging tests prevent PRs from being merged fast
  </td></table>
  <br />
 

### Issue

Hanging `IncrementalSyncObserverTests` test.

### Solution

There are 2 test that can easy reproduce this issue by running it 100 times:
`ifAppIsLiveThenDontWait` and `waitForDecryptionToFinish`. I will explain fixes for each of them separately.

1. `ifAppIsLiveThenDontWait`
Just(...) emits and completes immediately upon subscription — there's a very short window to process the value.
If `sut.waitUntilCanSendMessage()` does any async setup after the subscription, it might miss the .liveSyncing(.ongoing) emission entirely. So even though subscription technically happens, the handler inside the Combine pipeline hasn't attached yet, and the value is lost.
**The solution** is to guarantee the value is emitted after the subscription is active.

3. `waitForDecryptionToFinish`
Sometimes, timing is such (when running test many times in a row) that `.waitUntilCanSendMessage()` subscribes after the Task sends `.processPendingEvents`. Then, it only sees the current value (which is still `.pullPendingEvents`). And since `.pullPendingEvents` doesn't satisfy the continuation → await sut.waitUntilCanSendMessage() never resumes → 💥 test hangs.
This issue cannot be reproduced by running the test only once, but it is possible to run it multiple times.
**The solution** is to add a flag to signal when the observer is active.

**My thoughts:**
In general, I would use still **XCTest** for such async tests. They have more functionality to handle complex use cases and timeout if needed without causing side hangs.

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Tests/Source/Synchronization/IncrementalSyncObserverTests.swift
+++ b/wire-ios-sync-engine/Tests/Source/Synchronization/IncrementalSyncObserverTests.swift
@@ -36,7 +36,12 @@ class IncrementalSyncObserverTests {
     func ifAppIsLiveThenDontWait() async {
         // Given
         syncAgent.isSyncV2Enabled = true
-        syncAgent.syncStatePublisher = Just(SyncState.liveSyncing(.ongoing)).eraseToAnyPublisher()
+        let subject = PassthroughSubject<SyncState, Never>()
+        syncAgent.syncStatePublisher = subject.eraseToAnyPublisher()
+        Task {
+            try? await Task.sleep(nanoseconds: 1_000_000)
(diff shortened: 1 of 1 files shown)


netbe (approved):
LGTM;)

johnxnguyen (approved):

PR #3273, opened 2025-06-26
feat: update serverTime - WPB-18318

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18318" title="WPB-18318" target="_blank">TaskWPB-18318</a>  [iOS] Retrieve serverTimestamp
  </td></table>
  <br />
 

### Issue

For consumable notifications, we don't get the serverTime from /notifications... Now we must use the dedicated endpoint to update our value.

- Implement new endpoint for v8+
- call on IncrementalSyncV2 

### Testing

See UTs.

---

### Checklist

---


WireDomain/Sources/WireDomain/Synchronization/Protocols/PullServerTimeSyncProtocol.swift
```diff
@@ -0,0 +1,27 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+// sourcery: AutoMockable
+/// An object to keep the server time up to date.
+public protocol PullServerTimeSyncProtocol {
+
+    /// Fetch servertime from remote, then store it locally.
```
johnxnguyen:
```suggestion
    /// Fetch server time from remote, then store it locally.
```

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -60,6 +63,9 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
 
     public func perform() async throws -> IncrementalSync.Token {
         logger.debug("performing live sync", attributes: .syncAttributes(initialSync: false))
+
+        try await pullServerTimeSync.pull()
```
johnxnguyen, line 77: question: so we only pull it once? The push channel will live longer so won't the time get out of date?

WireNetwork/Sources/WireNetwork/APIs/UpdateEventsAPI/UpdateEventsAPIV8.swift
```diff
@@ -16,6 +16,26 @@
...
+
 final class UpdateEventsAPIV8: UpdateEventsAPIV7 {
     override var apiVersion: APIVersion { .v8 }
+
+    override func getServerTime() async throws -> Date {
+        let path = "\(pathPrefix)/time"
+
+        var requestBuilder = try URLRequestBuilder(path: path).withMethod(.get)
+
+        let request = requestBuilder.build()
```
johnxnguyen: suggestion: keep it simple:

```suggestion
        let request = try URLRequestBuilder(path: "\(pathPrefix)/time")
            .withMethod(.get)
            .build()
```

WireNetwork/Tests/WireNetworkTests/APIs/UpdateEventsAPI/UpdateEventsAPITests.swift
```diff
@@ -242,6 +248,41 @@ final class UpdateEventsAPITests: XCTestCase {
...
+        // when
+        // then
+
+        XCTAssertEqual(suts.count, unsupportedVersions.count)
+
+        for sut in suts {
+            await XCTAssertThrowsErrorAsync(UpdateEventsAPIError.unsupportedEndpointForAPIVersion) {
+                try await sut.getServerTime()
+            }
+        }
```
johnxnguyen, line 272: thought: I think the new swift testing framework here would make this nicer, we could just pass in unsupported api versions as arguments and avoid the for loop. (This is just a comment and not a request)

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: yes, to be honest I didn't think of it, well the sync closes when we go on background, so I don't see how we would go out of date so much.

I would keep it like this, unless you have another recommendation

johnxnguyen: What would happen though if the app is open for more than 30 seconds and receives a call event?  Wouldn't it be considered stale? Perhaps we can pull once at the start, then once we're up to date pull for every event.

netbe: pulling for each event sounds not efficient, if that's really needed then better for backend to put the time within the event, no?

netbe: actually that would have been the case also before since we received the servertime only when pulling pending events via /notifications not when opening the websocket

johnxnguyen: So we never used up to date time stamps with the legacy push channel?

johnxnguyen (approved):

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
jullianm: Maybe we could check with other platforms how they handled it since this is new feature we could all align (if we have a doubt)

netbe: @jullianm I checked with Android they're going to do as before so fetch before opening websocket

jullianm (approved):
Nice work

PR #3279, opened 2025-06-30
feat: Replace compositions with separate string entries - WPB-18448

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18448" title="WPB-18448" target="_blank">BugWPB-18448</a>  [iOS] [Replicated] When a group contains federated users and guests, display issues occur with spelling mistakes and double spaces
  </td></table>
  <br />
 

### Issue

There is a spelling mistakes in German (gäste instead of Gäste) and double spaces in the "Federated users and guests ..." bar.
The problem is that localized strings are being composed from multiple strings.

### Solution

The solution is to move all of these compositions and have a separate string entry for every possible case.

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+GuestBar.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+GuestBar.swift
@@ -28,40 +28,67 @@ extension ConversationViewController {
 
         let state = conversation.externalParticipantsState
 
-        if state.isEmpty {
+        guard
+            !state.isEmpty,
+            let labelKey = label(for: state)
+        else {
             return .hidden
-        } else {
-            return .visible(labelKey: label(for: state), identifier: identifier(for: state))
         }
+        return .visible(labelKey: labelKey, identifier: identifier(for: state))
+
     }
 
-    func label(for state: ZMConversation.ExternalParticipantsState) -> String {
-        var states: [String] = []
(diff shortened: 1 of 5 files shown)


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+GuestBar.swift
```diff
@@ -28,40 +28,67 @@ extension ConversationViewController {
...
+            return BannerStrings.externalsPresent
+
+        case [.visibleGuests]:
+            return BannerStrings.guestsPresent
+
+        case [.visibleServices]:
+            return BannerStrings.servicesActive
+
+        default:
+            return nil
```
samwyndham, line 91: suggestion (optional): Is this an expected state? If no, you could log an error / assert a failure

samwyndham (approved):
Nice work. Much better than composing the strings!

jullianm (approved):
Looks good , much better to have separate entries, we had same in NSE it can be very confusing. 👍

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+GuestBar.swift
KaterinaWire: it's not expected, but we won't show the top bar if there is no label.

PR #3300, opened 2025-07-02
fix: missing groups and channels - WPB-18477

jullianm:
[WPB-18477](https://wearezeta.atlassian.net/browse/WPB-18477)

### Issue

**Context**: Some groups conversation / channels are sometimes missing in the list - sometimes we still receive notifications for these missing conv, sometimes we don't.

**Causes**: After inspecting local database on a user that has a missing group, we noticed the group was created locally but metadata is missing

**Solution**: Solution is to set the `needsToBeUpdatedFromBackend` flag to true right after we create the conversation and sets it to false only after all the metadata has been updated.
If for any reasons, the object metadata was not fully updated to the store, flag will remain to true and the conversation that has this flag to true will be **automatically** resynced : see legacy code in in `ConversationRequestStrategy`

```
self.updateSync = KeyPathObjectSync(
            entityName: ZMConversation.entityName(),
            \.needsToBeUpdatedFromBackend
```

Also, we introduced an app version migration logic on 4.1.0 so that it triggers a resources sync and resolve the missing group conv / channels for existing users.

Screenshot 2025-07-02 at 14 10 03

### Testing

- on prod, iOS app in debug action menu, trigger "Create MLS group conv with missing metadata", created conv will not show up in the iOS app but it will on desktop app
- on desktop app, add someone and have them send messages
- on iOS app, conv is missing but we still get notifications (don't do anything)
- back in debug actions, trigger "perform app versions migration" 

-> group conv should now be visible

### Checklist

---

[WPB-18477]: https://wearezeta.atlassian.net/browse/WPB-18477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jullianm:
@johnxnguyen @netbe it's a first step to avoid having this issue in the future. it won't resolve the already missing groups or channels for existing users, for that we can ask customer support to tell them to trigger a resources sync or maybe add some logic inside the codebase to repair the missing groups (not sure if it's worth it)

WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift
```diff
@@ -1082,23 +1083,26 @@ public final class ConversationLocalStore: ConversationLocalStoreProtocol {
...
             }
 
-            /// If we discover this group is actually a fake one on one,
-            /// then we should link the one on one user.
+            // If we discover this group is actually a fake one on one,
+            // then we should link the one on one user.
             linkOneOnOneUserIfNeeded(for: localConversation)
+            
+            // All metadata has been updated, object does not need to be updated from backend
+            localConversation.needsToBeUpdatedFromBackend = false
```
netbe, line 1104: question: so we flip the flag off at the end of this method? shouldn't the legacy code do that instead?

netbe (commented):
question: which missing metadata is causing the list not to show the conversation?

WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift
jullianm: groupType, conversationType but overall all the metadata is missing

jullianm: if we don't set it to false, legacy code will try to refetch metadata for a conversation that is already up to date which I think it's not necessary no?

netbe:
> @johnxnguyen @netbe it's a first step to avoid having this issue in the future. it won't resolve the already missing groups or channels for existing users, for that we can ask customer support to tell them to trigger a resources sync or maybe add some logic inside the codebase to repair the missing groups (not sure if it's worth it)

@jullianm you can add an app migration to trigger a resources sync on 4.1

jullianm:
@netbe  I'll add the logic to trigger resources sync

WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift
```diff
@@ -1015,6 +1015,7 @@ public final class ConversationLocalStore: ConversationLocalStoreProtocol {
         )
 
         await context.perform { [self] in
+            localConversation.needsToBeUpdatedFromBackend = true
```
johnxnguyen: thought: I think this should be whenever we create a new conversation, so deep inside the `fetchOrCreateConversation` method.

WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift
johnxnguyen: legacy code should already flip the flag, but doesn't hurt to do it here as well. I'm wondering though, does the legacy code also deal with `groupType`?

netbe: @johnxnguyen it does we checked

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1550,3 +1567,15 @@ extension ZMUserSession {
 
     }
 }
+
+extension ZMUserSession {
+
+    private func makeAppVersionMigrations() -> [any AppVersionMigration] {
+        [
+            AppVersionMigration_4_1_0(
+                performResourceSync: triggerResourcesSync
```
netbe: question: does it capture self here? could there be a memory issue?

WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift
netbe: I guess that's a good idea, but going forward we should think how updating a conversation will look like out of legacy. this flag kind of introduce a side effect and is hard to control when the update is finished

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
jullianm: hmm, yes there could be, if I'm not mistaking closure are reference type so if the ZMUserSession wants to deallocate and the closure is being run. it will not be able to do so. not sure, better to make it weak. I'll do the change

WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift
jullianm: done

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
jullianm: https://github.com/wireapp/wire-ios/pull/3300/commits/2f26b82afb60d12c7aae312dbb43e0b4b91bd8a2

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1095,20 +1095,21 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+                            journal: journal
+                        )
 
-                    self.triggerMigrationsNeedsActionsIfNeeded(
-                        journal: journal,
-                        userSession: userSession
-                    )
+                        Task {
+                            await userSession.performAppMigrationsIfNeeded()
+                        }
```
netbe: issue: why did this change? this will not wait anymore for the migration to be finished

netbe: see on develop, I marked the startBackgroundSession as mainActor:

```
 Task {
                    if self.shouldEnableSyncV2(journal: journal) {
                        await self.enableSyncV2(
                            journal: journal,
                            coreDataStack: coreDataStack
                        )
                    }
                   // this is run on MainActor
                    let userSession = await self.startBackgroundSession(
                        for: account,
                        with: coreDataStack,
                        journal: journal
                    )
                   // ignore this
                    await userSession.migrateToConsumableNotificationsIfNeeded()

                    await userSession.triggerSync()

                    await MainActor.run {
                        onCompletion(userSession)
                    }
                }

File: WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift (hunk 2 above)
Comment (johnxnguyen) [reply WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift]: @netbe even with new sync it will be necessary to use such a flag, but certainly we should have more control about when these sync requests are made.

File: WireDomain/Sources/WireDomain/AppVersionMigration/AppVersionMigrationService.swift [hunk 5]
```diff
@@ -100,7 +100,7 @@ extension JournalProtocol {
 
 public extension JournalProtocol {
 
-    internal(set) var lastCompletedAppVersionMigration: SemanticVersion? {
+    var lastCompletedAppVersionMigration: SemanticVersion? {
```
johnxnguyen, line 109: question: why do we make this public settable?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_1_0.swift
```diff
@@ -0,0 +1,38 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireDomain
+import WireLogging
+
+struct AppVersionMigration_4_1_0: AppVersionMigration {
+
+    var version: SemanticVersion { SemanticVersion("4.1.0") }
```
johnxnguyen: suggestion: could be a `let`

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_1_0.swift
```diff
@@ -0,0 +1,38 @@
...
+
+    init(
+        performResourceSync: @escaping () -> Void
+    ) {
+        self.performResourceSync = performResourceSync
+    }
+
+    func perform() async throws {
+        // Performs a resources sync to ensure all conversations are up to date.
+        performResourceSync()
```
johnxnguyen: suggestion: we could be more targeted and just pull conversations.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_1_0.swift
```diff
@@ -0,0 +1,38 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireDomain
+import WireLogging
+
+struct AppVersionMigration_4_1_0: AppVersionMigration {
```
johnxnguyen: suggestion: I think a comment describing what this migration solves is useful for the future.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -366,6 +367,12 @@ public final class ZMUserSession: NSObject {
         mlsService: mlsService
     )
 
+    private lazy var appVersionMigrationService: AppVersionMigrationService = .init(
+        journal: journal,
+        currentVersion: SemanticVersion(stringLiteral: currentAppVersion),
+        allMigrations: makeAppVersionMigrations()
+    )
```
johnxnguyen: suggestion: I don't think we really need to retain this... just create it when needed, use it, then discard.

wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DebugActions/DeveloperDebugActionsViewModel.swift
```diff
@@ -81,7 +82,12 @@ final class DeveloperDebugActionsViewModel: ObservableObject {
             .init(title: "Delete domains in the database", action: deleteDomains),
             .init(title: "Find Conversation with MLS Group", action: showSearchMLSConversations),
             .init(title: "Clear collapsed messages cache", action: clearCollapsedMessagesCache),
-            .init(title: "Simulate access token failure", action: simulateAccessTokenFailure)
+            .init(title: "Simulate access token failure", action: simulateAccessTokenFailure),
+            .init(
+                title: "Create MLS group conversation with missing metadata",
+                action: createMLSGroupConversationWithMissingMetadata
+            ),
+            .init(title: "Perform app version migrations", action: performAppVersionMigrations)
```
johnxnguyen: question: is this temporary to help testing?

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
jullianm: done thx

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_1_0.swift
jullianm: cannot do, it's a computed property

wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DebugActions/DeveloperDebugActionsViewModel.swift
jullianm: yes

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_1_0.swift
jullianm: done https://github.com/wireapp/wire-ios/pull/3300/commits/de329fd263d6ebf5af47349eaef052d8302cf1e5, if add some actions for 4_1_0 we can add it to the doc

WireDomain/Sources/WireDomain/AppVersionMigration/AppVersionMigrationService.swift
jullianm: to access it for debug actions, once we tested it, I'll revert this change

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_1_0.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/3300/commits/de329fd263d6ebf5af47349eaef052d8302cf1e5
I constructed the PullAllResourcesSync from scratch. Maybe that's better to make it public in userSessionComponent and access it directly from there..

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
jullianm: done https://github.com/wireapp/wire-ios/pull/3300/commits/de329fd263d6ebf5af47349eaef052d8302cf1e5

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1094,18 +1095,15 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+                        for: account,
+                        with: coreDataStack,
+                        journal: journal
+                    )
 
-                        self.triggerMigrationsNeedsActionsIfNeeded(
-                            journal: journal,
-                            userSession: userSession
-                        )
+                    await userSession.performAppMigrationsIfNeeded()
```
johnxnguyen: issue: these migrations may take some time, so I think we should move the UI to the migration screen.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_1_0.swift
johnxnguyen: issue: the comment doesn't give me any new information that I don't already have by looking at the implementation. It would be more helpful to provide context about why the migration is necessary. For example:

```
/// Issue: some users had conversations in their database that weren't fully up do date with the backend.
/// Fix: re-sync all conversations.
```

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_1_0.swift
johnxnguyen: How about: 

```swift
let version: SemanticVersion = "4.1.0"
```

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1550,3 +1568,54 @@ extension ZMUserSession {
...
+                pullAllConversationsSync: pullAllConversationsSync
+            )
+
+            appVersionMigrations.append(appVersionMigration4_1_0)
+        }
+
+        return appVersionMigrations
+    }
+
+    private func makePullAllConversationsSync() -> (any PullAllConversationsSyncProtocol)? {
```
johnxnguyen: question: can we not make user of the client session component to obtain these dependencies?

wire-ios/Wire-iOS UnitTests/Developer/DeveloperTools/DebugActions/DeveloperDebugActionsViewModelTests.swift
```diff
@@ -27,7 +27,7 @@ final class DeveloperDebugActionsViewModelTests: XCTestCase {
 
         // when
         // then
-        XCTAssertEqual(viewModel.debugItems.count, 12)
+        XCTAssertEqual(viewModel.debugItems.count, 14)
```
johnxnguyen, line 30: thought: why do we even have these tests?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_1_0.swift
KaterinaWire: I think "Issue" can sometimes contain sensitive information. We can of course write without details, but is it necessary? Jira ticket should contain this information.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_1_0.swift
jullianm: done

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_1_0.swift
jullianm: done

wire-ios/Wire-iOS UnitTests/Developer/DeveloperTools/DebugActions/DeveloperDebugActionsViewModelTests.swift
jullianm: not sure, doesn't seem very useful as this is not prod stuff

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
jullianm: done

wire-ios/Wire-iOS UnitTests/Developer/DeveloperTools/DebugActions/DeveloperDebugActionsViewModelTests.swift
jullianm: any objections removing these test class @netbe  ?

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/3300/commits/e0cf8f5ebc97f49a327a0610636c4c799170ad4a

important point though: I noticed that when we do perform the app version migration action (i.e pulling conversations) the loading screen remains until the whole app is synced, so user will see the loading screen (screenshot below) until app is fully synced (including resolved missing groups/channels).

works as usual (with the sync status bar) if no app version migration actions is performed

is that acceptable ?

IMG_0083

KaterinaWire: We will need to migrate "something that we maybe shouldn't mention here " soon, which will mean the user should not use the app until the migration is complete

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1094,18 +1096,17 @@ public final class SessionManager: NSObject, SessionManagerType {
...
-                            with: coreDataStack,
-                            journal: journal
-                        )
+                    let userSession = await self.startBackgroundSession(
+                        for: account,
+                        with: coreDataStack,
+                        journal: journal
+                    )
 
-                        self.triggerMigrationsNeedsActionsIfNeeded(
```
johnxnguyen, line 1104: issue: we still need this line

johnxnguyen (changes requested):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_1_0.swift
johnxnguyen: @KaterinaWire certainly we should not reveal sensitive details until patches are released, but even still it would be good to fill in that info post release.

WireDomain/Sources/WireDomain/Repositories/Conversations/Protocols/ConversationRepositoryProtocol.swift
```diff
@@ -66,10 +66,6 @@ public protocol ConversationRepositoryProtocol {
         domain: String?
     ) async -> ZMConversation
 
-    /// Fetches and persists all conversations
-
-    func pullConversations() async throws
-
```
johnxnguyen, line 72: dead code

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1159,21 +1182,6 @@ public final class SessionManager: NSObject, SessionManagerType {
         }
     }
 
-    /// Executes post migration slow sync or sync resources
-    private func triggerMigrationsNeedsActionsIfNeeded(
```
johnxnguyen, line 1163: moved to user session

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1094,18 +1107,28 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+
+                    let migrationService = userSession.makeAppVersionMigrationService()
+                    if migrationService.isMigrationNeeded {
+                        await self.delegate?.sessionManagerWillMigrateAccount()
+
+                        do {
+                            try await migrationService.performAppMigrations()
+                        } catch {
+                            WireLogger.session.error(
+                                "Failed to perform app version migrations: \(String(describing: error))"
```
KaterinaWire, line 1124: **note:** The security team asked what kind of error we are going to log and whether it contains sensitive information. In this migration we are not going to throw any errors, but for a future migration we don't know. Maybe it is better to log the version of the migration that failed? @johnxnguyen and @netbe  what do you think?

KaterinaWire (approved):

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnxnguyen: Yes, logging at least the migration version is helpful.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1095,21 +1107,28 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+                            )
                         }
+                    }
 
-                        self.triggerMigrationsNeedsActionsIfNeeded(
-                            journal: journal,
-                            userSession: userSession
-                        )
+                    userSession.triggerSyncsIfNeeded()
```
netbe, line 1129: question: why is this non-blocking? I thought we were aiming to have all syncs asynchronous and to wait for them before returning the userSession, no?

WireDomain/Sources/WireDomain/Synchronization/PullAllConversationsSync.swift
```diff
@@ -19,29 +19,32 @@
 import Foundation
 import WireAPI
 
-struct PullAllConversationsSync: PullAllConversationsSyncProtocol {
+public final class PullAllConversationsSync: PullAllConversationsSyncProtocol {
```
netbe, line 22: question: why does it become a class

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1095,21 +1107,28 @@ public final class SessionManager: NSObject, SessionManagerType {
...
-                            await userSession.performAppMigrationsIfNeeded()
+                    let userSession = await self.startBackgroundSession(
+                        for: account,
+                        with: coreDataStack,
+                        journal: journal
+                    )
+
+                    let migrationService = userSession.makeAppVersionMigrationService()
+                    if migrationService.isMigrationNeeded {
+                        await self.delegate?.sessionManagerWillMigrateAccount()
```
netbe, line 1118: question: do we have a counter part where we dismiss the migration UI?

netbe (commented):
left some comments for understanding

WireDomain/Sources/WireDomain/Synchronization/PullAllConversationsSync.swift
johnxnguyen: to mutate the journal. Maybe the journal can be a class instead?

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnxnguyen: Currently no, since the next step is to show "authenticated" state which happens after activating the user session. But in the future it I think it would make sense to do so.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnxnguyen: It previously wasn't blocking either. But I don't think such syncs should be blocking because they depend on the network and may take time to complete. Even now, when opening the app, we show the conversation list while the incremental sync is ongoing.

WireDomain/Sources/WireDomain/Synchronization/PullAllConversationsSync.swift
netbe: up to you

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
netbe: hmm i see your point, not a blocker for now as it's more general design for future. But in triggerSyncsIfNeeded, we also have triggerInitialSync which should/is blocking

netbe (approved):
Approving now

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnxnguyen: yes agreed (although that may take a long time too..) and we can address it during deterministic session loading.

johnxnguyen (approved):

PR #3302, opened 2025-07-03
fix: nonce force unwrapping - WPB-18559

dmitrysimkin:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18559" title="WPB-18559" target="_blank">BugWPB-18559</a>  [iOS] Crash on message.nonce
  </td></table>
  <br />
 

### Issue

Remove nonce force unwrapping in Coversation table view datasource

### Testing

_Describe how to test._

---

### Checklist

---

### DIFF
--- a/WireFoundation/Sources/WireFoundation/ThreadSafeDictionary.swift
+++ b/WireFoundation/Sources/WireFoundation/ThreadSafeDictionary.swift
@@ -31,9 +31,8 @@ public class ThreadSafeDictionary<Key: Hashable, Value> {
         }
     }
 
-    public func get(for key: Key?) -> Value? {
-        guard let key else { return nil }
-        return queue.sync {
+    public func get(for key: Key) -> Value? {
+        queue.sync {
             self.dictionary[key]
         }
     }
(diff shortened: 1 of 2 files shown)


netbe (approved):
LGTM;)

WireFoundation/Sources/WireFoundation/ThreadSafeDictionary.swift
```diff
@@ -31,8 +31,9 @@ public class ThreadSafeDictionary<Key: Hashable, Value> {
         }
     }
 
-    public func get(for key: Key) -> Value? {
-        queue.sync {
+    public func get(for key: Key?) -> Value? {
```
samwyndham: issue: Don't allow a nil key. Instead have the caller do this check.

(opinionated reason): It doesn't make sense to call `get` if we don't have a key. `ThreadSafeDictionary` shouldn't have to adjust its API for this caller. By looking at the API we have the impression that passing a nil value is valid and might have some behavior when in fact we just return early. If we follow this pattern `ThreadSafeDictionary.remove(for:)` could also accept an optional key. And what if we no longer need this behavior in the future. Should we re-adjust ThreadSafeDictionary.

samwyndham (changes requested):

WireFoundation/Sources/WireFoundation/ThreadSafeDictionary.swift
dmitrysimkin: Updated

samwyndham (approved):

PR #3311, opened 2025-07-04
feat: modify pushChannelV2 upToDate logic - WPB-18485

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18485" title="WPB-18485" target="_blank">TaskWPB-18485</a>  [iOS] Implement "end of queue" event and remove message_count
  </td></table>
  <br />
 

### Issue

This PR includes two things:
1. Tackling a tech debt, Moving PushChannelV2API to own folder: b6bd3f51a60574173ccaf6889d97b8a6c1aff3c3 - See ticket: https://wearezeta.atlassian.net/browse/WPB-18374
2. Adapting the consumable-notifications pushChannel:

* Remove messageCount message
* Add new synchronisation notification which is received via the websocket and inform that we reach the end of the queue, This means the client is now up to date. The syncMarker id is used to ensure we reached the correct synchronisation notification even if we get interrupted.
* Make consumable-notifications available only for api v9 and above

**Note:** because the first point won't compile on it's own, I included it with the second

### Testing

1. enable consumable-notifications Feature Flag
2. be on api version 9 - (set Preferred API version to v9)
3. login

verify sync bar should end

 

---

### Checklist

---


samwyndham (approved):
Nice work!

WireDomain/Sources/WireDomain/Synchronization/PullPendingUpdateEventsV2Sync.swift
```diff
@@ -46,7 +49,8 @@ public struct PullPendingUpdateEventsSyncV2: PullPendingUpdateEventsSyncV2Protoc
         updateEventsStore: any UpdateEventsLocalStoreProtocol,
         journal: Journal,
         decryptor: any UpdateEventDecryptorProtocol,
-        coreCryptoProvider: any CoreCryptoProviderProtocol
+        coreCryptoProvider: any CoreCryptoProviderProtocol,
+        syncMarkerGenerator: @escaping SyncMarkerGenerator = { UUID().uuidString }
```
dmitrysimkin, line 53: question: Is this default block only for tests? if so better not add to production code

dmitrysimkin (approved):
LGTM

WireDomain/Sources/WireDomain/Synchronization/PullPendingUpdateEventsV2Sync.swift
netbe: it's the default for prod, it's here to let tests override it

PR #3309, opened 2025-07-04
feat: remove draft attachments - WPB-17621

samwyndham:
### Issue

This PR adds the ability to remove a draft attachment, either by tapping the `X` button beside a draft or by selecting `Remove File` on the attachments error menu.

https://github.com/user-attachments/assets/f64e2043-9688-4797-806e-9426006de396

Specifically it:

- Adds a `DeleteDraftUseCase`.
- Wires up the use case.

Additionally it:

- Fixes a bug uploading small files.
- Fixes the need for a manual `DraftsRepositoryProtocolMock.swift` replacing it with an auto generated one.

### Testing

Not currently manually testable

---

### Checklist

- [ ] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [ ] Description is filled and free of optional paragraphs.
- [ ] Adds/updates automated tests.

---

### UI accessibility checklist

_If your PR includes UI changes, please utilize this checklist:_
- [ ] Make sure you use the API for UI elements that support large fonts.
- [ ] All colors are taken from WireDesign.ColorTheme or constructed using WireDesign.BaseColorPalette.
- [ ] New UI elements have Accessibility strings for VoiceOver.

PR #3315, opened 2025-07-04
feat: retry draft attachment upload - WPB-17622

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17622" title="WPB-17622" target="_blank">TaskWPB-17622</a>  [iOS] Retry draft attachment upload
  </td></table>
  <br />
 

### Issue

This PR adds the ability to retry uploading a draft attachment by selecting `Retry` on the attachments error menu.

https://github.com/user-attachments/assets/66fbc8e8-5838-4cfb-968a-2b626472cb30

To achieve this it:

- extends `UploadDraftUseCase` with a method to upload a file based on a drafts node ID - this is used to fetch metadata (eg. file location etc) from the DraftsRepository.
- Moves logic for uploading drafts from the `DraftsRepository` to `UploadDraftUseCase`. As mentioned in https://github.com/wireapp/wire-ios/pull/3309, an anti-pattern was emerging where all the logic was being added to the `DraftsRepository` and the use cases were just empty shells. In this PR I continue to change things so that DraftsRepository will essentially be storage of WireCellsDraft items with CRUD access and business logic moved to the use case where it belongs.
- Removes the concept of keeping failed upload state within the upload manager as this is not needed.

####

This PR also addresses https://wearezeta.atlassian.net/browse/WPB-17616

### Testing

Not currently manually testable

---

### Checklist


WireCells/Sources/WireCellsImplementation/DraftsRepository.swift
```diff
@@ -26,19 +26,13 @@ import WireLogging
...
-        assetURL: URL,
-        assetSize: Int,
-        cellName: String,
-        fileName: String,
-        fileType: UTType?,
-        deleteAfterUpload: Bool
-    ) async
     func drafts(for cellName: String) async -> AsyncStream<[WireCellsDraft]>
     func publishAll(for cellName: String) async throws
     func clearPublishedDrafts(for cellName: String) async
```
samwyndham, line 31: These methods will also eventually be removed from this protocol and there logic moved to existing use cases.

WireCells/Sources/WireCellsImplementation/DraftsRepository.swift
```diff
@@ -76,56 +70,6 @@ package actor DraftsRepository: DraftsRepositoryProtocol {
         continuations.values.forEach { $0.finish() }
     }
 
-    package func add(
```
samwyndham, line 79: This logic has moved to `UploadDraftUseCase`

WireCells/Sources/WireCellsImplementation/WireCellsNodeUploadManager.swift
samwyndham, line 1: Here I remove the responsibility of retrying an upload and managing the state of failed uploads. This is the correct thing to do because an upload may fail for example due to a file name conflict. Trying to re-upload from here wouldn't allow fixing the file name so the upload would always fail.

WireCells/Tests/WireCellsImplementationTests/UseCases/UploadDraftUseCaseTests.swift
```diff
@@ -28,14 +28,115 @@ final class UploadDraftUseCaseTests {
...
 
+    // MARK: - invoke(nodeID:)
+
+    @Test
+    func invokeWithNodeID_whenUnknownDraft() async {
+        // Given
+        let nodeID = UUID()
+
+        // When, Then
+        let sut = sut
```
johnxnguyen, line 52: question: what is the purpose of this assignment?

WireCells/Tests/WireCellsImplementationTests/UseCases/UploadDraftUseCaseTests.swift
```diff
@@ -28,14 +28,115 @@ final class UploadDraftUseCaseTests {
...
+        await #expect(throws: (any Error).self) {
```
johnxnguyen: question: is there a particular error to assert?

WireCells/Tests/WireCellsImplementationTests/UseCases/UploadDraftUseCaseTests.swift
```diff
@@ -28,14 +28,115 @@ final class UploadDraftUseCaseTests {
...
+        let nodeID = UUID()
+
+        draftsRepository.fetchDraftNodeIDUUIDCellNameStringWireCellsDraftReturnValue = WireCellsDraft.fixture(
+            nodeID: nodeID,
+            status: .uploading(progress: 0.5),
+            name: "original.txt"
+        )
+
+        uploadManager
+            .uploadNodeIDUUIDVersionIDUUIDAssetPathURLAssetSizeUInt64DestNodePathString_NodeWireCellsNodeStreamAsyncStreamWireCellsUploadStatusReturnValue
```
johnxnguyen, line 70: though: oof these long mock names are kind of annoying

johnxnguyen (approved):
Nice work

WireCells/Tests/WireCellsImplementationTests/UseCases/UploadDraftUseCaseTests.swift
samwyndham: In this case the `#expect` macro fails to compile without it 😠

caldrian (approved):

WireCells/Tests/WireCellsImplementationTests/UseCases/UploadDraftUseCaseTests.swift
samwyndham: Yep. Crazy annoying.

WireCells/Tests/WireCellsImplementationTests/UseCases/UploadDraftUseCaseTests.swift
samwyndham: Fixed. I left the other similar error in this test case unchanged as it is thrown by the system and didn't want to be to exact there.

PR #3335, opened 2025-07-11
chore: store logs of app extensions in shared container - WPB-15114

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15114" title="WPB-15114" target="_blank">TaskWPB-15114</a>  Export logs from NSE on prod
  </td></table>
  <br />
 

### Issue

Currently logs created by our app extensions are written into their sandbox only.
This PR changes the logger to instead place the files in the shared container.

---

### Checklist

---


wire-ios-sync-engine/Source/UserSession/CacheFileRelocator.swift
```diff
@@ -20,10 +20,15 @@ import Foundation
 
 struct CacheFileRelocator {
 
-    // whitelisted files, so the FileRelocator doesn't consider to check these system files.
-    // - com.apple.nsurlsessiond is used by the system as cache while sharing an item.
-    // - .DS_Store is the hidden file for folder preferences used in macOS (only for simulator)
-    private let whitelistedFiles = ["com.apple.nsurlsessiond", ".DS_Store"]
+    // Whitelisted files, so the FileRelocator doesn't consider to check these system files.
```
samwyndham: nitpick:

```suggestion
    // Whitelisted files, so the FileRelocator doesn't consider to check these files.
```

wire-ios/WireCommonComponents/Logging/CocoaLumberjackLogger.swift
```diff
@@ -24,11 +24,17 @@ import WireSystem
...
+    private let fileLogger: DDFileLogger
     private var tags = [LogAttributesKey: String]()
     private let tagsQueue = DispatchQueue(label: "CocoaLumberjackLogger.tagsQueue", attributes: .concurrent)
 
-    init() {
+    /// - Parameter logsDirectory: If `nil` the default logs directory of `CocoaLumberjack` is used, otherwise the provided URL.
+    init(logsDirectory: URL?) {
+        let logFileManager = DDLogFileManagerDefault(
+            logsDirectory: logsDirectory?.path(),
+            defaultFileProtectionLevel: .none
```
samwyndham: issue(maybe): I'm not really sure what to use here but I expect we shouldn't be using .none. See here: https://developer.apple.com/documentation/foundation/fileprotectiontype

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
```diff
@@ -43,12 +43,57 @@ public enum WireAnalytics {
...
+            CocoaLumberjackLogger(logsDirectory: target.logsDirectory)
             WireAnalytics.Datadog.shared
         }
 
         // pass tags to Datadog through WireLogger
         WireLogger.system.addTag(.processId, value: "\(ProcessInfo.processInfo.processIdentifier)")
         WireLogger.system.addTag(.processName, value: ProcessInfo.processInfo.processName)
     }
+
+    public enum Target {
```
samwyndham: suggestion: _maybe_ this is better as a struct?

```swift
struct Target {
    let logsDirectory: URL?

    static let app = Target(
        logsDirectory: // ...
    )
}
```

samwyndham (commented):
Nice work... maybe there is an issue regarding `defaultFileProtectionLevel`. What do you think?

wire-ios/WireCommonComponents/Logging/CocoaLumberjackLogger.swift
caldrian: I tried to keep the current behavior. If I'm not mistaken, we currently also have no file protection set on the log files.
Do you think because we are putting the files in a different location now it is more important to set a value?
Probably it won't hurt to just set one. 
@samwyndham

caldrian: Actually, since the notification service extension could run just right after booting the device, I'm not sure we can set one.
What do you think @netbe ?

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
caldrian: I like that you challenge this enum.
In my opinion we often use enums where structs would save us a lot of `switch` statements for example.

However, just in this situation I think I prefer the enum as it is.
Could you elaborate on why you consider a struct the better alternative, please?
@samwyndham

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportRouter.swift
```diff
@@ -93,8 +93,7 @@ final class SettingsDebugReportRouter: NSObject, SettingsDebugReportRouterProtoc
         let body = mailComposeViewController.prefilledBody()
         mailComposeViewController.setMessageBody(body, isHTML: false)
 
-        activityIndicator.stop()
-        let topMostViewController = UIApplication.shared.topmostViewController(onlyFullScreen: false)
+        activityIndicator.start()
```
netbe, line 96: question: what happened here?

wire-ios/WireCommonComponents/Logging/CocoaLumberjackLogger.swift
netbe: we use completeUntilFirstUserAuthentication everywhere, and since iOS Y this is the default value if not specified if i recall correctly.

I am not sure about this case @caldrian, do you expect not to have the user authenticated?

netbe (commented):
good first draft, I'd be interested to know if that actually wrote the files or if we are able to retrieve them

wire-ios/WireCommonComponents/Logging/CocoaLumberjackLogger.swift
caldrian: I deleted the argument so that the default behavior applies, which is `.completeUntilFirstUserAuthentication` in our case (for the main app due to it's background running abilities and for the app extensions).

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/SettingsDebugReportRouter.swift
caldrian: The `topMostViewController` variable was never used, so I just removed the line.
The `activityIndicator` should actually be started before, and stopped afterwards, but both times `stop` was called.

wire-ios/WireCommonComponents/Logging/CocoaLumberjackLogger.swift
caldrian: > do you expect not to have the user authenticated?

As I understood it in the case that the notification service extension runs before the device is unlocked, a new log-file is created.

netbe: but when you boot, you have to authenticate = enter your pin code

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
```diff
@@ -43,12 +43,56 @@ public enum WireAnalytics {
...
+        switch self {
+
+        case .app:
+            let cachesDirectory = try? fileManager.url(
+                for: .cachesDirectory,
+                in: .userDomainMask,
+                appropriateFor: nil,
+                create: true
+            )
+            return cachesDirectory?.appending(path: "Logs", directoryHint: .isDirectory)
```
netbe: todo: I think we should update LogFilesProvider to use this location too then. So it's defined only in one place.
To answer your question about moving also the main app logs to shared group, we can do it, but we have to make sure we clean up:
- after sending logs 
- after logout

netbe (commented):
left another comment

wire-ios/WireCommonComponents/Logging/CocoaLumberjackLogger.swift
caldrian: > but when you boot, you have to authenticate

But at this point notifications might have already arrived, or am I wrong?

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
samwyndham: I don't have a strong argument here - I tend to prefer using structs for configuration where the enums don't have associated values. It could be a prejudice I picked up after reading [this](https://www.jessesquires.com/blog/2016/07/31/enums-as-configs/) many years ago but I don't think those arguments apply here. Feel free not to change anything.

samwyndham (approved):

WireLogging/Sources/WireLogging/AggregatedLogger.swift
```diff
@@ -31,10 +32,6 @@ final class AggregatedLogger: LoggerProtocol {
 
     // MARK: - LoggerProtocol
 
-    var logFiles: [URL] {
-        loggers.reduce(into: []) { $0 += $1.logFiles }
-    }
-
```
caldrian, line 37: Not needed anymore since the only type implementing this was `CocoaLumberjackLogger` and it basically just listed all contents of the `logsDirectory` argument we are now passing in.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_1_1.swift
```diff
@@ -46,8 +46,8 @@ final class AppVersionMigration_4_1_1: AppVersionMigration {
         journal[.isConversationSyncRequired] = true
 
         // Deletes all raw log files collected from the logger sources.
-        // This removes files from `WireLogger.logFiles` and `ZMSLog.pathsForExistingLogs`,
-        try logFilesProvider.removeLogFiles()
+        // This removes files from CocoaLumberjackLogger and `ZMSLog.pathsForExistingLogs`,
```
netbe, line 49: **suggestion:** as of 4.1.1, no ZMSLog file should be there anynore, we might want to update this

wire-ios-sync-engine/WireSyncEngine.xcodeproj/project.pbxproj
```diff
@@ -3242,7 +3230,6 @@
 				63497702268A20EC00824A05 /* AVSParticipantsChange.swift in Sources */,
 				8717DFA71F6EF44E0087EFE4 /* SessionManager+Push.swift in Sources */,
 				D52257232062637500562561 /* DeletableAssetIdentifierProvider.swift in Sources */,
-				015E2F8B2E26A53A00470025 /* AppVersionMigration_4_2_0.swift in Sources */,
```
netbe: todo: not sure what happens here, just make sure the 2 app version migrations are correctly in the project

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/LogFilesProvider.swift
```diff
@@ -46,88 +46,93 @@ struct LogFilesProvider: LogFilesProviding {
...
     }
 
     // MARK: - Interface
 
     func generateLogFilesData() throws -> Data {
+        let fileManager = FileManager.default
         defer {
             // because we don't rotate file for this one, we clean it once sent
             // this regenerated from os_log anyway
             if let url = LogFileDestination.main.log {
```
netbe: question: is LogFileDestination still relevant?

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
```diff
@@ -39,16 +40,71 @@ public enum WireAnalytics {
...
         WireLogger.system.addTag(.processName, value: ProcessInfo.processInfo.processName)
     }
+
+    private static func logsDirectory(for target: LogTarget) -> URL? {
+        guard let appGroupIdentifier = Bundle.main.applicationGroupIdentifier else { return nil }
+        return FileManager.default.sharedLogsDirectoryURL(for: appGroupIdentifier)?
+            .appending(component: target.rawValue, directoryHint: .isDirectory)
+    }
+
+    /// With release 4.2.0 the log files of all targets (main app, notification service extension and share extension)
```
netbe, line 62: suggestion: we can move this to the 4.2.0 migration so it's run once only

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
```diff
@@ -39,16 +40,71 @@ public enum WireAnalytics {
...
+                try fileManager.createDirectory(at: newRootDirectory, withIntermediateDirectories: true)
+            }
+
+            for previousLogFileURL in previousLogFileURLs {
+                let newLogFileURL = newRootDirectory.appending(
+                    path: previousLogFileURL.lastPathComponent,
+                    directoryHint: .notDirectory
+                )
+                if fileManager.fileExists(atPath: newLogFileURL.path()) {
+                    continue // don't overwrite, instead delete the old file (unlikely anyway)
```
netbe: question: as you update, you'll start writing in the new location, even before you migrate old files, no?

we can simply move the old file in the new location to be zipped and let the new file be zipped when we send

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
```diff
@@ -39,16 +40,71 @@ public enum WireAnalytics {
...
+                }
+                try fileManager.moveItem(at: previousLogFileURL, to: newLogFileURL)
+            }
+
+            try fileManager.removeItem(at: previousRootDirectory)
+
+        } catch {
+            let logger = os.Logger(subsystem: Bundle.main.bundleIdentifier!, category: "analytics")
```
netbe: question: why using os.Logger here? also why analytics category?

netbe (commented):

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
caldrian: Thanks for the comment, I changed the category to "system".
I'm not sure about which other way to log, since we're here about to initialize `WireLogger`.

wire-ios-sync-engine/WireSyncEngine.xcodeproj/project.pbxproj
caldrian: I had converted some groups to folders, but reverted now.

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
caldrian: The `AppVersionMigration_4_2_0` code runs once in the main app and cannot access the sandboxes of the share and notification service extensions.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_1_1.swift
caldrian: What would you like to update?
Shouldn't the migrations be kept in place for users who skipped one version?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Debug Report/LogFilesProvider.swift
caldrian: I haven't noticed `LogFileDestination` before.
Also I don't find any place where we write to these files. I think we can clean it up.

caldrian: I created a separate PR: https://github.com/wireapp/wire-ios/pull/3354

wire-ios/WireCommonComponents/Analytics/WireAnalytics.swift
caldrian: > you'll start writing in the new location, even before you migrate old files, no?

No, I put this code in the initialization, so it will run the first time after updating.

But it will run for each target, so each of the tree executables (app, nse, se) will have to move its own log files into the shared container.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_1_1.swift
netbe: right I might have missed that we're in a migration here 🫠

netbe (approved):
Looks good to me

PR #3340, opened 2025-07-15
chore: unify tracking identifier name and type - WPB-17530

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17530" title="WPB-17530" target="_blank">StoryWPB-17530</a>  [iOS] New Registration Flows - Personal account countly tracking
  </td></table>
  <br />
 

### Issue

Currently we use three different terms for an analytics identifier:
- tracking id/identifier
- analytics id/identifier
- devide id/identifier

Moreover, we are not consistent with the type. Sometimes we use String, sometimes UUID.

This PR changes the name and type to trackingID/UUID wherever possible.

---

### Checklist

---


WireAnalytics/Sources/WireAnalytics/AnalyticsUser/AnalyticsUser.swift
```diff
@@ -22,7 +22,7 @@ public struct AnalyticsUser: Equatable, Sendable {
 
     /// A unique id.
 
-    public let analyticsIdentifier: String
+    public let trackingID: String
```
caldrian: I chose trackingID because in the Protobuf declarations it is also called tracking identifier.

WireAnalytics/Sources/WireAnalytics/AnalyticsUser/AnalyticsUser.swift
```diff
@@ -31,14 +31,14 @@ public struct AnalyticsUser: Equatable, Sendable {
     /// Create a new `AnalyticsUser`.
     ///
     /// - Parameters:
-    ///   - analyticsIdentifier: A uniqe id.
+    ///   - trackingID: A uniqe id.
```
samwyndham, line 36:
```suggestion
    ///   - trackingID: A unique id.
```

wire-ios-data-model/Source/Model/User/TrackingIDProvider.swift
```diff
@@ -19,42 +19,43 @@
...
-        let message = DataTransfer(trackingIdentifier: identifier)
+    private func broadcast(trackingID: String, context: NSManagedObjectContext) {
+        let message = DataTransfer(trackingIdentifier: trackingID)
         do {
             try ZMConversation.sendMessageToSelfClients(message, in: context)
         } catch {
+            let redactedID = UUID(uuidString: trackingID)?.safeForLoggingDescription ?? "invalid UUID"
             WireLogger.messaging
-                .error("Error broadcasting analytics ID: \(identifier.safeForLoggingDescription) \(error)")
+                .error("Error broadcasting analytics ID: \(redactedID) \(error)")
```
samwyndham:
```suggestion
                .error("Error broadcasting tracking ID: \(redactedID) \(error)")
```

samwyndham (approved):
Great! Two things:

- I assume our setup allows for renaming `User.analyticsIdentifier` in the core data model. I think we should do that too, perhaps when we are updating the model in another PR. Consider making a jira ticket for the specifically.
- Why do you choose String instead of UUID? I'm actually not sure what is best - I'm just interested.  (Out of scope)I also wonder if it would be better to have another define another type of `Identifier` that enforces using the transport string etc as I've made mistakes here and wondered why things are broken.

wire-ios-data-model/Source/Utilis/Protos/GenericMessage+Helper.swift
```diff
@@ -895,9 +895,9 @@ public extension DataTransfer {
 // MARK: - TrackingIdentifier
 
 extension TrackingIdentifier {
-    init(_ uuid: UUID) {
+    init(_ id: String) {
         self = TrackingIdentifier.with {
-            $0.identifier = uuid.transportString()
```
KaterinaWire: **question:** we make an `uuid ` as `lowercased()` here, are you sure it will stay the same?

caldrian: I think we always used `transportString()` except in this one location: https://github.com/wireapp/wire-ios/pull/3340/files#diff-a439038e3cb01f32be13c4b0cb65a797814d1e2f035d3c2d13218c927aa66859L52
And that one is probably a mistake.

KaterinaWire (approved):

PR #3344, opened 2025-07-16
feat: use wire auth token for cells - WPB-18754

samwyndham:
### Issue

This PR updates cells network clients to use the API token we use for other wire endpoints. Previously the app used fixed tokens that were input in iOS settings. Using fixed tokens is still possible if necessary via feature flag.

To achieve the above this PR:

- Shuffles things around so that the some cells state is tied to the lifetime of a logged in user (previously these were static).
- Ensures that `AuthenticationManager` is used to provide authentication tokens - the particular one used is that which belongs to the `ClientSessionComponent` as we shouldn't use multiple of these per logged in user due to race conditions

### Testing

Wire cells functionality in the iOS client is currently very limited so the following steps are very specific.

1. Follow steps defined here to login: https://wearezeta.atlassian.net/wiki/spaces/PAD/pages/1948680479/2025-06-12+Wire+Cells+iOS+configuration. Use `imai` envionment with the login credentials for `sam`. **Remember to set the correct feature flags.**
2. Navigate to the `iOS Dev 2` group
3. Attach a file using _attach_ icon - the file will start uploading.
4. Once the file is uploaded tap send
5. Confirm the message is sent - right now the message will display as just dummy text listing the file name.

---

### Checklist


WireNetwork/Sources/WireNetwork/Authentication/AuthenticationManager.swift
```diff
@@ -44,13 +44,13 @@ public actor AuthenticationManager: AuthenticationManagerProtocol {
...
     private var currentToken: CurrentToken?
-    private let clientID: String?
+    private let clientID: String
     private let cookieStorage: any CookieStorageProtocol
     private let networkService: any NetworkServiceProtocol
     private let onAuthenticationFailure: () -> Void
 
     public init(
-        clientID: String?,
+        clientID: String,
```
samwyndham: There was no need for this to be optional

wire-ios/Wire-iOS Tests/Mocks/WireCells/WireCellsClearPublishedDraftsUseCaseProtocolMock.swift
```diff
@@ -0,0 +1,27 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireCellsAPI
+
+final class WireCellsClearPublishedDraftsUseCaseProtocolMock: WireCellsClearPublishedDraftsUseCaseProtocol {
+
+    func invoke() async {
+        fatalError("Implement")
```
samwyndham, line 24: These are currently just stubs which is enough for now. I think these are going to be shuffled around with the move to clean architecture and I would rather use Sourcery to generate things (I can't can't currently) so am leaving them as they are.

wire-ios/Wire-iOS/Resources/Settings.bundle/DeveloperSettings.plist
```diff
@@ -36,34 +36,20 @@
...
+			<key>Key</key>
+			<string>ZMWireCellsAccessToken</string>
+			<key>DefaultValue</key>
+			<string></string>
+			<key>AutocapitalizationType</key>
+			<string>None</string>
+			<key>AutocorrectionType</key>
+			<string>No</string>
+		</dict>
 	</array>
```
samwyndham, line 53: This manual setting is no longer needed

WireCells/Sources/WireCellsImplementation/NodesAPI/RestAPI.swift
```diff
@@ -32,17 +32,16 @@ final class RestAPI: Sendable {
...
-    private let configuration: CellsSDKAPIConfiguration
+    private let serverURL: URL
+    private let accessTokenProvider: any AccessTokenProvider
 
-    init(serverURL: URL, accessToken: String) {
-        let config = CellsSDKAPIConfiguration()
-        config.basePath = serverURL.absoluteString
-        config.customHeaders = ["Authorization": "Bearer \(accessToken)"]
-        self.configuration = config
+    init(serverURL: URL, accessToken: any AccessTokenProvider) {
```
caldrian, line 38: suggestion: make `RestAPI` a struct and remove `init`

WireCells/Sources/WireCellsImplementation/NodesAPI/RestAPI.swift
```diff
@@ -32,17 +32,16 @@ final class RestAPI: Sendable {
...
+        self.serverURL = serverURL
+        self.accessTokenProvider = accessToken
```
caldrian, line 40: question: Did you forget to rename `accessToken` to `accessTokenProvider` or is it intentionally like this?

caldrian (approved):
Great work!

jullianm (approved):
Looks good overall, what confuses me though is that some logic seems to be related to "networking" logic (AWSClient, NodesAPI, RestAPI) yet still live in WireCells. Shouldn't it be moved to WireNetwork ?

WireCells/Sources/WireCellsImplementation/NodesAPI/RestAPI.swift
samwyndham: Good idea. I don't know why I keep defaulting to classes here. I will change it a follow up PR.

WireCells/Sources/WireCellsImplementation/NodesAPI/RestAPI.swift
samwyndham: It is intentional as I tend to like shorter names. Internally I used accessTokenProvider so that I didn't call `accessToken.accessToken`. However perhaps this is confusing so I will update the name in a follow up PR. Thank you!

PR #3350, opened 2025-07-17
fix: ask user whether to erase data on logout - WPB-18862

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18862" title="WPB-18862" target="_blank">BugWPB-18862</a>  [iOS] Random device session expired
  </td></table>
  <br />
 

### Issue

**Context**: A customer is blocked on login screen after session expired message, cannot reauthenticate then forced to logout hence losing all history data.

**Causes**: This is a hotfix to prevent that issue from happening so.. root cause is not clearly identified yet. 

**Solution**: Temporary solution to fix this blocking issue for that customer is provide a pop up on logout to ask whether data should be erased or not - which also aligns with what other platforms currently do -

Simulator Screenshot - iPhone 14 - 18 5 - 2025-07-18 at 12 10 07

**Note:** translations have been validated and added directly here as it is for a hotfix release.

### Testing

- in debug actions, trigger "simulate access token failure" 
- tap on logout top right button
- "new" pop up shows up, tap "no"
- session will logout then login again
- all conversations history should be there (not erased)

### DIFF
--- a/wire-ios/Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
+++ b/wire-ios/Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
@@ -509,7 +509,7 @@ extension AuthenticationCoordinator {
             let yesAction = AuthenticationCoordinatorAlertAction(
                 title: L10n.Localizable.Self.Settings.AccountDetails.LogOut.EraseData.Alert.yes,
                 coordinatorActions: [.showLoadingView, .deleteSession(eraseData: true)],
-                style: .default
+                style: .destructive
             )
 
             let noAction = AuthenticationCoordinatorAlertAction(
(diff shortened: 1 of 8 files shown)


wire-ios/Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
```diff
@@ -502,32 +506,41 @@ extension AuthenticationCoordinator {
...
+                coordinatorActions: [.showLoadingView, .deleteSession(eraseData: false)],
+                style: .default
             )
 
             let alertModel = AuthenticationCoordinatorAlert(
-                title: L10n.Localizable.Self.Settings.AccountDetails.LogOut.Alert.title,
-                message: L10n.Localizable.Self.Settings.AccountDetails.LogOut.Alert.message,
-                actions: [.cancel, signOutAction]
+                title: L10n.Localizable.Self.Settings.AccountDetails.LogOut.EraseData.Alert.title,
+                message: nil,
```
netbe: suggestion: keep title and message. 
title: repeating the intention of user Logout
message: explaining the choices: Delete data...

netbe (commented):
overall looks good, just about the wording I have some doubts.

wire-ios/Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
jullianm: done, thanks!

netbe (approved):
LGTM;)

KaterinaWire (approved):

PR #3371, opened 2025-07-22
chore: update `WireMessaging` to follow architecture guidelines - WPB-18936

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18936" title="WPB-18936" target="_blank">TaskWPB-18936</a>  [iOS] Update `WireMessaging` to follow architecture guidelines 
  </td></table>
  <br />
 

### Issue

This PR updates the `WireMessaging` package to approximately follow the architecture guidelines as discussed here: https://wearezeta.atlassian.net/wiki/spaces/IOS/pages/1969651758/Architecture+guidelines. It aims to:

- ensure `WireMessaging` has the correct targets
- move existing source code in the package to the correct targets
- refactor code as minimally as possible - i.e. I haven't made any substantial changes to the types in those files.

The final result looks like:

Screenshot 2025-07-22 at 09 21 45

Note that:

- `WireMessagingData` is currently empty (non of the existing files fit that catergory)
- We will most likely need to introduce other test targets and test support targets. I didn't bother to create empty ones.

To achieve the above I mainly just moved files around. However I did make a the following additional changes:

- Split `WireConversationChannelIconAsset` & `WireConversationGroupIconAsset` so that their main definitions reside in the _domain_ target but code responsible for their images lives in the _ui_ target.
- Added specific tests for `WireConversationChannelIconAsset` & `WireConversationGroupIconAsset` to ensure that their images can be generated as their implementations rely on some string mangling. 
- Dropped the `Wire` prefix from all types within that package as it doesn't add any information.

Approximately the following was done:

- `WireMessagingUIBindings` &`WireMessagingBindings` -> `WireMessagingAssembly`
- `WireMessagingAPI` & `WireMessagingImplementation` -> `WireMessagingDomain`
- `WireMessagingResources` -> `WireMessagingDomain` & `WireMessagingUI`

### Testing

- Run automated tests

---

### Checklist


WireMessaging/Sources/WireMessagingDomain/ConversationChannelIcon.swift
```diff
@@ -38,16 +36,4 @@ package enum WireConversationChannelIconAsset: String {
...
-        "channel-icon-\(rawValue)"
-    }
-
-    package var image: Image {
-        Image(imageName, bundle: .module)
-    }
-
-    package var uiKitImage: UIImage {
-        UIImage(resource: .init(name: imageName, bundle: .module))
-    }
```
samwyndham, line 52: Moved to an extension in the _UI_ target

WireMessaging/Sources/WireMessagingDomain/ConversationGroupIcon.swift
```diff
@@ -74,16 +72,4 @@ package enum WireConversationGroupIconAsset {
...
-        "group-icon-\(index)"
-    }
-
-    package var image: Image {
-        Image(imageName, bundle: .module)
-    }
-
-    package var uiKitImage: UIImage {
-        UIImage(resource: .init(name: imageName, bundle: .module))
-    }
```
samwyndham, line 88: Moved to an extension in the _UI_ target

WireMessaging/Package.swift
```diff
@@ -64,26 +51,17 @@ let package = Package(
             plugins: [.plugin(name: "SwiftGenPlugin", package: "WirePlugins")]
         ),
         .target(
-            name: "WireMessagingImplementationSupport",
+            name: "WireMessagingDomainSupport",
```
johnxnguyen, line 54: question: seeing as WireMessageDomain is a product, I can understand the need for this support target so other modules can access mocks for public domain protocols. But I imagine that ideally the domain target is't publicly accessible and there would be no need for protocol mocks to be accessible outside the package. What do you think?

dmitrysimkin (approved):
LGTM 👍

WireMessaging/Package.swift
samwyndham: Yes I think you are right. I imagine that ideally `WireMessagingDomain` wouldn't be visible outside the package. But I expect `WireMessagingDomainSupport` might still be necessary so that mocks can be accessed from both `WireMessagingUITests` & `WireMessagingDataTests`. An alternative would be to have a single test target for the package, in which case the support target might not be necessary at all.

johnxnguyen: I think though the test targets could also depend on domain and the mocks are available in debug builds... not sure what's better.

johnxnguyen (approved):

WireMessaging/Package.swift
samwyndham: I see. So having the mocks residing in `WireMessagingDomain`. I'm going to merge this PR now to avoid further conflicts, but I will take a look at your suggestion - it looks like a worthwhile improvement - and try to to follow up in a new PR.

PR #3379, opened 2025-07-23
fix: missing messages during consumable-notifications sync system - WPB-18702

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18702" title="WPB-18702" target="_blank">BugWPB-18702</a>  [iOS] missing messages in async notifications
  </td></table>
  <br />
 

### Issue

With the consumable-notifications, backend sends push notifications even if the web socket is opened in the main app. While receiving messages, this will result in some messages not being displayed right away to the user. Some messages will be decrypted and stored by the NSE and will be later processed as pending events when the user put suspend and resume the sync.

To avoid this behavior, we create a lock on a file, so the NSE knows if the main app web socket is open.

This file is stored on the shared app container so both the main app and NSE can access it.

Introducing the `PushChannelState`:

- IncrementalSyncV2 set the lock on the file when its pushChannel is open and unlocks it when its pushChannel is closed.

- VerifyUserStep (on the NSE side) use the PushChannelState to verify if the lock is set or not. If it is set, it throws an error, and the NSE stops. Otherwise it continues the flow and a web socket will be opened to retrieve pending events from backend.

**Other changes**

* Reworked the batching logic: use an actor to ensure no messages are missed - 
* Disable batching once we're up to date (under developerFlag)

### Testing

1. spam conversation
2. check that no missing message occurs
3. check logs in datadog to monitor NSE activity
---

### Checklist

---


WireDomain/Sources/WireDomain/Notifications/Components/VerifyUserStep.swift
```diff
@@ -81,6 +89,9 @@ final class VerifyUserStep: Component<VerifyUserDependency>, VerifyUserStepProto
         }
 
         if journal[.isConsumableNotificationsEnabled] {
+
+            try verifyMainPushChannelClosed(clientID: selfClientID)
```
jullianm, line 93: question: why don't we do that check (and return nil) earlier in `loadNotificationService()` from `NotificationService`, I think we should not even initialize the NSE if main push channel is open.

jullianm:
@netbe looks good, a couple of questions just so I understand:
We only have this issue with PushChannelV2 (with consumable notifications enabled), right ? 
If so, shouldn't it be on backend side to fix that issue, not client side ?

netbe:
> @netbe looks good, a couple of questions just so I understand: We only have this issue with PushChannelV2 (with consumable notifications enabled), right ? If so, shouldn't it be on backend side to fix that issue, not client side ?

@jullianm it's not a trivial thing to do on backend + in the future we could have one client using multiple connections at the time to retrieve events from different queues. So this solution has been put it aside

WireDomain/Sources/WireDomain/Notifications/Components/VerifyUserStep.swift
netbe: I needed the clientID so that's why i put it here. Is there a way to retrieve the clientID + having the accountManager earlier? @jullianm

wire-ios-share-engine/Sources/SharingSession.swift
```diff
@@ -377,7 +377,8 @@ public final class SharingSession {
             wireAPIBackendEnvironment: wireAPIBackendEnvironment,
             minTLSVersion: .minVersionFrom(minTLSVersion),
             apiVersion: wireAPIVersion,
-            sharedUserDefaults: sharedUserDefaults
+            sharedUserDefaults: sharedUserDefaults,
+            sharedContainerURL: URL("unused")!
```
johnxnguyen, line 381: question: I don't quite understand the purpose of this URL.

WireDomain/Sources/WireDomain/Account/AccountManager.swift
```diff
@@ -210,6 +210,13 @@ public final class AccountManager: NSObject {
         }
     }
 
+    func selectedMainWebsocketFileURL() throws -> URL? {
```
johnxnguyen: thought: I feel like this should be in the `AccountData` directory instead (where also the crypto store, core data store, and other account related data should go). I would also consider calling it "Websocket file lock".

WireDomain/Sources/WireDomain/Notifications/Components/VerifyUserStep.swift
```diff
@@ -37,6 +37,7 @@ final class VerifyUserStep: Component<VerifyUserDependency>, VerifyUserStepProto
     enum Failure: Error {
         case noAccountFound
         case missingSelfClientID
+        case mainAppPushChannelOpened
```
johnxnguyen, line 40: suggestion: `pushChannelAlreadyOpen`

WireDomain/Sources/WireDomain/Notifications/Components/VerifyUserStep.swift
```diff
@@ -65,6 +66,13 @@ final class VerifyUserStep: Component<VerifyUserDependency>, VerifyUserStepProto
         super.init(parent: parent)
     }
 
+    func verifyMainPushChannelClosed(clientID: String) throws {
```
johnxnguyen, line 69: suggestion: verifyPushChannelIsClosed

WireDomain/Sources/WireDomain/Synchronization/PushChannelState.swift
```diff
@@ -0,0 +1,51 @@
...
+    func markAsOpen()
+    func markAsClosed()
+}
+
+struct PushChannelState: PushChannelStateProtocol {
+    let fileContext: SafeFileContext
+    init(sharedContainerURL: URL, clientID: String) {
+        let url = sharedContainerURL.appendingPathComponent(clientID)
+        if !FileManager.default.fileExists(atPath: url.path) {
+            let created = FileManager.default.createFile(atPath: url.path, contents: Data())
```
johnxnguyen, line 32: question: I'm confused... the file to lock is this one or the one in AccountStore?

WireDomain/Sources/WireDomain/Notifications/Components/VerifyUserStep.swift
netbe: it's to avoid confusion, which PushChannel we talked about? NSE or MainApp?

WireDomain/Sources/WireDomain/Account/AccountManager.swift
netbe: it just pass through it.
yes but we have to distinguish between main websocket and nse websocket.

netbe: ok so actually this is not used. the PushaCahnnelState given the sharedContainerURL constructs the filelock location

WireDomain/Sources/WireDomain/Account/AccountStore.swift
```diff
@@ -154,6 +154,15 @@ struct AccountStore {
         }
     }
 
+    func mainWebsocketLockFileURL(accountID: UUID) throws -> URL {
```
netbe: todo: to remove

WireDomain/Sources/WireDomain/Synchronization/PushChannelState.swift
netbe: ah maybe I started with one approach and finished with the other.

So actually i'll remove the code in AccountManager

WireDomain/Sources/WireDomain/Notifications/Components/VerifyUserStep.swift
netbe: to me it's important to distinguish which pushChannel we're talking about: the NSE or the MainApp

WireDomain/Sources/WireDomain/Account/AccountStore.swift
netbe: done https://github.com/wireapp/wire-ios/pull/3379/commits/e445fae23796acf097debd88dabdd568d652add3

WireDomain/Sources/WireDomain/Synchronization/PushChannelState.swift
netbe: https://github.com/wireapp/wire-ios/pull/3379/commits/e445fae23796acf097debd88dabdd568d652add3

johnxnguyen (approved):

WireDomain/Sources/WireDomain/Notifications/Components/VerifyUserStep.swift
jullianm: no you're right so I think we can keep it here, probably we could rename the class `VerifyStep` because it's no more only related to verifying user

jullianm (approved):
looks good

WireDomain/Sources/WireDomain/Notifications/Components/VerifyUserStep.swift
netbe: i'll change in a next PR

PR #3398, opened 2025-07-28
fix: crash in WireLinkPreview - WPB-19076

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19076" title="WPB-19076" target="_blank">BugWPB-19076</a>  [iOS] Crash on WireLinkPreview
  </td></table>
  <br />
 

### Issue

The crash attached to the jira ticket shows a crash in PreviewDownloader accessing the `containerByTaskID` proprety. This proprety can be accessed via multiple threads. This PR addresses this.

Aslo fixed a coupled of force unwrapped crashes.

### Testing

Check sending message with link previews still work.

---

### Checklist

---


wire-ios-link-preview/WireLinkPreview/PreviewDownloader.swift
```diff
@@ -136,15 +137,19 @@ final class PreviewDownloader: NSObject, URLSessionDataDelegate, PreviewDownload
     func cancel(task: URLSessionDataTaskType) {
         // When we manually cancel the task, `urlSession(session:task:didCompleteWithError:) will be called,
         // but we do not want to call the completion handler in that case.
-        cancelledTaskIDs.insert(task.taskIdentifier)
+        var set = cancelledTaskIDs.get()
+        set.insert(task.taskIdentifier)
+        cancelledTaskIDs.set(set)
```
samwyndham: I believe that this `Atomic` pattern isn't actually thread safe because the underlying data can be mutated between the `get()` call and the `set()` call. I read this is one reason why swift does't yet have a high level `atomic` wrapper.

How about using https://developer.apple.com/documentation/os/osallocatedunfairlock instead which is similar to your `ThreadSafeValue` but has a closure API so you can read + update the value in the same call

samwyndham (commented):
I left a comment to consider first.

dmitrysimkin (approved):
LGTM  👍

wire-ios-link-preview/WireLinkPreview/PreviewDownloader.swift
netbe: sure, I gave it a try

samwyndham (approved):
Nice work!

PR #3400, opened 2025-07-28
feat: Add `WireCellsLocalAssetRepository` - WPB-18848

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18848" title="WPB-18848" target="_blank">TaskWPB-18848</a>  [iOS] Create repository that returns local URL to assets
  </td></table>
  <br />
 

### Issue

This PR introduces a `WireCellsLocalAssetRepository`. This repository is responsible for downloading and accessing files from wire cells. A `WireCellsLocalAsset` holds metadata about an asset (id, size, etc) and what its current download state is. The repository provides a way of observing these assets for a given ID. This will allow table view cells in the conversation to have live updates as things change - for example showing a download progress bar or updating if the name of a file is changed remotely.

Currently I'm not super happy about the implementation and may revisit this in a little when I come to using it in the project. However, the public APIs of `WireCellsLocalAssetRepository` and `WireCellsLocalAsset` are how I imagine they should be.

### Testing

Run automated tests

---

### Checklist


WireCells/Sources/WireCellsAPI/Model/WireCellsLocalAsset.swift
```diff
@@ -0,0 +1,106 @@
...
+
+    /// The content type of the asset as defined by the backend.
+    ///
+    /// This is a MIME type (e.g. "image/png", "application/pdf").
+
+    public let contentType: String?
+
+    /// The size of the asset in bytes as defined by the backend or `-1` if unknown.
+
+    public let size: UInt64?
```
dmitrysimkin: question: if -1 as default, does it need to be optional?

WireCells/Sources/WireCellsAPI/Model/WireCellsLocalAssetMetadata.swift
```diff
@@ -0,0 +1,50 @@
...
+        self.eTag = eTag
+        self.path = path
+        self.contentType = contentType
+        self.size = size
+        self.isDownloaded = isDownloaded
+    }
+
+    public var cacheKey: String {
+        "\(nodeID.uuidString)-\(eTag)"
+    }
```
dmitrysimkin: question: is it data layer or domain layer property?

WireCells/Sources/WireCellsAPI/Model/WireCellsNodeDTO.swift
dmitrysimkin, line 1: question: usually I'm against DTO in the name because it's not clear to which layer it belongs and what it maps form or to? 
I'd suggest some other prefix+model, e.g. StoredModel, CachedModel, NetworkModel, etc. 
Is it possible to find better name for this model?

WireCells/Sources/WireCellsAPI/Persistence/FileCache.swift
```diff
@@ -16,6 +16,12 @@
 // along with this program. If not, see http://www.gnu.org/licenses/.
 //
 
-import XCTest
+public import Foundation
 
-final class WireCellsTests: XCTestCase {}
+// sourcery: AutoMockable
+public protocol FileCache: Sendable {
```
dmitrysimkin: Question: FileCache is about persistence but lives under WireCellsAPI which is network related code and FileCache is more like related to file system? Probably better put to some other folder?

WireCells/Sources/WireCellsAPI/Persistence/WireCellsLocalAssetMetadataStore.swift
```diff
@@ -0,0 +1,28 @@
...
+//
+
+public import Foundation
+
+// sourcery: AutoMockable
+@MainActor
+public protocol WireCellsLocalAssetMetadataStore {
+
+    func assetMetadata(nodeID: UUID) throws -> WireCellsLocalAssetMetadata?
+    func upsertAssetMetadata(_ metadata: WireCellsLocalAssetMetadata) throws
```
dmitrysimkin: question: honestly never heard of `upsert`, what it actually doing?

WireCells/Sources/WireCellsImplementation/WireCellsLocalAssetRepository.swift
```diff
@@ -0,0 +1,252 @@
...
+            try await fileCache.saveFile(at: tempURL, key: metadata.cacheKey)
+
+            if try metadataStore.assetMetadata(nodeID: nodeID)?.eTag == eTag {
+                // downloaded file is up-to-date so update the metadata
+                metadata.isDownloaded = true
+                try metadataStore.upsertAssetMetadata(metadata)
+            } else {
+                // remove the out-of-date file
+                try await fileCache.deleteFile(forKey: metadata.cacheKey)
+            }
```
dmitrysimkin: question: To me it looks like more of a use case not a repo, with business logic with states and api, file cache and metadata store as individual repositories, wdyt @samwyndham ?

WireCells/Sources/WireCellsAPI/Model/WireCellsLocalAsset.swift
samwyndham: Oh the comment is wrong here. When it gets stored in core data it will be -1. In the domain it will be optional. I will fix it. Thanks!

WireCells/Sources/WireCellsAPI/Model/WireCellsLocalAssetMetadata.swift
samwyndham: I guess that whole object should just be data layer. I will move it

WireCells/Sources/WireCellsAPI/Model/WireCellsNodeDTO.swift
samwyndham: Agreed. This is still left over from the original implementation and I haven't got around to changing it yet. It should be the `NetworkModel` I believe. I'll fix it. Eventually it will move somewhere else.

WireCells/Sources/WireCellsAPI/Persistence/FileCache.swift
samwyndham: FYI `WireCellsAPI` is not network related code - it's currently a big mix. It used to be a way of the different layers talking to each other. Julian is working on moving existing code from WireCells to WireMessaging after which we will follow our clean architecture approach. I will will move this to `WireMessagingData` once his PR is done.

WireCells/Sources/WireCellsAPI/Persistence/WireCellsLocalAssetMetadataStore.swift
samwyndham: Upsert means insert or update... I believe it is common when working with DBs. So if an entity exists with the same id it updates it and if not it inserts it (creates it).

WireCells/Sources/WireCellsImplementation/WireCellsLocalAssetRepository.swift
samwyndham: So making the repositories simple right and moving the logic to the use case. I will have a think about it. It might well make more sense.

WireCells/Sources/WireCellsAPI/Model/WireCellsLocalAsset.swift
```diff
@@ -0,0 +1,106 @@
...
+
+    /// The download state of the asset.
+
+    public let downloadState: DownloadState
+
+    public init(
```
caldrian: Question: will a public init be needed?

WireCells/Sources/WireCellsAPI/Model/WireCellsLocalAsset.swift
```diff
@@ -0,0 +1,106 @@
...
+
+        /// The asset has been downloaded and is available at the given cache key.
+
+        case downloaded(cacheKey: String)
+
+        /// The asset download failed with an error.
+
+        case failed(error: any Error)
+
+        public static func == (lhs: DownloadState, rhs: DownloadState) -> Bool {
```
caldrian: question: Do we really need `Equatable` conformance?
If we want to have it, I'm not sure that the current implementation, ignoring the actual error type, is behaving like any user would expect.

optional: One alternative I could think of is offering properties, like `isPending: Bool` etc.
But there are other possibilities, depending on what the goal actually is.

WireCells/Sources/WireCellsAPI/Persistence/WireCellsLocalAssetMetadataStore.swift
```diff
@@ -0,0 +1,28 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+public import Foundation
+
+// sourcery: AutoMockable
+@MainActor
```
caldrian: question: is it really needed to restrict every conforming type to be isolated to the `@MainActor`?
What about moving the attribute to the funcs?

WireCells/Sources/WireCellsImplementation/URLSession+FileDownloading.swift
```diff
@@ -0,0 +1,40 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+public import WireCellsAPI
+
+extension URLSession: FileDownloading {
```
caldrian: Personal opinion: Declaring protocols and making third party (or SDK) types conform to one's own protocols is an anti-pattern for me. As usual I'm having difficulties phrasing my thoughts why it feels wrong.

But one example what I don't like is that from now on every target which links WireCells suddenly offers a `URLSession.download(from:)` method.
Depending on whether we want this code to be a general helper for any other part of the codebase, or whether this is something WireCells specific, we should either move this into WireUtilities/WireFoundation or prevent the namespace of `URLSession` from being spoiled with this API.

I suggest to make a simple structure wrapping an URLSession and conforming to the `FileDownloading` protocol.

WireCells/Sources/WireCellsImplementation/URLSessionTaskProgressDelegate.swift
```diff
@@ -0,0 +1,39 @@
...
+
+    private let progress: @Sendable (Double) -> Void
+    private var cancellables: Set<AnyCancellable> = []
+
+    init(progress: @Sendable @escaping (Double) -> Void) {
+        self.progress = progress
+    }
+
+    func urlSession(_ session: URLSession, didCreateTask task: URLSessionTask) {
+        precondition(cancellables.isEmpty, "Delegate must not be reused across multiple tasks.")
```
caldrian: question: does it mean if one wants to download two files the following will have to be done in a strictly sequential order?
- set URLSession delegate to a new instance of `URLSessionTaskProgressDelegate`
- start a task and wait for it to be finished
- replace the delegate of the URLSession
- start another task

WireCells/Sources/WireCellsImplementation/WireCellsLocalAssetRepository.swift
```diff
@@ -0,0 +1,252 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Combine
+import Foundation
+import WireCellsAPI
+
+/// Repository for accessing & updating `WireCellsLocalAsset`s.
```
caldrian:
```suggestion
/// Repository for accessing & updating `WireCellsLocalAssets.
```
Since you're the native speaker feel free to ignore this.

caldrian (approved):

WireCells/Sources/WireCellsImplementation/URLSessionTaskProgressDelegate.swift
samwyndham: I will add some documentation but this delegate object is intended only to be passed into the async URLSession methods that accept a `URLSessionTaskDelegate` - e.g `download(from:delegate:)`. It is not intended to be set on the URLSession. It is my understanding that these delegates are scoped to that particular method call. So we can still perform downloads in parallel.

WireCells/Sources/WireCellsAPI/Persistence/FileCache.swift
samwyndham: Sorry I was wrong. This protocol now lives in `WireMessagingDomain` which I think is right.

WireCells/Sources/WireCellsAPI/Model/WireCellsLocalAsset.swift
samwyndham: Fixed.

WireCells/Sources/WireCellsAPI/Model/WireCellsLocalAssetMetadata.swift
samwyndham: Moved to the data layer.

WireCells/Sources/WireCellsAPI/Model/WireCellsNodeDTO.swift
samwyndham: Renamed with `NetworkModel` suffix

WireCells/Sources/WireCellsImplementation/WireCellsLocalAssetRepository.swift
samwyndham: The "`" is not an apostrophe but the closes the markdown section.

WireCells/Sources/WireCellsImplementation/WireCellsLocalAssetRepository.swift
samwyndham: OK the issue here is that this code in the repository needs to be tied to the lifetime of a repository which will be long lived (1 per logged in user) as downloads should continue when you leave a particular screen and come back

My assumption is that use cases should be short lived - e.g. Instantiated for a particular screen. If this code moves to for example the `WireCellsObserveDownloadStateUseCase`, that use case will have to exist for the lifetime of the user session which feels wrong. What do you think @dmitrysimkin?

dmitrysimkin: Use cases are for business logic, if it's business logic to live as long as user session then it can be. But feel free to decide @samwyndham

dmitrysimkin (approved):

WireCells/Sources/WireCellsAPI/Model/WireCellsLocalAsset.swift
samwyndham: `package` is enough. I've updated it.

WireCells/Sources/WireCellsAPI/Model/WireCellsLocalAsset.swift
samwyndham: Currently the goal is purely for testing although I think it is very useful elsewhere. Lets discuss further when you are back. It is a common problem I'm not sure the best way to solve.

WireCells/Sources/WireCellsAPI/Persistence/WireCellsLocalAssetMetadataStore.swift
samwyndham: I will try to move to functions in a follow up PR.

WireCells/Sources/WireCellsImplementation/URLSession+FileDownloading.swift
samwyndham: 👍 I will change this in a follow up PR

PR #3410, opened 2025-07-30
feat: new conversation view for chat bubbles - WPB-19034

WilhelmOks:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19034" title="WPB-19034" target="_blank">StoryWPB-19034</a>  [iOS] Set up new conversation content view in WireMessaging
  </td></table>
  <br />
 

### Issue

Added a new developer feature flag "chatBubbles".
Added a new empty ConversationMessagesViewController which is intended to contain messages with chat bubbles later.
If the chatBubbles feature flag is on, the new ConversationMessagesViewController is used instead of the old ConversationContentViewController.
 
### Testing

The chatBubbles feature flag can be enabled via development tools by shaking the device.
If the chatBubbles flag is enabled, the new conversation UI should appear.
If it is disabled, the old UI should appear and look and function exactly like it was before.


wire-ios-utilities/Source/DeveloperFlag.swift
```diff
@@ -122,6 +126,8 @@ public enum DeveloperFlag: String, CaseIterable {
             "IgnoreIncomingEventsEnabled"
         case .useWireAuthentication:
             "WireAuthenticationEnabled"
+        case .chatBubbles:
+            "ChatBubblesEnabled"
```
johnxnguyen: suggestion: this isn't needed unless we plan to control the flag via a build configuration (which for now we don't), so you can remove it.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/ConversationRootViewController.swift
```diff
@@ -52,24 +59,28 @@ final class ConversationRootViewController: UIViewController {
...
+                classificationProvider: ZMUserSession.shared(),
+                networkStatusObservable: NetworkStatus.shared,
+                getParticipantImageSourceUseCase: GetParticipantImageSourceUseCase(
+                    repository: GetParticipantImageSourceRepository(
+                        userSession: userSession
+                    )
+                ),
+                wireCellsFactory: wireCellsFactory
+            )
+        }
```
johnxnguyen: question: the `ConversationViewController` encompasses the entire conversation screen including the input bar at the bottom. `ConversationContentViewController` on the other hand encompasses only the table view containing the messages. Did you mean to replace this latter view controller? If so, you would do so [here](https://github.com/wireapp/wire-ios/blob/91b0b00d7e30b572dbd04c32de9a2c2bec0f62eb/wire-ios/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.swift#L80).

WilhelmOks: Ah, that sounds like a much better idea. I'll try that.

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationView.swift
```diff
@@ -0,0 +1,43 @@
...
+    }
+}
+
+public extension ConversationView {
+    var viewController: UIViewController {
+        let viewController = UIHostingController(rootView: Self())
+        viewController.view.backgroundColor = .clear
+        return viewController
+    }
+}
```
dmitrysimkin: question: are we making new conversation screen in SwiftUI? I though until we decided otherwise new conversation screen is in UIKit using collection view + diffable datasource? cc @johnxnguyen @netbe

johnxnguyen (approved):
Looks good 👍🏻

dmitrysimkin (approved):

PR #3430, opened 2025-08-06
feat: Chat bubbles datasource - WPB-19058

dmitrysimkin:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19058" title="WPB-19058" target="_blank">StoryWPB-19058</a>  [iOS] Setup ConversationView new datasource
  </td></table>
  <br />
 

### Issue

Created a skeleton for new conversation screen datasource, for now with mock data.

Key points:
- Created `ConversationMessagesViewModel` as MainActor to serve the view on main thread
- Created `ConversationMessagesDataSource` as actor so it automatically work on background thread as we need all messages calculations be offloaded from main thread
- `ConversationMessagesDataSource` emits updates with `NSDiffableDataSourceSnapshot` in order to do as much as possible on background thread and give UI ready snapshot to just apply
- Added initial models `MessageModel` and `UserModel` in domain layer
- Created `ChatBubbles` preview app for easier development

### Testing

Run ChatBubbles preview app (new scheme created)

### Checklist

---


WireMessaging/Sources/WireMessagingDomain/Conversation/MessageModel.swift
```diff
@@ -0,0 +1,47 @@
...
+
+public struct MessageModel: Sendable {
+
+    public enum Kind: Sendable {
+        case system(SystemMessageModel)
+        case image(ImageMessageModel)
+        case text(TextMessageModel)
+    }
+
+    public let sender: UserModel
```
samwyndham, line 27: question: Do all messages have a sender? For example a system message?

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
```diff
@@ -0,0 +1,194 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+public import UIKit
+import WireMessagingDomain
+
+public enum MessagesSection: Sendable {
+    // one section for now, later we'd have probably one section for a day
```
samwyndham, line 24: thought: I like the idea that eventually sections will be per day. I was wondering how to do this in a lazy efficient way using NSFetchedResultsController. We can't just group by `serverTimestamp`. I wonder it this could be done with a transient property 🤔. Anyway this is just me thinking and is out of scope for now

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
```diff
@@ -0,0 +1,194 @@
...
+
+/// Actor to synchronise access to all that needed to conversation screen
+/// Does all calculations in background
+public actor ConversationMessagesDataSource: @preconcurrency ConversationMessagesDataSourceProtocol {
+
+    // AsyncStream because Combine's AnyPublisher is not Sendable
+    private var updatesStreamContinuation: AsyncStream<MessagesUpdate>.Continuation?
+    public func updatesStream() async -> AsyncStream<MessagesUpdate> {
+        AsyncStream { continuation in
+            self.updatesStreamContinuation = continuation
```
samwyndham, line 43: suggestion: Does this creates a strong reference cycle? Anyway there is now a better way (in my opinion) of creating streams without closures:

```swift
let (stream, continuation) = AsyncStream.makeStream(of:MessagesUpdate.self)
updatesStreamContinuation = continuation
return stream
```

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
```diff
@@ -0,0 +1,194 @@
...
+public protocol ConversationMessagesDataSourceProtocol: Sendable {
+    func updatesStream() async -> AsyncStream<MessagesUpdate>
+    func loadInitialMessages() async
+}
+
+/// Actor to synchronise access to all that needed to conversation screen
+/// Does all calculations in background
+public actor ConversationMessagesDataSource: @preconcurrency ConversationMessagesDataSourceProtocol {
+
+    // AsyncStream because Combine's AnyPublisher is not Sendable
```
samwyndham, line 39: thought: One thing to be careful of with AsyncStream is that it each update can only be read once - this is fine if there is only a single consumer of updatesStream() but if there are multiple they will miss updates. Personally I think it might be better to use a publisher internally. I believe these can easily be converted to streams via the publishers `values` property. However behavior here needs to be confirmed.

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
```diff
@@ -0,0 +1,194 @@
...
+public typealias MessagesSnapshot = NSDiffableDataSourceSnapshot<MessagesSection, MessageType>
+
+public protocol ConversationMessagesDataSourceProtocol: Sendable {
+    func updatesStream() async -> AsyncStream<MessagesUpdate>
+    func loadInitialMessages() async
+}
+
+/// Actor to synchronise access to all that needed to conversation screen
+/// Does all calculations in background
+public actor ConversationMessagesDataSource: @preconcurrency ConversationMessagesDataSourceProtocol {
```
samwyndham, line 37: praise: actor is a good choice.

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
```diff
@@ -0,0 +1,194 @@
...
+        forceRecalculate: Bool = false,
+    ) async -> IndexPath {
+        .init()
+    }
+
+    // load older messages
+    func loadOlderMessages() {}
+
+    // load newer messages
+    func loadNewerMessages() {}
```
samwyndham, line 87: thought: I'm  wondering if _paging_ is the right way to go. A properly configured NSFetchedResultsController doesn't need it and will do things lazily and efficiently. One reason where paging can be awkward can be if we want to jump around in history - e.g. a jump to start of conversation feature. This is of course doable but might be a bit more complex to implement. NSFetchedResultsController also does not require diffing but still emits `NSDiffableDataSourceSnapshot`

These two approaches will be require quite a different API though. The NSFetchedResultsController approach would be more similar to a classic UITableViewDataSource.

WireMessaging/Sources/WireMessagingUI/Conversation/MessageCollectionViewCell.swift
```diff
@@ -0,0 +1,40 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import SwiftUI
+import UIKit
+
+class MessageCollectionViewCell: UICollectionViewCell {
```
samwyndham, line 22: suggestion: No reason to subclass. Use a regular UICollectionViewCell and configure it withe help methods. That way we also don't need the casting after dequeuing the cell.

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesViewController.swift
```diff
@@ -17,21 +17,122 @@
...
+    }
+
+    private func setupDataSource() {
+        dataSource =
+            DataSource(collectionView: collectionView) { collectionView, indexPath, message -> UICollectionViewCell? in
+                guard let cell = collectionView.dequeueReusableCell(
+                    withReuseIdentifier: MessageCollectionViewCell.reuseIdentifier,
+                    for: indexPath
+                ) as? MessageCollectionViewCell else {
+                    return UICollectionViewCell()
```
samwyndham, line 118: suggestion: Crashing is preferable if we get a random cell type as we can't do anything with it.

samwyndham (commented):
Nice work @dmitrysimkin! Right now I've added a few comments to think about.

WireMessaging/Sources/WireMessagingDomain/Conversation/MessageModel.swift
dmitrysimkin: True, it's optional in ZMMessage, will update in next PR

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
dmitrysimkin: I'll check, tnx for pointing it out. and I'll improve without closure

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
dmitrysimkin: I prefer publisher as well but it does not have `Sendable` conformance out of the box. I'll monitor its behaviour while i'd be working on it in next PRs and change/fix is there would be issues. Thanks for the concern

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
dmitrysimkin: yeah Interesting to think about it. i just made same interface as we have in old datasource. Let me try find more effecient way working with NSFRC at later stages of implementation

WireMessaging/Sources/WireMessagingDomain/Conversation/MessageModel.swift
OlivellaO: thought: We could consider that even system messages have a sender (the system), but i guess if we are using UserModel as a sender that's not possible (and out of scope). It will make MessageModel a bit more standard and reusable tho. Anyway, i was just a thought that cross my mind when reviewing this.

johnxnguyen: I think the model here shouldn't be concerned with messages, but rather cells. Cells are mostly for messages, but they can also be for other things like time dividers and even a loading cell at the top of the history. Perhaps we can do something like:

```swift
enum CellModel {

  case systemMessage(SystemMessageModel)
  case error(ErrorModel)
  case userMessage(UserMessageModel)
  case timeDivider(TimeDividerModel)
  ...

}

```

WireMessaging/Sources/WireMessagingUI/Conversation/Message Views/TextMessageView/TextMessageViewModel.swift
```diff
@@ -0,0 +1,46 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import SwiftUI
+public import Foundation
+
+public class TextMessageViewModel: ObservableObject, Identifiable, Hashable, @unchecked Sendable {
```
johnxnguyen, line 22: question: is `Identifiable` and `Hashable` necessary for diffable datasource?

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
johnxnguyen: Or perhaps when we set the timestamp we can compute its day and store to make it easier for the fetched results controller to group.

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
```diff
@@ -0,0 +1,194 @@
...
+
+public enum MessagesSection: Sendable {
+    // one section for now, later we'd have probably one section for a day
+    case main
+}
+
+public typealias MessagesSnapshot = NSDiffableDataSourceSnapshot<MessagesSection, MessageType>
+
+public protocol ConversationMessagesDataSourceProtocol: Sendable {
+    func updatesStream() async -> AsyncStream<MessagesUpdate>
```
johnxnguyen, line 31: suggestion: `streamMessageUpdates`

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
johnxnguyen: @samwyndham are you suggesting to let the fetched results controller load all messages of the conversation? How does it do that efficiently if we only really will look at the last 50 unless we scroll back in time?

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesViewModel.swift
```diff
@@ -0,0 +1,48 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+@MainActor
+public protocol ConversationMessagesViewModelProtocol {
+    func updatesStream() async -> AsyncStream<MessagesUpdate>
```
johnxnguyen, line 23: suggestion: `streamMessageUpdates`

WireMessaging/Sources/WireMessagingUI/Conversation/MessageUpdateType.swift
```diff
@@ -0,0 +1,30 @@
...
+import Foundation
+
+public enum MessagesUpdate: Sendable {
+
+    case initiallyLoaded(MessagesSnapshot)
+    case messageAdded(MessagesSnapshot)
+    // later to be added more updates like:
+    // loaded new messages, new or older
+    // re-sent failed message
+    // to be added other updates that happens to a conversation view
```
johnxnguyen, line 28: question: can you elaborate a bit more on this? For example, how was a `failedMessage` update be handled?

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesViewController.swift
```diff
@@ -17,21 +17,122 @@
...
+    private func observeUpdates() async {
+        let stream = await viewModel.updatesStream()
+        for await update in stream {
+            switch update {
+            case let .initiallyLoaded(snapshot):
+                await dataSource.apply(snapshot)
+            case let .messageAdded(snapshot):
+                await dataSource.apply(snapshot)
+            }
+        }
```
johnxnguyen, line 66: question: how would you envision this working with a SwiftUI list? Just wondering when the time comes, what changes we would make to connect the datasource.

WireMessaging/Sources/WireMessagingUI/Conversation/MessageCollectionViewCell.swift
dmitrysimkin: Got rid of MessageCollectionViewCell in https://github.com/wireapp/wire-ios/pull/3439

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesViewController.swift
dmitrysimkin: No need to crash after using `UICollectionviewCell`

WireMessaging/Sources/WireMessagingUI/Conversation/Message Views/TextMessageView/TextMessageViewModel.swift
dmitrysimkin: `Hashable` constraint was added by `UICollectioViewDiffableDataSource`, but Identifiable looks like an artefact from my POC tries, I'll remove it

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
dmitrysimkin: why stream at the beginning? usually it's prefix like `...Publisher` or `...Subject` and it's not singe message updates but plural messages updates. it feels to me still `updatesStream` might be better name, wdyt @johnxnguyen ?

WireMessaging/Sources/WireMessagingUI/Conversation/MessageUpdateType.swift
dmitrysimkin: Similar to message added, if failed message resent successfully it will update internal data in datasource and will emit new snapshot, where new message will be moved from it's original place to the end of the conversation

For now I added different cases and they all (2) have just snapshot, so probably specific cases might not be need in the future but I'd think in UI we'd might need to introduce some side effect based on specific update case

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesViewController.swift
dmitrysimkin: We already have list of view models inside snapshot and in case of SwiftUI we'd just use list of view models. Or we can map steam to some other @Published property of a state or smth like this

WireMessaging/Sources/WireMessagingDomain/Conversation/MessageModel.swift
dmitrysimkin: In Domain it should be about business models. For example if in real work for a generic message can have no sender it should be optional. But for a presentation layer for a specific message type such as text message for instance it should be a sender, so in UI layer it would non-optional

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
dmitrysimkin: It was indeed a retain cycle, fixed in https://github.com/wireapp/wire-ios/pull/3439

WireMessaging/Sources/WireMessagingUI/Conversation/Message Views/TextMessageView/TextMessageViewModel.swift
dmitrysimkin: Removed Identifiable in https://github.com/wireapp/wire-ios/pull/3439

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
samwyndham: I was thinking that too but it would be problematic when changing times zones 😞

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
samwyndham: Yeah one option is just to import Combine with the pre-concurrency flag I guess which is annoying. 

Anyway it will be fine as long as there is only one subscriber which I expect there is. Consider adding a warning to the doc comments to say there should only be one subscriber

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
samwyndham: I thought about this some more. I was very attached to my NSFetchedResultsController approach because it is what I'm used and it is how I imagined solving this problem but perhaps there is no real reason to use it.

> @samwyndham are you suggesting to let the fetched results controller load all messages of the conversation? How does it do that efficiently if we only really will look at the last 50 unless we scroll back in time?

Yes I believe so because it operates in a lazy fashion. It knows how many items there are without loading them - it just needs to do a count which SQLite is efficient at. So something like `controller.fetchedObjects?.count` is cheap even when there are many many items. Only when you do `controller.fetchedObjects?.first.name` will the data actually be fetched from the DB. This is what I have understood in the past. I haven't tried profiling it at all but is how I would expect things to work.

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
dmitrysimkin: Added comment in next PR

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
dmitrysimkin: If we don't use paging and use just all `fetchedObjects` with like >1000 messages e.g., maybe they will be loaded lazily from DB but we need map DB objects to view models, and we don't want to map all objects to UI model, right?   @samwyndham how didi you imagine to also lazily map models?

samwyndham: Yeah, it could be tricky to use NSFetchedResultController + differable data source. I think there are ways to achieve it such as mapping the snapshot to not contain the model but only an id then having a method that fetches the mapped model from the data source by id.

But I think we should go with your approach.

WireMessaging/Sources/WireMessagingDomain/Conversation/MessageModel.swift
samwyndham: How about `ConversationElement`?

samwyndham (approved):
Nice work Dmitry

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
johnxnguyen: I think naming functions as verbs rather than nouns is a good practice because it describes what the function does rather than what it produces. In this case "stream message updates" signifies that the datasource will begin to stream updates which the caller can the iterate over.

Just a personal preference and not a blocker.

WireMessaging/Sources/WireMessagingDomain/Conversation/MessageModel.swift
johnxnguyen: I like this

johnxnguyen (approved):

WireMessaging/Sources/WireMessagingUI/Conversation/ConversationMessagesDataSource.swift
dmitrysimkin: I'll update it to `makeUpdatesStream` in next PR

WireMessaging/Sources/WireMessagingDomain/Conversation/MessageModel.swift
dmitrysimkin: I'll rename `MessageType` to `ConversationElement` which is same level model (UI) and leave domain layer as still `UserModel`, `DomainModel` as for example in domain there is no such thing as time divider, it's only UI thing

PR #3435, opened 2025-08-07
chore: personal-account-lifecycle test - WPB-19192

findms:
### Issue

_Please describe the issue._

### Testing

image

_Describe how to test._

---

### Checklist

---


dmitrysimkin (approved):

WireNetwork/Sources/WireNetwork/APIs/SelfUserAPI/SelfUserAPIV0.swift
```diff
@@ -86,17 +86,18 @@ class SelfUserAPIV0: SelfUserAPI, VersionedAPI {
             .parse(code: response.statusCode, data: data)
     }
 
-    func updateHandle(handle: String) async throws {
+    func updateHandle(accessToken access_token: String, handle: String) async throws {
```
johnxnguyen: question: could you explain this change? My guess is that you want to inject an access token, but why not mock it from the authentication manager?

findms: @johnxnguyen While creating 2 users , for 2nd user updating handle uses the previous user token as active and doesn't generate new token which leads to update handle for 2nd user, it updates the 1st one and creating a mismatch with different handle for different name. By injecting the current user access_token explicitly use the valid cookie/token to update handle. Not sure how we can invalidate the active token or mock it while executing request

findms: Tried removing old user cookie and store new cookie and getting stored as well but accesstoken with user1 is valid it simply doesn't go that way to create new for user2

johnxnguyen (approved):

PR #3438, opened 2025-08-07
feat: chat bubble wrapper for text messages - WPB-19273

WilhelmOks:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19273" title="WPB-19273" target="_blank">TaskWPB-19273</a>  [iOS] Conversation Message Content embedding into Simple Chat Bubbles
  </td></table>
  <br />
 

### Issue

Text Message cells are wrapped into bubbles.

The color of the bubble depends on if it's my own message or from someone else and also depends on the color selected by the user as their accent color.

The corner radius of all cell types (except small icons like reactions) like reply and file has been adjusted to be the same as the new chat bubble corner radius.

The correct alignment will be implemented in WPB-19271 and is not in scope of this PR.

### Testing

If the chatBubblesSimple DeveloperFlag is disabled, everything should look and behave exactly as it was before.

If the chatBubblesSimple DeveloperFlag is enabled:

* text messages should be inside of a chat bubble
  * the color of the chat bubble from other users should be gray
  * the color of the chat bubble from the self user should be their selected accent color
* dark and light theme should look like in the design
* the corner radius of all content cells like image and reply should be the same radius as the chat bubble


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationFileMessageCell.swift
```diff
@@ -55,9 +55,15 @@ final class ConversationFileMessageCell: UIView, ConversationMessageCell {
     }
 
     private func configureSubview() {
-        containerView.shape = .rounded(radius: 12)
+        let cornerRadius: CGFloat = if DeveloperFlag.chatBubblesSimple.isOn {
+            ConversationMessageContainerView.bubbleCornerRadius
+        } else {
+            12
+        }
```
dmitrysimkin, line 62: suggestion: Since it's same in several files I'd suggest to introduce a constant and reuse `if DeveloperFlag.chatBubblesSimple.isOn ...`

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationTextMessageCell.swift
```diff
@@ -76,16 +88,35 @@ final class ConversationTextMessageCell: UIView, ConversationMessageCell, TextVi
...
+            insets = ConversationMessageContainerView.bubbleEdgeInsets
+        } else {
+            let margins = conversationHorizontalMargins
+            insets = UIEdgeInsets(top: 0, left: margins.left, bottom: 0, right: margins.right)
+        }
         messageTextView.fitIn(view: self, insets: insets)
     }
 
+    private func configureTextColor(forOwnMessage ownMessage: Bool) {
+        if DeveloperFlag.chatBubblesSimple.isOn {
```
dmitrysimkin: nit-pick: use guard

dmitrysimkin (commented):
Looks good, nice work 👍 
Before approving, can you also please some snapshot tests @WilhelmOks ?

WilhelmOks:
> Looks good, nice work 👍 Before approving, can you also please some snapshot tests @WilhelmOks ?

Assuming this is a test which compares the visual output to an image from the design:
I think our designs for the simple solution are not nearly exact enough to make a pixel perfect comparison.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationFileMessageCell.swift
WilhelmOks: Not sure what you mean by reuse the if condition.

Do you mean extracting it into a property like this?

```
var: cornerRadius: CGFloat {
        let cornerRadius: CGFloat = if DeveloperFlag.chatBubblesSimple.isOn {
            ConversationMessageContainerView.bubbleCornerRadius
        } else {
            12
        }
}
```

but then it would still be repeated in multiple files.

dmitrysimkin:
> > Looks good, nice work 👍 Before approving, can you also please some snapshot tests @WilhelmOks ?
> 
> Assuming this is a test which compares the visual output to an image from the design: I think our designs for the simple solution are not nearly exact enough to make a pixel perfect comparison.

What do you mean by it's a test? As I understand it supposed to be released, right? Tests help make sure that while working on other parts of application already existed functionality behaves in same way. Not critical for me, we just usually add snapshot tests to views we add. cc @johnxnguyen

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationFileMessageCell.swift
dmitrysimkin: I mean more like this
```
struct ChatBubblesConstants {
    var cornerRadius: CGFloat {
        if DeveloperFlag.chatBubblesSimple.isOn {
            return ConversationMessageContainerView.bubbleCornerRadius
        } else {
            return 12
        }
    }
}
```

And then on caller side: `containerView.layer.cornerRadius = ChatBubblesConstants.cornerRadius`
Do you think there would be a benefit?

WilhelmOks:
> > > Looks good, nice work 👍 Before approving, can you also please some snapshot tests @WilhelmOks ?
> > 
> > 
> > Assuming this is a test which compares the visual output to an image from the design: I think our designs for the simple solution are not nearly exact enough to make a pixel perfect comparison.
> 
> What do you mean by it's a test? As I understand it supposed to be released, right? Tests help make sure that while working on other parts of application already existed functionality behaves in same way. Not critical for me, we just usually add snapshot tests to views we add. cc @johnxnguyen

Hm, probably I misunderstood what is a snapshot test.
Can you provide to me a quick explanation or a guide how such a snapshot test would be implemented in the Wire project?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationTextMessageCell.swift
```diff
@@ -49,9 +50,20 @@ final class ConversationTextMessageCell: UIView, ConversationMessageCell, TextVi
...
-    weak var message: ZMConversationMessage?
+    weak var message: ZMConversationMessage? {
+        didSet {
+            guard let message, DeveloperFlag.chatBubblesSimple.isOn else { return }
+            let isOwnMessage = message.isSentBySelfUser
+            let userColor = message.senderUser?.accentColor ?? .clear
+            container?.bubbleStyle = isOwnMessage ? .ownMessage(userColor: userColor) : .otherMessage
+            configureTextColor(forOwnMessage: isOwnMessage)
+        }
+    }
```
johnxnguyen, line 65: suggestion: this seems to be more appropriate placed inside the  `func configure(with object: Configuration, animated: Bool)` method, but I then you'd need to add an extra field to the configuration for the bubble style. What do you think?

johnxnguyen:
> Hm, probably I misunderstood what is a snapshot test. Can you provide to me a quick explanation or a guide how such a snapshot test would be implemented in the Wire project?

@WilhelmOks snapshot tests kind of what you think, but the implemented views aren't compared to the designed view, rather you implement the view according to the design, snapshot it (record a reference image from your view), then future runs of the test will assert that the view still matches the reference. 

Take a look at `ConversationTextMessageTests.swift`, basically you create your view, run the verify method, it if it doesn't find a reference already it will record one, otherwise it will compare the view with the reference.

johnxnguyen (commented):
Looks good! Left a suggestion/question.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationTextMessageCell.swift
WilhelmOks: My first attempt was to set both values, `isBubble` and `bubbleStyle`, in `configure(with object: Configuration, animated: Bool)`  but the problem is that for `bubbleStyle` I need the info if this is the self user message. And `message` is nil at this point.
So I moved the code to didSet of the `message` property, because there I can be certain that the `message` has been set and is not nil.

wire-ios/Wire-iOS Tests/ConversationMessageCell/SimpleChatBubblesSnapshotTests.swift
```diff
@@ -0,0 +1,74 @@
...
+//
+
+import GenericMessageProtocol
+import XCTest
+
+@testable import Wire
+
+final class SimpleChatBubblesSnapshotTests: ConversationMessageSnapshotTestCase {
+    
+    private let record: Bool? = nil
```
dmitrysimkin, line 26: might be removed if not used anymore

dmitrysimkin (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationFileMessageCell.swift
WilhelmOks: I see some issues with that:
* the old value for the corner radius isn't always 12. For some views like the reply cell it's smaller.
* ChatBubblesConstants for me suggests that it's about the new chat bubbles. So putting the old value there (12) feels wrong.
* special treatment for the corner radius while the other numeric values remain number literals.

wire-ios/Wire-iOS Tests/ConversationMessageCell/SimpleChatBubblesSnapshotTests.swift
WilhelmOks: The record constant?
I feel like it's useful for later when we change the bubble layout and need to generate new snapshots for all tests again in one go.
But I have no strong opinion on that.

johnxnguyen (approved):

PR #3441, opened 2025-08-11
fix: unable to pull more than 1000 conversations - WPB-19404

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19404" title="WPB-19404" target="_blank">BugWPB-19404</a>  [iOS] Can't pull conversations if there are more than 1000
  </td></table>
  <br />
 

### Issue
If a user has more than 1000 conversations and tries to log in, then the initial sync will fail when trying to pull conversations because the endpoint only accepts between 1 and 1000 conversation ids.

To resolve this, we now validate the argument containing ids (at the api layer) and will pull perform multiple requests if needed to fetch > 1000 conversations.

### Testing
1. Have more than 1000 conversations
2. Log in fresh
3. Assert you can complete initial sync and see your conversations in the list.

---

### Checklist

---

### DIFF
--- a/WireDomain/Sources/WireDomain/Synchronization/PullAllConversationsSync.swift
+++ b/WireDomain/Sources/WireDomain/Synchronization/PullAllConversationsSync.swift
@@ -59,30 +59,37 @@ public final class PullAllConversationsSync: PullAllConversationsSyncProtocol {
             }
         }
 
-        let conversations = try await api.getConversations(for: conversationIDs)
+        while !conversationIDs.isEmpty {
+            // Fetch max 1000 at a time.
+            let idsToFetch = Array(conversationIDs.prefix(1000))
+            conversationIDs.removeFirst(idsToFetch.count)
 
-        for conversation in conversations.found {
-            await store.storeConversation(
-                conversation.toDomainModel(),
-                timestamp: .now,
-                isFederationEnabled: isFederationEnabled,
-                isMLSEnabled: isMLSEnabled
-            )
-        }
+            let conversations = try await api.getConversations(for: idsToFetch)
 
-        for id in conversations.notFound {
-            await store.storeConversation(
-                needsBackendUpdate: true,
-                conversationID: id.id,
-                conversationDomain: id.domain
-            )
-        }
+            for conversation in conversations.found {
+                await store.storeConversation(
+                    conversation.toDomainModel(),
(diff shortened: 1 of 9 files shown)


WireDomain/Sources/WireDomain/Synchronization/PullAllConversationsSync.swift
```diff
@@ -59,30 +59,37 @@ public final class PullAllConversationsSync: PullAllConversationsSyncProtocol {
             }
         }
 
-        let conversations = try await api.getConversations(for: conversationIDs)
+        while !conversationIDs.isEmpty {
```
samwyndham: suggestion: Break `conversationIDs` into batches and fetch all batches simultaneously using a task group. That way all the risky network operations are done (a tiny bit) faster and there is a bit less mutating style code.

WireNetwork/Sources/WireNetwork/APIs/ConversationsAPI/ConversationsAPIV2.swift
```diff
@@ -22,6 +22,12 @@ class ConversationsAPIV2: ConversationsAPIV1 {
     override var apiVersion: APIVersion { .v2 }
 
     override func getConversations(for identifiers: [QualifiedID]) async throws -> ConversationList {
+        guard (1 ... 1000).contains(identifiers.count) else {
```
samwyndham: knowledge share: not saying you should use it but I just learnt you can do the following instead:

```swift
guard 1...1000 ~= identifiers.count else {
```

samwyndham (approved):
Nice work. Left an optional suggestion.

WireNetwork/Sources/WireNetwork/APIs/ConversationsAPI/ConversationsAPIV2.swift
johnxnguyen: done

WireDomain/Sources/WireDomain/Synchronization/PullAllConversationsSync.swift
johnxnguyen: done

jullianm (approved):
Looks good, nice work

PR #3445, opened 2025-08-12
fix: incremental sync can be slow - WPB-19298

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19298" title="WPB-19298" target="_blank">BugWPB-19298</a>  [iOS] New incremental sync is unusually slow
  </td></table>
  <br />
 

### Issue
It has been observed that the incremental sync (as visible by the sync bar) can take an unusually long time. Upon investigation I found that the sync bar is shown when the incremental sync starts, but continues to be shown after it is finished and won't be dismissed until several other potentially long tasks are completed, such as joining pending mls groups and resolving all 1-1 conversations.

It's not clear if there are issues in these extra tasks, certainly there is opportunity to optimize them. For example, if there are many 1-1s then resolving them all may take a significant amount of time, especially on a poor network. There may be error that occur which then block the the sync from completing etc. Solving these problems are out of the scope of this pull request however.

Instead, this PR adjusts the display and dismissal of the syncing bar to match directly the length of the incremental sync, i.e when all pending events are processed. This is alignment with other platforms.

Additionally I have also tried to resolve some potential concurrency issues, such as blocking the main thread and potentially blocking the sync context with long running tasks.

### Testing
1. Log in and open the app.
2. Use the app as normal, put in background, receive many events
3. Come to foreground, assert that the sync takes a reasonable amount of time.

---

### Checklist

---


wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1201,74 +1198,64 @@ extension ZMUserSession: SyncAgentDelegate {
...
 
-            showSyncBar(true)
+        Task {
+            await showSyncBar(false)
+        }
 
-            NotificationInContext(
-                name: .quickSyncCompletedNotification,
-                context: notificationContext
-            ).post()
```
johnxnguyen, line 1227: this was dead

KaterinaWire (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1050,17 +1050,14 @@ extension ZMUserSession: ZMNetworkStateDelegate {
...
 
     func updateNetworkState() {
-        let state: NetworkState = if isNetworkOnline {
-            if isPerformingSync {
-                .onlineSynchronizing
-            } else {
-                .online
-            }
-        } else {
+        networkState = switch (isNetworkOnline, isPerformingSync) {
```
samwyndham, line 1053: praise: much easier to read (for me)

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1201,74 +1198,64 @@ extension ZMUserSession: SyncAgentDelegate {
...
-                showSyncBar(false)
-            }
+            // TODO: [WPB-18175] Port MLS client creation and related MLS operations from here to the InitialSync
 
-            recurringActionService.performActionsIfNeeded()
-            performPostQuickSyncE2EIActions()
+            await recurringActionService.performActionsIfNeeded()
         }
+
+        performPostQuickSyncE2EIActions()
```
samwyndham, line 1258: sanity check: So this doesn't need to happen after the code in `WaitingGroupTask` right?

samwyndham (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: on develop it doesn't wait for the group task anyway, it was typically running before the group. I wanted to remove the group altogether but it breaks too many tests.

PR #3452, opened 2025-08-13
fix: further long sync mitigations - WPB-19298

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19298" title="WPB-19298" target="_blank">BugWPB-19298</a>  [iOS] New incremental sync is unusually slow
  </td></table>
  <br />
 

### Issue
After investigation, we found that at the end of each sync (when app comes to foreground) we repeat some possibly expensive work, some of which is pointless due to broken mls groups. For instance:
- we try to join pending mls groups (will fail for all broken groups)
- we try to resolve all 1-1 conversations
- we try to commit pending proposals

These first two are sufficient to do once per launch, there is not much benefit in retrying so frequently. (Note that 1-1 conversations are resolved on an individual basic when opening a chat).

The last one is hides a nasty issue: some of the broken mls groups have pending proposals which will always fail. But the recovery strategy is to perform another sync and retry committing the proposal. This will fail each time causing another attempt, finally giving up after 3 failures. This will happen for each broken group, so if there are 10 broken groups, we will cause 10 x 3 syncs, then retry this all again when brining the app to the foreground once more.

To solve this, we shouldn't attempt to commit pending proposals for broken groups, so if we detect a broken group, we store the id, and then use this id to exclude the group from the list of those needing to commit pending proposals. 

### Testing
1. Have broken groups
2. Perform sync
3. Assert that only one sync starts and ends in a reasonable time when coming to foreground.

---

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/Use cases/ResolveOneOnOneConversationsUseCase.swift
+++ b/wire-ios-sync-engine/Source/Use cases/ResolveOneOnOneConversationsUseCase.swift
@@ -23,7 +23,8 @@ import WireLogging
 // sourcery: AutoMockable
 public protocol ResolveOneOnOneConversationsUseCaseProtocol {
 
-    func invoke() async throws
+    @discardableResult
+    func invoke() async throws -> Bool
 
 }
 
@@ -36,9 +37,8 @@ struct ResolveOneOnOneConversationsUseCase: ResolveOneOnOneConversationsUseCaseP
     let resolver: any OneOnOneResolverInterface
     let pullSelfUserClientsFactory: PullSelfUserClientsFactory
 
-    static var didResolve = false
-
-    func invoke() async throws {
+    @discardableResult
+    func invoke() async throws -> Bool {
         let oldProtocols = await context.perform {
             let selfUser = ZMUser.selfUser(in: context)
             return selfUser.supportedProtocols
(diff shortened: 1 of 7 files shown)


KaterinaWire (approved):

wire-ios-sync-engine/Source/Use cases/ResolveOneOnOneConversationsUseCase.swift
```diff
@@ -36,6 +36,8 @@ struct ResolveOneOnOneConversationsUseCase: ResolveOneOnOneConversationsUseCaseP
     let resolver: any OneOnOneResolverInterface
     let pullSelfUserClientsFactory: PullSelfUserClientsFactory
 
+    static var didResolve = false
```
samwyndham: issue(maybe): How does this behave if we log in a new additional user? Or we log out and re-login. Invoke will exist early. Is that an issue? Or is it OK to wait for the next app run?

johnxnguyen: Good question, probably best to avoid storing it statically and instead keep it local to the scope of the user session. I'll think about how to do that.

johnxnguyen: done fb5df03e3e705ef18fae73aed16343147c519161

samwyndham (approved):

PR #3454, opened 2025-08-13
feat: add `WireCellsFetchNodesUseCase` - WPB-19447

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19447" title="WPB-19447" target="_blank">TaskWPB-19447</a>  [iOS] Create a use case for fetching wire cells nodes for a given conversation
  </td></table>
  <br />
 

### Issue

This PR adds `WireCellsFetchNodesUseCase`. FYI a _node_ is how we refer to data about a given file stored in wire cells.
This use case in configurable and will be used in multiple placed depending on its configuration. For example:

- It can fetch all nodes for all conversations for use in a _all files_ tab.
- It can fetch nodes for a single conversation or a single folder in that conversation.
- It can fetch all deleted nodes (useful for the trash can / recycle bin feature)

The particular configurations are defined in the domain target (and can only be defined there). Currently there is only a single configuration suitable for the files tab in the conversation view.

The response is paged. It returns a token that can be used to fetch the next page. If that token is nil, then there are no more pages.

### Testing

Run automated tests.

---

### Checklist


WireMessaging/Package.swift
```diff
@@ -14,7 +14,7 @@ let package = Package(
         .library(name: "WireMessagingUI", targets: ["WireMessagingUI"])
     ],
     dependencies: [
-        .package(url: "https://github.com/pydio/cells-sdk-swift.git", from: "0.1.1-alpha10"),
+        .package(url: "https://github.com/pydio/cells-sdk-swift.git", from: "0.1.1-alpha15"),
```
samwyndham, line 17: We also update the cells SDK to the latest version

WireMessaging/Sources/WireMessagingData/WireCells/NodesAPI/NodesAPI.swift
```diff
@@ -63,20 +63,6 @@ package final actor NodesAPI: NodesAPIProtocol {
         await awsClient.upload(path: path, node: node.toDTO(), versionID: versionID)
     }
 
-    package func getFiles(
```
samwyndham, line 66: This method won't be used.

WireMessaging/Sources/WireMessagingData/WireCells/NodesAPI/NodesAPI.swift
```diff
@@ -120,6 +106,13 @@ package final actor NodesAPI: NodesAPIProtocol {
         return dto.toModel()
     }
 
+    package func getNodes(
+        _ request: WireCellsGetNodesRequest
+    ) async throws -> (nodes: [WireCellsNode], nextOffset: Int?) {
+        let (nodes, nextOffset) = try await restAPI.getNodes(request)
+        return (nodes: nodes.map { $0.toModel() }, nextOffset: nextOffset)
```
dmitrysimkin: suggestion: `toModel` is not clear to which layer model it maps, I assume it maps to domain? so it could be renamed to `toDomainModel()`

dmitrysimkin (approved):

WireMessaging/Sources/WireMessagingDomain/WireCells/UseCases/WireCellsFetchNodesUseCase.swift
```diff
@@ -0,0 +1,95 @@
...
+                deletionStatus: configuration.deletionStatus,
+                text: searchTerm,
+                type: configuration.nodeType
+            ),
+            limit: configuration.pageSize,
+            offset: offset
+        )
+        let (nodes, nextOffset) = try await repository.getNodes(request)
+
+        return (nodes, nextOffset.map { WireCellsPageToken(offset: $0) })
```
jullianm, line 92: question: we have a `PayloadPager` generic struct in `WireNetwork` that we use to iterate through one or more pages returned from an api endpoint. I'm wondering if we could use a similar mechanism (we can't reuse exactly the same because it's very specific to Wire BE) here to fetch the nodes.

jullianm (commented):
Looks good, left a question about the pagination mechanism as we already have a `PayloadPager` generic struct in our code I wonder whether we could use a similar mechanism to fetch the nodes as it involves pagination too.

WireMessaging/Sources/WireMessagingDomain/WireCells/UseCases/WireCellsFetchNodesUseCase.swift
samwyndham: I guess there is no point having the iterator functionality because we would only fetch the next page when the user scrolls to near the bottom of the list. But there could be some kind of object returned like:

```swift
struct Pager<T> {
  var hasMore: Bool { //... }
  func nextPage() async throws -> [T] { //... }
}
```

samwyndham: I don't think the `AsyncSequence` protocol makes sense because there is no way of checking if there are more results without trying to load more

samwyndham: I do wonder if it is worth the abstraction though. I guess it will add extra code with limited benefit.

jullianm: makes sense, let's keep it like this

jullianm (approved):

PR #3455, opened 2025-08-13
feat(chat bubbles): reactively update sender name - WPB-19514

dmitrysimkin:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19514" title="WPB-19514" target="_blank">TaskWPB-19514</a>  [iOS] Handle reactive changes in sender name
  </td></table>
  <br />
 

### Issue

Added logic to handle when user name changed, by introducing one publisher per user (one subject and sever cell's view models as subscribers.

### Testing

https://github.com/user-attachments/assets/d91381fb-2650-4771-b4ce-4bca5ea97e2c


WireMessaging/Sources/WireMessagingAssembly/ConversationsAssembly.swift
```diff
@@ -19,18 +19,21 @@
...
 public import WireMessagingDomain
+import Combine
 
 public enum WireMessagingAssembly {
 
     @MainActor
     public static func makeConversationScreen(
-        loadMessagesRepo: any (LoadConversationMessagesRepositoryProtocol & MonitorMessagesRepositoryProtocol)
+        loadMessagesRepo: any (LoadConversationMessagesRepositoryProtocol & MonitorMessagesRepositoryProtocol),
+        senderNameObserverProvider: SenderNameObserverProvider?
```
dmitrysimkin, line 29: For passing as separate parameter but later I'd image it would be factory to get individual message publishers

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/MessagesIndividualUpdatesFactory.swift
```diff
@@ -0,0 +1,50 @@
...
+final actor MessagesIndividualUpdatesFactory {
+
+    private let context: NSManagedObjectContext
+    private var dict = [NSManagedObjectID: SenderObserver]()
+
+    init(context: NSManagedObjectContext) {
+        self.context = context
+    }
+
+    func makeSenderNameObserver(user: UserModel?) -> SenderNameObserverProtocol? {
```
samwyndham, line 32: issue: In my opinion it doesn't make sense to allow a nil `user`. The caller of `makeSenderNameObserver` should do the nil check beforehand.

WireMessaging/Sources/WireMessagingAssembly/ConversationsAssembly.swift
samwyndham, line 29: question: Why does this need to be `nil`? Having this as nil seems to propagate throughout the code.

WireMessaging/Sources/WireMessagingDomain/Conversation/UserModel.swift
```diff
@@ -18,16 +18,26 @@
 
 public import Foundation
 
+// To be refined later
 public struct UserModel: Sendable {
-    // To be refined later
+
+    // 'objectID' to abstract id from data layer hide behind abstract 'any Sendable'
+    // used as a way to map domain models back to data models
+    public let objectID: any Sendable
```
samwyndham, line 26: thought: I understand wanting to hide the fact that this uses core data underneath but I wonder if it is worth it. It means casting in lots of places. I personally think using `NSManagedObjectID` is fine. But perhaps it is annoying for tests? How about doing something like the following. I assume in most cases where we need a real NSManagedObjectID we will be inside a throwing function:

```swift
struct ObjectID: Sendable {
    private let _managedObjectID: NSManagedObjectID?

    init(managedObjectID: NSManagedObjectID) {
        self.init(managedObjectID)
    }

    private init(_ managedObjectID: NSManagedObjectID?) {
        self._managedObjectID = managedObjectID
    }

    // Used in tests where we never need the real NSManagedObjectID
    #if DEBUG
    static func makeFake() -> ObjectID {
        ObjectID(nil)
    }
    #endif

    var managedObjectID: NSManagedObjectID {
        get throws {
            guard let _managedObjectID else {
                // throw some error
            }
            return _managedObjectID
        }
    }
}
```

samwyndham: Alternatively I believe the following works for creating _mock_ NSManagedObjectID subclasses:

```swift
class MockManagedObjectID: NSManagedObjectID, @unchecked Sendable {

    override func uriRepresentation() -> URL {
        URL(string: "http://example.com")!
    }
}

let id: NSManagedObjectID = MockManagedObjectID()

```

dmitrysimkin: In my option this just it makes more boilerplate code, makes tests harder to write and most importantly leaks core data in domain. We have one layer of models which are abstracted and provide managed `objectID` (other one in tests) so casting back is pretty safe. We can even use force unwrapping, since we'd have a lot of tests to cover if something is not passed correctly, wdyt @samwyndham ?

samwyndham: I'm generally OK with force unwrapping in such situations but I wonder if it might be annoying in tests.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/MessagesIndividualUpdatesFactory.swift
dmitrysimkin: I'll update update optionality in all places

dmitrysimkin: @samwyndham Addressed in https://github.com/wireapp/wire-ios/pull/3492

WireMessaging/Sources/WireMessagingAssembly/ConversationsAssembly.swift
dmitrysimkin: fixed in https://github.com/wireapp/wire-ios/pull/3492

WireMessaging/Sources/WireMessagingDomain/Conversation/UserModel.swift
dmitrysimkin: Updated to use force casting in https://github.com/wireapp/wire-ios/pull/3492

WireMessaging/Sources/WireMessagingUI/Conversation/Message Views/SenderView/SenderViewModel.swift
```diff
@@ -16,18 +16,27 @@
...
+    private var cancellables: Set<AnyCancellable> = []
+
+    package init(
+        state: State,
+        namePublisher: AnyPublisher<String, Never>?
+    ) {
         self.state = state
+        namePublisher?.sink { [weak self] name in
+            self?.state = .exists(AttributedString(name))
+        }.store(in: &cancellables)
```
jullianm, line 40: question: no risk of having a binding still active when a view is reused ? probably works differently in SwiftUI but on UIKit we could have these issues with reusable cells if we didn't clear out the subscriptions which could mess the UI.

jullianm (approved):
Nice work! left a question

samwyndham (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/SenderObserver.swift
```diff
@@ -0,0 +1,68 @@
...
+
+    var authorChangedPublisher: AnyPublisher<String, Never>?
+
+    // later to add publisher for avatar
+
+    init(
+        userID: NSManagedObjectID,
+        viewContext: NSManagedObjectContext
+    ) {
+        viewContext.performAndWait { [weak self, viewContext] in
```
netbe, line 33: thought: I am wondering if we can avoid this, and either make it async or marking the class @MainActor

netbe (commented):
Also looks good, just wonder if we can remove the performAndWait

PR #3472, opened 2025-08-18
fix: mitigate aggressively attempting to commit proposals in broken conversations - WPB-19594

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19594" title="WPB-19594" target="_blank">TaskWPB-19594</a>  [iOS] Mitigate superfluous attempts to commit pending proposals in broken MLS conversations
  </td></table>
  <br />
 

### Issue

This PR attempts to limit the amount of times we attempt to recover from a commit proposal failure in an MLS conversation for a given MLS group ID by _quarantining_ any group IDs when the retry limit is reached AND MLS reset conversation mechanism is disabled.

The reason for this PR is to mitigate repeated attempts to commit proposals which are likely to fail due to a conversation being broken.

My assumption here is that once the MLS reset conversation mechanism is enabled we won't need this mitigation - I have therefore added a check that it is disabled before quarantining. 

### Testing

I'm not sure how to test this. 

---

### Checklist

---

### DIFF
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -1525,6 +1525,13 @@ public final class MLSService: MLSServiceInterface {
                 )
 
                 guard retryCount <= maxRetryAttempts else {
+                    // If MLS conversation reset is DISABLED we quarantine the group to avoid repeated commit attempts
+                    // otherwise assume that things will sort themselves out through conversation reset.
+                    let feature = await featureRepository.fetchAllowedGlobalOperations()
+                    if feature.status == .disabled || feature.config.mlsConversationReset == false {
+                        brokenGroupIDs.insert(groupID)
+                    }
+
                     throw MLSRetryError.retryLimitReached
                 }
 


wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1525,6 +1525,13 @@ public final class MLSService: MLSServiceInterface {
                 )
 
                 guard retryCount <= maxRetryAttempts else {
+                    // If MLS conversation reset is DISABLED we quarantine the group to avoid repeated commit attempts
+                    // otherwise assume that things will sort themselves out through conversation reset.
+                    let feature = await featureRepository.fetchAllowedGlobalOperations()
```
samwyndham, line 1530: @johnxnguyen @typfel Do you think this check is reasonable?

johnxnguyen: Yes but do we need to remove this at some point?

johnxnguyen (approved):

wire-ios-data-model/Source/MLS/MLSService.swift
samwyndham: My assumption is that we will need to remove it once the `mlsConversationReset` feature flag is removed - therefore no specific TODO is necessary

typfel (approved):

PR #3488, opened 2025-08-21
feat:  allow different backends in auth flow - WPB-17734

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17734" title="WPB-17734" target="_blank">TaskWPB-17734</a>  [iOS] Use default backend environment for new authentication flows
  </td></table>
  <br />
 

### Issue
This PR contains changes that allow switching to a different backend in the authentication flow and using this backend environment for the account once authentication completes. Changes include:
- refactor WireAuthentication to use the `NetworkStack` from `WireNetwork` and deleting the network stack it defined.
- always starting the authentication flow with the default backend environment
- when authentication completes, take the environment, cookies, metadata, and proxy credentials and persistent them fro the account.

I highly recommend to review this PR one commit at a time.

### Testing
1. Enable `multibackend` developer flag
2. Log in with account on production
3. Log in with second account on staging.
4. Assert both accounts can send and receive messages, even after app is relaunched.

---

### Checklist

---


jullianm (approved):
Looks good

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -933,9 +933,13 @@ public final class SessionManager: NSObject, SessionManagerType {
     }
 
     @MainActor
-    fileprivate func activateSession(for account: Account) async -> ZMUserSession? {
+    fileprivate func activateSession(
+        for account: Account,
+        newEnvironment: NewEnvironment? = nil
```
samwyndham, line 938: question: Is this or similar methods ever called with the using the default param? I think it is good to avoid too default params unless they are really helpful for common APIs. Otherwise it is easy to forget to pass a param, especially in cases where there are a chain of methods each with their own default param.

WireAuthentication/Tests/WireAuthenticationUITests/Resources/ReferenceImages/LoginViaEmailViewTests/testColorSchemeVariantsWithProxySettings.dark.png
samwyndham, line 1: question: Are these screenshot updates necessary or can we change the mock data. I think https://example.com is more representative of reality than host.

samwyndham (approved):
Nice work @johnxnguyen!

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnxnguyen: Yes some invocations omit the `newEnvironment` parameter. I agree with you and will try to remove in another PR. I would like to refactor the API a bit so that is a separate method to create a new session explicitly after the authentication flow.

WireAuthentication/Tests/WireAuthenticationUITests/Resources/ReferenceImages/LoginViaEmailViewTests/testColorSchemeVariantsWithProxySettings.dark.png
johnxnguyen: Actually these snapshots were updated because of bug fix... previously we were displaying the backend rest api url as the proxy host instead of the proxy host in the config. In general, the mock data can be greatly improved to be more realistic, but I'd do it separately from here.

samwyndham: Oh cool. I think I did something similar once. Must have snuck back in. Or maybe I introduced the bug

PR #3489, opened 2025-08-21
fix: do not cancel batch task if processing events - WPB-19620

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19620" title="WPB-19620" target="_blank">BugWPB-19620</a>  [iOS] Call goes into infinite sync after accepting call inside app
  </td></table>
  <br />
 

### Issue

**Context**: With consumable notifications on, sometimes incoming messages are no longer showing up, only way to see them is to go background/foreground and even after doing so subsequent incoming messages would not show up again.

**Causes**: While debugging, I noticed the batch buffer mechanism from the `PushChannelV2` (see `setUpBatchTask`)  that is yielding events is no longer triggered hence no longer yielding events from its buffer. Reason behind is that the Task it is wrapped into has been cancelled so any subsequent events (i.e incoming messages) that is added to the buffer will **remain** in the buffer until going to the background/foreground again where the same events will be correctly received and processed.

**Solution**: Solution is to no longer tear down that batch Task in the critical place that causes the issue. As a measure of precaution I also avoided to tear down that task in another place where it **could** potentially raise an issue. 
The Task will still be torn down when closing the push channel anyways.

### Testing

Not easily reproducible however when I did get the issue, it was consistently reproducible even after going to background/foreground, batch task would be cancelled early on and no subsequent incoming messages would be received. Commenting out the code that cancel that task fixed the issue right away.

STR :
- consumable notifications on (feature flag in developer tools)
- 1:1
- send message from web client
- app don't receive any
- go background/foreground
- messages are showing up
- send messages again from web client
- messages are not showing up
- and so on..

### Checklist

---


WireNetwork/Sources/WireNetwork/Network/PushChannel/PushChannelV2.swift
```diff
@@ -112,7 +115,6 @@ public final class PushChannelV2: PushChannelV2Protocol {
                         if !drained.isEmpty {
                             continuation.yield(.events(drained))
                             WireLogger.pushChannel.debug("reached batch of size '\(drained.count)' yield")
-                            tearDownBatchTask()
```
jullianm, line 115: this is the one that caused the issue, removing it fixed it right away. as a measure of precaution I also removed the code right above to avoid that kind of issue so that now the task is being torn down only when closing push channel

johnxnguyen (commented):
I think @netbe should review this when he's back because he's most familiar with this code.

WireNetwork/Sources/WireNetwork/Network/PushChannel/PushChannelV2.swift
netbe: yes sounds about right, that we should not teardown the task here

netbe (approved):
looks good

johnxnguyen (approved):

wire-ios/Configuration/Backend.bundle/bella.json
```diff
@@ -5,7 +5,7 @@
         "backendWSURL": "https://nginz-ssl.bella.wire.link",
         "blackListURL": "https://clientblacklist.wire.com/staging",
         "teamsURL": "https://teams.bella.wire.link",
-        "accountsURL": "https://account.anta.bella.link",
+        "accountsURL": "https://account.bella.link",
```
netbe, line 8: minor fix of url

wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DeveloperToolsViewModel.swift
```diff
@@ -180,6 +180,10 @@ final class DeveloperToolsViewModel: ObservableObject {
                             title: "Client ID",
                             value: selfClient?.remoteIdentifier?.uppercased() ?? "None"
                         )),
+                        .text(TextItem(
+                            title: "Team ID",
+                            value: selfUser.teamIdentifier?.uuidString ?? "None"
+                        )),
```
netbe, line 186: added this info in debug

wire-ios-sync-engine/Source/Synchronization/IncrementalSyncObserver.swift
```diff
@@ -105,8 +105,10 @@ final class IncrementalSyncObserver: IncrementalSyncObserverProtocol {
 
             var cancellable: AnyCancellable?
             await withCheckedContinuation { continuation in
+                var resumed = false
                 cancellable = $decryptionState.sink { newDecryptionState in
-                    if newDecryptionState == .done {
+                    if newDecryptionState == .done, !resumed {
+                        resumed = true
```
netbe, line 111: fix crash when subscription will get called twice before cancellation

PR #3494, opened 2025-08-21
fix: iPhone keep ringing after joining the call - WPB-18769

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18769" title="WPB-18769" target="_blank">BugWPB-18769</a>  [iOS] iPhone keep ringing after joining the call from desktop
  </td></table>
  <br />
 

### Issue

There is a group conversation between _User1_ and _User2_.

- _User1_ opens Wire (app is on the foreground). _User1_ gives a permission to use microphone.
- _User1_ logs into desktop app.
- _User2_ initiate the call in the group conversation. 
- _User1_ receives the call on iPhone and moves the app to the background. (the phone is ringing)
- _User1_ joins the call from the desktop app

**Actual result:** the app is ringing.
**Expected result:** the app stops ringing.

When a user receives a notification about a call answered from another client, we always:
- ignore notifications from the self user's clients;
- we take the conversation ID from the top level of the event, in this case _selfConversation_. As a result, we check if there is a received incoming call for this conversation, obviously, there is no call for `selfConversation`. So we decide not stop any calls, since there are no calls for `selfConversation`

What was solved:
- added more logs;
- do not ignore events where the self user is the sender for some event types;
- when there is a calling event, take a conversation ID from the calling content.

### Testing

The use case is described above.

### Checklist

---


WireDomain/Sources/WireDomain/Notifications/Builders/Conversation/ConversationEventNotificationBuilder.swift
```diff
@@ -129,8 +131,12 @@ extension ConversationEventNotificationBuilder {
 
             let eventTimeStamp = time
             let lastReadTimestamp = await conversationLocalStore.lastReadServerTimestamp(conversation)
+            // `selfUser` can be the sender for `mlsMessageAdd` and `proteusMessageAdd` events:
+            // calls can be answered on another device,
+            // so sending notifications is still relevant.
+            let isSenderSelfUserAllowed = isMessageAddEvent ? true : !isSenderSelfUser
```
johnxnguyen: question: why do we do checks in `ConversationEventNotificationBuilder` and not `ConversationCallingEventNotificationBuilder`? I would expect that if we detect a calling event then we handle it the latter and no changes are needed here.

KaterinaWire: we filter our events where selfUser is sender in  `ConversationEventNotificationBuilder`, line #139
validation in the `ConversationCallingEventNotificationBuilder` will be the next step, but we have filter here

johnxnguyen: Could we change the condition to:

```
if isCallMessage {
    // We should try to handle all call messages
    return true
} else if isSenderSelfUser {
    return false
} else if isConversationMuted {
    return false
} else {
    return true
}
```

WireDomain/Sources/WireDomain/Notifications/Builders/Conversation/ConversationMessageAddEventNotificationBuilder.swift
```diff
@@ -61,7 +61,7 @@ struct ConversationMessageAddEventNotificationBuilder: ConversationMessageAddEve
             )
 
             senderID = mlsMessageEvent.senderID
-            conversationID = mlsMessageEvent.conversationID
+            conversationID = getConversationID(for: message, eventConversationID: mlsMessageEvent.conversationID)
```
johnxnguyen: Do we really need to do this? Don't we know when handling the `UserNotification` if it's a call notification or not? And if so, we can check the `message` to find the call conversation id?

KaterinaWire: the event type is `mlsMessageAdd` or `proteusMessageAdd`, so we use `ConversationMessageAddEventNotificationBuilder`. Yes, we know if it's a calling event. We check message.hasCalling in the getConversationID(...).
Let's have 1:1 to clarify your question

jullianm (approved):
looks good

netbe:
to retarget for 4.5 or develop

WireDomain/Sources/WireDomain/Notifications/Builders/Conversation/ConversationCallingEventNotificationBuilder.swift
```diff
@@ -37,16 +38,27 @@ struct ConversationCallingEventNotificationBuilder: ConversationCallingEventNoti
             return nil
         }
 
+        var resolvedConversationID: ConversationID {
+            let callingConversationID = calling.qualifiedConversationID
+            guard !callingConversationID.id.isEmpty,
```
KaterinaWire, line 43: callingConversationID is empty for Proteus conversations, so we need to use conversation ID from the event

WireDomain/Sources/WireDomain/Notifications/Builders/Conversation/ConversationEventNotificationBuilder.swift
KaterinaWire: discussed in 1:1

WireDomain/Sources/WireDomain/Notifications/Builders/Conversation/ConversationMessageAddEventNotificationBuilder.swift
KaterinaWire: discussed in 1:1

johnxnguyen (approved):

PR #3510, opened 2025-08-27
feat: deterministically load Notification Service Extension - WPB-17732

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17732" title="WPB-17732" target="_blank">TaskWPB-17732</a>  [iOS] Deterministically load the notification session
  </td></table>
  <br />
 

### Issue
In this PR we load the Notification Service Extension deterministically similar to the user session and the sharing session. Due to the way we built the NSE, it was a bit hard to make the necessary changes, so I have refactored it by:
- Adding a new `NSEFlow` object which acts as the root component and handles the extracting the raw notification request to loading the user scope (child component)
- Adding `NSEUserScope` which handles setting of the stack for processing the notification request and performing various checks before continue to the client scope (child component)
- Adding `NSEClientScope` which handles the notification request by pulling events, generating notifications, and showing the notifications.

The idea is that when multibackend support is on by default we'll remove the existing `NotificationServiceExtensionFlow` component and its children.

### Testing
1. Enable DeveloperFlag.multibackend.
2. Log in with two accounts, each from a different backend.
3. Put the app in the background, receive messages from both accounts.
4. Assert you see notifications for both accounts.

---

### Checklist

---


wire-ios-sync-engine/Source/SessionManager/SessionFactories.swift
```diff
@@ -145,6 +145,7 @@ open class AuthenticatedSessionFactory {
 
         let userSession = userSessionBuilder.build()
         userSession.setup(
+            apiVersion: nil,
```
johnxnguyen, line 148: This PR also fixes a little bug where the old network stack (operation loop + request strategies) didn't know which api version to use. This is because when multibackend support is enabled, we are no longer storing the api version in `BackendInfo`. You'll see later that code has changed. Here we pass `nil` because this is the legacy code path and it's not needed here.

wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
```diff
@@ -492,6 +492,7 @@ final class UserSessionLoader {
         )
 
         userSession.setup(
+            apiVersion: backendMetadata.apiVersion,
```
johnxnguyen, line 495: Instead here this is a new code path and where we take the api version resolved during session loader and pass it on to the legacy network stack.

wire-ios-sync-engine/Source/Synchronization/ZMOperationLoop+APIVersion.swift
```diff
@@ -23,8 +23,18 @@ import WireDataModel
...
-        guard let current = BackendInfo.apiVersion else { return nil }
-        return .init(value: current)
+        if DeveloperFlag.multibackend.isOn {
+            guard
+                let rawAPIVersion = apiVersion?.int32Value,
+                let apiVersion = APIVersion(rawValue: rawAPIVersion)
+            else {
+                return nil
+            }
+            return .init(value: apiVersion)
```
johnxnguyen, line 33: For simplicity, the operation loop has an optional `apiVersion` which is to be used in case multibackend is on.

wire-ios/Wire Notification Service Extension/NotificationService.swift
```diff
@@ -62,22 +62,66 @@ final class NotificationService: UNNotificationServiceExtension {
     }
 
     private func loadNotificationService() -> NotificationServiceProtocol? {
-        // API version decides which service to use, if we don't have it
-        // yet then we simply supress the notification request.
-        guard let apiVersion = BackendInfo.apiVersion else {
-            WireLogger.notifications.warn("no resolved api version, not loading service")
+        let info = Bundle.appMainBundle.infoDictionary
```
johnxnguyen, line 65: This is the top level object for the NSE. Although these environment vars are only needed for the new code, it doesn't hurt to have them here.

wire-ios/Wire Notification Service Extension/NotificationService.swift
```diff
@@ -62,22 +62,66 @@ final class NotificationService: UNNotificationServiceExtension {
...
+            WireLogger.notifications.warn("no shared user defaults, not loading service")
             return nil
         }
 
-        /// With v8, the new extension is available, but not necessarily
-        /// turned on yet. Regardless, we will use it and later check
-        /// if the new sync is enabled.
-        if apiVersion >= .v8 {
+        if DeveloperFlag.multibackend.isOn {
+            // Only new extension is supported in multibackend mode.
```
johnxnguyen, line 95: With multibackend, the api version is not known yet (it will be resolved later) so will only use the new notification extension (introduce with the sync refactoring).

WireDomain/Sources/WireDomain/Notifications/Components/NSEClientScope.swift
```diff
@@ -0,0 +1,613 @@
...
+    var cryptoboxMigrationManager: CryptoboxMigrationManager { get }
+
+}
+
+/// The scope of a user client within the NSE flow.
+///
+/// Within this scope the necessary use cases to process
+/// the notification request are builds and invoked.
+
+final class NSEClientScope: Component<NSEClientScopeDependency> {
```
johnxnguyen, line 44: The implementation for this has been largely taken from the `PullEventStep`, `GenerationNotificationsStep`, and `ShowNotificationStep`.

WireDomain/Sources/WireDomain/Notifications/Components/NSEFlow.swift
```diff
@@ -0,0 +1,105 @@
...
+import NeedleFoundation
+import UserNotifications
+import WireDataModel
+import WireFoundation
+import WireNetwork
+
+/// The flow for processing notification requests received
+/// in the notification service extension.
+
+final class NSEFlow: BootstrapComponent {
```
johnxnguyen, line 29: This is the new NSEFlow, its implementation is largely taken from `NotificationServiceExtensionFlow` and `ProcessNotificationRequestStep` and will replace it once multibackend is on by default.

WireDomain/Sources/WireDomain/Notifications/Components/NSEUserScope.swift
```diff
@@ -0,0 +1,287 @@
...
+    var preferredAPIVersion: WireNetwork.APIVersion? { get }
+
+}
+
+/// The scope of user within the NSE flow.
+///
+/// Within this scope, validation for the user is performed
+/// and the next scope is prepared: `NSEClientScope`.
+
+final class NSEUserScope: Component<NSEUserScopeDependency> {
```
johnxnguyen, line 42: The implementation for this has largely been taken from `VerifyUserStep`.

WireNetwork/Sources/WireNetwork/Authentication/CookieStorage.swift
```diff
@@ -163,8 +163,8 @@ public actor CookieStorage: CookieStorageProtocol {
     }
 
     private func updateCookieInKeychain(_ cookieData: Data) async throws {
-        let updateQuery = updateQuery(cookieData: cookieData)
-        try await keychain.updateItem(query: fetchQuery, attributesToUpdate: updateQuery)
+        let updateQuery: Set<KeychainQueryItem> = [.data(cookieData.base64EncodedData())]
+        try await keychain.updateItem(query: baseQuery, attributesToUpdate: updateQuery)
```
johnxnguyen, line 167: This fixes another bug where updating the cookie failed. The fix is to use simply the base query to identify the object needing update, and the update query should only contain the data to update.

WireDomain/Sources/WireDomain/Notifications/Components/NSEClientScope.swift
```diff
@@ -0,0 +1,613 @@
...
+    ) {
+        self.clientID = clientID
+        self.restNetworkService = restNetworkService
+        self.webSocketNetworkService = webSocketNetworkService
+        self.apiVersion = apiVersion
+        self.coreDataStack = coreDataStack
+        super.init(parent: parent)
+    }
+
+    func processPayload(
```
samwyndham, line 74: thought (not a request for change). I wonder if these _scopes_ which are all Needle components have too much logic. To me it feels a bit funning that for example `processPayload` is part of an object that is basically for resolving dependencies.

WireNetwork/Sources/WireNetwork/APIs/UpdateEventsAPI/UpdateEventsAPIBuilder.swift
```diff
@@ -37,7 +37,7 @@ public struct UpdateEventsAPIBuilder {
     /// - Parameter version: An api version.
     /// - Returns: A versioned `UpdateEventsAPI`.
 
-    public func makeAPI(for version: APIVersion) -> any UpdateEventsAPI {
+    public func makeAPI(for version: APIVersion) -> some UpdateEventsAPI {
```
samwyndham, line 40: praise: Removing `any`

wire-ios/Wire Notification Service Extension/NotificationService.swift
```diff
@@ -62,22 +62,66 @@ final class NotificationService: UNNotificationServiceExtension {
...
 
     private func loadNotificationService() -> NotificationServiceProtocol? {
-        // API version decides which service to use, if we don't have it
-        // yet then we simply supress the notification request.
-        guard let apiVersion = BackendInfo.apiVersion else {
-            WireLogger.notifications.warn("no resolved api version, not loading service")
+        let info = Bundle.appMainBundle.infoDictionary
+
+        guard let currentAppVersion = info?["CFBundleShortVersionString"] as? String else {
+            WireLogger.notifications.warn("no current app version, not loading service")
```
samwyndham: suggestion: log these issues as error or critical as they should always be there rights?

samwyndham (approved):
Nice work!

WireDomain/Sources/WireDomain/Notifications/Components/NSEClientScope.swift
johnxnguyen: True, we could factor them out into use cases... however then it starts to get a bit odd when we need to descend into a child scope.

In WireAuthentication, each Needle component is tied to a view + view model, and so the logic happens inside the view model and the component is just creating objects. But here in the NSE we do't have any UI and so no view models or ways to move from one step to another (like navigating through views), which is why this approach (and the last approach) treats the components as steps in a flow.

wire-ios-sync-engine/Source/Synchronization/ZMOperationLoop.h
```diff
@@ -38,6 +38,7 @@ extern NSString * const ZMPushChannelIsOpenKey;
 @property (nonatomic, readonly) id<TransportSessionType> transportSession;
 @property (nonatomic) BOOL isDeveloperModeEnabled;
 @property (nonatomic) BOOL isSyncV2Enabled;
+@property (nonatomic, strong, nullable) NSNumber *apiVersion;
```
netbe, line 42: suggestion: add comment on how this apiVersion is used for multibackend but not in legacy

wire-ios/Wire Notification Service Extension/NotificationService.swift
```diff
@@ -62,22 +62,66 @@ final class NotificationService: UNNotificationServiceExtension {
...
 
     private func loadNotificationService() -> NotificationServiceProtocol? {
-        // API version decides which service to use, if we don't have it
-        // yet then we simply supress the notification request.
-        guard let apiVersion = BackendInfo.apiVersion else {
-            WireLogger.notifications.warn("no resolved api version, not loading service")
+        let info = Bundle.appMainBundle.infoDictionary
+
+        guard let currentAppVersion = info?["CFBundleShortVersionString"] as? String else {
+            WireLogger.notifications.critical("no current app version, not loading service")
```
netbe:
```suggestion
            WireLogger.notifications.critical("no current app version, not loading service", attributes: .safePublic)
```

wire-ios/Wire Notification Service Extension/NotificationService.swift
netbe: question: I didn't follow why those were injected instead of having them as computed property, what's the benefit?

netbe (approved):
Approving, and as discussed it might be worth to organise the files in a way to distinguish old needle steps and new multibackend steps, so we can remove them later

johnxnguyen:
> Approving, and as discussed it might be worth to organise the files in a way to distinguish old needle steps and new multibackend steps, so we can remove them later

@netbe created a ticket for that https://wearezeta.atlassian.net/browse/WPB-19817

wire-ios/Wire Notification Service Extension/NotificationService.swift
johnxnguyen: difference is that here we're in the wire-ios project and before these were in the WireDomain project. I find it it gets a little confusing when accessing the main bundle (which one is intended?), so I moved extracting these from the app main bundle at this level because we already have the code to determine that here (whereas the same code was duplicated in WireDomain as part of the needle components).

wire-ios/Wire Notification Service Extension/NotificationService.swift
johnxnguyen: done

wire-ios-sync-engine/Source/Synchronization/ZMOperationLoop.h
johnxnguyen: done

jullianm (approved):
Looks good overall, a thought though.. it seems components now have much more responsibilities than before and I'm not sure they're meant to be used this way but I suppose there's nothing forbidding us to adapt them to meet our needs.

I would still consider splitting them a little bit more if possible and if that makes sense. (non blocking for this PR)

PR #3529, opened 2025-09-01
fix: remove unnecessary call to updateDelegates - WPB-19740

OlivellaO:
### Issue

On ConversationMessageSectionController we had a function updateDelegates() that was trying (sometimes) to update the message from the background thread. This was causing a crash.

Now we've updated the function so the naming is correct (updateMessage), it's getting the message as a parameter and i also remove the calls from the background thread that were not necessary and were causing the crash.

### Testing

- Open group chat and make keyboard up
- Try to minimise the keyboard

Expected - App shouldn’t crash

Also test with different type cells (text, self deleting, file, image, etc.) and in 1 to 1 chats.

---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
@@ -75,14 +75,11 @@ final class ConversationMessageSectionController: NSObject, ZMMessageObserver {
     private let useInvertedIndices: Bool
 
     /// The object that controls actions for the cell.
-    var actionController: ConversationMessageActionController? {
-        didSet { updateDelegates() }
-    }
+    var actionController: ConversationMessageActionController?
 
     /// The message that is being presented.
-    var message: ConversationMessage {
+    private(set) var message: ConversationMessage {
         didSet {
-            updateDelegates()
             changeObservers.removeAll()
             startObservingChanges(for: message)
         }
@@ -91,9 +88,7 @@ final class ConversationMessageSectionController: NSObject, ZMMessageObserver {
     var selfUser: any UserType
 
     /// The delegate for cells injected by the list adapter.
-    weak var cellDelegate: ConversationMessageCellDelegate? {
-        didSet { updateDelegates() }
-    }
+    weak var cellDelegate: ConversationMessageCellDelegate?
 
     /// The object that receives informations from the section.
     weak var sectionDelegate: ConversationMessageSectionControllerDelegate?
@@ -500,7 +495,8 @@ final class ConversationMessageSectionController: NSObject, ZMMessageObserver {
         self.cellDescriptions = cellDescriptions
     }
 
-    private func updateDelegates() {
+    func updateMessage(_ message: ConversationMessage) {
+        self.message = message
         actionController?.message = message
         cellDescriptions.forEach { cellDescription in
             cellDescription.message = message
@@ -512,7 +508,7 @@ final class ConversationMessageSectionController: NSObject, ZMMessageObserver {
     func recreateCellDescriptions(in context: ConversationMessageContext) {
         self.context = context
         createCellDescriptions(in: context)
-        updateDelegates()
+        updateMessage(message)
     }
 
     func isBurstTimestampVisible(in context: ConversationMessageContext) -> Bool {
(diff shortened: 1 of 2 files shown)


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
```diff
@@ -75,14 +76,12 @@ final class ConversationMessageSectionController: NSObject, ZMMessageObserver {
...
 
     /// The object that controls actions for the cell.
-    var actionController: ConversationMessageActionController? {
-        didSet { updateDelegates() }
-    }
+    var actionController: ConversationMessageActionController?
 
     /// The message that is being presented.
-    var message: ConversationMessage {
+    private(set) var message: ConversationMessage {
```
samwyndham, line 81: praise: Making private(set)

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
```diff
@@ -75,14 +76,12 @@ final class ConversationMessageSectionController: NSObject, ZMMessageObserver {
     private let useInvertedIndices: Bool
 
     /// The object that controls actions for the cell.
-    var actionController: ConversationMessageActionController? {
-        didSet { updateDelegates() }
-    }
+    var actionController: ConversationMessageActionController?
```
samwyndham, line 78: issue/question: Do we no longer need to call update message if the actionController has changed? Otherwise how do the cell descriptions get an updated `actionController`?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
```diff
@@ -500,7 +497,7 @@ final class ConversationMessageSectionController: NSObject, ZMMessageObserver {
         self.cellDescriptions = cellDescriptions
     }
 
-    private func updateDelegates() {
+    func updateMessage(_ message: ZMConversationMessage) {
```
samwyndham: suggestion: I'm not sure this is really a more accurate method name because this method doesn't just update the message, it also update the action controller etc. I can't think of a good name but perhaps keeping `func updateMessage(_ message: ZMConversationMessage)` makes sense because it feels less specific. I'm not sure 🤔

OlivellaO: Yes, i was having similar thought before the renaming. Since the call was mostly done when message change that's why i finally rename it and made "message" set private.
Also it seems that other update like the action controller are "side effects" of updating the message.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
OlivellaO: Maybe i can add the call to updateMessage(message) now that message is private and will not be set from the background thread.
Or maybe we don't need that anymore, do you know how i can test that cell descriptions are getting the updated actionController even when not calling updateMessage(message) here?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
```diff
@@ -18,6 +18,7 @@
 
 import Foundation
 import WireFoundation
+import WireLogging
```
OlivellaO: I don't need this import, i will remove it on next PR update.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
samwyndham: OK this is just set one time as it can't be passed in the init method

samwyndham (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageSectionController.swift
samwyndham: OK it makes sense as now it sets the message and there are some side effects performed after.

samwyndham (approved):

netbe (approved):
LGTM;)

PR #3531, opened 2025-09-02
fix: consumable-notifications missing messages - WPB-19919

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19919" title="WPB-19919" target="_blank">BugWPB-19919</a>  [iOS] missing messages when connection get lost
  </td></table>
  <br />
 

### Issue

The missing messages with consumable-notifications were due to 2 things:

- the PushChannel was created multiple times in main app when resuming the app in background, we didn't cancel the task early enough. This was resulting in one pushChannel still open and connected to no IncrementalSync which resulted in completly loosing the event.
- When resuming the app, if the NSE was running, the app was not waiting for the NSE to finish. So events collected by the NSE were stored but processed only in the next time the app started the sync (another trip to background would trigger a new sync).

**Solution**

* Fix debugging NSE: changed name in Debug configuration and updated logging
* Modify the PushChannelState to use fileContext in NSE and mainApp
* Add priority for main App over NSE to start the sync: This uses Darwin notifications to communicate between NSE and mainApp.
* Add cancelling checks of tasks

### Testing

* On anta, have 2 users
* Spam on 1 conversation from web
* Put app in background and foreground while receiving the messages
---

### Checklist

---


WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -82,12 +82,17 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
     public func perform() async throws -> IncrementalSync.Token {
         logger.debug("performing live sync", attributes: .syncAttributes(initialSync: false))
 
+        
+        while (pushChannelState.isOpen()) {
```
netbe: if NSE is running and main app is opened, we end up here until NSE timeout (25 sec max)

WireDomain/Sources/WireDomain/Notifications/Components/NSEClientScope.swift
```diff
@@ -79,12 +79,12 @@ final class NSEClientScope: Component<NSEClientScopeDependency> {
         let eventStream: AsyncStream<[UpdateEvent]>
 
         if dependency.journal[.isConsumableNotificationsEnabled] {
-            guard !PushChannelState(
-                sharedContainerURL: dependency.appContainerURL,
-                clientID: clientID
-            ).isOpen() else {
-                throw Failure.mainAppPushChannelOpened
+            // make sure no pushChannel is open
```
netbe, line 90: the filelock is now set both side, here too in NSE so we always have only one connection opened.

If we have multiple connections opened only the first one will get the events

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
samwyndham: suggestion: put a `Task.sleep(nanoseconds:)` in there to prevent the underlying thread being used for nothing.

WireDomain/Sources/WireDomain/Notifications/Components/Legacy/SyncEventsStep.swift
```diff
@@ -53,15 +54,32 @@ final class SyncEventsStep: Component<SyncEventsDependency>, SyncEventsStepProto
...
+            clientID: selfClientID,
+            postingNotificationName: DarwinNotification
+                .releasingPushChannelAccess,
+            observingNotificationName: DarwinNotification
+                .requestingPushChannelAccess
+        )
         super.init(parent: parent)
     }
 
+    private var currentTask: Task<Void, any Error>?
```
netbe, line 68: this task is used to embed the sync in NSE and to be cancelled if mainApp request access to the websocket

WireDomain/Sources/WireDomain/Synchronization/PushChannelState.swift
```diff
@@ -37,15 +46,15 @@ struct PushChannelState: PushChannelStateProtocol {
...
     }
 
-    func isOpen() -> Bool {
-        fileContext.isLocked()
-    }
-
-    func markAsOpen() {
-        fileContext.acquireDirectoryLock()
+    func markAsOpen() throws {
+        if !fileContext.tryAcquireLock() || Self.locked {
```
netbe: in order to avoid race condition we try to acquire the lock instead of checking if isLocked before. Also acquiring the lock within the same process will always succeed hence the check with the locked static property

WireLogging/Sources/WireLogging/SystemLogger.swift
```diff
@@ -82,16 +82,10 @@ public class SystemLogger: LoggerProtocol {
...
-                finalMessage += extraInfo
-            }
-        #endif
-        if mergedAttributes[.public] as? Bool == true {
+        #if DEBUG
             os_log(osLogType, log: logger, "%{public}@", finalMessage)
-        } else {
+        #else
             os_log(osLogType, log: logger, "\(finalMessage)")
-        }
```
netbe, line 95: in system log, we don't need the tags, moreover the logs can be visible in console on debug configuration.

Note SystemLogger is only needed in debug builds

WireNetwork/Sources/WireNetwork/Network/PushChannel/PushChannelService.swift
```diff
@@ -71,6 +71,11 @@ public final class PushChannelService: PushChannelServiceProtocol {
         var request = request
         let accessToken = try await authenticationManager.getValidAccessToken()
         request.setAccessToken(accessToken)
+
+        // We don't want to proceed if not necessary (in case we've
+        // gone to the background)
+        try Task.checkCancellation()
```
netbe, line 77: porting same logic from PUshCHannelV1 creation

WireNetwork/Sources/WireNetwork/Network/PushChannel/PushChannelV2.swift
```diff
@@ -73,6 +73,10 @@ public final class PushChannelV2: PushChannelV2Protocol {
     }
 
     public func open() async throws -> AsyncThrowingStream<Element, any Error> {
+        // We don't want to proceed if not necessary (in case we've
```
netbe, line 76: porting same logic from PUshCHannelV1 creation

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -280,6 +289,20 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
...
+        }
+        WireLogger.sync.debug("push channel already opened, request NSE release", attributes: .syncAttributes)
+        pushChannelMonitor.notify()
+    }
+
+    private func retrySyncAfterPushChannelClosed() {
+        WireLogger.sync.debug("retry sync after NSE push channel closed", attributes: .syncAttributes)
+        pushChannelMonitor.stopMonitoring()
+        resume()
+    }
```
netbe: communication with NSE: If the main app opens, we want the NSE to release the lock immediately. For the user that way the app shows messages directly within the app

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1232,6 +1233,8 @@ extension ZMUserSession: SyncAgentDelegate {
 
         WaitingGroupTask(context: syncContext) { [weak self] in
             guard let self else { return }
+            try Task.checkCancellation()
```
netbe: going in background and foreground, this is executed multiple times, if the enclosing task is cancelled we should return early too

wire-ios/Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -50,7 +50,7 @@
 		0619A95C2D3FE78700876BDE /* WireAuthenticationUI in Frameworks */ = {isa = PBXBuildFile; productRef = 0619A95B2D3FE78700876BDE /* WireAuthenticationUI */; };
 		061F34192B1E2ED50099CBAB /* AppAuth in Frameworks */ = {isa = PBXBuildFile; productRef = 061F34182B1E2ED50099CBAB /* AppAuth */; };
 		061F341B2B1E2ED50099CBAB /* AppAuthCore in Frameworks */ = {isa = PBXBuildFile; productRef = 061F341A2B1E2ED50099CBAB /* AppAuthCore */; };
-		065A5AE5279B07AC002D4D1E /* Wire Notification Service Extension.appex in Embed App Extensions */ = {isa = PBXBuildFile; fileRef = 065A5ADE279B07AC002D4D1E /* Wire Notification Service Extension.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
+		065A5AE5279B07AC002D4D1E /* NSE.appex in Embed App Extensions */ = {isa = PBXBuildFile; fileRef = 065A5ADE279B07AC002D4D1E /* NSE.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
```
netbe: this is to test NSE locally

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: this has been changed

WireDomain/Sources/WireDomain/Notifications/NotificationServiceExtension.swift
```diff
@@ -227,25 +229,45 @@ extension NotificationServiceExtension {
         }
     }
 
+    private func logSyncEventsStepError(_ error: SyncEventsStep.Failure) {
+        switch error {
+        case .pushChannelAlreadyOpened:
+            logger.error(
+                "Main app is running in foreground with push channel open",
+                attributes: .newNSE
```
netbe, line 241: note here as this error will happen quite often we ignore it for prod

wire-ios/Wire-iOS.xcodeproj/project.pbxproj
netbe: reverted because it breaks CI

WireDomain/Sources/WireDomain/Helpers/DarwinNotificationManager.swift
```diff
@@ -0,0 +1,71 @@
...
+
+public enum DarwinNotification {
+    public static let requestingPushChannelAccess = "com.wire.RequestingPushChannelAccess"
+    public static let releasingPushChannelAccess  = "com.wire.ReleasingPushChannelAccess"
+}
+
+public class DarwinNotificationManager {
+    public init() {}
+
+    private var callbacks: [String: () -> Void] = [:]
```
samwyndham, line 29: suggestion: This structure means that there can only be one callback per notification name meaning that calling `startObserving` with the same name will overwrite previous calls. I think that behavior is surprising. Some possible suggestions:

- Document this behavior in `startObserving`.
- Change name of `startObserving` to make it more explicit: e.g. `setCallback`
- Change API to be cancelation token based. E.g. startObserving would return a cancelation token.

WireDomain/Sources/WireDomain/Helpers/PushChannelMonitor.swift
```diff
@@ -0,0 +1,58 @@
...
+    ) {
+        self.postingNotificationName = postingNotificationName + "_" + clientID
+        self.observingNotificationName = observingNotificationName + "_" + clientID
+    }
+
+    deinit {
+        darwinNotificationManager.stopObserving(name: observingNotificationName)
+    }
+
+    public func startMonitoring(action: @escaping () -> Void) {
```
samwyndham: suggestion:

```suggestion
    public func startMonitoring(onNotification: @escaping () -> Void) {
```

WireDomain/Sources/WireDomain/Notifications/Components/Legacy/SyncEventsStep.swift
```diff
@@ -53,15 +54,32 @@ final class SyncEventsStep: Component<SyncEventsDependency>, SyncEventsStepProto
...
     private var selfClientID: String
 
+    private let pushChannelMonitor: PushChannelMonitor
+
     init(
         parent: any Scope,
         selfClientID: String
     ) {
         self.selfClientID = selfClientID
+        self.pushChannelMonitor = PushChannelMonitor(
```
samwyndham: thought (no need to change anything): For me it is a bit confusing that the same object manages both posting and observing and that they have different notification names.

WireDomain/Sources/WireDomain/Notifications/Components/Legacy/SyncEventsStep.swift
samwyndham: question: Can this be updated from multiple threads? Considering it's purpose we should avoid that. An easy fix might be to make this property `@MainActor`

WireDomain/Sources/WireDomain/Notifications/Components/Legacy/SyncEventsStep.swift
```diff
@@ -72,19 +90,37 @@ final class SyncEventsStep: Component<SyncEventsDependency>, SyncEventsStepProto
...
+
+        currentTask = Task {
+            try Task.checkCancellation()
+            do {
+                try await useCase.invoke()
+            } catch {
+                // either we timeout during decrypting/storing events OR an issue with the sync
+                // In both cases, we end up with a stream of notifications that has not been shown, so we need to
+                // continue
+                // to show them
```
samwyndham: nitpick: Funny new lines whitespace.

WireDomain/Sources/WireDomain/Synchronization/PushChannelState.swift
```diff
@@ -16,15 +16,24 @@
...
 }
 
 struct PushChannelState: PushChannelStateProtocol {
+    enum Failure: Error {
+        case alreadyLocked
+    }
+
+    /// The file context only prevents other processes to get the lock
+    /// same process will succeed, so we need an extra state
+    private static var locked: Bool = false
```
samwyndham: issue: I think we need some guarantee of thread safety when reading/setting `locked`

samwyndham (changes requested):
Nice work and a hard topic! I've added one issue which I think is worth addressing

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -121,6 +128,8 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
                 try await retrier.retry { [self] in
                     try await performSync()
                 }
+            } catch is CancellationError {
+                // ignore error
```
johnxnguyen, line 132: thanks for this

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -280,6 +291,20 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
                 self?.resume()
             }
     }
+
+    private func handlePushChannelAlreadyOpened() {
+        pushChannelMonitor.startMonitoring { [weak self] in
+            self?.retrySyncAfterPushChannelClosed()
+        }
+        WireLogger.sync.debug("push channel already opened, request NSE release", attributes: .syncAttributes)
+        pushChannelMonitor.notify()
```
johnxnguyen: suggestion: what do you think about making this more idiomatic async/await:

```
// Assuming you don't want this to block
Task { [weak self] in
  await pushChannelMonitor.waitUntilPushChannelClosed()
  self?.resume()
}

File: WireDomain/Sources/WireDomain/Helpers/DarwinNotificationManager.swift [hunk 19]
```diff
@@ -0,0 +1,71 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public enum DarwinNotification {
+    public static let requestingPushChannelAccess = "com.wire.RequestingPushChannelAccess"
+    public static let releasingPushChannelAccess  = "com.wire.ReleasingPushChannelAccess"
```
johnxnguyen: suggestion: align with conversation of other Apple notifications:

```
didRequestPushChannelAccess
didReleasePushChannelAccess
```

WireDomain/Sources/WireDomain/Helpers/DarwinNotificationManager.swift
```diff
@@ -0,0 +1,71 @@
...
+}
+
+public class DarwinNotificationManager {
+    public init() {}
+
+    private var callbacks: [String: () -> Void] = [:]
+
+    /// Method to post a Darwin notification
```
johnxnguyen: suggestion: this comment is not giving any new information, consider:

```
/// Post an in inter-process Darwin notification.
///
/// You can post this notification in one process (such as the main app) observe observe
/// it from another process (such as an app extension).

```

WireDomain/Sources/WireDomain/Helpers/PushChannelMonitor.swift
```diff
@@ -0,0 +1,58 @@
...
+        darwinNotificationManager.startObserving(name: observingNotificationName) {
+            action()
+        }
```
johnxnguyen: question: do you think we should have a max waiting time for the response?

WireDomain/Sources/WireDomain/Notifications/Components/NSEClientScope.swift
johnxnguyen: question: is there no need here to observe the Darwin notifications?

WireNetwork/Sources/WireNetwork/Network/PushChannel/PushChannelV2.swift
```diff
@@ -96,15 +103,17 @@ public final class PushChannelV2: PushChannelV2Protocol {
                     switch result {
                     case let .event(event):
                         WireLogger.pushChannel.debug(
-                            "push channel received events: \(event.events.map(\.name))",
+                            " push channel received events: \(event.events.map(\.name))",
```
johnxnguyen:
```suggestion
                            "push channel received events: \(event.events.map(\.name))",
```

johnxnguyen (commented):
looks good, left some comments

WireDomain/Sources/WireDomain/Notifications/Components/NSEClientScope.swift
netbe: still needed, but we do it on demand:
1. the main app sees the push channel is open (via file lock)
2. observe the darwin notifiactions from NSE to release the lock
3. main app send darwin notification to. request to release the lock
4. NSE release the filelock and send back a notification to release the lock
5. main app receives the notification and restart the sync

WireDomain/Sources/WireDomain/Notifications/Components/Legacy/SyncEventsStep.swift
netbe: not sure, i'll look into adding it for safety

WireDomain/Sources/WireDomain/Notifications/Components/Legacy/SyncEventsStep.swift
netbe: I am not to keen on the name but as it is, this object can serve both usage in main app and nse

WireDomain/Sources/WireDomain/Helpers/PushChannelMonitor.swift
netbe: you mean in case we don't get the notification?

WireDomain/Sources/WireDomain/Helpers/DarwinNotificationManager.swift
netbe: i'll go with documentation

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
netbe: checking the idea

WireDomain/Sources/WireDomain/Notifications/Components/Legacy/SyncEventsStep.swift
netbe: so i try but failed to make it work

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -292,21 +292,31 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
...
         resume()
     }
 }
 
+private extension PushChannelMonitorProtocol {
+
+    func waitUntilPushChannelClosed() async {
+        await withCheckedContinuation { continuation in
+            var resumed = false
+
```
samwyndham: question: Should we also log that we are about to start waiting or can we know this from other logs?

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -292,21 +292,31 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
...
+            startMonitoring {
```
samwyndham: question: Is it possible that the other process is killed? If that happens are we stuck here indefinitely?

samwyndham (approved):

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
netbe: i'll add one

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
netbe: right, i think @johnxnguyen had the same concern, let me address this

WireDomain/Sources/WireDomain/Helpers/PushChannelMonitor.swift
netbe: I added max waiting time of 5 sec at the callsite see https://github.com/wireapp/wire-ios/pull/3531/commits/bfb0397e28837593042736b1623bcfd4f51e7a00

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
netbe: https://github.com/wireapp/wire-ios/pull/3531/commits/bfb0397e28837593042736b1623bcfd4f51e7a00

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -280,6 +291,51 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
...
+
+            Task {
+                // in case the other process is killed we resume anyway
+                // so we're not stuck
+                do {
+                    try await Task.sleep(for: .seconds(5))
+                    guard !resumed else { return }
+                    resumed = true
+                    WireLogger.sync.debug(
+                        "timed out waiting for push channel to be closed",
```
johnxnguyen: suggestion(non-blocking): I think it'd be nice to move the timeout logic inside the monitor.

WireDomain/Sources/WireDomain/Helpers/PushChannelMonitor.swift
```diff
@@ -0,0 +1,59 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+// sourcery: AutoMockable
+public protocol PushChannelMonitorProtocol {
+    func startMonitoring(onNotification: @escaping () -> Void)
+    func stopMonitoring()
+    func notify()
+}
```
johnxnguyen: thought: the api is a little confusing because the object has two responsibilities (that I see):
1. the main app signals that it wants to take over the push channel and listens for a response
2. the nse listens for signals from the main app and responds.

What do you think about having two separate objects to make this clearer?

```swift
protocol MainAppPushChannelCoordinator {

    // Send a notification signaling informing extensions to yield the push channel
    // and wait for that request to be acknowledged.
    func signalToExtensionsToYieldPushChannel() async

}

protocol AppExtensionPushChannelCoordinator {

  // Listens for notifications from main app, when one arrives it produces a `YieldRequest`
  // which can be retained while push channel is closed and `acknowledged()` to send
  // a notification back.
  func listenForYieldRequests() async -> YieldRequest

}
```

johnxnguyen (commented):
looks good just some final comments

WireDomain/Sources/WireDomain/Helpers/PushChannelMonitor.swift
netbe: see https://github.com/wireapp/wire-ios/pull/3531/commits/c02819a684d924c01870c88576fa2d14a9e8b43d

johnxnguyen (approved):

PR #3533, opened 2025-09-02
feat: show multibackend error alerts - WPB-18012

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18012" title="WPB-18012" target="_blank">StoryWPB-18012</a>  Use Cloud Temple s3 bucket for asset storage instead of minio.
  </td></table>
  <br />
 

### Issue
This PR contains changes to connect together the api version resolving errors (i.e backend obsolete or client obsolete) thrown by the network stack with the error alerts informing the user.

This is done in three places:
1. Loading a user session in the main app
2. Loading a sharing session in the share extension
3. In the authentication flow

In all three places an alert is shown whether the backend is obsolete or the client is obsolete, with various actions to switch account, logout, or open the App Store to update.

Also included is a debug helper to make simulating the errors easier.

### Testing
1. Enable multibackend support
2. Log in with accounts on backends that throw api resolving errors (or use the debug helper)
3. Try to load the user session, or use the share extension, or start an authentication flow
4. Assert you see error alerts and the actions work as expected.

---

### Checklist

---


wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -766,11 +766,11 @@ public final class SessionManager: NSObject, SessionManagerType {
                             return
                         }
 
+                        accountManager.select(account)
```
johnxnguyen, line 769: This fixes an issue where in case loading the session fails we still have the previous account selected which doesn't really agree with the UI state (e.g the account switcher should only show non selected accounts, but the switcher is shown when failing to load the session).

I think it makes more sense that we select the account, then try to load the session. That way in case of a session failure, we can switch to another account.

wire-ios/Wire-iOS/Sources/UserInterface/Blacklist/BlacklistReason+BlockerViewControllerContext.swift
```diff
@@ -22,10 +22,12 @@ import WireSyncEngine
 extension BlacklistReason {
     var blockerViewControllerContext: BlockerViewControllerContext {
         switch self {
-        case .appVersionBlacklisted, .clientAPIVersionObsolete:
```
johnxnguyen, line 25: Before obsolete client and blacklisted client shared the same "block" screen... with multibackend support they're different (at least for now).

WireAuthentication/Sources/WireAuthenticationUI/Views/Root/RootDestination.swift
johnxnguyen, line 1: No longer needed since we show the account switcher as a root sheet rather than push on the stack.

WireNetwork/Sources/WireNetwork/Network/NetworkStack/NetworkStack.swift
```diff
@@ -25,6 +25,9 @@ import WireLogging
 
 public actor NetworkStack {
 
+    public static var envNameForObsoleteBackend: String?
+    public static var envNameForObsoleteClient: String?
```
johnxnguyen, line 29: A way to simulate which envs throw which errors. I will need to add a developer tool option to set this at run time, but I'll do it in another PR.

wire-ios/Wire-iOS Share Extension/Resources/Localization/Base.lproj/Localizable.strings
```diff
@@ -77,3 +77,8 @@
 "share_extension.voiceover.conversation_under_legal_hold" = "Under legal hold.";
 "share_extension.voiceover.conversation_secure" = "All clients verified.";
 "share_extension.voiceover.conversation_secure_with_ignored" = "Not all clients verified.";
+
+// Update required
+"share_extension.error.update_required.title" = "Update required";
+"share_extension.error.update_required.obsolete_client" = "You are missing out on new features. Get the latest version of  Wire to continue using the app with this account.";
```
netbe:
```suggestion
"share_extension.error.update_required.obsolete_client" = "You are missing out on new features. Get the latest version of Wire to continue using the app with this account.";
```

WireNetwork/Sources/WireNetwork/Network/NetworkStack/NetworkStack.swift
```diff
@@ -100,6 +103,14 @@ public actor NetworkStack {
             return backendMetadata
         }
 
+        #if DEBUG
```
netbe, line 108: question: what is it for ? maybe leave a comment too if this stays for a bit

netbe (approved):
Looks good, left a question

wire-ios/Wire-iOS Share Extension/Resources/Localization/Base.lproj/Localizable.strings
```diff
@@ -77,3 +77,8 @@
...
+"share_extension.error.update_required.obsolete_backend" = "Your Wire server needs to be updated. Please notify your system administrator.";
```
KaterinaWire, line 84: **question:** are these strings from the product team? I'm wondering about "Wire server" phrase, is it relevant for for on-prem customers.

WireNetwork/Sources/WireNetwork/Network/NetworkStack/NetworkStack.swift
johnxnguyen: to simulate api resolving errors and only do so in debug builds. I'll add a comment

wire-ios/Wire-iOS Share Extension/Resources/Localization/Base.lproj/Localizable.strings
johnxnguyen: It's taken from the designs: https://www.figma.com/design/sOAbcd6UVtlZ5ejK4BoEvX/iOS?node-id=9001-5174&t=GPybn16O19jCeTT4-4

WireNetwork/Sources/WireNetwork/Network/NetworkStack/NetworkStack.swift
johnxnguyen, line 108:
```suggestion
        // To simulate api version errors, set these static vars with the names of backend environments.
        // As a precaution to not affect prod builds, only make this available for debug builds.
        #if DEBUG
```

KaterinaWire (approved):

PR #3536, opened 2025-09-04
fix: Unable to tap user avatar to show profile - WPB-19923

SyamalaHari:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19923" title="WPB-19923" target="_blank">BugWPB-19923</a>  [iOS] Unable to tap user avatar to show profile
  </td></table>
  <br />
 

### Issue

Problem is with ConversationSenderMessageDetailsCell: tappedOnAvatar is never called

avatar.leadingAnchor.constraint(equalTo: leadingAnchor, constant: -(24 + 12)) pushes the avatar 36 points to the left, outside the frame of its superview. Even if the avatar is visually displayed (because the cell's clipsToBounds property might be false), the touch system will ignore any taps that occur outside the parent's frame.

So aligned it with in ConversationSenderMessageDetailsCell and added flag isAlreadyAlignedInsideCell so that again it is not aligned in ConversationMessageCellTableViewAdapter

### Testing

Avatar must be tappable now like before

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSenderMessageDetailsCell.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSenderMessageDetailsCell.swift
@@ -312,7 +312,7 @@ final class ConversationSenderMessageCellDescription: ConversationMessageCellDes
 
     let containsHighlightableContent: Bool = false
     lazy var shouldAlignMessageContentForBubbles: Bool = ZMUserSession.shared()?.isChatBubbleSimpleEnabled ?? false
-    lazy var isCellAlreadyAligned: Bool = true
+    let isCellAlreadyAligned: Bool = true
 
     let accessibilityIdentifier: String? = nil
     var accessibilityLabel: String?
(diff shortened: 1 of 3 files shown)


johnxnguyen (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSenderMessageDetailsCell.swift
```diff
@@ -312,6 +312,8 @@ final class ConversationSenderMessageCellDescription: ConversationMessageCellDes
 
     let containsHighlightableContent: Bool = false
     lazy var shouldAlignMessageContentForBubbles: Bool = ZMUserSession.shared()?.isChatBubbleSimpleEnabled ?? false
+    lazy var isCellAlreadyAligned: Bool = true
```
netbe: question: so this cell is always true, right? if so what about changing to a constant?

netbe (commented):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSenderMessageDetailsCell.swift
SyamalaHari: Yes, this can be changed to let. Thanks for the suggestion

samwyndham (approved):

PR #3537, opened 2025-09-04
fix: mls migration notice for calling keeps appearing - WPB-19845

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19845" title="WPB-19845" target="_blank">BugWPB-19845</a>  [iOS] MLS migration notice for calling keeps appearing on iOS
  </td></table>
  <br />
 

### Issue

mls-call

The system message above keeps being inserted in 1:1 MLS conversations during 1:1 calls, even though the conversation has not migrated to MLS (it was already MLS when the call started)

### Solution

During calls, we observe and react to changes to the conversation. If the conversation's changed keys contain the `messageProtocol` key and the value is `mls`, we assume that the protocol has changed from something `!= mis` to `mls`. 

However, this isn't always true. In some cases, we are notified of a change in `messageProtocol` key while the underlying value remains the same (perhaps the conversation is updated from the backend and its properties values remain the same).

The solution adopted here is to save the conversation's message protocol in call snapshots at the beginning of the call, and check if the value has changed when we observe a change in the conversation.

---

### Checklist


samwyndham (approved):
Approving but I wonder if there is a fundamental thing we need to fix in perhaps `ExplicitChangeDetector` or `PotentialChangeDetector` 🤔

KaterinaWire (approved):

David-Henner:
> Approving but I wonder if there is a fundamental thing we need to fix in perhaps `ExplicitChangeDetector` or `PotentialChangeDetector` 🤔

What I noticed while debugging is that both of this change detectors would eventually lead to detect a change in a conversation. Either an explicit change, where all (or almost all? 28 keys) of the conversation's key where updated, or a potential change, which would flag the `ObjectChangeInfo`'s `considerAllKeysChanged` property as `true`. 

Potential changes would be detected after the incremental sync finishes, while the explicit changes were detected after a context save triggered by a request generator.

It's interesting that the `ExplicitChangeDetector` would flag 28 keys as changed though, would be worth investigating. Still, the `NotificationDispatcher` detects changes to the conversation in `didMergeChanges` and I'm not sure where that comes from

PR #3539, opened 2025-09-04
fix: set accent color for .selected IconButton state - WPB-19908

OlivellaO:
### Issue

**I wasn't able to reproduce the issue, this PR is to make sure most usual button states have a background color defined so no default color is used. If you think the issue might be because of another reason or you think we don't need this update on Button state, don't hesitate to contact me**

---

Sometimes the send button from a chat is showing with the wrong background color, as if there was no color assign to the state and maybe using a default iOS color.

When checking the code the state ".selected" has no color assign and it could be that at some point the button is on that state.

### Testing

This is how it should show:
Captura de pantalla 2025-09-04 a las 13 42 24

This is how sometimes is showing up:
image-20250901-133314

---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Helpers/IconButton+Factory.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Helpers/IconButton+Factory.swift
@@ -55,11 +55,13 @@ extension IconButton {
             accessibilityId: "sendButton",
             backgroundColor: [
                 UIControl.State.normal.rawValue: UIColor.accent(),
+                UIControl.State.selected.rawValue: UIColor.accent(),
                 UIControl.State.highlighted.rawValue: UIColor.accentDarken,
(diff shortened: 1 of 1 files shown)


netbe (commented):
question: does this happen if accentColor is changed from another client?

samwyndham (commented):
question: My assumption is that this won't fix the the issue as I expect those buttons aren not in a selected state.. Have you tried reproducing the issue based on @netbe's comment?

I think the issue is more likely to be related to `UIColor.accent()` - that has a bunch of complicated logic, especially in `indexedAccentColor` for example.

OlivellaO:
@netbe @samwyndham I don't think is related with changing color from another client (that works) but maybe dark & light colors are getting mess up at some point. That's the only reason i can think about it.

I'm closing this PR since it doesn't fix anything.

PR #3543, opened 2025-09-05
feat: new reachability for multibackend - WPB-20050

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20050" title="WPB-20050" target="_blank">TaskWPB-20050</a>  [iOS] Check backend reachability per account
  </td></table>
  <br />
 

### Issue
This PR introduces a new `NetworkReachability` that makes use of `NWPathMonitor` to report whether the app has an internet connection. This is in contrast to the old `ZMReachability` which is written in Objective C, uses a lower level C style network API, and checks reachability of specific hosts. 

A key difference between the two is that although `NetworkReachability` is much simpler, it only tells us if there is a connection available and not whether a connection to the Wire backends is available. However, I observed that we only use the existing reachability to hide and show the "no internet" banner, so I think this is fine. If we want to check host reachability, we could extend the code to also send pings to specific hosts.

### Testing
1. Enabled mutilbackend developer flag
2. Log in
3. For the phone in airplane mode, observe the "No internet" banner.
4. Reconnect to the internet, observe the banner goes away.

---

### Checklist

---

### DIFF
--- a/WireNetwork/Sources/WireNetwork/Network/NetworkStack/NetworkReachability.swift
+++ b/WireNetwork/Sources/WireNetwork/Network/NetworkStack/NetworkReachability.swift
@@ -1,0 +1,45 @@
+//
+// Wire
+// Copyright (C) 2025 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+@preconcurrency import Combine
+import Foundation
+import Network
+
+public final class NetworkReachability: Sendable {
+
+    private let monitor = NWPathMonitor()
+    private let isReachableSubject = CurrentValueSubject<Bool, Never>(true)
+    private let queue = DispatchQueue(label: "NetworkReachabilityQueue")
+
+    public var isReachablePublisher: AnyPublisher<Bool, Never> {
+        isReachableSubject.eraseToAnyPublisher()
+    }
+
+    public init() {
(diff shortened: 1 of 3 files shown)


WireNetwork/Sources/WireNetwork/Network/NetworkStack/NetworkReachability.swift
```diff
@@ -0,0 +1,45 @@
...
+    private let isReachableSubject = CurrentValueSubject<Bool, Never>(true)
+    private let queue = DispatchQueue(label: "NetworkReachabilityQueue")
+
+    public var isReachablePublisher: AnyPublisher<Bool, Never> {
+        isReachableSubject.eraseToAnyPublisher()
+    }
+
+    public init() {
+        monitor.pathUpdateHandler = { [weak self] path in
+            print("[DEBUG] reachability changed: \(path.status)")
```
johnxnguyen:
```suggestion
```

samwyndham (approved):
Nice. I'm pro moving to NWPathMonitor!

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -489,6 +492,12 @@ public final class ZMUserSession: NSObject {
 
         super.init()
 
+        if DeveloperFlag.multibackend.isOn {
+            self.isNetworkReachableCancellable = networkReachability.isReachablePublisher
+                .sink { [weak self] isReachable in
+                    isReachable ? self?.didReceiveData() : self?.didGoOffline()
+                }
```
netbe, line 499: question: do we need to shutdown ReachabilityWrapper if multibackend is on?

netbe (commented):
@johnxnguyen I believe there might be other cases besides the internet banner:

1. Looking at`reachability.mayBeReachable` in the code, we also used reachability to reopen the websocket. i believe we don't at the moment -> but that could go under -> https://wearezeta.atlassian.net/browse/WPB-19904
2. ZMTransportRequestScheduler seems to change of status if we're offline

johnxnguyen:
I'm not sure actually this PR is necessary for the initial release: the old ZMReachability should still work for individual accounts because there are separate instances configured with the selected backend. I'll leave this PR open to finish later.

netbe:
@johnxnguyen closing this since @KaterinaWire took most of the reusable code, reopen it if needed

PR #3555, opened 2025-09-09
chore: account management -  lock with passcode - WPB-20119

findms:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20119" title="WPB-20119" target="_blank">TaskWPB-20119</a>  [iOS/QA] Account Management - Lock with passcode
  </td></table>
  <br />
 

### Test

_Please describe the Test._
This is a test for the functionality in account management lock with passcode.
What this test does:

 - Created a team owner with 1 member in team
 - Team member performs login in app
 - Team member navigate to settings> options and enabled the swich lock with passcode
 - Set a password
 - Go to background
 - Wait for 2 seconds
 - Activate the app and validate if lock screen is being presented or conversation screen
 - Enter passcode and land inside of the app

Ran other UI Tests locally on this branch and here is the result:
image

### Testing

_Describe how to test._

---

### Checklist

---


wire-ios/WireUITests/TeamManageTests.swift
```diff
@@ -188,4 +188,37 @@ final class TeamManageTests: WireUITestCase {
...
+        )
+
+        let page = try await app.loginUser(email: teamMember.email, password: teamMember.password)
+            .acceptPopupOnTeamMemberSetup()
+            .setUsername(teamMember.username)
+            .openSettings()
+            .openOptionsMenu()
+            .enableLockWithPasscode()
+            .SetPasscode(passcode)
+            .backgroundAndResume(app: app, forDelay: 60)
```
netbe: issue: this will take time, either we inject a value in the app so we can select a passcode lock with a time of 2 sec? or we think of running these tests not on every PRs

wire-ios/WireUITests/Pages/OptionsOnSettingsPage.swift
```diff
@@ -0,0 +1,62 @@
...
+    var optionsLabel: XCUIElement {
+        app.staticTexts["Options"]
+    }
+
+    var lockWithPasscodeSwitch: XCUIElement {
+        app.descendants(matching: .any)["Lock With Passcode"].firstMatch
+    }
+
+    var conversationsPageLabel: XCUIElement {
+        app.staticTexts["Conversations"]
```
netbe, line 36: though: we need to use Constants at some point so having a target containing these ids which is accessible to the main and the tests

netbe (commented):
the time of waiting 1 min is not going to help regarding the CI, see my comment

wire-ios/WireUITests/TeamManageTests.swift
johnxnguyen: It would be best to override the timeout, perhaps that are other test overrides that we can bundle together.

findms: Tried few other workarounds to avoid this: 
background and foreground - lock doesn't apply
Kill the reopen - session not maintained

but it works only with this for now which is indeed not the best way but it also validates if the passcode gets apply after 1 minute.

is it possible to make this change to apply passcode on background and foreground? or this will be a feature change.

wire-ios/WireUITests/Pages/OptionsOnSettingsPage.swift
findms: @netbe Make sense in monorepo to make a single locator constants with all these locators to be used in main + UI tests. Not sure if creating a new separate file i.e UIElements.swift with all the locator would be best way.
I will create a separate task to pick it up as next to handle this and reach out to you offline.

wire-ios/WireUITests/TeamManageTests.swift
findms: Pushed a change where for if DEBUG changed the timeout to apply lock to 2 seconds and now delay of 60 seconds reduced to 2 seconds though not sure if this is right approach.

johnxnguyen: We could add a developer tool to override the timeout, or have a config file that's used to store automation overrides. What do you prefer?

wire-ios/WireUITests/Pages/OptionsOnSettingsPage.swift
netbe: a single file will work, you need to put it in a common target i think WireFoundation makes sense

wire-ios/WireUITests/TeamManageTests.swift
findms: @johnxnguyen Having config make more sense to me if we have any more automation needs we can use it same

wire-ios/WireUITests/Pages/OptionsOnSettingsPage.swift
findms: Created https://wearezeta.atlassian.net/browse/WPB-20289

wire-ios-data-model/Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -47,7 +47,13 @@ public final class AppLockController: AppLockType {
     }
 
     public var timeout: UInt {
-        legacyConfig?.timeout ?? config.timeout
+        #if DEBUG
+            if let s = ProcessInfo.processInfo.environment["UITEST_APPLOCK_TIMEOUT"],
+               let v = UInt(s) {
+                return v
+            }
```
netbe:
```suggestion
            // override timeout for XCUITests
            if let applockTimeInSeconds = UInt(ProcessInfo.processInfo.environment["UITEST_APPLOCK_TIMEOUT"]) {
                return applockTimeInSeconds
            }
```

netbe (approved):
Looks good, left a suggestion

wire-ios-data-model/Source/Model/FeatureConfig/AppLock/AppLockController.swift
findms: changes pushed according to suggestion.

johnxnguyen (approved):

PR #3560, opened 2025-09-10
chore: remove dependency on backend info - WPB-19987

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19987" title="WPB-19987" target="_blank">TaskWPB-19987</a>  [iOS] Remove dependency on BackendInfo
  </td></table>
  <br />
 

### Issue
The app uses some information about the connected backend, including the resolved api version, its domain, and whether federation and mls are enabled. Since the app can currently connect only to a single backend, this information was stored globally (in user defaults) and accessed with static properties via `BackendInfo`.

Now with multibackend support, each account may connect to a different backend, but each account still needs access to this metadata. We can no longer store this information in user defaults and access it globally because the backend for account A may override this data for account B. Instead, this metadata should be stored on a per-account basis (already implemented) and accessed on a per-account basis (this PR).

In this PR, we minimize the dependency on `BackendInfo` by injecting values that are taken either from the stored backend metadata for the account (if multibackend flag is on) or the global metadata from `BackendInfo` (in case multibackend flag is off). There has been a lot of changes to make this happen because `BackendInfo` was used so widely across the app. 

#### Introduced `resolvedBackendMetadata`
ZMUserSession exposes `resolvedBackendMetadata` which will hold the values to inject. We use this a much as we can as it means we have fewer sources of truth (couldn't avoid this multiplicity), but you'll still see some places where `BackendInfo` is still used namely legacy code paths that shouldn't be activated when multibackend is enabled.

#### API version aware codable conformances
This was a particularly tricky problem to solve. There are 2 legacy payloads (`Payload.Conversation` and `Payload.NewConversation`) that have different encoding/decoding depending on the api version. The difficulty is how to inject the current api version, which was problematic in cases where we were trying to decode `Conversation` as part of an update event. Update event payloads are not versioned so we didn't have a version nearby to inject. I solved this similar to how we do in `WireNetwork`; have separate payloads for update events.

#### Managed object context user info
There were some places (mostly on managed objects) where we used `BackendInfo.domain` and `BackendInfo.isFederationEnabled` where we couldn't easily inject a value, or if we did it would result in an exploding of changes (since these call sites are so widespread). I chose to take simpler (and ugly) workaround and to inject these values into the `userInfo` of all managed object contexts so we can reach for them there if needed. It's ugly I know, but the better solution involves a lot more architectural changes.

#### Is mls enabled?
This value is used to know if we should register an mls client or show some different UI and was stored as `BackendInfo.isMLSEnabled`. But it's not actually part of the metadata return from the backend. Instead it's computed depending on whether the backend has any mls keys. I chose to keep this separate from the rest of the metadata, instead storing it in the journal. Its value is also now computed when loading the session.

### AVSWrapper
There are several callbacks that convert raw strings into `AVSIdentifier`s which now need to have `isFederationEnabled` injected. But because these are C style callbacks they can't capture a local values. Instead these now pass the raw strings to the call center and decoding to `AVSIdentifier`s happens there now.

### Testing
Use the app with multibackend on and also off. 

---

### Checklist

---


wire-ios-data-model/Source/MLS/Migration/ProteusToMLSMigrationCoordinator.swift
```diff
@@ -65,34 +65,39 @@ public class ProteusToMLSMigrationCoordinator: ProteusToMLSMigrationCoordinating
     private let featureRepository: LegacyFeatureRepositoryInterface
     private let actionsProvider: MLSActionsProviderProtocol
     private var storage: ProteusToMLSMigrationStorageInterface
+    var apiVersion: WireTransport.APIVersion?
```
netbe, line 68: question: does it needs to be a var ? is it set after init?

wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack.swift
```diff
@@ -150,20 +150,27 @@ public class CoreDataStack: NSObject, CoreDataStackProtocol {
     private let eventsMigrator: CoreDataMigrator<CoreDataEventsMigrationVersion>
     private var hasBeenClosed = false
 
+    private let localDomain: String?
+    private let isFederationEnabled: Bool
```
netbe, line 154: suggestion: it feels weird to inject these inside the CoreDataStack, I understand the need and given the timeline let's leave it like this, but I would recommend have a follow-up ticket to find a better solution

wire-ios-data-model/Source/ManagedObjectContext/NSManagedObjectContext+BackendMetadata.swift
```diff
@@ -0,0 +1,44 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension NSManagedObjectContext {
```
samwyndham, line 21: question: Did you consider storing the `BackendMetadataProvider` instead here? 

Maybe the advantage would be that it is always up-to-date vs here you have to make sure that you keep things up to date. However I haven't thought about this deeply.

wire-ios-request-strategy/Sources/Payloads/Conversation/Payload+Conversation.swift
samwyndham, line 1: question: Anything to check in detail here or is it mainly copy/pasted?

wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
```diff
@@ -509,6 +530,35 @@ final class UserSessionLoader {
...
+                authenticationManager: authenticationManager
+            )
+            let api = MLSAPIBuilder(apiService: apiService).makeAPI(for: apiVersion)
+            let keys = try await api.getBackendMLSPublicKeys()
+            return keys.removal.isValid
+        } catch
+        URLError.notConnectedToInternet,
+            URLError.networkConnectionLost,
+            MLSAPIError.unsupportedEndpointForAPIVersion,
+            MLSAPIError.mlsNotEnabled {
```
samwyndham, line 556: question: What does this error (`MLSAPIError.mlsNotEnabled`) mean?

samwyndham (commented):
Great work @johnxnguyen. I asked a couple of questions. Will approve after. Due to size of the PR I kind of skipped test & mock changes.

wire-ios-data-model/Source/ManagedObjectContext/NSManagedObjectContext+BackendMetadata.swift
johnxnguyen: Yes but we shouldn't expect the values to change after being set so there shouldn't be a problem with out of date values. Secondly, I added this helper very reluctantly and still want to avoid its usage because for me it's an anti pattern. Having to inject values everywhere is not pretty but I think it's a step in the right direction... with better organization and architecture I think we can greatly improve the situation.

wire-ios-request-strategy/Sources/Payloads/Conversation/Payload+Conversation.swift
johnxnguyen: Copy pasted, but the api versioning is removed for the `CreatedConversation` payload.

wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
johnxnguyen: Honestly not sure, I saw that it was a possible error, my guess is that that backend may return it instead of nil keys.

samwyndham: OK, just a sanity check that we don't want to return `false` in this case instead of throwing?

samwyndham (approved):

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
```diff
@@ -66,13 +66,18 @@ public class ConversationRequestStrategy: AbstractRequestStrategy, ZMRequestGene
     let conversationEventProcessor: ConversationEventProcessor
 
     let removeLocalConversation: RemoveLocalConversationUseCaseProtocol
+    private let apiVersion: WireTransport.APIVersion?
```
netbe, line 69: question: why is apiVersion optional in request strategy like this one?

wire-ios-sync-engine/Source/Data Model/Conversation+ReadReceiptMode.swift
```diff
@@ -55,7 +55,7 @@ public extension ZMConversation {
             if domain == nil {
                 WireLogger.conversation.warn("ZMConversation.setEnableReadReceipts: conversation.domain == nil")
             }
-            let domain = domain ?? BackendInfo.domain ?? "None"
+            let domain = domain ?? userSession.resolvedBackendMetadata.domain ?? "None"
```
netbe: suggestion: safely unwrapped instead of "None" with completion failure like for apiVersion

wire-ios-sync-engine/Source/SessionManager/SessionFactories.swift
```diff
@@ -114,7 +114,8 @@ open class AuthenticatedSessionFactory {
             accountDirectory: coreDataStack.accountContainer,
             syncContext: coreDataStack.syncContext,
             cryptoboxMigrationManager: cryptoboxMigrationManager,
-            coreCryptoKeyMigrationManager: coreCryptoKeyMigrationManager
+            coreCryptoKeyMigrationManager: coreCryptoKeyMigrationManager,
+            localDomain: BackendInfo.domain
```
netbe, line 118: question: just checking this is intended to be BackendInfo.domain ?

wire-ios-sync-engine/Source/Synchronization/Strategies/CallingRequestStrategy.swift
```diff
@@ -182,7 +190,10 @@ public final class CallingRequestStrategy: AbstractRequestStrategy, ZMSingleRequ
             guard let jsonData = response.rawData else { return }
 
             let apiVersion = APIVersion(rawValue: response.apiVersion)!
```
netbe: suggestion: safely unwrap instead of force unwrap

wire-ios-sync-engine/Source/Synchronization/Strategies/CallingRequestStrategy.swift
```diff
@@ -451,7 +469,7 @@ extension CallingRequestStrategy: WireCallCenterTransport {
                 // With MLS we will fetch all clients for each group participant at once
                 // directly from the backend.
                 let userIDs = conversation.localParticipants.map { user in
-                    QualifiedID(uuid: user.remoteIdentifier, domain: user.domain ?? BackendInfo.domain!)
+                    QualifiedID(uuid: user.remoteIdentifier, domain: user.domain ?? localDomain!)
```
netbe: suggestion: if possible safely unwrap the localDomain

wire-ios-sync-engine/Source/Synchronization/Strategies/TypingStrategy.swift
```diff
@@ -119,28 +119,41 @@ public class TypingStrategy: AbstractRequestStrategy, TearDownCapable, ZMEventCo
...
+    private let localDomain: String?
+    private let isFederationEnabled: Bool
 
     @available(*, unavailable)
     override init(withManagedObjectContext moc: NSManagedObjectContext, applicationStatus: ApplicationStatus) {
         fatalError()
     }
 
-    public convenience init(applicationStatus: ApplicationStatus, managedObjectContext: NSManagedObjectContext) {
+    public convenience init(
```
netbe, line 130: thought: I guess typingStrategy is only use in legacy sync, so could rely on BackendInfo

wire-ios-sync-engine/Source/UserSession/ZMClientRegistrationStatus.swift
```diff
@@ -683,16 +690,15 @@ public class ZMClientRegistrationStatus: NSObject, ClientRegistrationDelegate {
...
+    public func needsToRegisterMLSClient(in context: NSManagedObjectContext) -> Bool {
+        guard !Self.needsToRegisterClient(in: context) else {
             return false
         }
         let hasRegisteredMLSClient = ZMUser.selfUser(in: context).selfClient()?.hasRegisteredMLSClient ?? false
         let mlsFeature = LegacyFeatureRepository(context: context).fetchMLS()
 
         let shouldRegisterMLSCLient = mlsFeature.isEnabled
-        let canRegisterMLSCLient = BackendInfo.isMLSEnabled
+        let canRegisterMLSCLient = DeveloperFlag.multibackend.isOn ? isBackendMLSEnabled : BackendInfo.isMLSEnabled
```
netbe, line 701: suggestion: do the condition be done on init

wire-ios-sync-engine/Source/UserSession/ZMUserSession/BackendMetadataProvider.swift
```diff
@@ -0,0 +1,72 @@
...
+//
+
+import WireTransport
+
+public class BackendMetadataProvider {
+
+    private let apiVersionOverride: WireTransport.APIVersion?
+    private let domainOverride: String?
+    private let isFederationEnabledOverride: Bool?
+    private let isBackendMLSEnabledOverride: Bool?
```
netbe, line 29: suggestion: add a comment on the purpose of these properties

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -99,7 +99,7 @@ public final class ZMUserSession: NSObject {
     public let e2eiActivationDateRepository: E2EIActivationDateRepositoryProtocol
 
     let lastEventIDRepository: LastEventIDRepositoryInterface
-    let conversationEventProcessor: ConversationEventProcessor
+    var conversationEventProcessor: ConversationEventProcessor!
```
netbe, line 102: question: why is this necessary?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1373,6 +1396,9 @@ extension ZMUserSession: SyncAgentDelegate {
     }
 
     private func fetchBackendMLSPublicKeys() async {
+        guard !DeveloperFlag.multibackend.isOn else {
+            return
```
netbe, line 1401:
```suggestion
            // fetching done on UserSessionLoader
            return
```

netbe (approved):
phew! long review, left some questions just to double check

wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
johnxnguyen: If we throw then we block the user session, which isn't what we want in this case.

wire-ios-data-model/Source/ManagedObjectContext/CoreDataStack.swift
johnxnguyen: agreed, I'll create a ticket

wire-ios-data-model/Source/MLS/Migration/ProteusToMLSMigrationCoordinator.swift
johnxnguyen: Yes but only for tests, it was a shortcut.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/ConversationRequestStrategy.swift
johnxnguyen: The api version version is always optional because that's what it is in `BackendInfo.apiVersion` and I wanted to keep the same behavior (even if I don't like it). It's also weird to inject the api version here given we also get it when asked to make a new request. But that is also due to poor architecture of these strategy objects: sometimes we get the api version injected in methods, but some other object don't have access to that.

For me the solution is to just rewrite all of these strategies with use cases / repositories / whatever based on our new network stack.

wire-ios-sync-engine/Source/SessionManager/SessionFactories.swift
johnxnguyen: Yes it is

wire-ios-sync-engine/Source/Synchronization/Strategies/TypingStrategy.swift
johnxnguyen: Probably but I'll leave it like this

wire-ios-sync-engine/Source/UserSession/ZMClientRegistrationStatus.swift
johnxnguyen: Probably is that `BackendInfo.isMLSEnabled` may be dynamic (it's set from various places) so better not to be stuck with an out of date value.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: Because it needs to be init after super unfortunately

wire-ios-sync-engine/Source/Data Model/Conversation+ReadReceiptMode.swift
johnxnguyen: done

wire-ios-sync-engine/Source/Synchronization/Strategies/CallingRequestStrategy.swift
johnxnguyen: done

wire-ios-sync-engine/Source/Synchronization/Strategies/CallingRequestStrategy.swift
johnxnguyen: done

wire-ios-sync-engine/Source/UserSession/ZMUserSession/BackendMetadataProvider.swift
johnxnguyen: done

PR #3566, opened 2025-09-10
fix: messages stuck in sending - WPB-20068

David-Henner:
<table>
<td>
  </td></table>
  <br />
 

### Issue

After rotating the core crypto database key, if the user is logged in multiple accounts, one of the accounts may be unable to send messages.

With multiple accounts, the first account to perform the migration overwrites core crypto's database key with a new key. However, the database key storage in the keychain is not scoped by account, so the second account to try and perform the migration is unable to decrypt the database because the key has been changed by the first account.

### Solution

- Rewrite the migration of version 4.3 to handle scoping the keychain items by account
- Add a migration to 4.5 to move unscoped keys to scoped storage

### Discussion

> Do we want to remove the old unscoped key ? 

We cannot do it during the migration because other accounts may need it, so we need another action to do it once all the accounts have migrated

### TODO

May need to update some unit tests

---

### Checklist

---


wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -23,87 +23,139 @@ import WireSystem
...
-            WireLogger.coreCrypto.error("Failed to delete legacy core crypto key: \(String(describing: error))")
-        } catch {
-            // key was not found. no action needed
-        }
+struct ScopedCoreCryptoKeychainItem: KeychainItemProtocol {
+
+    var userID: UUID
+
+    var id: String {
+        "com.wire.mls.key.\(userID.uuidString)"
```
netbe:
```suggestion
        "com.wire.cc.key.\(userID.uuidString)"
```

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -23,87 +23,139 @@ import WireSystem
...
+        if let oldKey = try? fetchUnscopedCoreCryptoKey() {
+            do {
+                WireLogger.coreCrypto.info("Updating core crypto key...", attributes: .safePublic)
+
+                let item = ScopedCoreCryptoKeychainItem(userID: userID)
+                let newKey = try KeychainManager.generateKey(numberOfBytes: 32)
+                try await coreCryptoKeyMigrationManager?.updateKey(path: path, oldKey: oldKey, newKey: newKey)
+
+                // In case another account needs to update it, we don't delete the old key because it's not scoped by account.
+                try KeychainManager.storeItem(item, value: newKey)
```
netbe: suggestion: check if we have multiple account. if not we can delete the old keys

question: in case of multiple accounts, I wonder when can we delete the old key? maybe here conditionally

netbe (commented):
left a couple of comments

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -23,90 +23,139 @@ import WireSystem
...
+                    "Failed to update core crypto key: \(String(describing: error))",
+                    attributes: .safePublic
+                )
+                throw error
+            }
+        }
+    }
+
+    public func migrateToScopedDatabaseKey(inactiveAccounts: Set<Account>) throws {
+        let scopedKeyExists = (try? fetchScopedCoreCryptoKey()) != nil
```
samwyndham: question: sanity check. Is it safe to treat an error as the key not existing?

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -23,90 +23,139 @@ import WireSystem
...
+
+        guard !scopedKeyExists else { return }
+
+        do {
+            WireLogger.coreCrypto.info("Migrating to scoped core crypto key...", attributes: .safePublic)
+
+            // Should we generate a new key and rotate it for extra safety ?
```
samwyndham: question: Do we need to answer this question before merging?

samwyndham (commented):
Looks good. I've added a couple to questions before final review

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: not sure I understand your question, are you asking if it's safe to treat the key not existing as a condition to perform the migration ?

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyMigrationManagerProtocol.swift
```diff
@@ -19,11 +19,19 @@
...
+
+    func migrateDatabaseKeyToBytes(path: String, oldKey: String, newKey: Data) async throws
+
+    func markMigrationToBytesAsSkipped()
+
+    func markMigrationToScopedKeyDone()
+
+    func markKeyRotationAsDone()
 
     func updateKey(path: String, oldKey: Data, newKey: Data) async throws
```
netbe: suggestion: it would be great to add documentation to this protocol

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -22,10 +22,10 @@ import WireSystem
 
 public class CoreCryptoKeyProvider {
 
-    private let coreCryptoKeyMigrationManager: CoreCryptoKeyMigrationManagerProtocol?
+    private let coreCryptoKeyMigrationManager: CoreCryptoKeyMigrationManagerProtocol
     private let userID: UUID
 
-    public init(coreCryptoKeyMigrationManager: CoreCryptoKeyMigrationManagerProtocol?, userID: UUID) {
+    public init(coreCryptoKeyMigrationManager: CoreCryptoKeyMigrationManagerProtocol, userID: UUID) {
```
netbe, line 28: question: how / why does it come non optional?

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -47,64 +47,105 @@ public class CoreCryptoKeyProvider {
...
+        let unscopedKey = try? fetchUnscopedCoreCryptoKey()
 
-                // In case another account needs to update it, we don't delete the old key because it's not scoped by account.
-                try KeychainManager.storeItem(item, value: newKey)
-            } catch {
-                WireLogger.coreCrypto.warn("Failed to update core crypto key: \(String(describing: error))", attributes: .safePublic)
-                throw error
-            }
+        WireLogger.coreCrypto.info("Checking if migration to scoped core crypto key is needed: \(coreCryptoKeyMigrationManager.isMigrationToScopedKeyNeeded)")
+        WireLogger.coreCrypto.info("scopedKey: \(String(describing: scopedKey?.hexString()))")
```
netbe: **issue:** to remove this will reveal the secret to logs

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -47,64 +47,105 @@ public class CoreCryptoKeyProvider {
...
+        WireLogger.coreCrypto.info("unscopedKey: \(String(describing: unscopedKey?.hexString()))")
```
netbe: **issue:** to remove this will reveal the secret to logs

```suggestion
        WireLogger.coreCrypto.info("unscopedKey \(unscopedKey == nil ? "not found" : "exists")")
```

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -47,64 +47,105 @@ public class CoreCryptoKeyProvider {
...
+
+        guard
+            coreCryptoKeyMigrationManager.isMigrationToScopedKeyNeeded,
+            let unscopedKey = unscopedKey,
+            scopedKey == nil
+        else { return }
```
netbe, line 61: question: in which condition do we end up in return? I mean do we need to set `coreCryptoKeyMigrationManager.isMigrationToScopedKeyNeeded = false` ?

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -47,64 +47,105 @@ public class CoreCryptoKeyProvider {
...
+            // Store the unscoped key as scoped key
+            let item = ScopedCoreCryptoKeychainItem(userID: userID)
+            try KeychainManager.storeItem(item, value: unscopedKey)
+
+            // Mark migration as done
+            coreCryptoKeyMigrationManager.markMigrationToScopedKeyDone()
+
+        } catch {
+            WireLogger.coreCrypto.warn(
+                "Failed to migrate to scoped core crypto key: \(String(describing: error))",
```
netbe: **todo:** check if error is safe for public log

**question:** is the error logged at the caller level?

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -47,64 +47,105 @@ public class CoreCryptoKeyProvider {
...
         do {
-            WireLogger.coreCrypto.info("Migrating to scoped core crypto key...", attributes: .safePublic)
+            WireLogger.coreCrypto.info("Updating core crypto key...", attributes: .safePublic)
 
-            // Should we generate a new key and rotate it for extra safety ?
-            let unscopedKey = try fetchUnscopedCoreCryptoKey()
+            // Generate a new key and update the database
+            let oldKey = try fetchScopedCoreCryptoKey()
+            let newKey = try KeychainManager.generateKey(numberOfBytes: 32)
+            WireLogger.coreCrypto.info("replacing old key \(oldKey.hexString()) with \(newKey.hexString())")
```
netbe: **issue:** remove keys from logs
```suggestion
            WireLogger.coreCrypto.info("replacing old key with newKey")
```

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -47,64 +47,105 @@ public class CoreCryptoKeyProvider {
...
+            try await coreCryptoKeyMigrationManager.updateKey(path: path, oldKey: oldKey, newKey: newKey)
+
+            // Store the new key in place of the old one
             let item = ScopedCoreCryptoKeychainItem(userID: userID)
-            try KeychainManager.storeItem(item, value: unscopedKey)
-            // We don't delete the unscoped key because it may be needed by another account to migrate.
+            try KeychainManager.deleteItem(item)
+            try KeychainManager.storeItem(item, value: newKey)
```
netbe: **thought:** I don't know how likely it is but if we fail deleting the item we're loosing the newkey. Maybe we want to catch the deleted error or store to a temporary key, what do you think?

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -47,64 +47,105 @@ public class CoreCryptoKeyProvider {
...
+
+            // Mark the rotation as done
+            coreCryptoKeyMigrationManager.markKeyRotationAsDone()
+
         } catch {
-            WireLogger.coreCrypto.warn("Failed to migrate to scoped core crypto key: \(String(describing: error))", attributes: .safePublic)
+            WireLogger.coreCrypto.warn(
+                "Failed to update core crypto key: \(String(describing: error))",
```
netbe, line 109: **todo:** check that error is safe for logging. Does a failing to access the keychain error contain the value for instance?

wire-ios-data-model/Source/Utilis/KeychainManager.swift
```diff
@@ -27,12 +27,18 @@ protocol KeychainItemProtocol {
...
+    }
+}
+
 public enum KeychainManager {
 
     // MARK: - Keychain access
 
     static func storeItem(_ item: KeychainItemProtocol, value: some Any) throws {
-        WireLogger.keychain.info("storing item (\(item.id))")
+        WireLogger.keychain.info("storing item (\(item.id):\(String(describing:(value as? Data)?.hexString())))")
```
netbe: **issue:** to remove the value

wire-ios-data-model/Source/Utilis/KeychainManager.swift
```diff
@@ -51,6 +57,7 @@ public enum KeychainManager {
             throw Error.failedToFetchItemFromKeychain(status)
         }
 
+        WireLogger.keychain.info("fetched item (\(item.id):\(String(describing:(value as? Data)?.hexString())))")
```
netbe: **todo:** to remove the value

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_3_0.swift
```diff
@@ -27,7 +27,7 @@ struct AppVersionMigration_4_3_0: AppVersionMigration {
     let coreCryptoProvider: CoreCryptoProviderProtocol
 
     func perform() async throws {
-        try await coreCryptoProvider.updateDatabaseKey()
+        // try await coreCryptoProvider.updateDatabaseKey()
```
netbe: suggestion: add comment why this migration should not be done anymore

netbe (commented):
left a couple of comments, especially regarding logs

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: There doesn't seem to be any current usage of the optional so I removed the optionality to avoid unnecessary unwrapping

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: it would contain the key, hence the user ID. But not the value

David-Henner: The error would be `"failed to store item in keychain, OSStatus: \(status)"`

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: The error would be `"failed to store item in keychain, OSStatus: \(status)"`

It's logged at the caller level with the item id but not made public

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -22,91 +22,154 @@ import WireSystem
...
-            try await coreCryptoKeyMigrationManager?.updateKey(path: path, oldKey: oldKey, newKey: newKey)
+            try await coreCryptoKeyMigrationManager.updateKey(path: path, oldKey: oldKey, newKey: newKey)
 
+            // Store the new key in place of the old one
+            let item = ScopedCoreCryptoKeychainItem(userID: userID)
             try KeychainManager.deleteItem(item)
             try KeychainManager.storeItem(item, value: newKey)
+
+            // Mark the rotation as done
+            coreCryptoKeyMigrationManager.markKeyRotationAsDone()
```
David-Henner, line 105: copying previous comment because it got lost after rebasing

> thought: I don't know how likely it is but if we fail deleting the item we're loosing the newkey. Maybe we want to catch the deleted error or store to a temporary key, what do you think?

David-Henner: @netbe if deletion fails, storing would also fail, as it won't overwrite an existing key. So I'm not sure it makes sense to catch the deleted error to continue with the execution. Perhaps a retry could add extra reinforcement but it could also just fail a second time depending on the nature of the error. 

Also not sure how to use a temporary key. I guess we would save it before replacing the old one and keep track of the status to retry later ? I'm concerned it would add too much complexity

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: [e090f79](https://github.com/wireapp/wire-ios/pull/3566/commits/e090f790a5a9ec0511e0094a767d5cd7e9e71c41)

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: [e090f79](https://github.com/wireapp/wire-ios/pull/3566/commits/e090f790a5a9ec0511e0094a767d5cd7e9e71c41)

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: [e090f79](https://github.com/wireapp/wire-ios/pull/3566/commits/e090f790a5a9ec0511e0094a767d5cd7e9e71c41)

wire-ios-data-model/Source/Utilis/KeychainManager.swift
David-Henner: [e090f79](https://github.com/wireapp/wire-ios/pull/3566/commits/e090f790a5a9ec0511e0094a767d5cd7e9e71c41)

wire-ios-data-model/Source/Utilis/KeychainManager.swift
David-Henner: [e090f79](https://github.com/wireapp/wire-ios/pull/3566/commits/e090f790a5a9ec0511e0094a767d5cd7e9e71c41)

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_3_0.swift
David-Henner: [a84f541](https://github.com/wireapp/wire-ios/pull/3566/commits/a84f541b83a1c48f3dc673b63dc275c50d091626)

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: resolving because I moved the discussion to another comment below

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: if there's already a scoped key, we should set it to `false`. Updated here [4afe24f](https://github.com/wireapp/wire-ios/pull/3566/commits/4afe24fdd1e537abbdfc319a905248023606dbe6)

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyMigrationManagerProtocol.swift
David-Henner: [65a700f](https://github.com/wireapp/wire-ios/pull/3566/commits/65a700f96084d465e7ac2f97df5e7c68a221af41)

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_3_0.swift
netbe: great thks!

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
netbe: @David-Henner this is bad too right, once the database is migrated and we're not able to store the key means the database can't be accessed anymore

netbe (approved):
Looks good, there's only the unresolved conversation that bugs me but i do not have a solution to it now

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
samwyndham: I mean that if an error is thrown it doesn't necessarily mean the key doesn't exist in the database. There could be some other error. I think the API of our KeychainManager isn't good here because it throws instead of returning `nil` when the data doesn't exist. So I think a better version of my question is?

If the error was thrown for another reason (not because the key doesn't exist), does that cause any logic issues?

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -22,91 +22,156 @@ import WireSystem
...
             // If there is no key,
             // then this is a fresh install and we do not need to perform migration.
-            coreCryptoKeyMigrationManager?.markMigrationAsSkipped()
+            coreCryptoKeyMigrationManager.markMigrationToBytesAsSkipped()
         }
     }
 
-    private func fetchCoreCryptoKey() throws -> Data {
-        let item = CoreCryptoKeychainItem()
+    private func fetchScopedCoreCryptoKey() throws -> Data {
```
samwyndham, line 143: suggestion: Make these helpers that return a key return optional values `Data?` and intercept the errors to see if the key doesn't exist or it is another error. I _think_ you can do something like (please check though):

```swift
        do {
            return try KeychainManager.fetchItem(item)
        } catch KeychainManager.Error.failedToFetchItemFromKeychain(errSecItemNotFound) {
            return nil
        } catch {
            throw error
        }
```

Then remove a lot of the `try?` from this file so that we are more certain whether there is a real error or the key just doesn't exist.

samwyndham (approved):
Nice work! Approving but please take a look at my suggestion.

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
samwyndham: I make a suggestion in my other comment related to this

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner:
```swift
    private func rotateKey(path: String) async throws {
        guard coreCryptoKeyMigrationManager.isKeyRotationNeeded else {
            return
        }

        handlePotentialPreviousFailures()          

        do {
            WireLogger.coreCrypto.info("Updating core crypto key...", attributes: .safePublic)
            
            // Generate a new key and update the database
            let newKey = try KeychainManager.generateKey(numberOfBytes: 32)

            // Store it temporarily
            let temporaryItem = ScopedCoreCryptoKeychainItem(userID: userID, temporary: true)
            try KeychainManager.storeItem(temporaryItem, value: newKey)
            
            // Update the database
            let oldKey = try fetchScopedCoreCryptoKey()
            try await coreCryptoKeyMigrationManager.updateKey(path: path, oldKey: oldKey, newKey: newKey)

            // Store the new key in place of the old one
            let item = ScopedCoreCryptoKeychainItem(userID: userID)
            try KeychainManager.deleteItem(item)
            try KeychainManager.storeItem(item, value: newKey)
            
            // Delete the temporary key
            try KeychainManager.deleteItem(temporaryItem)

            // Mark the rotation as done
            coreCryptoKeyMigrationManager.markKeyRotationAsDone()

        } catch {
            WireLogger.coreCrypto.warn(
                "Failed to update core crypto key: \(String(describing: error))",
                attributes: .safePublic
            )
            throw error
        }
    }

    private func handlePotentialPreviousFailures() {
        let temporaryItem = ScopedCoreCryptoKeychainItem(userID: userID, temporary: true)
        let item = ScopedCoreCryptoKeychainItem(userID: userID)
        
        do {
            let key = try KeychainManager.fetchItem(item)
            
            if let temporaryKey = try? KeychainManager.fetchItem(temporaryItem) {
                if key == temporaryKey {
                    // we saved the temp key but failed to delete the temporary item
                    // retry to delete the temp item
                }
                
                if key != temporaryKey {
                    // we saved a temporary key and it isn't the same as the current key
                    // means we failed to save the new key
                    // retry to replace the key with the temporary key
                    // then delete the temporary key
                }
            } else {
                // there is no temporary key, this is the first attempt
            }
        }
    }
```
How about this ?

David-Henner: I can also update this 

```swift
try KeychainManager.deleteItem(item)
try KeychainManager.storeItem(item, value: newKey)
```

to this

```swift
try KeychainManager.updateItem(item, value: newKey)
// doesn't exists yet as an API of our `KeychainManager` but I can add it and use `SecItemUpdate` to do the update
```

so we don't risk deleting a key and loosing it if we fail to store the new one

netbe: yes that would work, meaning also that handlePotentialPreviousFailures will in the 2 first if condition mark the rotation as done and exit.

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: I'll consider this in the follow up PR

PR #3574, opened 2025-09-12
fix: deadlock on filelock - WPB-19260

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19260" title="WPB-19260" target="_blank">BugWPB-19260</a>  [iOS] Crash on background deadlock file
  </td></table>
  <br />
 

### Issue

When consumable-notifications is enabled, there is higher chance to get a deadlock in the background.

Assumption this is caused by opened file lock which guards the websocket.

* **[Verified]** When receiving a call (even on main app), we call processCallEvents which will call again the syncAgent.resume() method. As a result the IncrementalSyncV2 will see the lock and asks the NSE to release the lock. As the NSE is not holding the lock but the main app is, this will end up in a recurring loop asking for the NSE to release the lock and retrying the sync (we retry the sync after 5 sec in case NSE missed the notification)

* Make sure PushChannelState is deallocated ~(hence the SafeFileContext which locks the file release the file descriptor which will release the lock) when going to background~. The PushChannelState was retained by the IncrementalSyncV2 (hold by the ClientSessionComponent), with this PR it creates the PushChannelState when running the sync.

**Steps done:**

* break loop in case it's the main app holding the lock
* fix issue where missingEvents is not triggering a fullsync (regular sync) since 4.1
* move backoffRetry inside the NonReentrantTaskManager method making sure no 2 syncs are run at the same time
* add NonReentrantTaskManager for initialsync

### Testing

1. have a call
2. spam messages
3. go to background

This has been tested with QA on build: Playground 4.6.0 build: 17084

---

### Checklist

---


WireDomain/Sources/WireDomain/Synchronization/PushChannelState.swift
```diff
@@ -43,17 +43,17 @@ struct PushChannelState: PushChannelStateProtocol {
                 fatal("could not create file")
             }
         }
-        self.fileContext = SafeFileContext(fileURL: url)
+        self.fileContext = SafeFileContext(fileURL: url, debug: "PushChannelState")
```
netbe: todo: remove after playtest

wire-ios-data-model/Source/Utilis/SafeFileContext.swift
```diff
@@ -17,29 +17,60 @@
...
+        if !debug.isEmpty {
+            Self.instanceCount += 1
+            WireLogger.pushChannel.debug("init SafeFileContext \(debug) \(Self.instanceCount)", attributes: .safePublic)
+            assert(Self.instanceCount < 3, "more than 2 instances of SafeFileContext is not supported")
+        } else {
+            Self.ccInstanceCount += 1
+            WireLogger.pushChannel.debug(
+                "SafeCoreCrypto init SafeFileContext \(Self.instanceCount)",
+                attributes: .safePublic
+            )
```
netbe: todo: remove after playtest

wire-ios-sync-engine/Source/Synchronization/IncrementalSyncObserver.swift
```diff
@@ -34,7 +34,11 @@ final class IncrementalSyncObserver: IncrementalSyncObserverProtocol {
 
     private let decryptionQueue = DispatchQueue(label: "decryptionQueue")
 
-    @Published private var decryptionState: DecryptionState = .notStarted
+    @Published private var decryptionState: DecryptionState = .notStarted {
+        didSet {
+            WireLogger.sync.debug("🍒 decrypted state changed to: \(decryptionState)")
```
netbe: todo: remove after playtest

wire-ios-sync-engine/Source/Synchronization/IncrementalSyncObserver.swift
```diff
@@ -54,6 +58,7 @@ final class IncrementalSyncObserver: IncrementalSyncObserverProtocol {
                 .syncStatePublisher
                 .receive(on: decryptionQueue)
                 .sink { [weak self] syncState in
+                    WireLogger.sync.debug("🍒 \(String(describing: syncState))")
```
netbe: todo: remove after playtest

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -291,13 +319,6 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
                 self?.resume()
             }
     }
-
-    private func handlePushChannelAlreadyOpened() async {
-        WireLogger.sync.debug("push channel opened, waiting until closed", attributes: .syncAttributes)
-        await pushChannelCoordinator.signalToExtensionsToYieldPushChannel()
-        WireLogger.sync.debug("retry sync after NSE push channel closed", attributes: .syncAttributes)
-        resume()
-    }
```
netbe, line 300: this is moved to callsite, simplified

wire-ios/Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
```diff
@@ -290,7 +290,9 @@ final class ConversationListViewController: UIViewController {
         userDefaultsObservation = UserDefaults.standard
             .observe(\.showUnreadConversationsFilter, options: [.new]) { [weak self] _, _ in
                 // Update navigation bar to reflect filter visibility changes
-                self?.updateNavigationItem()
+                DispatchQueue.main.async {
```
netbe, line 293: this fixes an issue where UI code is called on background thread

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
```diff
@@ -410,7 +410,9 @@ public final class ClientSessionComponent {
             syncStateSubject: syncStateSubject,
             coreCryptoProvider: coreCryptoProvider,
             journal: journal,
-            pushChannelState: PushChannelState(sharedContainerURL: sharedContainerURL, clientID: selfClientID)
+            createPushChannelState: { [selfClientID] in
+                PushChannelState(sharedContainerURL: sharedContainerURL, clientID: selfClientID)
```
johnxnguyen, line 414: Question:  Why does it need to be created on demand?

johnxnguyen (approved):

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
samwyndham: I guess it relates to the later comment from @netbe 

> // makes sure that the file descriptor within pushChannelState is released when in background
        // so we're not killed by OS

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -337,26 +362,7 @@ extension SyncAgent: MLSSyncDelegate {
...
-                            .perform()
-                        delegate?.syncAgentDidFinishIncrementalSync(self, isRecovering: false)
-                    }
-                }
-            } catch IncrementalSyncV2.Failure.pushChannelAlreadyOpened {
-                await handlePushChannelAlreadyOpened()
-            } catch {
-                WireLogger.sync.error("failed to perform recovery incremental sync: \(String(describing: error))")
-                throw error
-            }
```
samwyndham, line 359: question: I see there is a tiny difference with `performIncrementalSync()` and the code deleted here (the error handling). I assume this difference is fine?

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
netbe: The filedescriptor is created on init and i thought at some point it was the issue (file opened while in background), as it does not harm I would leave it as it is for now

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -222,53 +224,79 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
...
+                        // swallow error from retrier and start resume
+                        resume()
+
+                    } catch IncrementalSync.Failure.missedEvents {
+
+                        WireLogger.sync.error(
+                            "failed to perform new incremental sync (missed events): recovering with a full sync"
+                        )
+
+                        journal[.isInitialSyncRequired] = true
```
samwyndham, line 264: question: Is this what the _journal_ is for? I kind of thought it was for one time events, but this is more something that can happen often.

samwyndham (approved):
Nice bug hunting!

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
netbe: good question, we can revise it should be one time only. if you look at the brokenMLSGRoupIDS they are updated frequently.

Anyway here i aligned the usage, we can change in a later PR if needed

PR #3575, opened 2025-09-12
fix: unable to switch to multibackend on iPad - WPB-20246

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20246" title="WPB-20246" target="_blank">BugWPB-20246</a>  [iOS/iPad] Not able to switch backend on iPad
  </td></table>
  <br />
 

### Issue
On iPad, when the multibackend flag is enabled and trying to add a second account on a different backend, there is an error alert preventing the switch to a different backend.

The cause is a check to prevent switching when there is an existing account that doesn't get bypassed when the multibackend flag is enabled. It appears that this check is not run on iPhone (there's another check in the WireAuthentication module, but only on iPad.

### Testing
1. Be on iPad, enabled multibackend developer flag
2. Log in to one account
3. Start to add a second account, switch to another backend
4. Assert you can switch and log in to the second account.

---

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/SessionManager/SessionManager+SwitchBackend.swift
+++ b/wire-ios-sync-engine/Source/SessionManager/SessionManager+SwitchBackend.swift
@@ -28,7 +28,7 @@ public extension SessionManager {
     typealias CompletedSwitch = (Result<BackendEnvironment, Error>) -> Void
 
     func canSwitchBackend() -> SwitchBackendError? {
-        guard !accountManager.hasAccounts else {
+        guard DeveloperFlag.multibackend.isOn || !accountManager.hasAccounts else {
             return .loggedInAccounts
         }
 


David-Henner (approved):

wire-ios-sync-engine/Source/SessionManager/SessionManager+SwitchBackend.swift
```diff
@@ -28,7 +28,7 @@ public extension SessionManager {
     typealias CompletedSwitch = (Result<BackendEnvironment, Error>) -> Void
 
     func canSwitchBackend() -> SwitchBackendError? {
-        guard !accountManager.hasAccounts else {
+        guard DeveloperFlag.multibackend.isOn || !accountManager.hasAccounts else {
```
netbe, line 31: question: do you know why is the check not run on iPhone ? is there something to align?

netbe (approved):
✅

wire-ios-sync-engine/Source/SessionManager/SessionManager+SwitchBackend.swift
johnxnguyen: No not sure, I thinks it's fine but will try to take a look later

PR #3578, opened 2025-09-12
fix: various multibackend issues - WPB-20257

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20257" title="WPB-20257" target="_blank">TaskWPB-20257</a>  [iOS] Check blacklist when loading user session
  </td></table>
  <br />
 

Also fixes: WPB-20258 WPB-20259

### Issue
This PR fixes a few issues with multibackend:
1. **The blacklist is now checked every time the session loads:** this prevents the case where the app is blacklisted, but relaunching the app would bypass the blacklist until the next time it's checked (every 6 hours).
2. **Unable to switch accounts on the blocker view controller:** this was because if the app was blacklisted we were ignoring any new app state changes, but this doesn't make sense anymore in multibackend environments. Additionally, it was possible to dismiss the account picker and be stuck.
3. **Sometimes when logging in and deleting a client it would get stuck:** and you'd be left in a broken state which would require a new install. This was due to a race condition where sometimes the user session was not yet available through a singleton and therefore there would be no use case to delete the client.

---

### Checklist

---

### DIFF
--- a/WireFoundation/Sources/WireFoundation/Utilities/DeveloperOverrides.swift
+++ b/WireFoundation/Sources/WireFoundation/Utilities/DeveloperOverrides.swift
@@ -25,27 +25,27 @@ public enum DeveloperOverrides {
     private static let buildNumberKey = "buildNumberOverride"
     public static var buildNumber: String? {
         get {
-            storage?.string(forKey: Self.buildNumberKey)
+            storage?.string(forKey: buildNumberKey)
         } set {
-            storage?.setValue(newValue, forKey: Self.buildNumberKey)
+            storage?.setValue(newValue, forKey: buildNumberKey)
         }
     }
 
     private static let obsoleteBackendEnvKey = "obsoleteBackendEnvOverride"
     public static var obsoleteBackendEnv: String? {
         get {
-            storage?.string(forKey: Self.obsoleteBackendEnvKey)
+            storage?.string(forKey: obsoleteBackendEnvKey)
         } set {
-            storage?.setValue(newValue, forKey: Self.obsoleteBackendEnvKey)
+            storage?.setValue(newValue, forKey: obsoleteBackendEnvKey)
         }
     }
 
     private static let obsoleteClientEnvKey = "obsoleteClientEnvOverride"
     public static var obsoleteClientEnv: String? {
         get {
-            storage?.string(forKey: Self.obsoleteClientEnvKey)
+            storage?.string(forKey: obsoleteClientEnvKey)
         } set {
-            storage?.setValue(newValue, forKey: Self.obsoleteClientEnvKey)
+            storage?.setValue(newValue, forKey: obsoleteClientEnvKey)
         }
     }
 
--- a/WireFoundation/Sources/WireFoundation/Utilities/DeveloperOverrides.swift
+++ b/WireFoundation/Sources/WireFoundation/Utilities/DeveloperOverrides.swift
@@ -20,7 +20,7 @@ import Foundation
 
 public enum DeveloperOverrides {
 
-    private static let storage = UserDefaults(suiteName: "developerOverrides")!
+    nonisolated(unsafe) private static let storage = UserDefaults(suiteName: "developerOverrides")!
 
     public static var buildNumber: String? {
         get {
--- a/WireFoundation/Sources/WireFoundation/Utilities/DeveloperOverrides.swift
+++ b/WireFoundation/Sources/WireFoundation/Utilities/DeveloperOverrides.swift
@@ -16,33 +16,36 @@
 // along with this program. If not, see http://www.gnu.org/licenses/.
 //
 
-import Foundation
+public import Foundation
 
 public enum DeveloperOverrides {
 
-    nonisolated(unsafe) private static let storage = UserDefaults(suiteName: "developerOverrides")!
+    public nonisolated(unsafe) static var storage: UserDefaults?
 
+    private static let buildNumberKey = "buildNumberOverride"
     public static var buildNumber: String? {
         get {
-            storage.string(forKey: "buildNumber")
+            storage?.string(forKey: Self.buildNumberKey)
         } set {
-            storage.setValue(newValue, forKey: "buildNumber")
+            storage?.setValue(newValue, forKey: Self.buildNumberKey)
         }
     }
 
+    private static let obsoleteBackendEnvKey = "obsoleteBackendEnvOverride"
     public static var obsoleteBackendEnv: String? {
         get {
-            storage.string(forKey: "obsoleteBackendEnv")
+            storage?.string(forKey: Self.obsoleteBackendEnvKey)
         } set {
-            storage.setValue(newValue, forKey: "obsoleteBackendEnv")
+            storage?.setValue(newValue, forKey: Self.obsoleteBackendEnvKey)
         }
     }
 
+    private static let obsoleteClientEnvKey = "obsoleteClientEnvOverride"
     public static var obsoleteClientEnv: String? {
         get {
-            storage.string(forKey: "obsoleteClientEnv")
+            storage?.string(forKey: Self.obsoleteClientEnvKey)
         } set {
-            storage.setValue(newValue, forKey: "obsoleteClientEnv")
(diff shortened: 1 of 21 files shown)


wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DeveloperToolsViewModel.swift
```diff
@@ -270,30 +270,37 @@ final class DeveloperToolsViewModel: ObservableObject {
...
+                        AnyView(DeepLinksView(viewModel: DeepLinksViewModel(
+                            router: self?.router,
+                            onDismiss: self?.onDismiss ?? { $0() }
+                        )))
+                    })),
+                    .destination(
+                        DestinationItem(title: "Overrides", makeView: {
+                            AnyView(DeveloperOverridesForm())
+                        })
+                    )
```
johnxnguyen, line 300: Only this part is new, the rest of the changes are from formatting.

wire-ios/Wire-iOS/Sources/UserInterface/Blacklist/BlockerViewController.swift
```diff
@@ -381,6 +381,9 @@ extension BlockerViewController {
             sheet.prefersGrabberVisible = false
             sheet.preferredCornerRadius = 24
         }
+
+        // Present swipe to dismiss.
+        accountSwitcher.isModalInPresentation = true
```
johnxnguyen, line 386: without this you can dismiss the account picker and end up in a UI trap.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/RemoveClients/RemoveClientsViewModel.swift
```diff
@@ -53,7 +50,13 @@ extension RemoveClientsViewController {
                 throw RemoveUserClientError.clientDoesNotExistLocally
             }
 
-            try await removeUserClientUseCase?.invoke(
+            // There's a race condition where we need to remove a client after
+            // login and the user session doesn't exist yet (at init of this view).
+            // So create the use case as late as possible. (Ideally we don't use
+            // this static accessor).
+            let useCase = ZMUserSession.shared()?.removeUserClient
```
johnxnguyen, line 57: admittedly this is not the best solution but I think we need to fix this by refactoring client deletion.

wire-ios/Wire-iOS/Sources/AppStateCalculator.swift
```diff
@@ -189,16 +189,7 @@ final class AppStateCalculator {
...
-        if DeveloperFlag.multibackend.isOn {
-            // If app has been blacklisted due to api version, ignore new state.
-            switch self.appState {
-            case .blacklisted(reason: .backendAPIVersionObsolete), .blacklisted(reason: .clientAPIVersionObsolete):
-                completion?()
-                return
-            default:
-                break
-            }
-        } else {
```
johnxnguyen, line 201: This was blocking account switching from a blacklisted account. I think it's not necessary anyway.

David-Henner (approved):

WireFoundation/Sources/WireFoundation/Utilities/DeveloperOverrides.swift
```diff
@@ -0,0 +1,49 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public enum DeveloperOverrides {
+
+    private nonisolated(unsafe) static let storage = UserDefaults(suiteName: "developerOverrides")!
```
netbe: question: so should it be in UserDefaults appgroup to work on extension?

wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
```diff
@@ -172,8 +172,19 @@ final class UserSessionLoader {
             cookieStorage: cookieStorage
         )
 
+        // Check if this build is blacklisted.
+        if try await isBuildBlacklisted(userSession: userSession) {
+            await userSession.close(deleteCookie: false)
```
netbe, line 177: question: what does this do? teardown the userSession only?

wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
```diff
@@ -172,8 +172,19 @@ final class UserSessionLoader {
...
+            throw Failure.buildIsBlacklisted
+        }
+
         // Perform pending migrations.
-        try await performPendingMigrations(userSession: userSession)
+        do {
+            try await performPendingMigrations(userSession: userSession)
+        } catch {
+            await userSession.close(deleteCookie: false)
+            throw error
```
netbe, line 186: question: so this is blocking loading the userSession if there is any error?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+Authentication.swift
```diff
@@ -85,6 +85,14 @@ extension ZMUserSession {
         completion()
     }
 
+    func close(deleteCookie: Bool) async {
+        await withCheckedContinuation { continuation in
+            close(deleteCookie: deleteCookie) {
+                continuation.resume()
+            }
```
netbe: suggestion: make sure the continuation is never called twice
```suggestion
        await withCheckedContinuation { continuation in
            var resumed = false
            close(deleteCookie: deleteCookie) {
                guard !resumed else { return }
                resumed = true
                continuation.resume()
            }
```

wire-ios/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/RemoveClients/RemoveClientsViewModel.swift
netbe: so correct solution would be inject the userSession in RemoveClientsViewController? can you create a follow up tikcet?

netbe (commented):
left some questions before approving

WireFoundation/Sources/WireFoundation/Utilities/DeveloperOverrides.swift
johnxnguyen: yes I can put it in the shared group

wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
johnxnguyen: Cancels notifications and tears down the session: https://github.com/wireapp/wire-ios/blob/4a65a139dccb0b8a2eed0cd8973f0a2016422aed/wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession%2BAuthentication.swift#L70

wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
johnxnguyen: Yes, but inside we'll need to think about which errors should be blocking and which not.

johnxnguyen:
> so correct solution would be inject the userSession in RemoveClientsViewController? can you create a follow up tikcet?

@netbe it's not so straightforward... there's a lot of mess here with the old authentication flows. I think a better solution is to rewrite client setup and make it part of session loading.

Created a ticket about this topic: https://wearezeta.atlassian.net/browse/WPB-20301

wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
netbe: ok, but should we align the behaviour then or create a follow up for this as current implementation does not block

netbe (approved):

wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
johnxnguyen: I'll follow up in another ticket specifically about this

PR #3580, opened 2025-09-15
fix: handle session loading errors - WPB-19625

johnxnguyen:
### Issue
Now with deterministic session loading there may be some errors thrown when loading the user session. This PR adds handling for these errors:

- The backend metadata can't be resolved due to internet connectivity, fall back to the previous metadata so the session can still be loaded. If no previous metadata exists, the session will be blocked.
- In cases where the session can't be loaded, show an alert and inform the user to try again.
- In the share extension, show an alert to the user informing them to open the main app if necessary.

### Testing
1. Enable multibackend
2. Open app while airplane mode is on
3. Assert you can still use the app, browse conversations etc
4. Turn airplane mode off
5. Assert you can send and receive messages

---

### Checklist

---


samwyndham (approved):
Nice work!

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1058,19 +1064,29 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+                    for: account,
+                    error: .clientIsObsolete
+                )
                 return nil
-            } catch URLError.notConnectedToInternet, URLError.networkConnectionLost {
-                // TODO: [WPB-19625] handle
-                fatalError()
             } catch {
-                // TODO: [WPB-19625] handle
+                WireLogger.sessionManager.critical("failed to load user session: \(String(describing: error))")
```
netbe: question: should this be critical or error? I would reserve critical to non recoverable error (crash). Also this won't show in prod builds (safePublic)

wire-ios-sync-engine/Source/UserSession/BlacklistReason.swift
```diff
@@ -27,4 +27,8 @@ public enum BlacklistReason {
 
     /// The API versions supported by the backend are too old in comparison to the ones supported by the client
     case backendAPIVersionObsolete
+
+    /// Some reason that the session couldn't be loaded.
+    case genericError
```
netbe, line 32: suggestion: associate the underlying error

wire-ios/Wire-iOS/Resources/Localization/Base.lproj/Localizable.strings
```diff
@@ -2147,3 +2147,9 @@ to reset this device and log in again";
 "device.details.section.mls.thumbprint" = "MLS Thumbprint: %@";
 "device.details.certificate_details.title" = "Certificate Details";
 "device.details.certificate_details.copy_to_clipboard" = "Copy to Clipboard";
+
+"account_blocked.generic_error.alert.title" = "Something went wrong";
+"account_blocked.generic_error.alert.message" = "Please contact support if the error persists.";
+"account_blocked.generic_error.alert.switch_accounts" = "Switch accounts";
+"account_blocked.generic_error.alert.send_logs" = "Send debug logs";
```
netbe, line 2154: suggestion: I guess decision goes to design / product but what about "Send report" or "Debug Report" ?

netbe (approved):
looks good, left some suggestions

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnxnguyen: sure can make it an `errorr`

johnxnguyen: done

wire-ios-sync-engine/Source/UserSession/BlacklistReason.swift
johnxnguyen: I wanted to but it's difficult because it needs to pass through `AppState` which needs to be `Equatable`.

wire-ios/Wire-iOS/Resources/Localization/Base.lproj/Localizable.strings
johnxnguyen: Yes copy needs to be finalized

PR #3586, opened 2025-09-15
fix: Update the ChatBubble background color when the accentColor changed from different clients - WPB-20325

SyamalaHari:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20325" title="WPB-20325" target="_blank">TaskWPB-20325</a>  Update the ChatBubble background color when the accentColor changed from different clients(web,iOS)
  </td></table>
  <br />
 

### Issue

ChatBubble background color in iOS is not updated  immediately when the accent Color is changed from different client (ex:Web ). It is only updated after refresh of the page

Update container background color inside configure function of cell. Wheneve accent color is changed from any client configure function is triggered, so updated background color here

### Testing

Change user accesnt color from web and that must be in sync with iOS client. Visually iOS client chat bubble color must immediately change
---

### Checklist

---


netbe (approved):
LGTM;)

samwyndham (approved):
Nice work

PR #3588, opened 2025-09-16
feat: add multibackend support for legacy NSE - WPB-20365

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20365" title="WPB-20365" target="_blank">TaskWPB-20365</a>  [iOS] Add multibackend support in legacy NSE
  </td></table>
  <br />
 

### Issue
Current multibackend support assumes that sync v2 is available, which is only the case when the api version is 8 or higher. To remove this assumption we must add multibackend support in the legacy NSE (for when the api version is less than 8).

### Testing
1. Set preferred api version 7 or less.
2. Log in with multbackend accounts.
3. Put app in background and receive messages for both accounts.
4. Assert you see notifications for both accounts.

---

### Checklist

---

### DIFF
--- a/wire-ios-notification-engine/Sources/LegacyNotificationSessionLoader.swift
+++ b/wire-ios-notification-engine/Sources/LegacyNotificationSessionLoader.swift
@@ -22,7 +22,7 @@ import WireDomain
 import WireFoundation
 import WireNetwork
 
-public struct NotificationSessionLoader {
+public struct LegacyNotificationSessionLoader {
 
     public enum Failure: Error {
 
--- a/wire-ios-notification-engine/Sources/LegacyNotificationSessionLoader.swift
+++ b/wire-ios-notification-engine/Sources/LegacyNotificationSessionLoader.swift
@@ -27,6 +27,7 @@ public struct LegacyNotificationSessionLoader {
     public enum Failure: Error {
 
         case mainAppRequired(message: String)
+        case shouldBeUsingNewNSE
         case failedToFetchBackendEnvironment(any Error)
         case failedToFetchProxyCredentials(any Error)
         case persistenceStoresNotFound
@@ -110,9 +111,8 @@ public struct LegacyNotificationSessionLoader {
             throw Failure.buildIsBlacklisted(buildNumber: buildNumber)
         }
 
-        // Return early if needed.
-        guard !shouldEnableSyncV2(metadata: metadata) else {
-            throw Failure.mainAppRequired(message: "sync v2 should be enabled")
+        guard !journal[.isSyncV2Enabled] else {
+            throw Failure.shouldBeUsingNewNSE
         }
 
         guard let selfClientID = await coreDataStack.syncContext.perform({
@@ -241,13 +241,6 @@ public struct LegacyNotificationSessionLoader {
         } catch {
             throw Failure.failedToCheckBuildBlacklist(error)
         }
-    }
-
-    private func shouldEnableSyncV2(metadata: ResolvedBackendMetadata) -> Bool {
-        return false
-        let isAvailable = metadata.apiVersion >= .v8
-        let isAlreadyEnabled = journal[.isSyncV2Enabled]
-        return isAvailable && !isAlreadyEnabled
     }
 
     private func makeNotificationSession(
--- a/wire-ios/Wire Notification Service Extension/LegacyNotificationService.swift
+++ b/wire-ios/Wire Notification Service Extension/LegacyNotificationService.swift
@@ -229,7 +229,7 @@ final class LegacyNotificationService: UNNotificationServiceExtension, Notificat
     @MainActor
     private func createSession(accountID: UUID) async throws -> NotificationSession {
         if DeveloperFlag.multibackend.isOn {
-            let loader = try NotificationSessionLoader(
+            let loader = try LegacyNotificationSessionLoader(
                 account: Account(userName: "", userIdentifier: accountID),
                 appContainerURL: appContainerURL,
                 appGroupID: appGroupID,
--- a/wire-ios/Wire Notification Service Extension/LegacyNotificationService.swift
+++ b/wire-ios/Wire Notification Service Extension/LegacyNotificationService.swift
@@ -48,7 +48,18 @@ final class CallEventHandler: CallEventHandlerProtocol {
 final class LegacyNotificationService: UNNotificationServiceExtension, NotificationSessionDelegate,
     NotificationServiceProtocol {
 
+    enum Failure: Error {
+
+        case failedToCreateSession(message: String)
+
+    }
+
     // MARK: - Properties
+
+    let appGroupID: String
+    let appContainerURL: URL
+    let currentAppVersion: String
+    let currentBuildNumber: String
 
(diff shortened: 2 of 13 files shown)


wire-ios/Wire Notification Service Extension/NotificationService.swift
```diff
@@ -101,8 +102,18 @@ final class NotificationService: UNNotificationServiceExtension {
             return nil
         }
 
-        if DeveloperFlag.multibackend.isOn {
-            // Only new extension is supported in multibackend mode.
+        guard let apiVersion = fetchLastKnownAPIVersion(
+            for: request,
+            appContainerURL: appContainerURL
+        ) else {
+            WireLogger.notifications.critical(
```
netbe: suggestion: do not log here as it's logged in fetchLastKnownAPIVersion

wire-ios/Wire Notification Service Extension/NotificationService.swift
```diff
@@ -101,8 +102,18 @@ final class NotificationService: UNNotificationServiceExtension {
...
+                "api version unknown, can't load service",
+                attributes: .safePublic
+            )
+            return nil
```
netbe, line 109: question: if nil should we fallback to new NotificationServiceExtension as it will fail later when resolving the backend metadata?

wire-ios/Wire Notification Service Extension/LegacyNotificationService.swift
```diff
@@ -48,41 +48,45 @@ final class CallEventHandler: CallEventHandlerProtocol {
 final class LegacyNotificationService: UNNotificationServiceExtension, NotificationSessionDelegate,
     NotificationServiceProtocol {
 
+    enum Failure: Error {
+
+        case failedToCreateSession(message: String)
```
netbe: question: is it used?

wire-ios/Wire Notification Service Extension/LegacyNotificationService.swift
```diff
@@ -230,17 +234,30 @@ final class LegacyNotificationService: UNNotificationServiceExtension, Notificat
...
-            accountIdentifier: accountID,
-            environment: BackendEnvironment.shared,
-            sharedUserDefaults: .applicationGroup,
-            minTLSVersion: SecurityFlags.minTLSVersion.stringValue
-        )
-
-        session.delegate = self
-        return session
+        if DeveloperFlag.multibackend.isOn {
+            let loader = try NotificationSessionLoader(
```
netbe: **suggestion:** if this is for Legacy only, rename it to LegacyNotificationSessionLoader

netbe (commented):

wire-ios/Wire Notification Service Extension/NotificationService.swift
johnxnguyen: Why continue if it would fail? I'd say it's better not to risk running something that shouldn't run. In any case, it would only be nil until the first metadata is resolved and stored.

wire-ios/Wire Notification Service Extension/LegacyNotificationService.swift
johnxnguyen: No, I'll delete it

wire-ios/Wire Notification Service Extension/NotificationService.swift
johnxnguyen: done

wire-ios/Wire Notification Service Extension/LegacyNotificationService.swift
johnxnguyen: done

wire-ios/Wire Notification Service Extension/NotificationService.swift
netbe: ok makes sense

netbe (approved):
LGTM;) tests might need to be updated: missing arguments for parameters 'appGroupID', 'appContainerURL', 'currentAppVersion', 'currentBuildNumber' in call

wire-ios-notification-engine/Sources/LegacyNotificationSessionLoader.swift
```diff
@@ -0,0 +1,362 @@
...
+
+        // Return early if needed.
+        guard try await !isBuildBlacklisted(networkService: networkServices.blacklist) else {
+            throw Failure.buildIsBlacklisted(buildNumber: buildNumber)
+        }
+
+        // Return early if needed.
+        guard !shouldEnableSyncV2(metadata: metadata) else {
+            throw Failure.mainAppRequired(message: "sync v2 should be enabled")
+        }
```
David-Henner: Should it be this instead ?
```suggestion
        guard shouldEnableSyncV2(metadata: metadata) else {
            throw Failure.mainAppRequired(message: "sync v2 should be enabled")
        }
```

I'm not sure how to read this otherwise, if `shouldEnableSyncV2` is `true` then we throw a failure because sync v2 should be enabled ? seems contradictory

wire-ios-notification-engine/Sources/LegacyNotificationSessionLoader.swift
```diff
@@ -0,0 +1,362 @@
...
+        } catch {
+            throw Failure.failedToFetchBackendEnvironment(error)
+        }
+    }
+
+    private func fetchProxyCredentials(for config: BackendEnvironment2.ProxyConfig) async throws -> WireNetwork
+        .ProxyCredentials? {
+        guard config.needsAuthentication else {
+            return nil
+        }
```
David-Henner, line 152: is this correct ?

wire-ios-notification-engine/Sources/LegacyNotificationSessionLoader.swift
```diff
@@ -0,0 +1,362 @@
...
+
+        do {
+            return try await useCase.invoke()
+        } catch {
+            throw Failure.failedToCheckBuildBlacklist(error)
+        }
+    }
+
+    private func shouldEnableSyncV2(metadata: ResolvedBackendMetadata) -> Bool {
+        return false
```
David-Henner: intentional ?

wire-ios-notification-engine/Sources/LegacyNotificationSessionLoader.swift
johnxnguyen: Thanks for pointing this out, indeed it was a mistake. I've changed it to return early if sync v2 is enabled.

wire-ios-notification-engine/Sources/LegacyNotificationSessionLoader.swift
johnxnguyen: Yes, we only need to fetch proxy credentials if authentication is needed, otherwise return no credentials.

wire-ios-notification-engine/Sources/LegacyNotificationSessionLoader.swift
johnxnguyen: nope, removed

David-Henner (approved):

PR #3590, opened 2025-09-16
feat: perform metadata migrations - WPB-14360

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14360" title="WPB-14360" target="_blank">TaskWPB-14360</a>  [iOS] Perform backend metadata migrations
  </td></table>
  <br />
 

### Issue
When comparing previous backend metadata to new metadata, the difference may require one of two migrations to be performed:
1. If federation is turned on, then the local domain needs to be assigned to all conversations and users in the local database
2. Moving to api v3 or higher then a the access token should be renewed with the self client id.

In this PR:
- when detecting federation being enabled, mark the federation migration as required.
- perform the federation migration after creating the user session
- No need to perform the access token migration: access tokens are only stored in memory, so given there is a self client, a new access token will always be requested with that client id in the lifetime of the app.

### Testing
1. Log in to backend that has federation disabled.
2. Export and inspect the database to assert no domains are stored on `ZMConversation` and `ZMUser`.
3. Enable federation on the backend.
4. Relaunch the app.
5. Export and inspect the database and assert all conversations and users have the local domain set as their domain.

---

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
+++ b/wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
@@ -296,10 +296,6 @@ final class UserSessionLoader {
             if !prevMetadata.isFederationEnabled, newMetadata.isFederationEnabled {
                 // TODO: [WPB-14630] mark federation migration needed
             }
-
-            if prevMetadata.apiVersion < .v3, newMetadata.apiVersion >= .v3 {
-                // TODO: [WPB-14630] mark access token migration needed
-            }
         }
 
         // Store new metadata.
--- a/wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
+++ b/wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
@@ -190,7 +190,10 @@ final class UserSessionLoader {
 
         // Perform pending migrations.
         do {
-            try await performPendingMigrations(userSession: userSession)
+            try await performPendingMigrations(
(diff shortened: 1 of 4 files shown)


wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
```diff
@@ -598,8 +606,31 @@ final class UserSessionLoader {
...
+                try await CoreDataStack.migrateLocalStorage(
+                    accountIdentifier: accountID,
+                    applicationContainer: sharedContainerURL,
+                    migration: {
+                        try $0.migrateToFederation(localDomain: localDomain)
+                    }
+                )
+                journal[.isFederationMigrationRequired] = false
+            } catch {
+                WireLogger.session.error(
```
netbe, line 628: question: sanity check: is the thrown error also logged higher at call site? just to avoid logging too much. + should we log on prod build

netbe (approved):
LGTM;) left a question

wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
johnxnguyen: yes all errors are logged at the call site in `SessionManager`.  I didn't make it safe public because I don't know what could be leaked in the error. I will revisit all the logs around session loading though in another PR.

David-Henner (approved):

PR #3591, opened 2025-09-16
fix: calling actions view moves down - WPB-20193

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20193" title="WPB-20193" target="_blank">BugWPB-20193</a>  [iOS] Calling UI moves down when minimized then maximized
  </td></table>
  <br />
 

### Issue

iOS-after-call-UI-minimized

The call actions view moves outside of the bottom safe area after minimising and maximising the call view.

### Causes

The call actions view is part of the `CallingBottomSheetViewController` which acts as a swipe-able sheet that only shows the call actions by default, and displays the participants list when the user swipes up. This mechanism relies on a constraint that offsets the position of the view.

The offset is based on the call actions view height constraint. When the offset is updated, the height constraint misses the height of the bottom safe area (it evaluates to 0 because the view hasn't been laid out yet).

In addition to this, the height constraint doesn't take into account the `SecurityLevelView` height, which can further lower the calculated offset.

### Solution

- Update the height constraint after ` viewDidLayoutSubviews()` so that the bottom safe area height is factored in
- Inform the `CallingBottomSheetViewController` it needs to update the offset when the call actions view height constraint changes
- Include the `SecurityLevelView` height in the call actions view height constraint

### Testing

- Start / join a call in a classified / unclassified conversation (it needs to have the classification banner)
- Minimise and maximise the call view

---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Calling/BottomSheetContainerViewController/CallingActionsInfoViewController/CallingActionsInfoViewController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Calling/BottomSheetContainerViewController/CallingActionsInfoViewController/CallingActionsInfoViewController.swift
@@ -20,6 +20,10 @@ import UIKit
 import WireDataModel
 import WireDesign
 
+protocol CallingActionsInfoViewControllerDelegate: AnyObject {
+    func actionsViewHeightChanged(to height: CGFloat)
+}
+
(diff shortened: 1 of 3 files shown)


netbe (approved):
LGTM;)

johnxnguyen (approved):

PR #3602, opened 2025-09-18
fix: reachability not working properly - WPB-20410

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20410" title="WPB-20410" target="_blank">BugWPB-20410</a>  [iOS] Network reachability is not functioning properly
  </td></table>
  <br />
 

### Issue
When turning on and off airplane mode, the "not internet" banner wouldn't always show and hide when it should. This is because of recent changes that create reachability per account (for multibackedn support) but not enabling them. The solution is to enable reachability. 

(By default reachability is disabled and only enabled once past login due to proxy mode. But this isn't necessary here since we have already guaranteed proxy mode to be properly configured).

This PR also contains fixes for reachability not set up correctly in the extensions.

### Testing
1. Log in
2. Turn on and off airplane mode
3. Assert that the no internet banner is shown correctly and the app can be used when it's not visible.

---

### Checklist

---


netbe (approved):
looks good, just wondering if NSE and ShareEx SessionLoader should call the same method

johnxnguyen:
> looks good, just wondering if NSE and ShareEx SessionLoader should call the same method

@netbe no because the extensions don't use the "safe reachability" wrapper that needs enabling as it's only relevant for the main app.

David-Henner (approved):

PR #3628, opened 2025-09-23
fix: push supported protocols - WPB-20151

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20151" title="WPB-20151" target="_blank">BugWPB-20151</a>  [iOS] client stuck in sync loop (400 errors, no incoming messages)
  </td></table>
  <br />
 

### Issue

Some users are stuck on InitialSync with a ParserError when pushing supported protocols. Since apiV8, this is actually a new error MLSProtocolError that prevents from removing MLS from supported protocols.

This PR handles this new error, and will correct the calculated supported protocols to include MLS and push them again.

Why is the calculated supported protocols ?

Probably because other self clients are proteus only, it's not clear at this point why they are not supporting mls.

**Solution:**

* Add new `SelfUserAPIError.mlsProtocolError` error
* Renamed Legacy objects used for the Legacy sync (we have 2 implementations of supported protocols, this will help cleaning up)

### Testing

1. test on a backend with supported protocols: mls only, mls migration not started
2. create a proteus only device (not registering mls)
3. login on another client 

Screenshot 2025-09-23 at 14 15 07

---

### Checklist

---


WireDomain/Sources/WireDomain/Synchronization/PushSupportedProtocolsSync.swift
```diff
@@ -45,8 +46,20 @@ public struct PushSupportedProtocolsSync: PushSupportedProtocolsSyncProtocol {
...
+        } catch let SelfUserAPIError.mlsProtocolError(errorMessage) {
+            WireLogger.supportedProtocols
+                .warn("Failed to push supported protocols: \(errorMessage), fallback to adding mls")
+
+            var newSupportedProtocols = supportedProtocols
+            newSupportedProtocols.insert(.mls)
+
+            try await api.pushSupportedProtocols(newSupportedProtocols)
+            await store.updateSelfUserSupportedProtocols(supportedProtocols: newSupportedProtocols.toDomainModel())
+        }
```
johnxnguyen, line 62: suggestion: wrap only the push inside a do catch:

```swift
var supportedProtocols = supportedProtocols
do {
  try await api.pushSupportedProtocols(supportedProtocols)
} catch let SelfUserAPIError.mlsProtocolError(message) {
  supportedProtocols.insert(.mls)
  try await api.pushSupportedProtocols(supportedProtocols)
}

await store.updateSelfuserSupportedProtocols(supportedProtocols: supportedProtocols.toDomainModel())
```

johnxnguyen (approved):

WireDomain/Sources/WireDomain/Synchronization/PushSupportedProtocolsSync.swift
netbe: 69ca24b1f0e07a8ff315c0397fdf5801cba24cdb

David-Henner (approved):

PR #3635, opened 2025-09-24
fix: request loop on supported-protocols - WPB-20151

netbe:
### Issue

Following https://github.com/wireapp/wire-ios/pull/3628, a request loop occurs on pushing the supporting-protocols because the calculation is still wrong.

```
  if remoteProtocols == [.mls], !allClientsMLSReady, migrationState.isOne(of: .notStarted, .ongoing) {
            result = [.proteus]
        }
```
this overrides the fact that the self client supports mls.

**Solution:**

move after this previous check:

```
 // SelfUser supports mls (other client) at the moment, so we should not remove it
        if currentSelfUserSupportedProtocols.contains(.mls) {
            result.insert(.mls)
        }
```

### Testing

See tests
---

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/Services/LegacySupportedProtocolsService.swift
+++ b/wire-ios-sync-engine/Source/Services/LegacySupportedProtocolsService.swift
@@ -106,7 +106,7 @@ public final class LegacySupportedProtocolsService: LegacySupportedProtocolsServ
         if currentSelfUserSupportedProtocols.contains(.mls) {
             result.insert(.mls)
         }
-        
+
         logger.debug("calculated supported protocols: \(result) - legacy")
 
         return result
--- a/wire-ios-sync-engine/Source/Services/LegacySupportedProtocolsService.swift
+++ b/wire-ios-sync-engine/Source/Services/LegacySupportedProtocolsService.swift
@@ -67,7 +67,7 @@ public final class LegacySupportedProtocolsService: LegacySupportedProtocolsServ
 
         logger
             .debug(
-                "remote protocols: \(remoteProtocols), migration state: \(migrationState), allClientsMLSReady: \(allClientsMLSReady) - legacy"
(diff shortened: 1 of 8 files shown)


wire-ios-sync-engine/Source/Services/LegacySupportedProtocolsService.swift
```diff
@@ -107,6 +102,11 @@ public final class LegacySupportedProtocolsService: LegacySupportedProtocolsServ
             result = [.proteus]
         }
 
+        // SelfUser supports mls (other client) at the moment, so we should not remove it
+        if currentSelfUserSupportedProtocols.contains(.mls) {
+            result.insert(.mls)
+        }
```
johnxnguyen, line 108: question: I'm trying to understand what went wrong here... 
- so the self user supports both proteus and mls and this was pushed to the backend at some point
- But recalculating, the previous condition was fulfilled so the supported protocols revert to proteus only?
- If that's the case, why is the previous conditioning being fulfilled now? I would expect that if we already support mls then there is migration is irrelevant.

johnxnguyen (approved):

KaterinaWire (approved):

wire-ios-sync-engine/Source/Use cases/LegacyResolveOneOnOneConversationsUseCase.swift
```diff
@@ -43,7 +43,7 @@ struct LegacyResolveOneOnOneConversationsUseCase: LegacyResolveOneOnOneConversat
             let selfUser = ZMUser.selfUser(in: context)
             return selfUser.supportedProtocols
         }
-
+        WireLogger.supportedProtocols.debug("old: \(oldProtocols)")
```
David-Henner: should this log be here ? if so, I think it should be more descriptive

David-Henner (approved):

wire-ios-sync-engine/Source/Use cases/LegacyResolveOneOnOneConversationsUseCase.swift
netbe: removed

PR #3614, opened 2025-09-19
revert: messages stuck in sending - WPB-20068

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20068" title="WPB-20068" target="_blank">BugWPB-20068</a>  [iOS] Messages stuck in sending
  </td></table>
  <br />
 

This reverts commit 9d865ecf365811f7c3f907a07f06d2b5baa700cd.

PR #3637, opened 2025-09-24
fix: messages stuck in sending - WPB-20068

David-Henner:
<table>
<td>
  </td></table>
  <br />
 

This issue was worked on first here (https://github.com/wireapp/wire-ios/pull/3566) but had to be reverted (https://github.com/wireapp/wire-ios/pull/3614) due to issues when logging out / logging in

### Issue

Version 4.3 of the app introduced a migration of the core crypto database key, a simple key rotation.

After rotating the core crypto database key, if the user is logged in multiple accounts, one of the accounts may be unable to send messages.

### Causes

The database key is saved in the keychain and isn't scoped by account. With multiple accounts, each account accesses the same database key. We didn't notice this until now because the key was never modified.

So, when account A performs the 4.3 migration and rotates the database key, they will update the key in the keychain. Then account B will try to do the migration, but will fail because they key has changed and they're now unable to decrypt the database. Any other operation that involves using the core crypto database will fail.

### Solution

- Migrate from an unscoped key to a scoped key
- Rotate the (now scoped) database key to a new scoped key

### Changes

- Introduced a **scoped key**. Scoping it with a **unique key identifier** 
- Stored the **unique key identifier** in UserDefaults, scoping this by user ID. This avoids stale keys being mistaken for valid keys after app re-install or re-login.
- Added a key migration to scope the database key.
- Added a migration to rotate the database key, with protections against loosing the key in case migration fails.
- Added logic to keep track of stale keys
- Added use case to remove the database keys on logout
- Disabled the 4.3 migration
- Cleaned up stale and duplicated code

### Future improvements

- Delete all keychain items on first install of the app
- Regularly check and delete stale keys

### Testing

**Scenario 1**
- Install app version <= 4.2
- Log into multiple accounts and send messages
- Install app version 4.3
- Send messages, notice it breaks
- Install latest version
- Log out of broken account, log back in
- Send messages

**Scenario 2**
- Install app version <= 4.2
- Log into multiple accounts and send messages
- Install latest version
- Verify you can send messages

**Scenario 3**
- Install latest version
- Log in with one or multiple accounts
- Send messages
- Log out / log back in
- Verify you can send messages

---

### Checklist

---


wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -17,96 +17,252 @@
...
+            return try createCoreCryptoKey()
+        } else {
+            throw Error.keyNotFound
+        }
+    }
+
+    private func performKeyMigrationsIfNeeded(path: String) async throws {
+        try await migrateDatabaseKeyToBytes(path: path)
+        try migrateToScopedDatabaseKey(path: path)
+        try? await rotateKeyIfNeeded(path: path)
```
netbe: **question:** why  try optional?

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -17,96 +17,252 @@
...
-            // key was not found. no action needed
-        }
+    enum Error: Swift.Error {
+        case keyNotFound
     }
 }
 
 struct CoreCryptoKeychainItem: KeychainItemProtocol {
+    static let scopedBaseId = "com.wire.cc.key"
+    static let unscopedId = "com.wire.mls.key"
```
netbe, line 265:
```suggestion
    /// deprecated key
    static let unscopedId = "com.wire.mls.key"
```

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -17,96 +17,252 @@
...
-            _ = try KeychainManager.fetchItem(legacyItem) as Data
-            WireLogger.coreCrypto.info("Found legacy core crypto key. Deleting...")
-            try KeychainManager.deleteItem(legacyItem)
-            WireLogger.coreCrypto.info("Deleted legacy core crypto key")
-        } catch let KeychainManager.Error.failedToDeleteItemFromKeychain(error) {
-            WireLogger.coreCrypto.error("Failed to delete legacy core crypto key: \(String(describing: error))")
-        } catch {
-            // key was not found. no action needed
-        }
+    enum Error: Swift.Error {
```
netbe:
```suggestion
    enum Failure: Error {
```

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -17,96 +17,252 @@
...
+            try await rotateKey(path: path)
         } catch {
-            if createIfNeeded {
-                return try createCoreCryptoKey()
-            } else {
+            WireLogger.coreCrypto.warn(
+                "Failed to rotate core crypto key: \(String(describing: error))",
+                attributes: .safePublic
+            )
+            throw error
```
netbe: suggestion: does the caller logs too, maybe we should log at the caller site and embed the error here in a specific error. This will avoid duplicate logs, what do you think ?

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -17,96 +17,252 @@
...
         }
     }
 
-    public func updateDatabaseKey(path: String) async throws {
-        if let oldKey = try? fetchCoreCryptoKey() {
-            let item = CoreCryptoKeychainItem()
-            let newKey = try KeychainManager.generateKey(numberOfBytes: 32)
-            try await coreCryptoKeyMigrationManager?.updateKey(path: path, oldKey: oldKey, newKey: newKey)
+    private func logRotation(_ message: String) {
+        WireLogger.coreCrypto.info("[key rotation] \(message)")
```
netbe: Question: does this need to be logged on prod?

```suggestion
        WireLogger.coreCrypto.info("[key rotation] \(message)", attribures: .safePublic)
```

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -17,96 +17,252 @@
...
+            do {
+                try await coreCryptoKeyMigrationManager.migrateDatabaseKeyToBytes(
+                    path: path,
+                    oldKey: oldKey.base64EncodedString(),
+                    newKey: oldKey
+                )
+            } catch {
+                WireLogger.coreCrypto.warn(
+                    "Failed to migrate core crypto key: \(String(describing: error))",
+                    attributes: .safePublic
```
netbe: todo: check same remark as above for logging and throwing at same level

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -17,96 +17,252 @@
...
+            }
+        } set {
+            defaults.setUUID(newValue, forKey: .uniqueKeyIdentifier)
+        }
+    }
+
+    public init(
+        coreCryptoKeyMigrationManager: CoreCryptoKeyMigrationManagerProtocol,
+        userID: UUID,
+        staleKeysTracker: StaleCoreCryptoKeysTrackerProtocol = StaleCoreCryptoKeysTracker(),
```
netbe: suggestion: pass StaleCOreCryptoKeysTracker the sharedCOntainer or do we exit on extensions?

netbe (approved):
looks good, left some minor comments. For testing, let's have a playtest with different playground builds

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyMigrationManagerProtocol.swift
```diff
@@ -19,12 +19,41 @@
 // sourcery: AutoMockable
 public protocol CoreCryptoKeyMigrationManagerProtocol {
 
-    var isMigrationNeeded: Bool { get }
+    /// Wether or not we need to update the database key from String to Data (bytes) format
```
johnxnguyen:
```suggestion
    /// Whether we need to update the database key from String to Data (bytes) format
```

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyMigrationManagerProtocol.swift
```diff
@@ -19,12 +19,41 @@
...
-    func performMigrationIfNeeded(path: String, oldKey: String, newKey: Data) async throws
+    var isMigrationToBytesNeeded: Bool { get }
 
-    func markMigrationAsSkipped()
+    /// Wether or not we need to migrate the stored database key from an unscoped keychain item to a keychain item
+    /// scoped by user
+
+    var isMigrationToScopedKeyNeeded: Bool { get }
+
+    /// Wether or not we need to re-encrypt the database with a new key and update it in keychain storage
```
johnxnguyen:
```suggestion
    /// Whether we need to re-encrypt the database with a new key and update it in keychain storage
```

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -17,126 +17,272 @@
...
             // then this is a fresh install and we do not need to perform migration.
-            coreCryptoKeyMigrationManager?.markMigrationAsSkipped()
+            coreCryptoKeyMigrationManager.markMigrationToBytesAsSkipped()
         }
     }
 
-    private func fetchCoreCryptoKey() throws -> Data {
-        let item = CoreCryptoKeychainItem()
-        return try KeychainManager.fetchItem(item)
+    private func fetchCoreCryptoKey(scoped: Bool = true) throws -> Data? {
```
johnxnguyen: suggestion: avoid default arguments so that it's explicitly clear at the callsite.

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -17,126 +17,272 @@
...
+
+            // Mark the rotation as done
+            coreCryptoKeyMigrationManager.markKeyRotationAsDone()
+            logRotation("marked key rotation as done")
+
+        } catch {
+
+            // Check if we need to clean up or rollback, otherwise throw the error
+            switch error {
+            case let CoreCryptoKeyMigrationManagerError.failedToUpdateKey(underlyingError):
```
johnxnguyen: suggestion: you can catch this error, anything else will be rethrown automatically.

```suggestion
        } catch let CoreCryptoKeyMigrationManagerError.failedToUpdateKey(underlyingError) {
            // Check if we need to clean up or rollback, otherwise throw the error
```

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
```diff
@@ -17,126 +17,272 @@
...
+                uniqueKeyId: uniqueKeyId,
+                userID: userID,
+                scoped: scoped
+            )
+            return try KeychainManager.fetchItem(item)
+        } catch KeychainManager.Error.failedToFetchItemFromKeychain(errSecItemNotFound) {
+            return nil
+        } catch {
+            throw error
+        }
```
johnxnguyen: No need to catch and throw, if you omit this it'll be rethrown automatically.

```suggestion
        }
```

wire-ios-data-model/Source/Core Crypto/CoreCryptoProvider.swift
```diff
@@ -229,7 +217,11 @@ public actor CoreCryptoProvider: CoreCryptoProviderProtocol {
     }
 
     func createCoreCrypto() async throws -> SafeCoreCrypto {
-        let coreCryptoKeyProvider = CoreCryptoKeyProvider(coreCryptoKeyMigrationManager: coreCryptoKeyMigrationManager)
+        let coreCryptoKeyProvider = CoreCryptoKeyProvider(
+            coreCryptoKeyMigrationManager: coreCryptoKeyMigrationManager,
+            userID: selfUserID,
+            storage: sharedUserDefaults
+        )
```
johnxnguyen, line 224: question: by doing the migration inside the core crypto provider it means that we may migrate the keys not only in the main app but also in the share and notification extensions?

wire-ios-data-model/Source/Core Crypto/StaleCoreCryptoKeyTracker.swift
```diff
@@ -0,0 +1,67 @@
...
+    func removeKey(id: UUID)
+}
+
+public struct StaleCoreCryptoKeysTracker: StaleCoreCryptoKeysTrackerProtocol {
+
+    private let defaults: UserDefaultsProtocol
+
+    let key = "staleCoreCryptoKeyIds"
+
+    public init(defaults: UserDefaultsProtocol = UserDefaults.standard) {
```
johnxnguyen: question: do we inject the user defaults or use standard? If we use standard... then it won't be shared with other processes, so keys marked as stale in the NSE won't be stale in the main app.

WireDomain/Sources/WireDomain/AppVersionMigration/AppVersionMigrationService.swift
```diff
@@ -68,9 +68,15 @@ public final class AppVersionMigrationService {
...
                 journal.lastCompletedAppVersionMigration = nextMigration.version
-                WireLogger.session.info("Completed migration to version \(nextMigration.version)")
+                WireLogger.session.info(
+                    "Completed migration to version \(nextMigration.version)",
+                    attributes: .safePublic
+                )
             } catch {
-                WireLogger.session.error("Failed migration to version \(nextMigration.version): \(error)")
+                WireLogger.session.error(
+                    "Failed migration to version \(nextMigration.version): \(error)",
```
johnxnguyen, line 77: thought: we should be careful to print errors publicly, 1) is `\(error)` going to provide any useful information?, and 2) will it leak any sensitive information?.

johnxnguyen (commented):
looks good, just have a few comments and questions.

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: not necessary

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: updated [3dd71ad](https://github.com/wireapp/wire-ios/pull/3637/commits/3dd71ad89cadc9edf4040d625e88dd5aea8f9827)

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: wasn't sure if it was worth flooding the logs, but I made it public for now and we can remove it later if needed [3dd71ad](https://github.com/wireapp/wire-ios/pull/3637/commits/3dd71ad89cadc9edf4040d625e88dd5aea8f9827)

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: updated [3dd71ad](https://github.com/wireapp/wire-ios/pull/3637/commits/3dd71ad89cadc9edf4040d625e88dd5aea8f9827)

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: we don't exit on extensions, I passed the shared defaults [3dd71ad](https://github.com/wireapp/wire-ios/pull/3637/commits/3dd71ad89cadc9edf4040d625e88dd5aea8f9827)

wire-ios-data-model/Source/Core Crypto/CoreCryptoProvider.swift
David-Henner: yes

wire-ios-data-model/Source/Core Crypto/StaleCoreCryptoKeyTracker.swift
David-Henner: we inject the shared defaults. I think the default parameter can be removed to make it more explicit though

WireDomain/Sources/WireDomain/AppVersionMigration/AppVersionMigrationService.swift
David-Henner: there may be a risk of leaking sensitive information if the migration doesn't take care to bubble up safe errors

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: [1f0a47f](https://github.com/wireapp/wire-ios/pull/3637/commits/1f0a47fdd9917d071734691e046f356d839e7cfd)

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: [1f0a47f](https://github.com/wireapp/wire-ios/pull/3637/commits/1f0a47fdd9917d071734691e046f356d839e7cfd)

wire-ios-data-model/Source/Core Crypto/CoreCryptoKeyProvider.swift
David-Henner: [1f0a47f](https://github.com/wireapp/wire-ios/pull/3637/commits/1f0a47fdd9917d071734691e046f356d839e7cfd)

johnxnguyen (approved):

PR #3645, opened 2025-09-25
fix: Update the Mention and Link text color - WPB-20472

SyamalaHari:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20472" title="WPB-20472" target="_blank">TaskWPB-20472</a>  Update the Mention and Link text color when the accentColor changed from different clients(web,iOS)
  </td></table>
  <br />
 

### Issue

In current implementation ChatBubble background color in iOS is not updated  immediately when the accent Color is changed from different client (ex:Web ). It is only updated after refresh of the page

Solution: Added observer to listen to the changes of accentColor and this observer is added only once inside the config using a flag. This helps us in not adding observer everytime we run config. 

### Testing

ChatBubble background color should be updated immediately when user changes accent Color from different clients (ex: web)

---

### Checklist

---


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationTextMessageCell.swift
```diff
@@ -67,6 +82,8 @@ final class ConversationTextMessageCell: UIView, ConversationMessageCell, TextVi
 
     weak var delegate: ConversationMessageCellDelegate?
     weak var actionController: ConversationMessageActionController?
+    private var accentColorChangeHandler: AccentColorChangeHandler?
+    private var addedObserver: Bool = false
```
WilhelmOks: I think this name is too generic. It should be clear what observer is added (the color change observer/handler).

Or even better, you could just check if `accentColorChangeHandler != nil` to know if the observer has been added. (And remove `addedObserver`)

SyamalaHari: Checking if accentColorChangeHandler != nil. Thanks for the suggestion 👍

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationTextMessageCell.swift
```diff
@@ -161,7 +177,15 @@ final class ConversationTextMessageCell: UIView, ConversationMessageCell, TextVi
...
         container?.isBubble = isChatBubbleSimpleEnabled
         updateContainerStyle()
-        configureTextColor(forOwnMessage: message?.isSentBySelfUser ?? false)
+        addAccentColorChangeObserver(userSession: object.userSession)
+    }
+
+    private func addAccentColorChangeObserver(userSession: UserSession?) {
+        guard accentColorChangeHandler == nil, let userSession else { return }
+        accentColorChangeHandler = AccentColorChangeHandler
+            .addObserver(userSession: userSession) { [unowned self] _ in
```
caldrian: Would `weak message` work as well? We had some crashes in the past where notification observers were called just a little bit after the UI has been torn down. I'm afraid having an unowned reference to a view might lead to the same issue.
```suggestion
            .addObserver(userSession: userSession) { [weak message] _ in
```

SyamalaHari: Thank you for the review, will do the changes as suggested

johnxnguyen (approved):

WilhelmOks (approved):

netbe:
removing myself as reviewer as this has already been approved

PR #3651, opened 2025-09-26
fix: package dependencies for WireMessaging - WPB-20676

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20676" title="WPB-20676" target="_blank">BugWPB-20676</a>  [iOS] build fails due to missing package dependency
  </td></table>
  <br />
 

### Issue

Under certain circumstances building the project fails with the following error:
Screenshot 2025-09-26 at 11 58 43
This PR adds the missing dependency.
Moreover a more-lightweight CI runner image is chosen.

### Testing

_Describe how to test._

---

### Checklist

---


.github/workflows/_reusable_run_tests.yml
```diff
@@ -42,7 +42,7 @@ env: # https://docs.fastlane.tools/getting-started/ios/setup/
 jobs:
   run-tests:
     name: Run tests
-    runs-on: ghcr.io/cirruslabs/macos-runner:sequoia
+    runs-on: ghcr.io/cirruslabs/macos-sequoia-xcode:16.4
```
netbe: question: when you say lightweight do you mean it boot up quicker?

netbe (commented):

.github/workflows/_reusable_run_tests.yml
caldrian: Could be, yes, since it has only one Xcode version installed. If you consider it too risky, I'll revert it.

WireMessaging/Package.swift
```diff
@@ -27,6 +27,7 @@ let package = Package(
         .target(
             name: "WireMessagingDomain",
             dependencies: [
+                .product(name: "CellsSDK", package: "cells-sdk-swift"),
```
jullianm, line 30: did you make sure the whole wire-ios project is building properly ? 

Asking because If I remember correctly this caused the whole Wire-iOS compilation to fail because WireMessagingDomain is being exposed to that target so was the CellsSDK (when adding it as a dependency) and caused the compilation issues at the main project level.

.github/workflows/_reusable_run_tests.yml
netbe: yes let's revert it please, I would keep the same image as before, Cirrus-CI was recommended this as far as i remember

netbe (approved):
approving counting on the CI change is reverted

WireMessaging/Package.swift
caldrian: Yes, it should work fine.

jullianm (approved):

PR #3657, opened 2025-09-29
fix: mls reset sending - WPB-20693

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20693" title="WPB-20693" target="_blank">TaskWPB-20693</a>  [iOS] Recover from re-initialized MLS conversation
  </td></table>
  <br />
 

### Issue

When on a conversation view where the conversation gets mls reset on another user's client, it's not possible to recover from sending a message.

### Solution

* Reestablish the session for pendingJoin conversation while sending a message.
* **[EDIT]:** adding new status pendingJoinAfterReset so in case the mls reset is not complete even when not on the conversation view, it's possible to fix the conversation in sending a message
  * Include in conversation's list conversations with pendingJoinAfterReset

### Testing

See ticket
---

### Checklist

---

### DIFF
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -931,27 +931,27 @@ public final class MLSService: MLSServiceInterface {
             context: context.notificationContext
         )
         
-        let (conversation, epoch) = await context.perform {
+        let (conversation, epoch, lastGroupID) = await context.perform {
             let conversation = ZMConversation.fetch(with: conversationInfo.qualifiedID.uuid,
                                                     domain: conversationInfo.qualifiedID.domain,
                                                     in: context)
-            return (conversation, conversation?.epoch)
+            return (conversation, conversation?.epoch, conversation?.mlsGroupID)
         }
         
-        guard let conversation else {
+        guard let conversation, let lastGroupID, let epoch else {
             throw MLSServiceError.conversationNotFound
         }
         
         let conversationExists = try await self.conversationExists(
-            groupID: groupID
+            groupID: lastGroupID
         )
 
         let shouldEstablishGroup = epoch == 0 && !conversationExists
 
-        if epoch == 0 {
-            try await internalEstablishPendingGroup(groupID: groupID, pendingGroup: conversation, context: context)
+        if shouldEstablishGroup {
+            try await internalEstablishPendingGroup(groupID: lastGroupID, pendingGroup: conversation, context: context)
         } else {
-            try await joinByExternalCommit(groupID: groupID)
+            try await joinByExternalCommit(groupID: lastGroupID)
         }
         
         await save(context)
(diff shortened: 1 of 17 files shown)


wire-ios-data-model/Source/MLS/MLSServiceInterface.swift
```diff
@@ -482,6 +482,13 @@ public protocol MLSServiceInterface: MLSEncryptionServiceInterface, MLSDecryptio
 
     func startProteusToMLSMigration() async throws
 
+    /// Restablish pending group
```
David-Henner:
```suggestion
    /// Re-establish pending group
```

wire-ios-data-model/Tests/MLS/MLSServiceTests.swift
```diff
@@ -571,6 +571,87 @@ final class MLSServiceTests: ZMConversationTestsBase, MLSServiceDelegate {
...
+        // THEN
+        try XCTAssertCount(mockActionsProvider.syncConversationQualifiedIDContext_Invocations, count: 1)
+        try XCTAssertCount(
+            mockActionsProvider.fetchConversationGroupInfoConversationIdDomainSubgroupTypeContext_Invocations,
+            count: 1
+        )
+        await uiMOC.perform {
+            XCTAssertEqual(conversation.mlsStatus, .ready)
+        }
+    }
```
David-Henner: This test and the one below are very similar. Only the conversation epoch appears to change. Perhaps the code duplication can be avoided.

I'm guessing we want to test that if the epoch is 0 we establish the group, but if the epoch is >0 we join via external commit ? However with how the test is set up we only verify that the status changes from `pendingJoin` to `ready`. Can we test whether the group was established or joined via external commit ?

David-Henner (commented):
looks good, I just have a comment regarding tests

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -915,6 +916,48 @@ public final class MLSService: MLSServiceInterface {
         }
     }
 
+    public func reEstablishPendingGroup(groupID: MLSGroupID) async throws {
```
johnxnguyen, line 919:
```suggestion
    public func reestablishPendingGroup(groupID: MLSGroupID) async throws {
```

johnxnguyen (approved):

wire-ios-data-model/Tests/MLS/MLSServiceTests.swift
netbe: https://github.com/wireapp/wire-ios/pull/3657/commits/201f8eeaa8707a6f7d74b98667c11f4401eec885

David-Henner (approved):

David-Henner (approved):

David-Henner:
checked [5b71855](https://github.com/wireapp/wire-ios/pull/3657/commits/5b718550c4260c43db1b87c58a4d54dead2d32bc)

johnxnguyen (approved):

PR #3667, opened 2025-10-01
fix: disable 4.3.0 migration - WPB-20068

David-Henner:
<table>
<td>
  </td></table>
  <br />
 

### Issue

Disabling the migration because it causes issues decrypting the core crypto database in case of multiple accounts being logged in

---

### Checklist

### DIFF
--- a/wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
+++ b/wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
@@ -1710,7 +1710,10 @@ extension ZMUserSession {
                 journal: journal,
                 sessionManager: sessionManager
             ),
-            AppVersionMigration_4_3_0(coreCryptoProvider: coreCryptoProvider)
+            // Disabling migration 4.3.0 because it causes database decryption
+            // issues when the user has multiple accounts.
+            //
+            // AppVersionMigration_4_3_0(coreCryptoProvider: coreCryptoProvider)
         ]
     }
 
--- a/wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
+++ b/wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
@@ -1709,7 +1709,7 @@ extension ZMUserSession {
                 lastEventIDRepository: lastEventIDRepository,
                 journal: journal,
                 sessionManager: sessionManager
-            ),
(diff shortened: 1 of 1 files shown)


wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1710,7 +1710,10 @@ extension ZMUserSession {
                 journal: journal,
                 sessionManager: sessionManager
             ),
-            AppVersionMigration_4_3_0(coreCryptoProvider: coreCryptoProvider)
+            // Disabling migration 4.3.0 because it causes database decryption
+            // issues when the user has multiple accounts.
+            //
+            // AppVersionMigration_4_3_0(coreCryptoProvider: coreCryptoProvider)
```
johnxnguyen, line 1716: issue: if we re-enable it, it won't run because the app will think it's performed the disabled migration successfully. Wouldn't it be better to remove the migration entirely?

David-Henner: I'm okay with removing it entirely. For 4.7 it's been disabled in a similar way ([here](https://github.com/wireapp/wire-ios/blob/744a91c341cc1f700ebdf61c8a5622e8e4df2409/wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_3_0.swift#L29)). @netbe suggested to disable it and add a comment so he may have an opinion towards keeping it ?

Either way we're not going to re-enable it

netbe: no i didn't think too much about it, removing it completly would make sense. I wonder if we need to keep a record of migration though like we do for database changes.

It could be useful in the future to know what was introduced and removed in case we encounter a case where a user did migration X but not Y

netbe (approved):
Either way looks good to me, but having some sort of trace of what happened is good in my opinion, see my comment

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: ok we can keep it commented out, then if we reintroduce it it'll be as a new version.

johnxnguyen (approved):

PR #3670, opened 2025-10-01
fix: Cells backend URL not properly mapped when multibackend ON - WPB-20737

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20737" title="WPB-20737" target="_blank">BugWPB-20737</a>  [iOS] Wire cells on non production backends is broken
  </td></table>
  <br />
 

### Issue

**Context**: When multibackend is ON and we switch backend to staging (or any other environment), the cells URL always resolve to prod.

**Issue**: When multibackend is ON, we still use the legacy backend environment which is set to production so the cells URL is not properly mapped.

**Solution**: Use the new backend environment if any (when multibackend enabled) otherwise use the legacy one (multibackend disabled)

### Testing

switch backend from the developer menu with multibackend enabled, make sure cells URL is correct

### Checklist

---


wire-ios/Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -107,6 +107,7 @@ final class AppRootRouter {
     func start(launchOptions: LaunchOptions) {
         lastLaunchOptions = launchOptions
         showInitial(launchOptions: launchOptions)
+        showInitial(launchOptions: launchOptions)
```
caldrian: Is this duplicated line a mistake (e.g. from a conflict)?

caldrian (approved):

wire-ios/Wire-iOS/Sources/AppRootRouter.swift
jullianm: yes totally a mistake, thanks for spotting it, don't know how I did that :D

jullianm: fixed thx

jullianm:
open a new PR with the right target branch https://github.com/wireapp/wire-ios/pull/3674

PR #3675, opened 2025-10-01
feat: adopt api v12 (legacy) - WPB-20349

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20349" title="WPB-20349" target="_blank">TaskWPB-20349</a>  [iOS] Adopt apiVersion 12
  </td></table>
  <br />
 

### Issue
In this PR:
- add v12 to WireTransport.APIVersion
- create stubs for new metadata on the POST /assets endpoint
- update the response payload for GET /feature-configs endpoint

---

### Checklist

---

### DIFF
--- a/wire-ios-request-strategy/Sources/Payloads/Conversation/Payload+Conversation.swift
+++ b/wire-ios-request-strategy/Sources/Payloads/Conversation/Payload+Conversation.swift
@@ -139,7 +139,7 @@ extension Payload {
             case .v0, .v1, .v2:
                 self.legacyAccessRole = try container.decodeIfPresent(String.self, forKey: .accessRole)
                 self.accessRoles = try container.decodeIfPresent([String].self, forKey: .accessRoleV2)
-            case .v3, .v4, .v5, .v6, .v7, .v8, .v9, .v10, .v11:
+            case .v3, .v4, .v5, .v6, .v7, .v8, .v9, .v10, .v11, .v12:
 
                 // v3 replaces the field "access_role_v2" with "access_role".
                 // However, since the format of update events does not depend on versioning,
@@ -159,13 +159,13 @@ extension Payload {
             case .v0, .v1, .v2, .v3, .v4:
                 self.cipherSuite = nil
                 self.epochTimestamp = nil
-            case .v5, .v6, .v7, .v8, .v9, .v10, .v11:
+            case .v5, .v6, .v7, .v8, .v9, .v10, .v11, .v12:
                 self.cipherSuite = try container.decodeIfPresent(UInt16.self, forKey: .cipherSuite)
                 self.epochTimestamp = try container.decodeIfPresent(Date.self, forKey: .epochTimestamp)
             }
 
             switch apiVersion {
-            case .v8, .v9, .v10, .v11:
+            case .v8, .v9, .v10, .v11, .v12:
                 self.groupType = try container.decodeIfPresent(ConversationGroupType.self, forKey: .groupType)
                 self.addPermission = try container.decodeIfPresent(ChannelPermission.self, forKey: .addPermission)
                 self.cellsState = try container.decodeIfPresent(CellsState.self, forKey: .cellsState)
@@ -197,7 +197,7 @@ extension Payload {
             case .v0, .v1, .v2:
                 try container.encodeIfPresent(legacyAccessRole, forKey: .accessRole)
                 try container.encodeIfPresent(accessRoles, forKey: .accessRoleV2)
-            case .v3, .v4, .v5, .v6, .v7, .v8, .v9, .v10, .v11:
+            case .v3, .v4, .v5, .v6, .v7, .v8, .v9, .v10, .v11, .v12:
                 if legacyAccessRole == nil {
                     try container.encodeIfPresent(accessRoles, forKey: .accessRole)
                 } else {
@@ -207,7 +207,7 @@ extension Payload {
             }
 
             switch apiVersion {
-            case .v8, .v9, .v10, .v11:
+            case .v8, .v9, .v10, .v11, .v12:
                 try container.encodeIfPresent(groupType, forKey: .groupType)
                 try container.encodeIfPresent(addPermission, forKey: .addPermission)
                 try container.encodeIfPresent(cellsState, forKey: .cellsState)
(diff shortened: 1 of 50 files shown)


wire-ios-request-strategy/Sources/Payloads/Conversation/Payload+Conversation.swift
```diff
@@ -139,7 +139,7 @@ extension Payload {
             case .v0, .v1, .v2:
                 self.legacyAccessRole = try container.decodeIfPresent(String.self, forKey: .accessRole)
                 self.accessRoles = try container.decodeIfPresent([String].self, forKey: .accessRoleV2)
-            case .v3, .v4, .v5, .v6, .v7, .v8, .v9, .v10, .v11:
+            case .v3, .v4, .v5, .v6, .v7, .v8, .v9, .v10, .v11, .v12:
```
netbe, line 142: **thought:** I wonder if we could write this simpler with `APIVersion.andNextVersions and `APIVersion.allCasesUpTo` which when we add a new apiversion will reduce our changes

Of course it would mean to deduplicate APIVersion declaration

wire-ios-request-strategy/Sources/Request Strategies/Feature configurations/Payload/FeatureConfigsPayload.swift
```diff
@@ -50,6 +50,7 @@ struct FeatureConfigsPayloadAPIV6: Decodable {
     let channels: FeatureStatusWithConfig<Feature.Channels.Config>?
     let consumableNotifications: FeatureStatus?
     let chatBubbles: FeatureStatus?
+    let assetAuditLog: FeatureStatus?
```
netbe: question: do we have to implement it here? As I understood it would be sufficient to implement it only in WireNetwork/ WireDomain. We should stop adding new fields in legacy.

It would make sense to remove the usage of FeatureConfig

wire-ios/Wire-iOS Tests/ReferenceImages/CallParticipantsListViewControllerTests/testCallParticipants_ConnectingState_Dark.1.png
netbe, line 1: question: why is this changed?

wire-ios/Wire-iOS/Sources/UserInterface/Components/AvatarImageView.swift
```diff
@@ -81,7 +82,7 @@ class AvatarImageView: UIView {
         imageView.isOpaque = false
         container.backgroundColor = .yellow
 
-        initialsLabel.textColor = .white
+        initialsLabel.textColor = SemanticColors.Button.textPrimaryEnabled
```
netbe, line 85: question: why is this changed?

netbe (commented):
There are some changes that don't seem to belong to here (screenshots)

Rest looks good

wire-ios-request-strategy/Sources/Payloads/Conversation/Payload+Conversation.swift
johnxnguyen: While working on these changes I was frustrated by the amount of mindless changes required... more so WireNetwork. Certainly we should find a better approach but I'm not sure what at the moment. 

I think though what you're proposing makes sense... but we would probably achieve it with an open ended range:

```swift
case .v3...:
  // From v3 up to v12
```

wire-ios-request-strategy/Sources/Request Strategies/Feature configurations/Payload/FeatureConfigsPayload.swift
johnxnguyen: This would still be used in cases where old sync is still active, so until we remove old sync I think we'll have to support this code path.

wire-ios/Wire-iOS Tests/ReferenceImages/CallParticipantsListViewControllerTests/testCallParticipants_ConnectingState_Dark.1.png
johnxnguyen: Not sure... probably I branched from develop then re-targeted my branch. I'll try a rebase.

johnxnguyen: Rebase fixed it

wire-ios/Wire-iOS/Sources/UserInterface/Components/AvatarImageView.swift
johnxnguyen: Unintended, rebase fixed it

wire-ios-request-strategy/Sources/Request Strategies/Feature configurations/Payload/FeatureConfigsPayload.swift
netbe: ok but it can't be a requirement, no? you would not expect even if running old sync recieve this featureflag unless backend backports it, which I doubt they do

netbe (approved):
looks good, left a reply. I don't think we need to take care of FeatureConfig in legacy, happy to discuss it

wire-ios-request-strategy/Sources/Request Strategies/Feature configurations/Payload/FeatureConfigsPayload.swift
johnxnguyen: Right that makes sense. I'll remove it

jullianm (approved):

wire-ios-request-strategy/Sources/Payloads/Conversation/Payload+Conversation.swift
johnxnguyen: ok so if we do this the compiler doesn't recognize the switch as being exhaustive. I'll leave it like this for now until we find a better solution.

wire-ios-request-strategy/Sources/Request Strategies/Feature configurations/Payload/FeatureConfigsPayload.swift
johnxnguyen: done

PR #3678, opened 2025-10-01
fix: voiceover not reading message content - WPB-17370

SyamalaHari:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17370" title="WPB-17370" target="_blank">BugWPB-17370</a>  [iOS] VoiceOver not reading message content 
  </td></table>
  <br />
 

### Issue

Issue: On the iOS version of the app, VoiceOver does not read or detect the message content in conversations. While it correctly reads other elements such as the timestamp and direct reply content, it consistently skips the actual message body.
Solution: Accessibility is added to cell itself but we are adding message text to container view. We have already disabled Cell's Accessibility by setting isAccessibilityElement to false, hence setting accessibilityLabel will not have any effect, hence removed it. Now we are setting accessibility label of container view to message text. 

### Testing

VoiceOver should read the full content of each message aloud when focused.

---

### Checklist

---


netbe (approved):
LGTM;)

OlivellaO (approved):

PR #3694, opened 2025-10-06
fix(backup-performance): fetching / updating conversations - WPB-20804

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20804" title="WPB-20804" target="_blank">TaskWPB-20804</a>  [iOS] Backup performance :: Updating `lastReadServerTimeStamp`
  </td></table>
  <br />
 

### Issue 

Using Xcode’s Instruments, I was able to measure backup restoration performance. Upon analyzing two samples of an import (one with 4k messages and another with 10k) I found out that **38%** of the time is spent on `[ZMConversation setLastReadServerTimeStamp:]`. Calling this method triggers a recalculation of unread messages that involves performing a fetch request, which is expensive. 

### Cause

For each message being restored:
- The conversation is being fetched (1st expensive operation)
- The `lastReadTimeStamp` is being set (2nd expensive operation)

### Solution

For each batch of messages, fetch a batch of conversations first. Then, process the messages using the cached conversations. Then update the timestamp for each conversation.

---

### Checklist

### DIFF
--- a/WireBackup/Sources/WireBackup/Protocols/BackupLocalStoreProtocol.swift
+++ b/WireBackup/Sources/WireBackup/Protocols/BackupLocalStoreProtocol.swift
@@ -52,7 +52,7 @@ public protocol BackupLocalStoreProtocol: Sendable {
     /// Returns all messages stored in the local database, including deleted ones.
     func fetchAllMessages() -> AsyncThrowingStream<MessageBackupModel, any Error>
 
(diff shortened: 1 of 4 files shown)


caldrian (approved):

caldrian (approved):
great job!

KaterinaWire (approved):

PR #3696, opened 2025-10-07
fix: tapping buttons in composite messages (e.g. polls) - WPB-19793

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19793" title="WPB-19793" target="_blank">BugWPB-19793</a>  [iOS] Selecting option in poll option doesn't work
  </td></table>
  <br />
 

### Issue

A performance improvement #2921 earlier this year broke composite messages.
Some dangling Core Data objects from a temporary background context remain in the view/model hierarchy.
So when a button of a composite message is tapped, it cannot be handled properly.

This PR adds a workaround which forces the recreation of conversation message views if the message is a composite message.

### Testing

Create a poll in a Proteus conversation and try to vote.

---

### Checklist

---


wire-ios-data-model/Source/Model/Message/Composite/CompositeMessageItemContent.swift
```diff
@@ -116,19 +116,17 @@ extension CompositeMessageItemContent: ButtonMessageData {
...
-        moc.performGroupedBlock { [weak self] in
-            guard let self else { return }
             let buttonState = buttonState ??
-                ButtonState.insert(with: buttonId, message: parentMessage, inContext: moc)
+                ButtonState.insert(with: buttonId, message: parentMessage, inContext: context)
             parentMessage.buttonStates?.resetExpired()
             guard parentMessage.isSenderInConversation else {
                 buttonState.isExpired = true
-                moc.saveOrRollback()
+                context.saveOrRollback()
```
caldrian, line 130: Two things done here:
1. rename `moc` to `context`
2. move accessing `nonce` into the `context.perform` closure, so that it works with any context

wire-ios/Wire-iOS/Wire-Info.plist
```diff
@@ -137,6 +137,8 @@
 		<string>remote-notification</string>
 		<string>fetch</string>
 	</array>
+	<key>UIDesignRequiresCompatibility</key>
+	<true/>
```
caldrian, line 141: This change allows using Xcode 26.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
```diff
@@ -199,7 +206,7 @@ final class ConversationTableViewDataSource: NSObject {
 
                     sectionController.selfUser = selfUserOnMainThread
 
-                    if sectionController.context != context || forceRecalculate {
+                    if sectionController.context != context || forceRecalculate || recreateCellDescriptions {
                         sectionController.recreateCellDescriptions(in: context)
```
caldrian, line 210: `calculateSections` fetches messages on a temporary background context, populates views (and related objects) and later replaces the message with instances from the main context. 
However, without calling `sectionController.recreateCellDescriptions` some dangling Core Data objects are not replaced.

Since I don't want to degrade performance again, I'm forcing the update only for composite messages.

wire-ios/Wire-iOS/Wire-Info.plist
netbe: suggestion: while at it, maybe worth to do the update then https://wearezeta.atlassian.net/browse/WPB-20169

wire-ios-data-model/Source/Model/Message/Composite/CompositeMessageItemContent.swift
netbe: question: was it a background context before that had no parent? or just no automaticallySaveToParent?

netbe (approved):
Fix looks good, I left some comments please have a look.

wire-ios-data-model/Source/Model/Message/Composite/CompositeMessageItemContent.swift
caldrian: Yes, some views or models belonging to views have core data entities set which have been fetched on a background context.

jullianm (approved):

PR #3711, opened 2025-10-08
fix(backup-performance): message creation - WPB-20905

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20905" title="WPB-20905" target="_blank">TaskWPB-20905</a>  [iOS] Backup performance :: creating messages
  </td></table>
  <br />
 

### Issue

Using the time profiler with a backup sample of ~15k messages, I observed that 34% of the time importing a backup is spent doing a fetch request to fetch messages we’re importing. (in `MessageLocalStore.fetchOrCreateClientMessage`) 

However, since we’re already checking if messages exist before creating them (as part of backup import), we don’t need to try to fetch them, we just need to create them.

### Solution

Add methods:
- `MessageLocalStore.createClientMessage`
- `MessageLocalStore.createAssetClientMessage`
 
And use these when processing messages from the backup in `ConversationProtobuMessageProcessor.processProtobufMessage`

### Result 

Measured the performance of importing a backup of ~15k messages, using a baseline build with no performance improvements and another build with the improvements from this PR (https://github.com/wireapp/wire-ios/pull/3694) + the current PR.

| Metric                           | Baseline  | Optimized | Times Faster         |
| -------------------------------- | --------- | --------- | -------------------- |
| **Total time**                   | 8.90 min  | 3.81 min  | **≈ 2.34×**          |
| **Page processing** (1k messages)             | 33.37 s   | 14.28 s   | **≈ 2.34×**          |
| **Message processing** (individual average)          | 19.81 ms  | 13.33 ms  | **≈ 1.49×**          |
| **Restoring last read**          | 183.6 s   | 12.23 s   | **≈ 15.0×**          |

**Note:** The change from this PR (https://github.com/wireapp/wire-ios/pull/3694) was between **1,2 - 1,31** times faster 

---

### Checklist

### DIFF
--- a/WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProtobufMessageProcessor.swift
+++ b/WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProtobufMessageProcessor.swift
@@ -241,12 +241,21 @@ public struct ConversationProtobufMessageProcessor: ConversationProtobufMessageP
     ) async throws {
         let (assetClientMessage, isNew): (ZMAssetClientMessage, Bool)
         do {
-            (assetClientMessage, isNew) = try await messageLocalStore.fetchOrCreateAssetClientMessage(
-                id: message.messageID,
-                conversation: conversation,
-                sender: (sender.id, sender.domain, sender.clientID),
-                date: date
-            )
+            if isProcessingBackup {
+                (assetClientMessage, isNew) = (try await messageLocalStore.createAssetClientMessage(
+                    id: message.messageID,
+                    conversation: conversation,
+                    sender: (sender.id, sender.domain, sender.clientID),
+                    date: date
+                ), true)
+            } else {
+                (assetClientMessage, isNew) = try await messageLocalStore.fetchOrCreateAssetClientMessage(
+                    id: message.messageID,
+                    conversation: conversation,
+                    sender: (sender.id, sender.domain, sender.clientID),
+                    date: date
+                )
+            }
         } catch let MessageLocalStore.Failure.invalidInsertion(reason: reason) {
             return WireLogger.eventProcessing.warn(
                 "failed to process asset message, dropping. Reason: \(reason)",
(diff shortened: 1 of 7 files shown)


WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProtobufMessageProcessor.swift
```diff
@@ -26,15 +26,18 @@ public struct ConversationProtobufMessageProcessor: ConversationProtobufMessageP
...
     let conversationLocalStore: any ConversationLocalStoreProtocol
     let userLocalStore: any UserLocalStoreProtocol
+    let isProcessingBackup: Bool
 
     public init(
         messageLocalStore: any MessageLocalStoreProtocol,
         conversationLocalStore: any ConversationLocalStoreProtocol,
-        userLocalStore: any UserLocalStoreProtocol
+        userLocalStore: any UserLocalStoreProtocol,
+        isProcessingBackup: Bool
```
caldrian, line 35: Could you please make the purpose of this flag somehow more clear?
E.g. leave a doc comment somewhere when the flag should be set to `true` or `false`.

Also we could rename it somehow, like
skipFetching...
skipChecking...
alwaysCreate...
...?
or similar.

caldrian (approved):

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProtobufMessageProcessor.swift
samwyndham: +1

samwyndham (approved):
LGTM. Nice performance improvement.

samwyndham:
> However, since we’re already checking if messages exist before creating them (as part of backup import), we don’t need to try to fetch them, we just need to create them.

question: If this checking somehow fails, what would happen if we attempt to create a message when it already exists?

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationProtobufMessageProcessor.swift
David-Henner: good point ! I was considering introducing a configuration object with flags such as `skipFetching` as I may need to introduce more conditions

David-Henner:
> > However, since we’re already checking if messages exist before creating them (as part of backup import), we don’t need to try to fetch them, we just need to create them.
> 
> question: If this checking somehow fails, what would happen if we attempt to create a message when it already exists?

Since there's no uniqueness constraint on the message entity in core data, it would create a duplicate entry in the db

PR #3729, opened 2025-10-13
fix: wire cells attachment alignment - WPB-21024

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21024" title="WPB-21024" target="_blank">BugWPB-21024</a>  [iOS] Fix attached documents alignment with chat bubbles
  </td></table>
  <br />
 

### Issue

This PR fixes wire cells attachment alignment when simple bubbles is enabled so that attachments for the self user are right aligned and for other users left aligned.

| Before | After |
| --- | --- |
| IMG_0385 | IMG_0386 |

### Testing

1. Ensure simplified chat bubbles is enabled.
2. Navigate to a wire cells enabled conversation.
3. Send a message with multiple files attached.
4. Have another user respond with a message with multiple files attached.
5. Note alignment

---

### Checklist


WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsAttachmentsPreviewView.swift
```diff
@@ -37,7 +37,7 @@ package struct WireCellsAttachmentsPreviewView: View {
                 itemRow(index: index)
             }
         }
-        .frame(maxWidth: .infinity, alignment: .leading)
+        .frame(maxWidth: .infinity, alignment: viewModel.alignment == .leading ? .leading : .trailing)
```
jullianm, line 40: question: is the ternary condition necessary ? can't we just do `viewModel.alignment` since we're working with the same type

jullianm (approved):
Looks good, left a (non blocking) question

caldrian (approved):

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsAttachmentsPreviewView.swift
samwyndham: The type is actually different. The input is a `HorizontalAlignment` instead of an `Alignment`. I might change this though. 🤔

PR #3731, opened 2025-10-13
feat: add Files tab on iPhone and iPad (UI only) - WPB-20934

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20934" title="WPB-20934" target="_blank">TaskWPB-20934</a>  [iOS] Add "Files" tab with a search screen
  </td></table>
  <br />
 

### Issue

This PR adds the Files tab, as per design, ensuring the UI is correctly displayed on both iPhone and iPad.
This PR is about UI only, there will be a separated ticket to implement the "All files" view.

**iPhone**
Capture d’écran   2025-10-13 à 16 53 40

**iPad**

Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-10-13 at 17 13 39

### Testing

- Updated snapshot tests (split controller, tab bar controller, side bar)
- Tested that it behaves as expected on iPad

### Checklist

---


WireUI/Sources/WireMainNavigationUI/Coordinator/MainCoordinator.swift
```diff
@@ -410,6 +451,10 @@ public final class MainCoordinator<Dependencies>: NSObject, MainCoordinatorProto
             let mainMenuItem = MainSidebarMenuItem(conversationListUI.conversationFilter)
             sidebar.selectedMenuItem = .init(mainMenuItem)
 
+        case .files:
+
+            sidebar.selectedMenuItem = .init(.files)
```
caldrian, line 455:
```suggestion
        case .files:
            sidebar.selectedMenuItem = .init(.files)
```

caldrian (approved):
Great job!

KaterinaWire:
@jullianm Do you think we should replace all the snapshots since we hide the tab by the developer flag? I've added a Meeting tab that is also hidden behind the developer flag, but t's not going to be on production yet. Maybe we should keep snapshot tests and add a new when the flag is on. What do you think?

jullianm:
@KaterinaWire when the feature is live we'll need to open a PR only to update snapshots, we can avoid that by doing it as part of the PR its related to even if the feature is not live, not to mention we may forget to update the snapshots when the feature is live as it could be weeks or months later. I don't have a strong opinion on this just to avoid extra work if we can

KaterinaWire:
> @KaterinaWire when the feature is live we'll need to open a PR only to update snapshots, we can avoid that by doing it as part of the PR its related to even if the feature is not live, not to mention we may forget to update the snapshots when the feature is live as it could be weeks or months later. I don't have a strong opinion on this just to avoid extra work if we can

@jullianm  We can have snapshots for both states of the develop flag. The problem is that I will need than to update the same snapshots and we will have "Conversations", "Files", "Meetings", "Settings". But it's not the current production state and won't be for the next release for sure. I believe we should keep tests that represents the current production or both states depends on the developer flags

jullianm:
@KaterinaWire as discussed, kept legacy snapshots and added a single snapshot for Files tabs here https://github.com/wireapp/wire-ios/pull/3731/commits/77bde9766b7852f73c62f622e209e12a865d2e45

KaterinaWire (approved):

PR #3764, opened 2025-10-20
feat: show un-established mls groups - WPB-21161

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21161" title="WPB-21161" target="_blank">TaskWPB-21161</a>  [iOS]  Display unestablished channels in conversation list
  </td></table>
  <br />
 

### Issue
Soon it'll be possible that a team admin can manage channels from the team admin panel despite not being part of the channel themself. This means that the responsibility to establish the mls group and add participants falls on the participants. For performance reasons, the group will be established by the user who first tries to send a message in the group.

Currently, clients don't show conversations with unestablished mls groups to the user. This PR changes this, so that the user will see conversations in which the underlying mls group is not yet established, so that they have the opportunity to try send a message and then establish the group.

### Testing
N/A

---

### Checklist

---

### DIFF
--- a/wire-ios-data-model/Source/Model/Conversation/ConversationPredicateFactory.swift
+++ b/wire-ios-data-model/Source/Model/Conversation/ConversationPredicateFactory.swift
@@ -188,14 +188,8 @@ public final class ConversationPredicateFactory: NSObject {
 
         // MLS
         let isMLS = NSPredicate(format: "\(ZMConversation.messageProtocolKey) == \(MessageProtocol.mls.int16Value)")
-        let isMLSStatusReady = NSPredicate(format: "\(ZMConversation.mlsStatusKey) == \(MLSGroupStatus.ready.rawValue)")
-        let isMLSStatusPendingJoinAfterReset =
-            NSPredicate(format: "\(ZMConversation.mlsStatusKey) == \(MLSGroupStatus.pendingJoinAfterReset.rawValue)")
 
-        let isValidMLSStatus = NSPredicate.any(of: [isMLSStatusReady, isMLSStatusPendingJoinAfterReset])
-        let isMLSAndReady = NSPredicate.all(of: [isMLS, isValidMLSStatus])
-
-        return .any(of: [isProteus, isMixed, isMLSAndReady])
+        return .any(of: [isProteus, isMixed, isMLS])
     }
 
     private func isValidConversation() -> NSPredicate {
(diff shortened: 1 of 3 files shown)


wire-ios-data-model/Source/Model/Conversation/ConversationPredicateFactory.swift
```diff
@@ -188,14 +188,8 @@ public final class ConversationPredicateFactory: NSObject {
...
         let isMLS = NSPredicate(format: "\(ZMConversation.messageProtocolKey) == \(MessageProtocol.mls.int16Value)")
-        let isMLSStatusReady = NSPredicate(format: "\(ZMConversation.mlsStatusKey) == \(MLSGroupStatus.ready.rawValue)")
-        let isMLSStatusPendingJoinAfterReset =
-            NSPredicate(format: "\(ZMConversation.mlsStatusKey) == \(MLSGroupStatus.pendingJoinAfterReset.rawValue)")
 
-        let isValidMLSStatus = NSPredicate.any(of: [isMLSStatusReady, isMLSStatusPendingJoinAfterReset])
-        let isMLSAndReady = NSPredicate.all(of: [isMLS, isValidMLSStatus])
-
-        return .any(of: [isProteus, isMixed, isMLSAndReady])
+        return .any(of: [isProteus, isMixed, isMLS])
```
netbe, line 192: question: is outOfSync mls conv should be shown too?

netbe (approved):
LGTM;) @johnxnguyen I assume this is aligned with other platforms too, right?

johnxnguyen:
> LGTM;) @johnxnguyen I assume this is aligned with other platforms too, right?

@netbe yes it should be, there's the same ticket for all platforms.

wire-ios-data-model/Source/Model/Conversation/ConversationPredicateFactory.swift
johnxnguyen: Do you mean the `pendingJoinAfterReset` status? I would say yes, because if the user tries to send a message in a conversation with that status (now that they can see it), doing so would [trigger the reestablishment of the group](https://github.com/wireapp/wire-ios/blob/f26eb8fcc9e9f3e5f2ea5aa5e5a414fa18922d2e/wire-ios-request-strategy/Sources/Message%20Sending/MessageSender.swift#L394). What do you think?

netbe: right pendingJoinAfterReset was just a workaround to filter these conv, so no issue.I just wonder about outOfSync state when this happen, but probably fine too

jullianm (approved):
👍

PR #3768, opened 2025-10-21
fix: gif picker leaving blank area in place of keyboard - WPB-20252

WilhelmOks:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20252" title="WPB-20252" target="_blank">BugWPB-20252</a>  [iOS][iOS26] White space shows in place of keyboard on opening gif/camera/markers 
  </td></table>
  <br />
 

### Issue

The virtual keyboard leaves a blank area when the keyboard is dismissed (by unfocussing the message text field) right before opening the gif picker.
The solution is to remove the unfocussing (resignFirstResponder).

### Testing

On iOS 26, open a conversation and tap into the text message input field so that the virtual keyboard appears.
Then tap on the gif button to open the gif/giphy picker view.
The keyboard should disappear automatically.
Dismiss/cancel the gif picker.
There should be no blank space where the keyboard was.
Instead, the keyboard should be visible again.

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController/ConversationInputBarViewController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController/ConversationInputBarViewController.swift
@@ -769,8 +769,8 @@ final class ConversationInputBarViewController: UIViewController,
             let conversation = conversation as? ZMConversation
         else { return }
 
-        presentMLSPrivacyWarningIfNeeded {
-            self.showGiphy(for: conversation)
+        presentMLSPrivacyWarningIfNeeded { [weak self] in
+            self?.showGiphy(for: conversation)
         }
     }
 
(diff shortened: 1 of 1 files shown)


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController/ConversationInputBarViewController.swift
```diff
@@ -769,8 +769,8 @@ final class ConversationInputBarViewController: UIViewController,
             let conversation = conversation as? ZMConversation
         else { return }
 
-        presentMLSPrivacyWarningIfNeeded {
-            self.showGiphy(for: conversation)
+        presentMLSPrivacyWarningIfNeeded { [weak self] in
```
WilhelmOks, line 772: Note for reviewers:
I just added this defensively to avoid possible memory leaks.

netbe (approved):
LGTM;)

samwyndham (approved):
LGTM too!

PR #3774, opened 2025-10-22
feat: cells image previews in conversations - WPB-16312

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16312" title="WPB-16312" target="_blank">TaskWPB-16312</a>  [iOS] Small image previews in a conversation message
  </td></table>
  <br />
 

### Issue

This PR introduces small and large image previews into wire cells conversations. This required adding some missing plumbing to ensure that file metadata is forwarded from multipart protobuf messages to the view layer.

IMG_0408

### Testing

1. Navigate to a cells enabled conversation
2. Send various files which include images. 
3. Test that the image previews look OK. For some formats no image preview will be available (heic).
4. Test both messages with single attachment (large image preview) and multiple attachments (small image preview)
5. Test that tapping on the previews opens the correct files.

---

### Checklist

- [x] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [x] Description is filled and free of optional paragraphs.
- [x] Adds/updates automated tests.

PR #3785, opened 2025-10-24
feat: large document previews - WPB-21022

OlivellaO:
### Issue

What was done in this PR:

1. Adding new View for large Documents preview files.
2. Extract header part from WireCellsDocumentAttachmentPreview into a new reusable View.

### Testing

Using a version that has Wire Cells enabled send a single document file on a chat.
When single document is shared, large document preview should be shown.

UI test can't be done, no way of mocking AsyncImage.

---

### Checklist

---


WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsLargeDocumentPreviewView.swift
```diff
@@ -0,0 +1,136 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import SwiftUI
+import WireDesign
+import WireFoundation
+
+struct WireCellsLargeDocumentPreviewView: View {
+    enum Constants {
```
WilhelmOks: I don't like the `Constants` wrapper and suggest to put the constants directly into the View instead.

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsLargeDocumentPreviewView.swift
```diff
@@ -0,0 +1,136 @@
...
+                WireCellsDocumentHeaderView(
+                    headerIcon: headerIcon,
+                    headerText: headerText,
+                    labelText: labelText,
+                    progress: progress,
+                    isError: isError
+                )
+                .background(ColorTheme.Backgrounds.surfaceVariant.color)
+                .frame(height: 74)
+                .frame(maxWidth: .infinity)
```
WilhelmOks, line 52: A blank line would be good to separate WireCellsDocumentHeaderView from previewContainer

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsLargeDocumentPreviewView.swift
```diff
@@ -0,0 +1,136 @@
...
+                previewContainer {
+                    if url != nil {
+                        asyncImage(url: url!)
```
WilhelmOks: It's better to use "if let" in this case. Avoid explicit unwraps.

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsLargeDocumentPreviewView.swift
```diff
@@ -0,0 +1,136 @@
...
+                    } else {
+                        errorView(text: "Invalid URL")
```
WilhelmOks: This can be presented to the end user and should be localized

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsLargeDocumentPreviewView.swift
```diff
@@ -0,0 +1,136 @@
...
+            .clipped()
+    }
+
+    @ViewBuilder
+    private func loadingView() -> some View {
+        ZStack {
+            Color(ColorTheme.Backdrop.background)
+            VStack {
+                ProgressView()
+                Text("Loading...")
```
WilhelmOks: should be localized

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsLargeDocumentPreviewView.swift
```diff
@@ -0,0 +1,136 @@
...
+        AsyncImage(url: url) { phase in
+            switch phase {
+            case .empty:
+                loadingView()
+            case .success(let image):
+                image
+                    .resizable()
+                    .scaledToFill()
+            case .failure:
+                errorView(text: "Unable to display preview")
```
WilhelmOks: should be localized

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsLargeDocumentPreviewView.swift
```diff
@@ -0,0 +1,136 @@
...
+            }
+        }
+    }
+
+    @ViewBuilder
+    private func previewContainer(@ViewBuilder content: () -> some View)
+        -> some View
+    {
+        Color.clear
+            .aspectRatio(CGFloat(8.0 / 3.0), contentMode: .fit)
```
WilhelmOks: I think the number 8/3 should be given a constant and a name, because it will be reused in many places, probably.

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Previews/WireCellsDocumentHeaderView.swift
```diff
@@ -0,0 +1,87 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import SwiftUI
+import WireDesign
+import WireFoundation
+
+struct WireCellsDocumentHeaderView: View {
+    enum Constants {
+        static let errorColor = ColorTheme.Base.error.color
```
WilhelmOks, line 25: I don't like the `Constants` wrapper and suggest to put the constants directly into the View instead.

WilhelmOks (changes requested):

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsLargeDocumentPreviewView.swift
OlivellaO: Not sure it will be reuse in other places. Videos comes with its aspectRatio for example.
But it could be a constant on the View and not a "magic number" somewhere on the code, i will update it.

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Previews/WireCellsDocumentHeaderView.swift
samwyndham: I think here @OlivellaO is trying to fit with the current style. I would consider keeping it as that is what is happening in other places.

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsLargeDocumentPreviewView.swift
samwyndham: You can put the localized string in `/WireMessaging/Sources/WireMessagingUI/Resources/Localization/en.lproj/Localizable.strings` then access it using the generated `L10n`. Something like `L10n.Localizable.Conversation.Message.Attachment.<something>`

samwyndham (commented):
Nice work! I will approve after @WilhelmOks comments are addressed

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsLargeDocumentPreviewView.swift
samwyndham: As this is a pattern used elsewhere I think you should keep it personally.

OlivellaO: I agree with Wilhelm here. There is any reason why we should keep it like this?

samwyndham: Either way it is subjective. My reason for keeping it is I see another 22 uses of the pattern in the project and during review other reviewers have asked me to move things to a Constants enum. But I don't feel strongly about the pattern itself so will leave it up to you

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Previews/WireCellsDocumentHeaderView.swift
OlivellaO: Yes, i did it like this because this is a copy & paste from the header part of the Previews. Cose was not modified, just extracted to a new struct.

But i agree with Wilhelm, if there is no point on having this enum here when we can have just a constant defined with that value why should we keep it?

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsLargeDocumentPreviewView.swift
samwyndham: issue: I just noticed that no URL should not be considered an error. It is a valid state. The localized string should be : "conversation.message.attachment.previewNotAvailable" = "Unable to display preview"; which already exists in my other PR: https://github.com/wireapp/wire-ios/pull/3774/ and is used like:

```swift
            Text(noPreviewMessage)
                .wireTextStyle(.subline1)
                .foregroundColor(ColorTheme.Backgrounds.surface.color)
                .multilineTextAlignment(.center)
                .padding()
``` 

There shouldn't be an `Image(systemName: "exclamationmark.triangle")` in this case because it is an expected state. The error should be reserved more for when `isError` is true - this will happen when a file itself can't be downloaded, not when the preview can't be downloaded.

samwyndham: I'm not actually sure we ever need the `Image(systemName: "exclamationmark.triangle")`. Did you see it in a design somewhere?

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsLargeDocumentPreviewView.swift
```diff
@@ -0,0 +1,135 @@
...
+            }
+        }
+    }
+
+    @ViewBuilder
+    private func errorView(text: String) -> some View {
+        ZStack {
+            Color(ColorTheme.Backdrop.background)
+            VStack {
+                Image(systemName: "exclamationmark.triangle")
```
samwyndham: issue: I don't think we want Image(systemName: "exclamationmark.triangle") here.

There are 2 kinds of errors that these preview views can have:

1. Failing to download a preview image.
2. Failing to download a actual file.

In the first case the error message should be "Unable to display preview" like in my other comment. In the second case (which we will know because`isError` will be true, for now lets use.   "Unable to download file". I will try to find out the real version.

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsLargeDocumentPreviewView.swift
OlivellaO: No, i didn't. I just did a UI to show the error, i will update it now with just a text.

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsLargeDocumentPreviewView.swift
OlivellaO: Then we have 3 kind of errors:
- URL is not provided
- Preview image can't be loaded (but URL was provided)
- File unable to download

So for the first 2 use cases I should use the "Unable to display preview" (those are the ones already implemented) and then i need to add an extra implementation for when `isError = true` so it will show "Unable to download file" instead.

@samwyndham In case i don't get an URL or preview can be loaded AND i get `isError = true` what should i show?

OlivellaO: I've updated the code to use isError (now called downloadError for better understanding).

Can you check it and let me know what you think about it? If it's ok you can resolve this conversation 😃

samwyndham: I think I need to clarify this behavior for all preview types with the designer which will take some time. Lets go with your implementation for now.

WilhelmOks (approved):

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsLargeDocumentPreviewView.swift
```diff
@@ -0,0 +1,134 @@
...
+            VStack {
+                WireCellsDocumentHeaderView(
+                    headerIcon: headerIcon,
+                    headerText: headerText,
+                    labelText: labelText,
+                    progress: progress,
+                    isError: downloadError
+                )
+                .background(ColorTheme.Backgrounds.surfaceVariant.color)
+                .frame(height: 74) // This might break the UI if text font is too big
```
samwyndham, line 51: Good point with this comment. Once the view is integrated we can play around with it. Perhaps we can use `idealHeight` or something else instead.

samwyndham (approved):
Nice work!

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsLargeDocumentPreviewView.swift
OlivellaO: I agree. I didn't want to play with this on this task since this is use by all the previews right now. Once we have al the large previews we can play with it and try to remove the hardcoded height 😃

WilhelmOks (approved):

PR #3792, opened 2025-10-27
fix: invisible groups - WPB-20123

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20123" title="WPB-20123" target="_blank">BugWPB-20123</a>  [iOS] Invisible groups, channels and 1:1 chats with functioning push notifications
  </td></table>
  <br />
 

### Issue

Sometimes conversations are not shown within the app. Some conversations were found not updated with data from backend for example having a invalid groupType hence not displayed.

Right now, the ConversationsRequestStrategy with the updateSync (KeyPathObjectSync) sync the conversation with the backend when the `NeedsToBeUpdatedFromBackend` property is set to true, but it does not wait for the request to finish to turn off the flag, so if the request fails or the app crashes, we'll never retry the request.

Solution:

* Using WorkAgent in #3800, we remove old KeyPathObjectSync
* `ConversationUpdatesGenerator` creates `UpdateConversationItem` workItems for the WorkAgent.
* remove pullConversation from the mlsWelcomeMessage event processing so processing events are without extra network calls.

Other changes:
* add logs for processing events

### Testing

1. Create a group
2. See ticket being created and executed

---

### Checklist

---


WireDomain/Sources/WireDomain/WorkAgent/Worker/SyncConversationsWorker.swift
```diff
@@ -0,0 +1,68 @@
...
+        kWorkerID
+    }
+    
+    private let repository: ConversationRepositoryProtocol
+
+    public init(repository: ConversationRepositoryProtocol) {
+        self.repository = repository
+    }
+
+    func performWork(for ticket: UpdateConversationTicket) async throws {
```
netbe: here the worker do the sync part with backend

WireDomain/Sources/WireDomain/Synchronization/ConversationUpdatesGenerator.swift
```diff
@@ -0,0 +1,106 @@
...
+        self.context = context
+        self.onConversationUpdated = onConversationUpdated
+        super.init()
+    }
+
+    /// Starts monitoring and triggers pulls for any needingToBeUpdatedFromBackend conversations.
+    public func start() async {
+        fetchedResultsController.delegate = self
+        do {
+            try fetchedResultsController.performFetch()
```
netbe, line 59: this class monitors the coredata change, and create ticket for scheduler

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1680,12 +1681,9 @@ public final class MLSService: MLSServiceInterface {
                     "Handling reset broken MLS conversation recovery strategy...",
                     attributes: [.mlsGroupID: groupID.safeForLoggingDescription]
                 )
-                var epoch: Int64 = 0
-                if let context, let conversation = fetchConversationInfo(
-                    with: groupID,
-                    in: context
-                )?.conversation {
-                    epoch = Int64(conversation.epoch)
```
netbe: this part getting the epoch outside the CoreData context was failing on debug build

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -221,7 +224,10 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
     /// Perform an incremental sync.
 
     func performIncrementalSync() async throws {
-
+        Task {
+            await conversationsMonitor.start()
```
netbe: we start generating the conversation sync with backend tickets only after initial sync is finished, in order to avoid having conversations syncing twice (in Worker and in InitialSync)

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
```diff
@@ -806,12 +806,23 @@ public final class ClientSessionComponent {
...
+    public lazy var workAgent: WorkAgent = {
+        let agent = WorkAgent(scheduler: PriorityOrderWorkScheduler())
+        agent.registerWorker(syncConversationsWorker)
+        // will start again whhen new tickets are enqueued
+        agent.shouldAutoStart = true
+        return agent
+    }()
 
-extension InitiateResetMLSConversationUseCase: ResetBrokenMLSConversationDelegate {
+    public lazy var conversationMonitor: ConversationUpdatesGeneratorProtocol = ConversationUpdatesGenerator(
```
johnxnguyen: suggestion: if we say that the worker is responsible for keeping local conversations in sync with the backend, then I would expect the worker to be capable of monitoring the database and raising tickets autonomously. In which case, the worker retains the monitor, the monitor simply informs a delegate or via callback that there are objects that it detected, and the worker creates the ticket.

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
johnxnguyen: suggestion: if the monitor is inside the worker, we could instead tell the workers to "start" and "stop" work. In this case, the conversation worker would start and stop monitoring. We could even add these start stop methods as optional protocol methods on `Worker`, then flip a flag on the work agent `isAcceptingTickets` or `isOpen`, which would then inform all its workers to start or stop. What do you think?

netbe: so workers would start when WorkAgent start to run, right?

in the case of conversations, I don't know if that would solve the issue where initialSync pulling conversations would create conversations with needsToBeUpdatedFromBackend = true and trigger the worker.

2) Did you mean that we should move setting needsToBeUpdatedFromBackend to true or false within the Worker?

WireDomain/Sources/WireDomain/WorkAgent/WorkItem/WorkItem.swift
```diff
@@ -39,6 +39,6 @@ protocol WorkItem: Sendable {
 
     /// Cancel the work item.
 
-    func cancel() async
+//    func cancel() async
```
netbe: @johnxnguyen this is not implemented, I wonder if we keep it

WireDomain/Sources/WireDomain/WorkAgent/WorkItem/UpdateConversationItem.swift
```diff
@@ -0,0 +1,74 @@
...
+
+    public init(
+        repository: ConversationRepositoryProtocol,
+        conversationID: WireNetwork.QualifiedID,
+    ) {
+        self.repository = repository
+        self.conversationID = conversationID
+    }
+
+    func start() async throws {
```
samwyndham, line 42: question (out of scope): @netbe @johnxnguyen Am I right that the way a `WorkAgent` has changed since we discussed it last Friday? The work now belongs to a `WorkItem` and not some other _actor_?

There were a few things that were nice about the _old_ approach:

1. Less need to pass dependencies to the place that the work item is created.
2. Possibility to persist work item across app runs.

Anyway this is out of scope for this PR.

WireDomain/Sources/WireDomain/WorkAgent/WorkItem/UpdateConversationItem.swift
```diff
@@ -0,0 +1,74 @@
...
+
+struct UpdateConversationItem: WorkItem {
+    private let repository: ConversationRepositoryProtocol
+
+    var id = UUID()
+    var priority: WorkItemPriority {
+        .medium
+    }
+
+    var conversationID: WireNetwork.QualifiedID
```
samwyndham: question: Why are `id` and `conversationID` publicly mutable?

WireDomain/Sources/WireDomain/Synchronization/ConversationUpdatesGenerator.swift
```diff
@@ -0,0 +1,116 @@
...
+extension ConversationUpdatesGenerator: NSFetchedResultsControllerDelegate {
+
+    public func controller(
+        _ controller: NSFetchedResultsController<NSFetchRequestResult>,
+        didChange anObject: Any,
+        at indexPath: IndexPath?,
+        for type: NSFetchedResultsChangeType,
+        newIndexPath: IndexPath?
+    ) {
+        guard let conversation = anObject as? ZMConversation else { return }
```
samwyndham: suggestion: Crash if this fails. The fetched results controller is configured in this file so we know that anObject right?

samwyndham (commented):
Looks good. I've just left a few comments

WireDomain/Sources/WireDomain/WorkAgent/WorkItem/UpdateConversationItem.swift
netbe: yes that's basically what change

WireDomain/Sources/WireDomain/WorkAgent/WorkItem/UpdateConversationItem.swift
netbe: no reason, true I'll make it a let

samwyndham (approved):
Nice work!

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -68,6 +68,7 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
     private let initialSyncTaskManager = NonReentrantTaskManager()
     private var incrementalSyncToken: IncrementalSync.Token?
     private var ongoingSyncTask: Task<Void, Never>?
+    private let conversationUpdatesGenerator: ConversationUpdatesGeneratorProtocol
```
johnxnguyen, line 71: issue: I fear this won't scale well as we add more generators, so I think this would be better outside of the `SyncAgent` and instead react to the delegate methods (possibly adding another for suspend) to start and stop the monitoring. What do you think?

WireDomain/Sources/WireDomain/Synchronization/ConversationUpdatesGenerator.swift
```diff
@@ -0,0 +1,118 @@
...
+        let conversations = fetchedResultsController.fetchedObjects ?? []
+        for conversation in conversations {
+            await context.perform {
+                if let id = conversation.qualifiedID {
+                    self.onConversationUpdated(UpdateConversationItem(
+                        repository: self.repository,
+                        conversationID: id.toAPIModel()
+                    ))
+                }
+            }
```
johnxnguyen, line 73: suggestion: an idea (I guess for the future) is to have a `PullConversationsSync` object that we can used to pull multiple conversations at once.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/AppVersionMigrations/AppVersionMigration_4_10_0.swift
```diff
@@ -0,0 +1,35 @@
...
+/// **Issue:**: Missing groups are not synced - [WPB-20123]
+struct AppVersionMigration_4_10_0: AppVersionMigration {
+
+    let version: SemanticVersion = "4.10.0"
+    let journal: any JournalProtocol
+
+    func perform() async throws {
+        // in order to fix conversations that does not show in the list
+        // we need to force a conversation sync
+        journal[.isConversationSyncRequired] = true
```
johnxnguyen, line 32: question: I couldn't see where this flag is checked. Is this migration used?

netbe: so it's the same flag as for migration 4.1.1. So just after loading the session we finish with fetching the conversations. This is needed to fix any issue someone might have to sync invalid conversations.

Screenshot 2025-11-10 at 16 54 27

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
netbe: well I agree it won't scale well, but do you suggest using the SyncAgentDelegate for this? If so, I fear to move the problem to ZMUserSession which handle too many things.

The reason for this to be here is that we want to sync conversations in same conditions as global sync:
- app in foreground once initial sync is done

So I see two options now:
1. An alternative could be keep a GeneratorsProvider (like StrategyDirectory😅) here that would start and stop
2. Add a dedicated delegate that forwards the events directly to ConversationUpdatesGenerator and later a kind of GeneratorsProvider

I don't see this as a blocker for now, or do you?

WireDomain/Sources/WireDomain/WorkAgent/WorkItem/WorkItem.swift
johnxnguyen: I guess we can delete it and bring it back when we implement it.

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
johnxnguyen: No not a blocker

johnxnguyen (approved):

PR #3800, opened 2025-10-28
feat: introduce WorkAgent - WPB-20123

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20123" title="WPB-20123" target="_blank">BugWPB-20123</a>  [iOS] Invisible groups, channels and 1:1 chats with functioning push notifications
  </td></table>
  <br />
 

### Issue
There is a need to perform tasks asynchronously and automatically, such as fetching conversation metadata for conversations marked as stale, or sending pending messages.

This PR introduces the `WorkAgent` as a system to generate, schedule, and perform work autonomously. Its design is modeled after ticketing systems common in public places such as banks and town halls where people generate a ticket with a computer system and wait for their ticket to be called. The system will schedule tickets according to priority and available resources and call people when their ticket is ready, who are then able to receive the service they have requested.

The system introduced here works in a similar way:
1. The `WorkAgent` is a central object that coordinates `Worker`s, `WorkTicket`s, and a `WorkScheduler`.
2. `Worker`s generate tickets and submit them to the `WorkAgent`.
3. The agent enqueues the ticket to the scheduler.
4. While the agent actively running, it will ask the scheduler for the next ticket, then call the associated worker.
5. The worker, when called, performs the work.

This is a simple an initial implementation to get started, but there are a variety of enhancements that we can make, such as:
- Handling cancellation: notifying workers tickets have been rejected
- Handling failed tickets: workers notify failures and suggest recovery options (e.g try again later)
- Preempting: cancel current tickets to make way for critical tickets
- Dequeue multiple tickets concurrently: currently one ticket is called at a time, but we could increase this number
- Smart scheduling based on environment conditions: don't call tickets that require network access if no network available, limit work load if resources (battery, memory) are constrained
- Add diagnostic metrics: keep track of ticket times, distribution of ticket priority, workloads etc.

### Testing
The system is not integrated yet.

---

### Checklist

---


wire-ios-mono.xcworkspace/xcshareddata/swiftpm/Package.resolved
johnxnguyen, line 1: hmm... this shouldn't be committed

WireDomain/Sources/WireDomain/WorkAgent/WorkTicket/WorkTicketPriority.swift
```diff
@@ -0,0 +1,85 @@
...
+    ///
+    /// Reserved for **urgent, must-run-now operations** that are essential
+    /// to app stability or correctness.
+    ///
+    /// These tasks should run immediately, even if that means preempting
+    /// other work. Use sparingly—for example, to recover from a failed
+    /// database write, handle an urgent network retry, or persist critical
+    /// user data before termination.
+
+    case critical
```
netbe: thought: I wonder why you would go with a WorkTicket if it needs to run immediately. Why can't you just do the work wherever you are then

wire-ios-sync-engine/Source/WorkAgent/WorkAgent.swift
```diff
@@ -0,0 +1,126 @@
...
+    var shouldAutoStart = false
+
+    // MARK: - Life cycle
+
+    private var task: Task<Void, Never>?
+    private let scheduler: any WorkScheduler
+    private var workers: [any Worker] = []
+    private let nonReentrantTaskManager = NonReentrantTaskManager()
+
+    private let logger = WireLogger(tag: "work-agent")
```
netbe: suggestion: declare in with other WireLogger instances

wire-ios-sync-engine/Source/WorkAgent/WorkAgent.swift
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    // MARK: - Operation
+
+    func registerWorker(_ worker: any Worker) {
+        workers.append(worker)
+    }
+
+    func submitTicket(_ ticket: any WorkTicket) {
+        logger.debug("ticket submitted: \(ticket)", attributes: .safePublic)
```
netbe: suggestion: let's create an attribute so then we can track easily the execution of a ticket

netbe (commented):
left a couple comments.

question: why was it added on SyncEngine rather than WireDomain? It would be easier also for integration to initialize on ClientSessionComponent

WireDomain/Sources/WireDomain/WorkAgent/WorkTicket/WorkTicketPriority.swift
johnxnguyen: I think if you can do it immediately and directly then you should. I'm struggling to think of any use case we have that would need a critical priority but I didn't think of one yet.

johnxnguyen:
> question: why was it added on SyncEngine rather than WireDomain? It would be easier also for integration to initialize on ClientSessionComponent

@netbe  I thought maybe it would make sense to keep it closer to the user session but happy to move it to WireDomain.

wire-ios-sync-engine/Source/WorkAgent/WorkAgent.swift
johnxnguyen: done

wire-ios-sync-engine/Source/WorkAgent/WorkAgent.swift
johnxnguyen: done

WireDomain/Sources/WireDomain/WorkAgent/WorkTicket/WorkTicketPriority.swift
netbe: if no use case, what about removing it?

netbe (approved):
Looks good

WireDomain/Sources/WireDomain/WorkAgent/WorkTicket/WorkTicketPriority.swift
johnxnguyen: sure

WireDomain/Sources/WireDomain/WorkAgent/WorkAgent.swift
```diff
@@ -0,0 +1,155 @@
...
+        task != nil
+    }
+
+    /// Whether dequeuing should begin after ticket submission.
+
+    var shouldAutoStart = false
+
+    // MARK: - Life cycle
+
+    private var task: Task<Void, Never>?
```
samwyndham, line 41: issue: I'm not confident that this won't be mutated from different threads. Could WorkAgent be an actor or could it's mutable properties have some multithreaded protection?

WireDomain/Sources/WireDomain/WorkAgent/WorkScheduler/PriorityOrderWorkScheduler.swift
```diff
@@ -0,0 +1,57 @@
...
+        case .medium:
+            mediumQueue.append(ticket)
+        case .high:
+            highQueue.append(ticket)
+        case .blocker:
+            blockerQueue.append(ticket)
+        }
+    }
+
+    func dequeueNextTicket() -> (any WorkTicket)? {
```
samwyndham: thought: I'm not really sure if GCD (for example) does anything special here but wondering if there is a danger that lower priority tasks get too blocked by higher priority tasks. Just thinking out loud.

WireDomain/Sources/WireDomain/WorkAgent/WorkAgent.swift
```diff
@@ -0,0 +1,155 @@
...
+    private let scheduler: any WorkScheduler
+    private var workers: [any Worker] = []
```
samwyndham: suggestion: Make this a dictionary.

samwyndham (approved):
Approving because I think this is needed soon but I have some questions that perhaps will be covered in tomorrow's meeting:

- Currently I don't quite understand where the actually work to be done is contained? I think I need to see an example with real work.
- Related to above, when reviewing I was kind of wondering is a ticket like a `DispatchWorkItem`? In the case of a DispatchWorkItem, the work to be done is defined in the item.. But now I'm thinking is the work to be done defined in the `Worker`? And perhaps a ticket provides information needed to perform the work such as userID? It would be good to see the names of a few possible Workers to understand this better.

WireDomain/Sources/WireDomain/WorkAgent/WorkAgent.swift
```diff
@@ -0,0 +1,155 @@
...
+        WireLogger.workAgent.info(
+            "starting",
+            attributes: .safePublic
+        )
+
+        task = Task {
+            let startTime = Date()
+            var completedTickets = 0
+
+            while let ticket = scheduler.dequeueNextTicket() {
```
samwyndham: thought: Dequeuing one work item at a time may be problematic when used in reality. Why enforce it at this stage at all? If we need to ensure one thing happens before another perhaps we need some kind of way of marking a ticket as a dependency of another ticket.

WireDomain/Sources/WireDomain/WorkAgent/WorkAgent.swift
```diff
@@ -0,0 +1,155 @@
...
+        await task?.value
+        task = nil
+    }
+
+    func stop() {
+        WireLogger.workAgent.info(
+            "stopping",
+            attributes: .safePublic
+        )
+        task?.cancel()
```
samwyndham, line 135: question: When we stop should the tasks while loop be cancelled too?

johnxnguyen:
> Approving because I think this is needed soon but I have some questions that perhaps will be covered in tomorrow's meeting:
> 
> * Currently I don't quite understand where the actually work to be done is contained? I think I need to see an example with real work.
> * Related to above, when reviewing I was kind of wondering is a ticket like a `DispatchWorkItem`? In the case of a DispatchWorkItem, the work to be done is defined in the item.. But now I'm thinking is the work to be done defined in the `Worker`? And perhaps a ticket provides information needed to perform the work such as userID? It would be good to see the names of a few possible Workers to understand this better.

@samwyndham yes you understand right, the work is to be done inside the `Worker` objects. imagine for example a `MessageSenderWorker` which is responsible for monitoring the database for unsent messages, raising a ticket for each message found, and when signal by the work agent, to perform the work described in that ticket (i.e send the message).

Another example could be a `KeyPackageReplenisher` that periodically checks (say once every 24 hours) if new key packages need to be generated and uploaded. Still another examples include "BlacklistChecker", "FeatureConfigFetcher", "ConversationMetadataFetcher".

The idea is that workers know what kind of work they need to do... they are autonomous in this regard, and will raise tickets whenever they want to perform work. But they wait for their ticket to be called in order to proceed with the work. This allows us to keep the work in the workers who have context, while the management and scheduling of the work is done elsewhere that doesn't need to know about the work being done (aside from provided metadata such as priority and requirements like network etc).

WireDomain/Sources/WireDomain/WorkAgent/WorkScheduler/PriorityOrderWorkScheduler.swift
johnxnguyen: Yes there certainly is that risk. This implementation is basic and intended as a starting point and we should build upon it to avoid starvation scenarios and allow fairness.

WireDomain/Sources/WireDomain/WorkAgent/WorkAgent.swift
johnxnguyen: Yep probably there's no current need to constrain execution like this, but I suppose there may be use cases where you'd want to limit the work load.

WireDomain/Sources/WireDomain/WorkAgent/WorkAgent.swift
johnxnguyen: Mapping id to worker?

WireDomain/Sources/WireDomain/WorkAgent/WorkAgent.swift
```diff
@@ -0,0 +1,155 @@
...
+                    WireLogger.workAgent.debug(
+                        "ticket complete",
+                        attributes: .init(ticket)
+                    )
+                } catch {
+                    WireLogger.workAgent.error(
+                        "ticket failed, dropping",
+                        attributes: .init(ticket)
+                    )
+                    continue
```
jullianm, line 114: question: if the work fails, ticket is removed from the queue and we move on to the next one, right ? 

thought:I wonder if it should be re-enqueued, maybe some tickets will fail because they might depend on something or connection is poor and they might succeed later on. 

Optional: We could also keep track of how many attempts were done on a specific ticket and log the attempts count as well.

WireDomain/Sources/WireDomain/WorkAgent/Worker/Worker.swift
```diff
@@ -0,0 +1,69 @@
...
+    func performWork(for ticket: Ticket) async throws
+
+}
+
+extension Worker {
+
+    func performWork(for ticket: any WorkTicket) async throws {
+        guard let ticket = ticket as? Ticket else {
+            throw WorkTicketRecoveryStrategy.drop
+        }
```
jullianm: nitpick: I believe we can do `extension Worker where Ticket: WorkTicket` to avoid dealing with typecasting here

jullianm (commented):
Nice work. I left a question/thought

WireDomain/Sources/WireDomain/WorkAgent/WorkAgent.swift
johnxnguyen: Yes, in the description I outlined some improvements such as rescheduling and recovery options, but I haven't included it yet.

WireDomain/Sources/WireDomain/WorkAgent/Worker/Worker.swift
johnxnguyen: I don't quite follow, could you elaborate? Note that `Ticket` already conforms to `WorkTicket` as it's part of the associated type requirement.

jullianm: I missed that my bad. discard my comment

jullianm (approved):

WireDomain/Sources/WireDomain/WorkAgent/WorkAgent.swift
johnxnguyen: yes, fixed

WireDomain/Sources/WireDomain/WorkAgent/WorkAgent.swift
johnxnguyen: made it an actor

WireDomain/Sources/WireDomain/WorkAgent/WorkAgent.swift
johnxnguyen: done

.github/workflows/_reusable_run_tests.yml
```diff
@@ -135,10 +135,10 @@ jobs:
...
           bundle exec fastlane generate_env
 
-      # - name: "Test new critical flows"
-      #   if: ${{ inputs.run_critical_flows }}
-      #   run: |
-      #     bundle exec fastlane run_uitests
+      - name: "Test new critical flows"
+        if: ${{ inputs.run_critical_flows }}
+        run: |
+          bundle exec fastlane run_uitests
```
netbe: todo: need to revert that

PR #3802, opened 2025-10-28
fix: displaying coredata popup - WPB-21424

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21424" title="WPB-21424" target="_blank">BugWPB-21424</a>  [iOS] CoreData error alert are not shown properly
  </td></table>
  <br />
 

### Issue

If user has an issue with it's database, we show an alert. The current alert offers to logout or retry. Neither of these options work.

* Replace popup with clear database, switch accounts.
* Logout cannot work because the session is not loaded if the db fails

### Testing

1. install 4.10.0 version with db change
2. install 4.9.0 version
3. popup shows

Simulator Screenshot - iPhone 14 - 18 5 - 2025-10-28 at 20 59 29

---

### Checklist

---


wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1104,6 +1104,16 @@ public final class SessionManager: NSObject, SessionManagerType {
                     error: .networkError(code: error.errorCode)
                 )
                 return nil
+            } catch let UserSessionLoader.Failure.failedToLoadPersistenceStack(error) {
+                WireLogger.sessionManager.error(
+                    "failed to load user session: \(String(describing: error))",
+                    attributes: .safePublic
```
samwyndham, line 1110: question: Is it safe public? I'm not sure what error can contain?

samwyndham (commented):
LGTM but a question about logging

johnxnguyen (approved):

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
netbe: I think it's fine, as it's the same log as before. Expect error about CoreData error could not read, see CoreDataMigratorError

samwyndham (approved):

PR #3806, opened 2025-10-29
fix: crash when uploading link preview upload - WPB-20714

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20714" title="WPB-20714" target="_blank">StoryWPB-20714</a>  [iOS] Upload asset with additional metadata
  </td></table>
  <br />
 

### Issue
If asset audit log is enabled, when I send a link in a conversation that generates a preview, then the app crashes and will keep crashing each time the app is relaunched.

The reason is that when trying to upload the link preview we always generate a nil request because we can't find the original image data and then later hit an assertion failure.

It appears that there is no need to get the original asset data, so the fix is to simply remove the guard. This fixes the crash, but the upload still fails because the mime type is empty And the backend rejects it. To resolve this, we provide a default `"application/octet-stream"` mime type ("consider this raw bytes") which the backend accepts and we can upload the link preview. 

### Testing
1. Log in to anta
2. send a link that generates a link preview
3. assert that the link preview is successfully sent and received on the other side (with image)

---

### Checklist

---

### DIFF
--- a/wire-ios-request-strategy/Sources/Helpers/AssetRequestFactory.swift
+++ b/wire-ios-request-strategy/Sources/Helpers/AssetRequestFactory.swift
@@ -47,13 +47,18 @@ public final class AssetRequestFactory: NSObject {
         public init(
             conversationID: QualifiedID,
             fileName: String,
-            mimeType: String
+            mimeType: String?
         ) {
             self.conversationID = conversationID
             self.fileName = fileName
-            self.mimeType = mimeType
-        }
 
+            if let mimeType, !mimeType.isEmpty {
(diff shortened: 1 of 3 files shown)


netbe (approved):
Looks good, but i don't see which assertion will make the app to keep on crashing

johnxnguyen:
> Looks good, but i don't see which assertion will make the app to keep on crashing

@netbe its in an upstream sync when it detects the transcoder returning a nil request.

samwyndham (approved):

PR #3521, opened 2025-08-29
chore: move db models to wire data package - WPB-19688

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19688" title="WPB-19688" target="_blank">TaskWPB-19688</a>  [iOS] Move DB data model to `WireData`
  </td></table>
  <br />
 

### Issue

This PR moves the `zmessaging` and `ZMEventModel` database models from `WireDataModel` framework to `WireData` package ensuring that the managed objects that were previously pointing to the `Current Product Module` now points to `WireDataModel` (where still reside their related subclasses).

Note: there was no need to re-generate the resources (i.e `store2-129-0.wiredatabase`) for the tests in `WireDataModel` as the tests were passing.

#### Sam's extra comments

I've assigned this PR to myself as @jullianm is out this week. Some things I've learnt which I didn't know before regarding the classes module:

Screenshot 2025-08-25 at 10 24 12

This can have 3 states:

- Global Namespace
- Current Product Module
- Custom (we enter a target)

`Global Namespace` is required for managed object subclasses that are obj-c or those that are used from obj-c. This is because obj-c uses a global namespace (hence the need to add prefixes).

`Current Product Module` is like `Custom` but automatically sets the target based on where the database models reside. Currently `Current Product Module` resolves to `WireDataModel` but with this PR we move the database models to `WireData` while keeping the classes in `WireDataModel`. This means that we have to updates settings of `Current Product Module` to `WireDataModel`.

Looking at this PR I was surprised that it updates each version of `zmessaging.xcdatamodeld`, not just `2.129` but `2.128`, `2.127`, etc. I now realize that this is because some of our migration tests depend on our managed object subclasses with older database versions. For example, `DatabaseMigrationTests_Conversations.testThatItPerformsMigrationFrom106_deleteConversationCascadesToParticipantRole()`. I think this is a little problematic as there is no reason why the current version of `ParticipantRole` is compatible with version `2.106.0` for example. It would be better to use just NSManagedObject with KVC. However it's probably best to leave things how they are.

Beyond @jullianm's work I added one small additional [test](https://github.com/wireapp/wire-ios/pull/3521/commits/342dad9e6bb21e09e4694d54806c051f05163a46) to assert that each managed object subclass of the latest data model can be instantiated. 

### Testing

- Ensure app is building and running properly
- Ensure tests in `WireDataModel` and `WireSyncEngine` are still passing
- Ensure database migration from 2.128.0 to 2.129 is working properly

[...]

PR #3819, opened 2025-10-31
chore: move db models to wire data package (second attempt) - WPB-19688

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19688" title="WPB-19688" target="_blank">TaskWPB-19688</a>  [iOS] Move DB data model to `WireData`
  </td></table>
  <br />
 

### Issue

This PR is a new version of https://github.com/wireapp/wire-ios/pull/3521. As that PR was quite far behind `develop` and the core data model has changed, I decided to combine changes from the original PR with a few updates.

This PR moves the `zmessaging` and `ZMEventModel` database models from `WireDataModel` framework to `WireData` package ensuring that the managed objects that were previously pointing to the Current Product Module now points to WireDataModel (where still reside their related subclasses).

To achieve this:

- Model files were moved to `WireData`
- Managed object subclass files were left in `WireDataModel` - these can be moved one by one as needed.
- The module for each entity was changed to WireDataModel` is previously it was set as `Current Product Module` - see notes below. 
- The locations of the model bundles were updated in tests.
- In Xcode project based modules that depend on core data, their test hosts were updated to link to the modules target. E.g. `WireRequestStrategyTestHost` was updated to link to `WireRequestStrategy`. Without doing this the `WireDataBundle.bundle` would crash trying to find the bundle. This change ensures that the compiled core data module resides in the test host and is therefore accessible in tests. This was the main difference to the original PR: https://github.com/wireapp/wire-ios/pull/3521
- A new test host was added for `WireDomain` to address the above issue as it was missing a test host.

**This PR is largely @jullianms good work - my changes mainly relate to resolving test host issue** 

#### Some additional notes

Some things I've learnt which I didn't know before regarding the classes module:

Screenshot 2025-08-25 at 10 24 12

This can have 3 states:

- Global Namespace
- Current Product Module
- Custom (we enter a target)

`Global Namespace` is required for managed object subclasses that are obj-c or those that are used from obj-c. This is because obj-c uses a global namespace (hence the need to add prefixes).

`Current Product Module` is like `Custom` but automatically sets the target based on where the database models reside. Currently `Current Product Module` resolves to `WireDataModel` but with this PR we move the database models to `WireData` while keeping the classes in `WireDataModel`. This means that we have to updates settings of `Current Product Module` to `WireDataModel`.

I was surprised at the need to update each version of `zmessaging.xcdatamodeld`, not just `2.129` but `2.128`, `2.127`, etc. This was necessary because some of our migration tests depend on our managed object subclasses with older database versions. For example, `DatabaseMigrationTests_Conversations.testThatItPerformsMigrationFrom106_deleteConversationCascadesToParticipantRole()`. I think this is a little problematic as there is no reason why the current version of `ParticipantRole` is compatible with version `2.106.0` for example. It would be better to use just NSManagedObject with KVC. However it's probably best to leave things how they are.

### Testing

- Ensure app is building and running properly
- Ensure all unit tests are passing [test run here](https://github.com/wireapp/wire-ios/actions/runs/18979382045)
- Ensure database migration from 2.130 to 2.131 is working properly

---

### Checklist


WireData/Sources/WireData/Schema/zmessaging.xcdatamodeld/zmessaging2.131.0.xcdatamodel/contents
```diff
@@ -1,18 +1,18 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="24299" systemVersion="24G231" minimumToolsVersion="Xcode 15.0" sourceLanguage="Swift" userDefinedModelVersionIdentifier="2.131.0">
-    <entity name="Action" representedClassName=".Action" syncable="YES">
+    <entity name="Action" representedClassName="WireDataModel.Action" syncable="YES">
```
caldrian, line 3: question: WireData cannot access WireDataModel. Are you sure referencing entity classes from WireDataModel is correct?

WireDomain/WireDomain Project.xcodeproj/project.pbxproj
```diff
@@ -26,6 +26,7 @@
 		C9E0C9BB2C91B76F00CE6607 /* WireTestingPackage in Frameworks */ = {isa = PBXBuildFile; productRef = C9E0C9BA2C91B76F00CE6607 /* WireTestingPackage */; };
 		CB6343F02DB7EF3D00A1C892 /* WireUpdateEventCoding in Frameworks */ = {isa = PBXBuildFile; productRef = CB6343EF2DB7EF3D00A1C892 /* WireUpdateEventCoding */; };
 		CB7979132C738508006FBA58 /* WireTransportSupport.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = CB7979122C738508006FBA58 /* WireTransportSupport.framework */; };
+		CBC41E722EB51569000769CE /* WireDomain.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 01D0DCA62C1C8C870076CB1C /* WireDomain.framework */; };
```
netbe, line 29: question: what's this change about?

wire-ios-data-model/Resources/zmessaging.xcdatamodeld/zmessaging2.110.0.xcdatamodel/contents
```diff
@@ -269,7 +269,7 @@
         <relationship name="removedUsers" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="User" inverseName="showingUserRemoved" inverseEntity="User"/>
         <relationship name="users" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="User" inverseName="systemMessages" inverseEntity="User"/>
     </entity>
-    <entity name="Team" representedClassName=".Team" syncable="YES">
```
netbe, line 272: thought: So I wonder if we should just duplicate / freeze the model for older versions, that way all the models don't need to be updated

But this could cause some confusion for us, what do you think @samwyndham ?

netbe (approved):
left a few comments, but otherwise looks good to me :) 

Question: Can we close https://github.com/wireapp/wire-ios/pull/3521 ?

wire-ios-data-model/Resources/zmessaging.xcdatamodeld/zmessaging2.110.0.xcdatamodel/contents
samwyndham: @netbe The issue is that we access the old models in tests, this is why they have also been updated. Keep in mind though that these changes are not schema changes so does not affect migrations.

WireDomain/WireDomain Project.xcodeproj/project.pbxproj
samwyndham: @netbe In this project I add a new test host. I _guess_ this particular change must be adding `WireDomain` as a dependency of the test host.

WireData/Sources/WireData/Schema/zmessaging.xcdatamodeld/zmessaging2.131.0.xcdatamodel/contents
samwyndham: @caldrian Yes I _think_ this is fine. My understanding is that `representedClassName` is only accessed dynamically when we try to create/access in this case an `Action`. As we don't access `Action` from `WireData` this is OK. I expect that if this doesn't work it would be immediately obvious when running the app.

samwyndham: Unfortunately I can't find documentation on this. The best I have is ChatGPTs answer to this question which I think has a good point - that we don't access `xcdatamodeld` directly. Instead this is compiled to a `.momd` that lives in the apps bundle and is accessed dynamically.

From chat GPT:

Let’s restate what’s going on:

- Target Bar contains your Core Data model (.xcdatamodeld).
- Target Foo contains your NSManagedObject subclasses.
- Foo imports Bar, so it can access the model if needed.
- The model in Bar references the classes in Foo via representedClassName.
- But Bar does not import Foo.

Now the key question:
Is it OK that Bar does not import Foo?
✅ Short answer: Yes, that’s perfectly fine — and actually desirable.

**Why It’s OK**
The .xcdatamodeld file doesn’t compile into code that needs to reference the managed object subclasses at build time. Instead, it compiles into a .momd resource (a Core Data model metadata file) that simply stores the class names as strings (e.g. "Foo.MyEntity").
When Core Data loads the model at runtime, it uses these class names as strings and asks the Objective-C runtime (or Swift runtime) to resolve them. That means:
- There’s no compile-time dependency between your model bundle (Bar) and your entity classes (Foo).
- The only requirement is that, at runtime, the classes referenced by representedClassName actually exist and can be found via the runtime (e.g. are linked into the process).

caldrian: Hm, I'm a bit skeptical but let's try it.
I guess once we want to have tests in WireData we have to move the entity files as well.

caldrian (approved):

wire-ios-data-model/Resources/zmessaging.xcdatamodeld/zmessaging2.110.0.xcdatamodel/contents
netbe: @samwyndham which test in particular ? it would make sense to remove or update the test

samwyndham: @netbe The tests I know about are `DatabaseMigrationTests_Conversations`. It is possible there are more that I haven't checked. If during migration we rely on on NSManagedObject subclasses somewhere else that would also break.

PR #3821, opened 2025-11-03
chore: simplify configuration when requesting nodes - WPB-21418

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21418" title="WPB-21418" target="_blank">TaskWPB-21418</a>  [iOS] As a user I want to see folders in the files view
  </td></table>
  <br />
 

### Issue

When fetching nodes via the /nodes endpoint there is a lot of configuration that can be done to the requests to handle filtering, scope of nodes fetched etc. This is defined in the CellsSDK a library we use generated from our swagger files.

When implementing the files view I needed a way to configure `WireCellsFetchNodesUseCase` with this information. My solution was to essentially duplicate the configuration from `CellsSDK` to `WireMessagingDomain` and map this in the data layer.

I thought it was nice that we could configure the API from `WireMessagingDomain`. However in practice I've found `/nodes` behavior to be surprising and we've had to set CellsSDK parameters slightly magically to have behavior we want.

As I'm making more changes here I don't want this leaking through to the domain layer so in this PR I have changed how things are done:

- Deleted duplicate configuration objects
- Rather than the domain passing a complex configuration object it passes a simple one which names what it wants - eg. `conversationFileView(root: WireCellsNodeLocator)`. It is then the data layer that maps that simple configuaration to the complex configuration required by the CellsSDK.

### Testing

1. Navigate to the files view of a conversation and check that files are loaded as expected.
2. Navigate to the all files view and check that files are loaded as expected.

---

### Checklist

### DIFF
--- a/WireMessaging/Sources/WireMessagingAssembly/WireMessagingFactory.swift
+++ b/WireMessaging/Sources/WireMessagingAssembly/WireMessagingFactory.swift
@@ -131,20 +131,14 @@ public extension WireMessagingFactory {
     @MainActor
     func makeFilesView(
         cellName: String,
-        isCellsStatePending: Bool,
-        nodeIDs: [UUID]
+        isCellsStatePending: Bool
     ) -> UIViewController {
-        let configuration: WireCellsFetchNodesUseCase.Configuration = nodeIDs
-            .isEmpty ? .conversationFileView(root: .path(cellName)) :
-            .nodesFileView(nodeIDs: nodeIDs)
-
+        let configuration: WireCellsFetchNodesUseCase.Configuration = .conversationFileView(root: .path(cellName))
         let filesView: UIHostingController<FilesView>
-
         filesView = makeFilesHostingController(
             configuration: configuration,
             isCellsStatePending: isCellsStatePending
         )
-
         return filesView
     }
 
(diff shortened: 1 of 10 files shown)


WireMessaging/Sources/WireMessagingAssembly/WireMessagingFactory.swift
```diff
@@ -131,26 +131,20 @@ public extension WireMessagingFactory {
     @MainActor
     func makeFilesView(
         cellName: String,
-        isCellsStatePending: Bool,
-        nodeIDs: [UUID]
```
samwyndham, line 135: `nodeIDs` was used in a temporary hack which we no longer use.

WireMessaging/Tests/WireMessagingTests/Resources/ReferenceImages/FilesBrowserViewTests/testFilesBrowserView_ReceivedItemsState.dark.png
samwyndham, line 1: These snapshot test references seem to be missing from develop so I've added them in this PR. They are unconnected to the rest of this PR.

jullianm (approved):
Looks good

WilhelmOks (approved):

PR #3836, opened 2025-11-05
fix: "keep ringing after joining the call from desktop" issue - WPB-18769

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18769" title="WPB-18769" target="_blank">BugWPB-18769</a>  [iOS] iPhone keep ringing after joining the call from desktop
  </td></table>
  <br />
 

### Issue

We get _"the phone keep ringing after joining the call from desktop"_ issue  from time to time. This issue can have various causes, but there is a use case with steps that allow us to reliably reproduce this issue.

There is a **group** conversation between **User1** and **User2**.
- User1 (client 1) opens Wire (app is on the **foreground**).
- User1 (client 2) logs into desktop app.
- User2 initiate the call in the group conversation.
- User1 (client 1) receives the call on iPhone and moves the app to the background (the phone is ringing).
- User1 (client 2) joins the call from the desktop app.
 
**Result:** the app is ringing.

### Solution

When the app is in the foreground, we receive an event via the web socket and process it (the app is ringing). When the app goes into the background, it likely closes the web socket, and a second event ("answered elsewhere") arrives via push channel (logs for this event were in the NotificationServiceExtension logs). Thus, the second event is delivered via push notification to NSE and process.
Event arrives with:
  - conversationID = selfConversation ID
  - Inside calling message: calling.qualifiedConversationID = group conversation ID
  - callContent.isAnsweredElsewhere = true

In the ConversationCallingEventNotificationBuilder validation method, we exclude events from selfUser, and this is the first reason why the app doesn't stop ringing (we just ignore this event).
I fixed this by checking the combination of selfUser and selfConversation. This way, we allow events to be processed if they are from selfUser and selfConversation.
!!! For reference, we even have  `testThatItDoesProcessEventsFromSelfUserArrivingInSelfConversation` in the EventDecoderTest.

The second issues is that we are using the conversation ID from the event (which is selfConversation), not from the calling message content.

### Testing

There is a group conversation between User1 and User2.
- User1 (client 1) opens Wire (app is on the foreground).
- User1 (client 2) logs into desktop app.
- User2 initiate the call in the group conversation.
- User1 (client 1) receives the call on iPhone and moves the app to the background (the phone is ringing).
- User1 (client 2) joins the call from the desktop app.
 
**Actual result:** the app is ringing.
**Expected result:** the app stops ringing

### Checklist

---


WireDomain/Sources/WireDomain/Notifications/Builders/Conversation/ConversationCallingEventNotificationBuilder.swift
```diff
@@ -37,16 +37,25 @@ struct ConversationCallingEventNotificationBuilder: ConversationCallingEventNoti
             return nil
         }
 
+        var resolvedConversationID: ConversationID {
+            let callingConversationID = calling.qualifiedConversationID
+            guard !callingConversationID.id.isEmpty,
+                  let conversationUUID = UUID(uuidString: callingConversationID.id)
+            else {
+                return conversationID
```
David-Henner, line 50: wouldn't this cause the same issue as you described, where the conversation ID is the id of the self conversation ? If so, should we treat this as an error instead ?

WireDomain/Sources/WireDomain/Notifications/Builders/Conversation/ConversationCallingEventNotificationBuilder.swift
```diff
@@ -37,16 +37,25 @@ struct ConversationCallingEventNotificationBuilder: ConversationCallingEventNoti
             return nil
         }
 
+        var resolvedConversationID: ConversationID {
```
David-Henner: wouldn't it be better to avoid using a computed property here ? `resolvedConversationID` is called 3 times below, maybe it's better to set it once ? what do you think ?

David-Henner (approved):
Good work ! Approving but please consider my comments

wire-ios-data-model/Support/Sources/ModelHelper.swift
```diff
@@ -21,7 +21,7 @@ import WireDataModel
 
 /// A helper object to make it easy to create and relate model objects.
 
-public struct ModelHelper {
+public struct ModelHelper: Sendable {
```
netbe, line 24: praise: nice

netbe (approved):
lgtm;) as it is a bug fix, consider put this to 4.10 maybe

johnxnguyen (approved):

KaterinaWire:
> lgtm;) as it is a bug fix, consider put this to 4.10 maybe

@netbe I will cherry-pick to 4.10

WireDomain/Sources/WireDomain/Notifications/Builders/Conversation/ConversationCallingEventNotificationBuilder.swift
KaterinaWire: I will add logs

WireDomain/Sources/WireDomain/Notifications/Builders/Conversation/ConversationCallingEventNotificationBuilder.swift
KaterinaWire: fixed

PR #3850, opened 2025-11-06
feat: Cells tags - WPB-19412

WilhelmOks:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19412" title="WPB-19412" target="_blank">TaskWPB-19412</a>  [iOS] Implement Tags - Add, remove, display tags
  </td></table>
  <br />
 

### Issue

* Context Menu Action:
   * Files in the Files tab should have a new option: Add or Remove Tags.
* Modal:
   * Input field to type tag names and Add new tags.
   * Typing filters the Suggested Tags list in real-time.
   * Display already Added tags.
   * Suggested tags stick to the keyboard in a horizontal carousel for easy selection.
* Feedback:
   * Error occurs when the tag name doesn’t meet the requirements: 
       * Length: Min 1 character - Max 30 characters
       * Disallowed characters , ; / \ " ' < > 
   * Show error alert if an action fails.
* File List Display:
   * Added tags should be visible directly in the file list.
   * If more than one tag is present, display the first tag + +x to indicate additional tags.

This PR now also includes the sorting requested by WPB-21607

### Implementation details

* The tags suggestions which are filtered by the text typed into the text field are shown in an "accessory view" (a toolbar element in SwiftUI) so that they stick to the virtual keyboard.
The styling of that area is given by the system and is not very customizable. The background color is very similar to the color of the tag bubbles, so I took the liberty to add a gray border around the bubble.

* When the save action succeeds, before the screen is dismissed automatically, the files list is reloaded in the background.

* The list of tag suggestions is loaded when the view model is initialized (so when the tag edit screen is opened)

### Testing

* Add or remove some tags
* Check if the tags are updated in the files list


WireMessaging/Sources/WireMessagingDomain/WireCells/UseCases/WireCellsGetTagSuggestionsUseCase.swift
```diff
@@ -0,0 +1,33 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+package import Foundation
```
samwyndham: nitpick: Fix import warning above

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FilesViewItemView.swift
```diff
@@ -30,6 +30,12 @@ struct FilesViewItemView: View {
     @StateObject private var viewModel: FilesItemViewModel
     @ScaledMetric private var imageHeight: CGFloat = 28
 
+    let additionalTagNumberFormatter = {
```
samwyndham: nitpick: Can this be private?

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FilesViewItemView.swift
```diff
@@ -50,16 +56,45 @@ struct FilesViewItemView: View {
...
+                                .padding(.vertical, 2)
+                                .padding(.horizontal, 5)
+                                .background {
+                                    RoundedRectangle(cornerRadius: 4)
+                                        .fill(ColorTheme.Base.primaryVariant.color)
+                                }
+                        }
+
+                        let additionalTags = viewModel.item.tags.count - 1
+
```
samwyndham: suggestion: Put this into the view model... Expose something optional like:

```swift
var additionalTagsValue: String?
```

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FilesViewModel.swift
```diff
@@ -84,9 +98,31 @@ package final class FilesViewModel: ObservableObject {
...
+        }
+
+        let fetchNodes: WireCellsFetchNodesUseCase
+        let deleteNodes: WireCellsDeleteNodesUseCase
+        let renameNode: any WireCellsRenameNodeUseCaseProtocol
+        let updateTags: any WireCellsUpdateTagsUseCaseProtocol
+        let getTagSuggestions: any WireCellsGetTagSuggestionsUseCaseProtocol
+    }
+
+    let useCases: UseCases
```
samwyndham, line 147: praise: This is looking neater for sure.

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FilesViewModel.swift
```diff
@@ -58,6 +61,17 @@ package final class FilesViewModel: ObservableObject {
         static let loadMoreThreshold = 5
     }
 
+    enum SheetNavigation: Identifiable {
```
samwyndham, line 78: praise: Nice idea. We will have a few kinds of sheets to handle.

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Tags/TagsEditView.swift
```diff
@@ -0,0 +1,331 @@
...
+
+    private let horizontalPadding: CGFloat = 16
+    private let tagBubbleSpacing: CGFloat = 8
+
+    struct UseCases {
+        let updateTags: any WireCellsUpdateTagsUseCaseProtocol
+        let getSuggestions: any WireCellsGetTagSuggestionsUseCaseProtocol
+    }
+
+    init(fileItem: FilesViewItem, useCases: UseCases, postSaveAction: @escaping () async -> Void) {
```
samwyndham, line 43: question: This is a different pattern then we have been using - usually we inject the view model. I'm fine to go with it for now though we might end up changing it to standardize. Do you have any pros of doing this rather than passing in the view model?

samwyndham (approved):
Nice work!

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FilesViewModel.swift
WilhelmOks: In the early days of SwiftUI, sheets used to break when there were more than one .sheet() in one View. I think this doesn't happen anymore but I like to be on the safe side and I think it's also a good pattern because you have all possible sheet navigations encapsulated in one type.

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Tags/TagsEditView.swift
WilhelmOks: I developed this habit a while ago because it allows the ViewModel to be hidden and an implementation detail of the View, rather than exposing it to the rest of the code.

Usually the ViewModel has only a few parameters. However in Wire there are a lot and I also started to wonder as well if it's good to repeat and relay them from View to ViewModel. In this case, I'm okay with changing it to a single ViewModel parameter, but generally I think we should try to avoid having to pass too many arguments. 

It's hard to figure out where most of the dependency instances can be taken from or how they can be created.
The dependencies are handed around in a lot of places, in large parameter lists, seemingly repeating themselves a lot of times. And this is doubled by the fact that also mocked dependencies need to be created for previews and tests.
Maybe we can pass them as environment values instead, as it is intended by SwiftUI. I'm not entirely sure about that.

I'd be glad to have a discussion about how we can improve that.

WireMessaging/Sources/WireMessagingDomain/WireCells/UseCases/WireCellsUpdateTagsUseCase.swift
```diff
@@ -0,0 +1,32 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+package import Foundation
+
+/// Sets a new list of tags for a file or folder, represented by the given node ID.
+package struct WireCellsUpdateTagsUseCase<NodesAPI: NodesAPIProtocol>: WireCellsUpdateTagsUseCaseProtocol {
+    private let nodesAPI: NodesAPI
```
jullianm, line 23: question: Is there a reason you're using generic over existential type like in our other use cases?
I believe we should have consistency across our use cases definition. I don't mind one approach or the other but probably something we should all agree on.
Also here I would consider using the `WireCellsNodesRepositoryProtocol` instead of the `NodesAPIProtocol` which we'll end up removing I believe.

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Tags/TagsEditView+ViewModel.swift
```diff
@@ -0,0 +1,157 @@
...
+//
+
+import Combine
+import Foundation
+import WireLogging
+import WireMessagingDomain
+
+extension TagsEditView {
+    @MainActor
+    final class ViewModel: ObservableObject {
```
jullianm, line 26: suggestion: move the view model to its own file outside of the view extension just like we do for other views. although eventually I prefer this approach, something to discuss and align on as well.

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FilesViewItemView.swift
```diff
@@ -81,6 +108,10 @@ struct FilesViewItemView: View {
                         }
                     }
 
+                    Button(action: editTags) {
```
jullianm: issue: this action will also be triggerable from the Files tab which I don't think is expected and is not doing anything currently. I think it should be hidden the same way we hid the "renaming" action from that view.  See `canRenameFile` flag from that view maybe the logic can be similar using a `canEditTags` flag

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FilesViewModel.swift
jullianm: very cool, thanks.

jullianm (changes requested):
Nice work, I think we should all align on the way we define our use cases, view models and inject them so we maintain consistency across our components. Also I believe  there's one issue to fix with the "add or remove tags" action which should be hidden when we're in the (all) "Files" tab.

WireMessaging/Sources/WireMessagingDomain/WireCells/UseCases/WireCellsUpdateTagsUseCase.swift
WilhelmOks: **generics**
I use generics to make a step towards being able to get rid of "any" later. I think this would be a bit cleaner, with better compile time suggestions and lower runtime impact.
But it will only be possible to remove "any", when we do it everywhere in the chain, because when something is "any", it cannot be passed into a generic or "some".

WilhelmOks: **NodesAPIProtocol**
In my implementation tags are not managed by a repository, so I'm using the nodes API directly.

jullianm (approved):
Looks good, nice work.

PR #3858, opened 2025-11-07
fix: committing pending proposals after leaving conversation - WPB-20223

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20223" title="WPB-20223" target="_blank">TaskWPB-20223</a>  [iOS] filter out conversations you left from joining pending proposals
  </td></table>
  <br />
 

### Issue

Sometimes users who have left a conversation still commit pending proposals.

### Solution

Check that the self user is a member of the conversation when:
- getting the list of conversations with pending proposals
- committing a scheduled pending proposal after the commit delay has passed

---

### Checklist


wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1460,22 +1474,26 @@ public final class MLSService: MLSServiceInterface {
...
+            ).filter(
+                \.isSelfAnActiveMember
+            ).compactMap {
+                guard
+                    let groupID = $0.mlsGroupID,
+                    let proposalDate = $0.commitPendingProposalDate
+                else {
+                    return nil
+                }
+                return (groupID, proposalDate)
```
netbe, line 1489: suggestion: what about moving the filtering to the method ?

netbe (approved):
LGTM;) left a suggestion

wire-ios-data-model/Source/MLS/MLSService.swift
David-Henner: My preference is to leave it as a utility method to perform the fetch request and leave the filtering outside. Mainly because it would require to have a very verbose name like `fetchConversationWithPendingProposalsWhereSelfIsAnActiveMember`.

johnxnguyen (approved):

PR #3778, opened 2025-10-22
feat(backup-performance): import messages from backup in batches - WPB-21300

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21300" title="WPB-21300" target="_blank">TaskWPB-21300</a>  [iOS] Refactor backup import to insert messages in batches
  </td></table>
  <br />
 

### Issue

This PR is an effort to improve the performance of importing messages from a backup. We now process messages in batches instead of processing messages sequentially by reusing single-message processing logic (for messages received in notifications).

Kalium's cross-platform backup API provides us with pagers for each entity saved in the backup file. Each page is a batch of 1000 entities. We use this to process messages in batches.

Page processing is structured in several steps:
1. Parse the backup messages models to collect the relevant data in order to restore messages. (ids and domains to fetch conversations and users, underlying generic messages for each message, message attributes to insert in Core Data..)
2. Fetch batches of users and conversations to establish relationships with messages
3. Insert batches of messages (one batch of client messages - text, location - and one batch of asset messages)
4. "Rehydrate" messages. i.e: establish relationships and set the generic message data + mark message as read
5. Save and reset the context 

---

### Checklist

- [x] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [x] Description is filled and free of optional paragraphs.
- [x] Adds/updates automated tests.

PR #3766, opened 2025-10-21
feat(backup-performance): update data model - WPB-21300

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21300" title="WPB-21300" target="_blank">TaskWPB-21300</a>  [iOS] Refactor backup import to insert messages in batches
  </td></table>
  <br />
 

### Issue

This PR adds attributes to the `ZMMessage` model in Core Data. They will be used to re-establish relationships with the message's sender and conversation during backup import, which will be the subject of another PR.

---

### Checklist

- [x] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [x] Description is filled and free of optional paragraphs.
- [ ] Adds/updates automated tests.

PR #3860, opened 2025-11-07
chore: remove unused attributes from Message entity - WPB-21300

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21300" title="WPB-21300" target="_blank">TaskWPB-21300</a>  [iOS] Refactor backup import to insert messages in batches
  </td></table>
  <br />
 

### Issue

I recently introduced new attributes for the `Message` entity in https://github.com/wireapp/wire-ios/pull/3766. It was a change I was relying on initially for https://github.com/wireapp/wire-ios/pull/3778, but I went with a different approach that doesn't rely on these attributes. Hence why I'm removing these attributes in this PR

---

### Checklist


netbe (changes requested):
since a new model was introduced on develop, better target develop on top of new Wire-Data datamodel

netbe (approved):
LGTM;)

caldrian (approved):

PR #3862, opened 2025-11-07
fix: crash setting supported protocols - WPB-21668

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21668" title="WPB-21668" target="_blank">BugWPB-21668</a>  [iOS] - Crash when setting supported protocols
  </td></table>
  <br />
 

### Issue

There’s a crash when core data merges conflicts after a context save. 

Screenshot 2025-11-07 at 17 12 54

### Cause

Core data is likely attempting to update `ZMUser.supportedProtocols` by calling its own synthesized setter but instead ends up calling the objc selector `ZMUser.setSupportedProtocols` defined in `ZMUser+SupportedProtocols.swift`

### Solution

Rename the method to avoid clashing with Core Data

---

### Checklist


KaterinaWire (approved):

samwyndham (approved):
Nice find!

PR #3889, opened 2025-11-14
fix(backup): duplicate messages after import - WPB-21759

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21759" title="WPB-21759" target="_blank">BugWPB-21759</a>  [iOS] Duplicate messages when importing universal backup
  </td></table>
  <br />
 

### Issue

In **ImportBackupUseCase** we filter out messages that already exist in the database:
 ```if !storedMessageIDs.contains(backupMessage.id) ... // then filter out```

The stored messages are returned by  **BackupLocalStore+Messages**:
  ```return Set(messages.compactMap(\.nonce).map(\.uuidString))```

  This returns UUIDs in uppercase format (e.g., "A1B2C3D4-...")

  But backup message IDs use `.transportString()` which is lowercase (e.g., "a1b2c3d4-...")

  The comparison is case-sensitive, so "a1b2c3d4-..." doesn't match "A1B2C3D4-...", causing the filter to fail and allowing duplicates through.

### Solution

  ~Change from~
 ~``` return Set(messages.compactMap(\.nonce).map(\.uuidString))```~

 ~to:~
  ~```return Set(messages.compactMap(\.nonce).map { $0.transportString() })```~

  ~This ensures both stored and backup message IDs use the same lowercase format for comparison.~

**Use UUIDs for comparison instead of Strings**

---

### Checklist

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Settings/Backup/BackupLocalStore/BackupLocalStore+Messages.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Settings/Backup/BackupLocalStore/BackupLocalStore+Messages.swift
@@ -33,13 +33,13 @@ extension BackupLocalStore {
         }
     }
 
-    func fetchAllMessageIDs() async throws -> Set<String> {
+    func fetchAllMessageIDs() async throws -> Set<UUID> {
         try await backupContext.perform { [backupContext] in
             let fetchRequest = ZMMessage.fetchRequest()
             fetchRequest.propertiesToFetch = ["nonce_data"]
 
             let messages = try backupContext.fetch(fetchRequest) as! [ZMMessage]
-            return Set(messages.compactMap(\.nonce).map { $0.transportString() })
+            return Set(messages.compactMap(\.nonce))
         }
     }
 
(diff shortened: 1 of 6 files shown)


caldrian (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Backup/BackupLocalStore/BackupLocalStore+Messages.swift
```diff
@@ -39,7 +39,7 @@ extension BackupLocalStore {
             fetchRequest.propertiesToFetch = ["nonce_data"]
 
             let messages = try backupContext.fetch(fetchRequest) as! [ZMMessage]
-            return Set(messages.compactMap(\.nonce).map(\.uuidString))
+            return Set(messages.compactMap(\.nonce).map { $0.transportString() })
```
netbe: thought: I wonder if we can have a type for messageNonce so comparing nonce hides the details of id being lowercase or uppercase

netbe (approved):
LGTM;)

wire-ios/Wire-iOS/Sources/UserInterface/Settings/Backup/BackupLocalStore/BackupLocalStore+Messages.swift
David-Henner: It's a core data attribute so it needs to stay as a UUID. Using a custom type would likely mean we'd need to make the nonce property private and encapsulate it in the custom type. So essentially might result in a couple of changes in different places to get it to be consistently adopted. But I like the thought. 

Also I guess it doesn't strictly need to be used everywhere and we could introduce it in the backup context

David-Henner: changed to **UUID**

PR #3894, opened 2025-11-17
fix: use correct removal key for 1:1 conversation - WPB-20294

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20294" title="WPB-20294" target="_blank">BugWPB-20294</a>  [iOS] not establishing 1:1 conversation using the correct removal key
  </td></table>
  <br />
 

### Issue

Currently the iOS app does not use the correct removal keys (from backend) when establish a 1:1 conversation. This PR fixes this.

Pass the mlsPublicKeys when establishing a 1:1 conversation

1. add missing MLSPublicKeys for APIV6 for `getMLSOneToOneConversation`
2. renamed MLSPublicKeys to p512 to p521
3. removed ed488 key as it's not returned by backend (keep stored one)
4. use OneOnOneResolver (not Legacy one) when opening a 1:1 conversation

### Testing

N/A
---

### Checklist


wire-ios-data-model/Source/MLS/MLSActionsProvider.swift
```diff
@@ -45,11 +45,6 @@ protocol MLSActionsProviderProtocol {
         in context: NotificationContext
     ) async throws -> [WireDataModel.KeyPackage]
 
-    func sendCommitBundle(
```
netbe, line 48: this is deadcode

WireDomain/Sources/WireDomain/Synchronization/OneOnOneResolver.swift
```diff
@@ -72,156 +72,170 @@ public struct OneOnOneResolver: OneOnOneResolverProtocol {
...
-
+        
         guard let userID else {
             throw Error.failedToActivateConversation
         }
-
+        
         // Sync the user MLS conversation from backend.
-        let mlsGroupID = try await pullMLSOneOnOneSync.pull(
+        let (mlsGroupID, mlsPublicKeys) = try await pullMLSOneOnOneSync.pull(
```
netbe, line 122: retrieves the mlsPublicKeys from backend

WireDomain/Sources/WireDomain/Synchronization/OneOnOneResolver.swift
```diff
@@ -72,156 +72,170 @@ public struct OneOnOneResolver: OneOnOneResolverProtocol {
...
     ) async {
+        
+        
+        let keys = mlsPublicKeys.flatMap { WireDataModel.BackendMLSPublicKeys(removal: $0.toDataModel()) }
+        
         do {
             try await setupMLSGroup(
                 mlsConversation: mlsConversation,
                 groupID: mlsGroupID,
+                mlsPublicKeys: keys,
```
netbe, line 169: pass the publicKeys to establish the 1:1 conversation

WireDomain/Sources/WireDomain/Synchronization/OneOnOneResolver.swift
```diff
@@ -84,29 +87,33 @@ public struct OneOnOneResolver: OneOnOneResolverProtocol {
...
+        return action
     }
 
-    private func resolveMLSConversation(for user: ZMUser) async throws {
+    @discardableResult
+    private func resolveMLSConversation(for user: ZMUser) async throws -> MLSGroupID {
         WireLogger.conversation.debug("Should resolve to mls 1-1 conversation")
 
-        let userID = await context.perform {
+        let userID = await context.perform { [user] in
```
David-Henner: Capturing a managed object won't solve the `@Sendable` warning. I think the only way to silence it is to pass the object ID and fetch it in the context within the perform block.

WireDomain/Sources/WireUpdateEventCoding/Models/UserEvent/StorableUserClientAddEvent.swift
```diff
@@ -35,10 +35,10 @@ struct StorableUserClientAddEvent: Equatable, Codable, Sendable {
             mlsPublicKeys: value.client.mlsPublicKeys.map {
                 StorableMLSPublicKeys(
                     ed25519: $0.ed25519,
-                    ed448: $0.ed448,
+                    ed448: nil,
                     p256: $0.p256,
                     p384: $0.p384,
-                    p512: $0.p512
+                    p521: $0.p521
                 )
```
David-Henner, line 42: why keep `ed448` if it's deprecated ?

WireNetwork/Sources/WireNetwork/APIs/Rest/ConversationsAPI/ConversationsAPIV6.swift
```diff
@@ -18,4 +18,65 @@
 
 class ConversationsAPIV6: ConversationsAPIV5 {
     override var apiVersion: APIVersion { .v6 }
+
+    // https://nginz-https.anta.wire.link/v12/api/swagger-ui/#/default/get-one-to-one-mls-conversation
```
David-Henner: is it safe to expose this link publicly?

jullianm (approved):
LGTM

WireDomain/Sources/WireUpdateEventCoding/Models/UserEvent/StorableUserClientAddEvent.swift
netbe: so here I wasn't sure to remove all the way down as this is stored in db. that's why I kept it as it is to be safe, what do you think?

WireNetwork/Sources/WireNetwork/APIs/Rest/ConversationsAPI/ConversationsAPIV6.swift
netbe: oups, i'll remove
```suggestion
```

WireDomain/Sources/WireUpdateEventCoding/Models/UserEvent/StorableUserClientAddEvent.swift
David-Henner: Okay, I was thinking it would be safe to keep in order to avoid the risk of breaking decoding from data in case there's a value for `ed448`. Since it's optional it's alright to set it to `nil` going forward. Seems like the safest choice if we assume it might break otherwise, but maybe it works if it's removed entirely ?

David-Henner (approved):
looks good !

PR #3897, opened 2025-11-17
fix: Call UI not maximizing after account switch - WPB-20468

KaterinaWire:
<table>
<td>
  </td></table>
  <br />
 

### Issue
When receiving a call on an inactive account, the app switches to the correct account, but only shows the green top bar.

From the UI perspective the flow is
  1. Receiving an incoming call on an inactive account and tapping the CallKit notification
  2. App opens and switches to the inactive account
  3. Calling UI briefly appears showing "Connecting..." state
  4. Calling UI is dismissed/covered when conversation view appears
  5. User sees only the conversation without the calling UI overlay

The calling UI is shown **before** the conversation navigation completes, and then the conversation appearing dismisses the calling UI.

```
func showConversation(
        _ conversation: ZMConversation,
        at message: ZMConversationMessage? = nil,
        in session: ZMUserSession
    ) {
        guard !conversation.isDeletedRemotely else {
            return
        }

        activateAccount(for: session) {
            self.presentationDelegate?.showConversation(conversation, at: message)
        }
    }
```
`updateActiveCallPresentationState()` is called inside the loadSession flow (activateAccount ->select() -> loadSession ) **before** the completion executes to show the conversation. The conversation navigation then interferes with the already shown calling UI.

### Solution

Display the calling UI after the conversation is fully visible.

1. Post notification when conversation view appears.
2. Observe notification in `SessionManager.activateAccount()` and triggers `updateActiveCallPresentationState()` at the right time - after conversation is visible

### Testing

  - Receive call on inactive account → ✅ Switches account and shows full call UI immediately

---

### Checklist

---


wire-ios-sync-engine/Source/SessionManager/SessionManager+Push.swift
```diff
@@ -112,6 +112,19 @@ extension SessionManager: UNUserNotificationCenterDelegate {
...
+                    queue: .main
+                ) { [weak self] _ in
+                    guard let self else { return }
+
+                    if let observer = conversationVisibleObserver {
+                        NotificationCenter.default.removeObserver(observer)
+                        conversationVisibleObserver = nil
+                    }
+                    presentationDelegate?.updateActiveCallPresentationStateIfNeeded()
+                }
```
johnxnguyen: question: can we move this inside the user session? The session manager shouldn't be concerned about this so it'd be good to avoid it if possible.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+ViewLifeCycle.swift
```diff
@@ -41,5 +41,15 @@ extension ConversationViewController {
...
 
         syncCellsState()
+        notifyConversationDidBecomeVisible()
+
+    }
+
+    private func notifyConversationDidBecomeVisible() {
+        NotificationCenter.default.post(
+            name: .conversationDidBecomeVisible,
+            object: conversation
```
johnxnguyen: suggestion: it doesn't seem necessary to pass the conversation, perhaps you can remove it?

wire-ios-sync-engine/Source/SessionManager/SessionManager+Push.swift
KaterinaWire: Well, I can share my thoughts on why I would prefer to keep it in `SessionManager`, but I'm open to your suggestions.
SessionManager owns presentationDelegate that already has API for UI communication in a context of account coordination/switching. 
This is a use case specifically for switching between accounts and SessionManager orchestrating it.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+ViewLifeCycle.swift
KaterinaWire: Yes, removed

wire-ios-sync-engine/Source/SessionManager/SessionManager+Push.swift
KaterinaWire: I also moved the code around and added more comments

johnxnguyen (commented):
I'm not a big fan of posting notifications like this but I understand an alternative solution must be tricky. Just so I understand, what's stopping us from awaiting the completion of showing the conversation directly? i.e something like 

```
showConversation(completion: {
  // do something
})
```

wire-ios/Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -510,7 +510,6 @@ extension AppRootRouter {
             sessionManager.processPendingURLActionDoesNotRequireAuthentication()
         case .authenticated:
             // This is needed to display an ongoing call when coming from the background.
```
netbe: todo: it seems the comment is for the removed code, remove comment

KaterinaWire: removed 🗑️

KaterinaWire:
> I'm not a big fan of posting notifications like this but I understand an alternative solution must be tricky. Just so I understand, what's stopping us from awaiting the completion of showing the conversation directly? i.e something like
> 
> ```
> showConversation(completion: {
>   // do something
> })
> ```

@johnxnguyen This wasn't my preferred solution either, but it was more robust and didn't require significant code refactoring that might reveal other hidden issues.
We trigger showConversation(...) on the callCenterDidChange():
```
func showConversation(
        _ conversation: ZMConversation,
        at message: ZMConversationMessage? = nil,
        in session: ZMUserSession
    ) {
        guard !conversation.isDeletedRemotely else {
            return
        }

        activateAccount(for: session) {
            self.presentationDelegate?.showConversation(conversation, at: message)
        }
    }
```

As far as I understand your question, you were asking if it is possible to add completion to the showConversation. The problem is that the conversation presentation happens through multiple asynchronous layers. 
The `presentationDelegate` transfers control to the `ZClientViewController`, which is well known for its ability to multitask. `ZClientViewController` calls select(conversation:scrollTo:focusOnView:animated:).
The method launches an async Task that:
      - dismisses modal controllers
      - navigates through the coordinator
      - updates the view model
And it returns before all these completes.
Eventually it presents the `ConversationViewController` and triggers `ConversationViewController.viewDidAppear`. 
To make it work I need to pass completion handler to the ZClientViewController.select(conversation:...) -> pass it into the async Task -> pass it to the coordinator navigation ->pass it to the view model -> somehow get it to the ConversationViewController -> call it in viewDidAppear.

johnxnguyen:
> > I'm not a big fan of posting notifications like this but I understand an alternative solution must be tricky. Just so I understand, what's stopping us from awaiting the completion of showing the conversation directly? i.e something like
> > ```
> > showConversation(completion: {
> >   // do something
> > })
> > ```
> 
> @johnxnguyen This wasn't my preferred solution either, but it was more robust and didn't require significant code refactoring that might reveal other hidden issues. We trigger showConversation(...) on the callCenterDidChange():
> 
> ```
> func showConversation(
>         _ conversation: ZMConversation,
>         at message: ZMConversationMessage? = nil,
>         in session: ZMUserSession
>     ) {
>         guard !conversation.isDeletedRemotely else {
>             return
>         }
> 
>         activateAccount(for: session) {
>             self.presentationDelegate?.showConversation(conversation, at: message)
>         }
>     }
> ```
> 
> As far as I understand your question, you were asking if it is possible to add completion to the showConversation. The problem is that the conversation presentation happens through multiple asynchronous layers. The `presentationDelegate` transfers control to the `ZClientViewController`, which is well known for its ability to multitask. `ZClientViewController` calls select(conversation:scrollTo:focusOnView:animated:). The method launches an async Task that: - dismisses modal controllers - navigates through the coordinator - updates the view model And it returns before all these completes. Eventually it presents the `ConversationViewController` and triggers `ConversationViewController.viewDidAppear`. To make it work I need to pass completion handler to the ZClientViewController.select(conversation:...) -> pass it into the async Task -> pass it to the coordinator navigation ->pass it to the view model -> somehow get it to the ConversationViewController -> call it in viewDidAppear.

@KaterinaWire yep I see, thanks for the elaboration.

johnxnguyen (approved):

netbe (approved):
lgtm:)

PR #3907, opened 2025-11-19
fix: file/folder naming issues - WPB-21700

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21700" title="WPB-21700" target="_blank">BugWPB-21700</a>  [iOS] Renaming a file with . as filename, disappears the file 
  </td></table>
  <br />
 

### Issue

This PR fixes some weird behaviors when dealing with "." in file/folder naming.
It introduces a Validator object to handle validation rules on the input.

### Testing

_Describe how to test._

---

### Checklist

---


WireMessaging/Sources/WireMessagingData/WireCells/NodesAPI/RestAPI.swift
```diff
@@ -90,7 +90,7 @@ final class RestAPI: Sendable {
 
         let parameters = RestActionParameters(
             awaitStatus: .finished,
-            awaitTimeout: "60s",
+            awaitTimeout: "5s",
```
jullianm, line 93: some requests would take 60s to timeout, it was mostly due to using "." in the naming so I believe it's fixed now that we introduced the new validation rules but just in case, we reduced the delay which is also the timeout set on the Web client.

WireMessaging/Sources/WireMessagingDomain/WireCells/UseCases/WireCellsRenameNodeUseCase.swift
```diff
@@ -50,15 +50,23 @@ package struct WireCellsRenameNodeUseCase: WireCellsRenameNodeUseCaseProtocol {
...
         }
 
-        let pathExtension = url.pathExtension
-        let directory = url.deletingLastPathComponent()
-        let targetPath = directory.appendingPathComponent("\(newFilename).\(pathExtension)")
+        let targetPath: URL
+
+        if isFolder {
+            let directory = url.deletingLastPathComponent()
+            targetPath = directory.appendingPathComponent("\(newFilename)")
```
jullianm, line 64: fix: ensuring the full folder name is displayed when dealing with `folders`

WireMessaging/Sources/WireMessagingData/WireCells/NodesAPI/RestAPI.swift
samwyndham: Hopefully they'll be logging timeouts on the backend to see if this is an issue. Five seconds seems too short but if the web is doing that I guess it's fine

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/CreateFolderViewModel.swift
```diff
@@ -26,15 +26,22 @@ private typealias Strings = L10n.Localizable.Conversation.WireCells
...
 @MainActor
 final class CreateFolderViewModel: ObservableObject {
+
     @Published var folderNameInput: String = ""
     @Published var errorMessage: String?
     @Published var isLoading: Bool = false
     @Published var isFocused: Bool = true
     @Published var didCreate: Bool = false
 
+    var isCreateDisabled: Bool {
```
samwyndham, line 36: Praise: moving this to the view model

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FileRenameViewModel.swift
```diff
@@ -145,18 +149,26 @@ final class FileRenameViewModel: ObservableObject {
...
             errorMessage = nil
+        case let .failure(failure):
+            isInputValid = false
+            switch failure {
+            case .tooLong:
+                errorMessage = inputTooLongErrorMessage
+            case .slashCharacter:
+                errorMessage = Strings.Files.RenameFile.wrongCharacterError
+            default:
+                errorMessage = nil
```
samwyndham, line 175: Question: any plans to have other error messages?

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/CreateFolderViewModel.swift
```diff
@@ -83,18 +91,26 @@ final class CreateFolderViewModel: ObservableObject {
 
     private func bindTextInput() {
         $folderNameInput
-            .sink { [weak self] input in
-                self?.validateTextInput(input)
-            }.store(in: &subscriptions)
+            .flatMap(filenameValidator.validate)
+            .sink(receiveValue: handleValidationResult)
```
samwyndham: Issue: I think here you implicitly capturing self causing a strong reference cycle. It's risky passing in a instance method when self is a class

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FilenameValidator.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    enum Failure: Error {
+        case tooLong
+        case dotPrefix
+        case singleDot
+        case slashCharacter
+        case empty
+    }
+
+    func validate(_ input: String) -> AnyPublisher<Result<Void, Failure>, Never> {
```
samwyndham, line 34: question: can this just be regular method instead of something that returns a publisher?

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FilenameValidator.swift
```diff
@@ -0,0 +1,52 @@
...
+        let result: Result<Void, Failure> = if input.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty {
+            .failure(.empty)
+        } else if input == "." {
+            .failure(.singleDot)
```
samwyndham: thought: maybe we don't need single dot as it will be caught by dot prefix

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FilenameValidator.swift
```diff
@@ -0,0 +1,52 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Combine
+
+struct FilenameValidator {
```
samwyndham, line 21: Praise: good to extract this

samwyndham (commented):
Nice work! I've got a few comments one of them it's an issue I think that needs to be fixed.

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/CreateFolderViewModel.swift
jullianm: right, always forget about these, fixed https://github.com/wireapp/wire-ios/pull/3907/commits/8e0b8eea9f9263834bf295fa38b0ffd206fd4979 thanks

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FilenameValidator.swift
jullianm: It can but since I'm binding the text input with a publisher I figured it'd make sense to go full Combine with that logic

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FilenameValidator.swift
jullianm: right, removed in https://github.com/wireapp/wire-ios/commit/8e0b8eea9f9263834bf295fa38b0ffd206fd4979 thanks

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FileRenameViewModel.swift
OlivellaO: suggestion: Should most of this be handled by the validator? Having to repeat handleValidationResult when we have to validate a text doesn't seems right.
Also if we need to add failure cases there will be a lot of modifications where we are using it, is not really scalable.
Couldn't we do something like a function that gets the text to validate and return success or failure (with an error message text)?

UI doesn't really need to know the type of error, just the error message to show if is not success, isn't it?

OlivellaO (approved):

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FileRenameViewModel.swift
jullianm: @samwyndham  https://github.com/wireapp/wire-ios/pull/3907/commits/993af1c3880ff6c5a9f9ea4e8cb3838d1575db8d made enum cases explicit and set error message when we have a "." prefix.

jullianm: @OlivellaO Maybe I'm missing your point and hopefully the following will clarify a bit my approach.

The idea is to validate the user input in real-time as he types using a Combine's stream, in that stream we :
1. receives the user's input as he types in 
2. validates it 
3. update properties to trigger a view update

View has no knowledge about the error message it just knows it has to display an error message.
Please note that in our new architecture approach the view model is a presentation layer so this is a good place to do some final setup before presenting the views.

The validator was extracted because it's used in two components.

OlivellaO: @jullianm I don't see anything wrong with your approach, and maybe i'm missing some specifics about the architecture approach. Also I don't know a lot specifically about Combine, but it's using the Observer pattern like other things i've used.

I just feel looking at the "repeated" code (only one localization string changes) that something can be done so we don't need to write all the switch statement on the view models every time we need the validator.

Maybe i'm overthinking it.

samwyndham (approved):
Nice work

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FileRenameViewModel.swift
jullianm: We could add an extension to the enum `var errorMessage: String { ... }` that switches over the cases and return the proper error message value.

jullianm: @OlivellaO I'll leave it as is for now and will add this suggestion to a Jira ticket Sam is about to create where we can list the things to improve (refactoring suggestions).

PR #3921, opened 2025-11-21
fix: don't crash when unable to parse blacklist - WPB-21680

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21680" title="WPB-21680" target="_blank">BugWPB-21680</a>  [iOS] Crash when checking blacklist
  </td></table>
  <br />
 

### Issue
When logging in to one of the QA environments the app would crash because it was unable to parse the client blacklist and would throw an error during session loading but would not tear down the session, leading to an assertion failure.

The specs (see attached ticket) now state that we should assume the blacklist to be empty in case we are unable to parse it.

---

### Checklist

---


samwyndham (approved):
Nice work!

WireDomain/Sources/WireDomain/UseCases/IsBuildBlacklistedUseCase.swift
```diff
@@ -45,22 +45,24 @@ public struct IsBuildBlacklistedUseCaseImpl: IsBuildBlacklistedUseCase {
...
 
-        guard let minLegalVersion = Int(blacklist.minimumLegalBuildNumber) else {
-            throw IsBuildBlacklistedUseCaseError.decodingFailed(
-                message: "blacklist version '\(blacklist.minimumLegalBuildNumber)' must be an integer"
-            )
+            return currentVersion < minLegalVersion || blacklist.illegalBuildNumbers.contains(String(currentVersion))
+        } catch {
+            // As per specs, if there is any failure obtaining the blacklist,
+            // whether it is missing, can't be decoded, or otherwise, then
+            // consider it empty (i.e all cliets valid).
```
netbe:
```suggestion
            // consider it empty (i.e all clients valid).
```

wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
```diff
@@ -184,7 +184,7 @@ final class UserSessionLoader {
         )
 
         // Check if this build is blacklisted.
-        if try await isBuildBlacklisted(userSession: userSession) {
```
netbe, line 187: question: I guess this also fix the case where if we can't parse the usersession is not tearDown and so we crash, right?

netbe (approved):
LGTM;)

wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
johnxnguyen: correct

PR #3933, opened 2025-11-24
chore: adding tags to the UITests - WPB-20997

findms:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20997" title="WPB-20997" target="_blank">TaskWPB-20997</a>  [iOS/QA] Explore Tags in XCTest
  </td></table>
  <br />
 

<hr />

<h2>Summary</h2>
<p><em>Briefly describe about this PR changes.</em></p>

- Added configuration in the env variable to run critical tests if test name has word critical
Configuration for critical test run added so basically it check if test name has critical(ignore casing) then running critical testplan will run all the tests which has critical word, if running normal configuration(as before) it will run all tests.

<hr />

<h2>Testing</h2>
<p><em>Describe how to verify the changes locally. Attach screenshots of local execution if relevant.</em></p>

tests name updated with critical like this
image

On running test using command `bundle exec fastlane run_uitests configuration:UITest-critical` , only test with critical word will run not others.

<h2>Additional Information</h2>
<p><em>any more info, add here.</em></p>

**How to run?**
To run tests marked with tags critical here is the command:
`bundle exec fastlane run_uitests configuration:UITest-critical`

if configuration is not being passed, it will run all tests as regression.
`bundle exec fastlane run_uitests`


netbe (commented):
interesting way to use kind of "tags" in XCTest as it would be in SwiftTesting. For our usecase, I wonder if it's worth it to have this rather to rely on a specific testplan that would contain the critical tests: the same way we have security tests.

what do you think @johnxnguyen ?

johnxnguyen:
> interesting way to use kind of "tags" in XCTest as it would be in SwiftTesting. For our usecase, I wonder if it's worth it to have this rather to rely on a specific testplan that would contain the critical tests: the same way we have security tests.
> 
> what do you think @johnxnguyen ?

@findms @netbe 

It's a shame that you can't add tags to XCTest cases like can for Swift Testing as that would be the preferred approach. For me, the better solution for XCTests is to have specific test plans (note that a single test can be in multiple test plans) as this will make it easier for people to read and it aligns with what we currently do already. I'd be cautious about adding additional variations for organizing and running tests.

netbe:
@findms should we close this? @johnxnguyen

findms:
@netbe @johnxnguyen 
I am going to revive this again as picked this task in discipline.

findms:
No longer needed as handled here https://github.com/wireapp/wire-ios/pull/4823

PR #3909, opened 2025-11-19
fix: new user unable to send messages - WPB-21849

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21849" title="WPB-21849" target="_blank">BugWPB-21849</a>  [iOS] New user unable to send messages after creating account
  </td></table>
  <br />
 

### Issue

Newly created user is unable to send / receive messages. They get the error `SQL error or missing database` from core crypto. This usually indicates that the key being used to decrypt the database is wrong.

### Cause

I was able to reproduce with a fresh user and got some useful logs

**Main app - filtered by [keychain] and [core-crypto] tags**
```
2025/11/19 09:40:08:721  [keychain] - [INFO] deleting item (PasscodeKeychainItem(itemIdentifier: "com.wire.passcode-ED656D53-30F9-401C-8E1B-E0321959F821"))[process_name:Wire][self_client_id:718a081***][process_id:2290]
2025/11/19 09:40:08:721  [keychain] - [ERROR] fetching items (class: keys) failed: osstatus -25300[process_name:Wire][self_client_id:718a081***][process_id:2290]
2025/11/19 09:42:00:844  [keychain] - [INFO] fetching item (com.wire.mls.key)[process_name:Wire][self_client_id:718a081***][process_id:2290][account_id:c93de3d***]
2025/11/19 09:42:00:846  [keychain] - [ERROR] fetching item (com.wire.mls.key) failed: osstatus -25300[process_name:Wire][self_client_id:718a081***][process_id:2290][account_id:c93de3d***]
2025/11/19 09:42:00:846  [core-crypto] - [INFO] Skip core crypto key migration[process_name:Wire][self_client_id:718a081***][process_id:2290][account_id:c93de3d***]
2025/11/19 09:42:00:848  [keychain] - [INFO] fetching item (com.wire.mls.key)[process_name:Wire][self_client_id:718a081***][process_id:2290][account_id:c93de3d***]
2025/11/19 09:42:00:849  [keychain] - [ERROR] fetching item (com.wire.mls.key) failed: osstatus -25300[process_name:Wire][self_client_id:718a081***][process_id:2290][account_id:c93de3d***]
2025/11/19 09:42:00:849  [core-crypto] - [INFO] Rotating core crypto key...[process_name:Wire][self_client_id:718a081***][process_id:2290][account_id:c93de3d***]
[...]

PR #3950, opened 2025-11-26
fix: core crypto key migration flags - WPB-21849

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21849" title="WPB-21849" target="_blank">BugWPB-21849</a>  [iOS] New user unable to send messages
  </td></table>
  <br />
 

### Issue

In a recent PR (https://github.com/wireapp/wire-ios/pull/3909) a fix was made to how core crypto key migrations are handled. Since then, we don't receive notifications and can't share files via the share extension. 

Context:
- We return early in the extensions if there are migrations pending.
- Migrations as marked as pending by default. 

Causes:
- On fresh install, when migrations aren't needed, they should be marked as done. This was missing for one of the migrations.
- In the NSE, the `guard` statement was wrong. 🤦 

---

### Checklist

### DIFF
--- a/WireDomain/Sources/WireDomain/Notifications/Components/NSEUserScope.swift
+++ b/WireDomain/Sources/WireDomain/Notifications/Components/NSEUserScope.swift
@@ -147,7 +147,7 @@ final class NSEUserScope: Component<NSEUserScopeDependency> {
             throw Failure.mainAppRequired(message: "cryptobox migration required")
         }
 
-        guard coreCryptoKeyMigrationManager.isAnyMigrationRequired else {
+        guard !coreCryptoKeyMigrationManager.isAnyMigrationRequired else {
             throw Failure.mainAppRequired(message: "core crypto key migration required")
         }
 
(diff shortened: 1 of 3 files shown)


WireDomain/Sources/WireDomain/Notifications/Components/NSEUserScope.swift
```diff
@@ -147,7 +147,7 @@ final class NSEUserScope: Component<NSEUserScopeDependency> {
             throw Failure.mainAppRequired(message: "cryptobox migration required")
         }
 
-        guard coreCryptoKeyMigrationManager.isAnyMigrationRequired else {
+        guard !coreCryptoKeyMigrationManager.isAnyMigrationRequired else {
```
netbe, line 150: 🫠 ah we missed this in review...

netbe (approved):
LGTM;)

KaterinaWire (approved):

PR #3956, opened 2025-12-01
feat: integrate collabora - WPB-21651

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21651" title="WPB-21651" target="_blank">TaskWPB-21651</a>  [iOS] Collabora integration
  </td></table>
  <br />
 

### Issue

This PR integrates Collabora allowing editing of certain file types. The backend returns a URL which we open in a WebView. Currently we filter the types which can be edited locally but this will soon be updated on the backend so that the backend tells us which files can be edited with collabora.

https://github.com/user-attachments/assets/04ab4cbc-7f75-4f46-b02f-734a0b159c13

### Testing

1. navigate to cells enabled conversation.
2. Send an editable file of one of the following types: ODF, DOCX, XLSX, PPTX
3. navigate to the files list view
4. Tap the menu button and select edit file.
5. Make some changes and close the file.
6. After a few seconds, view the file. You should notice that it has been updated with your changes. (It takes some seconds for the updates to be applied.)

---

### Checklist


WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FilesViewProtocol.swift
```diff
@@ -66,7 +66,8 @@ extension FilesViewProtocol {
             viewModel: viewModel.itemViewModel(index: index),
             canRenameFile: !isBrowsing, // action not allowed when browsing files
             canEditTags: !isBrowsing, // action not allowed when browsing files
-            canMoveToFolder: !isBrowsing && viewModel.isFoldersEnabled // action not allowed when browsing files
+            canMoveToFolder: !isBrowsing && viewModel.isFoldersEnabled, // action not allowed when browsing files
+            canEditFile: !isBrowsing // action not allowed when browsing files
```
WilhelmOks: thought: with that many `canDoX` boolean parameters maybe we can improve this by putting them into an array and a single parameter. Something like this:

```
enum ItemAction {
    case rename, editTags, moveToFolder, editFile
}

//...

    func theFunction(
        //...
        supportedItemActions: isBrowsing ? [] : [.rename, .editTags, .moveToFolder, .editFile],
        //...
    )
```

WilhelmOks (approved):

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FilesViewProtocol.swift
jullianm: Agreed. @samwyndham maybe we don't need to do these changes in this PR and add it to the Jira ticket.

WireMessaging/Sources/WireMessagingData/WireCells/NodesAPI/NodesAPI.swift
```diff
@@ -25,7 +25,8 @@ package enum NodesAPIError: Error {
     case moveFailed
 }
 
-package final actor NodesAPI: NodesAPIProtocol, WireCellsNodesRepositoryProtocol {
+package final actor NodesAPI: NodesAPIProtocol, WireCellsNodesRepositoryProtocol,
+    WireCellsEditingURLRepositoryProtocol {
```
jullianm, line 29: praise: it's a good idea to use protocol composition, maybe we should generalize this approach as improvements (later on)

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsAttachmentsPreviewItemViewModel.swift
```diff
@@ -143,13 +144,29 @@ final class WireCellsAttachmentsPreviewItemViewModel: ObservableObject {
...
+    func startPolling() {
+        pollingTask?.cancel()
+        pollingTask = Task { [weak self] in
+            while !Task.isCancelled {
+                await self?.refresh()
+
+                try? await Task.sleep(for: .seconds(30))
+            }
+        }
+    }
```
jullianm, line 156: suggestion/question: I used a Combine publisher for the polling mechanism in my PR, I feel like it's a bit simpler to use a Publisher for that, no need to have a reference to a task and cancel it when view disappears. the stream automatically cancels when view is dismissed and also no need for a Task.sleep too. My suggestion is to follow that approach, what do you think ?

```
        Timer.publish(every: .thirtySeconds, on: .main, in: .common)
            .autoconnect()
            .sink { [weak self] _ in
                Task { await self?.fetch() }
            }.store(in: &subscriptions)
```

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/EditFileView.swift
```diff
@@ -0,0 +1,99 @@
...
+            VStack {
+                switch viewModel.state {
+                case .idle:
+                    EmptyView()
+                case .loading:
+                    ProgressView()
+                case let .loaded(url):
+                    WebView(url: url)
+                case let .error(title, message):
+                    MoveToFolderEmptyStateView(
```
jullianm, line 44: suggestion:  rename the file to `EmptyStateView`

jullianm (approved):
Looks good, nice work, approving with some suggestions

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/EditFileView.swift
samwyndham: Agree. I will fix in a follow up PR. Thanks

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsAttachmentsPreviewItemViewModel.swift
samwyndham: I'll look into doing this in a follow up PR. There was an issue using timer and I'm trying to remember what it was.

PR #3962, opened 2025-12-02
fix: unable to select backup from Google Drive - WPB-21859

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21859" title="WPB-21859" target="_blank">BugWPB-21859</a>  [iOS] Can't select Wire backup file in Google Drive
  </td></table>
  <br />
 

### Issue
If I have a backup file stored on Google Drive, when I want to restore a backup I can navigate to Google Drive via the native iOS document picker but all Wire backup files are greyed out and can't be selected.

The reason is because we restricted allowed file types by configuration the document picker with Wire backup UTIs (files with `.wbu`, .`.iOS-wbu`, and `.iOS_wbu` extensions). For some reason, Google Drive doesn't understand these. (Side note, this issue has been experienced by others, seems to be a limitation of Google Drive on iOS).

The fix/workaround is to allow all file times for selection and then validate the correct file type after picking. The universal backup library already validates the file type so to achieve this we simply catch the "wrong file type" error and show an alert.

There are however some further changes: when picking a file hosted on iCloud, the URL of the selected file will already be materialized (i.e downloaded and pointing to local data). But in the case of Google Drive (or any other 3rd party cloud file system) the URL may be a placeholder for remote content. Therefore we must materialize the URL ourselves and this may take some time. Therefore I have also adjusted the import progress view to display a step for loading the backup file.

### Testing

1. Install Google Drive on your device.
2. Place a large backup file in Google Drive
3. Place another file (such as a document or image) in Google Drive
4. Log in and start the import backup flow
5. Navigate to Google Drive in the document picker and select a non-backup file
6. Assert you see an error stating "Wrong file type".
7. Now select the wire backup
8. Assert that it successfully imports the backup.

---

### Checklist

---


netbe (approved):
LGTM;)

WireUI/Sources/WireSettingsUI/Account/BackupImportExport/Import/ImportBackupView.swift
```diff
@@ -44,13 +44,19 @@ struct ImportBackupView: View {
             .foregroundStyle(Color.primaryText)
             .fileImporter(
                 isPresented: $isFileImporterPresented,
-                allowedContentTypes: WireBackupUTIs,
+                // Workaround: Google Drive doesn't recognize Wire backup
+                // UTIs so if we only allow them, the user won't be able
+                // to select Wire backups stored in Google Drive.
+                // Intesad, we allow any file and then we'll validate the
```
David-Henner:
```suggestion
                // Instead, we allow any file and then we'll validate the
```

David-Henner (approved):

PR #3971, opened 2025-12-03
fix: can't send screenshot via share extension - WPB-20501

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20501" title="WPB-20501" target="_blank">BugWPB-20501</a>  [iOS] Can't send screenshot via ShareExt in iOS 26
  </td></table>
  <br />
 

### Issue

Given I'm on iOS 26, when I take a screenshot and try to share via Wire, then it always fails.

The reason is that we try to load the attachment as an `UIImage` but it's actually `Data`. It's likely that detail changed from iOS 18 to 26.

The solution is to handle both cases.

### Testing

1. Take a screenshot anywhere in iOS
2. Tap the share button, select Wire
3. Choose a conversation and share it.
4. Assert it succeeds.

---

### Checklist

---


netbe (approved):
LGTM;)

caldrian (approved):

PR #3991, opened 2025-12-09
fix: No CallKit notification for incoming call - WPB-21875

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21875" title="WPB-21875" target="_blank">BugWPB-21875</a>  [iOS] No CallKit notification for incoming call for second inactive account
  </td></table>
  <br />
 

### Issue

No **CallKit** notification (just a regular notification) for incoming call for second inactive account when CallKit is enabled. The issue is reproduced after terminating the app -> opening the app -> moving the app to the background.

One of the criteria checked for displaying CallKit notifications is `isUserSessionLoaded`.
The problem: `loadedUserSessions` only contains accounts that are loaded into backgroundUserSessions. When `Account #2 `is not active after app restart, it's not loaded into background sessions, so `isUserSessionLoaded` returns **false** for `Account #2`. As a result, we display the regular notification instead

### Solution

Remove this check.
This condition is too restrictive because we load the session while processPendingCallEvents:
  ```
if let session = await withSession(for: account) {
      await session.processPendingCallEvents()
  }
```

Additionally, one of the steps involves checking the `VerifyUserSessionUseCase` to ensure correct user authentication.

### Testing

- enable CallKit
- make the account that will receive the call inactive
- terminate the app
- open the app
- move the app to the background
- receive an incoming call for an inactive account

### Checklist

---


netbe (approved):
LGTM;)

johnxnguyen (changes requested):
Looks good, but we should also remove the relevant code in `VoipPushHelper` where we set the loaded users sessions.

KaterinaWire:
> Looks good, but we should also remove the relevant code in `VoipPushHelper` where we set the loaded users sessions.

@johnxnguyen Removed from `VoipPushHelper` and also legacy code from  the `NotificationSession`

johnxnguyen (approved):

PR #3999, opened 2025-12-11
feat: multipart message reply - WPB-20938

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20938" title="WPB-20938" target="_blank">TaskWPB-20938</a>  [iOS] Support reply for multipart message
  </td></table>
  <br />
 

### Issue

This PR introduces the ability for the user to reply to a multipart message, the following use cases were covered:
- Reply to a single video/image attachment -> display a small thumbnail preview
- Reply to a single (other than the above) attachment -> file type icon + file name
- Reply to multiple attachments -> generic icon + number of files attached

See video 

https://github.com/user-attachments/assets/4bc7e48d-f17a-4a41-8ce4-2cb10c434bcb

### Testing

- Add message with different attachments (use cases covered in the PR description)
- Reply to these messages
- On scroll, thumbnails already downloaded should show up immediately

### Checklist

---


wire-ios/Wire-iOS/Sources/UserInterface/Components/MessageReplyAttachmentsView.swift
```diff
@@ -0,0 +1,204 @@
...
+import WireDataModel
+import WireDesign
+import WireMessagingDomain
+
+/// A lightweight view that renders a preview for one or more message attachments.
+/// It supports:
+///  - Image and video thumbnail previews
+///  - Generic file previews with icons
+///  - Automatic preview loading and caching
+final class MessageReplyAttachmentsView: UIView {
```
jullianm, line 30: we could probably turn this into a SwiftUI view and interface it to use it with legacy UIKit code, didn't choose this approach here as I felt it could introduce some weird layout issues and delay the implementation of the feature.

OlivellaO (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Components/MessageReplyAttachmentsView.swift
samwyndham: I think you made the right choice. If we don't have access to the view controller directly, we shouldn't be messing with the view hierarchy

wire-ios/Wire-iOS/Sources/UserInterface/Components/MessagePreviewView.swift
```diff
@@ -268,6 +285,11 @@ final class MessagePreviewView: UIView {
             senderLabel.isAccessibilityElement = true
             senderLabel.accessibilityIdentifier = "SenderLabel_ReplyPreview"
         }
+
+        if message.isMultipart {
```
samwyndham, line 289: question: Is this view reused with table view cell reuse? If so, does it need to be cleaned up?

wire-ios/Wire-iOS/Sources/UserInterface/Components/MessageReplyAttachmentsView.swift
```diff
@@ -0,0 +1,204 @@
...
+
+        return imageView
+    }
+
+    private func makeLoadingOverlay() -> UIView {
+        let overlay = UIView()
+        overlay.backgroundColor = ColorTheme.Backdrop.background
+
+        let indicator = UIActivityIndicatorView(style: .medium)
+        indicator.transform = CGAffineTransform(scaleX: 0.5, y: 0.5)
```
samwyndham, line 174: question: Was the standard size really too big?

wire-ios/Wire-iOS/Sources/UserInterface/Components/MessageReplyAttachmentsViewModel.swift
```diff
@@ -0,0 +1,139 @@
...
+            .compactMap(UIImage.init(data:))
+            .handleEvents(receiveOutput: { [weak self] image in
+                guard let self else { return }
+                cache.setObject(image, forKey: cacheKey)
+            })
+            .map { PreviewImageInfo(image: $0, isVideo: isVideo) }
+            .sink { _ in } receiveValue: { [weak self] in
+                guard let self else { return }
+                previewImageInfo = $0
+            }.store(in: &subscriptions)
```
samwyndham: suggestion: Personally, I prefer that this was done with async / await API. I think it's easier to read. I think it's best to reserve combine for places where it's really better than async-await.

samwyndham (approved):
Nice work!

wire-ios/Wire-iOS/Sources/UserInterface/Components/MessagePreviewView.swift
jullianm: yes, this does need to be cleaned up as cell is being reused and results in asset showing up in reused cells on unrelated messages.

wire-ios/Wire-iOS/Sources/UserInterface/Components/MessageReplyAttachmentsView.swift
jullianm: I think it would be acceptable to keep the standard size however (in my opinion) it looks much better with a smaller size as the preview thumbnail is quite small already.

wire-ios/Wire-iOS/Sources/UserInterface/Components/MessageReplyAttachmentsViewModel.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/3999/commits/23bb23ed2e825128e191a8ec5935ee41409bee5b

PR #4000, opened 2025-12-11
fix: do not close webSocket if there is an ongoing call - WPB-21358

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21358" title="WPB-21358" target="_blank">BugWPB-21358</a>  [iOS] Audio drops after a while when the app is in background (screen locked)
  </td></table>
  <br />
 

### Issue

We close the webSocket connection when the app goes into the background. As a result, the app doesn't receive epoch changes during the call and can no longer decode audio.  So client loses audio in the middle of the call.

### Solution

Don't close the webSocket if there is an ongoing call. And close it when the user ends the call via CallKit.

### Testing

---

### Checklist

---


caldrian (approved):

wire-ios-sync-engine/Source/SessionManager/SessionManager+CallKitManagerDelegate.swift
```diff
@@ -89,4 +89,13 @@ extension SessionManager: CallKitManagerDelegate {
         }
     }
 
+    func didEndAllCalls() {
+        WireLogger.calling.info("all calls ended, suspending background tasks")
```
netbe:
```suggestion
        WireLogger.calling.info("all calls ended, suspending background tasks", attributes: .safePublic)
```

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+LifeCycle.swift
```diff
@@ -64,9 +64,15 @@ public extension ZMUserSession {
 
     @objc
     func applicationDidEnterBackground(_ note: Notification?) {
```
netbe, line 66: **thought:** I know it was like this, but I wonder if the responsability should not be within the SyncAgent.

We could observe applicationDidEnterBackground within SyncAgent and therefore test this.

something to keep in mind (non-blocking)

netbe (approved):
LGTM;) added a suggestion

PR #4006, opened 2025-12-12
fix: concurrency issue - WPB-21441

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21441" title="WPB-21441" target="_blank">StoryWPB-21441</a>  [iOS] [Integration] Toggle Apps in Conversation Flow
  </td></table>
  <br />
 

### Issue

This PR fixes an issue where `SearchDirectory.tearDown()` is called on a wrong thread from `deinit`.
Moreover, this PR contains a few more cosmetic changes.

---

### Checklist

---


johnxnguyen (approved):

wire-ios-sync-engine/Source/Use cases/SearchUsersUseCase.swift
```diff
@@ -16,30 +16,24 @@
...
-    ) async throws -> SearchResult
-}
-
-public class SearchUsersUseCase: SearchUsersUseCaseProtocol {
+public final class SearchUsersUseCase: SearchUsersUseCaseProtocol {
 
     // MARK: - Properties
 
-    private let context: NSManagedObjectContext
+    public let context: NSManagedObjectContext
```
netbe: question: why is this public now?

wire-ios-sync-engine/Source/Use cases/SearchUsersUseCase.swift
```diff
@@ -16,30 +16,24 @@
...
+
     private let searchDirectory: SearchDirectory
     private let isFederationUsageAllowed: Bool
     private var activeSearchTask: SearchTask?
     private let isMLSEnabled: Bool
 
     deinit {
-        searchDirectory.tearDown()
+        DispatchQueue.main.async { [searchDirectory] in
+            searchDirectory.tearDown()
```
netbe: suggestion: can we make the tearDown itself safe to be called from different threads?
The issue seems to be the access of the viewContext associated value

pass the searchUserObserverCenter as a dependency of SearchDirectory and then the teardown does not need to be called asynchronously, no?

netbe (commented):
left a suggestion, before approving, let me know what you think @caldrian

wire-ios-sync-engine/Source/Use cases/SearchUsersUseCase.swift
caldrian: `SearchUserObserverCenter` itself is not thread-safe, so even by having a reference to the instance without accessing `viewContext`, we would still modify the state of the instance which is typically accessed only from the main thread, from a background thread.

netbe (approved):
LGTM;)

wire-ios-sync-engine/Source/Use cases/SearchUsersUseCase.swift
caldrian: The `SearchResult` which is returned from the use case contains Core Data objects, which means they must be accessed from a certain thread. By providing the context in `SearchUsersUseCaseProtocol` a caller can use it to access the objects using `moc.perform { ... }`

wire-ios-sync-engine/Source/Use cases/SearchUsersUseCase.swift
caldrian: Moved the code into `SearchDirectory.tearDown()`.

wire-ios-sync-engine/Source/Use cases/SearchUsersUseCase.swift
caldrian: Moved `context` into `SearchResult`.

PR #4020, opened 2025-12-17
fix: messages stuck with stale delivery state - WPB-22214

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22214" title="WPB-22214" target="_blank">BugWPB-22214</a>  [iOS] - Messages shown as pending delivery and ordering of messages wrong - even though receiver received the message
  </td></table>
  <br />
 

### Issue

Messages sent in 1:1 conversations may sometimes end up showing as `sending` while they're already `sent`

### Cause

If a message is rendered with a `sending` delivery state, the message section will be cached in `ConversationTableViewDataSource`. When the delivery state changes the sections will be recalculated, but since the message section is cached the cache - containing the wrong delivery state - will be used.

### Solution

Use a cache identifier that includes deliveryState to invalidate the cache on state change

---

### Checklist

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
@@ -232,7 +232,7 @@ final class ConversationTableViewDataSource: NSObject {
                     }
 
                     // Section model uses nonce for DifferenceKit identity (not cache identifier)
-                    // This ensures the same message is recognized as the same section across delivery state changes
+                    // This ensures the same message is recognized as the same section across delivery state changes.
                     sections.append(ArraySection(
                         model: identifier.nonce,
                         elements: sectionController.tableViewCellDescriptions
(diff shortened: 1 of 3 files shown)


netbe (approved):
LGTM:) @David-Henner should we/ can we include this in 4.12 ?

johnxnguyen (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
```diff
@@ -55,7 +55,13 @@ final class ConversationTableViewDataSource: NSObject {
 
     var registeredCells: [AnyClass] = []
 
-    var sectionControllers = ThreadSafeDictionary<UUID, ConversationMessageSectionController>()
+    /// Cache for section controllers, keyed by message identifier
```
David-Henner, line 58:
```suggestion
    /// Cache for section controllers, keyed by message identifier.
```

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
```diff
@@ -214,8 +231,10 @@ final class ConversationTableViewDataSource: NSObject {
                         sectionController.recreateCellDescriptions(in: context)
                     }
 
+                    // Section model uses nonce for DifferenceKit identity (not cache identifier)
+                    // This ensures the same message is recognized as the same section across delivery state changes
```
David-Henner:
```suggestion
                    // This ensures the same message is recognized as the same section across delivery state changes.
```

PR #4021, opened 2025-12-17
fix: oneToOne call fails to connect when answered - WPB-21338

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21338" title="WPB-21338" target="_blank">BugWPB-21338</a>  [iOS] 1:1 call fails to connect when answered on iPhone; other devices continue to ring
  </td></table>
  <br />
 

### Issue

In some cases, when the user answers the call, the call can't be established. The problem can be easily reproduced by closing (killing) the app and receiving a call on **inactive** account.
This fix what implemented on iOS and AVS sides.

### Solution

1. On the AVS side: [Delay invocation of start/answer until AVS becomes ready](https://github.com/wireapp/wire-avs/releases/tag/10.1.36)
2. "AVS becomes ready" means the client has submitted to AVS the call configuration received from the backend. In this PR, I added a call to retrieve the call configuration and update AVS.

### Checklist

---


wire-ios-sync-engine/Source/SessionManager/SessionManager+CallKitManagerDelegate.swift
```diff
@@ -74,13 +74,28 @@ extension SessionManager: CallKitManagerDelegate {
...
+    /// - Parameter session: The user session to request config for
+    private func requestCallConfigIfNeeded(for session: ZMUserSession) async {
+        guard session != activeUserSession,
+              let callCenter = session.callCenter else {
+            WireLogger.calling.warn("Cannot request call config: callCenter not available")
+            return
+        }
+
+        WireLogger.calling.info("Proactively requesting call config for background session")
+        callCenter.requestCallConfig()
```
netbe: question: does this need to be finished before processPendingCallEvents?

We could start moving away from CallingRequestStrategy to have a `async` method

netbe (commented):
left a question

wire-ios-sync-engine/Source/SessionManager/SessionManager+CallKitManagerDelegate.swift
KaterinaWire: The [changes](https://github.com/wireapp/wire-avs/releases/tag/10.1.36) on AVS helps with this, they will wait for us. I was thinking to make requestCallConfig() as async, but AVS helps.
Ideally yes, we need to have a config before processing events. Does it need to be in this PR, not sure

netbe: ok but when do we set AVS is ready?

KaterinaWire: When we call `callCenter.requestCallConfig() `and receives the response from the backend we pass it to AVS using: 
```
public func update(callConfig: String?, httpStatusCode: Int) {
        wcall_config_update(handle, httpStatusCode == 200 ? 0 : EPROTO, callConfig ?? "")
    }
```
on the AVS side they added a check:
```
bool wcall_is_ready(struct calling_instance *inst,
		    int conv_type)
{
        if (!inst)
	        return false;

	switch (conv_type) {
	case WCALL_CONV_TYPE_ONEONONE:
	       return inst->call_config != NULL;

	default:
	       return true;
	}
}
```
it means once AVS receive call_config it can start processing events

netbe (approved):
LGTM;)

David-Henner (approved):

PR #4028, opened 2025-12-18
fix: slow request strategies tests - WPB-22487

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22487" title="WPB-22487" target="_blank">BugWPB-22487</a>  [iOS] fix request strategies slow tests
  </td></table>
  <br />
 

### Issue

While introducing ProteusClientSimulator the ReqeustStrategies tests became really slow (4-6 seconds) per test.
This is due to calling CoreCrypto instance on each setup of tests even if there is no cryptographic tests.

Solution:

Make initialization lazy so only the required tests setup CC. This will result in tests being finished between 45-90 seconds (pretty similar as before see 4.11)

Note: ultimately we would use a mock, but this would be another optimization.

### Testing

run tests

---

### Checklist

---

### DIFF
--- a/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoderTest.swift
+++ b/wire-ios-request-strategy/Sources/Synchronization/Decoding/EventDecoderTest.swift
@@ -69,7 +69,6 @@ class EventDecoderTest: MessagingTestBase {
         super.tearDown()
     }
 
-
     private func decryptAndStoreEvents(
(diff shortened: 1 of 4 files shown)


David-Henner (approved):
Thanks for fixing this ! I really should've noticed my changes were responsible for the slow tests 😬

KaterinaWire (approved):

PR #4034, opened 2025-12-19
fix: establish groups pending join if epoch is 0 - WPB-22496

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22496" title="WPB-22496" target="_blank">BugWPB-22496</a>  [iOS] Don't join pending group via external commit if epoch is 0
  </td></table>
  <br />
 

### Issue
It was discovered that there is a case where the iOS client tries to join an mls group via external commit (because the user is a participant of the conversation but not yet a member of the underlying mls group) even when the epoch is 0. If the epoch is 0 it's likely because the group has been reset by another user but they failed to establish it. In this case, the iOS client should establish the group, otherwise if they try to join by external commit it would likely not succeed.

Note: we do have a `reEstablishPendingGroup` method that handles this correctly, but it only is invoked when trying to send a message and only in the case that the mls reset event has been received. This is not the case when the group has been reset but not established and then a user logs into to a new iOS client, since all the mls groups would be get a `pendingJoin` status rather than `pendingJoinAfterReset` status.

### Testing
1. Be in an mls group but logged out.
2. Have another reset the group but not establish it (e.g via curl call to backend mls reset endpoint)
3. Log in
4. Assert that the groups can be joined.

---

### Checklist

---

### DIFF
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -876,13 +876,13 @@ public final class MLSService: MLSServiceInterface {
                     }
 
                     do {
-                        let (epoch, isSelfConversation) = await context.perform {
-                            (pendingGroup.epoch, pendingGroup.isSelfConversation)
+                        let epoch = await context.perform {
+                            pendingGroup.epoch
                         }
                         let conversationExists = try await self.conversationExists(
                             groupID: mlsGroupID
                         )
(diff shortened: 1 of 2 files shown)


typfel (approved):

netbe (approved):
Looks good,  @johnxnguyen  maybe as an improvement for later we factor out the reEstablish method to be reused in performPendingJoin?

PR #4052, opened 2025-12-23
feat: introduces API v14 to fetch new cells feature config - WPB-22364

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22364" title="WPB-22364" target="_blank">TaskWPB-22364</a>  [iOS] Support backend url configuration from feature flag settings
  </td></table>
  <br />
 

### Issue

This PR introduces API v14 which is required to fetch wire cells backend URL from a newly introduced feature config.

A key change in this PR is to resolve the cells server URL dynamically since we now get its value after pulling all the resources during initial sync. 

To avoid dealing with optionals in many parts of the codebase and keep the structure as close as it is currently, the solution was to inject a URL **resolver** closure in the factory which will be invoked at runtime by the `RestAPI` and the `AWSClient`.

Note: at the time of this PR, API v14 is still under development

### Testing

- on dev env (IMAI), fresh launch the app with API v13 or below, cells related features should be hidden to the user.
- open dev tool, set preferred API version to 14, open the app again, cells features should now be visible.

### Checklist

---


WireMessaging/Sources/WireMessagingData/WireCells/NodesAPI/AWSClient.swift
```diff
@@ -258,11 +255,22 @@ final class AWSClient: Sendable {
...
 
     func getPreSignedUrl(objectKey: String) async throws -> String {
+        let s3 = try await makeS3Client()
         let expiration = TimeInterval(Constants.preSignedUrlExpiryInHours * 60 * 60)
         let input = GetObjectInput(bucket: Constants.bucket, key: objectKey)
         let signed = try await s3.presignedURLForGetObject(input: input, expiration: expiration)
         return signed.absoluteString
     }
+
+    private func makeS3Client() async throws -> any S3ClientProtocol {
```
samwyndham: suggestion / issue: I'm not sure it's safe to have multiple S3 clients. Luckily I think the AWS SDK already provides a way to lazily provide the endpoint. See that the config can optionally take an `EndpointResolver` instead of an endpoint. We are doing something similar to lazily provide the access token using `CredentialIdentityResolver`. If you haven't already done so, please give that a try instead of creating a new S3 client each time

wire-ios-data-model/Source/Model/FeatureConfig/Cells/Feature.CellsInternal.swift
```diff
@@ -0,0 +1,58 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public extension Feature {
+
+    struct CellsInternal: Codable {
```
samwyndham, line 23: question: What does internal mean here ?

wire-ios-data-model/Source/Model/FeatureConfig/Cells/Feature.CellsInternal.swift
```diff
@@ -0,0 +1,58 @@
...
+            config: Config = .init()
+        ) {
+            self.status = status
+            self.config = config
+        }
+
+        public struct Config: Codable {
+            public let backend: Backend
+
+            public init(backend: Backend = .init(url: nil)) {
```
samwyndham: Question: Why is the default needed here?

samwyndham (changes requested):
Nice work. I've added a comment for a small change to do in the AWS client. Feel free to discuss.

WireMessaging/Sources/WireMessagingData/WireCells/NodesAPI/AWSClient.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/4052/commits/d99e2faa4d4204dc467132cb447350834b7628a4, thanks for suggestion, lazily init very handy here!

wire-ios-data-model/Source/Model/FeatureConfig/Cells/Feature.CellsInternal.swift
jullianm: it's the BE naming. I suppose it means this is not user facing config

wire-ios-data-model/Source/Model/FeatureConfig/Cells/Feature.CellsInternal.swift
jullianm: not needed, before we were actually creating default config but this code is not relevant anymore.
done in https://github.com/wireapp/wire-ios/pull/4052/commits/429590639ea9dffdaa3f9adb47fa5d2d99a3a621

samwyndham (approved):
Nice work!

KaterinaWire (approved):

PR #4053, opened 2025-12-30
fix: no whitespace allowed in file renaming - WPB-21698

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21698" title="WPB-21698" target="_blank">BugWPB-21698</a>  [iOS] Whitespaces are not being trimmed while using in file name and able to save a filename with just whitespace as well
  </td></table>
  <br />
 

### Issue

This PR adds a fix for File renaming where the user could enter and save a file with whitespaces.

### Testing

- when renaming a file in a cells conv and typing any whitespaces in the input field, save button should be disabled

### Checklist

---


samwyndham (commented):
I'm not sure this is the best approach. I was thinking we could allow white space at the beginning or end of the file name, but just trim it before sending. What do you think?

jullianm:
> I'm not sure this is the best approach. I was thinking we could allow white space at the beginning or end of the file name, but just trim it before sending. What do you think?

@samwyndham  yes you're right. done in https://github.com/wireapp/wire-ios/pull/4053/commits/223dc963cb1ec693e01115d0b611e8d2b2826aef thanks

OlivellaO (approved):

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FileRenameViewModel.swift
```diff
@@ -116,10 +116,14 @@ final class FileRenameViewModel: ObservableObject {
             let nodeID = model.nodeID
             let nodeFilePath = model.filepath
 
+            let newFilename = filenameInput.trimmingCharacters(
```
WilhelmOks: suggestion: 
Extract the trimming into a function to not repeat the details of the transformation (the trimming) in different places.
This will reduce the risk of bugs when the transform requirements will change and only one place will be updated.
The new function could be named something like "transformFilename" or "cleanFilename".

I see the two occurrences are in different files, so I suggest to apply the transform (the trimming) before passing it into the validator. This also seems to be more semantically correct, because the validator is only supposed to check the validity, not to apply the transformation (before checking).

Maybe a new var makes sense:

```
var cleanFilename: String {
    filenameInput.trimmingCharacters(in: .whitespacesAndNewlines)
}
```

WilhelmOks (approved):
Approved with a suggestion.

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/FileRenameViewModel.swift
jullianm: done in https://github.com/wireapp/wire-ios/pull/4053/commits/f1f662971e3ed5e4ba6cb7fe8ea27c2359242965 thanks

PR #4057, opened 2026-01-02
fix: app re-authenticates with email/password instead of SSO - WPB-22220

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22220" title="WPB-22220" target="_blank">BugWPB-22220</a>  [iOS] [Core] - Multibackend - When session gets terminated - App asks for password in SSO scenario
  </td></table>
  <br />
 

### Issue

When their session expires and they need to re-authenticate, users that have an email address registered and use SSO are presented with the email / password authentication flow instead of the SSO authentication flow.

### Cause

`LoginCredentials` has a property `hasPassword` that is used to determine whether to use SSO or email & password authentication. However, this property is set to `email != nil`. But some SSO users may have an email address, so it isn't accurate to assume that having an email necessarily means the user relies on email + password authentication.

### Solution

Remove the `hasPassword` property and rely on the existing `usesCompanyLogin` (translates to `usesSSO`) to check whether the user should authenticate via SSO

### Testing

- Have an SSO user with an email saved 
- Log in
- Expire cookie
- Re-authenticate

The app should redirect to the SSO authentication flow

---

### Checklist


samwyndham (approved):
Looks good. I was trying to understand what was the purpose of even having this  hasPassword field. As far as I can tell, it seems safe to remove it.

wire-ios/Wire-iOS/Sources/Authentication/AuthenticationInterfaceBuilder.swift
```diff
@@ -25,6 +25,7 @@ import WireCommonComponents
 import WireCountly
 import WireDataModel
 import WireFoundation
+import WireLogging
```
KaterinaWire: **question:** Did you want to add some logs?

KaterinaWire (approved):

wire-ios/Wire-iOS/Sources/Authentication/AuthenticationInterfaceBuilder.swift
David-Henner:
```suggestion
```

David-Henner: it was a leftover from debugging, removed.

PR #4059, opened 2026-01-05
fix: recycle bin item menu - WPB-22298

WilhelmOks:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22298" title="WPB-22298" target="_blank">BugWPB-22298</a>  [iOS] - Recycle Bin Context menu and Expected tap Behavior
  </td></table>
  <br />
 

### Issue

File and Folder items in the recycle bin should not have the "open" and "download" menu items.
The only menu items should be "restore" and "delete".

In this PR I have also refactored the code of the file items and their menus:
* using `ItemAction` enum to build the menu and perform/delegate menu actions
* moved some swift files into thematically named folders for better overview

### Testing

* delete files or folders to the recycle bin
* go to the recycle bin
* open the menu of a file or folder item
* there should only be the menu options "restore" and "delete"

The code for which menu items appear in the menu has changed so it's also a good idea to test different views like the All-Conversations-View, to check if the menu items are correct.


WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/Item/FilesItemViewModel.swift
```diff
@@ -260,6 +253,47 @@ final class FilesItemViewModel: ObservableObject {
     }
 }
 
+extension Collection<FilesItemViewModel.ItemAction> {
+    static func menuActions(
```
samwyndham: suggestion: Have this method on `FilesItemViewModel.ItemAction` instead to avoid (minimally) polluting Collections API.

samwyndham (approved):
Nice work!

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/Item/FilesItemViewModel.swift
WilhelmOks: I don't really see it as API pollution because this function will only be available on Collections with ItemActions Elements, not on every collection.

WilhelmOks: But I have another improvement in mind, which would also address your suggestion:
instead of passing the list of menu actions into FilesItemView, I should pass the same params of this function of the Collection extension instead and then the view or the viewmodel will decide how to build this collection.
This will also get rid of the awkward exceptional "isEditable" check in the View.

WilhelmOks: done

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/Item/FilesItemView.swift
```diff
@@ -244,44 +152,103 @@ struct FilesViewItemView: View {
...
-        Task { await viewModel.open() }
-    }
+    @ViewBuilder
+    private func menuContent() -> some View {
+        menuItem(.open) { item in
+            Button {
+                viewModel.performMenuAction(item)
+            } label: {
+                Label(Strings.Files.Item.Menu.open, systemImage: "arrow.up.forward.square")
+            }
```
OlivellaO, line 162: Can we do something so we are not repeating for each action the "menuItem(.actionType) { Button(performAction, Label) }"

It just feel a lot of repeated code.

OlivellaO (approved):

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Files/Item/FilesItemView.swift
WilhelmOks: Yeah I also thought about that but I think this kind of repetition is fine. Especially since there is one exception for the Download menu item.

jullianm (approved):
Looks good

PR #4060, opened 2026-01-05
fix: conversation sorted after backup import - WPB-21925

JaCaLlaSchwarz:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21925" title="WPB-21925" target="_blank">BugWPB-21925</a>  [iOS] Conversation list isn't ordered based on latest messages after backup import
  </td></table>
  <br />
 

During restoring backup process, conversation timestamp is set with its latest message timestamp.

### Issue

When user is logged in empty conversations are pulled from server and its .lastModifiedDate is set to login time, so all conversations are at the same timestamp. On recovering backup messages, no order can be set on adding messages to conversations because all message timestamps are earlier to login timestamp, so is not possible set an order in the conversations.

SOLUTION: On pulling conversations, timestamp is set to minimum (instead of current), and when messages are added to conversations conversation timestamp is set when message time is greater than conversation time, in that way most recent conversations are presented at the beginning of the list.

### Testing

Pre: Select Staging environment and Login in with provided credentials
1. Write any message in any conversation, BUT NOT the first.
2. Get back to conversations lists, now the conversation where we wrote the message is at top.
3.  Do a backup (Adjustments/Account/Copy or restore/do not encrypt/Save in files)
4. Logout
5. Select staging environment.
6. Login with provided credentials, OBSERVE that conversation order is not kept because conversation timestamp was set to login timestamp.
7. Restore step 3 backup ((Adjustments/Account/Copy or restore/do not encrypt/Save in files)
8. Get back at conversation list and check that top conversation is same as step 2.

Video:
https://github.com/user-attachments/assets/1ead3aef-cc2d-4cf2-9793-46c5bcdc5f25

## Unit test

* On PullAllConversationsSyncTests.testPull(), is assert(ed) that conversation timestamp is set to minimum, for enabling update conversation timestamp when messages timestamp are closer to now.
image

### Checklist

---


David-Henner (approved):

WireDomain/Sources/WireDomain/Synchronization/PullAllConversationsSync.swift
```diff
@@ -97,7 +97,7 @@ public final class PullAllConversationsSync: PullAllConversationsSyncProtocol {
         for conversation in conversations.found {
             await store.storeConversation(
                 conversation.toDomainModel(),
-                timestamp: .now,
+                timestamp: Date.distantPast,
```
netbe: question: this modifies the lastServerTimeStamp, because it kind of change the meaning of the field, if in the future we rely on this, is it really needed ?

netbe (commented):

WireDomain/Sources/WireDomain/Synchronization/PullAllConversationsSync.swift
JaCaLlaSchwarz: When user re-logins it fetches his/her empty-conversation from backend. And app what it did was to set conversation time to login-timestamp, so it was imposible to sort conversations, because on importing data from backup conversation messages will be more recent.
By moving conversation retrieval on login, too Date.distantPast, in importing conversations from backup is possible setup conversation timestamp, because they are greater than Date.distantPast.

netbe: I don't fully get it why it's needed, let's sync 1:1. My concern is that the PullAllConversationsSync is not used only for initial sync on login but on occasional sync "resources sync" and I wonder if then the sorting will be broken too

netbe: ok after discussion let's go with this

netbe (approved):
LGTM;)

WireDomain/Sources/WireDomain/Synchronization/PullAllConversationsSync.swift
David-Henner:
```suggestion
                timestamp: conversation.lastEventTime ?? Date.distantPast,
```

David-Henner: To be in line with Android, we should do this 👆 

@JaCaLlaSchwarz you'd need to update the tests

netbe (changes requested):
should retarget 4.14

WireDomain/Sources/WireDomain/Synchronization/PullAllConversationsSync.swift
JaCaLlaSchwarz: Sure,

PR #4062, opened 2026-01-05
fix:  for CI failures where setUsernamePage doesn't load and flaky AccountSettingPage identifier removed - WPB-22561

findms:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22550" title="WPB-22550" target="_blank">TaskWPB-22550</a>  [iOS/QA] Test- Share Extension - Share Image to 1:1 via share extension
  </td></table>
  <br />
 

<hr />

<h2>Summary</h2>
<p><em>Briefly describe about this PR changes.</em></p>
2 changes:

- Occasionally on CI, setUsernamePage doesn't load and test fails, so removed the usage of this page as not needed on U so setting username using API call which improved the performance as well by reducing the test run time.
- Also less frequent but happening where AccountSettingPage doesn't load as tapping on _Accountcell_ identifier remains on same page and page doesn't change. So removed the identifier using label for now.Might add identifier later.

<hr />

<h2>Testing</h2>
<p><em>Describe how to verify the changes locally. Attach screenshots of local execution if relevant.</em></p>

image

**HTML Report**
[index.html](https://github.com/user-attachments/files/24432971/index.html)

<h2>Additional Information</h2>
<p><em>any more info, add here.</em></p>

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptor.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptor.swift
@@ -233,7 +233,6 @@ final class SettingsGroupCellDescriptor: SettingsInternalGroupCellDescriptorType
         cell.icon = icon
         if let cell = cell as? SettingsTableCell {
             cell.showDisclosureIndicatorAccessoryView()
-            cell.accessibilityIdentifier = settingsTopLevelMenuItem?.accessibilityID
         }
     }
 
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptor.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptor.swift
@@ -233,6 +233,7 @@ final class SettingsGroupCellDescriptor: SettingsInternalGroupCellDescriptorType
         cell.icon = icon
         if let cell = cell as? SettingsTableCell {
             cell.showDisclosureIndicatorAccessoryView()
+            cell.accessibilityIdentifier = settingsTopLevelMenuItem?.accessibilityID
         }
     }
 
--- a/wire-ios/WireUITests/AccountManagementTests.swift
+++ b/wire-ios/WireUITests/AccountManagementTests.swift
@@ -25,28 +25,13 @@ final class AccountManagementTests: WireUITestCase {
 
     @MainActor
     func test_Account_Management_Lock_With_Passcode() async throws {
+
         let passcode = UserGenerator.generateAppPasscode()
 
-        do {
-            let (_, teamOwner) = try await userHelper.registerUserAsTeamOwner()
-            let ownerAccessToken = try await userHelper.fetchAccessToken(
-                email: teamOwner.email,
-                password: teamOwner.password
-            )
-            let teamID = try XCTUnwrap(teamOwner.teamID)
+        let user = try await userHelper.createPersonalUser()
 
-            let (_, userInfo) = try await userHelper.registerUsersAsTeamMember(
-                ownerAccessToken: ownerAccessToken,
-                teamID: teamID
-            )
-            teamMember = userInfo
-        } catch {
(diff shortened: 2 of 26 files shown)


KaterinaWire (approved):

wire-ios/WireUITests/Pages/ConversationsPage.swift
```diff
@@ -62,11 +62,21 @@ class ConversationsPage: PageModel {
...
+            nameButton.tap()
+        } else {
+            // Note: Sometimes predicate doesn't work so fallback to identifier.
+            let profileImage = app.buttons[Locators.ConversationsPage.accountProfileImageView.rawValue].firstMatch
+            XCTAssertTrue(
+                profileImage.waitForExistence(timeout: 2),
+                "Not able to open user profile"
+            )
+            profileImage.tap()
         }
```
netbe: question: why it does not work? should we only rely on the accountProfileImageView then?

wire-ios/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptor.swift
```diff
@@ -233,7 +233,6 @@ final class SettingsGroupCellDescriptor: SettingsInternalGroupCellDescriptorType
         cell.icon = icon
         if let cell = cell as? SettingsTableCell {
             cell.showDisclosureIndicatorAccessoryView()
-            cell.accessibilityIdentifier = settingsTopLevelMenuItem?.accessibilityID
```
netbe, line 236: question: where do we set the accessibilityIdentifier now?

wire-ios/WireUITests/AccountManagementTests.swift
```diff
@@ -25,28 +25,13 @@ final class AccountManagementTests: WireUITestCase {
...
-            )
-            let teamID = try XCTUnwrap(teamOwner.teamID)
-
-            let (_, userInfo) = try await userHelper.registerUsersAsTeamMember(
-                ownerAccessToken: ownerAccessToken,
-                teamID: teamID
-            )
-            teamMember = userInfo
-        } catch {
-            throw XCTSkip("error in setup of test: \(error)")
```
netbe, line 44: question: the test was skipped before we failed to have the requirements for the test (aka have a team owner or a member, was it intentional to remove this?

netbe (commented):
left a couple questions before approving

netbe:
@findms maybe would be good to have one test where setUsername is needed too, wdyt?

wire-ios/WireUITests/Pages/ConversationsPage.swift
findms: This is something which I have observed locally while running..so far haven't seen on CI..where opening UserProfile by passing name predicate fails to open..and for the reason why we rely fully on accountProfileImageView identifier  because when there is 1 account, it works fine with but when there is 2 account for multibackendtest or personalusertestflow this identifier doesn't work and not able to open user profile view..so we have to go back to predicate by passing name. I put this else condition to avoid the race condition if one doesn't locate try another.

wire-ios/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptor.swift
findms: Earlier we were extending this settingsTopLevelMenuItem and setting accessbilityID and has been working fine so far..but recently checked on past 1 week this accessbilityID gets located but on performing tap() it doesn't navigate to next page, tried adding some wait but this is completely random. So to keep it simple reverted to use Label and maybe look for adding identifier later.

wire-ios/WireUITests/AccountManagementTests.swift
findms: Yes it is intended and should not fail the way frequently it was failing after changing the flow now so removed. Can put it back as safe guard, wdyt?

findms:
> @findms maybe would be good to have one test where setUsername is needed too, wdyt?

Currently 2 tests are there where username is being set.
1. registering personal user
2. Inviting team member to team

netbe (approved):
LGTM;)

PR #4086, opened 2026-01-08
fix: cancelling uploads - WPB-18574

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18574" title="WPB-18574" target="_blank">TaskWPB-18574</a>  [iOS] Find a way to cancel uploads via AWS SDK
  </td></table>
  <br />
 

### Issue

Currently when using wire drive (eg wire cells) the user can cancel uploading files. However this only cancels it from a UX perspective. The file is still actually uploaded to the server. This is because the server is S3 compatible and we use the AWS SDK to upload files. However, this SDK provides no way to cancel uploads 🙀. I opened a ticket for the SDK to improve this but I'm not sure it will be done anytime soon. https://github.com/awslabs/aws-sdk-swift/issues/1977. Amazon has just released a more suitable SDK for our purposes (https://github.com/aws/aws-sdk-swift-s3-transfer-manager) but It is currently in developer preview, so not suitable for production.

Therefore, this PR introduces a hack to force cancellation of uploads. It is very tied to the implementation of the AWS SDK.

This PR adds a cancel method to `ObservableStream` (another hack to allow observing upload progress) that just makes the stream return nil data. This causes the upload to error. Elsewhere we catch this error and return a cancellation error.

### Testing

1. navigate to a wire cells enabled conversation.
2. Begin uploading a large file.
3. Tap the cross icon on the upload preview to cancel the upload.
4. The upload preview disappears.
5. While connected to a debugger, check out the logs display a cancellation message like: `Upload cancelled for node: <path>`

---

### Checklist

### DIFF
--- a/WireMessaging/Sources/WireMessagingData/WireCells/NodesAPI/ObservableStream.swift
+++ b/WireMessaging/Sources/WireMessagingData/WireCells/NodesAPI/ObservableStream.swift
@@ -17,15 +17,16 @@
 //
 
 import Foundation
+import os
 import Smithy
 
 /// Wraps a `Smithy.Stream` allowing its progress to be observed. This is a workaround to enable reading upload progress
 /// using the AWS SDK.
-
 final class ObservableStream: Smithy.Stream, Sendable {
 
     private let wrappedStream: any Smithy.Stream
     private let readContinuation: AsyncStream<Int>.Continuation
+    private let cancellationState = OSAllocatedUnfairLock(uncheckedState: false)
 
     let readProgress: AsyncStream<Int>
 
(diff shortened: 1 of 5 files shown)


WireMessaging/Sources/WireMessagingData/WireCells/NodesAPI/ObservableStream.swift
```diff
@@ -17,15 +17,16 @@
...
 import Smithy
 
 /// Wraps a `Smithy.Stream` allowing its progress to be observed. This is a workaround to enable reading upload progress
 /// using the AWS SDK.
-
 final class ObservableStream: Smithy.Stream, Sendable {
 
     private let wrappedStream: any Smithy.Stream
     private let readContinuation: AsyncStream<Int>.Continuation
+    private let cancellationState = OSAllocatedUnfairLock(uncheckedState: false)
```
samwyndham, line 29: Using a OSAllocatedUnfairLock here to keep free sendable conformance

WireMessaging/Sources/WireMessagingData/WireCells/NodesAPI/ObservableStream.swift
```diff
@@ -99,4 +110,21 @@ final class ObservableStream: Smithy.Stream, Sendable {
...
+        cancellationState.withLock { $0 }
+    }
+
+    /// Cancels the stream, causing further read operations to return nil or throw a CancellationError.
+    ///
+    /// This is a hack to work around the lack of cancellation support in the AWS SDK. It relies on implementation
+    /// details of the SDK and its dependencies but it does currently work - when called while a file is being uploaded,
+    /// the upload stops and an error will soon after be thrown.
+    func cancel() {
+        cancellationState.withLock {
```
jullianm, line 126: question: Just so I understand, why do we need to rely on a lock mechanism here ? 
My thought would be to replace this low-level API by an actor, so when the `cancel()` method here is called, we update the cancellation token flag to true and it will still throw the `CancellationError` and stops the stream like you already did.

`cancellationToken.isCancelled = true`

```
actor CancellationToken {
    var isCancelled = false
}
```

jullianm (commented):
Fix looks good but I have a question specifically about the usage of the lock mechanism

WireMessaging/Sources/WireMessagingData/WireCells/NodesAPI/ObservableStream.swift
samwyndham: Hey @jullianm. The reason that we can't use an actor is that some of the methods that need to inspect the cancelled state are not async so we can't use an await call which we would need to access the actor. We also can't make observable stream and actor for the same reason

jullianm (approved):
LGTM

WilhelmOks (approved):

PR #4093, opened 2026-01-12
fix: ipad image preview automatically closing - WPB-22073

WilhelmOks:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22073" title="WPB-22073" target="_blank">BugWPB-22073</a>  [iPad Mini 5/ iPadOS 18.6.2] Files: Cannot view images the doc viewer collapses after a short moment
  </td></table>
  <br />
 

### Issue

On an iPad, the image preview was closing itself automatically. The reason was that the QuickLookPreview, which was used to show the image preview, doesn't support being presented from within a Split View Controller.
I'm not sure if this is a bug or a deliberate limitation, but it's only happening with the QuickLookPreview, not with any other custom view presented in a sheet or in fullscreen.

To solve this, I implemented a workaround which, instead of using SwiftUI's .quickLookPreview() modifier, finds the topmost presented View Controller via UIKit and then presents the preview on top of it, with UIKit as well.
This works on iPhone as well, so there is no need for a iPhone/iPad check.

### Testing

Open the app in an iPad.
Tap on an image, which was sent in a conversation.
The image should open in a fullscreen preview and remain open. It should not automatically close itself.


jullianm (approved):
Looks good

samwyndham (approved):
Nice work

PR #4095, opened 2026-01-12
fix: Inform AVS about background state - WPB-22217

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22217" title="WPB-22217" target="_blank">BugWPB-22217</a>  [Calling] [iOS] Call notification and missed call indication from many people for Team meeting call
  </td></table>
  <br />
 

### Issue

The app can receive multiple "Missed call from ..." messages even for calls that are still ongoing.
When the app is not receiving `CONFCHECK` messages on an ongoing basis, for instance due to app not having internet connection, then AVS may timeout that call when going back to online and connection is being reestablished and result in executing `closedCallHandler` for that call even if the call is still ongoing and that client is not taking part in that call.

### Solution

We need to inform the AVS about app being inactive ("in the background")  or not having the network connection, so that it can properly determine the current state and decide whether it should timeout because of missing `CONFCHECK` messages or if it's an expected behavior because the app is in the background + not in the call/inactive/closed web-socket. 
`"inactive" means that there is no open webSocket. `

### Testing

The issue with unexpected system messages can be reproduced by following these steps:
- **user A** (this is our main test device) moves the app in the background
- user B starts a group call
- user C joined the call
- user A receives a CallKit notification about the incoming call and close it (don't join the call)
- user A turns on the Airplane mode
- the call needs to be still ongoing while the app is in the background without connection (user B can send a message during this time)
- after a few minutes (3-4 should be enough) user A turns Internet back on and opens the app again
- the call is still ongoing and user A should **NOT** receive "Missed call from UserName" message for that call

### Checklist

---


wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1252,6 +1272,7 @@ extension ZMUserSession: SyncAgentDelegate {
         }
 
         performPostQuickSyncE2EIActions()
+        notifyAVSDidProcessEvents()
```
netbe: suggestion: move to top of method. some actions here could take sometimes (resolveOneOnOneConversationsIfNeeded)

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -626,6 +634,7 @@ public final class ZMUserSession: NSObject {
             )
         )
         self.clientSessionComponent = clientSessionComponent
+        observeSyncStateForAVS()
```
netbe: suguestion: pass the subject directly instead of guard the optional clientSessionComponent
```suggestion
        observeSyncStateForAVS(syncStateSubject)
```

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+AVS.swift
```diff
@@ -0,0 +1,51 @@
...
+
+import Foundation
+
+extension ZMUserSession {
+
+    func notifyAVSWillProcessEvents() {
+        managedObjectContext.perform { [weak self] in
+            guard let self else { return }
+            callCenter?.avsWrapper.notify(isProcessingNotifications: true)
+        }
```
netbe, line 27: thought: it would be good at some point to get rid of accessing the callCenter via nsmanagedobjectContext. AVSWrapper could be an actor instead

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1119,6 +1127,17 @@ extension ZMUserSession: ZMNetworkStateDelegate {
...
         }
     }
+
+    private func observeSyncStateForAVS() {
+        guard let clientSessionComponent else { return }
+
+        syncStateCancellable = clientSessionComponent.syncStateSubject
+            .sink { [weak self] syncState in
+                guard let self else { return }
+                isLiveSyncOngoing = syncState == .liveSyncing(.ongoing)
```
netbe: suggestion:
I would suggest to directly call `notifyAVSOfLiveSyncState(syncState == .liveSyncing(.ongoing))` to remove another state that needs to be maintained within ZMUserSession

netbe (approved):
LGTM;) left a could of suggestions

johnxnguyen (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
KaterinaWire: done

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
KaterinaWire: fixed

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
KaterinaWire: removed
cc @johnxnguyen @netbe

PR #4104, opened 2026-01-13
fix: Wire drive preview unavailable on first sent - WPB-21901

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21901" title="WPB-21901" target="_blank">BugWPB-21901</a>  [iOS] When a file is first sent, its preview is unavailable.
  </td></table>
  <br />
 

### Issue

This PR fixes an issue where the preview is marked as unavailable when sent for the first time. This is because previews may not be immediately available server side.
We make sure to poll the server more frequently until they are.

### Testing

Attach an image on a wire cells enabled conv and send it
Before: Preview in the conv showed as unavailable for 30 seconds or when going back and entering the conv again
Now: Preview is right away displayed when ready

### Checklist

---


OlivellaO (approved):

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsAttachmentsPreviewItemViewModel.swift
```diff
@@ -156,10 +156,16 @@ final class WireCellsAttachmentsPreviewItemViewModel: ObservableObject {
     func startPolling() {
         pollingTask?.cancel()
         pollingTask = Task { [weak self] in
+            var attemptToGetInitialPreviews = 0
```
jullianm: to avoid saturating the server if we fail to get the previews within the 10sec

samwyndham: @jullianm The node response JSON includes a bool value ("processing") that we can inspect to know that we are processing the previews. I think we should use this somehow to avoid polling unnecessarily... For example some files have no previews and this polling implementation may be aggressive in those cases. Perhaps we can do this implementation for now but add a ticket for a gentler implementation. What do you think?

jullianm: makes sense, created the ticket here https://wearezeta.atlassian.net/browse/WPB-22763

jullianm: fixed in this commit https://github.com/wireapp/wire-ios/pull/4104/commits/ed072f9accca65ad1becc95c80346f1a111695c0, also re-enabled the share link for folders. @samwyndham

WireMessaging/Sources/WireMessagingData/WireCells/Model/WireCellsNodeNetworkModel.swift
```diff
@@ -174,26 +174,31 @@ package extension RestNode {
...
-            return nil
+        var url: URL?
+
+        if let urlString = value.preSignedGET?.url, let contentType = value.contentType {
+            guard let previewURL = URL(string: urlString),
+                  let type = UTType(mimeType: contentType),
+                  type.conforms(to: .image) else {
+                return nil
+            }
+            url = previewURL
```
samwyndham, line 196: isssue (I think): Isn't the logic wrong here? Don't we want to allow for a nil URL rather then returning nil? Also perhaps it would make sense for `PreviewDTO` to be an enum:

```swiift
enum PreviewDTO {

  case processing
  case ready(url: URL, dimension: Int?)

}
```

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsAttachmentsPreviewItemViewModel.swift
```diff
@@ -155,10 +163,14 @@ final class WireCellsAttachmentsPreviewItemViewModel: ObservableObject {
...
-                await self?.refresh()
+            guard let self else { return }
 
-                try? await Task.sleep(for: .seconds(30))
+            while !Task.isCancelled {
+                await refresh()
+                // Image previews may not be immediately available after upload.
+                // Poll the server at a higher frequency if the preview is still being processed.
+                let refreshInitialPreviews = (imagePreviewURL == nil && isProcessing)
+                try? await Task.sleep(for: .seconds(refreshInitialPreviews ? 1 : 30))
```
samwyndham: issue: I think we still want to have a cut off like you had in your previous commit. For example imagine you successfully downloaded the node and isProcessing is true and there is currently no internet. Polling would be stuck happening every second. 

In general I think most previews should be ready in 1 second. How about having a polling time that doubles each time to a maximum of 32 seconds. So, 1, 2, 4, 8, 16, 32 then stays at 32?

samwyndham (changes requested):
Nice work. I think there could be a couple of small changes. What do you think?

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsAttachmentsPreviewItemViewModel.swift
jullianm: If we successfully downloaded the node and lost the connection, the `isProcessing` flag will be false hence falling back to the default sleep value of 30 seconds. The API response only returns the processing field with a value of true, the other fields are omitted. Conversely, if not processing, `processing` field is omitted. So, I think we're safe here by ensuring we rely on the processing flag to update the sleep duration.

```
  "Previews": [
    {
      "Processing": true,
      "ContentType": "image/jpg"
    }
  ],
```

WireMessaging/Sources/WireMessagingData/WireCells/Model/WireCellsNodeNetworkModel.swift
jullianm: this is the same logic as before, adjusted a bit to handle the processing flag. If we do get a value from the server that means we're not processing, so the criteria for the URL should be met. 
The enum makes total sense though I would keep it for a separate (later improvement) PR.

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsAttachmentsPreviewItemViewModel.swift
jullianm: I misread your comment, I thought we were talking about downloading the preview. 
You're right indeed, this is probably an edge case though but let's handle that case just to be on the safe side.

jullianm: fixed in https://github.com/wireapp/wire-ios/pull/4104/commits/5237cfa5f52e670bd469fcf335aa006595dd84b3

WireMessaging/Sources/WireMessagingData/WireCells/Model/WireCellsNodeNetworkModel.swift
samwyndham: Ah got it. I was misreading the code

samwyndham (approved):
Nice work!

PR #4114, opened 2026-01-14
fix: deleted file previews visible in replies - WPB-19434

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19434" title="WPB-19434" target="_blank">TaskWPB-19434</a>  [iOS] Display "File not available" when files are not available
  </td></table>
  <br />
 

### Issue

This is a fix for an issue found during QA of `WPB-19434`. When a cells asset is in the recycle bin, "File not available" should be displayed in any reply messages that contain it. However that is not the case. Instead it is treated like it is not in the recycle bin. This PR fixes that issue with a caveat that the table view cell might first need to be recycled to see the update.

One thing that is required in various places is the ability to access a cached node synchronously from the main thread. I tried various different APIs to achieve this in `WireCellsFetchNodeUseCase` but in the end simplified that use case and introduced `WireCellsFetchCachedNodeUseCaseProtocol`. `WireCellsFetchCachedNodeUseCase` is now being used in UI in all places that `WireCellsNodeCacheProtocol` was previously being used.

As the height of a reply message cell can change dynamically when the app discovers that an asset has been deleted / restored I had to include delegation code to have the `UITableView` re-layout using `begin/endUpdates()`

https://github.com/user-attachments/assets/71368d62-b46f-41c5-a4cd-31e5c200c210

### Testing

1. Navigate to a cells enabled converation
2. Send a message with multiple attachments
3. Reply to that message
4. Navigate to the files list
5. Play around with deleting & restoring files
6. Assert that 1. Only non-deleted files are taken account of in the reply preview and 2. If all files are deleted we display a "File not available" message.

### Checklist

---


WireMessaging/Sources/WireMessagingDomain/WireCells/Protocols/WireCellsNodeCacheProtocol.swift
```diff
@@ -22,9 +22,15 @@ package import Foundation
...
 /// A `nil` value indicates that the node was not found on the server, for example it may have been deleted. This is
 /// different from the value never having been added to the cache.
-package struct WireCellsNodeCacheItem {
+public struct WireCellsNodeCacheItem: Equatable, Sendable {
 
-    package let node: WireCellsNode?
+    public let node: WireCellsNode?
+
+    /// Whether the node is deleted or in the recycle bin.
+    public var isDeleted: Bool {
```
WilhelmOks: maybe a more clear name would be `isDeletedOrRecycled`

WireMessaging/Tests/WireMessagingTests/WireCells/ImplementationTests/UseCases/WireCellsFetchNodeUseCaseTests.swift
```diff
@@ -35,54 +35,36 @@ final class WireCellsFetchNodeUseCaseTests {
     }
 
     @Test
-    func invoke_whenNothingCached() async throws {
+    func invoke_whenNodeNoteFound() async throws {
```
WilhelmOks: seems to be a typo and should be `invoke_whenNodeNotFound` instead

WilhelmOks (approved):
Approved with suggestions

wire-ios/Wire-iOS/Sources/UserInterface/Components/MessageReplyAttachmentsView.swift
```diff
@@ -39,12 +40,28 @@ final class MessageReplyAttachmentsView: UIView {
...
         super.init(frame: .zero)
 
-        setup(attachments: attachments)
+        let cachedVisibleAttachments = viewModel.cachedVisibleAttachments(attachments: attachments)
+        setup(attachments: cachedVisibleAttachments)
+
+        Task { @MainActor in
+            do {
+                let latestVisibleAttachments = try await viewModel.latestVisibleAttachments(attachments: attachments)
+                if latestVisibleAttachments != cachedVisibleAttachments {
```
jullianm, line 55: question: I wonder if the cached visible attachments logic is needed. Can't we compare the reply current attachments to the `latestVisibleAttachments` instead? I figured this would be sufficient to know whether we need to update the reply attachments or not.

jullianm (commented):
Looks good, left one question about the cached nodes logic introduced here. Not sure why it is needed.

wire-ios/Wire-iOS/Sources/UserInterface/Components/MessageReplyAttachmentsView.swift
samwyndham: @jullianm Some context about how this is intended to work. I found implementing this, we need to have have initial data at init time so we do at least need to fetch `cachedVisibleAttachments` and as we have that I think it makes sent to use it for comparison before updating the data. 

However I did do this PR attempting to make minimal changes and there are certainly possibly better solutions. Of course it is hard here because we are interacting with the legacy system.

jullianm (approved):

WireMessaging/Tests/WireMessagingTests/WireCells/ImplementationTests/UseCases/WireCellsFetchNodeUseCaseTests.swift
samwyndham: Fixed in https://github.com/wireapp/wire-ios/pull/4114/commits/582f34ff232ab7ad77c06b5cf635c89df4a09e0d

WireMessaging/Sources/WireMessagingDomain/WireCells/Protocols/WireCellsNodeCacheProtocol.swift
samwyndham: Implemented in https://github.com/wireapp/wire-ios/pull/4114/commits/50468f60401951db0b7b5e1bed55c61d324581ca

PR #4120, opened 2026-01-15
fix: generating thumbnail failure UX - WPB-21900

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21900" title="WPB-21900" target="_blank">BugWPB-21900</a>  [iOS] Sometimes we see continuous spinner when attaching an image
  </td></table>
  <br />
 

### Issue

When uploading an asset generating a thumbnail (locally) might fail. In that situation the user sees a progress indicator (spinner) constantly even if the file has successfully been uploaded.

There are two issues I identified:
1. Generation of some image thumbnails failed because they were incorrectly marked as jpeg when in fact they were png (and probably other types also).
2. If thumbnail generation fails the UX is really bad because we just show an spinner constantly.

To fix part 1 I changed the code which was incorrectly marking images as jpeg. To fix part 2 I just removed the spinner. This should only show while generation occurs and as generation takes almost zero time (it's local) we should likely never see the spinner anyway.

This PR also fixes the size of small image previews in the conversation, making them 120 points instead of 74 points. 

### Testing

1. Navigate to a cells conversation
2. Attach images by various means - taking a photo, selecting from the in app gallery etc. Try to send a png - for example a screenshot on my device. This is what currently fails.
3. A thumbnail should be correctly generated.

---

### Checklist


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/CameraKeyboard/CameraKeyboardViewController.swift
samwyndham, line 1: This file has the main bug fix

WireMessaging/Sources/WireMessagingUI/WireCells/Components/AttachmentsPreviewView/WireCellsAttachmentsPreviewItemView.swift
samwyndham, line 1: This fixes the size of small image previews in the conversation

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Previews/WireCellsImageAttachmentPreview.swift
samwyndham, line 1: There was no reason to show a progress view because image generation is super quick

jullianm (approved):
Nice work

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Previews/WireCellsImageAttachmentPreview.swift
OlivellaO: Are we sure about removing this ProgressView()? If for some reason image generation is not super quick we will show nothing to the user, isn't it?
I think is better to keep it if doesn't hurt.

OlivellaO (approved):

WireMessaging/Sources/WireMessagingUI/WireCells/Components/Previews/WireCellsImageAttachmentPreview.swift
samwyndham: Yes I think so because image generation is done locally in milliseconds and keeping it adds quite a lot of complexity because we must derive 3 states - image is being generated, generation failed, generation succeeded. I don't think it is worth the complexity because I don't think that it will be shown in practice.

PR #4124, opened 2026-01-15
fix: conversation sorted after backup import - WPB-21925

JaCaLlaSchwarz:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21925" title="WPB-21925" target="_blank">BugWPB-21925</a>  [iOS] Conversation list isn't ordered based on latest messages after backup import
  </td></table>
  <br />
 

During restoring backup process, conversation timestamp is set with its latest message timestamp. (Video attached)

### Issue

When user is logged in empty conversations are pulled from server and its .lastModifiedDate is set to login time, so all conversations are at the same timestamp. On recovering backup messages, no order can be set on adding messages to conversations because all message timestamps are earlier to login timestamp, so is not possible set an order in the conversations.

SOLUTION: On pulling conversations, timestamp is set to minimum (instead of current), and when messages are added to conversations conversation timestamp is set when message time is greater than conversation time, in that way most recent conversations are presented at the beginning of the list.

### Testing

Pre: Select Staging environment and Login in with provided credentials
1. Write any message in any conversation, BUT NOT the first.
2. Get back to conversations lists, now the conversation where we wrote the message is at top.
3.  Do a backup (Adjustments/Account/Copy or restore/do not encrypt/Save in files)
4. Logout
5. Select staging environment.
6. Login with provided credentials, OBSERVE that conversation order is not kept because conversation timestamp was set to login timestamp.
7. Restore step 3 backup ((Adjustments/Account/Copy or restore/do not encrypt/Save in files)
8. Get back at conversation list and check that top conversation is same as step 2.

Video:
https://github.com/user-attachments/assets/1ead3aef-cc2d-4cf2-9793-46c5bcdc5f25

## Unit test

* On PullAllConversationsSyncTests.testPull(), is assert(ed) that conversation timestamp is set to minimum, for enabling update conversation timestamp when messages timestamp are closer to now.
image

### Checklist

---

### DIFF
--- a/WireDomain/Tests/WireDomainTests/Synchronization/PullAllConversationsSyncTests.swift
+++ b/WireDomain/Tests/WireDomainTests/Synchronization/PullAllConversationsSyncTests.swift
@@ -65,7 +65,7 @@ final class PullAllConversationsSyncTests: XCTestCase {
         })
 
         api.getConversationsFor_MockValue = .init(
-            found: [Scaffolding.remoteConversation1],
+            found: [Scaffolding.remoteConversation1, Scaffolding.remoteConversation4],
             notFound: [Scaffolding.conversationID2],
             failed: [Scaffolding.conversationID3]
         )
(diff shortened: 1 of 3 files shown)


David-Henner (approved):

WireDomain/Tests/WireDomainTests/Synchronization/PullAllConversationsSyncTests.swift
```diff
@@ -88,6 +88,7 @@ final class PullAllConversationsSyncTests: XCTestCase {
         XCTAssertEqual(storeFoundInvocations[0].conversation, Scaffolding.localConversation1)
         XCTAssertEqual(storeFoundInvocations[0].isFederationEnabled, Scaffolding.isFederationEnabled)
         XCTAssertEqual(storeFoundInvocations[0].isMLSEnabled, Scaffolding.isMLSEnabled)
+        XCTAssertEqual(storeFoundInvocations[0].timestamp, Date.distantPast)
```
netbe, line 91: todo: can you add a test where the remoteConversation has a lastEventTime?

netbe (approved):
LGTM:)

PR #4129, opened 2026-01-16
fix: conversations marked as read only when failing to establish mls group - WPB-22774

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22774" title="WPB-22774" target="_blank">BugWPB-22774</a>   [iOS] Messages missing in chat & Input field disappeared
  </td></table>
  <br />
 

### Issue

When resolving a 1:1 conversation, for a mls conversation we try to establish the group and then migrate potential messages from previous proteus conversations if any. In case the group fails to establish, we set the conversation as readonly.

As the mls conversation is already migrated, next time we don't resolve this conversation again and can't remove the readonly flag once the conversation successfully establish.

### Testing

1. user A has a 1:1 with a user B
2. user B has no more key packages
3. trigger a mls reset
4. user A can't establish the new group

Result:
~~the conversation is set to readonly~~
the conversation is still accessible

---

### Checklist

---

### DIFF
--- a/WireDomain/Sources/WireDomain/Synchronization/OneOnOneResolver.swift
+++ b/WireDomain/Sources/WireDomain/Synchronization/OneOnOneResolver.swift
@@ -86,10 +86,11 @@ public struct OneOnOneResolver: OneOnOneResolverProtocol {
         let selfUser = await userLocalStore.fetchSelfUser()
         let commonProtocol = await getCommonProtocol(between: selfUser, and: user)
 
-        // If there are no common protocols, there can be no communication
-        // yet, so mark it read only FIRST! Otherwise we unblock the conversation
-        // since it can be resolved.
-        // only when conversation messageProtocol's none and iMLSEnabled we'll set conversation to readOnly
+        // If MLS is enabled and there are no common protocols, then
+        // there is no possibility to communicate and the 1-1 will be marked
+        // read only (follow case below). However in all other cases
+        // the conversation should be unmarked so that if they were
+        // previously blocked they are now unblocked.
         if !(mlsProvider.isMLSEnabled && commonProtocol == nil) {
             await setReadOnly(
                 to: false,
--- a/WireDomain/Sources/WireDomain/Synchronization/OneOnOneResolver.swift
+++ b/WireDomain/Sources/WireDomain/Synchronization/OneOnOneResolver.swift
@@ -158,24 +158,26 @@ public struct OneOnOneResolver: OneOnOneResolverProtocol {
         mlsPublicKeys: MLSPublicKeys?,
         user: ZMUser,
         userID: WireDataModel.QualifiedID
-    ) async {
+    ) async throws {
+        // Establish the group if needed.
+        if try await !mlsProvider.service.conversationExists(groupID: mlsGroupID) {
+            let keys = mlsPublicKeys.flatMap {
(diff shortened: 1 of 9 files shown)


wire-ios-data-model/Source/MLS/OneOnOne/LegacyOneOnOneResolver.swift
```diff
@@ -98,9 +98,22 @@ public final class LegacyOneOnOneResolver: OneOnOneResolverInterface {
 
         let messageProtocol = try await protocolSelector.getProtocolForUser(with: userID, in: context)
 
+        // If there are no common protocols, there can be no communication
+        // yet, so mark it read only. Otherwise we unblock the conversation
+        // since it can be resolved.
+        // only when conversation messageProtocol's none and iMLSEnabled we'll set conversation to readOnly
```
johnxnguyen:
```suggestion
        // If MLS is enabled and there are no common protocols, then
        // there is no possibility to communicate and the 1-1 will be marked
        // read only (follow case below). However in all other cases
        // the conversation should be unmarked so that if they were
        // previously blocked they are now unblocked.
```

WireDomain/Sources/WireDomain/Synchronization/OneOnOneResolver.swift
```diff
@@ -85,7 +85,19 @@ public struct OneOnOneResolver: OneOnOneResolverProtocol {
 
         let selfUser = await userLocalStore.fetchSelfUser()
         let commonProtocol = await getCommonProtocol(between: selfUser, and: user)
-
+        
+        // If there are no common protocols, there can be no communication
+        // yet, so mark it read only FIRST! Otherwise we unblock the conversation
+        // since it can be resolved.
+        // only when conversation messageProtocol's none and iMLSEnabled we'll set conversation to readOnly
```
johnxnguyen: This comment is confusing, I suggested another

WireDomain/Sources/WireDomain/Synchronization/OneOnOneResolver.swift
```diff
@@ -86,6 +86,18 @@ public struct OneOnOneResolver: OneOnOneResolverProtocol {
         let selfUser = await userLocalStore.fetchSelfUser()
         let commonProtocol = await getCommonProtocol(between: selfUser, and: user)
 
+        // If there are no common protocols, there can be no communication
+        // yet, so mark it read only FIRST! Otherwise we unblock the conversation
+        // since it can be resolved.
+        // only when conversation messageProtocol's none and iMLSEnabled we'll set conversation to readOnly
```
netbe:
```suggestion
        // If MLS is enabled and there are no common protocols, then
        // there is no possibility to communicate and the 1-1 will be marked
        // read only (follow case below). However in all other cases
        // the conversation should be unmarked so that if they were
        // previously blocked they are now unblocked.
```

WireDomain/Sources/WireDomain/Synchronization/OneOnOneResolver.swift
netbe: fixed

johnxnguyen (approved):

caldrian (approved):

PR #4169, opened 2026-01-21
fix: incorrect "X added you" (to conversation) notifications - WPB-22820

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22820" title="WPB-22820" target="_blank">BugWPB-22820</a>  [iOS] Repeated “Added to Group” push notifications for existing members
  </td></table>
  <br />
 

### Issue

Some users are experiencing push notifications being presented, indicating they've been added to a conversation, when they're already members of that conversation. They reported seeing these notifications more than once for the same conversation.

### Cause

Clients receive and process `member-join` events with their self user being included in the payload as one of the added members, even though they're already members of the conversation. This causes notifications to be shown incorrectly. 

### Solution

The cause for these `member-join` is unclear, but we can prevent them from triggering notifications by verifying whether the user is already a member of the conversation.

### Testing

No clear steps to reproduce. We'll rely on unit tests to verify that notifications aren't created incorrectly.

We can still manual verify that notifications are being shown when needed, by adding a user to a conversation and verifying that their client receives a notification.

---

### Checklist


wire-ios-request-strategy/Sources/Notifications/Push Notifications/Notification Types/Content/ZMLocalNotification+Events.swift
```diff
@@ -394,14 +394,60 @@ private class NewSystemMessageNotificationBuilder: EventNotificationBuilder {
...
+            guard let conversation else {
+                logSkippingNotification(
+                    reason: "conversation doesn't exist",
+                    conversationID: nil
+                )
+                return false
+            }
+
+            // Check if self user is already a participant
+            guard !conversation.localParticipants.contains(selfUser) else {
```
samwyndham: question: Could there be a situation where self user is added to this conversation just before this notification code is triggered?

samwyndham (approved):
Nice work

wire-ios-request-strategy/Sources/Notifications/Push Notifications/Notification Types/Content/ZMLocalNotification+Events.swift
```diff
@@ -394,14 +394,60 @@ private class NewSystemMessageNotificationBuilder: EventNotificationBuilder {
...
+
+        case .participantsAdded where concernsSelfUser == true:
+            // Check if conversation exists
+            guard let conversation else {
+                logSkippingNotification(
+                    reason: "conversation doesn't exist",
+                    conversationID: nil
+                )
+                return false
+            }
```
johnxnguyen: suggestion: I think this could be easier to read we first guard that the self user is concerned:

```swift
guard concernsSelfUser else {
  return false
}

switch contentType {
case .participantsAdded:
  // ...

case .participantsRemoved:
  // ...
}
```

wire-ios-request-strategy/Sources/Notifications/Push Notifications/Notification Types/Content/ZMLocalNotification+Events.swift
```diff
@@ -394,14 +394,60 @@ private class NewSystemMessageNotificationBuilder: EventNotificationBuilder {
 
     override func shouldCreateNotification() -> Bool {
         // we don't want to create notifications when other people join or leave conversation
-        let concernsSelfUser = event.userIDs.contains(ZMUser.selfUser(in: moc).remoteIdentifier)
+        let selfUser = ZMUser.selfUser(in: moc)
```
johnxnguyen: question: is this code still live? Maybe only when the app is open?

wire-ios-request-strategy/Sources/Notifications/Push Notifications/Notification Types/Content/ZMLocalNotification+Events.swift
David-Henner: It's difficult to say with certainty because showing notifications is handled by the NSE while processing the event and applying its effects (adding the user) is performed by the main app. I tested it and it works, i.e: I get the notification when I'm supposed to. But there could perhaps be a case where the order of execution between the main app and the NSE cause the user to be added before the notification is created?

wire-ios-request-strategy/Sources/Notifications/Push Notifications/Notification Types/Content/ZMLocalNotification+Events.swift
David-Henner: Looks like it's unused & the code is already removed on `develop`.  I removed the changes and left only the ones being used by the new sync system

johnxnguyen (approved):

PR #4172, opened 2026-01-21
fix: sharing IPS file changes name and sends "corrupt" data  - WPB-22695

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22695" title="WPB-22695" target="_blank">BugWPB-22695</a>  [iOS] File name replaced when forwarding files from iPhone / .ips files get corrupted
  </td></table>
  <br />
 

### Issue
When sending a file through the share extension, the file name is replaced with a random UUID string. Why exactly is not clear, but it is now expected that the original file name is retained.

This PR uses the original filename if available, otherwise falls back on the UUID + file extension.

Additionally, for some files (such as IPS files shared from iOS settings), sharing through the extension results in a file that seems to contain corrupt data. In reality it is a property list containing a url of the file wanted to be shared. To handle this case, we need to attempt to interpret the share attachment as a property list and try to extract a file url so the contents of that file is what is actually shared.

### Testing
1. Send a file through the share extension, such as from the Files app
2. Verify the file is sent and received and the filename was not changed.

1. Send an IPS file through the extension (from iOS settings > privacy & security > analytics & improvements > Analytics data > share
2. Verify on the receiver side that the shared file is a valid ops file.

---

### Checklist

---


wire-ios/Wire-iOS Share Extension/Sources/Content/UnsentSendable.swift
```diff
@@ -352,12 +340,95 @@ class UnsentFileSendable: UnsentSendableBase, UnsentSendable {
...
+    /// - Returns: A file URL if the data is a property list containing one, nil otherwise
+    private func extractFileURLFromPropertyList(_ data: Data) -> URL? {
+        // Check if this is a binary property list (starts with "bplist")
+        guard
+            data.count > 6,
+            let header = String(data: data.prefix(6), encoding: .utf8),
+            header.hasPrefix("bplist")
+        else {
+            return nil
+        }
```
samwyndham: question: Is this check necessary. Can we just try to open it and treat opening failing as it is not a property list

wire-ios/Wire-iOS Share Extension/Sources/Content/UnsentSendable.swift
```diff
@@ -434,10 +505,13 @@ class UnsentFileSendable: UnsentSendableBase, UnsentSendable {
     }
 
     private func nameForFile(withUTI UTI: String, name: String?) -> String? {
-        if let fileExtension = UTType(UTI)?.preferredFilenameExtension {
-            return "\(UUID().uuidString).\(fileExtension)"
+        if let name {
+            name
+        } else if let fileExtension = UTType(UTI)?.preferredFilenameExtension {
+            "\(UUID().uuidString).\(fileExtension)"
+        } else {
```
netbe, line 503: question: did you check with security if that was a concern?

netbe (approved):
Looks good, glad this is resolved

wire-ios/Wire-iOS Share Extension/Sources/Content/UnsentSendable.swift
```diff
@@ -327,20 +327,8 @@ class UnsentFileSendable: UnsentSendableBase, UnsentSendable {
...
-                self?.prepareForSending(
-                    withUTI: UTIString,
-                    name: name,
-                    data: data,
-                    completion: prepareColsure
-                )
-            } else if let dataURL = data as? URL {
-
+            // First try to interpret the result as a URL
+            if let dataURL = data as? URL {
```
samwyndham, line 331: question: Did you check if this will catch the IPS file case? I would hope that iOS might magically give us a URL if we request it.

samwyndham (approved):

samwyndham:
@johnxnguyen For test steps it might be worth confirming that pass kit files are still working

wire-ios/Wire-iOS Share Extension/Sources/Content/UnsentSendable.swift
johnxnguyen: yes, it doesn't unfortunately

wire-ios/Wire-iOS Share Extension/Sources/Content/UnsentSendable.swift
johnxnguyen: I will try

wire-ios/Wire-iOS Share Extension/Sources/Content/UnsentSendable.swift
johnxnguyen: Nope but will double check. However we were also using the name previous to this change if there wasn't a preferred file extension so I imagine it's not a problem.

johnxnguyen: Also... when sharing in the main app we don't remove the name

wire-ios/Wire-iOS Share Extension/Sources/Content/UnsentSendable.swift
johnxnguyen: removed!

wire-ios/Wire-iOS Share Extension/Sources/Content/UnsentSendable.swift
johnxnguyen: Seeing as the main app already uses the name, I'll merge this and change it back in share extension and also main app in a separate PR if we need to.

PR #4174, opened 2026-01-22
fix: update self-deleting system message - WPB-21395

OlivellaO:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21395" title="WPB-21395" target="_blank">BugWPB-21395</a>  [iOS] "Cells" text from system message seems moving to next line when there is space avaiable.
  </td></table>
  <br />
 

### Issue

When creating a chat with "Cells" enabled, the self-deleting message was not using the fill width available.

In order to fix that and avoid creating more abstraction or system cells, i've updated DynamicFontLabel to allow to use Attributed String instead of just String (an extra init() was added)

Adding that extra init() we can use it for ConversationWarningSystemMessageCell (updating a bit ConversationEncryptionInfoSystemMessageCellDescription) instead of ConversationSystemMessageCell for ConversationTimerSystemMessageCellDescription.

This will use the full width of the parent view.

As needed extra, padding was added to ConversationWarningSystemMessageCell so text is not showing under the scrollbar.

As a nice to have extra, ConversationFileCollaborationSystemMessageCellDescription was updated to also use ConversationWarningSystemMessageCell so if the text gets too long it will also use the full width properly.

### Testing

Steps to test:
- Create a conversation with Cells enabled
- Observe system message

---

### Checklist

---


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationEncryptionInfoSystemMessageCellDescription.swift
```diff
@@ -36,9 +36,17 @@ final class ConversationEncryptionInfoSystemMessageCellDescription: Conversation
     init(isWireCellsEnabled: Bool) {
         typealias connectionView = L10n.Localizable.Conversation.ConnectionView
 
+        let encryptationText = NSAttributedString(
```
WilhelmOks: typo: should be `encryptionText`

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationWarningSystemMessageCell.swift
```diff
@@ -57,15 +61,24 @@ final class ConversationWarningSystemMessageCell<
...
         lineView.isHidden = true
-        imageView.image = .init(resource: .attention)
         imageView.tintColor = IconColors.backgroundDefault
     }
 
     override func configureConstraints() {
         super.configureConstraints()
-        encryptionLabel.fitIn(view: topContentView)
-        sensitiveInfoLabel.fitIn(view: bottomContentView)
+        let padding: CGFloat = -6.0
```
WilhelmOks: you can make the padding positive when you apply the layout-constraints the other way around:

`topContentView.trailingAnchor.constraint(equalTo: encryptionLabel.trailingAnchor, constant: padding),`
etc.

WilhelmOks (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationEncryptionInfoSystemMessageCellDescription.swift
OlivellaO: Fixed, thanks!

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationWarningSystemMessageCell.swift
OlivellaO: Update the constraint code to make it nicer and less boilerplate

samwyndham (commented):
question: I'm a bit confused why the need to move away from `ConversationSystemMessageCell` and not change `ConversationSystemMessageCell` so that it has a greater width.

Also reminder to re-target this branch against develop

PR #4177, opened 2026-01-23
feat: file upload dialog for autorename - WPB-22584

WilhelmOks:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22584" title="WPB-22584" target="_blank">TaskWPB-22584</a>  [iOS] Disallow \ and " in file names when uploading a file
  </td></table>
  <br />
 

### Issue

The characters `\` and `"` are not allowed in filenames of uploaded files.
Before a file is uploaded and added as an attachment to a message, the code will check if the filename contains those characters and if it does, it will show an alert with two buttons:
* Cancel will keep the message in the state before adding the files to upload.
* Replace will add the files as attachments and upload them.

Files which are uploaded, are automatically renamed by replacing invalid characters with the underscore character `_`.

### Testing

* open a Wire Drive enabled conversation
* upload one or multiple files as attachments to a message
* if a filename has invalid characters, it should trigger the alert
  * cancelling the alert should do nothing and the message should have no attachments
  * confirming the alert should add the attachments and and upload the files
* send the message
* if the message contained attached files, those files should appear in the file list
  * check if the filenames of the new files don't contain the invalid characters `\` or `"` and are replaced by `_`.

---

### Checklist

---

### DIFF
--- a/WireMessaging/Sources/WireMessagingDomain/WireCells/UseCases/UploadDraftUseCase.swift
+++ b/WireMessaging/Sources/WireMessagingDomain/WireCells/UseCases/UploadDraftUseCase.swift
@@ -38,6 +38,8 @@ package struct UploadDraftUseCase: WireCellsUploadDraftUseCaseProtocol, WireCell
     private let intermediaryFilesDirectory: URL
     private let filenameGenerator: FilenameGenerator
 
+    package let charactersToReplace: [Character] = ["\\", "\""]
+
     package init(
         cellName: String,
         draftRepository: any DraftsRepositoryProtocol,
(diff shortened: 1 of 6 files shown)


WireMessaging/Sources/WireMessagingDomain/WireCells/UseCases/UploadDraftUseCase.swift
```diff
@@ -38,6 +38,8 @@ package struct UploadDraftUseCase: WireCellsUploadDraftUseCaseProtocol, WireCell
     private let intermediaryFilesDirectory: URL
     private let filenameGenerator: FilenameGenerator
 
+    package let charactersToReplace: [Character] = ["\\", "\""]
```
samwyndham, line 41: suggestion: I think exposing this is slightly breaking the idea of use cases but I'm not really sure what a better approach would be. Maybe making is `static` then it feels more like a constant. Feel free to change if you want.

samwyndham (approved):
Nice work

WireMessaging/Sources/WireMessagingDomain/WireCells/UseCases/UploadDraftUseCase.swift
WilhelmOks: I agree that this doesn't feel entirely right, however I don't see how making this static/constant would make it better. On the contrary, I think it would make it worse because static would make it bound to the protocol (rather than class/struct) and would require to make an extension on it, because protocols can't have stored properties or property getter implementations. So, it would be available on the protocol type but not on the concrete types which are conforming to this protocol.

The reason why I put it into the use case is because this info is required in two different modules and is required for the same feature and making a copy of it would introduce the likelihood of adding bugs when one copy is modified later but the other is not.

jullianm (approved):
Nice work

OlivellaO (approved):

PR #4181, opened 2026-01-23
fix: shared "Drive" tab not displaying conversation names - WPB-23032

jullianm:
WPB-23032

### Issue

This PR fixe an issue where, in the shared "Drive" tab, we don't display the conversation name the file was shared in.
It also takes this opportunity to do some preparation work for the sorting & filtering feature coming up (filtering by conversations and participants)

https://github.com/user-attachments/assets/fbda3486-a60d-4dbe-8996-92b3f0cf804e

### Testing

go to shared "Drive" tab, the subtitle of the cell should display the file owner and the conversation name it's linked (i.e "userA in conversationB")
conversation files list should remain unchanged

### Checklist

---


WilhelmOks (approved):
nice work

wire-ios-data-model/Source/Model/Conversation/ConversationLike.swift
```diff
@@ -61,7 +61,7 @@ public protocol ConversationLike: AnyObject {
     var hasMoreHistory: Bool { get }
 
     /// The name of the `cell` used for Wire Drive file management.
-    var wireCellName: String { get }
+    var wireDriveName: String { get }
```
samwyndham: thought: I wonder if we should still use the term `cell` in some cases because this is what it really is? Or maybe `node` because I guess it is the name of the root node of the conversation.  So maybe `wireDriveNodeName`? What do you think?

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Messaging.swift
```diff
@@ -258,6 +258,12 @@ public extension ZMConversation {
         return context.fetchOrAssert(request: request)
     }
 
+    static func fetchDriveConversations(in context: NSManagedObjectContext) -> [ZMConversation] {
```
samwyndham, line 261: sanity check: Do we definitely need to do this. The conversation name isn't returned as part of the node response somewhere?

samwyndham (approved):
Nice work. Left a sanity check comment as I thought I recalled that conversation name might be part of the nodes response. Maybe something to double check with Charles if you haven't already.

WilhelmOks (approved):

wire-ios-data-model/Source/Model/Conversation/ConversationLike.swift
jullianm: Yes you're right, I went autopilot mode here I shouldn't have renamed it. I'll fix that, probably `wireDriveNodeName` makes sense

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Messaging.swift
jullianm: No it's not, as far as I remember during the sorting & filtering refinement I asked Charles if we would get the conversation linked to the node in the response and he answered that we would need to get all conversations locally and pass their IDs to the API to filter properly (I think they don't map it on BE side). However, let me double check that because I might be wrong. 

Though I think we'd still need that code as we'll need to know the participants that are part of the conversation (out of scope for this ticket but I'm kind of preparing the sorting & filtering feature too).

wire-ios-data-model/Source/Model/Conversation/ConversationLike.swift
jullianm: https://github.com/wireapp/wire-ios/pull/4181/changes/1d0ecfb728dddb700bbe89b985c9d83ad791f3ca

PR #4186, opened 2026-01-26
fix: update self-deleting system message - WPB-21395

OlivellaO:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21395" title="WPB-21395" target="_blank">BugWPB-21395</a>  [iOS] "Cells" text from system message seems moving to next line when there is space avaiable.
  </td></table>
  <br />
 

### Issue

When creating a chat with "Cells" enabled, the self-deleting message was not using the fill width available.

In order to fix that and avoid creating more abstraction or system cells, i've updated DynamicFontLabel to allow to use Attributed String instead of just String (an extra init() was added)

Adding that extra init() we can use it for ConversationWarningSystemMessageCell (updating a bit ConversationEncryptionInfoSystemMessageCellDescription) instead of ConversationSystemMessageCell for ConversationTimerSystemMessageCellDescription.

This will use the full width of the parent view.

As needed extra, padding was added to ConversationWarningSystemMessageCell so text is not showing under the scrollbar.

As a nice to have extra, ConversationFileCollaborationSystemMessageCellDescription was updated to also use ConversationWarningSystemMessageCell so if the text gets too long it will also use the full width properly.

### Testing

Steps to test:
- Create a conversation with Cells enabled
- Observe system message

---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS/Sources/Helpers/DynamicFontLabel.swift
+++ b/wire-ios/Wire-iOS/Sources/Helpers/DynamicFontLabel.swift
@@ -45,7 +45,21 @@ class DynamicFontLabel: UILabel, DynamicTypeCapable {
         self.adjustsFontForContentSizeCategory = true
     }
 
-    @available(*, deprecated, message: "Use `init(text:style:color)` instead")
+    init(
+        attributedText: NSAttributedString,
+        style: WireTextStyle = .body1,
+        color: UIColor
+    ) {
+        // Not needed when we use a font style.
+        self.onRedrawFont = { nil }
+        super.init(frame: .zero)
+        self.attributedText = attributedText
+        self.textColor = color
+        self.font = .font(for: style)
+        self.adjustsFontForContentSizeCategory = true
+    }
+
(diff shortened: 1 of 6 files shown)


WilhelmOks (approved):
nice work

wire-ios/Wire-iOS/Sources/Helpers/DynamicFontLabel.swift
```diff
@@ -45,7 +45,21 @@ class DynamicFontLabel: UILabel, DynamicTypeCapable {
         self.adjustsFontForContentSizeCategory = true
     }
 
-    @available(*, deprecated, message: "Use `init(text:style:color)` instead")
+    init(
+        attributedText: NSAttributedString,
```
johnxnguyen, line 49: question: will the label still adapt when the preferred content size is changed (in iOS settings)?

OlivellaO: Yes, i've test it and it will

jullianm (approved):
LGTM

jullianm:
@OlivellaO don't forget to retarget to 4.14

PR #4191, opened 2026-01-27
fix: crash + keyboard issue on iPad - WPB-21539

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21539" title="WPB-21539" target="_blank">BugWPB-21539</a>  [iOS/iPad] Keyboard remain open on switching from conversation to Files view
  </td></table>
  <br />
 

### Issue

This PR fixes an issue where the keyboard remains open when switching to the "All" files section (in sidebar) on iPad.
This wrong keyboard behavior is also causing a weird UI issue and eventually ends up with a crash which is most likely related to that [Jira ticket](https://wearezeta.atlassian.net/browse/WPB-22075), specifically:

- Go to conversations, have the keyboard open
- Open the side bar, tap on "Folders" then tap again to dismiss the folders view.
- Observe that the "All" button from the "Files" section is being replaced by "Settings" and "Support" ones
- Go to settings then open the side bar again, "All" button is back under the Files section.
- Go to "All" conversations, then back to the "All" files -> This is causing a crash reproducible every single time.

Note: Doing the exact same STR without having the keyboard open is working fine.

The solution to avoid these is to make sure the keyboard is always hidden when we open the side bar (aka the split view is changing to an expanded state). This fixes the crash, the weird UI issue mentioned above and also avoids having a weird UX where the "Settings" and "Support" sections are hidden by the keyboard.

### Testing

See STR in description

### Checklist

---

### DIFF
--- a/WireUI/Sources/WireMainNavigationUI/Coordinator/MainCoordinator.swift
+++ b/WireUI/Sources/WireMainNavigationUI/Coordinator/MainCoordinator.swift
@@ -532,6 +532,11 @@ public final class MainCoordinator<Dependencies>: NSObject, MainCoordinatorProto
             splitViewController.conversationUI != nil
         }
     }
+
+    public func splitViewController(_ svc: UISplitViewController, willChangeTo displayMode: UISplitViewController.DisplayMode) {
+        guard .oneOverSecondary == displayMode else { return }
+        splitViewController.view.endEditing(true)
+    }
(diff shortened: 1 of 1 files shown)


caldrian (approved):

samwyndham (approved):
LGTM

PR #4195, opened 2026-01-27
fix: filter out drafts - WPB-19863

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19863" title="WPB-19863" target="_blank">TaskWPB-19863</a>  [iOS] Draft Assets Management - MVP
  </td></table>
  <br />
 

### Issue

When attaching a file to a message **without** sending it, the attached files are showing up in the list and available to other users. 
Reason is that the `isDraft` flag (set to false) passed in the API request is not yet handled by the backend meaning the drafts are not properly filtered. As a consequence, we need to handle drafts filtering on client side.

This PR introduces the logic to filter out these drafts.

Note: Since we're paginating the nodes, we need to make sure duplicates are removed. This logic is already handled in the code.

### Testing

Attach a file to a message but do not send the message. That file should no longer shows up in the files list (whether in the shared "Drive" tab or in the conversation files list)

### Checklist

---

### DIFF
--- a/WireMessaging/Sources/WireMessagingDomain/WireCells/UseCases/WireCellsFetchNodesUseCase.swift
+++ b/WireMessaging/Sources/WireMessagingDomain/WireCells/UseCases/WireCellsFetchNodesUseCase.swift
@@ -87,11 +87,7 @@ package final class WireCellsFetchNodesUseCase: Sendable {
             configuration: configuration
         )
 
-        let task = Task {
-            var (nodes, nextOffset) = try await repository.getNodes(request)
-            nodes = nodes.filter { !$0.isDraft }
-            return (nodes: nodes, nextOffset: nextOffset)
-        }
+        let task = Task { try await repository.getNodes(request) }
         currentTask = task
 
         return task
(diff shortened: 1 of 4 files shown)


WireMessaging/Sources/WireMessagingDomain/WireCells/UseCases/WireCellsFetchNodesUseCase.swift
```diff
@@ -87,7 +87,11 @@ package final class WireCellsFetchNodesUseCase: Sendable {
             configuration: configuration
         )
 
-        let task = Task { try await repository.getNodes(request) }
+        let task = Task {
+            var (nodes, nextOffset) = try await repository.getNodes(request)
+            nodes = nodes.filter { !$0.isDraft }
```
johnxnguyen: suggestion: since this logic is repeated, perhaps it can be part of the `getNodes` method:

```swift
var (nodes, nextOffset) = try await repository.getNodes(request, includeDrafts: false)
```

johnxnguyen (approved):

WireMessaging/Sources/WireMessagingDomain/WireCells/UseCases/WireCellsFetchNodesUseCase.swift
jullianm: done https://github.com/wireapp/wire-ios/pull/4195/commits/91e65dfe2cf4decd43be1f5f010f707d849dbe05 thanks

WilhelmOks (approved):

PR #4214, opened 2026-01-28
fix: app freeze and crashed when internet is reconnected during the call - WPB-19210

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19210" title="WPB-19210" target="_blank">BugWPB-19210</a>  [iOS] App freeze and crashed when internet is reconnected during the call
  </td></table>
  <br />
 

### Issue

1:1 call freezes and does not recover after ~30 seconds of being offline.

### Solution

Call the `wcall_network_changed` method when network connectivity is restored thus largely improving the re-connect success rate.
In order to improve the reconnection in a 1:1 call after network has been lost **AVS team** did the following:
- Toggle the `aestab` flag in 1:1 calls when a reconnect per above is being done. This allows the clients to display the "Connecting.." banner for the users in the 1:1 call, more graciously informing the user about what is happening.

These changes to AVS revealed an issue that had been present in our code for a long time:
for 1:1 calls via SFT (`useSFTForOneToOneCalls` feature flag) we should end the call if [aestab goes from 1 to 0](https://wearezeta.atlassian.net/wiki/spaces/PAD/pages/1314750477/2024-07-29+1+1+calls+over+SFT), but we didn't have a check for the `useSFTForOneToOneCalls` feature flag. This fix is a part of this PR

### Testing

- establish 1:1 video call
- turn on airplane mode for ~30 seconds

### Checklist

---


David-Henner (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -707,6 +708,10 @@ public final class ZMUserSession: NSObject {
         userSessionComponent = nil
         syncStateCancellable?.cancel()
         syncStateCancellable = nil
+        networkInterfaceSwitchCancellable?.cancel()
+        networkInterfaceSwitchCancellable = nil
+        isNetworkReachableCancellable?.cancel()
+        isNetworkReachableCancellable = nil
```
samwyndham: question: I don't really understand why this code is needed because:

1. networkInterfaceSwitchCancellable & isNetworkReachableCancellable will become nil automatically  when this class deallocates. 
2. the streams should cancel when the their _cancellables_ are deallocated.

Did you notice a bug where this wasn't happening?

samwyndham (commented):
Looks good but I have a question about some of the clean up code.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
KaterinaWire: You are right, we don't need it. This is redundant cancellable cleanup. `AnyCancellable` automatically calls cancel() in its own deinit, so explicitly canceling and niling out these properties is unnecessary. When `ZMUserSession` deallocates, the stored cancellables will be deallocated and automatically cancel their subscriptions.                                                                                                                                                                                                                                                                                                                                                                                                                                        Since both subscriptions use `[weak self]` captures, there are no retain cycles to break.

samwyndham (approved):
Nice work!

PR #4218, opened 2026-01-29
fix: crash loading multiple user session for same account concurrently - WPB-22928

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22928" title="WPB-22928" target="_blank">BugWPB-22928</a>  [iOS] WireSystem: fatal(_:file:line:)
  </td></table>
  <br />
 

### Issue

There is a crash caused by multiple concurrent calls to `SessionManager.withSession(for:newEnvironment:notifyAboutMigration:)` with the same Account when there is no existing session in `backgroundUserSession` for that account.

Each successive call may cause a new `ZMUserSession` to be created and associated with an Account in the session manager. `SessionManager.configure(session:for:)` is called before the session is returned. It contains a precondition that no session is already loaded for that account. This is the crash. Specifically:

1. `SessionManager.withSession(for:newEnvironment:notifyAboutMigration:)` is called which starts creating a new session for the given account.
2. `SessionManager.withSession(for:newEnvironment:notifyAboutMigration:)` is called again before  `1` completes and also starts setting up a new session for the same account.
3. `1` completes and return the new session.
4. `2` creates a new session but crashes when it tries to configure it because there is already an existing session for that account.

There are likely many situations where `SessionManager.withSession(for:newEnvironment:notifyAboutMigration:)` can be called for the same account currently. In the specific case I looked into there was:

1. One call after app instantiation.
2. 5 seconds later one kicked off by `SessionManager.lookupConversationAndProcessPendingCallEvents(by:completionHandler:)`

Perhaps this happened when receiving a call on bad internet?

#### Fix

The fix is to put the user session loading code into a Task and have concurrent calls wait on the task to get a session.

### Testing

I'm not sure of a good way of testing this. I never managed to reproduce the crash. Anyone have any ideas?

---

### Checklist

---

### DIFF
--- a/WireDomain/Sources/WireDomain/Utilities/NonReentrantTaskManager.swift
+++ b/WireDomain/Sources/WireDomain/Utilities/NonReentrantTaskManager.swift
@@ -24,7 +24,7 @@ import Foundation
 /// at any given time, regardless how many times the block is enqueued.
 /// Repeated invocations don't result in repeated executions.
 
-public actor NonReentrantTaskManager<Success, Failure>: Sendable where Success: Sendable, Failure: Error {
+public actor NonReentrantTaskManager<Success, Failure> where Success: Sendable, Failure: Error {
 
     private var state: TaskState = .idle
 
--- a/WireDomain/Sources/WireDomain/Utilities/NonReentrantTaskManager.swift
+++ b/WireDomain/Sources/WireDomain/Utilities/NonReentrantTaskManager.swift
@@ -39,6 +39,11 @@ public actor NonReentrantTaskManager<Success, Failure> where Success: Sendable, 
 
 }
 
+// MARK: - Public API
+
+/// Due to limitations of `Task` - initializers are constrained to either `any Error` or `Never` type Failures - we need
+/// to provide two separate APIs for throwing and non-throwing implementations.
+
 public extension NonReentrantTaskManager where Failure == any Error {
 
     /// Perform a non-reentrant async block that cannot throw.
--- a/WireDomain/Sources/WireDomain/Utilities/NonReentrantTaskManager.swift
+++ b/WireDomain/Sources/WireDomain/Utilities/NonReentrantTaskManager.swift
@@ -24,13 +24,30 @@ import Foundation
 /// at any given time, regardless how many times the block is enqueued.
 /// Repeated invocations don't result in repeated executions.
 
-public actor NonReentrantTaskManager {
+public actor NonReentrantTaskManager<Success, Failure>: Sendable where Success: Sendable, Failure: Error {
 
     private var state: TaskState = .idle
 
     public init() {}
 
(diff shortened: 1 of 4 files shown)


wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -381,6 +381,13 @@ public final class SessionManager: NSObject, SessionManagerType {
 
     private let sharedUserDefaults: UserDefaults
 
+    /// Currently active `Tasks` loading sessions by user ID
+    ///
+    /// Used by `withSession(for:newEnvironment:notifyAboutMigration:)` to avoid creating & loading multiple sessions
+    /// for the same account concurrently.
+    @MainActor
+    private var inFlightWithSessionTasks: [UUID: Task<ZMUserSession?, Never>] = [:]
```
johnxnguyen: question: could we use `NonReentrantTaskManager` or a variation of it?

samwyndham: Oh I'll take a look. I forgot about that

samwyndham: I've updated this in my latest push.

WireDomain/Sources/WireDomain/Utilities/NonReentrantTaskManager.swift
```diff
@@ -24,21 +24,38 @@ import Foundation
...
+    private enum TaskState {
+
+        case idle
+        case inFlight(Task<Success, Failure>)
+
+    }
+
+}
+
+public extension NonReentrantTaskManager where Failure == any Error {
```
samwyndham, line 47: Unfortunately I needed to duplicate the implementation here for 2 different generic options due a  limitation with `Task` instantiation - it is only possible to instantiate a Task where `Failure` is `Never` or `any Error`.

If you can think of a better way let me know.

WireDomain/Sources/WireDomain/Utilities/NonReentrantTaskManager.swift
```diff
@@ -24,21 +24,38 @@ import Foundation
 /// at any given time, regardless how many times the block is enqueued.
 /// Repeated invocations don't result in repeated executions.
 
-public actor NonReentrantTaskManager {
+public actor NonReentrantTaskManager<Success, Failure> where Success: Sendable, Failure: Error {
```
samwyndham, line 27: I've extended this actor to allow returning a value and to allow not throwing an Error

WireDomain/Sources/WireDomain/Utilities/NonReentrantTaskManager.swift
johnxnguyen: suggestion: leave a comment explaining the duplication.

johnxnguyen (approved):

netbe (approved):
LGTM;)

PR #4241, opened 2026-02-02
chore: improve code readability of SearchTask.swift - WPB-20362

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20362" title="WPB-20362" target="_blank">TaskWPB-20362</a>  [iOS/Integration] - Searching for Apps specifically
  </td></table>
  <br />
 

### Issue

As a preparation for subsequent PRs this PR just tries to improve the readability of SearchTask.swift.

---

### Checklist

---


wire-ios-sync-engine/Source/UserSession/Search/SearchTask.swift
```diff
@@ -677,7 +676,7 @@ extension SearchTask {
         var url = URLComponents()
         url.path = "/users"
         url.queryItems = [URLQueryItem(name: "handles", value: handle)]
-        let urlStr = url.string?.replacingOccurrences(of: "+", with: "%2B") ?? ""
```
caldrian, line 680: Doesn't seem to be needed, also no other client does it.

wire-ios/Wire-iOS/Sources/UserInterface/UserProfile/SearchUserViewController.swift
```diff
@@ -137,7 +137,7 @@ final class SearchUserViewController: UIViewController {
 
             navigationController?.setViewControllers([profileViewController], animated: true)
             resultHandled = true
-        } else if isCompleted {
```
caldrian, line 140: Because there's a guard at the top the func this will always evaluate to true.

wire-ios-sync-engine/Source/UserSession/Search/SearchTask.swift
jullianm: question: I wonder why we introduced this. it doesn't break the API request or introduce weird behaviors right ? 

Asking because we had a percent encoding [bug](https://wearezeta.atlassian.net/browse/WPB-23207) in Wire drive that came from when we wrapped the path into a URL (for convenient path manipulation API) and then back to a string with `.absoluteString`. 

Probably unrelated but this line reminds me of this bug.

KaterinaWire: @caldrian have you tried to test and search similar to "katerina+test1@wire.com"

caldrian: Thank you both!
You're probably right. `+` is sometimes interpreted as space, so escaping makes sense.
Reverted this change and added one more escaping.

jullianm (approved):
LGTM

KaterinaWire (approved):

PR #4244, opened 2026-02-02
fix: files percent encoding - WPB-23207

WilhelmOks:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23207" title="WPB-23207" target="_blank">BugWPB-23207</a>  [iOS] Files and folders with spaces and special characters created on iOS have wrong names on Android and Web
  </td></table>
  <br />
 

### Issue

Renaming files was adding percent encoding like `%20` for space ` `.
This was because the file path was wrapped into a URL and then was converted to a String using `.absoluteString`.
I fixed that by calling `.removingPercentEncoding` before that string is used.

The percent encoding isn't visible in iOS but you can check it in the web app.

I also tested creating folders, moving folders and uploading drafts. None of those cases have that issue in 4.14, so no code needs to be changed there.

We should also test if creating new files for Collabora (templates) is affected, but that feature is not in 4.14.

### Testing

Creating, moving and renaming files and folders should not cause the file names to have percent encodings. Percent encodings are only visible in the web app (and probably Android)

---

### Checklist

---

### DIFF
--- a/WireMessaging/Sources/WireMessagingDomain/WireCells/UseCases/WireCellsRenameNodeUseCase.swift
+++ b/WireMessaging/Sources/WireMessagingDomain/WireCells/UseCases/WireCellsRenameNodeUseCase.swift
@@ -67,10 +67,12 @@ package struct WireCellsRenameNodeUseCase: WireCellsRenameNodeUseCaseProtocol {
             let directory = url.deletingLastPathComponent()
             targetPath = directory.appendingPathComponent("\(newFilename).\(pathExtension)")
         }
+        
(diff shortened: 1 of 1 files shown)


jullianm (approved):
Nice work, thanks for fixing this.

WilhelmOks:
Will create a new PR with base `release/cycle-4.15`

WilhelmOks:
reopening since this should go into 4.14.1 hotfix

johnxnguyen (approved):

PR #4242, opened 2026-02-02
chore: replace old search requests - WPB-20362

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20362" title="WPB-20362" target="_blank">TaskWPB-20362</a>  [iOS/Integration] - Searching for Apps specifically
  </td></table>
  <br />
 

### Issue

This PR replaces legacy networking with new code in WireNetwork.

### Testing

From the user's perspective nothing should change.

---

### Checklist

- [ ] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [ ] Description is filled and free of optional paragraphs.
- [ ] Adds/updates automated tests.

---

### UI accessibility checklist

_If your PR includes UI changes, please utilize this checklist:_
- [ ] Make sure you use the API for UI elements that support large fonts.
- [ ] All colors are taken from WireDesign.ColorTheme or constructed using WireDesign.BaseColorPalette.
- [ ] New UI elements have Accessibility strings for VoiceOver.

PR #4245, opened 2026-02-02
chore: refactor SearchTask - WPB-23110

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23110" title="WPB-23110" target="_blank">BugWPB-23110</a>  [iOS] fix Swift Concurrency API misuse, leaking continuation
  </td></table>
  <br />
 

### Issue

Currently `SearchTask`'s different subtasks for searching contacts is implemented in a fragile way (see `tasksRemaining` property). Cancelling a task can lead to an async continuation leaking.

This PR refactors the code in order to make use of Swift Concurrency for a more stable implementation.

### Testing

Ideally there shouldn't be any differences from the user's perspective.

---

### Checklist

---


wire-ios-sync-engine/Source/UserSession/Search/SearchRequest.swift
```diff
@@ -66,6 +66,7 @@ public struct SearchOptions: OptionSet {
 }
 
 public extension SearchOptions {
+
     mutating func updateForSelfUserTeamRole(selfUser: UserType) {
```
jullianm, line 70: suggestion: since we're explicitly mutating self, it probably makes sense to turn the struct into a class (unless there's a specific reason to keep it a struct). What do you think?

wire-ios-sync-engine/Source/UserSession/Search/SearchDirectory.swift
```diff
@@ -116,13 +92,13 @@ public class SearchDirectory: NSObject {
     }
 }
 
-extension SearchDirectory: TearDownCapable {
```
jullianm, line 119: question: do you know why we had this protocol conformance in the first place or is it just unused code ? Looks like it's only called from the search related objects so I guess it's OK. Probably worth checking though whether it was not called automatically (using the protocol) from somewhere else in the code (probably not).

wire-ios-sync-engine/Source/UserSession/Search/SearchTask.swift
```diff
@@ -16,221 +16,194 @@
...
-    public func addResultHandler(_ resultHandler: @escaping ResultHandler) {
-        resultHandlers.append(resultHandler)
-    }
-
     /// Cancel a previously started task
     public func cancel() {
-        resultHandlers.removeAll()
+        guard status == .running else { return assertionFailure() }
+        status = .cancelled
+    }
```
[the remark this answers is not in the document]
caldrian: fixed

wire-ios-sync-engine/Source/UserSession/Search/SearchRequest.swift
caldrian: `SearchOptions` is an `OptionSet`. I don't think we should change anything here.

wire-ios-sync-engine/Source/UserSession/Search/SearchDirectory.swift
caldrian: I thought it's never used, but I'll just reverted this change. Thanks for the hint.

wire-ios-sync-engine/Source/UserSession/Search/SearchTask.swift
```diff
@@ -492,85 +460,81 @@ extension SearchTask {
...
+                var updatedResult = searchResult
+                updatedResult.extendWithMembershipPayload(payload: payload)
+                updatedResult.filterBy(
+                    searchOptions: searchRequest.searchOptions,
+                    query: searchRequest.query.string,
+                    contextProvider: contextProvider
+                )
 
-        })
+                continuation.resume(returning: { $0 = $0.union(withDirectoryResult: searchResult) })
```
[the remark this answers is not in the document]
caldrian: fixed

jullianm (approved):
LGTM

wire-ios-sync-engine/Source/UserSession/Search/SearchTask.swift
```diff
@@ -16,221 +16,199 @@
...
+            taskGroup.addTask {
+                await self.performUserLookup()
+            }
+
+            var result = SearchResult()
+            while let aggregator = await taskGroup.next() {
+                aggregator(&result)
+            }
+
+            if Task.isCancelled { // TODO: [WPB-20362] make this code throwing
```
caldrian, line 122: The TODO is resolved in PR https://github.com/wireapp/wire-ios/pull/4242/

wire-ios-sync-engine/Source/Use cases/SearchUsersUseCase.swift
```diff
@@ -72,24 +72,14 @@ public final class SearchUsersUseCase: SearchUsersUseCaseProtocol {
             team: team
         )
 
-        return try await withCheckedThrowingContinuation { continuation in
-            // TODO: [WPB-23110] SWIFT TASK CONTINUATION MISUSE: invoke(query:options:messageProtocol:) leaked its continuation without resuming it. This may cause tasks waiting on it to remain suspended forever.
-            guard !Task.isCancelled else {
-                continuation.resume(throwing: CancellationError())
-                self.activeSearchTask = nil
-                return
```
netbe, line 80: question: why do we need to retain the task here? is the `===` on purpose ?

wire-ios-sync-engine/Source/UserSession/Search/SearchTask.swift
```diff
@@ -492,85 +465,81 @@ extension SearchTask {
...
+            let teamID,
             !teamMembersIDs.isEmpty
         else {
-            completeRemoteSearch(searchResult: searchResult)
-            return
+            return { $0 = $0.union(withDirectoryResult: searchResult) }
         }
 
-        let request = type(of: self).fetchTeamMembershipRequest(
+        let request = Self.fetchTeamMembershipRequest(
```
netbe, line 508: suggestion: it seems you could use the WireNetwork getTeamMembers on the teamsAPI

wire-ios-sync-engine/Source/UserSession/Search/SearchTask.swift
```diff
@@ -711,35 +677,31 @@ extension SearchTask {
 
 extension SearchTask {
 
-    func performRemoteSearchForServices() {
+    func performRemoteSearchForServices() async -> SearchResultAggregator {
+
         let searchContext = contextProvider.newBackgroundContext()
```
netbe, line 682: question: so we use a complete new context here? i wonder if it could be an issue with EAR @David-Henner

netbe (commented):
Looks good, left some comments and questions before approving

wire-ios-sync-engine/Source/UserSession/Search/SearchTask.swift
David-Henner: My changes will ensure the background contexts created via the contextProvider have access to EAR capabilities, so it's safe to use.

wire-ios-sync-engine/Source/Use cases/SearchUsersUseCase.swift
caldrian: We keep an instance of the task in order to be able to call `cancel()` on it when the query changes and a new search is started (when the user edited the search field while the last search hasn't finished).

For every new search an instance of `SearchTask` is created.
`activeSearchTask` should always reference the latest non-cancelled search task. So after `task.start()` returns, we want to clean up the `activeSearchTask` property but only if there hasn't been another search task started already. (we only set it to `nil` if the property still references the very same instance that was set in the beginning of `invoke`).

wire-ios-sync-engine/Source/UserSession/Search/SearchTask.swift
caldrian: Yup, that's part of the subsequent PR: https://github.com/wireapp/wire-ios/pull/4242

netbe (approved):
LGTM:) cool cleanup

PR #4254, opened 2026-02-03
fix: commit pending proposals recovery strategy - WPB-23159

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23159" title="WPB-23159" target="_blank">TaskWPB-23159</a>  [iOS] Change Commit Pending Proposals recovery
  </td></table>
  <br />
 

### Issue

* clear commit pending proposal date when mls reset
* change recovery strategy to a backoff retry instead of retryaftersync
* avoid sending commit to non existing groups
* clear scheduled items from workAgent queue when sync is suspended. These will be recreated if needed.
* cleanup setup initiate mls reset use case
* improve debugging workAgent item (changed ids)

### Testing

-
---

### Checklist

---

### DIFF
--- a/WireDomain/Sources/WireDomain/WorkAgent/WorkItem/RepairFaultyMLSRemovalKeysWorkItem.swift
+++ b/WireDomain/Sources/WireDomain/WorkAgent/WorkItem/RepairFaultyMLSRemovalKeysWorkItem.swift
@@ -30,7 +30,7 @@ struct RepairFaultyMLSRemovalKeysWorkItem: WorkItem {
     var id: String {
         "repairFaultyMLSRemovalKeys_\(UUID().uuidString)"
     }
-    
+
     var priority: WorkItemPriority {
         .low
     }
--- a/WireDomain/Sources/WireDomain/WorkAgent/WorkItem/RepairFaultyMLSRemovalKeysWorkItem.swift
+++ b/WireDomain/Sources/WireDomain/WorkAgent/WorkItem/RepairFaultyMLSRemovalKeysWorkItem.swift
@@ -27,7 +27,10 @@ import WireLogging
 
 struct RepairFaultyMLSRemovalKeysWorkItem: WorkItem {
 
-    let id = UUID()
+    var id: String {
+        "repairFaultyMLSRemovalKeys_\(UUID().uuidString)"
+    }
+    
     var priority: WorkItemPriority {
         .low
     }
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -1442,19 +1442,9 @@ public final class MLSService: MLSServiceInterface {
 
     enum RecoveryStrategy: Equatable {
 
-        /// Perform a quick sync, then retry the action in its entirety.
+        /// Retry the action in its entirety.
         ///
-        /// Core Crypto can not automatically recover by itself. It needs
-        /// to process incoming handshake messages then generate a new commit.
-
-        case retryAfterQuickSync
-
-        /// Repair (re-join) the group and retry the action
-        ///
-        /// We may have missed a few commits so we will rejoin the group
-        /// and try again.
-
-        case retryAfterRepairingGroup
+        case retryAfterBackoff
 
         /// Add the missing users to the group, then retry the action in
         /// its entirety.
@@ -1488,10 +1478,8 @@ public final class MLSService: MLSServiceInterface {
             }
 
             switch error {
-            case .mlsClientMismatch, .mlsCommitMissingReferences:
-                self = .retryAfterQuickSync
-            case .mlsStaleMessage:
-                self = .retryAfterRepairingGroup
+            case .mlsClientMismatch, .mlsCommitMissingReferences, .mlsStaleMessage:
+                self = .retryAfterBackoff
             case .mlsInvalidLeafNodeIndex, .mlsInvalidLeafNodeSignature:
                 self = .resetBrokenMLSConversation
             case let .groupOutOfSync(missingUsers):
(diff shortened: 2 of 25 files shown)


wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1517,49 +1504,16 @@ public final class MLSService: MLSServiceInterface {
...
+                    )
+                    try await operation()
                 }
 
-                var currentRetryCount = retryCount
-                currentRetryCount += 1
-
-                try await retryOnCommitFailure(for: groupID, operation: operation, retryCount: currentRetryCount)
-
-            case .retryAfterRepairingGroup:
```
johnxnguyen, line 1547: question: do we not need to handle this anymore?

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1517,49 +1504,16 @@ public final class MLSService: MLSServiceInterface {
...
-                    // If MLS conversation reset is DISABLED we quarantine the group to avoid repeated commit attempts
-                    // otherwise assume that things will sort themselves out through conversation reset.
-                    let feature = await featureRepository.fetchAllowedGlobalOperations()
-                    if feature.status == .disabled || feature.config.mlsConversationReset == false {
-                        brokenGroupIDs.insert(groupID)
-                    }
+            case .retryAfterBackoff:
 
-                    throw MLSRetryError.retryLimitReached
+                try await BackoffRetrier(policy: .init(maxRetries: 2)).retry { [logger] in
```
johnxnguyen: issue: the backoff retrier internally sets up observation of the network reachability and will reset under certain conditions, we may need to decouple that from the retrier and place it in the sync agent where I think it's relevant.

wire-ios-data-model/Source/MLS/MLSService.swift
netbe: This translates to mlsStaleMessage so we should do a backoff in case of POST commit bundle EXCEPT:
when sending an application message OR commitPendingProposals in MessageSender.

but I realise that MessageSender should not use the same commitPendingProposal method than internally to MLSService...

TLDR:
2 cases:

- MessageSender's commitPendingProposals (do the CC operation and handle the error in caller with repairGroup for staleMEssage otherwise retry
- MLSService's commitPendingProposals never repair

samwyndham (commented):
@netbe Remember to re-target this to 4.15 release branch

wire-ios-data-model/Source/MLS/MLSService.swift
netbe: for now i simply disable it for this case

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -1517,49 +1512,16 @@ public final class MLSService: MLSServiceInterface {
...
-                    // otherwise assume that things will sort themselves out through conversation reset.
-                    let feature = await featureRepository.fetchAllowedGlobalOperations()
-                    if feature.status == .disabled || feature.config.mlsConversationReset == false {
-                        brokenGroupIDs.insert(groupID)
-                    }
-
-                    throw MLSRetryError.retryLimitReached
+                try await BackoffRetrier(policy: .init(maxRetries: 2), monitoringNetwork: false).retry { [logger] in
+                    logger.warn(
+                        "failed to send commit due to \(reason). retrying operation with backoff - attempt: \(retryCount)...",
```
samwyndham: sanity check: Is `retryCount` here correct considering the use of the BackoffRetrier?

WireDomain/Sources/WireDomain/WorkAgent/WorkItem/RepairFaultyMLSRemovalKeysWorkItem.swift
```diff
@@ -27,7 +27,10 @@ import WireLogging
 
 struct RepairFaultyMLSRemovalKeysWorkItem: WorkItem {
 
-    let id = UUID()
+    var id: String {
```
samwyndham, line 30: question: What is the purpose of making this a string?

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
```diff
@@ -532,3 +566,32 @@ private extension Payload.ClientListByQualifiedUserID {
...
+            from: Data(reason.utf8)
+        ) else {
+            return nil
+        }
+
+        switch error {
+        case .mlsClientMismatch:
+            self = .mlsClientMismatch
+        case .mlsCommitMissingReferences:
+            self = .mlsCommitMissingReferences(message: "")
```
samwyndham, line 584: question: Would it make sense to pass the `reason` as the message?

samwyndham (commented):
Looks good. I've added just a couple of comments before approval

WireDomain/Sources/WireDomain/WorkAgent/WorkItem/RepairFaultyMLSRemovalKeysWorkItem.swift
netbe: when we log item in workAgent only the id is visible, which makes it hard to know what kind of item is dropped / failed / started. I wanted something to help with it so basically including more info in the logs

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
netbe: reason is a JSON string which unfortunately does not contain the message. I think there's a mismatch in error definition.

Backend sends error in this format:
```
{ 
code: 400,
message: "",
label: ""
}
```
The message _only_ helps to debug.

I think the issue is that MLSAPIError don't always attached the message to the case. so it can't be passed here.

So it's a bigger topic to review the SendMLSMessageFailure and MLSAPIError mapping

samwyndham (approved):
Nice work

WireDomain/Sources/WireDomain/WorkAgent/WorkItem/RepairFaultyMLSRemovalKeysWorkItem.swift
samwyndham: I see. I would personally be tempted to make `WorkItem` conform to `LogConvertible` or something like that. But I don't have a good argument why. Feel free to disregard this comment.

johnxnguyen (approved):

PR #4258, opened 2026-02-03
fix: strategy for repairing broken mls groups - WPB-23166

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23166" title="WPB-23166" target="_blank">TaskWPB-23166</a>  [iOS] Cleanup recovery strategy for mlsService
  </td></table>
  <br />
 

### Issue

When a broken mls group is detected ("wrong epoch" error when decrypting a message), the MLS service will attempt to perform an incremental sync and then repair the group. Triggering syncs as a recovery mechanism is no longer correct, instead the group ID should be noted and the group then repair explicitly later in time.

This PR removes the automatic sync behavior and introduces the handling to track the broken group id and schedule a repair work item.

### Testing
?

---

### Checklist

---


WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -320,6 +323,16 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
...
             for envelope in envelopes {
+
+                if DeveloperFlag.ignoreIncomingEvents.isOn {
+                    logger.warn(
+                        "ignore incoming events",
+                        attributes: .incrementalSyncV3 + [.eventEnvelopeID: envelope.id]
+                    )
+                    await acknowledgeUntilEnvelope(envelope, through: pushChannel, batchSize: 1)
+                    continue
+                }
```
samwyndham, line 334: question: What is this for? Testing?

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
```diff
@@ -350,6 +351,7 @@ public final class ClientSessionComponent {
     // MARK: High level syncs
 
     public private(set) lazy var syncStateSubject = CurrentValueSubject<SyncState, Never>(.idle)
+    private(set) lazy var liveBrokenGroupSubject = PassthroughSubject<Set<String>, Never>()
```
samwyndham, line 354: sanity check: is there is any issue with this being a passthrough subject vs a current value subject 🤔. Does it matter that new subscribers won't get the existing IDs?

I assume that the setup is called before we have any broken groups have been discovered.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
netbe: yes, it was there before in legacy sync. It skips events so when you set it off again you're out of sync in mls conversations

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
netbe: yes, we only want to get live events so the current value does not matter, it's just a signal to repair groups when the websocket is opened

johnxnguyen (commented):
I started this PR but @netbe finished it. I can't approve it explicitly but I have reviewed it and do approve.

WireDomain/Sources/WireDomain/Components/ClientSessionComponent.swift
```diff
@@ -876,4 +880,14 @@ public final class ClientSessionComponent {
         workAgent: workAgent
     )
 
+    public func setupLiveMLSBrokenGroups() -> AnyCancellable {
+        liveBrokenGroupSubject.sink { [weak self]  liveMLSBrokenGroups in
+            guard let self else { return }
+            WireLogger.mls.debug("detected during live sync \(liveMLSBrokenGroups.count) broken MLS groups")
+            let item = RepairBrokenMLSGroupsItem(repairAgent: mlsGroupRepairAgent)
+            Task { [weak self] in
+                await self?.workAgent.submitItem(item)
```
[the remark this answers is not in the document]
netbe: fixed

netbe (approved):

samwyndham (approved):

PR #4261, opened 2026-02-04
fix: ci failure tests - WPB-23003

findms:
### Issue

**Flaky tests observed in v4.15.**

Changes:
- Wait to perform action when sync is in progress
- timeout increased for a test where splash screen shows for long period of time and test fail due to that
- branch name missing in info on summary - added

### Testing

_Describe how to test._

ran locally - works fine
image

---

### Checklist

---


wire-ios/WireUITests/Pages/OptionsOnSettingsPage.swift
```diff
@@ -53,13 +53,12 @@ class OptionsOnSettingsPage: PageModel {
         let springboard = XCUIApplication(bundleIdentifier: "com.apple.springboard")
         let passcodeField = springboard.secureTextFields["Passcode field"].firstMatch
 
-        guard passcodeField.waitAndTap()
-        else {
-            XCTFail("Passcode SecureTextField did not appear")
-            throw XCTSkip("Passcode field not available")
+        // Sometimes the splashscreen stays visible for unexpectedly long.
```
netbe, line 56: suggestion: it's good as a workaround, but to me we should investigate what's going on:
- when does this happen?
- and extract the app logs when this happens

it could hide a bigger issue

wire-ios/WireUITests/Pages/ConversationsPage.swift
```diff
@@ -76,7 +76,21 @@ class ConversationsPage: PageModel {
...
+        }
+
+        // Fallback: Sometimes observed on CI that accountProfileImageView isn't tappable after login.
+        let userAccountButton = app.buttons
+            .containing(NSPredicate(format: "value CONTAINS[c] %@", "account"))
+            .firstMatch
+
+        if userAccountButton.waitAndTap(timeout: 5.0) {
+            return try UserProfilePage()
+        }
```
netbe: suggestion: it would be interesting to know why it's not tappable, because the issue could be somewhere else

netbe (approved):
the changes are ok, but please look at the suggestions, this could hide bigger issues

David-Henner (commented):
Maybe there's another issue somewhere ? I don't feel comfortable working around it unless we're sure there's not something broken

wire-ios/WireUITests/Pages/OptionsOnSettingsPage.swift
findms: Completely agree to avoid any workarounds but currently the 2 tests flaky behaviour is weird on Ci not on local:
- For the test where it fails to open user profile - it simply closes the app when app is in sync doesn;t wait for the next page element which is ~15 seconds

https://github.com/user-attachments/assets/2574689b-beb9-48d6-b759-bc07b1e41123

- Foe the seconds issue: with app lock with passcode - coming from background to foreground, sometime it takes loner on splashscreen so increase the time out and also observed that, without performing tap() if we directly typeText() it works as fallback in case of secureText is not able to locate.

findms: Found it.. the tap is actually being performed not on the exact button.
image

wire-ios/WireUITests/Pages/ConversationsPage.swift
findms: removed this fallback as not needed.

wire-ios/WireUITests/Pages/OptionsOnSettingsPage.swift
findms: Locally I fail to replicate. This same function is being used few places but all works fine except one test.

findms:
> Maybe there's another issue somewhere ? I don't feel comfortable working around it unless we're sure there's not something broken

hey @David-Henner I have revert the fallbacks, but found a potential reason when app is in sync mode fails to perform tap().

David-Henner (approved):

findms:
Cherrypicked the changes to another PR pointing to develop.

PR #4266, opened 2026-02-04
fix: open file guid filename - WPB-23112

WilhelmOks:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23112" title="WPB-23112" target="_blank">BugWPB-23112</a>  [iOS] Opening a pdf file shows GUID in place of file name
  </td></table>
  <br />
 

### Issue

Files are downloaded locally and are given a long unreadable id as the filename. When we open them for preview, we see that unreadable filename.

My first idea was to just change the appearance of the filename in the preview view, but that's not possible because the "quicklook preview" system component that we are using always shows the filename from the source URL that it was given.

We can't just download the files with their readable filename because there would be conflicts with same filenames from different conversations, different folders and different file versions (from editing).

Temporary copying a file just for preview would also be not optimal, especially when the file is large.

So my solution is to download the files into folders with the names of the former filenames (unreadable ids) and inside those folders the filenames are the readable proper names of the files.
The quicklook preview doesn't show the folder names but only the clean filenames.
And there should be no conflicts of filenames because each of them is in a unique folder.

### Testing

* choose a file that you want to open in fullscreen preview
* delete it and then restore again to clear the cached download
* open the file
  * it should start the download and then open automatically. The file name should be the readable original filename.
* open the file again
  * it should open right away because it's already downloaded. The file name should be the readable original filename.

---

### Checklist

---


samwyndham (commented):
question: We are using the UIKit version of quicklook previews now right? Did you attempt to set the filename using https://developer.apple.com/documentation/quicklookui/qlpreviewitem/previewitemtitle

WilhelmOks:
> question: We are using the UIKit version of quicklook previews now right? Did you attempt to set the filename using https://developer.apple.com/documentation/quicklookui/qlpreviewitem/previewitemtitle

There is a place where we use the UIKit version, but that is just a workaround for a bug that was causing the quicklook preview to automatically close on ipad.
The code here is using the SwiftUI modifier. I did a quick research and apparently it's not possible to change the displayed title with the SwiftUI version.
And even if it was possible, it's not clear if the new filename would apply to all of the functions like "Save in Files", that are integrated in the system Quick Look view.

WireMessaging/Sources/WireMessagingData/WireDrive/WireDriveLocalAssetRepository.swift
```diff
@@ -131,11 +131,23 @@ package final class WireDriveLocalAssetRepository: WireDriveLocalAssetRepository
...
+            let filename = node.path.split(separator: "/").last.flatMap(String.init) ?? "-"
+
             var asset = try verifyAsset(nodeID: nodeID, eTag: eTag)
-            try await fileCache.saveFile(at: tempURL, key: asset.cacheKey)
+
+            let extensionComponents = asset.cacheKey.split(separator: ".")
+            let pathWithoutExtension: String = if extensionComponents.count > 1 {
+                String(extensionComponents.dropLast().joined(separator: "."))
+            } else {
+                asset.cacheKey
```
caldrian, line 142: Does it make sense to cover the `else` branch with the unit tests?

caldrian (approved):

WireMessaging/Sources/WireMessagingData/WireDrive/WireDriveLocalAssetRepository.swift
WilhelmOks: Now that you mention it, I think the else case will never trigger, actually. Because the cacheKey is built always with the pattern: 
`uuid + "_" + some_other_id + "." + file_extension`
and then the new key becomes `uuid + "_" + some_other_id + "/" + filename_with_extension`
The else case is just a defensive fallback.
We can't test it.

WilhelmOks: Actually I don't know what would happen if we had files without a file extension.

WilhelmOks: I added a unit test for files without extensions. Thanks for the suggestion.

wire-ios-data-model/Source/Model/Message/FileAssetCache.swift
```diff
@@ -945,14 +945,19 @@ private struct FileCache: Cache {
...
+        let destinationFolder = toUrl.deletingLastPathComponent()
+        if !FileManager.default.fileExists(atPath: destinationFolder.path) {
+            try? FileManager.default.createDirectory(at: destinationFolder, withIntermediateDirectories: true)
+        }
+
         let coordinator = NSFileCoordinator()
 
         var error: NSError?
-        coordinator.coordinate(writingItemAt: toUrl, options: .forReplacing, error: &error) { url in
+        coordinator.coordinate(writingItemAt: toUrl, options: .forMoving, error: &error) { url in
```
jullianm, line 960: question: When deleting the asset `deleteAssetData(_ key: String)` we remove the item (the asset) but not the directory the file was in. Shouldn't we also clean up and delete the directory if it no longer contains anything?

jullianm (commented):
left one question before approving

wire-ios-data-model/Source/Model/Message/FileAssetCache.swift
WilhelmOks: It's the cache folder and I think it's fine to have empty folders there. They don't take up a meaningful amount of space and the system can clean it up automatically.

OlivellaO (approved):

jullianm (approved):
Nice work!

PR #4267, opened 2026-02-04
fix: duplicate green banner - WPB-23037

WilhelmOks:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23037" title="WPB-23037" target="_blank">BugWPB-23037</a>  [iOS] Duplicate initial system messages
  </td></table>
  <br />
 

### Issue

For some 1:1 conversations, the green welcome banner is appearing twice.
The code for the 1:1 green banner is in a complete different place than the code for group and channel conversations.
This PR adds a check for 1:1 in the latter code (channels + groups) and only adds the green banner if the conversation type is not 1:1.
With the assumption that the banner is added twice because somehow both parts of the code were executed, the new check should prevent the second banner from being added.

### Testing

We couldn't figure out how to reliably reproduce the duplicate banners, but this code has fixed the issue locally in one of the 1:1 conversations for @jullianm 

---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCellDescription.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCellDescription.swift
@@ -174,13 +174,15 @@ enum ConversationSystemMessageCellDescription {
         case .newConversation:
             var cells: [AnyConversationMessageCellDescription] = []
 
-            let welcomeCell = ConversationWelcomeSystemMessageCellDescription(
-                variant: (
-                    wireCells: conversation.isWireDriveEnabled,
-                    isChannel: conversation.isChannel
+            if conversation.conversationType != .oneOnOne {
+                let welcomeCell = ConversationWelcomeSystemMessageCellDescription(
+                    variant: (
+                        wireCells: conversation.isWireDriveEnabled,
+                        isChannel: conversation.isChannel
+                    )
                 )
-            )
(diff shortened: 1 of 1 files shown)


jullianm (approved):
LGTM, I confirm this fix remove the duplicated banners on some of my impacted 1:1s

netbe (commented):
@WilhelmOks I think the intention of the ticket was to prevent migrating the system messages from proteus conversations when migrating to mls (look at OneOnOneSource.swift:134):

All 1:1 have a proteus conversation and since mls is enabled a mls one. At creation of conversations the system messages are inserted.

When we resolve 1:1 conversations, we try to figure out which 1:1 conversation to display if mls is enabled and can be supported on both sides we migrate all previous messages of the proteus conversation to the final mls conversation which is displayed.

WilhelmOks:
> @WilhelmOks I think the intention of the ticket was to prevent migrating the system messages from proteus conversations when migrating to mls (look at OneOnOneSource.swift:134):
> 
> All 1:1 have a proteus conversation and since mls is enabled a mls one. At creation of conversations the system messages are inserted.
> 
> When we resolve 1:1 conversations, we try to figure out which 1:1 conversation to display if mls is enabled and can be supported on both sides we migrate all previous messages of the proteus conversation to the final mls conversation which is displayed.

@netbe thanks for the insight. We discussed it in our team and consulted with Marie.
If I understand correctly, it means that the fix in this PR is not a complete fix because it only removed duplicate green banners but other duplicate system messages remain.
Also this issue seems to be an old one and not caused by the implementation of the green banner.
I got assigned this ticket because we assumed that it was an issue with the green banner only or caused by the green banner.
Now we will reassign it to Core and I will close this PR. Sorry for the confusion.
Please contact me if you have questions about the implementation of the green banner or need any assistance.

PR #4271, opened 2026-02-05
chore: add logging to help investigate hang - WPB-22986

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22986" title="WPB-22986" target="_blank">BugWPB-22986</a>  [iOS] WireSyncEngine: ZMUserSession.setUpSyncAgent(clientID:)
  </td></table>
  <br />
 

### Issue

One of our common hangs is one which takes place in `ZMUserSession.setUpSyncAgent(clientID:)` due to a call to `NSManagedObjectContext.performAndWait(_:)` - see ticket for more details.

I'm unable to reproduce any issue here so with this PR I'm just adding some logging. My hope is to catch incidences of a hang in Beta so that I can investigate better what is happening via beta logs.

In this PR I also made some changes to `WireLogger.measureTime` to allow for working with non async code.

---

### Checklist

- [x] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [x] Description is filled and free of optional paragraphs.
- [ ] Adds/updates automated tests.

PR #4287, opened 2026-02-09
fix: duration logging attribute name - WPB-22986

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22986" title="WPB-22986" target="_blank">BugWPB-22986</a>  [iOS] WireSyncEngine: ZMUserSession.setUpSyncAgent(clientID:)
  </td></table>
  <br />
 

### Issue

In https://github.com/wireapp/wire-ios/pull/4271 I added some logging of the duration it takes to make client related sync strategies. This is to help me debug a hang.

I thought this would log a `duration` but instead it logged a `sync_duration` in DataDog. This is because we have a mapper on DD that renames `duration` to `sync_duration` which is not what I want for the property I'm logging.

In this PR:

- I rename  `LogAttributesKey.duration` to `LogAttributesKey.syncDuration` - I chose not to change it's underlying string to avoid breaking anything in our DD pipelines.
- Update our measureTime methods to allow passing in the `LogAttributesKey` used for duration.
- Introduce a new `LogAttributesKey` - `timeInterval` - which is intended to be a generic time interval key
- Use `timeInterval` when logging "make client strategies".

#### Todo

- [x] Add a re-mapper on DD to ensure that `time_interval` is logged as a Double.

### Testing

Check that the "completed make client strategies" has a `time_interval` attribute .

---

### Checklist


netbe (approved):
Looks good, just curious how you filter it in datadog

caldrian (approved):

PR #4301, opened 2026-02-11
chore: remove session manager completion handlers - WPB-23401

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23401" title="WPB-23401" target="_blank">TaskWPB-23401</a>  [iOS] Use async / await in SessionManager
  </td></table>
  <br />

### Issue

This PRs converts a number of UserSession methods with completion handlers to async methods. Necessarily this requires small changes in behavior that are hard to understand due to for example the addition of `Task` in various places which may slightly change the order of execution. I've tried to look at where these have been added and think it is safe (often it is in a kind of fire and forget context) but these changes should be well tested.

### Testing

These are not specific test steps but more scenarios

- Login multiple accounts on a couple of different backends
- Switch between different accounts.
- Logging out
- Cancel logging out of one account when logged in with multiple
- Switch backend from the conversation list & login.
- Using simulator, log in to multiple accounts, then simulate a memory warning in the simulator menu, then switch accounts.
- Have multiple logins, Enable 'createLegacyBackups', create a backup, logout, rel-ogin, restore, switch accounts a few times.
- Login with production + staging, select staging, go Dev tools > Overrides, enter `staging` for the obsolete backend, app restarts, error is shown - select switch accounts to go to production.
- Answer a a call to a non active account

Legacy authentication:
- Cancel logging in a second account using legacy authentication (x button)

---

### Checklist


wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -688,70 +688,62 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+    @MainActor
+    public func select(_ account: Account) async -> ZMUserSession? {
         guard !isSelectingAccount else {
-            completion?(nil)
-            return
+            return nil
         }
 
-        confirmSwitchingAccount { [weak self] isConfirmed in
+        let isConfirmed = await confirmSwitchingAccount()
```
johnxnguyen, line 700: thought: not really part of the scope here but I'm curious to see what you think. Calling this method will ask a ui delegate to present an alert and request user confirmation. Do you think this kind of back and forth communication makes sense? Or would it be better to have a single direction: ui tries to select an account, if confirmation is needed, throw error, ui handles, gets confirmation, tries again passing in a confirmation flag. Something liKe:

```swift
public func select(_ account: Account, authoritative: Bool) async throws -> ZMUserSession? {
  if confirmationNeeded && !authoritative  {
    throw Error.confirmationNeeded
  }

  // continue...
}
```

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -688,70 +688,62 @@ public final class SessionManager: NSObject, SessionManagerType {
...
 
-            guard let delegate = self?.delegate else {
-                completion?(nil)
-                return
-            }
+        guard let delegate = delegate else {
+            return nil
+        }
+
+        return await withCheckedContinuation { continuation in
```
johnxnguyen: suggestion: if you follow the path `delegate.sessionManagerWillOpenAccount` you'll see that it does nothing but immediately invoke the `userSessionCanBeTornDown` completion handler. I think you can simplify this by removing the delegate method as well as the `AppState.loading` case.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -778,9 +770,10 @@ public final class SessionManager: NSObject, SessionManagerType {
         WireLogger.sessionManager.debug("Deleting account \(account.userIdentifier)...")
         if let secondAccount = accountManager.inactiveAccounts.first {
             // Deleted an account but we can switch to another account
-            select(secondAccount, tearDownCompletion: { [weak self] in
-                self?.tearDownSessionAndDelete(account: account, eraseData: eraseData)
-            })
+            Task {
```
johnxnguyen, line 755: question: why not make `delete(account:reason:eraseData)` async?

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -688,70 +688,62 @@ public final class SessionManager: NSObject, SessionManagerType {
...
-            let error = NSError(userSessionErrorCode: .addAccountRequested, userInfo: userInfo)
-            self?.delegate?.sessionManagerWillLogout(
+    @MainActor
+    public func addAccount(userInfo: [String: Any]? = nil) async {
+        let isConfirmed = await confirmSwitchingAccount()
+        guard isConfirmed else { return }
+
+        let error = NSError(userSessionErrorCode: .addAccountRequested, userInfo: userInfo)
+        await withCheckedContinuation { continuation in
+            delegate?.sessionManagerWillLogout(
```
netbe: question: is it a plan to go a bit further and make this async in AppStateCalculator?

netbe: issue: if the delegate is nil, I guess you would wait forever here and crash, no?
also to check within the AppStateCalculator completion is not called if AppStateCalculatorDelegate is nil

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -688,70 +688,62 @@ public final class SessionManager: NSObject, SessionManagerType {
...
-            }
+        guard let delegate = delegate else {
+            return nil
+        }
+
+        return await withCheckedContinuation { continuation in
             delegate.sessionManagerWillOpenAccount(
                 account,
                 from: selectedAccount,
                 userSessionCanBeTornDown: { [weak self] in
```
netbe: todo: double check this is always executed otherwise continuation won't be called

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1274,23 +1267,24 @@ public final class SessionManager: NSObject, SessionManagerType {
...
         }
         tearDownObservers(account: accountId)
         backgroundUserSessions[accountId] = nil
 
         dispatchGroup.enter()
-        userSession.close(deleteCookie: false) { [weak self] in
-            self?.notifyUserSessionDestroyed(accountId)
-            completion?()
-            self?.dispatchGroup.leave()
+        await withCheckedContinuation { continuation in
```
netbe, line 1270: thought: so related to this, when userSession.close is called. we  call teardown, it might happen that because of leak this never calls the continuation.  I think it's still fine, as it won't crash just hang

netbe (commented):
left a couple comments before approving

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
samwyndham: Implemented in https://github.com/wireapp/wire-ios/pull/4301/commits/551497b5f7221b8afd7b077c6e4c8bc4ffb64088. Note that this also required the AppState Equatable implementation to be updated so that an authenticated state is not the same if it's user sessions are different which I think makes sense. Any concern there? This change was necessary so that changing from one authenticated state to another still did something.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
samwyndham: @johnxnguyen That method is called from so many non async contexts. It feels kind of risky as some of it's code paths don't need to be async. Overall I'm afraid of making many changes here:

Screenshot 2026-03-25 at 12 47 52

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
samwyndham: Yeah for me ideally it would be like you suggest. Perhaps not even part of this method but something which is called before. I think making that change might be difficult though as there are so many ways this might be triggered.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
samwyndham: > issue: if the delegate is nil, I guess you would wait forever here and crash, no?
also to check within the AppStateCalculator completion is not called if AppStateCalculatorDelegate is nil

Good catch. Fixed in 8ddeb5361b5d12524e0c54d16626aa6600b815c7

> question: is it a plan to go a bit further and make this async in AppStateCalculator?

I think the long term goal is to remove the need for them: https://wearezeta.atlassian.net/browse/WPB-23398. However I haven't put much thought into this yet.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
samwyndham: This code was removed in 551497b5f7221b8afd7b077c6e4c8bc4ffb64088

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnxnguyen: it makes sense

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
samwyndham: it looks like in `ZMUserSession.close(deleteCookie:completion:)` the call to tear down is non-blocking (it's using the asynchronous ManagedObjectContext method) so it does seem that the closure will be called even if tear down doesn't happen. Either way it would be nice to fix that leak.

johnxnguyen (approved):

netbe (approved):
LGTM:)

PR #4304, opened 2026-02-11
fix: triggering MLS resets for a federated conversation will establish a broken group - WPB-22510

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22510" title="WPB-22510" target="_blank">BugWPB-22510</a>  [iOS] Triggering MLS resets for a federated conversation will establish a broken group
  </td></table>
  <br />
 

### Issue

Following https://github.com/wireapp/wire-ios/pull/4048, we need to make sure that we only establish groups for local domain conversations.

* Make localDomain non optional for mlsService
* check localDomain and conversation domain before calling establishGroup.

### Testing

1. connect to a federated user
2. make sure it join the conversation via external commit

---

### Checklist

---


wire-ios-data-model/Tests/MLS/MLSServiceTests.swift
```diff
@@ -1006,6 +1006,41 @@ final class MLSServiceTests: ZMConversationTestsBase, MLSServiceDelegate {
...
+        let conversation = await uiMOC.perform { [uiMOC] in
+            let conversation = ZMConversation.insertNewObject(in: uiMOC)
+            conversation.remoteIdentifier = conversationID
+
+            conversation.mlsGroupID = groupID
+            conversation.messageProtocol = .mls
+            conversation.mlsStatus = .pendingJoin
+            conversation.conversationType = .group
+
+            conversation.epoch = 0
```
johnxnguyen, line 1023: question: if the epoch is 0 then wouldn't it establish the group? If so, then the assertion that it joined is confusing... I would expect it all to fail.

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
```diff
@@ -395,7 +395,8 @@ public final class MessageSender: MessageSenderInterface {
         }
 
         do {
-            if mlsStatus?.isOne(of: .pendingJoinAfterReset, .pendingJoin) == true {
+            if mlsStatus?.isOne(of: .pendingJoinAfterReset, .pendingJoin) == true,
+               conversationID.domain == mlsService.localDomain {
```
johnxnguyen, line 399: question: I wonder if instead of checking this here, we check it inside the `reEstablishPendingGroup` method?

wire-ios-data-model/Tests/MLS/MLSServiceTests.swift
netbe: no because here it's the case of federation, so the localDomain is different from the conversation domain

wire-ios-request-strategy/Sources/Message Sending/MessageSender.swift
netbe: well good question, as the first part of the check if outside, I would keep checks outside for now.

johnxnguyen (approved):

David-Henner (approved):
LGTM :)

David-Henner:
@netbe did you look at the new issues from sonarqube ?

PR #4307, opened 2026-02-12
chore: setup callingservice and test multiple group call - WPB-23001

findms:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23001" title="WPB-23001" target="_blank">TaskWPB-23001</a>  [iOS/QA] Setup callingServiceClient for calling tests
  </td></table>
  <br />
 

### Issue

**Changes in this PR:**

- Setup CallingServiceClient to createinstances, startcall, destroyinstance
- NewPage backoffice to unlock and enable services
- Test - Owner initiate a group call with 5 member total where 4 join the call via calling service and 1 via app
- Verify that all members were able to join the call

this is how the test looks:

https://github.com/user-attachments/assets/a728979c-8aa7-4db3-a659-c5bbe048decb


wire-ios/WireUITests/Helper/CallingServiceClient.swift
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    init() {
+        do {
+            let envVariables = try EnvironmentVariables()
+            self.callingServiceURL = envVariables.callingServiceURL
+            self.callingServiceUsername = envVariables.callingServiceUsername
+            self.callingServicePassword = envVariables.callingServicePassword
+        } catch {
+            preconditionFailure("CallingServiceClient failed to fetch EnvVariables: \(error)")
```
netbe: suggestion: throw instead so testcase can fail gracefully

wire-ios/WireUITests/Helper/CallingServiceClient.swift
```diff
@@ -0,0 +1,418 @@
...
+        }
+    }
+
+    enum Constants {
+        static let CONNECT_TIMEOUT: TimeInterval = 600_000
+        static let RESPONSE_TIMEOUT: TimeInterval = 600_000
+        static let CALLING_RESPONSE_TIMEOUT: TimeInterval = 3_600_000
```
netbe: question: are these seconds!!?

wire-ios/WireUITests/Helper/CallingServiceClient.swift
```diff
@@ -0,0 +1,418 @@
...
+        ]
+
+        if let auth = basicAuthHeader() {
+            headers[HttpClient.HeaderKey.authorization] = auth
+        }
+
+        let jsonBody: Data = if let body {
+            try JSONEncoder().encode(body)
+        } else {
+            Data()
```
netbe, line 115: question: is it expected to send random data? why is body optional?

wire-ios/WireUITests/Helper/CallingServiceClient.swift
```diff
@@ -0,0 +1,418 @@
...
+
+        let (data, code) = try await sendHttpRequest(endpoint: endpoint, body: body, method: .post)
+
+        guard code.statusCode == 200 else {
+            throw RuntimeError(
+                "CallingService failed to createInstance: HTTP \(code.statusCode). \(String(data: data, encoding: .utf8) ?? "")"
+            )
+        }
+        let instance = try JSONDecoder().decode(CallingServiceInstance.self, from: data)
+        await createdInstances.add(instance.id)
```
netbe, line 170: thought: maybe keeping track of instances and creating instances should be separate

wire-ios/WireUITests/Helper/CallingServiceClient.swift
```diff
@@ -0,0 +1,418 @@
...
+        let (data, code) = try await sendHttpRequest(endpoint: endpoint, body: CallingServiceEmptyBody(), method: .put)
+        guard code.statusCode == 200 else {
+            throw RuntimeError(
+                "CallingService call failed: PUT \(path) HTTP \(code.statusCode). \(String(data: data, encoding: .utf8) ?? "")"
+            )
+        }
+        return try JSONDecoder().decode(T.self, from: data)
+    }
+
+    func start(instanceId: String, request: StartCallBody) async throws -> CallReponse {
```
netbe: typo: 
```suggestion
    func start(instanceId: String, request: StartCallBody) async throws -> CallResponse {
```

wire-ios/WireUITests/Helper/CallingServiceClient.swift
```diff
@@ -0,0 +1,418 @@
...
+        )
+        guard code.statusCode == 200 else {
+            throw RuntimeError(
+                "CallingService call failed: GET \(path) HTTP \(code.statusCode). \(String(data: data, encoding: .utf8) ?? "")"
+            )
+        }
+        return try JSONDecoder().decode(T.self, from: data)
+    }
+
+    private func performCallPut<T: Decodable>(
```
netbe: question: do we always pass CallResponse ? if so we can get rid of the generic

wire-ios/WireUITests/Helper/CallingServiceClient.swift
```diff
@@ -0,0 +1,418 @@
...
+struct StartCallBody: Encodable {
+    let conversationId: String
+    let timeout: Double
+}
+
+typealias CallRequest = StartCallBody
+
+struct CallingServiceEmptyBody: Encodable {}
+
+struct CallReponse: Decodable {
```
netbe:
```suggestion
struct CallResponse: Decodable {
```

wire-ios/WireUITests/Helper/CallingServiceClient.swift
```diff
@@ -0,0 +1,418 @@
...
+}
+
+struct InstanceType: Encodable {
+    let name: String
+    let version: String
+}
+
+struct CallingServiceInstance: Decodable {
+    let id: String?
+    let instanceStatus: String?
```
netbe: question: is having nil for both still make sense ? I would assume at least the id to be there

wire-ios/WireUITests/Helper/CallingTestDefaults.swift
```diff
@@ -0,0 +1,25 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+enum CallingTestDefaults {
+    static let backend = "DEV" // temp on DEV as staging is down
+    static let isBeta = true
+
+    static let instanceTypeName = "chrome"
+    static let instanceTypeVersion = "103.0.5060.53"
```
netbe: question: should be moved to environment variable?

wire-ios/WireUITests/Helper/HttpClient.swift
```diff
@@ -0,0 +1,75 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+final class HttpClient {
```
netbe, line 21: question: don't we have a client that we can reuse in WireNetwork instead?

wire-ios/WireUITests/Helper/UserHelper.swift
```diff
@@ -246,6 +254,46 @@ class UserHelper {
...
+            headers: [
+                HttpClient.HeaderKey.accept: HttpClient.ContentType.json,
+                HttpClient.HeaderKey.contentType: HttpClient.ContentType.jsonUtf8,
+                HttpClient.HeaderKey.authorization: "Bearer \(accessToken.token)"
+            ]
+        )
+
+        if response.statusCode == 200 {
+            return
+        }
```
netbe: todo: to remove there is no handling of the response

wire-ios/WireUITests/Helper/UserHelper.swift
```diff
@@ -377,6 +465,80 @@ class UserHelper {
         return [teamMember1.name, teamMember2.name]
     }
 
+    /// Registers a team with a given number of members and add them to a group
+    /// - Parameters:
+    ///   - memberCount: count of members
+    ///   - groupName : optional groupName
+    /// - Returns: teamOwner info, teamMembers info, qualifiedIds of members, conversationId if group created
+    func registerTeamWithXMembersAndOptionalGroupConversation(
```
netbe:
```suggestion
    func registerTeam(
```

wire-ios/WireUITests/Helper/UserHelper.swift
```diff
@@ -377,6 +465,80 @@ class UserHelper {
...
+        var conversationId: UUID?
+        if let groupName {
+            try await createGroupConversations(
+                qualifiedIds: qualifiedIDs,
+                owner: teamOwner,
+                groupName: groupName
+            )
+
+            let (resolvedConversationId, _) = try await getConversationId(matching: .groupName(groupName))
+            guard let resolvedConversationId else {
```
netbe: question: is it possible that you get nil for resolvedConversationId and getConversationId does not throw?

wire-ios/WireUITests/Pages/OngoingCallPage.swift
```diff
@@ -0,0 +1,37 @@
...
+        endCallButton
+    }
+
+    var endCallButton: XCUIElement {
+        app.buttons[Locators.OngoingCallPage.endOngoingCallButton.rawValue]
+    }
+
+    func endOngoingCall() throws -> ConversationsPage {
+        endCallButton.tap()
+        return try ConversationsPage()
```
netbe, line 34: question: does ending a call always go back to the conversation List? is there not a case it returns to Conversation content if it started there?

wire-ios/WireUITests/CallingTests.swift
```diff
@@ -0,0 +1,132 @@
...
+            users: users,
+            backend: CallingTestDefaults.backend,
+            beta: CallingTestDefaults.isBeta,
+            instanceTypeName: CallingTestDefaults.instanceTypeName,
+            instanceTypeVersion: CallingTestDefaults.instanceTypeVersion
+        )
+    }
+
+    private func requireOwnerInstanceId(from instances: [CallingServiceInstance]) throws -> String {
+        let ownerId = try XCTUnwrap(instances.first?.id, "Owner instanceId is nil")
```
netbe, line 80: question: is the owner always the first instance?

wire-ios/WireUITests/CallingTests.swift
```diff
@@ -0,0 +1,132 @@
...
+
+        let ongoingCallPage = try incomingCallPage.acceptIncommingCall()
+        XCTAssertTrue(app.staticTexts[groupName].waitForExistence(timeout: 10), "Conversation title mismatch")
+
+        return ongoingCallPage
+    }
+
+    private func verifyParticipantsVisible(_ participants: [UserInfo]) {
+        for user in participants {
+            let participantIdentifier = "audioView.\(user.name).minimized.inactive"
```
netbe: suggestion: use locator with a constant with associatedValue. i.e: Locators.callView.participantIdentifier(userName)

netbe (commented):
Great to see this coming to life (video), left some comments before approving

wire-ios/WireUITests/Helper/CallingTestDefaults.swift
findms: for now it's okay to keep as it is.Later maybe we can fetch as an request and get the response based on that we can use these value not hardcoding.

wire-ios/WireUITests/Helper/CallingServiceClient.swift
findms: true. id is never nil so we can make it non optional.

wire-ios/WireUITests/Helper/CallingServiceClient.swift
findms: updated

wire-ios/WireUITests/Helper/CallingServiceClient.swift
findms: Yeah..from what I see from now CallResponse is always returning. Updated with generic to CallResponse.

wire-ios/WireUITests/Helper/CallingServiceClient.swift
findms: updated

wire-ios/WireUITests/Helper/CallingServiceClient.swift
findms: yes ..as GET instance request has no body.

wire-ios/WireUITests/Helper/CallingServiceClient.swift
findms: Was milliseconds in double but now changed as seconds.

wire-ios/WireUITests/Helper/CallingServiceClient.swift
findms: make sense..updated

wire-ios/WireUITests/Helper/UserHelper.swift
findms: corrected with guard.

wire-ios/WireUITests/Helper/UserHelper.swift
findms: no..group will be created there will be  conversationId ..so no nil if group is being created.

wire-ios/WireUITests/Pages/OngoingCallPage.swift
findms: for this test it will always land on conversation list..for other cases we will be handling that case based on the test need.

wire-ios/WireUITests/Helper/UserHelper.swift
findms: updated

wire-ios/WireUITests/CallingTests.swift
findms: updated

wire-ios/WireUITests/CallingTests.swift
findms: yes. Owner will always be the 1st instance as users are being processed sequentially unless any race condition is found.

wire-ios/WireUITests/Helper/HttpClient.swift
netbe: I couldn't find a good match, I think all our clients HTTPClientImpl or NEtworkService are too specific, so good to use a specific one here

wire-ios/WireUITests/Pages/OngoingCallPage.swift
netbe: I would suggest to assert the onGoingCallPage is dismissed then instead of checking here the ConversationList is present

wire-ios/WireUITests/Helper/UserHelper.swift
netbe: in that case it's better not to make resolvedConversationId optional

wire-ios/WireUITests/Helper/WireUITestCase.swift
```diff
@@ -26,7 +26,11 @@ class WireUITestCase: XCTestCase {
     var app: XCUIApplication!
     let springboard = XCUIApplication(bundleIdentifier: "com.apple.springboard")
     let userHelper = UserHelper()
-    let callingServiceClient = CallingServiceClient()
+    lazy var callingServiceClient: CallingServiceClient = {
+        do { return try CallingServiceClient() } catch {
+            fatalError("CallingServiceClient initialization failed: \(error)")
+        }
+    }()
```
netbe: suggestion: we usually do this in setup
```
var callingServiceClient: CallingServiceClient!
```

and in the setup method:

```
 override func setUpWithError() throws {
   callingServiceClient = try CallingServiceClient()
```

wire-ios-build-assets
netbe, line 1: todo: this is not intended should not be changed

netbe (changes requested):
submodule should not be included here, the rest looks good

wire-ios-build-assets
findms: ah..yeah switching branches this sneak it..will remove this.

findms: fixed

wire-ios/WireUITests/Pages/OngoingCallPage.swift
findms: Added another assertion where conversationList is being checked after ending the call.

wire-ios/WireUITests/Helper/UserHelper.swift
findms: since we are calling getConversationId, it will always return with id so not an optional..updated.

wire-ios/WireUITests/Helper/WireUITestCase.swift
findms: updated

KaterinaWire (approved):

wire-ios/Tests/TestPlans/UITests.xctestplan
```diff
@@ -9,6 +9,30 @@
...
+            "key" : "CALLING_BACKEND",
+            "value" : "DEV"
+          },
+          {
+            "key" : "CALLING_INSTANCE_TYPE_NAME",
+            "value" : "chrome"
+          },
+          {
+            "key" : "CALLING_INSTANCE_TYPE_VERSION",
+            "value" : "103.0.5060.53"
```
netbe: todo: it needs to be read from the env.xcconfig same as CALLINGSERVICE_URL

findms: moved it to 1password

netbe (approved):
LGTM';)

PR #4300, opened 2026-02-11
fix: improve error logging for notification generation - WPB-23407

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23407" title="WPB-23407" target="_blank">TaskWPB-23407</a>  [iOS] Investigate 7.48K logs for "Failed generating notification: failedToDecryptMLSMessage"
  </td></table>
  <br />
 

### Issue

The highest logged error at the moment is `Failed generating notification: failedToDecryptMLSMessage`.

Screenshot 2026-02-11 at 10 11 50

This error message is misleading because notifications are generated from already decrypted events. Upon investigation, I found that there are several things possible errors that are all bundled into a single generic error. One possible explanation of the high number of logs is that we expect there to always be a decrypted mls message for processing, which is not the case when the update event contained only mls protocol messages (not visible to the user).

This PR introduces more granular error reporting as well as not logging expected "failures" so that we can understand better if there is actually a problem to address or not.

### Testing
Monitor DD dashboard for the error in subsequent versions.

---

### Checklist

- [x] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [x] Description is filled and free of optional paragraphs.
- [ ] Adds/updates automated tests.

---

### UI accessibility checklist

_If your PR includes UI changes, please utilize this checklist:_
- [ ] Make sure you use the API for UI elements that support large fonts.
- [ ] All colors are taken from WireDesign.ColorTheme or constructed using WireDesign.BaseColorPalette.
- [ ] New UI elements have Accessibility strings for VoiceOver.

PR #4310, opened 2026-02-13
feat: support EAR with incremental sync - WPB-17302

David-Henner:
### Issue

This PR add support for EAR with the incremental sync. This means we encrypt the update events when saving them in the events database, and decrypt them when fetching them for processing.

Most of the events can be accessible only when the app is unlocked. However, calling events need to be processed while in the background, so their access level is less restrictive and can be encrypted / decrypted without the app being unlocked. This is why we have primary and secondary encryption keys. 

### Changes

**IncrementalSync**
  - Added `earService` and `notificationContext` dependencies
  - ~New: `waitForDatabaseUnlock()` method with 3-second timeout~
  - Key Flow Changes:
    - If we're in foreground and database is locked, abort sync
    - Fetch public / private keys before syncing
    - Pass public keys when pulling events
    - Pass private keys when processing stored events
    - Pass `backgroundAccessibleOnly` to live & stored event processing 

**UpdateEventsLocalStore**
  - Encryption on Write (`persistEventEnvelope`):
    - Accepts optional `publicKeys` parameter
    - Encrypts event data with primary or secondary key based on event type
    - Sets `isEncrypted` and `isBackgroundAccessible` properties
  - Decryption on Read (`fetchStoredEventEnvelopes`):
    - Accepts optional `privateKeys` and `backgroundAccessibleOnly` parameters
    - Filters events based on background accessibility (for processing of calling events in NSE)
    - Decrypts encrypted events with appropriate private key - primary vs secondary

  **Core Data**

  `StoredUpdateEventEnvelope` entity now includes:
  - `isEncrypted`: indicates if event data is encrypted
  - `isBackgroundAccessible`: indicates if the event should be accessible in the background (for calling events)

  **New Files Added**

  1. `EAREncryptionHelper.swift` - Encryption/decryption utilities for event data
  2. `ProtobufMessageDecoder.swift` - Helper for decoding protobuf messages 
  3. `UpdateEventEnvelope+BackgroundAccessible.swift` - Extension determining if events are calling-related
  4. `DatabaseEncryptionLockNotification.swift` - Notification type for database lock/unlock events (moved from sync-engine)

### Testing

- Log in 
- Go to `Settings > Account > Encrypt messages at rest` - enable
- Background the app
- Receive messages or other events
- Open the app
- Verify the messages / events are processed 

---

### Checklist


WireDomain/Sources/WireDomain/Notifications/Components/NSEUserScope.swift
```diff
@@ -155,6 +155,13 @@ final class NSEUserScope: Component<NSEUserScopeDependency> {
             throw Failure.mainAppRequired(message: "no self client id")
         }
 
+        let earService = await EARService(
+            accountID: accountID,
+            coreDataStack: coreDataStack,
+            sharedUserDefaults: dependency.sharedUserDefaults,
+            authenticationContext: AuthenticationContext(storage: LAContextStorage())
```
netbe, line 162: question: isn't the LAContextStorage() requiring a biometric prompt?

WireDomain/Sources/WireDomain/Repositories/UpdateEvents/UpdateEventsLocalStore.swift
```diff
@@ -33,6 +33,7 @@ final class UpdateEventsLocalStore: UpdateEventsLocalStoreProtocol {
     enum Error: Swift.Error {
         case failedToFetchStoredEvents(Swift.Error)
         case failedToDeleteStoredEvents(Swift.Error)
+        case failedToEncryptEventData
```
netbe: suggestion: add associated error?

WireDomain/Sources/WireDomain/Repositories/UpdateEvents/UpdateEventsLocalStore.swift
```diff
@@ -93,43 +94,96 @@ final class UpdateEventsLocalStore: UpdateEventsLocalStoreProtocol {
...
+                        }
+
+                        let isBackgroundAccessible = storedEnvelope.isBackgroundAccessible
+                        let key = isBackgroundAccessible ? privateKeys.secondary : privateKeys.primary
+
+                        guard let key, let decryptedData = EAREncryptionHelper.decrypt(
+                            data: data,
+                            privateKey: key
+                        ) else {
+                            WireLogger.ear.error("failed to decrypt stored event", attributes: .safePublic)
```
netbe: question: can we log the eventId as envelope_event_id attribute?

WireDomain/Sources/WireDomain/Repositories/UpdateEvents/UpdateEventsLocalStore.swift
```diff
@@ -220,4 +274,43 @@ final class UpdateEventsLocalStore: UpdateEventsLocalStoreProtocol {
...
+
+            if let encryptedData = EAREncryptionHelper.encrypt(
+                data: data,
+                publicKey: key
+            ) {
+                data = encryptedData
+                storedEventEnvelope.isEncrypted = true
+                storedEventEnvelope.isBackgroundAccessible = isBackgroundAccessible
+            } else {
+                WireLogger.ear.error("failed to encrypt event", attributes: .safePublic)
```
netbe: suggestion: add event id in attributes if possible

WireDomain/Sources/WireDomain/Repositories/UpdateEvents/UpdateEventEnvelope+BackgroundAccessible.swift
```diff
@@ -0,0 +1,75 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireNetwork
+
+extension UpdateEventEnvelope {
+    var isBackgroundAccessible: Bool {
+        !events.isEmpty && events.allSatisfy(\.hasCallingContent)
```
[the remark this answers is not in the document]
David-Henner: this is intentionally restrictive to avoid setting the wrong access for any event that isn't meant to be accessed in the background.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -351,16 +386,82 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
+            // Task 2: Timeout
+            group.addTask { [earService, logger] in
+                try await Task.sleep(for: timeout)
+
+                if earService.isLocked {
+                    logger.error(
+                        "timed out waiting for database unlock, cancelling sync",
+                        attributes: .incrementalSyncV2, .safePublic
+                    )
+                    throw Failure.databaseUnlockTimeout
```
netbe: question: is the failure propagated?
does it mean that if it timesout? the incremental sync is cancelled?
why would it timeout?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -192,7 +195,12 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
             // If we need to abort, do it before processing the next batch.
             try Task.checkCancellation()
 
-            let envelopesWithObjectIDs = try await updateEventsStore.fetchStoredEventEnvelopes(limit: batchSize)
+            // TODO: [WPB-23558] Support EAR in incremental sync v2
```
netbe, line 198: praise: thanks for adding todos

WireDomain/Sources/WireDomain/Notifications/Components/NSEClientScope.swift
```diff
@@ -88,6 +91,7 @@ final class NSEClientScope: Component<NSEClientScopeDependency> {
     ) async throws {
         // Pull pending update events.
         let eventStream: AsyncStream<[UpdateEvent]>
+        let publicKeys = try earService.fetchPublicKeys()
```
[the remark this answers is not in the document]
David-Henner: no, the public keys are accessible even if the database is locked

netbe (commented):
left some comments, the main point I am a bit fuzzy about is when DatabaseEncryptionLockNotification is posted, and if we could use publisher instead of notification

WireDomain/Sources/WireDomain/Notifications/Components/NSEUserScope.swift
David-Henner: it's `LAContext()` that does IIRC

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
David-Henner: outdated [b04d292](https://github.com/wireapp/wire-ios/pull/4310/commits/b04d2922550a25568536a98b9722fdf360b9100b)

WireDomain/Sources/WireDomain/Helpers/ProtobufMessageDecoder.swift
```diff
@@ -0,0 +1,78 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import GenericMessageProtocol
+
+struct ProtobufMessageDecoder {
```
johnxnguyen, line 22: question: it looks like this was restored to its original implementation, but [changes were made recently](https://github.com/wireapp/wire-ios/pull/4300) to throw errors. I think we should preserve these changes.

WireDomain/Sources/WireDomain/Repositories/UpdateEvents/UpdateEventsLocalStore.swift
```diff
@@ -93,43 +94,96 @@ final class UpdateEventsLocalStore: UpdateEventsLocalStoreProtocol {
...
+                    if storedEnvelope.isEncrypted {
+
+                        WireLogger.ear.info("decrypting stored event.")
+
+                        guard let privateKeys else {
+                            WireLogger.ear.error(
+                                "failed to decrypt stored event: no private keys",
+                                attributes: .safePublic
+                            )
+                            return nil
```
johnxnguyen: issue: returning `nil` here means we just skip over this event and will proceed to process the rest. This could lead to some unexpected issues given the assumption that events are processed in order. What do you think about throwing an error and aborting? Perhaps this comes with other issues. what do you think @netbe ?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -69,13 +73,46 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
     }
 
     public func perform() async throws -> Token {
+        try await perform(
+            appState: await UIApplication.shared.applicationState
+        )
+    }
+
+    func perform(appState: UIApplication.State) async throws -> Token {
+        let inBackground = appState == .background
```
johnxnguyen: question: why do we need to check if we're in the BG?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -271,15 +295,69 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
     }
 
-    private func processStoredEvents() async throws -> Set<UUID> {
+    private func processLiveEventEnvelope(
+        envelope: UpdateEventEnvelope,
+        index: Int64,
+        publicKeys: EARPublicKeys?,
+        backgroundAccessibleOnly: Bool
+    ) async {
+        guard !backgroundAccessibleOnly || (backgroundAccessibleOnly && envelope.isBackgroundAccessible) else {
```
johnxnguyen: question: why do we abort if we get a live event that is not background accessible? To avoid processing it before we processed earlier stored events?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -271,15 +295,69 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
+            logger.info(
+                "skipping processing of live event envelope: not accessible in the background",
+                attributes: .safePublic,
+                .incrementalSyncV2 + [.eventEnvelopeID: envelope.id] // Can it be safePublic?
```
johnxnguyen: I think it can be safe public

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -357,10 +435,15 @@ extension IncrementalSyncV1: SyncMigratorProtocol {
...
         logger.debug("pulling pending update events", attributes: .incrementalSyncV2)
         syncStateSubject.send(.incrementalSyncing(.pullPendingEvents))
-        try await updateEventsSync.pull()
+        try await updateEventsSync.pull(publicKeys: try earService.fetchPublicKeys())
 
         logger.debug("processing stored update events", attributes: .incrementalSyncV2)
         syncStateSubject.send(.incrementalSyncing(.processPendingEvents))
-        _ = try await processStoredEvents()
+        let isLocked = earService.isLocked
+        let privateKeys = try earService.fetchPrivateKeys(includingPrimary: !isLocked)
```
johnxnguyen: question: so we will attempt to migrate event if the database is locked? I would expect that a prerequisite would be to have an unlocked database.

WireDomain/Tests/WireDomainTests/LocalStores/UpdateEventsLocalStoreTests.swift
johnxnguyen, line 1: question: which of these test should be including en the security tests plan?

David-Henner: discussed 1:1

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
David-Henner: fixed here [c192c71](https://github.com/wireapp/wire-ios/pull/4310/commits/c192c713bd3f2afb2e29c720d9dbb4a433213f62)

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
David-Henner: added comment [927a65c](https://github.com/wireapp/wire-ios/pull/4310/commits/927a65c9cabea19b351bc8e39f582fb174eb6bc8)

WireDomain/Sources/WireDomain/Repositories/UpdateEvents/UpdateEventsLocalStore.swift
David-Henner: added critical logs and threw errors [6a2e408](https://github.com/wireapp/wire-ios/pull/4310/commits/6a2e408a81e15b9a0906b6afb30280c7981418a2)

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
David-Henner: made an attempt to update the flow as discussed [65fd560](https://github.com/wireapp/wire-ios/pull/4310/commits/65fd560de72bdd14afa0de0d51f0038eb4ce3d5c)

WireDomain/Sources/WireDomain/Repositories/UpdateEvents/UpdateEventsLocalStore.swift
David-Henner: added [2ea7235](https://github.com/wireapp/wire-ios/pull/4310/commits/2ea7235c941e8c74fc085a17c1d72cc87a151982)

WireDomain/Sources/WireDomain/Repositories/UpdateEvents/UpdateEventsLocalStore.swift
David-Henner: not available on the stored event

WireDomain/Sources/WireDomain/Repositories/UpdateEvents/UpdateEventsLocalStore.swift
David-Henner: [6a2e408](https://github.com/wireapp/wire-ios/pull/4310/commits/6a2e408a81e15b9a0906b6afb30280c7981418a2)

wire-ios-sync-engine/Source/Calling/WireCallCenterV3+Notifications.swift
```diff
@@ -17,7 +17,23 @@
...
+
+    func post(in context: NotificationContext, object: AnyObject? = nil) {
+        NotificationInContext(
+            name: type(of: self).notificationName,
+            context: context,
+            object: object,
+            userInfo: [type(of: self).userInfoKey: self]
+        ).post()
+    }
+}
```
netbe, line 36: question: why moving this to this file? is it the only usage?

wire-ios-sync-engine/Source/SessionManager/SessionManager+CallKitManagerDelegate.swift
```diff
@@ -75,7 +75,7 @@ extension SessionManager: CallKitManagerDelegate {
             }
 
             await requestCallConfigIfNeeded(for: userSession)
-            await userSession.processPendingCallEvents()
+            await userSession.processPendingCallEvents(only: false)
```
netbe: question: when is this called exactly? I see it's called when CXEndCallAction and CXAnswerCallAction, is it guaranteed that the app is in foreground at that point?

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -109,8 +109,10 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
     /// otherwise the incremental sync will be performed.
     ///
     /// This method logs any errors and does not wait for the sync to finish.
+    ///
+    /// - Parameter forCallEventsOnly: if the sync should be resumed only for calling events
 
-    func resume() {
+    func resume(forCallEventsOnly: Bool = false) {
```
netbe: nitpic:
 func resume(callEventsOnly: Bool = false) {

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+UserSession.swift
```diff
@@ -114,7 +114,7 @@ extension ZMUserSession: UserSession {
 
     public func unlockDatabase() throws {
         try earService.unlockDatabase()
-
+        syncAgent?.resume()
         DatabaseEncryptionLockNotification(databaseIsEncrypted: false).post(in: notificationContext)
```
netbe, line 118: question: is the notification still used?

WireDomain/Sources/WireDomain/Helpers/EAREncryptionHelper.swift
```diff
@@ -19,29 +19,43 @@
...
+        var cfError: Unmanaged<CFError>?
+        guard let result = SecKeyCreateDecryptedData(
             privateKey,
             .eciesEncryptionCofactorX963SHA256AESGCM,
             data as CFData,
-            nil
-        ) as? Data
+            &cfError
+        ) as? Data else {
+            // cfError is guaranteed to be populated if result in nil
```
netbe, line 56: typo: 
cfError is guaranteed to be populated if result is nil

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -299,6 +285,56 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
+    private func processLiveEventEnvelope(
+        envelope: UpdateEventEnvelope,
+        index: Int64,
+        publicKeys: EARPublicKeys?
+    ) async {
+        // If we're processing events in the background, and the event is not accessible: skip it.
+        //
+        // If we don't skip it, we may end up processing events that require access to the database key (e.g: messages).
+        // But the database key is only accessible once the app is unlocked.
+        guard !earService.isLocked || envelope.isBackgroundAccessible else {
```
netbe, line 295: question: is this the right condition? it seems like it will skip events even when live with no EAR

netbe (commented):
still have some questions before approving

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
David-Henner: if live with no EAR `earService.isLocked` will return false, so it'll pass the guard

netbe: ok, sorry, i should have been more precise. it's in the case of EAR enabled that I don't understand

David-Henner: so, if `earService.isLocked` is `true` then we know the database key is `nil` thus we can't process events that require it. We nullify the database key from cache every time the app goes to the background.
if `envelope.isBackgroundAccessible` is `true`, it means we can use the secondary private key (accessible without user authentication, aka "the background") to decrypt it 

So the guard essentially verifies that either the database is unlocked, or the event is processable in the background.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+UserSession.swift
David-Henner: yes, it was used here originally and it's still being used https://github.com/wireapp/wire-ios/blob/4a3f82af869c710e6b9b4cc999127ee37c4b3616/wire-ios-sync-engine/Source/SessionManager/SessionManager.swift#L1192

wire-ios-sync-engine/Source/Calling/WireCallCenterV3+Notifications.swift
David-Henner: this is just something I reverted back to how it was on `develop`, if you look at all the changes it's not there. I suspect it's used only by the calling code

WireDomain/Sources/WireDomain/Helpers/EAREncryptionHelper.swift
David-Henner: fixed

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
David-Henner: done

wire-ios-sync-engine/Source/SessionManager/SessionManager+CallKitManagerDelegate.swift
David-Henner: thanks for bringing this up, I realized it can be either foreground or background. so we can't always pass `false`. I'll think about another approach

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -69,13 +73,39 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
+            )
+            throw Failure.databaseLocked
+        }
+
+        return try await internalPerform(backgroundAccessibleOnly: false)
+    }
+
+    public func performInBackgroundForCallingEvents() async throws -> Token {
+        // if EAR is enabled, we'll process only the calling events (they are marked as `backgroundAccessible`)
+        try await internalPerform(backgroundAccessibleOnly: earService.isEAREnabled)
```
johnxnguyen: thought: I think we still want to only process call events here, regardless of whether EAR is enabled, since it's the fastest path to establishing a call in the background.

WireDomain/Tests/WireDomainTests/Synchronization/IncrementalSyncTests.swift
```diff
@@ -407,6 +420,131 @@ final class IncrementalSyncTests: XCTestCase {
...
+        updateEventsStore.calculateLastUnreadMessages_MockMethod = {}
+        databaseSaver.save_MockMethod = {}
+        mlsGroupRepairAgent.repairConversations_MockMethod = {}
+
+        // When
+        _ = try await sut.perform()
+
+        // Then: Should fetch keys with includingPrimary: true
+        XCTAssertEqual(earService.fetchPrivateKeysIncludingPrimary_Invocations.count, 1)
+        XCTAssertTrue(earService.fetchPrivateKeysIncludingPrimary_Invocations[0])
```
johnxnguyen: suggestion: here and elsewhere, instead of asserting the count and continuing to access by index, which may cause an exception, you can assert the count which will throw if it's not correct:

```swift
try XCTAssertCount(earService.fetchPrivateKeysIncludingPrimary_Invocations, count: 1)
// This will never be out of bounds
XCTAssertTrue(earService.fetchPrivateKeysIncludingPrimary_Invocations[0])
```

johnxnguyen (commented):
Looking good

wire-ios-sync-engine/Source/SessionManager/SessionManager+CallKitManagerDelegate.swift
netbe: actually we always want to only processPendingCallEvents so it should be always true

David-Henner: we need to test this because I remember something being wrong when I systematically processed only call events

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
David-Henner: done [e4277eb](https://github.com/wireapp/wire-ios/pull/4310/commits/e4277eb72ae51903a90c768994c91ed5aee8c944)

wire-ios-sync-engine/Source/SessionManager/SessionManager+CallKitManagerDelegate.swift
David-Henner: [e4277eb](https://github.com/wireapp/wire-ios/pull/4310/commits/e4277eb72ae51903a90c768994c91ed5aee8c944)

WireDomain/Tests/WireDomainTests/Synchronization/IncrementalSyncTests.swift
David-Henner: done [eaf102c](https://github.com/wireapp/wire-ios/pull/4310/commits/eaf102cf794d6f6ecddf3fd3241770df8ff7548b)

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1317,13 +1318,13 @@ extension ZMUserSession: SyncAgentDelegate {
...
-    func processPendingCallEvents() async {
+    func processPendingCallEvents(only onlyCallEvents: Bool) async {
         WireLogger.sync.debug(
-            "process pending call events",
+            "process pending call events (onlyCallEvents: \(onlyCallEvents)",
             attributes: .incrementalSync
         )
 
-        syncAgent?.resume()
+        syncAgent?.resume(callEventsOnly: true)
```
netbe: question: so the bool is not used and can be removed, @David-Henner right?

David-Henner: ah yes, I missed this. I'll update

David-Henner: [9f975d3](https://github.com/wireapp/wire-ios/pull/4310/commits/9f975d3670a2da0e8b13735884be443d7aeba93f)

netbe (approved):
LGTM;)

johnxnguyen (approved):

PR #4315, opened 2026-02-17
fix: background fetch tasks accumulate & interfere with sync - WPB-23451

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23451" title="WPB-23451" target="_blank">BugWPB-23451</a>  [iOS] Background fetch tasks accumulate and run in foreground
  </td></table>
  <br />
 

### Issue
Background fetch was been softly disabled with the introduction of sync v2, but the app still wakes periodically and runs some code, including starting a Swift `Task` that suspends any ongoing sync. These tasks have been observed to accumulate, meaning they don't start until the app comes to the foreground and then all run at once. In the logs, this appears as many as 20 or 30 "suspending sync" messages.

Since background fetch shouldn't run (yet) with sync v2, the solution here is to simply remove background fetch support entirely. When we add it back, we should do so with the [BackgroundTasks framework](https://developer.apple.com/documentation/UIKit/using-background-tasks-to-update-your-app).

### Testing
I don't think there's any direct way we can test this, but we should monitor how the app behaves when left in the background for extended periods of time.

---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS/Wire-Info.plist
+++ b/wire-ios/Wire-iOS/Wire-Info.plist
@@ -135,7 +135,6 @@
 		<string>audio</string>
 		<string>voip</string>
 		<string>remote-notification</string>
-		<string>fetch</string>
 	</array>
 	<key>UIDesignRequiresCompatibility</key>
 	<true/>
--- a/wire-ios-sync-engine/Source/Synchronization/ApplicationStatusDirectory.swift
+++ b/wire-ios-sync-engine/Source/Synchronization/ApplicationStatusDirectory.swift
@@ -85,7 +85,7 @@ public final class ApplicationStatusDirectory: NSObject, ApplicationStatus {
         switch operationStatus.operationState {
         case .foreground:
             .foreground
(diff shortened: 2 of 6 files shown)


wire-ios/Wire-iOS/Wire-Info.plist
```diff
@@ -135,7 +135,6 @@
 		<string>audio</string>
 		<string>voip</string>
 		<string>remote-notification</string>
-		<string>fetch</string>
```
netbe, line 138: thought: I wonder if we can remove capability from 1 version to another

netbe (approved):
Looks good, I just wonder if there is a restriction removing capabilities on the appstore

wire-ios/Wire-iOS/Wire-Info.plist
johnxnguyen: I think it's ok, and actually better than having the capability but not using it.

samwyndham (approved):
LGTM

PR #4326, opened 2026-02-17
fix: Allow call join from the lock screen - WPB-21406

KaterinaWire:
<table>
<td>
[[iOS] Can't join call via CallKit from Lock Screen](https://wearezeta.atlassian.net/browse/WPB-21406)
  </td></table>
  <br />
 

### Issue
Users cannot join incoming MLS group calls via CallKit from the phone's lock screen, the call gets stuck in "Connecting" state.
`MLSRequestStrategy` only had `.allowsRequestsWhileOnline` configuration, which prevents it from processing requests when the app is in background.

### Solution
Add `.allowsRequestsWhileInBackground` to `MLSRequestStrategy` configuration to allow MLS operations (like fetching subgroups for conference calls) to execute when the app is in background state, such as when accepting an incoming group call with the phone locked.

### Testing
- move the app to the background
- lock the screen
- receive an incoming call
- join the call

### Checklist

---


netbe (approved):
LGTM:)

David-Henner (approved):

PR #4339, opened 2026-02-19
fix: "no connection" banner not hiding when internet back - WPB-23482

OlivellaO:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23482" title="WPB-23482" target="_blank">BugWPB-23482</a>  [iOS] When internet off and on No internet banner is still visible on global Drive and Shared Drive screen
  </td></table>
  <br />
 

### Issue

"No internet" bar was still showing on a specific use case when user wasn't on File Browser screen when the connection was lost.
The issue was that when internet was going out but list file was empty the network state wasn't getting updated. 

Now that logic was move out from the function in charge of checking for internet changes and updating the connection state so it will always update the state correctly.

UI was updated to show Offline bar base on param that combines both "states" (internet status && list items empty or not) now that internet status will always get updated.

### Testing

Steps to reproduce:
- Open conversation list
- Switch internet off
- Open global Drive
- Open conversation list
- Switch internet on
- Open global Drive

No internet bar should be hidden since internet is back on.

---

### Checklist

---

### DIFF
--- a/WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesViewModel.swift
+++ b/WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesViewModel.swift
@@ -254,6 +254,10 @@ package final class FilesViewModel: ObservableObject {
         loadMoreTask != nil
     }
 
+    var shouldShowOfflineBar: Bool {
+        connectionState == .offline && !state.items.isEmpty
+    }
+
     enum ConnectionState {
         case offline
         case online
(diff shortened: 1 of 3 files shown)


WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesViewModel.swift
```diff
@@ -254,6 +254,10 @@ package final class FilesViewModel: ObservableObject {
         loadMoreTask != nil
     }
 
+    var shouldShowOfflineBar: Bool {
+        connectionState == .offline && !state.items.isEmpty
```
WilhelmOks, line 258: Why don't we show the offline bar for when the list is empty?

WilhelmOks: ok, I see you just moved the logic from another place to here. That's fine then.

WilhelmOks (approved):

jullianm (approved):
Nice work. Thanks for fixing it!

PR #4340, opened 2026-02-20
chore: make `AccountManager` concurrency safe - WPB-23570

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23570" title="WPB-23570" target="_blank">TaskWPB-23570</a>  [iOS] Make `AccountManager` thread safe
  </td></table>
  <br />
 

### Issue

This PR makes `AccountManager`, `AccountStore` & `Account` `Sendable` and therefore thread safe. It is part of a larger effort to make Session management safer across concurrent contexts.

I intentionally chose to do this with the aid of locks instead of Actors or @MainActor annotation as:

- As these types are accessed from many places in nonisolated contexts, marking them as isolated would be a large effort and may lead to bugs.
- I don't believe there to be performance issues doing this - in fact performance is likely better using locks than using the `await` keyword.

The negative is there is some additional complexity added to these types themselves.

Note:

I tried to add some tests checking for threading issues but was unsuccessful. I could do it by enabling the thread sanitizer in the test plan and having tests fail on error but there were too many existing tests that would fail in such cases. 

### Testing

1. Logging in with multiple account on an older app
2. Migrating to the new app
3. There shouldn't be any issues - user should not be logged out.

---

### Checklist

---


wire-ios-data-model/Source/Model/Account/Account.swift
```diff
@@ -27,36 +28,86 @@ public extension Notification.Name {
 /// An `Account` holds information related to a single account,
 /// such as the accounts users name,
 /// team name if there is any, picture and uuid.
-public final class Account: NSObject, Codable {
+public final class Account: NSObject, Codable, @unchecked Sendable {
```
johnxnguyen, line 33: suggestion: add a comment explaining why `@unchecked Sendable` is ok here.

wire-ios-data-model/Source/Model/Account/Account.swift
```diff
@@ -70,31 +121,33 @@ public final class Account: NSObject, Codable {
...
     public func updateWith(_ account: Account) {
         guard userIdentifier == account.userIdentifier else { return }
-        userName = account.userName
-        teamName = account.teamName
-        imageData = account.imageData
-        teamImageData = account.teamImageData
-        loginCredentials = account.loginCredentials
-        handle = account.handle
-        backendName = account.backendName
+        lock.withLock {
```
johnxnguyen, line 144: thought: I'm wondering if it's feasible to make `Account` a immutable structure and to modify this method to return a mutated copy:

```swift
public func updateWith(_ account: Account) -> Account {
  guard userIdentifier == account.userIdentifier else { return 
  return Account(
    userIdentifier: userIdentifier,
    userName: account.userName,
    teamName: account.teamName ?? self.teamName,
    ...
  )
}
```

samwyndham: @johnxnguyen It will require wider changes because Account is relied upon to be a reference type - E.g. it is expected to be mutated in one place and those mutations to be relied upon in another place.

wire-ios-data-model/Source/Model/Account/Account.swift
samwyndham: Added in https://github.com/wireapp/wire-ios/pull/4340/commits/22d7eafbc7a5119aeac6b35da29c90af833bb096

jullianm (approved):
LGTM 👍

johnxnguyen (approved):

PR #4270, opened 2026-02-05
feat: respond to QR code backend feature flag - WPB-22095

JaCaLlaSchwarz:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22095" title="WPB-22095" target="_blank">TaskWPB-22095</a>  [iOS] Respond to QR code backend feature flag and turn QR code feature on / off according to flag setting
  </td></table>
  <br />
 

### Issue
Respond to QR code backend feature flag and turn QR code feature on / off according to flag setting

#### Case: Flag disabled (default) / (attached video:  WhenDisabled_WPB22095)
1. Select profile and check that QR button is presented. ('3)
2. Open with Safare user deeplink and check that user profile screen is presented ('25)

#### Case: Flag enabled / (attached video:  WhenEnabled_WPB22095)
1. Select profile and check that QR button is NOT presented. ('3)
2. Open with Safare user deeplink and check that on conversationo is presented an alert ('10)

### Evidences<table style="width: 100%; border-collapse: collapse; border: 1px solid black;">
  <thead>
    <tr style="background-color: #f2f2f2;">
      <th style="border: 1px solid black; padding: 8px;">Disabled</th>
      <th style="border: 1px solid black; padding: 8px;">Enabled</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td style="border: 1px solid black; padding: 8px;">BeforeFix_WPB20582</td>
      <td style="border: 1px solid black; padding: 8px;">AfterFix_WPB20582</td>
    </tr>
    <tr>
      <td style="border: 1px solid black; padding: 8px;">BeforeFix_WPB20582</td>
      <td style="border: 1px solid black; padding: 8px;">AfterFix_WPB20582</td>
    </tr>
  </tbody>
</table>

### Testing
- [x] testThatItFetchesSimplifiedUserConnectionRequestQRCode
- [x] testThatItFetchesUserProfileQRCodeItReturnsADefaultValueWhenObjectDoesNotExist
- [x] testThatItStoresSimplifiedUserConnectionRequestQRCode
- [x] testPull

#### Manual test
See upper Test section

### Checklist

- [x] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [x] Description is filled and free of optional paragraphs.
- [x] Adds/updates automated tests.

---

### UI accessibility checklist

_If your PR includes UI changes, please utilize this checklist:_
- [ ] Make sure you use the API for UI elements that support large fonts.
[...]

PR #4351, opened 2026-02-23
fix: enable SimplifiedUserConnectionRequestQRCode by default - WPB-22095

JaCaLlaSchwarz:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22095" title="WPB-22095" target="_blank">TaskWPB-22095</a>  [iOS] Respond to QR code backend feature flag and turn QR code feature on / off according to flag setting
  </td></table>
  <br />
 

### Issue
Default value was not validated when app access to an REST API version without feature flag defined.
This is a fix over: https://github.com/wireapp/wire-ios/pull/4270

#### Case: Default value
1. Login with a user to an environment where does not exist the feature flag is not defined.
2. go to user profile
EXPECTED: By default QR must be presented.

### Testing
- [x] testThatItFetchesSimplifiedUserConnectionRequestQRCode
- [x] testThatItFetchesUserProfileQRCodeItReturnsADefaultValueWhenObjectDoesNotExist
- [x] testThatItStoresSimplifiedUserConnectionRequestQRCode
- [x] testPull

#### Manual test
See upper Test section

### Checklist

---


WireDomain/Sources/WireDomain/Repositories/FeatureConfig/FeatureConfigRepository.swift
```diff
@@ -54,6 +54,10 @@ public final class FeatureConfigRepository: FeatureConfigRepositoryProtocol {
     public func isFeatureEnabled(
         _ feature: Feature.Name
     ) async -> Bool {
+        await isFeatureEnabled(feature, defaultValue: false)
```
David-Henner: why is it `false` here ?

JaCaLlaSchwarz: Because on catch error always returned false . Old line 65

image


In case of our flag is true, so it was created a function that did the same but returning in case of error default value.

David-Henner (approved):

netbe (changes requested):
please add the jira ticket to the PR and provide a unit test for this default behaviour

netbe (approved):
you need to test

PR #4354, opened 2026-02-24
chore: test-addfavouriteAndSortByfavourite - WPB-23601

findms:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23601" title="WPB-23601" target="_blank">TaskWPB-23601</a>  [iOS] Automate - Able to add favourite conversation TC-8869 
  </td></table>
  <br />
 

### Issue

New test - add conversation as favourite and filter by favourite
++ small change for handling any alert stuck from previous test failure
++ CallingTests name changed to ZCallingTests to run at the end, just to avoid if this test may impact others due to alert keep showing

### Testing

ran test locally - ran fine.

image


wire-ios/WireUITests/PersonalUserTests.swift
```diff
@@ -136,4 +136,73 @@ final class PersonalUsersTests: WireUITestCase {
 
         XCTAssertNotEqual(accountNameUserA, accountNameUserB, "Account name didn't change after deleting")
     }
+
+    /// testiny: https://app.testiny.io/IOS/testcases/tcf/1389/tc/8869
```
jullianm, line 140: question: does this map the result of the test (success or failure) on the testiny use cases page ?

jullianm (approved):
LGTM, left one question for understanding

wire-ios/WireUITests/PersonalUserTests.swift
findms: @jullianm  No..not yet..but the future plan are to remove this commented testiny id and use the TC-id in test function so that testiny CLI can mark pass/fail in test run. We are working on it.

wire-ios/WireUITests/ZCallingTests.swift
```diff
@@ -20,7 +20,7 @@ import WireFoundation
 import WireLocators
 import XCTest
 
-final class CallingTests: WireUITestCase {
+final class ZCallingTests: WireUITestCase {
```
samwyndham, line 25: suggestion: Add a comment here explaining the reason for the naming (so it happens last).

samwyndham (approved):
Nice work!

wire-ios/WireUITests/ZCallingTests.swift
findms: Added

SyamalaHari (approved):

wire-ios/WireUITests/ZCallingTests.swift
netbe: suggestion: add more details:  "just to avoid if this test may impact others due to alert keep showing"

netbe (approved):
LGTM:) just left a suggestion

wire-ios/WireUITests/ZCallingTests.swift
findms: Added this info as well

PR #4361, opened 2026-02-25
fix: made searchview & offline bar sticky on FileBrowserView - WPB-23484

OlivellaO:
### Issue

Offline bar on File Browser list was not moving and going behind the Searchview when scrolling. To fix that issue and make it more consistent with Conversation list now both Searchview and offline bar will stick to the top when scrolling.
Loading from pull to refresh will show under both.

### Testing

- Go to Drive screen
- Switch off phone internet
- Offline bar will show under searchview
- Try to scroll. Both Searchview and offline bar should stick to the top.

---

### Checklist

---


WilhelmOks (approved):

jullianm (changes requested):
fix looks good for the `FilesBrowserView` but I believe we need to implement the same changes in the `FilesView` as well as this view still has the issue.

jullianm (approved):
Nice work

PR #4363, opened 2026-02-25
fix: E2EI flow can be interrupted - WPB-23171

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23171" title="WPB-23171" target="_blank">BugWPB-23171</a>  [iOS] E2EI enrollment flow is interrupted
  </td></table>
  <br />
 

### Issue
When logging in to an account that requires End-To-End-Identity, if the user interrupts the E2EI enrollment flow (either put app in BG then FG, or restart app) then the user can bypass the flow and end up in a state where they should have E2EI but don't and can't get it. 

The logic to determine what is the top level app state (e.g authenticated -> go to conversation list, or e2ei required -> go to e2ei flow) is spread out in various places and is a fragile network of competing code paths. This broke at some point due to the various refactoring made to session loading and authentication flows.

The solution was to go step by step correcting various regressions, which I won't explain here but instead via comments on the changes.

### Testing
There are three main flows to test:

1. Log in to an account that requires E2EI without interruption (log in -> conversation list), verify success.
2. Log in to an account that requires E2EI, but before completion go BG, then FG, complete and verify success.
3. Log in to an account that requires E2EI, but before completion terminate app. Relaunch, see E2EI flow, complete and verify success.

---

### Checklist

---


wire-ios-request-strategy/Sources/Request Strategies/Feature configurations/Payload/FeatureConfigsPayloadProcessor.swift
```diff
@@ -210,109 +210,4 @@ struct FeatureConfigsPayloadProcessor {
         }
     }
 
-    func processEventPayload(
```
johnxnguyen, line 213: This is dead code.

wire-ios-data-model/Source/MLS/MLSClientManager.swift
```diff
@@ -53,13 +54,19 @@ public final class MLSClientManager: MLSClientManagerProtocol {
...
     ) async {
         guard isBackendMLSEnabled, mlsFeature.isEnabled else {
             WireLogger.mls.info("MLS feature in not enabled.")
             return
         }
 
+        if !hasRegisteredMLSClient && isE2EIRequired {
+            WireLogger.mls.info("MLS client needs to be initialized via E2EI.")
+            return
+        }
```
johnxnguyen: This early abort is necessary to prevent the MLS client being initialized with basic credentials (as opposed to E2EI credentials) if in fact we need E2EI. Otherwise, when we complete the E2EI enrollment flow it will fail because CoreCrypto complains that the MLS client was already initialized with basic credentials.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -71,6 +71,7 @@ public protocol SessionActivationObserver: AnyObject {
 // sourcery: AutoMockable
 public protocol SessionManagerDelegate: AnyObject, SessionActivationObserver {
     func sessionManagerWillLogout(
+        accountID: UUID?,
```
johnxnguyen, line 74: This delegate method eventually leads to the `unauthenticated` app state, which may require access to the relevant user session. Before, the `ZMUserSession.shared()` instance was used, which only works of course if there is a shared user session. However recent changes caused the shared user session to be nil (since it was set at a later point). To remove this dependency on a singleton, which isn't guaranteed to be the correct user session anyway, I decided to pass along the account id which will be used later to obtain the correct user session when needed.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1020,6 +1026,10 @@ public final class SessionManager: NSObject, SessionManagerType {
                     newSession: userSession,
                     coreDataStack: userSession.coreDataStack
                 )
+
+                userSession.operationLoop?.resumeEnqueuing()
+                await userSession.start()
```
johnxnguyen: Another issue I noticed is that when we load the user session via `UserSessionLoader` it will not only initialize the `ZMUserSession` instance but also call a method to `setup` the session. Within this method was code that triggered side affects, such as determining the client registration status and proceeding with network requests and initialization of MLS clients. The main problem is that this is happening too early, such as before the `SessionManager` is set as the delegate (to handle various events) or even the user session is cached (and accessed from other code paths).

To address this implicit side effect triggering when loading a session, I have:
- Disabled the `ZMOperationLoop` upon init and explicitly enabled it via `resumeEnqueuing` when ready. This ensures that no requests/flows are started early.
- Moved the code within `ZMUserSession` that triggers side effects into a `start` method and invoking this explicitly when ready.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1594,6 +1604,14 @@ extension SessionManager: UnauthenticatedSessionDelegate {
             return
         }
 
+        // The journal may have old values for the same user
+        // from a previous installation or login session.
+        var journal = Journal(
+            userID: account.userIdentifier,
+            storage: sharedUserDefaults
+        )
+        journal[.isInitialSyncRequired] = true
```
johnxnguyen, line 1612: This was another problem that I observed, so the change here is to always require an initial sync after login.

wire-ios-sync-engine/Source/Synchronization/ZMOperationLoop.m
```diff
@@ -171,6 +171,12 @@ - (void)executeNextOperation
     }];
 }
 
+- (void)resumeEnqueuing
+{
+    self.shouldStopEnqueueing = NO;
+    [self newRequestsAvailable];
+}
+
```
johnxnguyen, line 179: As mentioned previously, this is to control when the operation loop starts.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -675,23 +654,49 @@ public final class ZMUserSession: NSObject {
...
-                        // clear all items, those will be regenerated when sync is resumed
-                        await clientSessionComponent?.workAgent.clearSchedulerQueue()
-                    }
-                }
-            }.store(in: &cancellables)
-            // Initialize the generator to enqueue repair work item if needed
-            clientSessionComponent.repairFaultyMLSRemovalKeysGenerator.submitWorkItemIfNeeded()
+    }
+
+    public func start() async {
```
johnxnguyen, line 659: This method contains the code in `setup` (called when initializing/loading the session) that triggers side effects. It is now called explicitly as a separate step from initialization.

wire-ios/Wire-iOS/Sources/Authentication/Delegates/AuthenticationStatusProvider.swift
```diff
@@ -20,9 +20,7 @@ import Foundation
 import WireSyncEngine
 
 final class AuthenticationStatusProvider {
-    var sharedUserSession: ZMUserSession? {
-        ZMUserSession.shared()
-    }
+    var sharedUserSession: ZMUserSession?
```
johnxnguyen: Rather than depend on a singleton, inject the relevant user session.

wire-ios-data-model/Source/MLS/MLSClientManager.swift
```diff
@@ -53,13 +54,19 @@ public final class MLSClientManager: MLSClientManagerProtocol {
...
+        isBackendMLSEnabled: Bool,
+        isE2EIRequired: Bool
     ) async {
         guard isBackendMLSEnabled, mlsFeature.isEnabled else {
             WireLogger.mls.info("MLS feature in not enabled.")
             return
         }
 
+        if !hasRegisteredMLSClient, isE2EIRequired {
+            WireLogger.mls.info("MLS client needs to be initialized via E2EI.")
```
netbe:
```suggestion
            WireLogger.mls.info("MLS client needs to be initialized via E2EI.", attributes: .safePublic)
```

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
netbe: question: would it make sense to move userSession.operationLoop?.resumeEnqueuing() inside start()

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -675,23 +654,49 @@ public final class ZMUserSession: NSObject {
...
+        // Proactively ensure we clean up invalid connection state.
+        do {
+            let connectionValidator = ConnectionValidator(context: syncContext)
+            try await connectionValidator.cleanUpAllInvalidConnections()
+        } catch {
+            WireLogger.session.error("failed to clean up invalid connections: \(String(describing: error))")
+        }
 
+        guard let clientSessionComponent else {
+            return
```
netbe, line 685: question: could this happen? component being nil

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1256,18 +1261,22 @@ extension ZMUserSession: SyncAgentDelegate {
             guard let self else { return }
             await fetchAndStoreFeatureConfig()
 
-            let (qualifiedSelfClientID, hasRegisteredMLSClient) = await syncContext.perform {
+            let (qualifiedSelfClientID, hasRegisteredMLSClient, isE2EIRequired) = await syncContext.perform {
                 let selfClient = ZMUser.selfUser(in: self.syncContext).selfClient()
                 let hasRegisteredMLSClient = selfClient?.hasRegisteredMLSClient == true
-                return (selfClient?.qualifiedClientID, hasRegisteredMLSClient)
+
+                let isE2EIRequired = LegacyFeatureRepository(context: self.syncContext).fetchE2EI().isEnabled
```
netbe: suggestion: use FeatureConfigRepository instead of legacy one, should be available in clientSessionComponent

netbe (approved):
left a few comments but looks good

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
samwyndham: praise: separating initialization & start code.

samwyndham (approved):
Nice work! I wonder if this change might improve the `setUpSyncAgent` hang I was looking at now that the run loop is paused until after setup is complete.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
johnxnguyen: @netbe yes good idea

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: yes, when there's no self client registered yet. But once registration happens, it'll be set up.

wire-ios-data-model/Source/MLS/MLSClientManager.swift
johnxnguyen: done

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: done

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
netbe: it'll be setup via the same method?

wire-ios/Wire-iOS/Sources/Authentication/Delegates/AuthenticationStatusProvider.swift
```diff
@@ -20,7 +20,11 @@ import Foundation
 import WireSyncEngine
 
 final class AuthenticationStatusProvider {
-    var sharedUserSession: ZMUserSession?
+
+    private let _sharedUserSession: ZMUserSession?
+    var sharedUserSession: ZMUserSession? {
+        _sharedUserSession ?? ZMUserSession.shared()
```
netbe, line 26: question: could you explain what's going on here?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: It gets set up when we set up the sync agent, which happens when the self client is registered: https://github.com/wireapp/wire-ios/blob/c7bbe9755172dc528c50194fb3d1125acd305d7a/wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift#L1379

wire-ios/Wire-iOS/Sources/Authentication/Delegates/AuthenticationStatusProvider.swift
johnxnguyen: Yes, initially the `AuthenticationStatusProvider` was using the `ZMUserSession.shared()` singleton as a way to provide the authentication coordinator with a user session (to complete client registration). I had made an earlier change to fix a problem where relying on this singleton didn't work (it wasn't set yet) by passing along the account id and fetching the correct user session when setting up the authentication coordinator. However... then I found out that in some cases (I can't remember which), the authentication coordinator needs to get the user session mid-flow, meaning if previously there was no shared user session (e.g first launch) and then it needed one, it would be out of luck and get completely stuck. The quick solution I chose is to fallback to the singleton as we were doing before. It's not ideal I know, but this whole authentication flow is not ideal and will be addressed soon.

netbe (approved):
re approving the last changes

PR #4373, opened 2026-02-26
chore: remove many calls to `ZMUserSession.shared()` - WPB-23738

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23738" title="WPB-23738" target="_blank">TaskWPB-23738</a>  [iOS] Remove calls to ZMUserSession.shared() where possible
  </td></table>
  <br />
 

### Issue

This PR removes many calls to the deprecated `ZMUserSession.shared()`.

| Before | After |
| --- | --- | 
| Screenshot 2026-02-26 at 13 48 50 | Screenshot 2026-02-26 at 13 50 12 |

I hope this will help lay down some work towards simplifying session management by removing the implicit dependency of SessionManager from many components. (`ZMUserSession.shared()` uses `SessionManager.shared`) internally.

I used Claude code to help me here which generally worked well but also made some funny choices that I changed. It also took me longer than I had hoped.

In general I've tried to keep changes to a minimum and to avoid behavior changes. `ZMUserSession` is has generally be passed through `init` methods as a `UserSession` and calls to `ZMUserSession.shared()` have often been changed to `UserSession as? ZMUserSession`.

I noticed that a lot of our tests depend on not using a real `ZMUserSession` which I haven't attempted to fix.

While this PR does get rid of a lot of calls to `ZMUserSession.shared()` that call is still used by for example `ZMUser.selfUser()` so may take quite some effort to completely get rid of it.

**I recommend checking this PR commit by commit. There are some overly large commits but many are small**

### Testing

Changes can't really be tested outside a full regression test at least the following should be tested before merge as it 

1. Login with multiple accounts (A + B)

Repeat the following with call kit enabled & disabled and in the foreground, background, and app quit
(6 states).

2. Ensure account A is seleced
3. Receive a call to account B
4. Answer call
5. Account should switch.

---

### Checklist


wire-ios/Wire-iOS/Sources/Managers/ProximityMonitorManager.swift
```diff
@@ -46,10 +48,12 @@ final class ProximityMonitorManager: NSObject {
...
     }
 
-    override init() {
+    init(userSession: UserSession) {
+        self.userSession = userSession
         super.init()
 
-        guard let userSession = ZMUserSession.shared() else {
+        // This is a hack for testing as a real ZMUserSession is not available
+        guard userSession is ZMUserSession else {
```
netbe, line 55: suggestion: change zmlog as WireLogger

wire-ios/Wire-iOS/Sources/UserInterface/Calling/BottomSheetContainerViewController/CallingBottomSheetViewController.swift
```diff
@@ -109,7 +109,8 @@ final class CallingBottomSheetViewController: BottomSheetContainerViewController
     override func viewDidLoad() {
         super.viewDidLoad()
 
-        guard let userSession = ZMUserSession.shared() else {
+        // This is a hack for testing as a real ZMUserSession is not available
+        guard userSession is ZMUserSession else {
             zmLog.error("UserSession not available when initializing \(type(of: self))")
```
netbe: suggestion: change zmlog as WireLogger

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/FileTransfer/AudioMessageView.swift
```diff
@@ -184,6 +178,17 @@ final class AudioMessageView: UIView, TransferView {
         }
     }
 
+    func setUserSession(userSession: UserSession) {
```
netbe, line 181: question: where is this method called?

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/FileTransfer/AudioMessageView.swift
```diff
@@ -184,6 +178,17 @@ final class AudioMessageView: UIView, TransferView {
...
+        self.userSession = userSession
+
+        if callStateObserverToken == nil, let userSession = userSession as? ZMUserSession {
```
netbe: question: I notice the check for callStateObserverToken == nil compared to before, is it needed?

netbe (approved):
this is quite a large PR so hard to be thorough. I left some comments but looking at the complexity it looks fine to me

johnxnguyen (approved):
Nice work

wire-ios/Wire-iOS/Sources/Managers/ProximityMonitorManager.swift
samwyndham: Fixed in https://github.com/wireapp/wire-ios/pull/4373/commits/5246a76782dba41ccaddc0ec8d7f007eeba051b9

wire-ios/Wire-iOS/Sources/UserInterface/Calling/BottomSheetContainerViewController/CallingBottomSheetViewController.swift
samwyndham: Fixed in https://github.com/wireapp/wire-ios/pull/4373/commits/5246a76782dba41ccaddc0ec8d7f007eeba051b9

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/FileTransfer/AudioMessageView.swift
samwyndham: So the reason for the check is that originally this happened one time in the initializer. But it doesn't really make sense considering now that the userSession can technically change due to the API change. 

In practice this shouldn't change but I have updated this code in `https://github.com/wireapp/wire-ios/pull/4373/commits/864bcc74bc08f7f78bff9a02118908c82e1ae2c3` so that if the userSession does change we re-observe.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/FileTransfer/AudioMessageView.swift
samwyndham: @netbe  This view is a subview of reusable cells in a couple of contexts:

- In an audio message in a conversation
- In an audio message in the search screen.

In both called when the parent cell is configured or re-configured (considering the cells are reused).

PR #4377, opened 2026-02-27
fix: userSession memory leak - WPB-18909

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18909" title="WPB-18909" target="_blank">BugWPB-18909</a>  [iOS] ZMUserSession is still retained in memory after logout
  </td></table>
  <br />
 

### Issue

This is an attempt to fix some of memory leaks involving userSession.

* add XCUITest to check memory after logout
* add memory label for userSession in DeveloperTools
* remove unused files

Note: this works on an account without activity, but logging out while receiving or having requests ongoing would still be a problem. 

### Testing

**Before fix:**

https://github.com/user-attachments/assets/6285c75d-ef16-46d2-a82e-0543308e99f9

** After fix:**
https://github.com/user-attachments/assets/40de0e70-8c91-4b71-8c4a-ccb0e2d6ce2d

---

### Checklist

---


wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+Clients.swift
```diff
@@ -42,7 +42,8 @@ public extension ZMUserSession {
 
     @objc
     func fetchAllClients() {
-        syncManagedObjectContext.performGroupedBlock {
+        syncManagedObjectContext.performGroupedBlock { [weak self] in
+            guard let self else { return }
             self.applicationStatusDirectory.clientUpdateStatus.needsToFetchClients(andVerifySelfClient: false)
             RequestAvailableNotification.notifyNewRequestsAvailable(self)
```
valerio-bettini, line 48: @netbe I've just bumped into this PR randomly mate.
Conscious I know 0% of the code yet :-) , but it looks weird we attach self to the notification payload. If that is not avoidable, can we send it as an optional? As this class can remain in memory because it's retained somewhere else, as we(I) don't know what's happening once self is sent through the notification

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+Clients.swift
```diff
@@ -55,7 +56,8 @@ public extension ZMUserSession {
         client.markForDeletion()
         client.managedObjectContext?.saveOrRollback()
 
-        syncManagedObjectContext.performGroupedBlock {
+        syncManagedObjectContext.performGroupedBlock { [weak self] in
+            guard let self else { return }
             self.applicationStatusDirectory.clientUpdateStatus.deleteClients(withCredentials: credentials)
             RequestAvailableNotification.notifyNewRequestsAvailable(self)
```
valerio-bettini, line 62: same here

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+Clients.swift
netbe: that's really legacy but also core principal method for operation loop 

so looking at the code it does not even use it:
Screenshot 2026-06-19 at 14 37 18

I can only guess it should be used as the object posting the notifcation instead of nil to know

PR #4381, opened 2026-03-03
chore: drop iOS 16 - WPB-22704

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22704" title="WPB-22704" target="_blank">EpicWPB-22704</a>  [iOS] Drop iOS 16
  </td></table>
  <br />
 

### Issue

Drop iOS 16:
- Bump deployment target to iOS 17 throughout the workspace
- Address deprecation warnings
- Remove iOS 17 availability checks

### Testing
NA

---

### Checklist

---


WireAnalytics/Package.swift
```diff
@@ -10,7 +10,7 @@ let isCountlyEnabled = hasEnvironmentVariable("ENABLE_COUNTLY", "true")
 
 let package = Package(
     name: "WireAnalytics",
-    platforms: [.iOS("16.4"), .macOS(.v12)],
```
netbe, line 13: **thought:** it would be nice to have only one place in the project to update the version. For packages we could have a swift file containing the iOS version as a constant and have it imported in here. Not sure how to do it for Xcode project: another common .xconfig ?

So a total of 2 places instead of current ones

WireCalling/Sources/WireCallingUI/Views/WireMeetings/MeetingsView/MeetingsView.swift
```diff
@@ -50,7 +50,7 @@ struct MeetingsView: View {
         .onAppear {
             viewModel.loadInitialData()
         }
-        .onChange(of: viewModel.selectedTab) { newValue in
+        .onChange(of: viewModel.selectedTab) { _, newValue in
```
netbe, line 53: question: out of curiosity is the omitted value the currentValue?

can we use https://developer.apple.com/documentation/swiftui/view/onchange(of:initial:_:)-8wgw9 in this case if we omit the value?

netbe (commented):
Image

I believe there are a few more places to update

WireCalling/Sources/WireCallingUI/Views/WireMeetings/MeetingsView/MeetingsView.swift
johnxnguyen: The first argument is the old value, so yes we can use this version.

johnxnguyen: I would keep it as it is for instances where we reference the new value because I think it's clearer, but in cases where we reference neither then I'll use the 0 argument version.

WireAnalytics/Package.swift
johnxnguyen: For Swift packages, it would be nice but I don't believe there's a simple way to do this, beside writing a script that performs the changes for you. Though I don't think that's really necessary if we can ask an agent to change the version.

WireAnalytics/Package.swift
```diff
@@ -10,7 +10,7 @@ let isCountlyEnabled = hasEnvironmentVariable("ENABLE_COUNTLY", "true")
...
+    platforms: [.iOS("17.0"), .macOS(.v12)],
```
samwyndham, line 13: nitpick suggestion: We can do the following. Maybe it is no better though as it could make searching harder. Feel free to ignore this comment.

```suggestion
    platforms: [.iOS(.v17), .macOS(.v12)],
```

samwyndham (approved):
How exciting!

netbe (approved):
LGTM;)

WireAnalytics/Package.swift
johnxnguyen: I think .v17 would resolve to the max iOS 17 version (not necessarily 17.0.

PR #4383, opened 2026-03-03
fix: missing build number in internal build icons - WPB-23820

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23820" title="WPB-23820" target="_blank">BugWPB-23820</a>  [iOS] Broken build number app icon
  </td></table>
  <br />
 

### Issue

The CI is currently not adding a build number to the beta build (etc) app icons. 

Fixing this confused me. I incorrectly assumed it was because of CI/CD image update that had caused this. But in the end it turned out to be due to changes in the compiled dependencies in the standard version of `imagemagick` installed via brew.

`imagemagick` has had a bunch of dependencies removed from it including `fontconfig` which `imagemagick` depends on to locate fonts on the system. There is a new brew `imagemagick-full` which keeps `fontconfig` but this requires a bit of setup getting it working. 

Instead of making changes above, in this PR I take the simple solution of providing the path to a font `Helvetica` on the system. This would break if Apple stops shipping Helvetica font of if the path of fonts move but I think that change is unlikely to happen.

### Test

[This build](https://github.com/wireapp/wire-ios/actions/runs/22633206272/job/65589094515) should have a build number on the icon.

### Checklist

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -897,7 +897,7 @@ class Build
             height = Integer(image_height) / 4.0
 
             # Add build number to the icon
-            %x( magick -background '#0008' -fill white -gravity center -size #{width}x#{height} caption:"#{extra_info}" "#{image}" +swap -gravity south -composite "#{image}" )
+            %x( magick -background '#0008' -fill white -font '/System/Library/Fonts/Helvetica.ttc' -gravity center -size #{width}x#{height} caption:"#{extra_info}" "#{image}" +swap -gravity south -composite "#{image}" )
             processed += 1
         end
         UI.important("Processed #{processed} icons in #{iconset_location} by adding '#{extra_info}'")


netbe (approved):
LGTM;) 

I notice the warnings in the build logs:
```
freetype 2.14.1_2 is already installed and up-to-date. To reinstall 2.14.1_2, run: brew reinstall freetype
imagemagick 7.1.2-15_1 is already installed and up-to-date. To reinstall 7.1.2-15_1, run: brew reinstall imagemagick
```

suggestion: I guess we could update the Brewfile and remove completely  

```
# CI-only dependencies
brew "imagemagick" if ENV['CI']
brew "ghostscript" if ENV['CI']
```

as this is not needed locally only to build apps

johnxnguyen (approved):

samwyndham:
> LGTM;)
> 
> I notice the warnings in the build logs:
> 
> ```
> freetype 2.14.1_2 is already installed and up-to-date. To reinstall 2.14.1_2, run: brew reinstall freetype
> imagemagick 7.1.2-15_1 is already installed and up-to-date. To reinstall 7.1.2-15_1, run: brew reinstall imagemagick
> ```
> 
> suggestion: I guess we could update the Brewfile and remove completely
> 
> ```
> # CI-only dependencies
> brew "imagemagick" if ENV['CI']
> brew "ghostscript" if ENV['CI']
> ```
> 
> as this is not needed locally only to build apps

@netbe This is actually coming from the caching that gerlero/brew-install@v1 uses I believe as the warning is generated in that step which is before the setup script has run. It is because of caching that that action does which I guess we want to keep

However I will remove those calls as suggested. I'll do it in another PR straight to develop to avoid breaking 4.16 branch in case there is something we have missed.

PR #4398, opened 2026-03-05
fix: missing 1on1 conversation after accepting federated connection - WPB-23838

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23838" title="WPB-23838" target="_blank">BugWPB-23838</a>  [iOS] Federated 1:1 conversation disappearing from the list
  </td></table>
  <br />
 

### Issue

When accepting a new connection from another backend after the 1-1 resolution, the conversation disappears from the list sometime:
The new mls 1-1 conversation is created then picked up by the `ConversationUpdatesGenerator` and tries to sync with backend via the POST /conversations/list endpoint which returns a result as notFound. The UpdateConversationItem marks as deleted the conversation in this case.

The reason why the server returns a result as notFound is that 1-1 conversations are only returned once the first commit bundle has been processed. This commit bundle is received during the 1-1 resolution which in some cases (race condition) happen after the UpdateConversationItem work.

Solution: don't mark 1-1 conversation as needsToBeUpdatedFromBackend when created so it's not picked up by `ConversationUpdatesGenerator`

### Testing

* Added XCUITest

https://github.com/user-attachments/assets/5c63a480-c297-4690-a2f5-04feb1e23913

---

### Checklist

---


WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift
```diff
@@ -1194,9 +1204,12 @@ public final class ConversationLocalStore: ConversationLocalStoreProtocol {
             return
         }
 
+        // for 1:1 we don't want to set needsToBeUpdatedFromBackend since backend only returns them once the first
```
netbe, line 1207: this is the key fix

WireDomain/Sources/WireDomain/WorkAgent/WorkItem/UpdateConversationItem.swift
```diff
@@ -55,6 +55,19 @@ struct UpdateConversationItem: WorkItem {
                 attributes: [.conversationId: conversationID.id.uuidString],
                 .init(self)
             )
+
+            guard await repository.conversationNeedsToBeUpdatedFromBackend(
```
netbe: if when dequeued from the WorkAgent the conversation has been updated elsewhere, there's no need to perform the operation anymore

johnxnguyen (approved):

findms (approved):
LGTM :)  For Later:  We might need to add testcase for this in testiny and add id

PR #4400, opened 2026-03-05
chore: add some concurrency safety to SessionManager state - WPB-23571

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23571" title="WPB-23571" target="_blank">TaskWPB-23571</a>  [iOS] Make `SessionManager` state thread safe
  </td></table>
  <br />
 

### Issue

With this PR I'm trying to add some concurrency safety around updating state in the session manager. I was quite optimistic before starting it but it has become a bit of a struggle. The ideal would be that SessionManager is an Actor  but this doesn't feel achievable in the short term. I had hoped to perhaps mark some properties as `@MainActor` but I don't think this is possible as it looks like much of this state might be accessed from various threads/queues. So in this PR I:

- Attempt to protect access to `activeUserSession`, `backgroundUserSessions` & `unauthenticatedSession` using an `OSAllocatedUnfairLock`.
- In a limited number of places ensure multiple access is done in the same `withLockUnchecked` call. I've been careful in these cases to avoid calling out to other methods to avoid possible deadlocks.
-  Marked a bunch of methods as @MainActor in AppRouteRouter that I know to be call only from @MainActor.

As you can see this code all gets additionally complex when we try to bring thread safety into it.

### Testing

No specific test steps. Instead:

- Exploritory testing with multiple accounts
- Regression suite.

---

### Checklist


wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -934,14 +952,13 @@ public final class SessionManager: NSObject, SessionManagerType {
...
 
-        activeUserSession = session
+        setActiveUserSession(session)
 
         WireLogger.sessionManager.debug(
             "Activated ZMUserSession for account - "
                 + account.userIdentifier.safeForLoggingDescription
         )
 
-        delegate?.sessionManagerDidChangeActiveUserSession(userSession: session)
```
samwyndham, line 916: This is a no-op so deleting the delegate method altogether

David-Henner (approved):
Nice work! Were there a lot of issues related to concurrency in the SessionManager before ?

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -266,37 +266,60 @@ public final class SessionManager: NSObject, SessionManagerType {
...
-            activeUserSession?.appLockController.beginTimer()
-            activeUserSession?.setAnalyticsEventTracker(nil)
-        }
+    struct State {
+        var activeUserSession: ZMUserSession?
+        var backgroundUserSessions: [UUID: ZMUserSession] = [:]
+        var unauthenticatedSession: UnauthenticatedSession?
+    }
+
+    private let state = OSAllocatedUnfairLock<State>(uncheckedState: State())
```
johnxnguyen, line 276: suggestion: add a comment to explain these changes for the next person

johnxnguyen (approved):

PR #4402, opened 2026-03-05
fix: crash in background with ExpiringActivity - WPB-23839

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23839" title="WPB-23839" target="_blank">BugWPB-23839</a>  [iOS] Crash CoreFoundation: __CFRunLoopServiceMachPort + 160
  </td></table>
  <br />
 

### Issue

 The app was being killed by the iOS watchdog with termination reason FRONTBOARD 0xBAADCA11 after running
   for ~6 hours in the background. The crash was observed in TestFlight (Wire 4.16.0 / build 17993, iPhone
   18,1, iOS 26.3).

```
  Exception Type:  EXC_CRASH (SIGKILL)
  Termination Reason: FRONTBOARD 0xbaadca11

  Thread 14 (crashing):
    semaphore_wait_trap
    _dispatch_semaphore_wait_slow
    closure #1 in closure #1 in ExpiringActivityManager.withExpiringActivity(reason:block:)
    ExpiringActivity.swift:79
```

  0xBAADCA11 ("bad call") is the watchdog timer — iOS killed the app because it blocked a system-managed
  thread indefinitely.

  **Root cause in ExpiringActivity.swift:**

  performExpiringActivity's callback must block its thread until the work is done (Apple API requirement).
   The implementation used a DispatchSemaphore for this:

  // expiring = false branch
  Task { try await self.startWork(block: block, semaphore: semaphore).value ... }
  semaphore.wait()  // ← blocks the callback thread

  // expiring = true branch
  Task { try await self.stopWork() }  // cancels the inner task
  // ← semaphore is never signaled if block() ignores Task.cancel()

  When expiring = true fires, stopWork() calls task.cancel(). If block() is a cooperative caller that
  checks Task.checkCancellation(), the inner task exits and its defer { semaphore.signal() } fires
  normally. But if block() does not cooperate with cancellation — which is a valid real-world scenario —
  the inner task keeps running, semaphore.signal() is never called, and semaphore.wait() blocks the
  callback thread indefinitely until the watchdog intervenes.

  A secondary bug was also present: if expiring = true fires before the actor executes startWork (a race
  condition), self.task is nil, stopWork() throws, and continuation.resume(throwing:) is called. When
  startWork later runs and block() completes, continuation.resume() is called a second time — undefined
  behaviour on CheckedContinuation.

**Solution:**

  Two changes:

  1. The expiring = true branch now always signals the semaphore and resumes the continuation directly,
  without relying on the inner task to do so:

```
  Task {
      do {
          try await self.stopWork()
          finish(.failure(CancellationError()))   // resume caller immediately
      } catch {
          finish(.failure(error))
      }
      semaphore.signal()   // always unblock callback thread
  }
```

  This guarantees the callback thread is unblocked and withExpiringActivity returns to the caller
  regardless of whether block() cooperates with cancellation.

  2. All continuation.resume() calls are routed through OnceAction, a simple NSLock-guarded wrapper that
  ensures the continuation is resumed at most once. This eliminates the double-resume race and makes the
  fix safe.
  
### Testing

- testBug1_ActivityCompletesAfterExpiry_WhenBlockIgnoresCancellation — verifies withExpiringActivity
  returns promptly on expiry even when block() loops with Task.yield() and never calls
  Task.checkCancellation().
 - testBug2_NoContinuationDoubleResume_WhenExpiryRacesBeforeStartWork — verifies no double-resume occurs
  when the expiring = true callback is processed by the actor before startWork.

---

### Checklist

---


wire-ios-system/Source/ExpiringActivity.swift
```diff
@@ -61,37 +61,57 @@ actor ExpiringActivityManager {
 
     func withExpiringActivity(reason: String, block: @escaping () async throws -> Void) async throws {
         try await withCheckedThrowingContinuation { continuation in
+            // Shared between both callback branches.
+            let semaphore = DispatchSemaphore(value: 0)
```
caldrian: Strictly speaking this violates the contract that threads are always able to make forward progress.

netbe: but it was already the case before, the semaphore.wait() didn't change. It's ok because performExpiringActivity operates on a private queue and is the one where we wait so we don't block the cooperative pool

caldrian (approved):

caldrian:
Could/should this go into `release/cycle-4.17`? @netbe

wire-ios-system/Source/ExpiringActivity.swift
```diff
@@ -61,37 +61,57 @@ actor ExpiringActivityManager {
...
                             try await self.stopWork()
+                            // Task was running and is now cancelled. Resume the continuation
+                            // immediately — we cannot rely on the task's defer to signal the
+                            // semaphore if block() ignores cooperative cancellation.
+                            finish(.failure(CancellationError()))
                         } catch {
-                            continuation.resume(throwing: error)
+                            finish(.failure(error))
                         }
+                        // Unblock the callback thread regardless. A double-signal (if the
```
typfel: Doesn't this mean we are not waiting until the task has finished "cancelling" anymore and we'll instead immediately exit function and let the app suspend. 

Couldn't this happen:
1. start expiring activity decrypting messages
2. activity expires and signal to cancel the decryption task
3. we exit the activity
4. crash because the decryption task didn't finish cancelling and therefore didn't have time to release the transaction lock.

wire-ios-system/Source/ExpiringActivity.swift
```diff
@@ -47,45 +47,62 @@ public func withExpiringActivity(reason: String, block: @escaping () async throw
...
+    
     init(api: any ExpiringActivityInterface) {
         self.api = api
     }
-
+    
     func withExpiringActivity(reason: String, block: @escaping () async throws -> Void) async throws {
         try await withTaskCancellationHandler {
             try await withCheckedThrowingContinuation { continuation in
+                // Shared between both callback branches.
```
netbe, line 65: @David-Henner since you added withTaskCancellationHandler I wonder if the PR still makes sense

netbe:
closing in favor of #4550

PR #4410, opened 2026-03-09
fix: missing messages after update from 3.120 to 4.16 - WPB-23849

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23849" title="WPB-23849" target="_blank">BugWPB-23849</a>  [iOS] Missing messages after app migration from 3.120 to 4.16
  </td></table>
  <br />
 

### Issue

When updating from `3.120` to `4.16`, events (messages) received during the update are missing once the app is opened.

### Cause

During this update, we're migrating legacy events to the new sync. This causes an initial sync to be triggered to ensure no updates were missed. Starting this sync pulls the last event ID from the backend and overrides the local last event ID. 

This leaves a gap between the last processed (legacy) event ID and the last event ID fetched from the backend.

### Solution

~Abort pulling the last event ID when there's already a local last event ID in storage.~

Perform a resources sync instead of an initial sync to avoid pulling the last event ID

### Testing

- Install 3.120
- Log in
- Spam a conversation
- Background the app & update to 4.16
- Open the app and check conversation

---

### Checklist


WireDomain/Sources/WireDomain/Synchronization/PullLastUpdateEventIDSync.swift
```diff
@@ -37,6 +37,14 @@ struct PullLastUpdateEventIDSync: PullLastUpdateEventIDSyncProtocol {
     }
 
     func pull() async throws {
+        if let id = store.lastEventID() {
+            WireLogger.sync.debug(
+                "skipping pull of last event id because it is already stored",
+                attributes: [.eventEnvelopeID: id]
+            )
+            return
```
johnxnguyen: Thought: we should be careful here... perhaps there is a valid case to override an existing last event id (such as one that has become stale). I wonder... can't we trigger a resource sync instead of an initial sync? Or simply pass in a bool to skip fetching the last event id?

David-Henner: The issue is that the sync is triggered indirectly by setting the `.isInitialSyncRequired` flag in the Journal before we have a session. 

https://github.com/wireapp/wire-ios/blob/5ccd00c9dca38138927c83066e87c19d3790627c/wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift#L416

David-Henner: @johnxnguyen I wonder, is it actually necessary to have an initial sync after migrating legacy events? Can't we just pick up syncing events from the last event that was migrated ?

johnxnguyen: Initial sync isn't needed? But rather a resource sync is, because we only migrate the events that contain data that can't be pulled via a resource sync.

David-Henner: [9bbfaa0](https://github.com/wireapp/wire-ios/pull/4410/commits/9bbfaa04ed42010be43db27ddb1525c53cd486dc)

wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
```diff
@@ -411,9 +411,9 @@ final class UserSessionLoader {
         do {
             if try await migrator.isMigrationNeeded() {
                 try await migrator.migrateLegacyUpdateEvents()
-                // Since we only migrate some events, we require an
-                // initial sync to ensure we didn't miss updates.
-                journal[.isInitialSyncRequired] = true
+                // Since we only migrate some events, we require a
```
netbe, line 414: question: what events are not covered? @johnxnguyen

netbe (approved):
Looks good to me, but I'll let @johnxnguyen confirm this is correct and does not have any issue

wire-ios-sync-engine/Source/SessionManager/UserSessionLoader.swift
johnxnguyen: Conversation create, conversation rename, receipt settings updated, etc.

johnxnguyen (approved):
Looks good

PR #4422, opened 2026-03-11
feat: opening and loading improvement - WPB-23919

WilhelmOks:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23919" title="WPB-23919" target="_blank">TaskWPB-23919</a>  [iOS] Refactor File Download state to prepare for loading & opening behavior
  </td></table>
  <br />
 

### Issue

This is just preparation code for the main feature tracked by WPB-23688. 
It doesn't change the behaviour of the app but it uses the new code in a small part of the existing code as a kind of a proof of concept and to see that it works.

The new code introduces a `WireDriveFileUITracker`, which intends to provide different parts of the app where Drive files are present, with UI states and state changes specifically and exclusively required by the UI.

Those new UI states are required for the new design, for example the new "ready to open" state that is only relevant to the UI.

Changes to the UI behaviour according to the new design will be implemented in other PRs.

### Testing

Check if the Drive files list works exactly like before, now that it uses `WireDriveFileUITracker`.


WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/Item/FilesItemViewModel.swift
```diff
@@ -112,6 +122,12 @@ final class FilesItemViewModel: ObservableObject {
         self.isBrowsing = isBrowsing
         self.isInRecycleBin = isInRecycleBin
 
+        self.fileTracker = .init(downloadState: .pending)
+        fileTracker.fileShouldOpen = {
+            // TODO: call the function to open the file here, later when we implement the new loading&opening design
```
caldrian:
```suggestion
            // TODO: [WPB-23688] call the function to open the file here, later when we implement the new loading&opening design
```

WilhelmOks: thanks, done

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/WireDriveFileUITracker.swift
```diff
@@ -0,0 +1,106 @@
...
+
+import Foundation
+public import WireMessagingDomain
+
+/// Intended for tracking the UI state of Drive files in different parts of the app.
+/// Consumes state changes from `WireDriveLocalAsset.DownloadState` and provides observable state changes specifically
+/// for the UI that represents the file.
+/// Also notifies if and when a file should be opened automatically after download.
+@MainActor
+public final class WireDriveFileUITracker: ObservableObject {
```
caldrian: question: Now that we have dropped iOS 16 could we use `@Observable`?
https://developer.apple.com/documentation/SwiftUI/Migrating-from-the-observable-object-protocol-to-the-observable-macro

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/WireDriveFileUITracker.swift
```diff
@@ -0,0 +1,106 @@
...
+
+    public init(downloadState: WireDriveLocalAsset.DownloadState) {
+        self.state = Self.stateFromDownloadState(downloadState)
+    }
+
+    public func handleDownloadState(_ downloadState: WireDriveLocalAsset.DownloadState) {
+        state = Self.stateFromDownloadState(downloadState)
+    }
+
+    private static func stateFromDownloadState(_ downloadState: WireDriveLocalAsset.DownloadState) -> State {
```
caldrian, line 96: suggestion: Could we do this in an extension to `WireDriveFileUITracker.State`?
I think it's convenience to override `init` [without argument labels](https://www.swift.org/documentation/api-design-guidelines/#argument-labels).
> initializers that perform value preserving type conversions, omit the first argument label

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/WireDriveFileUITracker.swift
```diff
@@ -0,0 +1,106 @@
...
+                    downloadingTimerStart = Date()
+                }
+            } else {
+                downloadingTimerStart = nil
+            }
+        }
+    }
+
+    /// This closure will be called when a file should be automatically opened after the download.
+    public var fileShouldOpen: (() -> Void)?
```
caldrian, line 84: question: Maybe I'm missing the complete picture now, but shouldn't we pass some argument to this closure? Maybe the file or `self`.

suggestion: Could we also name it `onOpenFile` for example?

caldrian (approved):

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/WireDriveFileUITracker.swift
WilhelmOks: I'd like to use that, but since this type is intended to be used in existing viewmodels which use the old style ObservableObject code, I'm not sure it would work together.
I can try and see if it works.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/WireDriveFileUITracker.swift
WilhelmOks: We don't need to pass any parameter here because each tracker only tracks one file, not multiple ones.
The receiver of this call has the context for the file which should be opened.

Imo, `onOpenFile` has the semantics of "this file has opened, and this is a notification about that".
But the actual semantics are supposed to be "this file needs to be opened now. call the code to open the file."

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/WireDriveFileUITracker.swift
WilhelmOks: The reason why I did it this way is because I don't see it merely as a state mapped to another state.
I see it as building state changes based on the changes of another state, managed by the tracker type.
Some of the state changes depend on the state of the tracker.
I think making it an initializer of `WireDriveFileUITracker.State` would convey the wrong idea of it being just a mapping without the involvement of the tracker.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/WireDriveFileUITracker.swift
WilhelmOks: I tested it and it works! It even fixed an issue from an oversight to add `@Published` to the state var.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/WireDriveFileUITracker.swift
```diff
@@ -0,0 +1,108 @@
...
+        /// `downloaded(showReadyToOpen: true)` indicates that the file should be shown in a specific "ready to open"
+        /// state in the UI.
+        /// Will automatically change to `downloaded(showReadyToOpen: false)` state after a few seconds.
+        case downloaded(showReadyToOpen: Bool)
+
+        /// The download has failed.
+        case failed(error: any Error)
+    }
+
+    public private(set) var state: State {
```
samwyndham, line 47: optional suggestion: I'm wondering if it is possible to make this a computed property which is a function of a few stored properties:

- `downloadState: WireDriveLocalAsset.DownloadState`
- `deadlineReached: Bool` which is set to true 3 seconds after handleDownloadState is called. 

I'm not yet familiar with `@Observable` but I think it might work and that it might simplify the logic in this file a bit.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/WireDriveFileUITracker.swift
```diff
@@ -0,0 +1,108 @@
...
+        didSet {
+            switch (oldValue, state) {
+            case (.downloading, .downloaded):
+                // state is changing from downloading to downloaded:
+
+                // if the download takes longer than this amount of time, the file is considered "big"
+                let downloadTimeForBigFiles: TimeInterval = 0.3
```
samwyndham, line 54: sanity check: Is this meant to be 0.3 seconds or 3 seconds?

samwyndham (approved):

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/WireDriveFileUITracker.swift
```diff
@@ -0,0 +1,108 @@
...
+
+                let fileIsBig = downloadingTimerStart
+                    .flatMap { Date().timeIntervalSince($0) > downloadTimeForBigFiles } ?? true
+
+                state = .downloaded(showReadyToOpen: fileIsBig)
+
+                if fileIsBig {
+                    Task {
+                        try? await Task.sleep(for: .seconds(3))
+                        self.state = .downloaded(showReadyToOpen: false)
```
netbe, line 64: question: isn't it a bit dangerous to modify state in a Task which is running in it's own didSet. Is there a risk of race condition?

netbe (commented):
left a comment

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/WireDriveFileUITracker.swift
WilhelmOks: 300ms aka 0.3 seconds

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/WireDriveFileUITracker.swift
WilhelmOks: It would work because the changes of the stored properties would trigger the observation notifications, effectively making the computed properties observable as well.

However I'm not in favor of introducing variables like `deadlineReached: Bool` because in this case the value of this boolean would be meaningful only for specific states, making it harder to reason about. For example, if the state is "failed", the value of deadlineReached is irrelevant.
Modeling all relevant states with a single enum with cases with associated values is the solution to this typical problem. It makes it impossible to represent invalid states and makes it clearer to the consumer of the api how to use it correctly and what the meaning is.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/WireDriveFileUITracker.swift
WilhelmOks: Initially this triggered a compilation warning for the reasons that you mentioned, however making it `@MainActor` made the warnings disappear, so I'm assuming that with `@MainActor`, a race condition is excluded.

PR #4299, opened 2026-02-11
feat: re-establish EAR support - WPB-23473

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23473" title="WPB-23473" target="_blank">TaskWPB-23473</a>  [iOS] Refactor database key access for message encryption & migrations
  </td></table>
  <br />
 

### Issue

This PR re-establishes Encryption at Rest after it was broken by the introduction of dynamic background contexts. The implementation decouples encryption logic from `NSManagedObjectContext` and centralizes database key management through a service-based architecture.

The previous EAR implementation had critical issues:

1. **Missing keys in dynamic contexts** - Ad-hoc background contexts created via `CoreDataStack.newBackgroundContext()` lacked the database key, causing decryption errors and disappearing messages
2. **Migration failures** - EAR enable/disable operations create temporary contexts that were missing the database key, causing migration failure
3. **Initialization deadlock** - Setting the database key on the contexts was causing a deadlock on the view context.

### Changes

  - Created `EARMessageEncryptionService` for database key management and centralized encryption                                                                                                                     
  - Created `EARMigrator` for migrations when enabling / disabling EAR                                                                                                                   
  - Made `EARService` initialization async  
  - Split unit tests from integration tests with `EARServiceTests` and `EARServiceIntegrationTests`
  - Added doc comments
 
---

### Checklist

- [x] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [x] Description is filled and free of optional paragraphs.
- [x] Adds/updates automated tests.

PR #4434, opened 2026-03-13
fix: EAR Security Tests - WPB-23474

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23474" title="WPB-23474" target="_blank">TaskWPB-23474</a>  [iOS] Fix EAR user session security tests
  </td></table>
  <br />
 

### Issue

In  https://github.com/wireapp/wire-ios/pull/4299 I disabled some of the tests covering the `EARService` integration in the `ZMUserSession` because the asynchronous init of `EARService` wasn't compatible with the synchronous setup methods of `SessionManagerEncryptionAtRestMigrationTests` and `ZMUserSessionTests_EncryptionAtRest`.

To solve this issue, I've introduced a `EARServiceFactory` that keeps the capability of creating a `EARService` instance in an `async` manner for production code, while adding an interface that allows tests to create an `EARService` instance synchronously, coupled with a subsequent `async` setup method. This allows the synchronous setup methods to create the service.
 
---

### Checklist

### DIFF
--- a/WireDomain/Sources/WireDomain/Notifications/Components/NSEUserScope.swift
+++ b/WireDomain/Sources/WireDomain/Notifications/Components/NSEUserScope.swift
@@ -155,7 +155,7 @@ final class NSEUserScope: Component<NSEUserScopeDependency> {
             throw Failure.mainAppRequired(message: "no self client id")
         }
 
-        let earService = await EARService(
+        let earService = await EARServiceFactory().createEARService(
             accountID: accountID,
             coreDataStack: coreDataStack,
             sharedUserDefaults: dependency.sharedUserDefaults,
--- a/WireDomain/Sources/WireDomain/Notifications/Components/NSEUserScope.swift
+++ b/WireDomain/Sources/WireDomain/Notifications/Components/NSEUserScope.swift
@@ -155,7 +155,7 @@ final class NSEUserScope: Component<NSEUserScopeDependency> {
             throw Failure.mainAppRequired(message: "no self client id")
         }
 
-        let earService = await EARServiceFactory().createEARService(
+        let earService = await EARServiceFactory.createEARService(
             accountID: accountID,
             coreDataStack: coreDataStack,
             sharedUserDefaults: dependency.sharedUserDefaults,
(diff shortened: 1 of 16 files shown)


WireDomain/Sources/WireDomain/Notifications/Components/NSEUserScope.swift
```diff
@@ -155,7 +155,7 @@ final class NSEUserScope: Component<NSEUserScopeDependency> {
             throw Failure.mainAppRequired(message: "no self client id")
         }
 
-        let earService = await EARService(
```
netbe, line 158: nit-pic: this could be a static function

wire-ios-data-model/Source/Authentication/EAR/EARServiceFactory.swift
```diff
@@ -0,0 +1,125 @@
...
+            keyEncryptor: EARKeyEncryptor(),
+            coreDataStack: coreDataStack,
+            canPerformKeyMigration: canPerformKeyMigration,
+            earStorage: earStorage,
+            messageEncryptionService: messageEncryptionService,
+            migrator: migrator,
+            authenticationContext: authenticationContext
+        )
+
+        if !databaseContexts.isEmpty {
```
netbe: suggestion:  move to private method that would be used also for test. That way it s clearer that we do the same for tests and prod

netbe (approved):
Lgtm;) did/ could you run the security tests for this PR?

WireDomain/Sources/WireDomain/Notifications/Components/NSEUserScope.swift
samwyndham: Yeah I think I agree here. Seeing the factory, I would kind of assume that it exists for another reason - something like injecting the factory but only creating the service at a later date of for returning a mock version of the service in tests but that is not the case. 

The fact that `createEARService` is always called directly after creating the factory feels like this would be clearer as a static method on EARService.

samwyndham (approved):
Nice work!

David-Henner:
Ran all tests: https://github.com/wireapp/wire-ios/actions/runs/23237103009

PR #4435, opened 2026-03-13
fix: change the condition to close the call - WPB-24070

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24070" title="WPB-24070" target="_blank">TaskWPB-24070</a>  [iOS] Remove the temporary workaround for 1:1 calls over SFT
  </td></table>
  <br />
 

### Issue

1. Remove the [workaround](https://wearezeta.atlassian.net/wiki/spaces/PAD/pages/1314750477/2024-07-29+1+1+calls+over+SFT#Short-term-solution) of checking `aestab` (for MLS conversations, aestab goes from 1 to 0)
2. Fix the `isSelfAnActiveMember(...)` method and use `QualifiedID` as a parameter because the pending proposals may belong to a sub-conversation and we were passing the sub-conversation ID to this method and the result was empty because we don't store sub-conversations in the DB.

### Checklist

---


WireDomain/Sources/WireDomain/Repositories/Conversations/Repository/ConversationRepository.swift
```diff
@@ -381,17 +381,17 @@ public final class ConversationRepository: ConversationRepositoryProtocol {
...
-        await conversationsLocalStore.execute(identifier: groupID) { conversation, _ in
+        await conversationsLocalStore.execute(conversationID: conversationID) { conversation, _ in
             isSelfAnActiveMember = conversation?.isSelfAnActiveMember ?? false
         }
         return isSelfAnActiveMember
     }
 
     public func clearPendingProposals(in groupID: WireDataModel.MLSGroupID) async {
-        await conversationsLocalStore.execute(identifier: groupID) { conversation, context in
+        await conversationsLocalStore.execute(groupID: groupID) { conversation, context in
```
netbe: Issue: so when committing to a subgroup this will not get cleared since we cannot fetch by sub group id.

This might be ok as the side effect would be we try again to commit and CC sees nothing to commit just do a no op and clear the flag...

WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift
```diff
@@ -1330,11 +1330,21 @@ public final class ConversationLocalStore: ConversationLocalStoreProtocol {
...
         block: @escaping @Sendable (ZMConversation?, NSManagedObjectContext) -> Void
     ) async {
         await context.perform { [context] in
-            let conversation = ZMConversation.fetch(with: identifier, in: context)
+            let conversation = ZMConversation.fetch(with: groupID, in: context)
+            block(conversation, context)
+        }
+    }
+
+    public func execute(
```
netbe, line 1332: todo: add doc comment

WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift
```diff
@@ -1330,11 +1330,21 @@ public final class ConversationLocalStore: ConversationLocalStoreProtocol {
     }
 
     public func execute(
```
netbe: Todo: add comment about subgroup as it cannot find the conversation by subgroup id

netbe (commented):
There might be one issue with the fix see my comment, and let's talk about it

David-Henner (approved):

WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift
KaterinaWire: Added at the protocol level

WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift
KaterinaWire: Added at the protocol level

netbe (approved):
LGTM;)

suggestion: add subconversationType to CommitPendingProposalItem to help when looking at the logs

PR #4444, opened 2026-03-17
feat: add accessibility identifiers to wire drive - WPB-23902

OlivellaO:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23902" title="WPB-23902" target="_blank">TaskWPB-23902</a>  [iOS] Add missing accessibility elements in Wire drive
  </td></table>
  <br />
 

### Issue

Wire Drive was missing some identifiers that we could use for testing.

This PR adds some identifiers like:
- One for each Files menu item (Open, share, version history, renaming, tags, delete)
- One for each file (fileX, "X" being the index of the file on the list)
- SearchView
- Create folder
- Create File
- Recycle Bin
- Toggle Password
- Share Password
- Reset Password
- Save Password
- Share Link

On this PR some existing identifiers were updated to follow the same convention as the others (camelCase) to make it easier to implement tests and have a common naming for all the identifiers.

Some updated identifiers on FilesInfoView where:
- preparing-files-title -> preparingFilesTitle
- preparing-files-message -> preparingFilesMessage
- no-files-search-title -> noFilesSearchTitle
- no-files-search-message -> noFilesSearchMessage
- no-files-title -> noFilesTitle
- no-files-all-conversation-message -> noFilesAllConversationMessage
- no-files-message -> noFilesMessage
- error-title -> errorTitle
- error-message -> errorMessage
- filesBrowser.retryButton -> retryButton
- load-more -> loadMore

**Feedback is welcome to know if i'm missing some identifiers that could be useful to have or i might added some that are not need it 😃**

### Testing

Some old test using the existing identifiers might need to get updated in order to use the new camelCase naming.
This PR can't be test it manually with the app since it only adds identifiers for automated test. No change was made on the code so everything should works as in develop.

---

### Checklist

---

### DIFF
--- a/WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/Create/CreateFileView.swift
+++ b/WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/Create/CreateFileView.swift
@@ -95,7 +95,7 @@ private extension CreateFileView {
                 Text(L10n.Localizable.General.cancel)
             }
         )
-        .accessibilityIdentifier(Locators.WireDrive.CreateFilePage.cancelButton.rawValue)
+        .accessibilityIdentifier(Locators.WireDrive.CreateFilePage.cancelButton)
     }
 
     @ViewBuilder var createButton: some View {
@@ -111,7 +111,7 @@ private extension CreateFileView {
                 }
             )
             .disabled(viewModel.isCreateDisabled)
-            .accessibilityIdentifier(Locators.WireDrive.CreateFilePage.createButton.rawValue)
+            .accessibilityIdentifier(Locators.WireDrive.CreateFilePage.createButton)
         }
     }
 }
--- a/WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/Create/CreateFileView.swift
+++ b/WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/Create/CreateFileView.swift
@@ -18,6 +18,7 @@
 
 import SwiftUI
 import WireDesign
+import WireLocators
 import WireReusableUIComponents
 
 private typealias Strings = L10n.Localizable.Conversation.WireCells
@@ -94,7 +95,7 @@ private extension CreateFileView {
                 Text(L10n.Localizable.General.cancel)
             }
         )
-        .accessibilityIdentifier("cancelButton")
+        .accessibilityIdentifier(Locators.WireDrive.CreateFilePage.cancelButton.rawValue)
     }
 
     @ViewBuilder var createButton: some View {
@@ -110,7 +111,7 @@ private extension CreateFileView {
                 }
             )
             .disabled(viewModel.isCreateDisabled)
-            .accessibilityIdentifier("createButton")
+            .accessibilityIdentifier(Locators.WireDrive.CreateFilePage.createButton.rawValue)
         }
     }
 }
--- a/WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesContentView.swift
+++ b/WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesContentView.swift
@@ -190,7 +190,7 @@ private extension FilesContentView {
 
     var confirmButton: some View {
         Button(L10n.Localizable.General.confirm, action: {})
-            .accessibilityIdentifier(Locators.WireDrive.FilesContentPage.confirm.rawValue)
+            .accessibilityIdentifier(Locators.WireDrive.FilesContentPage.confirm)
     }
 }
 
@@ -237,7 +237,7 @@ private extension FilesContentView {
             placement: .navigationBarDrawer(displayMode: .always),
             prompt: Strings.Files.Search.title
         )
-        .accessibilityIdentifier(Locators.WireDrive.FilesContentPage.search.rawValue)
+        .accessibilityIdentifier(Locators.WireDrive.FilesContentPage.search)
     }
 }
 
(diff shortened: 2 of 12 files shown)


OlivellaO:
Feedback is welcome to know if i'm missing some identifiers that could be useful to have or i might added some that are not need it 😃

findms (approved):
lgtm 👍

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/Create/CreateFileView.swift
```diff
@@ -94,7 +95,7 @@ private extension CreateFileView {
                 Text(L10n.Localizable.General.cancel)
             }
         )
-        .accessibilityIdentifier("cancelButton")
+        .accessibilityIdentifier(Locators.WireDrive.CreateFilePage.cancelButton.rawValue)
```
WilhelmOks: Instead of using `.rawValue` everywhere, I suggest to make an extension like this:

```
extension View {
    func accessibilityIdentifier<T: RawRepresentable>(_ id: T) -> some View where T.RawValue == String {
        accessibilityIdentifier(id.rawValue)
    }
}
```

and then use the `.accessibilityIdentifier()` like this:
```
        .accessibilityIdentifier(Locators.WireDrive.CreateFilePage.cancelButton)
```

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/Item/FilesViewItemView.swift
```diff
@@ -273,6 +273,7 @@ struct FilesItemView: View {
     ) -> some View {
         if viewModel.menuActions.contains(itemAction) {
             menuItem(itemAction)
+                .accessibilityIdentifier(itemAction.rawValue)
```
WilhelmOks: I think this needs some name spacing because it could collide with other ids with the same name in this screen.
Maybe something like this:
```
    .accessibilityIdentifier("file_menu_\(itemAction.rawValue)")
```

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesContentView.swift
```diff
@@ -203,7 +206,7 @@ private extension FilesContentView {
         Task { await viewModel.loadMoreIfNeeded(index: index) }
     }
 
-    func loadMore() {
+    func loadMoreTask() {
```
WilhelmOks, line 209: What's the reason for renaming it? I think `loadMore` is a better name than `loadMoreTask`. For me, the latter name suggests that a task is returned, not performed/executed.
The fact that the action is wrapped in a Task doesn't change the semantics of `loadMore` for me.

WireUI/Sources/WireLocators/Locators.swift
```diff
@@ -329,5 +329,61 @@ public enum Locators {
             case cancelButton
             case removeFilterButton
         }
+
+        public enum ShareLinkPasswordPage: String {
```
WilhelmOks, line 333: Imo, we should not use the term "Page" to refer to those parts of the View.
Page is already reserved for the iOS concept of pages like in a `UIPageViewController` or the `TabView` with `.tabViewStyle(.page)`.
There is no official name for that, but I'd rather prefer "Screen" or completely omit it.

WilhelmOks (changes requested):

WireUI/Sources/WireLocators/Locators.swift
jullianm: @WilhelmOks This is a naming convention we have for other screens as well. Screen would probably be a better naming I agree although I would not handle it in this PR as it will imply many other changes and I think it's worth sharing the naming change suggestion with the rest of iOS team as well.

jullianm (commented):
Looks good so far but I think we're missing identifiers for some views:
- `FileVersioningView`
- `TagsEditView`
.. and the sort & filter views.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesContentView.swift
OlivellaO: To keep consistency. We have reloadTask, loadMoreIfNeededTask and the renamed loadMore (now loadMoreTask).

All 3 functions do the same (perform a task) so it make sense that their naming its also the same.

WireUI/Sources/WireLocators/Locators.swift
OlivellaO: Agree with both. I don't like Page either but i'm using it to keep the naming consistent.

WilhelmOks: as @jullianm pointed out, that seems to be an already established naming convention, so no need to change that here.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesContentView.swift
WilhelmOks: Ok, I see. But I'd rather rename the other ones to not use `Task` :)
Just a suggestion. Feel free to ignore.

WilhelmOks (approved):
Approved with suggestions

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/Create/CreateFileView.swift
OlivellaO: Done!

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesContentView.swift
OlivellaO: I'm keeping the Task suffix for now. I agree with you that maybe we should rename all three functions to avoid using Task as suffix.

jullianm (approved):
LGTM, as discussed the remaining identifiers will be done in a follow-up PR

PR #4450, opened 2026-03-18
fix(ear): messages can't be displayed - WPB-24136

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24136" title="WPB-24136" target="_blank">BugWPB-24136</a>  [iOS] Messages can't be displayed due to decryption errors
  </td></table>
  <br />
 

### Issue

Sometimes, with EAR enabled, after opening the app, some messages can't be displayed.

Screenshot 2026-03-18 at 11 53 24

### Cause

Messages couldn't be encrypted at rest because the database key was missing.

This happens because the app went to the background and locked the database (nullified the database key from cache) while the incremental sync was running and processing messages. 

When this happens, we cancel the sync https://github.com/wireapp/wire-ios/blob/48209b1ada766ebed55fb552e69d4de8d111136d/wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession%2BLifeCycle.swift#L46

However, the cancellation doesn't get propagated properly from the suspend method to the inner sync task. Hence, the sync continues and tries to process messages while the database key is nil

### Solution 

- Propagate task cancellation properly in `NonReentrantTaskManager` and `ExpiringActivity`
- ~Add a `Task.checkCancellation` before processing individual events in the incremental sync~
- Add a guard against having the database locked before processing any non-backgroundAccessible event, and handle cleaning up the processed events from the current batch

### Testing

1. Log in & Enable EAR
3. Receive many messages (use a spambot to send 100s of messages)
4. Open the app and immediately put it in the background once the sync starts
5. Reopen the app

---

### Checklist

### DIFF
--- a/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
+++ b/WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
@@ -360,28 +360,24 @@ public struct IncrementalSync: IncrementalSyncProtocol {
                 "fetched \(envelopes.count) stored envelopes for processing",
                 attributes: .incrementalSyncV2
             )
+            
+            let processed = try await processEventEnvelopes(envelopesWithObjectIDs: envelopesWithObjectIDs)
+            processedEnvelopeIDs.formUnion(processed)
+        }
 
-            for envelope in envelopes {
-                try Task.checkCancellation()
+        return processedEnvelopeIDs
+    }
+    
+    private typealias EnvelopeWithObjectID = (envelope: UpdateEventEnvelope, objectID: NSManagedObjectID)
 
-                for event in envelope.events {
-                    do {
-                        logger.debug(
-                            "processing pending event: \(event.name)",
-                            attributes: .incrementalSyncV2 + [.eventEnvelopeID: envelope.id]
-                        )
-                        try await processor.processEvent(event)
-                    } catch {
-                        logger.error(
-                            "failed to process stored event, dropping: \(error)",
-                            attributes: .incrementalSyncV2 + [.eventEnvelopeID: envelope.id]
-                        )
-                    }
-                }
-            }
-
-            processedEnvelopeIDs.formUnion(envelopes.map(\.id))
-            try await updateEventsStore.deleteNextPendingEvents(with: envelopesObjectIDs)
+    private func processEventEnvelopes(
+        envelopesWithObjectIDs: [EnvelopeWithObjectID]
+    ) async throws -> Set<UUID> {
+        
+        func finishProcessing(processedObjectIds: [NSManagedObjectID]) async throws {
+            if processedObjectIds.isEmpty { return }
+            
+            try await updateEventsStore.deleteNextPendingEvents(with: processedObjectIds)
             await updateEventsStore.calculateLastUnreadMessages()
 
             do {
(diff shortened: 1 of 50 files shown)


WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -362,6 +362,8 @@ public struct IncrementalSync: IncrementalSyncProtocol {
             )
 
             for envelope in envelopes {
+                try Task.checkCancellation()
```
johnxnguyen: issue: if we processed some events, then cancel, we won't delete any of the events already processed, so next time the sync happens some events will be processed again. I think in the case of new message events this is fine as we have tolerance to reprocess such events. But I'm not sure that all events are like this.

Though... now I'm wondering if this is the case since we don't hit the `databaseSaver.save()` line. What do you think?

David-Henner: [55ec7a0](https://github.com/wireapp/wire-ios/pull/4450/commits/55ec7a03b43307c15d541d6e242f96e361a8fd71)

In this commit:
- Added logic to handle the interruption of event processing, in order to delete processed events and save the database.
- Replaced `Task.checkCancellation` by throwing when the database is locked and the envelope isn't background accessible. The reasoning is that if there's an ongoing call when going to the background, we don't cancel the sync, so the database key would be removed from cache but we'd continue processing events that require it

I think it starts to make the sync too complex with special cases for EAR, and another idea I had was to systematically cancel the sync when going to the background. And, in case there's an ongoing call, start a call events only sync. 

What do you think ?

wire-ios-system/Source/ExpiringActivity.swift
```diff
@@ -60,34 +60,38 @@ actor ExpiringActivityManager {
...
-                            try await self.stopWork()
-                        } catch {
-                            continuation.resume(throwing: error)
+                        }
+                        semaphore.wait()
+                    } else {
+                        WireLogger.backgroundActivity.warn("Background activity is expiring: \(reason)")
+                        Task {
+                            do {
+                                try await self.stopWork()
```
samwyndham, line 85: issue: I know this is not from this PR, but in this code path if stopWork does not throw, withExpiringActivity will never complete.

wire-ios-system/Source/ExpiringActivity.swift
```diff
@@ -60,34 +60,38 @@ actor ExpiringActivityManager {
...
-                            try await self.startWork(block: block, semaphore: semaphore).value
-                            WireLogger.backgroundActivity.debug("Expiring activity completed: \(reason)")
-                            continuation.resume()
-                        } catch {
-                            WireLogger.backgroundActivity.warn("Expiring activity ended with an error: \(error)")
-                            continuation.resume(throwing: error)
-                        }
+        try await withTaskCancellationHandler {
+            try await withCheckedThrowingContinuation { continuation in
+                api.performExpiringActivity(withReason: reason) { expiring in
```
samwyndham, line 65: suggestion: I think we can greatly simplify this code and remove the need for the child tasks and the issue I highlighted in my other comment + the need for the semaphore which is problematic.

I suggest creating a new method that just changes `performExpiringActivity` to an async method without any additional behaviour:

```swift
extension ExpiringActivityInterface {
    func performExpiringActivity(withReason reason: String) async -> Bool {
        await withUnsafeContinuation { continuation in
            performExpiringActivity(withReason: reason) { result in
                continuation.resume(returning: result)
            }
        }
    }
}
```

Then everything can be done a lot more simply:

```swift
    func withExpiringActivity(reason: String, block: @escaping () async throws -> Void) async throws {
        let task = Task {
            let isExpiring = await api.performExpiringActivity(withReason: reason)
            // etc
```

samwyndham: Sorry this comment is nonsense because I didn't understand the purpose of api. I was treating it like something to be done before the block but that is not the case.

johnxnguyen (approved):

samwyndham (commented):
Nice work

samwyndham (approved):

PR #4451, opened 2026-03-18
fix: moving cacheKey generation to a single place - WPB-24133

WilhelmOks:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24133" title="WPB-24133" target="_blank">BugWPB-24133</a>  [iOS] Downloaded files need download again after app restart
  </td></table>
  <br />
 

### Issue

When a Drive file, which was already downloaded is tapped after an app restart, it starts downloading again.

This is caused by the fact that the cacheKey is being built differently in two places.
Those two places diverged after WPB-23112.
With this PR I moved to logic into one single place and added a doc comment.

### Testing

* tap on a Drive file to start the download
* wait for the download to finish
* force close the app (not just background mode)
* open app again
* tap on the previously downloaded file to open it

Old behaviour before fix:
* The file starts downloading again

New behaviour after fix:
* The file opens

### DIFF
--- a/WireMessaging/Sources/WireMessagingData/WireDrive/Model/WireDriveLocalAsset+CacheKey.swift
+++ b/WireMessaging/Sources/WireMessagingData/WireDrive/Model/WireDriveLocalAsset+CacheKey.swift
@@ -26,7 +26,7 @@ extension WireDriveLocalAsset {
     /// The file name of the URL path will be shown as the title in the QuickLook preview when the file os opened,
     /// so it's important to make it readable and correspond to the actual file name, rather than a generated id.
     static func cacheKey(nodeID: UUID, eTag: String, path: String) -> String {
-        let filename = path.split(separator: "/").last.flatMap(String.init) ?? "-"
+        let filename = URL(fileURLWithPath: path).lastPathComponent
         return "\(nodeID.uuidString)-\(eTag)/\(filename)"
     }
 
--- a/WireMessaging/Sources/WireMessagingData/WireDrive/Model/WireDriveLocalAsset+CacheKey.swift
+++ b/WireMessaging/Sources/WireMessagingData/WireDrive/Model/WireDriveLocalAsset+CacheKey.swift
@@ -21,12 +21,13 @@ import WireMessagingDomain
 
 extension WireDriveLocalAsset {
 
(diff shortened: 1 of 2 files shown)


OlivellaO (approved):

samwyndham (approved):
LGTM

jullianm (approved):
LGTM 👍

WireMessaging/Sources/WireMessagingData/WireDrive/Model/WireDriveLocalAsset+CacheKey.swift
```diff
@@ -21,12 +21,13 @@ import WireMessagingDomain
...
+    /// This cache key is also used to build the URL which points to the downloaded file.
+    /// The file name of the URL path will be shown as the title in the QuickLook preview when the file os opened,
+    /// so it's important to make it readable and correspond to the actual file name, rather than a generated id.
     static func cacheKey(nodeID: UUID, eTag: String, path: String) -> String {
-        var result = "\(nodeID.uuidString)-\(eTag)"
-        if let pathExtension = URL(string: path)?.pathExtension, !pathExtension.isEmpty {
-            result.append(".\(pathExtension)")
-        }
-        return result
+        let filename = path.split(separator: "/").last.flatMap(String.init) ?? "-"
```
JaCaLlaSchwarz: Suggestion:

let filename = URL(fileURLWithPath: path).lastPathComponent

WilhelmOks: Thanks for the suggestion.
I chose to extract the filename manually because of two reasons:
* we had issues with this URL-path api in the past, producing percent encoding.
* the path to the file is not a standard url file path. It's some custom format with `@` in it and information about the conversation. I don't know if it would work, but even if it would, it might be risky and fail for some special cases.

WilhelmOks: Since the previous code was using the URL api as well, but with `URL(string: )` rather than `URL(fileURLWithPath: )`, I tried it with `URL(fileURLWithPath: path)` and it works and doesn't produce percent encoding.
Updated the code.

PR #4457, opened 2026-03-18
feat: handle cancellation when downloading asset - WPB-23688

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23688" title="WPB-23688" target="_blank">TaskWPB-23688</a>  [iOS] Opening & Loading behavior
  </td></table>
  <br />
 

### Issue

As part of the loading & opening behavior feature, user can now cancel a file being downloaded.

This PR adjusts the logic to handle task cancellation from a parent task (in view model) and ensures the cancellation is propagated to the child task (repository) **eventually** ensuring the URL task from the URLSession is also properly cancelled (url session delegate).

To achieve that, we needed to adjust the code a bit and use the dedicated `withTaskCancellationHandler` API.

### Testing

- upload a rather large file
- tap to start downloading, progress will show up
- at any time (before downloading finishes), tap to cancel
- view goes back to initial state.

### Checklist

---


WireMessaging/Sources/WireMessagingUI/WireDrive/Components/AttachmentsPreviewView/WireDriveAttachmentsPreviewItemViewModel.swift
```diff
@@ -259,7 +274,7 @@ final class WireDriveAttachmentsPreviewItemViewModel: ObservableObject {
 
     private var canOpen: Bool {
```
jullianm, line 269: from that point on, the changes are only related to lint and format.

WireMessaging/Sources/WireMessagingData/WireDrive/WireDriveLocalAssetRepository.swift
```diff
@@ -122,25 +131,30 @@ package final class WireDriveLocalAssetRepository: WireDriveLocalAssetRepository
...
                 asset.fileSize = fileSize
                 try store.upsertAsset(asset)
             }
-            
+
             timerTask.cancel()
 
-            let (tempURL, _) = try await download.value
+            if Task.isCancelled {
+                urlDownloadTask.cancel()
```
jullianm, line 144: this is important to explicitly cancel the download task here too. If we don't do so, UI wise it look like the file stopped from being downloaded but it actually continues (while debugging I made sure of that by checking the inner `URLSessionTask` stopped emitting progress values (see `URLSessionTaskProgressDelegate` where the final cancellation occurs).

WilhelmOks (commented):
I think the approach with leaving the task handle in the viewmodel isn't optimal.
We will probably lose the handle when we close the screen after starting a download and then come back to it again to cancel the download. I'm suspecting that the cancellation won't work.

Can you change it so that the download task handle is managed by the use case or the repository instead?
That way we won't lose it and we can use the existing cancelDownload() from the repository.
Also, if we don't use this existing method, we'd probably need to remove it because it's not working and will only lead to confusion about which is the correct method to cancel a download.

jullianm:
@WilhelmOks your suggestion makes sense, it's probably best to handle that cancellation directly from the usecase/repository and remove the extra Task cancellation operation from the view model. I will adjust it in a follow-up commit.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/Item/FilesViewItemView.swift
```diff
@@ -119,9 +119,9 @@ struct FilesItemView: View {
     @ViewBuilder
     private func icon() -> some View {
         let state: WireDriveFileUITracker.State = viewModel.fileTracker.state
-        //let state: WireDriveFileUITracker.State = .downloading(progress: 0.3, isLargeFile: false)
-        //let state: WireDriveFileUITracker.State = .failed(error: URLError(.badURL))
-        //let state: WireDriveFileUITracker.State = .downloaded(showReadyToOpen: true)
+        // let state: WireDriveFileUITracker.State = .downloading(progress: 0.3, isLargeFile: false)
+        // let state: WireDriveFileUITracker.State = .failed(error: URLError(.badURL))
+        // let state: WireDriveFileUITracker.State = .downloaded(showReadyToOpen: true)
```
OlivellaO, line 124: There is a reason why we need to keep all this comments?
I know they where not added on your PR, it's just a formatting change, but still if there is no reason to keep them we can remove them.

OlivellaO (approved):

WilhelmOks (approved):

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/Item/FilesViewItemView.swift
jullianm: Yes, we'll make sure they're removed in the main PR 👍

WilhelmOks: yup, definitely should be removed before we make the final PR

WilhelmOks: done in branch `feat/WPB-23688-opening-and-loading-ui`

PR #4471, opened 2026-03-23
fix: block team member - WPB-21260

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21260" title="WPB-21260" target="_blank">BugWPB-21260</a>  [iOS] Option to block teammate visible but without effect
  </td></table>
  <br />
 

### Issue

Currently the option to block a team member was showing during long press on conversation in conversationList.

We were checking the teamID of the conversation which might have been fine with old proteus conversations, but now 1-1 conversations don't have the teamID set.

Check explicitly if the other user in the oneOnOne is from a different team.

### Testing

1. login
2. long press a 1:1 within the same team 
-> no block option
3. long press a 1:1 within another team 
-> block option should show

---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationAction.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationAction.swift
@@ -65,7 +65,7 @@ extension ZMConversation {
         var actions = [Action]()
         actions.append(contentsOf: availableStandardActions())
         actions.append(.clearContent)
-        if teamRemoteIdentifier == nil, let connectedUser {
+        if let connectedUser, connectedUser.teamIdentifier != ZMUser.selfUser()?.teamIdentifier {
             actions.append(.block(isBlocked: connectedUser.isBlocked))
         }
         return actions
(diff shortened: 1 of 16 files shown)


jullianm (approved):
LGTM

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationAction.swift
```diff
@@ -65,7 +65,7 @@ extension ZMConversation {
         var actions = [Action]()
         actions.append(contentsOf: availableStandardActions())
         actions.append(.clearContent)
-        if teamRemoteIdentifier == nil, let connectedUser {
+        if let connectedUser, connectedUser.teamIdentifier != ZMUser.selfUser()?.teamIdentifier {
```
KaterinaWire: **question:** what if both users are not team members, should they be able to block each other?

KaterinaWire (approved):

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationAction.swift
netbe: good catch pushing a fix

PR #4472, opened 2026-03-23
fix: critical flow fixes - WPB-24211

findms:
### Issue
**Some intermittent failures observed recently**

### Testing

Ran locally
image


wire-ios/WireUITests/Pages/FirstTimePage.swift
```diff
@@ -70,9 +70,16 @@ class FirstTimePage: PageModel {
     func acceptPopup(with testCase: XCTestCase) throws -> ConversationsPage {
         handleNotificationPermissionAlert(testCase: testCase)
         // Sometimes the OK button take longer to disappear after tapping.
-        if conversationsButton.exists || conversationsButton.waitForExistence(timeout: 10) {
-            conversationsButton.tap()
+        guard conversationsButton.waitForExistence(timeout: 15) else {
```
findms, line 64: Sometimes, in few days once observed where tapping button keep loading - not sure if due to backend is slow that time or some app issue but for now increasing the timeout in worst case as frequency is very low of this.

wire-ios/WireUITests/Pages/PhotosAppPage.swift
```diff
@@ -73,7 +81,8 @@ class PhotosAppPage: PageModel {
     }
 
     @discardableResult
-    func openFirstImage() throws -> PhotosAppPage {
+    func openFirstImage(with testCase: XCTestCase) throws -> PhotosAppPage {
+        handleNotificationPermissionAlert(testCase: testCase)
```
findms: In recent run 2 times observed that photos app notification alert is being shown which is new which is showing up so handled it here.

wire-ios/WireUITests/Pages/FirstTimePage.swift
netbe: question: so what are we talking about ? first initial launch that would take 15 seconds to launch?
I'd be careful with this as it could hide another issue more critical.

So i would suggest to have the timeout <= "time to do initial sync". Not sure 15 seconds is what we expect

wire-ios/WireUITests/Pages/PhotosAppPage.swift
netbe: question: hmm weird, if you run the test individually, fresh install does it still happen here?

netbe (commented):
The fixes look good, but I am a bit afraid to approve without more information about what goes wrong exactly.

Extracting the logs from the run should be done so we're sure we are not missing some issue.

wire-ios/WireUITests/Pages/PhotosAppPage.swift
findms: No with fresh run also,this works fine on photos app as well.

wire-ios/WireUITests/Pages/FirstTimePage.swift
findms: no..not everytime..out of the blue sometimes when user perform login, tap on OK -> loader keep showing up for more than 10 seconds which is not on local case, could be due the first test which is more heavy lifting to do ..etc..
I will check with you offline how I can fetch the logs for this delay.

findms: @netbe  this is what I got from datadog:
`FirstTimePage.swift:72: Failed to tap "bottomBarRecentListButton" Button: No matches found for Elements matching predicate '"bottomBarRecentListButton" IN identifiers' from input {(
    Button, identifier: 'okButton', label: 'OK', value: 1, Disabled
)}
Possibly caused by runtime issues:
Automation type mismatch: computed Button from legacy attributes vs ActivityIndicator from modern attribute. Input attributes and values: {
    "XC_kAXXCAttributeAutomationType" = 36;
    "XC_kAXXCAttributeElementBaseType" = UIResponder;
    "XC_kAXXCAttributeElementType" = "SwiftUI.AccessibilityNode";
    "XC_kAXXCAttributeTraits" = 8589935361;
}
See test report attachments for more detail.`

findms: For the notification, Found a case where addUIInterruptionMonitor was returning true no matter if notification alert is shown or not. Fixed it in latest commit.

wire-ios/WireUITests/FederationTests.swift
```diff
@@ -60,6 +60,7 @@ final class FederationTests: WireUITestCase {
             .acceptConnectionRequest()
 
         // THEN
+        XCTAssertTrue(activeConversationPage.classifiedBanner.exists)
```
findms, line 63: This additional assert to buy some time as it taps back button so quickly that it remains on same page and fails to find conversationsPage

wire-ios/WireUITests/Pages/PhotosAppPage.swift
samwyndham: suggestion: I'm thinking it is likely that this isn't tied to opening the image but rather test run order or some race condition. Do we generally always want to allow? If yes how about adding the handler in the `WireUITestCase` setup method. That way it will work for any test right?

samwyndham (approved):
Nice work

wire-ios/WireUITests/Pages/PhotosAppPage.swift
findms: Good suggestion. removed notificationHandler from specific pages to setup in `WireUITestCase`

jullianm (approved):
Looks good

PR #4485, opened 2026-03-25
chore: do not check blacklist when loading user session - WPB-24175

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24175" title="WPB-24175" target="_blank">TaskWPB-24175</a>  [iOS] Don't check whether build is blacklisted as part of loading user session
  </td></table>
  <br />
 

### Issue

When loading a user session we currently check the blacklist. Under poor networking conditions this can lead to poor UX - a user might have to wait 60 seconds on the loading screen before a time out occurs and they are navigated to the conversation screen.

This PR:

- Removes this check during loading. We already have wiring in place for this check to be done at other times.
- Introduces `CheckBlacklistWorker` which intermittently checks the blacklist and cause the block view to be shown if the build is blacklisted for a particular logged in server. The app does a check just after loading the users session then again every 6 hours. It also does re-checks if those have failed (based on network condition changes and every 30 min).
- Removes the `checkBuildBlacklistAction` recurring action as this is handled by the above worker
- Adds a UI test for the basic case of the checklist showing after login.
- Fixes a race condition that could cause the user to go to the authenticated app state from the blacklist app state.

### Testing

1. Login
2. Set build number to `0` via Developer menu -> Overrides & quit the app.
3. Restart
4. Update required view should display

---

### Checklist


WireFoundation/Sources/WireFoundation/WirePrimitives/UITestConfig.swift
```diff
@@ -0,0 +1,56 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// Configuration values for UITests.
+public struct UITestConfig: Codable {
```
samwyndham, line 22: As discussed with @findms & @johnxnguyen here I introduce a new way of sharing launch environment data from the tests to the app. It is a JSON object so very easy to add to and is in `WireFoundation` so can be accessed easily in different targets. The key property `environment` which is used to get the launch environment values is `#if DEBUG` to avoid test code being shipped to production.

WireDomain/Sources/WireDomain/Synchronization/CheckBlacklistWorker.swift
```diff
@@ -0,0 +1,134 @@
...
+            onIsBuildBlacklisted: onIsBuildBlacklisted
+        )
+    }
+
+    deinit {
+        triggerTask?.cancel()
+    }
+
+    public nonisolated func start() {
+        Task { await self.startAndWait() }
```
netbe, line 53: question: should we retain the task here and cancel it on stop?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -580,6 +590,7 @@ public final class ZMUserSession: NSObject {
         startEphemeralTimers()
         restoreDebugCommandsState()
         configureRecurringActions()
+        checkBlacklistWorker.start()
```
netbe: question: does the worker gets properly stopped when session is teardown?

netbe (approved):
LGTM;) only point is about the worker being stopped

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: question: this looks like a side effect, so can we move this to the user sessions `start` method?

wire-ios/WireUITests/BlacklistTests.swift
```diff
@@ -16,25 +16,26 @@
...
+    override func setUpWithError() throws {
+        uiTestConfig.isBuildBlacklisted = true
 
-    /// The name of the associated Core Data entity.
+        try super.setUpWithError()
+    }
 
-    public static let entityName = "AppInfo"
+    @MainActor
+    func test_blacklistAfterLogin__TC_9483() async throws {
```
johnxnguyen: praise: nice to see!

WireDomain/Sources/WireDomain/UseCases/IsBuildBlacklistedUseCase.swift
```diff
@@ -18,11 +18,12 @@
...
 import WireFoundation
-import WireNetwork
+@preconcurrency import WireNetwork
 
-public protocol IsBuildBlacklistedUseCase {
+// sourcery: AutoMockable
+public protocol IsBuildBlacklistedUseCase: Sendable {
 
-    func invoke() async -> Bool
+    func invoke() async -> (isBuildBlacklisted: Bool, error: Error?)
```
johnxnguyen, line 26: question: why doesn't it throw instead of returning an optional error?

WireDomain/Tests/WireDomainTests/Synchronization/CheckBlacklistWorkerTests.swift
```diff
@@ -0,0 +1,138 @@
...
+            isBuildBlacklistedUseCase: mockUseCase,
+            trigger: trigger,
+            onIsBuildBlacklisted: { callbackCounter.count += 1 }
+        )
+    }
+
+    // MARK: - Tests
+
+    @Test
+    func `calls use case on trigger`() async {
```
johnxnguyen, line 54: TiL: \`you can name methods like this\`

johnxnguyen (commented):
Looks good, just some questions.

WireDomain/Sources/WireDomain/UseCases/IsBuildBlacklistedUseCase.swift
samwyndham: @johnxnguyen I didn't want to dilute the current behavior of IsBuildBlacklistedUseCase.swift and open the possibility for users of IsBuildBlacklistedUseCase to misinterpret it's results:

```swift
// As per specs, if there is any failure obtaining the blacklist,
// whether it is missing, can't be decoded, or otherwise, then
// consider it empty (i.e all clients valid).
```

wire-ios/WireUITests/BlacklistTests.swift
findms: nitpic: to make it uniform `func testBlacklistAfterLogin_TC_9483() async throws {`

findms (approved):
LGTM 👍
@samwyndham  Would be great to add a screenshot how it went on local run.

WireDomain/Sources/WireDomain/Synchronization/CheckBlacklistWorker.swift
samwyndham: This method is only a wrapper.  The actual task gets started and retained in `startAndWait()`

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -580,6 +590,7 @@ public final class ZMUserSession: NSObject {
...
```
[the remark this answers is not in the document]
samwyndham: Fixed in https://github.com/wireapp/wire-ios/pull/4485/commits/338e9db7e7f08f5b3002dad737858bd93d5fefa3

wire-ios/Wire-iOS/Sources/AppStateCalculator.swift
```diff
@@ -323,9 +334,7 @@ extension AppStateCalculator: SessionManagerDelegate {
     }
 
     func sessionManagerDidReportLockChange(forSession session: UserSession) {
-        if session.isBuildBlacklisted {
-            transition(to: .blacklisted(reason: .appVersionBlacklisted))
-        } else if session.isLocked {
+        if session.isLocked {
             transition(to: .locked(session))
         } else {
             transition(to: .authenticated(session))
```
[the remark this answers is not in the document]
samwyndham: This check is now done in `func validAppState(from appState: AppState)` so that it handles more situations.

wire-ios/WireUITests/BlacklistTests.swift
samwyndham: Updated in https://github.com/wireapp/wire-ios/pull/4485/commits/18f2aacb4ce7b3ed9aec1e3c7af8420845a6b3ee

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
samwyndham: Moved in https://github.com/wireapp/wire-ios/pull/4485/commits/a7631a9aa5d740deb31457ee111a07a1802ef28e

samwyndham: Good catch. Fixed in https://github.com/wireapp/wire-ios/pull/4485/commits/338e9db7e7f08f5b3002dad737858bd93d5fefa3

johnxnguyen (approved):

PR #4489, opened 2026-03-25
fix: sent indicator way off on shared file (image/video) - WPB-23929

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23929" title="WPB-23929" target="_blank">BugWPB-23929</a>  [iOS][Files in conversation] Send indicator appears much below the video in conversation for the sender
  </td></table>
  <br />
 

### Issue

This PR fixes a layout issue on some shared file (image/video) where the height of the custom `FlowLayout` is not properly computed resulting in a sent indicator looking way off (see screenshots)

**BEFORE**

Capture d’écran   2026-03-25 à 11 50 22

Capture d’écran   2026-03-25 à 11 49 50

**AFTER**

Capture d’écran   2026-03-25 à 11 52 46
Capture d’écran   2026-03-25 à 11 51 52

### Testing

_Describe how to test._

---

### Checklist

---

### DIFF
--- a/WireMessaging/Sources/WireMessagingUI/FlowLayout.swift
+++ b/WireMessaging/Sources/WireMessagingUI/FlowLayout.swift
@@ -49,7 +49,7 @@ struct FlowLayout: Layout {
 
     private func layout(subviews: Subviews, containerWidth: CGFloat) -> (items: [CGRect], size: CGSize) {
         let sizes = subviews.map { subview in
-            let size = subview.sizeThatFits(.unspecified)
+            let size = subview.sizeThatFits(ProposedViewSize(width: containerWidth, height: nil))
             return CGSize(width: min(size.width, containerWidth), height: size.height)
         }
 


samwyndham (approved):
@jullianm Please check that this doesnit break other places that the flow layout is used are not broken - I'm thinking especially sending/receiving multiple files.

WilhelmOks (approved):

WilhelmOks:
> @jullianm Please check that this doesnit break other places that the flow layout is used are not broken - I'm thinking especially sending/receiving multiple files.

That's a good point.

I was curious and checked where FlowLayout is being used.

It's not used by the file upload (drafts). Instead, it uses an HStack within a horizontal ScrollView.

The other places where it's used are only the two Views where we show the tags. I checked it in the SwiftUI Preview and it looked fine.

Interestingly, we also have unit tests for the FlowLayout. But sadly they didn't catch the issue with the wrong size calculation.
Maybe it would be a good idea to add a unit test to check for this and prevent that kind of regression in the future.

jullianm:
@samwyndham @WilhelmOks checked the various places where flow layout is used and it looks fine.

PR #4491, opened 2026-03-25
fix: attachment preview header layout issues - WPB-23931

WilhelmOks:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23931" title="WPB-23931" target="_blank">BugWPB-23931</a>  [iOS][Files in conversation] Remove top padding from video and files thumbnails 
  </td></table>
  <br />
 

### Issue

In the file attachment header, the space between the first line (file icon and file type) and the second line (file name) was too large.
The issue was that in many places the size was set to take up maximum space, either through `maxHeight: .infinity` or through `Spacer()`. Then to reduce the height, an explicit height was set (`height: 74`). 
This not only caused the layout issues but also broke the UI when the font size was set to very large.
I fixed it by removing the indefinite growing and the explicit heights.

Additionally I added a gray background and some padding in SwiftUI Previews to better see if there are issues with the layout.

### Testing

Send messages with attachments of different files and check if the header layout is ok.
Especially the space between the file type and the file name should not be too large.
Most important to check are video files and documents.


jullianm (commented):
It does fix the issue in conversation previews however it also changes the layout (height) of draft previews which might be acceptable but I think it's better to keep the draft previews as it was before

WilhelmOks:
> It does fix the issue in conversation previews however it also changes the layout (height) of draft previews which might be acceptable but I think it's better to keep the draft previews as it was before

In the latest commit I changed it for the draft previews to keep the same bubble height as the other elements in the list

jullianm (approved):
Looks good!

caldrian (approved):

PR #4494, opened 2026-03-25
fix: can't receive message in new groups after updating certificate - WPB-24195

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24195" title="WPB-24195" target="_blank">BugWPB-24195</a>  [iOS] User did not receive message in a group until they sent one after updating a certificate
  </td></table>
  <br />
 

### Issue

- User A logs in and enrolls a new E2EI certificate
- User A updates their E2EI certificate, from device details
- User B adds user A to a new conversation and sends messages

**Result:** User A sees the conversation but cannot see the messages

### Cause

When user B adds user A to a new conversation, user A receives a welcome message. Upon processing the welcome message, core crypto throws an `OrphanWelcome` error. This causes the group to fail to setup and any messages received in that group to fail to be decrypted.

To receive an `OrphanWelcome` error means that the key package user B used to add user A to the group was not recognized by core crypto when the welcome message was processed. 

Upon investigation, we discovered that CC's `clientKeyPackages` method was returning existing key packages that match the ciphersuite and credential type, but these KPs do not match the new x509 credentials. Hence, CC returns stale key packages that we end up uploading to the backend.

Then, when calling `deleteStaleKeyPackages`, these stale key packages are deleted from local storage.

Finally, when user B creates a group with user A and fetches a key package, a stale key package is returned and used for the welcome message that user A's client fails to process, since the stale key packages were deleted in the last step. 

### Solution

Change the order of the key package rotation flow, so that we first call `deleteStaleKeyPackages` to delete the key packages that don't match the new `x509` credentials, and force the `clientKeyPackages` method to generate new key packages.

---

### Checklist


typfel (approved):

netbe (approved):
LGTM:)

1. do we need to update any documentation regarding this?
2. do we need to add a test too?

PR #4519, opened 2026-03-31
fix: cannot be added to group after MLS is enabled - WPB-24496

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24496" title="WPB-24496" target="_blank">BugWPB-24496</a>  [iOS] Can not add user from iOS to group on another client after enabling MLS
  </td></table>
  <br />
 

### Issue

- User A logs in
- MLS is enabled for user A's team
- User B adds user A to a group

Result: user B fails to add user A to the group

### Causes

When user B adds user A to a group, they fail to fetch their key packages.

This happens because user A's client attempts to generate and upload key packages after the incremental sync completes. However, at this time MLS is not enabled and the operation fails.

If the client triggers another incremental sync once MLS is enabled, we should attempt to upload key packages again, as long as it has been more than 24h since the last time we checked. Since the last attempt failed recently, the 24h check won't pass and we won't retry to upload key packages.

### Solutions

1. Check if MLS is enabled before trying to upload key packages
2. Only update the timestamp if there were already enough key packages or if the replenishment succeeded

---

### Checklist

### DIFF
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -601,7 +601,7 @@ public final class MLSService: MLSServiceInterface {
         func logWarn(abortedWithReason reason: String) {
             logger.warn("aborting key packages upload: \(reason)")
         }
-        
+
         guard await featureRepository.fetchMLS().isEnabled else {
             return logWarn(abortedWithReason: "MLS is not enabled")
         }
@@ -640,7 +640,7 @@ public final class MLSService: MLSServiceInterface {
                 context: context.notificationContext
             )
             logger.info("success: uploaded key packages for client \(clientID)")
-            
+
             userDefaults.set(.now, forKey: .keyPackageQueriedTime)
         } catch {
             logger.warn("failed to upload key packages for client \(clientID). \(String(describing: error))")
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -602,10 +602,6 @@ public final class MLSService: MLSServiceInterface {
             logger.warn("aborting key packages upload: \(reason)")
         }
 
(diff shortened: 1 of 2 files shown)


wire-ios-data-model/Tests/MLS/MLSServiceTests.swift
```diff
@@ -1567,6 +1567,32 @@ final class MLSServiceTests: ZMConversationTestsBase, MLSServiceDelegate {
...
+
+        // these should not be called
+        mockActionsProvider.countUnclaimedKeyPackagesClientIDCiphersuiteContext_MockMethod = { _, _, _ in
+            XCTFail("shouldn't count key packages when MLS is disabled")
+            return 0
+        }
+
+        mockActionsProvider.uploadKeyPackagesClientIDKeyPackagesContext_MockMethod = { _, _, _ in
+            XCTFail("shouldn't upload key packages when MLS is disabled")
+        }
```
samwyndham: suggestion: Assert the invocations is empyt. It will be simpler code I think:

```swift
XCTAssertTrue(
    mockActionsProvider.countUnclaimedKeyPackagesClientIDCiphersuiteContext_Invocations.isEmpty),
    "shouldn't count key packages when MLS is disabled"
)
```

samwyndham (approved):
LGTM

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -602,6 +602,10 @@ public final class MLSService: MLSServiceInterface {
             logger.warn("aborting key packages upload: \(reason)")
         }
 
+        guard await featureRepository.fetchMLS().isEnabled else {
```
netbe: suggestion: move to shouldQueryUnclaimedKeyPackagesCount

netbe (approved):
LGTM, left a suggestion

PR #4532, opened 2026-04-02
fix: infinite sync after MLS got enabled while logged out - WPB-24543

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24543" title="WPB-24543" target="_blank">BugWPB-24543</a>  [iOS] Infinite sync re-logging in after MLS is enabled
  </td></table>
  <br />
 

### Issue
   
After logging out and back in on a team where MLS was enabled while the user was logged out, the incremental sync enters an infinite sync loop because of an error:

```
failedToFetchStoredEvents(
   WireDomain.EAREncryptionHelper.CryptoError.decryptionFailed(
      Error Domain=NSOSStatusErrorDomain Code=-50 "ECIES: Failed to aes-gcm decrypt data (err -69)"
      UserInfo={numberOfErrorsDeep=0, NSDescription=ECIES: Failed to aes-gcm decrypt data (err -69)}
   )
)
```

The sync eventually stops thanks to the `BackoffRetrier` and an alert is shown with the error above. Afterwards we also observe a few EAR decryption errors.

### Causes

**A — Zombie websocket stores events with stale EAR keys.**
`ZMUserSession.tearDown()` set `syncAgent = nil` without calling `suspend()` first. This left the websocket alive after logout. When the user logged back in, a live `user.clientAdd` event arrived on this zombie connection and was encrypted into the event store using EAR public keys that had been fetched at the start of the previous session's `IncrementalSync`.

**B — EAR keys were not deleted on logout.**
`Account.deleteKeychainItems()` removed the app lock passcode and CoreCrypto keys but not EAR keys. On re-login, `clientRegistrationDidSucceed` called `enableEncryptionAtRest`, which deleted the old EAR keys and generated new ones. Any events already stored with the old public key became permanently undecryptable.

**In short:** the zombie websocket stores a freshly arrived event using old EAR keys, then key rotation destroys those keys, leaving the stored event orphaned.

### Solutions

- **Close the websocket on logout** by calling `suspend()` on the sync agent before teardown, ensuring the push channel and its captured EAR public keys are released.
- **Delete EAR keys on account cleanup** by adding `RemoveEARKeysUseCase` to `Account.deleteKeychainItems()`, ensuring no orphaned EAR keys survive in the keychain after logout.

### Changes Made

- **`SyncAgent`**: Added `tearDown()` method that nils the delegate and fires a task to call `suspend()`, which cancels the ongoing sync and closes the push channel.
- **`ZMUserSession.tearDown()`**: Replaced `syncAgent?.delegate = nil` with `syncAgent?.tearDown()`.
- **`RemoveEARKeysUseCase`** *(new)*: Use case that deletes all five EAR keychain entries (primary/secondary public, primary/secondary private, database key) for a given account ID.
- **`Account.deleteKeychainItems()`**: Added `RemoveEARKeysUseCase().invoke(accountID:)` alongside the existing CoreCrypto key cleanup.
- **Tests**: Added `RemoveEARKeysUseCaseTests`, `EARKeyRepositoryTests`, extended `AccountTests` and `SyncAgentTests`.                                                                                                                                    
                
### Testing

- login with proteus user on iOS and any other client
- create some groups and / or 1:1s
- logout the iOS user
- enable MLS for the team
- login the iOS user
- Try to send a message in any conversation or
- try to add the user to a group

---

### Checklist

### DIFF
--- a/wire-ios-data-model/Source/UseCases/RemoveEARKeysUseCase.swift
+++ b/wire-ios-data-model/Source/UseCases/RemoveEARKeysUseCase.swift
@@ -23,7 +23,7 @@ import WireLogging
 public protocol RemoveEARKeysUseCaseProtocol {
 
     /// Removes all encryption at rest keys from the keychain for the given account.
-    func invoke(accountID: UUID) throws
+    func invoke(accountID: UUID)
 
 }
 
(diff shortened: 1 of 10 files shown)


netbe (approved):
LGTM:)

I just wonder if this for 4.16.1 or 4.17.0?

wire-ios-data-model/Source/UseCases/RemoveEARKeysUseCase.swift
```diff
@@ -0,0 +1,57 @@
...
+        self.keyRepository = EARKeyRepository()
+    }
+
+    init(keyRepository: EARKeyRepositoryInterface) {
+        self.keyRepository = keyRepository
+    }
+
+    public func invoke(accountID: UUID) throws {
+        do {
+            try keyRepository.deletePublicKey(description: .primaryKeyDescription(accountID: accountID))
```
caldrian: Could we also try to delete as many as possible and not stop on the first failure?
```
do {
  try ...
} catch {
  // log
}
do {
  try ...
} catch {
  // log
}
...
```

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
```diff
@@ -101,6 +101,13 @@ final class SyncAgent: NSObject, SyncAgentProtocol {
         setupBindings()
     }
 
+    func tearDown() {
+        delegate = nil
+        Task { [syncAgent = self] in
+            await syncAgent.suspend()
```
caldrian: question: why capturing `self` as `syncAgent` instead of putting `await self.suspend()`

caldrian (approved):

wire-ios-data-model/Source/UseCases/RemoveEARKeysUseCase.swift
David-Henner: good point, I've updated [a57f6f8](https://github.com/wireapp/wire-ios/pull/4532/commits/a57f6f8121f262a3d6479751f216ea05712c75b6)

wire-ios-sync-engine/Source/Synchronization/SyncAgent.swift
David-Henner: no good reason, I think I got confused about strong reference capturing. Updated [a57f6f8](https://github.com/wireapp/wire-ios/pull/4532/commits/a57f6f8121f262a3d6479751f216ea05712c75b6)

PR #4539, opened 2026-04-08
fix: crash resolving WireDrive backend URL - WPB-24584

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24584" title="WPB-24584" target="_blank">BugWPB-24584</a>  [iOS] Crash when resolving wire drive backend URL
  </td></table>
  <br />
 

### Issue

From crash log, the mention `doDecrementSlow` indicates there's race condition happening when resolving the wire drive backend URL.

Race condition was reproducible with this debug code which indicates that multiple threads might attempt to initialize the property simultaneously resulting in undefined behavior (crash).

```
while true {
        DispatchQueue.global().async {
            do {
                _ = try driveURLResolver()
            } catch {
                   
            }
        }
}
```

**Root cause:** race condition, lazy var are thread-unsafe
**Solution**: Remove the lazy initialization

### Testing

_Describe how to test._

---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/MainController/ZClientControllerBuilder.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/MainController/ZClientControllerBuilder.swift
@@ -35,7 +35,7 @@ final class ZClientControllerBuilder {
     private(set) var trackingManager: TrackingManager?
     let legacyEnvironment: WireTransport.BackendEnvironment
     let newEnvironment: WireNetwork.BackendEnvironment2?
-    private lazy var wireDriveBackendURL: URL? = {
+    private var wireDriveBackendURL: URL? {
         let contextProvider = userSession.contextProvider
         let syncContext = contextProvider.syncContext
         let featureRepository = LegacyFeatureRepository(context: syncContext)
@@ -43,7 +43,7 @@ final class ZClientControllerBuilder {
         return syncContext.performAndWait {
             featureRepository.fetchCellsInternal()?.config.backend.url
         }
-    }()
+    }
 
     init(
         account: Account,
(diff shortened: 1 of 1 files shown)


wire-ios/Wire-iOS/Sources/UserInterface/MainController/ZClientControllerBuilder.swift
```diff
@@ -35,15 +35,15 @@ final class ZClientControllerBuilder {
     private(set) var trackingManager: TrackingManager?
     let legacyEnvironment: WireTransport.BackendEnvironment
     let newEnvironment: WireNetwork.BackendEnvironment2?
-    private lazy var wireDriveBackendURL: URL? = {
+    private var wireDriveBackendURL: URL? {
         let contextProvider = userSession.contextProvider
         let syncContext = contextProvider.syncContext
         let featureRepository = LegacyFeatureRepository(context: syncContext)
 
         return syncContext.performAndWait {
```
netbe: suggestion: is there a good reason for fetching on syncContext? As the build(router:) is on MainActor it could be better to fetch from viewContext

netbe (approved):
Looks good to fix the issue, made a suggestion too

WilhelmOks (approved):

wire-ios/Wire-iOS/Sources/UserInterface/MainController/ZClientControllerBuilder.swift
jullianm: yeah it makes sense to use the view context here rather than the sync context for the reason you mentioned and also because I think this operation is tightly related to UI updates (fetch the URL -> perform an API request -> update UI). Changed in last commit, thanks

PR #4542, opened 2026-04-08
fix: skipping the rename request if the name didn't change - WPB-22329

WilhelmOks:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22329" title="WPB-22329" target="_blank">BugWPB-22329</a>  [iOS] Error when user does not change the name in Renaming flow
  </td></table>
  <br />
 

### Issue

When renaming a file or a folder, I added a check if the name has changed.
If it didn't change, the rename BE request is skipped.
The result is that no error message appears.

### Testing

* Go to a Drive enabled conversation
* Select a file or folder for renaming
* don't change the name and press Save
* the screen should close immediately


OlivellaO (approved):
LGTM

jullianm (approved):

PR #4549, opened 2026-04-10
fix: empty folder text - WPB-23213

WilhelmOks:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23213" title="WPB-23213" target="_blank">BugWPB-23213</a>  [iOS] Incorrect copy for folder empty state
  </td></table>
  <br />
 

### Issue

Added different text for empty folders (different to the text of the empty roots).

Also refactored `FilesInfoView` to make the switch logic for the different states more clear and easier to maintain.
And removed `.accessibilityLabel()` from `Text()` since it's redundant.

Moved `LoadMoreView` from `FilesInfoView` into a separate file.

Second phase:

Unified `MoveToFolderEmptyStateView` into `FilesInfoView` (It has been used by the Move-to-Folder view and also by the Edit-File view).

Added snapshot tests for `FilesInfoView`.

Introduced `Variants` to `SnapshotHelper` to make it easier to test light/dark themes and different sizes.
@johnxnguyen @caldrian please review this part in particular as it also affects the code outside of Wire Drive.

### Testing

* Go to the files list of a Drive enabled conversation or to the Drive tab
* Empty folders should have different text than a completely empty files list (root)
* Empty folders in the Recycle Bin should also have different text


WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesInfoView.swift
```diff
@@ -173,20 +193,14 @@ struct FilesInfoView: View {
...
+    @ViewBuilder
+    private func learnMoreLink() -> some View {
+        if let link = info.learnMoreLinkData {
+            Link(destination: link.destination) {
+                Text(link.title)
+                    .foregroundColor(SemanticColors.Label.baseSecondaryText.color)
+                    .underline()
+            }
         }
     }
```
jullianm, line 249: @WilhelmOks could we introduce snapshot tests for this view ? might be worth tracking any UI changes on this view since we have several different wordings.

WireFoundation/Sources/WireTesting/Utilities/SnapshotHelper.swift
```diff
@@ -211,6 +211,87 @@ public struct SnapshotHelper {
         }
     }
 
+    public enum Variants {
+        case colorSchemes
+        case sizes(_ set: SizeSet = .smallestLarge)
```
jullianm, line 216: question: Shouldn't we add a third variant for snapshots solely focused on accessibility with the various sizes (accessibility1, accessibility2 to accessibility5) ? It could be more clear at callsite whether we're testing accessibility or _regular_ sizes.

WireFoundation/Sources/WireTesting/Utilities/SnapshotHelper.swift
```diff
@@ -211,6 +211,87 @@ public struct SnapshotHelper {
...
+                [.xSmall, .xxxLarge]
+            case .smallestLargest:
+                [.xSmall, .accessibility5]
+            case .smallestLargeLargest:
+                [.xSmall, .xxxLarge, .accessibility5]
+            case .smallestNormalLarge:
+                [.xSmall, defaultSize, .xxxLarge]
+            case .smallestNormalLargest:
+                [.xSmall, defaultSize, .accessibility5]
+            }
```
jullianm: thought: Following up on my first comment, I think it might be better to distinguish between accessibility snapshots and _regular_ (sizes) snapshots as we're mixing both here.

WireMessaging/Tests/WireMessagingTests/WireDrive/UITests/Components/Files/FilesInfoViewTests.swift
```diff
@@ -0,0 +1,122 @@
...
+    func testEmpty() async {
+        let configs: [(name: String, scope: FilesInfoView.Scope)] = [
+            (name: "all_conv_files", scope: .files(conversation: .all)),
+            (name: "one_conv_files_root", scope: .files(conversation: .one, isFolder: false)),
+            (name: "one_conv_files_folder", scope: .files(conversation: .one, isFolder: true)),
+            (name: "recycle_bin_root", scope: .recycleBin(isFolder: false)),
+            (name: "recycle_bin_folder", scope: .recycleBin(isFolder: true)),
+            (name: "search", scope: .search),
+            (name: "move_to_folder", scope: .moveToFolder)
+        ]
```
jullianm: suggestion: use Apple's API `@Test(arguments: [])` from `Testing` framework. You can check `WireDriveDraftMetadataRepositoryTests` as an example.

This suggestion is optional as we don't always use this framework in our tests however I think it could be good candidate here.

jullianm (commented):
Nice work, left a few feedbacks before approving. As we're introducing a new helper method for snapshot tests, I'd suggest adding devs from the iOS team as reviewers as well.

OlivellaO (approved):
LGTM

WireFoundation/Sources/WireTesting/Utilities/SnapshotHelper.swift
```diff
@@ -211,6 +211,87 @@ public struct SnapshotHelper {
...
+        case colorSchemes
+        case sizes(_ set: SizeSet = .smallestLarge)
+
+        public enum SizeSet {
+            case all
+            case smallestLarge
+            case smallestLargest
+            case smallestLargeLargest
+            case smallestNormalLarge
+            case smallestNormalLargest
```
OlivellaO, line 224: This size naming is really confusing... I know it was like this before, it's just not clear at all 😅
I'm not asking to change anything, just complaining 😄

WilhelmOks: It wasn't like this before. It's something new introduced by me.
I don't see why this is confusing, maybe you can explain or make a suggestion for better naming?
I'm genuinely interested in improving the naming if it's confusing.

WilhelmOks:
> Nice work, left a few feedbacks before approving. As we're introducing a new helper method for snapshot tests, I'd suggest adding devs from the iOS team as reviewers as well.

Makes sense, I added Sam and John.

WireFoundation/Sources/WireTesting/Utilities/SnapshotHelper.swift
WilhelmOks: Imo, it should not matter in tests if we are testing accessibility or not accessibility. The end user can select all of the sizes, including the extreme accessibility sizes, so we should test them in any case. My intention was to hide the details on the call side if it's about accessibility or not, because it shouldn't matter.

WireFoundation/Sources/WireTesting/Utilities/SnapshotHelper.swift
WilhelmOks: See my answer on the other comment.

WireMessaging/Tests/WireMessagingTests/WireDrive/UITests/Components/Files/FilesInfoViewTests.swift
WilhelmOks: Good suggestion. I did the change.

WireFoundation/Sources/WireTesting/Utilities/SnapshotHelper.swift
OlivellaO: I really didn't think about a suggestion. I had to go thru the case let .sizes(set) to see what any of this cases mean.
I can see they will all test .xSmall and then also test another 1 or 2 sizes. So Large = xxxLarge, Largest = .accessibility5 and Normal = defaultSize but you need to go thru the implementation at least once to get to know what corresponds each thing.
Maybe adding documentation to this changes (SizeSet, Variants & new verify) so when using it you will be able to quickly check the right SizeSet you need for the test you want to perform.

WireFoundation/Sources/WireTesting/Utilities/SnapshotHelper.swift
jullianm: makes sense

jullianm (approved):
Nice work!

johnxnguyen (approved):

PR #4550, opened 2026-04-10
chore: replace expiring activity manager - WPB-24585

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24585" title="WPB-24585" target="_blank">BugWPB-24585</a>  [iOS] Crash on ExpiringActivityManager
  </td></table>
  <br />
 

### Issue

This PR replaces the `ExpiringActivityManager` actor we have by some code which just cancels a provided task if the system tells us to stop background work.

---

### Checklist

---


WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/registerExpiringActivity.swift
```diff
@@ -0,0 +1,75 @@
...
+
+            // System granted time. Block this callback until the task finishes,
+            // so the system knows we're still doing work.
+            let semaphore = DispatchSemaphore(value: 0)
+            defer { semaphore.wait() }
+
+            Task<Void, Never> {
+                // We only need to wait for the task to complete; the result is irrelevant.
+                _ = try? await task.value
+                semaphore.signal()
```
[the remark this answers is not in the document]
caldrian: I'm not able to verify this. I'll use `Task.detached` to be on the safe side.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -165,24 +166,17 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
+                let task = Task {
+                    await processLiveEvents(
+                        liveEventStream: liveEventStream,
+                        processedEnvelopeIDs: processedEnvelopeIDs,
+                        publicKeys: publicKeys
                     )
-                    await pushChannel.close()
                 }
+                registerExpiringActivity(reason: "processLiveStream IncrementalSync", task: task)
+                await task.value
```
netbe: question: hmm are you not waiting and executing twice the task?
- 1 time here
- 1 time in the registerExpiringActivity detached task?

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/registerExpiringActivity.swift
```diff
@@ -0,0 +1,75 @@
...
+            // so the system knows we're still doing work.
+            let semaphore = DispatchSemaphore(value: 0)
+            defer { semaphore.wait() }
+
+            Task.detached {
+                // We only need to wait for the task to complete; the result is irrelevant.
+                _ = try? await task.value
+                semaphore.signal()
+            }
+
```
netbe: nit-pic: 

```suggestion

            Task.detached {
                // We only need to wait for the task to complete; the result is irrelevant.
                _ = try? await task.value
                semaphore.signal()
            }
            semaphore.wait()
```

netbe (commented):
left a question, maybe there's something i misunderstood or we wait twice for the same task

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
caldrian: Yes exactly. In `registerExpiringActivity` we need to wait for the task to finish (ignoring the actual outcome/result) in order to block the execution of the closure. Otherwise the system will not call the closure again with `isExpiring` true.

caldrian: The task is only executed once.

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/registerExpiringActivity.swift
```diff
@@ -0,0 +1,74 @@
...
+        } else {
+
+            // System granted time. Block this callback until the task finishes,
+            // so the system knows we're still doing work.
+            let semaphore = DispatchSemaphore(value: 0)
+            Task.detached {
+                // We only need to wait for the task to complete; the result is irrelevant.
+                _ = try? await task.value
+                semaphore.signal()
```
samwyndham: issue (I think): As far as I understand this block is executed 1 or 2 times depending on if it expires. In the second case we have a problem because:

1. Just because the `task` is cancelled it doesn't mean it will finish immediately (due to cooperative cancellation)
2. If the `task` does keep running the semaphore won't be cancelled.
3. The first execution of the block may not end quickly so the system will kill the process.

suggestion: Ensure that the semaphore is signaled when isExpiring is true

samwyndham (changes requested):
I have't reviewed the PR fully but I think I have found an issue worth addressing.

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/registerExpiringActivity.swift
caldrian: good catch! fixed

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityPerformerProtocol.swift
```diff
@@ -0,0 +1,66 @@
...
+    func performExpiringActivity(
+        reason: String,
+        using block: @escaping @Sendable (_ isExpiring: Bool) -> Void
+    )
+
+}
+
+extension ExpiringActivityPerformerProtocol {
+
+    func performTaskCancellationAsExpiringActivity(
```
johnxnguyen, line 44: question: do I understand correctly that this method doesn't block while the semaphore is being waited on?

johnxnguyen (commented):
Looks good, I like the simplicity, but I have a question:

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityPerformerProtocol.swift
caldrian: Correct, `performTaskCancellationAsExpiringActivity` doesn't block.
It just passes a task to be cancelled when the system tells us to stop background work.

We only block in the closure which is passed into `performExpiringActivity` and executed on a global concurrent queue and only if the bool `isExpiring` is false.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
```diff
@@ -156,26 +157,16 @@ public struct IncrementalSyncV2: LiveSyncProtocol {
...
-                        liveEventStream,
-                        pushChannel: pushChannel,
-                        syncMarker: syncMarker
-                    )
-
-                    WireLogger.sync.debug("Live stream ended, close push channel", attributes: logAttributes)
-                    await pushChannel.close()
-                    await pushChannelState.markAsClosed()
-                }
-            } catch {
```
caldrian, line 173: I literally just deleted the `do`/`catch` (including the code in the `catch` block).

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -165,23 +166,14 @@ public struct IncrementalSync: IncrementalSyncProtocol {
...
-                        attributes: .incrementalSyncV2
+                // because we might be interrupted when in background, we wrap the sync in an expiringActivity that
+                // will cancel the task - not keeping any db operation (sqlite file opened) in suspend mode
+                await withExpiringActivity(reason: "processLiveStream IncrementalSync") {
+                    await processLiveEvents(
+                        liveEventStream: liveEventStream,
+                        processedEnvelopeIDs: processedEnvelopeIDs,
+                        publicKeys: publicKeys
                     )
-                    await pushChannel.close()
```
caldrian, line 184: Also here I literally just deleted the `do`/`catch` and expect the push channel to be closed properly.

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityPerformerProtocol.swift
johnxnguyen: Ok makes sense. I think it'd be good to make this explicitly clean in the doc comments so people don't make false assumptions.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
johnxnguyen: looking at the implementation of `processLiveEvents` it looks like it doesn't handle cancellation. If the activity expires then we'd need to check for cancellation to close the push channel, right @netbe ?

johnxnguyen (commented):
Looks good but I think the push channel may still need to be closed when the activity expires.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
netbe: hmm, the issue was there before regardless of this change, is it what you mean?

netbe: Image

adding this would make sense

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityPerformerProtocol.swift
```diff
@@ -0,0 +1,76 @@
...
+
+                // System granted time. Block this callback until the task finishes,
+                // so the system knows we're still doing work.
+                Task.detached {
+                    // We only need to wait for the task to complete; the result is irrelevant.
+                    _ = try? await task.value
+                    semaphore.signal()
+                }
+                semaphore.wait()
+
```
netbe, line 85: thought: I wonder if we should add logs as before in ExpiringActivityManager.

we could add some monitoring in datadog to check all non expiringActivity:
a. log when entering the activity
b. log when exiting (expiration or normal end of task)

on dd: add dashboard doing a - b, if we have a number != 0 we have an issue

netbe (commented):
This looks good, left a thought, suggestion.

samwyndham (approved):
LGTM

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
johnxnguyen: > egardless of this change,

It was fine because because in the `catch` when the activity expires the push channel was closed explicitly.

David-Henner (commented):
I was confused by the task being waited on from both within `performTaskCancellationAsExpiringActivity` and `withExpiringActivity` - maybe we could add a comment about it - other than that it looks good

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
caldrian: I'm not sure we reached the `catch` block, except in the one edge case that the expiring activity was never started (throwing `ExpiringActivityNotAllowedToRun()` error).

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -201,6 +193,9 @@ public struct IncrementalSync: IncrementalSyncProtocol {
     ) async {
         do {
             for try await var envelope in liveEventStream {
+
+                guard !Task.isCancelled else { return }
+
```
caldrian, line 203: I added this check here and in IncrementalSyncV2.
I hope it's not an issue that the `envelope` is consumed from the stream and then discarded.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
caldrian: I think fixing that code in regards to cleaning up properly etc. is out of scope of this PR.

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityPerformerProtocol.swift
caldrian: log statements added

WireDomain/Sources/WireDomain/Synchronization/IncrementalSyncV2.swift
caldrian: Maybe I should revert these changes and we should fix the code in another PR.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
netbe: suggestion: no it's fine, we'll be able to refetch it since we didn't update the lastEventID.  maybe add a log in the return so we can see it has been cancelled on datadog

netbe (approved):
LGTM:) left a small suggestion

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
caldrian: done

johnxnguyen (approved):

PR #4563, opened 2026-04-13
feat: re-create folder structure locally when offline - WPB-24437

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24437" title="WPB-24437" target="_blank">TaskWPB-24437</a>  [iOS] Fetch and display the assets available offline when connection is down
  </td></table>
  <br />
 

### Issue

This PR re-constructs the folder structure locally in offline mode using the asset path to keep a consistent UI with the online mode (which is BE driven). 
Also, local asset path is now being updated when moving a node to a folder otherwise it will not have a correct path when re-constructing the folder structure.

### Testing

Do the following to prepare for testing:

1) Make a file available offline (flat, no folders)
2) Make a file available offline then move it to a folder
3) Create multiple nested folders and move files inside those various folders, then make the them available offline

Switch to plane mode and ensure the file/folder stucture is preserved for each of these files in the shared drive screen.
In the drive tab, folders should not show up as files are flattened.

### Checklist

---

### DIFF
--- a/WireMessaging/Sources/WireMessagingData/WireDrive/WireDriveLocalAssetRepository.swift
+++ b/WireMessaging/Sources/WireMessagingData/WireDrive/WireDriveLocalAssetRepository.swift
@@ -67,9 +67,9 @@ package final class WireDriveLocalAssetRepository: WireDriveLocalAssetRepository
     @MainActor
     package func offlineAssets(
         conversationName: String?,
-        filePath: String?
+        assetsPath: String?
     ) async throws -> [WireMessagingDomain.WireDriveLocalAsset] {
-        try await store.offlineAssets(conversationName: conversationName, filePath: filePath)
+        try await store.offlineAssets(conversationName: conversationName, assetsPath: assetsPath)
     }
 
     /// Refreshes the local asset metadata for a given `nodeID` and deletes any cached file if necessary.
--- a/WireMessaging/Sources/WireMessagingData/WireDrive/WireDriveLocalAssetRepository.swift
+++ b/WireMessaging/Sources/WireMessagingData/WireDrive/WireDriveLocalAssetRepository.swift
@@ -65,8 +65,11 @@ package final class WireDriveLocalAssetRepository: WireDriveLocalAssetRepository
     }
 
     @MainActor
-    package func offlineAssets(conversationName: String?) async throws -> [WireMessagingDomain.WireDriveLocalAsset] {
-        try await store.offlineAssets(conversationName: conversationName)
+    package func offlineAssets(
+        conversationName: String?,
+        filePath: String?
+    ) async throws -> [WireMessagingDomain.WireDriveLocalAsset] {
+        try await store.offlineAssets(conversationName: conversationName, filePath: filePath)
     }
 
     /// Refreshes the local asset metadata for a given `nodeID` and deletes any cached file if necessary.
--- a/WireMessaging/Sources/WireMessagingData/WireDrive/WireDriveLocalAssetStore.swift
+++ b/WireMessaging/Sources/WireMessagingData/WireDrive/WireDriveLocalAssetStore.swift
@@ -48,13 +48,13 @@ package final class WireDriveLocalAssetStore: WireDriveLocalAssetStoreProtocol {
 
     package func offlineAssets(
         conversationName: String?,
-        filePath: String?
+        assetsPath: String?
     ) async throws -> [WireMessagingDomain.WireDriveLocalAsset] {
         let context = contextProvider.newBackgroundContext()
         return try await context.perform {
             let managedAssets = try context.fetchLocalOfflineAssets(
                 conversationName: conversationName,
-                filePath: filePath
+                assetsPath: assetsPath
             )
 
             return managedAssets.compactMap { managed in
@@ -201,7 +201,7 @@ private extension NSManagedObjectContext {
         return try fetch(request).first
     }
 
-    func fetchLocalOfflineAssets(conversationName: String?, filePath: String?) throws -> [ManagedLocalAsset] {
+    func fetchLocalOfflineAssets(conversationName: String?, assetsPath: String?) throws -> [ManagedLocalAsset] {
         let request = ManagedLocalAsset.fetchRequest() as! NSFetchRequest<ManagedLocalAsset>
         let isAvailableOfflinePredicate = NSPredicate(format: "isAvailableOffline == YES")
         let isDownloadedPredicate =
@@ -217,8 +217,8 @@ private extension NSManagedObjectContext {
             predicates.append(NSPredicate(format: "conversationName == %@", conversationName))
         }
 
-        if let filePath {
-            predicates.append(NSPredicate(format: "path CONTAINS %@", filePath))
+        if let assetsPath {
+            predicates.append(NSPredicate(format: "path BEGINSWITH %@", assetsPath))
         }
 
         request.predicate = NSCompoundPredicate(andPredicateWithSubpredicates: predicates)
--- a/WireMessaging/Sources/WireMessagingData/WireDrive/WireDriveLocalAssetStore.swift
+++ b/WireMessaging/Sources/WireMessagingData/WireDrive/WireDriveLocalAssetStore.swift
@@ -46,10 +46,16 @@ package final class WireDriveLocalAssetStore: WireDriveLocalAssetStoreProtocol {
         }
     }
 
-    package func offlineAssets(conversationName: String?) async throws -> [WireMessagingDomain.WireDriveLocalAsset] {
+    package func offlineAssets(
+        conversationName: String?,
+        filePath: String?
+    ) async throws -> [WireMessagingDomain.WireDriveLocalAsset] {
         let context = contextProvider.newBackgroundContext()
         return try await context.perform {
-            let managedAssets = try context.fetchLocalOfflineAssets(conversationName: conversationName)
+            let managedAssets = try context.fetchLocalOfflineAssets(
+                conversationName: conversationName,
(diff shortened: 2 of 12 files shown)


WireMessaging/Sources/WireMessagingData/WireDrive/WireDriveLocalAssetRepository.swift
```diff
@@ -65,8 +65,11 @@ package final class WireDriveLocalAssetRepository: WireDriveLocalAssetRepository
     }
 
     @MainActor
-    package func offlineAssets(conversationName: String?) async throws -> [WireMessagingDomain.WireDriveLocalAsset] {
-        try await store.offlineAssets(conversationName: conversationName)
+    package func offlineAssets(
+        conversationName: String?,
+        filePath: String?
```
OlivellaO: Are conversationName or filePath optional? I don't understand why they are nullable.
If they are mandatory shouldn't be nullable IMO.

WilhelmOks: I think they are optional because for the Drive tab we don't have a specific conversation, we want to fetch files for all conversations and for all paths because they are flattened.
This could be described in a doc comment.

WireMessaging/Sources/WireMessagingDomain/WireDrive/Protocols/WireDriveLocalAssetRepositoryProtocol.swift
```diff
@@ -30,7 +30,8 @@ package protocol WireDriveLocalAssetRepositoryProtocol: Sendable {
 
     /// Returns offline available local assets for a given conversation or for all conversations if nil.
     @MainActor
-    func offlineAssets(conversationName: String?) async throws -> [WireMessagingDomain.WireDriveLocalAsset]
+    func offlineAssets(conversationName: String?, filePath: String?) async throws
```
WilhelmOks: Doc comment missing new param filePath

WireMessaging/Sources/WireMessagingData/WireDrive/WireDriveLocalAssetStore.swift
```diff
@@ -211,6 +217,10 @@ private extension NSManagedObjectContext {
             predicates.append(NSPredicate(format: "conversationName == %@", conversationName))
         }
 
+        if let filePath {
+            predicates.append(NSPredicate(format: "path CONTAINS %@", filePath))
```
WilhelmOks: "CONTAINS" sounds dangerous, it could match wrong folders because it will match not only the suffix but also everything in the middle.
Is there a way to write the predicate with suffix instead of "contains"?

WireMessaging/Sources/WireMessagingDomain/WireDrive/Protocols/WireDriveLocalAssetStoreProtocol.swift
```diff
@@ -27,7 +27,8 @@ package protocol WireDriveLocalAssetStoreProtocol: Sendable {
     func asset(nodeID: UUID) throws -> WireDriveLocalAsset?
 
     /// Returns offline available local assets for a given conversation or for all conversations if nil.
-    func offlineAssets(conversationName: String?) async throws -> [WireMessagingDomain.WireDriveLocalAsset]
+    func offlineAssets(conversationName: String?, filePath: String?) async throws
```
WilhelmOks: Doc comment missing new param filePath

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesViewModel.swift
```diff
@@ -667,42 +670,81 @@ package final class FilesViewModel: ObservableObject {
...
                 let fileUrl = URL(fileURLWithPath: asset.path)
-                let fileName = fileUrl.lastPathComponent
                 let fileExtension = fileUrl.pathExtension
                 let fileType = UTType(filenameExtension: fileExtension)
 
+                func nextFolderPath(from fullPath: String, basePath: String) -> String? {
+                    let baseComponents = basePath.split(separator: "/")
+                    let fullComponents = fullPath.split(separator: "/")
+
+                    // no more folders.
```
WilhelmOks: nitpick suggestion:
instead of using a comment:

```
    let noMoreFolders = fullComponents.count == baseComponents.count + 1
    if noMoreFolders {
        return nil
    }
```

WireMessaging/Sources/WireMessagingData/WireDrive/WireDriveLocalAssetRepository.swift
OlivellaO: But you gonna need a filePath to know where the oflineAssets are stored, isn't it? i can understand conversationName, but unless there is a default i don't see it's weird to not provide a filePath.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesViewModel.swift
```diff
@@ -667,42 +670,81 @@ package final class FilesViewModel: ObservableObject {
...
                 let creationDate = (try? url?.resourceValues(forKeys: [.creationDateKey]))?.creationDate ?? Date()
                 return (item: item, creationDate: creationDate)
             }
-            
+
             let sortedItems = itemsWithCreationDates.sorted { lhs, rhs in
                 lhs.creationDate.compare(rhs.creationDate) == .orderedDescending
             }
             .map(\.item)
+            .reduce(into: [FilesViewItem]()) { result, item in
```
WilhelmOks, line 742: I'm a bit confused why this is a "reduce" instead of a "filter"

WilhelmOks (commented):
Added some suggestions and questions.

WireMessaging/Sources/WireMessagingData/WireDrive/WireDriveLocalAssetRepository.swift
WilhelmOks: for the Drive tab, filePath is actually not required because we need it only to know which folder the files are in, but there are no folders there. We fetch all files from all conversations, no matter which folder.

The filePath is the folder info from the directory of the conversation stored on the server, not the folder info on the local file system.

jullianm: @OlivellaO I updated the parameter name because I agree this is slightly confusing, hope it's more clear now. 
As @WilhelmOks mentioned it, the filePath is actually the `assetsPath` where we want to locate the assets belonging to a specific folder (if they're part of one). This helps us reconstruct the folder structure and it's only useful for the shared drive screen. 
https://github.com/wireapp/wire-ios/pull/4563/changes/a99f20d1dedda89ff0883ac9256a7129a2380cc8

WireMessaging/Sources/WireMessagingDomain/WireDrive/Protocols/WireDriveLocalAssetRepositoryProtocol.swift
jullianm: https://github.com/wireapp/wire-ios/pull/4563/changes/a99f20d1dedda89ff0883ac9256a7129a2380cc8

WireMessaging/Sources/WireMessagingData/WireDrive/WireDriveLocalAssetStore.swift
jullianm: While it shouldn't cause any issues with the current implementation, I do agree this is error prone to use CONTAINS. I updated it to use `BEGINSWITH` as this is actually the prefix we need and not the suffix here.
https://github.com/wireapp/wire-ios/pull/4563/changes/a99f20d1dedda89ff0883ac9256a7129a2380cc8

WireMessaging/Sources/WireMessagingDomain/WireDrive/Protocols/WireDriveLocalAssetStoreProtocol.swift
jullianm: https://github.com/wireapp/wire-ios/pull/4563/changes/a99f20d1dedda89ff0883ac9256a7129a2380cc8

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesViewModel.swift
jullianm: https://github.com/wireapp/wire-ios/pull/4563/changes/a99f20d1dedda89ff0883ac9256a7129a2380cc8

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesViewModel.swift
jullianm: How can we remove duplicates using filter ?

OlivellaO (approved):
LGTM

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesViewModel.swift
WilhelmOks: Sorry, you are right, we can't use filter to remove duplicates. This method with reduce is fine.

Btw, there is a function for that, `uniqued()` in Apple's official SPM package `swift-algorithms`.
But we'd need to import that package, which is probably overkill. Might be worth it if we end up using this function a lot.

WilhelmOks (approved):

PR #4566, opened 2026-04-14
fix: app state transition deadlock - WPB-24530

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24530" title="WPB-24530" target="_blank">BugWPB-24530</a>  [iOS] This message can not be displayed shown when phone is locked for a while
  </td></table>
  <br />
 

### Issue

After ending a call in a group that was migrated from proteus to MLS, going to the background then foreground results in EAR decryption errors due to the database key being missing. The app lock screen doesn't appear when foregrounding the app. 

### Cause

A chain of three failures:

- **sync failure:** When the user ends an MLS call, `didEndAllCalls()` suspends the sync. The task cancellation propagates to an ongoing `URLSessionDataTask` and an `NSURLErrorCancelled` propagates back to the SyncAgent's `ongoingSyncTask`. It then triggers `syncAgentDidFailSyncing`, which _in developer builds_ triggers `transition(to: .syncFailure)`. 

- **`appStateTransitionGroup` deadlock:** `AppRootRouter.transition(to:completion:)` handles the `.syncFailure` case by showing an alert. The alert button handlers would call `appStateTransitionGroup.leave()`. If the app backgrounds before the user taps - or the alert is dropped because the `rootViewController` is already presenting the call UI - `leave()` is never called and the group stays at count 1.

- **AppLock never presented:** With the group stuck, the subsequent `enqueueTransition(to: .locked)` - triggered when the database key is cleared on backgrounding - blocks indefinitely on
  `appStateTransitionGroup.wait()`. The AppLock view is never shown and never evaluates biometrics, so the database key cannot be fetched from the keychain.

### Solution

Move `appStateTransitionGroup.leave()` from the alert button handlers to immediately after `presentSyncErrorAlert` in `transition(to:completion:)`

### Testing

see ticket

---

### Checklist


netbe (approved):
Looks good. if only we could remove this appStateTransitionGroup seems fragile

samwyndham (approved):
Looks good. Additionally probably we should filter out cancelation errors from causing this alert as it's normal behavior and not really an error

jullianm:
Looks good. Maybe we could put that sync debug alert under a dev flag (instead of always showing it in developer mode `Bundle.developerModeEnabled`) so people can purposely enable/disable it if needed. 
Probably out of scope for this PR though.

jullianm (approved):

PR #4572, opened 2026-04-15
fix: prevent main app from blocking notification extension - WPB-23511

johnxnguyen:
<table>
<td>
  </td></table>
  <br />
 

### Issue
It's been observed that the Notification Service Extension (NSE) sometimes gets stuck and expires after 30 seconds of activity. Once this occurs, it typically repeats for every push received by the NSE until the main app is opened. 

After some thorough investigation, I found that the NSE gets stuck waiting on a Core Crypto transaction. There is a Core Crypto instance in both the main app and the NSE and to ensure that there is safe access to shared mutable state, Core Crypto internally uses file locks to make the transaction blocks are executed sequentially. The problem is that if the main app starts a CC transaction and then goes to the background before the transaction completes then execution would freeze while the lock is acquired and wouldn't be released. At the same time the push channel is closed and so new events will be delivered as pushes to the NSE which would be stuck indefinitely waiting for the lock to be released.

The solution is to ensure that all CC transactions can complete even before the app is suspended, which we can achieve by wrapping all transactions inside expiring activates. This ensures that even when we go to the background the file lock will be released and allow the NSE to run.

Changes include:
- modify the expiring activity function to return the wrapped task's result
- add an `extendedTransaction` function to Core Crypto that performs a transaction in an expiring activity
- use the `extentedTransaction` 

### Testing

The best way to verify this fix is to let it run in beta for a little while and observe how many NSE expirations occur.

---

### Checklist

---

### DIFF
--- a/wire-ios-data-model/WireDataModel.xcodeproj/project.pbxproj
+++ b/wire-ios-data-model/WireDataModel.xcodeproj/project.pbxproj
@@ -10,6 +10,7 @@
 		0176B8812E7AE25B005D448B /* WireLogging in Frameworks */ = {isa = PBXBuildFile; productRef = 0176B8802E7AE25B005D448B /* WireLogging */; };
 		01D2B7A32D64A8F50030D28D /* WireCoreCrypto in Frameworks */ = {isa = PBXBuildFile; productRef = 01D2B7A22D64A8F50030D28D /* WireCoreCrypto */; };
 		1657FA9A2D9C2EB800A7B337 /* WireCoreCryptoUniffi in Frameworks */ = {isa = PBXBuildFile; productRef = 1657FA992D9C2EB800A7B337 /* WireCoreCryptoUniffi */; };
+		34A4F28F2F8FBC4B006FA043 /* WireUtilitiesPackage in Frameworks */ = {isa = PBXBuildFile; productRef = 34A4F28E2F8FBC4B006FA043 /* WireUtilitiesPackage */; };
 		34DC44B52E01C210004D5DD5 /* WireNetwork in Frameworks */ = {isa = PBXBuildFile; productRef = 34DC44B42E01C210004D5DD5 /* WireNetwork */; };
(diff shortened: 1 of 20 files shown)


caldrian (approved):
🤞

wire-ios-data-model/WireDataModel.xcodeproj/project.pbxproj
```diff
@@ -10,6 +10,7 @@
 		0176B8812E7AE25B005D448B /* WireLogging in Frameworks */ = {isa = PBXBuildFile; productRef = 0176B8802E7AE25B005D448B /* WireLogging */; };
 		01D2B7A32D64A8F50030D28D /* WireCoreCrypto in Frameworks */ = {isa = PBXBuildFile; productRef = 01D2B7A22D64A8F50030D28D /* WireCoreCrypto */; };
 		1657FA9A2D9C2EB800A7B337 /* WireCoreCryptoUniffi in Frameworks */ = {isa = PBXBuildFile; productRef = 1657FA992D9C2EB800A7B337 /* WireCoreCryptoUniffi */; };
+		34A4F28F2F8FBC4B006FA043 /* WireUtilitiesPackage in Frameworks */ = {isa = PBXBuildFile; productRef = 34A4F28E2F8FBC4B006FA043 /* WireUtilitiesPackage */; };
```
netbe, line 13: question: is this an intented change?

netbe (approved):
looks good

With new implementation of withExpiringActivity, this looks safe but I wonder if it still necessary to wrap IncrementatlSync or other places with withExpiringActivity

wire-ios-data-model/WireDataModel.xcodeproj/project.pbxproj
johnxnguyen: yes because this is where the `withExpiringActivity` method is defined.

johnxnguyen:
> looks good
> 
> With new implementation of withExpiringActivity, this looks safe but I wonder if it still necessary to wrap IncrementatlSync or other places with withExpiringActivity

I think we should look at it case by case. Certainly we still want to gracefully suspend and I think an expiring activity is necessary, but that may only be to surround the suspend function and not the sync itself.

PR #4590, opened 2026-04-20
fix: FilesViewModel - WPB-24805

WilhelmOks:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24805" title="WPB-24805" target="_blank">TaskWPB-24805</a>  [iOS] improve FilesViewModel of WireDrive
  </td></table>
  <br />
 

### Issue

The swift file FilesViewModel is almost 1k loc long, is shared by multiple Views and has accumulated a lot of code for different kinds of functionality.
This makes it hard to read and maintain it.

### Solution

Many different kinds of refactoring were implemented resulting in the `FilesViewModel.swift` file being about 50% smaller:

* extracted `FilesViewItem` into a separate file
* refactored `SheetNavigation` to use Model data instead of SwiftUI Views as payload, eliminating the need for `import SwiftUI` in ViewModel code
* refactored sub-views and sheets to consistently use callbacks instead of Combine bindings to notify the parent view about specific events
* unified creation of sub-views or sheets to only create the ViewModels rather than SwiftUI Views
* extracted all functions which create ViewModels into the extension `FilesViewModel+make_ViewModels.swift`
* extracted the big `UseCases` nested struct into the extension `FilesViewModel+UseCases.swift`
* extracted the big function for loading offline files into `LoadOfflineAvailableFilesUIAction`
* extracted logic related to loading paginated lists of (file) items into the components `IncrementalListLoader` and `FilesListLoader`
* moved and grouped all of the remaining functions using `// MARK: label` comments

### Testing

The logic should be exactly the same after the refactoring.

* Go to the files list, via the Drive tab or the Shared Drive of a conversation
* Check if the list is loaded correctly and the pagination is right
* Check offline mode and if the list of offline available files is loaded correctly
* Check deleting and restoring files via recycle bin


WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Handlers/OfflineAvailableFilesHandler.swift
```diff
@@ -0,0 +1,150 @@
...
+//
+
+import Foundation
+import WireFoundation
+import WireLogging
+import WireMessagingDomain
+import WireMessagingDomainSupport
+import UniformTypeIdentifiers
+
+struct OfflineAvailableFilesHandler {
```
OlivellaO: Should we add some unit test to this?

WilhelmOks: Not sure if it would make sense for this particular case, but in general, I don't see why not.

OlivellaO (approved):
LGTM

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesViewModel+make_ViewModels.swift
jullianm, line 1: suggestion: rename `FilesViewModel+make_ViewModels` (which is not a naming convention we have in our codebase) to `FilesViewModelBuilder` (as in legacy code) or `FilesViewModelFactory`

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesViewModel+make_ViewModels.swift
```diff
@@ -0,0 +1,173 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireFoundation
+import WireMessagingDomain
+
+extension FilesViewModel {
```
jullianm, line 22: praise: nice to have these init all in on place.  left one suggestion about the naming of the file `.swift`

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/FilesListStateController.swift
```diff
@@ -0,0 +1,115 @@
...
+        } catch {
+            guard loader.state.isLoaded else { return }
+
+            // set items to how they were before the change:
+            loader.state = .received(items: items)
+        }
+    }
+}
+
+private extension Collection<FilesViewItem> {
```
jullianm, line 125: suggestion: this private extension is not necessary as the `latestModified` called from only one place. I suggest removing the extension.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/IncrementalListLoader.swift
```diff
@@ -0,0 +1,136 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Combine
+import Observation
+
+/// A generic loader for paginated lists of items.
+@MainActor
+class IncrementalListLoader<Item: Identifiable & Hashable & Sendable>: Observable, ObservableObject {
```
jullianm: question: what's the benefits of introducing generics here ? I'm not against it but I tend to think it introduces unnecessary complexity given this `IncrementalListLoader` is only used from one place thus the types could be explicitly declared without the need of generics.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/FilesListLoader.swift
```diff
@@ -0,0 +1,115 @@
...
+//
+
+import Combine
+import Foundation
+import Observation
+import WireLogging
+
+/// Loads and manages a paginated list of `FilesViewItem` elements.
+@MainActor
+class FilesListLoader: Observable, ObservableObject {
```
jullianm: `FilesListLoader` naming was a bit confusing for me at first, I thought we were dealing with a loader (progress indicator). Since the code comment also mentions that it `manages` the list, we could give it a name that better represents this component responsibility.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/IncrementalListStateController.swift
```diff
@@ -0,0 +1,136 @@
...
+    func fetchItems(offset: Int) async throws -> ItemsData {
+        if let onFetch {
+            let itemsData = try await onFetch(offset)
+            try Task.checkCancellation()
+            return itemsData
+        } else {
+            return (items: [], isLastPage: true)
+        }
+    }
+}
```
jullianm, line 136: thought: I feel like this component introduces an unnecessary layer of complexity, for example at callsite from the `FilesViewModel` we do : `filesListLoader.loader.reload(refreshing: refreshing)` which doesn't say much about what the `loader` inside the `filesListLoader` is about and induces some redundancy too.

I believe it would be more clear to do `filesListLoader.reload(refreshing: refreshing)`.

Consequently, I'd suggest to merge `FilesListLoader` and `IncrementalListLoader` into a single component to remove this unnecessary layer of complexity.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/Item/FilesItemView.swift
```diff
@@ -365,6 +365,20 @@ struct FilesItemView: View {
             )
         }
 
+        #if DEBUG
```
jullianm: note: the DEBUG code was removed on develop, it made the CI fails for unknown reason.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesView.swift
```diff
@@ -193,22 +194,21 @@ private extension FilesView {
...
-        case let .renameFile(fileRenameView):
-            fileRenameView
-        case let .create(folderView):
-            folderView
-        case let .versionHistory(versionHistoryView):
-            versionHistoryView
-        case let .moveToFolder(fileItem):
-            viewModel.moveToFolderView(item: fileItem)
+        case let .shareLink(item):
+            ShareLinkView(viewModel: viewModel.shareLinkViewModel(item: item))
```
jullianm, line 198: praise: thanks for fixing that

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/UI Actions/LoadOfflineAvailableFilesUIAction.swift
```diff
@@ -0,0 +1,127 @@
...
+//
+
+import Foundation
+import UniformTypeIdentifiers
+import WireFoundation
+import WireLogging
+import WireMessagingDomain
+import WireMessagingDomainSupport
+
+struct LoadOfflineAvailableFilesUIAction {
```
jullianm: Initially I had suggested this approach having in mind that we could have multiple UIActions as an intermediate layer between the use cases and the view models however in this PR we only have that single UIAction so I think we should no longer go with this approach and move this code to the loader.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/UI Actions/LoadOfflineAvailableFilesUIAction.swift
```diff
@@ -0,0 +1,127 @@
...
+            let isDuplicate = result.map(\.filePath).contains(item.filePath) // removes duplicated folders
+
+            if !isDuplicate {
+                result.append(item)
+            }
+        }
+    }
+}
+
+private extension Sequence {
```
jullianm: suggestion: the Sequence extension is not necessary here, it is private and only used in one place.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesViewModel.swift
```diff
@@ -326,146 +129,109 @@ package final class FilesViewModel: ObservableObject {
...
-            openFolder(item: item)
         }
     }
 
+    // MARK: load files
+
+    func reload(refreshing: Bool = false) async {
+        guard networkMonitor.currentStatus != nil else { return }
+
+        await filesListLoader.loader.reload(refreshing: refreshing)
```
jullianm: issue: it seems to me the naming of these objects is not clear enough so we can easily identify what exactly they are and what they are responsible for. `loaders` are a bit misleading (they're not progress indicators, more like files managers).

As per my suggestion in another comment, I think we have one too many loaders. One would be enough namely the `FilesListLoader` which acts more as a manager rather than simply a loader, and could be renamed accordingly.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesViewModel.swift
```diff
@@ -17,108 +17,39 @@
...
-    let size: UInt64?
-}
 
 private typealias Strings = L10n.Localizable.Conversation.WireCells
 private typealias Accessibility = L10n.Accessibility.Conversation.WireCells
 
-@MainActor
 /// View model for the `FilesView`.
+@MainActor
 package final class FilesViewModel: ObservableObject {
```
jullianm, line 29: praise: nice clean up of this view model

jullianm (changes requested):
Nice clean up of the view model, I left a couple of feedbacks/issues specifically related to the loaders components.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesViewModel+make_ViewModels.swift
WilhelmOks: With those names I'd expect to find there equally named types using the builder pattern or the factory pattern. But in this case it's just an extension of the FilesViewModel. So I think those would not be appropriate names.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/FilesListStateController.swift
WilhelmOks: It's not about extracting repeated code. The code of the extension is still required. Removing it would break the code.
And since it's private I think it's fine to have it as an extension.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/IncrementalListLoader.swift
WilhelmOks: Yes it's not needed and is just there to extract the essence of paginated list loading into a generic and reusable type, not polluted with the specifics of the Wire Drive file items.
I thought it might be easier to maintain that way.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/FilesListLoader.swift
WilhelmOks: Personally I don't associate the term "loader" with a progress indicator. For me the name means that it encapsulates the loading logic. It manages the different loading states and provides functions for starting and cancellation and potentially error handling and completion notifications.
I saw this pattern in other code bases and decided to use it here as well.
Do you have a suggestion for a better name?

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/IncrementalListStateController.swift
WilhelmOks: Indeed, the call side with the "double loader" was also worrying me.
Instead of merging the two, what do you think about just providing a better api for FilesListLoader so that the call side has all the public functions that IncrementalListLoader has?
It would then be this: `filesListLoader.reload(refreshing: refreshing)` on the call side and the `.loader` will be private.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/Item/FilesItemView.swift
WilhelmOks: Ah, I didn't know that. I will remove it then.

WilhelmOks: done

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/UI Actions/LoadOfflineAvailableFilesUIAction.swift
WilhelmOks: Yes I also thought we could extract more actions but when I tried I realized that the other actions don't make sense to extract.
With the current pattern, that code would go directly into the normal `FilesViewModel.swift` file.
It's quite large though and I'd like to extract it into somewhere, but I don't have a good idea where, without breaking the pattern.
With pattern I mean that the code for fetching the online files is in `setupFilesLoader()` in `FilesViewModel` and by symmetry the code for the offline files should go there as well.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/UI Actions/LoadOfflineAvailableFilesUIAction.swift
WilhelmOks: What do you mean, not necessary? It contains code that is needed.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesViewModel.swift
WilhelmOks: See my answer in the other comment.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesViewModel+make_ViewModels.swift
jullianm: the underscore is a bit strange. any reasons you chose this naming convention ?

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/FilesListStateController.swift
jullianm: I meant creating a Collection extension just for one place in the code, probably not necessary but I'm fine with that

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/IncrementalListLoader.swift
jullianm: do we intend to reuse it ? For the maintenance purpose I think it's fine to have explicit types instead of generics if it's not reused across components. I personally don't like introducing generics if there's not a real added value for it.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/FilesListLoader.swift
jullianm: I think this class is doing more than "loading":
- handles online/offline source selection
- owns pagination/incremental loading
- manages error presentation behavior
- performs optimistic item mutation rollback (removeItem)
- exposes observable UI state

I would suggest `FilesListStateController` since our flow revolves around our `State` enum.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/IncrementalListStateController.swift
jullianm: yeah that'd be definitely better to keep the loader private. Hiding the implementation details at call site.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/UI Actions/LoadOfflineAvailableFilesUIAction.swift
jullianm: you created a static helper func for nodes coming from the BE, can't we do same for offline asset ?
After all the logic is similar: take an input and returns a `FilesViewItem` as the output.

```
extension FilesViewItem {
    static func fromNode(_ node: WireDriveNode) -> FilesViewItem?
```

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/UI Actions/LoadOfflineAvailableFilesUIAction.swift
jullianm: creating a Sequence extension just for that, probably not necessary but I don't have strong argument against it  either.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/IncrementalListLoader.swift
WilhelmOks: As suggested in the other comment, I will make this generic type a private implementation detail of the FilesListLoader (which will be renamed). And then it should not matter if it's generic or not. 

If I understand you correctly, you'd like those two loader types to be one single non-generic type.
I think trying to merge those two into one type wouldn't be that productive.
It's not necessarily about being reusable but more about modularity and readability for me, and that one type is generic by nature, so just removing the generics for the sake of not having generics seems not right to me.

Maybe it's a bit overengineered, but I see no point in "underengineering" it now since it's already done and working.

What do you think?

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Files/FilesViewModel+make_ViewModels.swift
WilhelmOks: I just wanted to separate the word `make` from the term `ViewModels`.
I have no strong opinion about this. Might as well go with `FilesViewModel+makeViewModels` or something else.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/UI Actions/LoadOfflineAvailableFilesUIAction.swift
WilhelmOks: the logic inside of LoadOfflineAvailableFilesUIAction is not just the mapping from one file item type to another, it also contains logic for fetching, sorting and removing duplicate folders.

But you are right that the mapping part seems to be the biggest chunk.
I'll try to extract that and then the overall logic might be a more managable size.

WilhelmOks: done

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/IncrementalListStateController.swift
WilhelmOks: done

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/Common/FilesListLoader.swift
WilhelmOks: done

jullianm (approved):
Nice work!

PR #4591, opened 2026-04-20
fix: too many login requests - WPB-24616

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24616" title="WPB-24616" target="_blank">TaskWPB-24616</a>  [iOS] Release iOS v3.120.9
  </td></table>
  <br />
 

### Issue

Two login requests could be sent concurrently, causing the access token from the first successful login to be silently dropped.

**Root cause**

`ZMLoginTranscoder.nextRequestForAPIVersion:` called `readyForNextRequestIfNotBusy` on  `timedDownstreamSync` immediately before requesting a new request. 

In `ZMSingleRequestSync.processResponse:`, `currentRequest` is cleared (nil) before status is set to Completed.  In that window, the transport scheduler (on a different thread) could call nextRequestForAPIVersion: on the login transcoder, see `currentRequest == nil`, and re-ready the sync — resetting status to Ready and  incrementing `requestUniqueCounter`. This caused a second login request to be generated while the first  was completing, and the first request's success response to be dropped via the counter mismatch check,  leaving the access token unstored.

**Solution**

Remove the `readyForNextRequestIfNotBusy` call from `nextRequestForAPIVersion:`. The sync is already ready at initialization (`ZMTimedSingleRequestSync` calls `readyForNextRequest` in its initializer) and is  re-readied by `setTimeInterval:0` in `didReceiveResponse:` after every response, so forcing it ready on  every request tick was both redundant and racy.

### Testing

  1. Log in with an account on a backend that requires email verification (2FA)
  2. Complete the login flow including the email verification code step.
  3. Verify in network logs that only a single `POST /login?persist=true` is sent at each stage of the flow
  (no duplicate login requests).
  4. Verify that login succeeds and the session is established without being stuck or requiring a manual
  retry.

---

### Checklist

---

### DIFF
--- a/.github/workflows/_reusable_app_release.yml
+++ b/.github/workflows/_reusable_app_release.yml
@@ -20,7 +20,7 @@ on:
         default: false
       produce_debug_builds:
         type: boolean
-        required: true
+        required: false
         default: false
 
     secrets:
--- a/wire-ios-sync-engine/Source/Synchronization/Transcoders/ZMLoginTranscoder.m
+++ b/wire-ios-sync-engine/Source/Synchronization/Transcoders/ZMLoginTranscoder.m
@@ -94,7 +94,6 @@ NSTimeInterval DefaultPendingValidationLoginAttemptInterval = 5;
     }
 
     if(authenticationStatus.currentPhase == ZMAuthenticationPhaseLoginWithEmail) {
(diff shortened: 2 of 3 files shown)


jullianm (approved):
LGTM

johnxnguyen (approved):

PR #4595, opened 2026-04-20
fix: avatar not loaded after login - WPB-24860

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24860" title="WPB-24860" target="_blank">BugWPB-24860</a>  [iOS] [Profile] Own profile picture is not displayed when starting the app for the first time by user
  </td></table>
  <br />
 

### Issue
After first login, the self user avatar is not loaded in the leading toolbar item of the conversation list. This is because the avatar view uses the image data set on the `Account` object, which is not set directly after login but only later. The solution is to fallback to the user image data (stored on `ZMUser`) which is set after login.

### Testing
- Log in fresh
- Check the avatar is loaded with the profile picture of the self user
Simulator Screenshot - A - 2026-04-20 at 23 14 15

---

### Checklist

---


caldrian (approved):

jullianm (approved):
LGTM

PR #4596, opened 2026-04-21
chore: unified cookie storage - WPB-23404

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23404" title="WPB-23404" target="_blank">TaskWPB-23404</a>  [iOS] Review cookie handling
  </td></table>
  <br />
 

### Issue

My apologies. This PR has turned into something much bigger than I intended 🙏.

Currently the app has two versions of CookieStorage - the legacy obj-c ZMPersistentCookieStorage & the new CookieStorage. There are some issues with this such as the legacy version using a cache and the new one not... E.g. if the cookie is updated via the new cookie storage, this may not be reflected in the legacy cookie due to it's cache.

This PR attempts to unify the cookie storage mechanisms and make some other safety improvements to try to make everything work well in a concurrent and multiprocess environment. Specifically it:

- Refactors `CookieStorage` making it a non actor and removing async methods so that it can be used from non async contexts. It's access it protected by a lock to ensure that storing, fetching & deleting are _generally_ atomic operations.
- Made `CookieStorage` act on all users not a single user. Eg. a call to `fetchCookies()` is now `fetchCookies(userID: UUID)`. The reasoning for this is that 1. in some cases cookie storage is not cleanly tied to a user session, 2. There are currently some static methods on LegacyCookieStorage (e.g. testing that any cookie exists). These methods should move to `CookieStorage` eventually and it would be better to make them non static to allow for injection etc.
- Moves in memory cache to `CookieStorage` and changes how it works. We now store an epoch as metadata to Keychain items when they are . When adding to the cache we also store the epoch. When fetching from the cache we compare an items epoch to that in the Keychain to determine if the cached item is still valid. This works across processes meaning that we can now uses caching in app extensions and an cookie update coming from an app extension would now not break the cache in the main app (I'm not sure this happens in practice but believe it could happen). 
- Refactored `ZMPersistentCookieStorage` into `LegacyCookieStorage` which now wraps `CookieStorage` and is a much simpler object. Interestingly I started this refactor by using Claude to convert `ZMPersistentCookieStorage` and its tests from obj-c to swift. This worked well.
- Moved CookieStorageTests to the app test target so that I could write tests against the real keychain - I think this is worthwhile vs mocking because we can see that the behavior is really what we expect.
- Removed the concept of cookie policies from the legacy store as in practice this was always `NSHTTPCookieAcceptPolicyAlways`.

This PR also fixes: https://wearezeta.atlassian.net/browse/WPB-24720

@johnxnguyen @netbe I accidently merged as dependent PR - https://github.com/wireapp/wire-ios/pull/4596/commits/84eafe5fb33b56b6da195e4f25839a7267c6ecf5 - into this before this was reviewed. This contains UI tests and has been independently reviewed. Feel free to skip over reviewing the UI test code.

### Testing

General regression test logging in & out of multiple accounts, and making use of app extensions. 

---

### Checklist


WireNetwork/Sources/WireNetwork/Authentication/CookieStorage.swift
```diff
@@ -17,166 +17,317 @@
...
-    /// - Parameter cookies: The cookies to store.
+    /// - Parameters:
+    ///   - cookies: The cookies to store.
+    ///   - userID: The unique identifier for the user whose cookies are being stored.
 
-    public func storeCookies(_ cookies: [HTTPCookie]) async throws {
-        let cookieData = try HTTPCookieCodec.encodeCookies(cookies)
-        try await storeCookieData(cookieData)
+    public func storeCookies(_ cookies: [HTTPCookie], userID: UUID) throws {
+        try storeCookies(cookies, userID: userID, epoch: UUID())
```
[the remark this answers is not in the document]
samwyndham: Fixed

wire-ios-sync-engine/Source/UnauthenticatedSession/UnauthenticatedSession.swift
```diff
@@ -162,8 +163,12 @@ extension UnauthenticatedSession: UserInfoParser {
...
         let account = Account(userName: "", userIdentifier: userInfo.identifier)
         let cookieStorage = transportSession.environment.cookieStorage(for: account)
-        cookieStorage.authenticationCookieData = userInfo.cookieData
-        authenticationStatus.authenticationCookieData = userInfo.cookieData
+        do {
+            try cookieStorage.storeCookies(userInfo.cookies)
+        } catch {
+            WireLogger.authentication.error("Failed to store cookies: \(error)")
+        }
+        authenticationStatus.didReceiveAuthenticationCookies = true
```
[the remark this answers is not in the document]
samwyndham: It's a good catch. However I will keep existing behavior which silently ignores if setting a cookie fails.

wire-ios-transport/Source/TransportSession/Unauthenticated Session/UnauthenticatedTransportSession.swift
```diff
@@ -280,13 +278,13 @@ extension ZMTransportResponse {
     @objc
     public func extractUserInfo() -> UserInfo? {
         guard
-            let data = extractCookies().data,
+            extractCookies().data != nil, // TODO: Validate this in another way.
             let cookies = extractCookies().array,
             let id = extractUserIdentifier()
```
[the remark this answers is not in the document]
samwyndham: I've added a Jira ticket to improve this.

wire-ios-mocktransport/Tests/Source/Registration/MockTransportSessionRegistrationTests.m
```diff
@@ -376,15 +353,14 @@ - (void)testThatRegistrationWithPhoneNumberReturns201AndSetsTheCookie
...
-    [[(id) self.sut.cookieStorage expect] setAuthenticationCookieData:ZM_ARG_SAVE(cookieData)];
+    [[(id) self.sut.cookieStorage expect] storeCookies:OCMOCK_ANY error:[OCMArg anyObjectRef]];
     
     // WHEN
     ZMTransportResponse *response = [self responseForPayload:payload path:@"/register" method:ZMTransportRequestMethodPost apiVersion:0];
     
     // THEN
     XCTAssertEqual(response.HTTPStatus, 200);
-    XCTAssertNotNil(cookieData);
+//    XCTAssertNotNil(cookieData);
```
johnxnguyen: question: does this need to be deleted?

wire-ios/Wire-iOS Tests/Authentication/CookieStorageTests.swift
johnxnguyen, line 1: praise: good and comprehensive tests!

johnxnguyen (approved):
Looks good! Left a couple comments.

wire-ios-sync-engine/Source/UnauthenticatedSession/UnauthenticatedSession.swift
```diff
@@ -162,8 +163,12 @@ extension UnauthenticatedSession: UserInfoParser {
     public func upgradeToAuthenticatedSession(with userInfo: UserInfo) {
         let account = Account(userName: "", userIdentifier: userInfo.identifier)
         let cookieStorage = transportSession.environment.cookieStorage(for: account)
-        cookieStorage.authenticationCookieData = userInfo.cookieData
-        authenticationStatus.authenticationCookieData = userInfo.cookieData
+        do {
+            try cookieStorage.storeCookies(userInfo.cookies)
+        } catch {
+            WireLogger.authentication.critical("Failed to store cookies: \(error)")
```
netbe: suggestion: if it's critical we could add a assertionFailure here for debug

wire-ios-sync-engine/Source/UnauthenticatedSession/UnauthenticatedSession.swift
```diff
@@ -177,8 +182,12 @@ extension UnauthenticatedSession: UserInfoParser {
     ) {
         let account = Account(userName: "", userIdentifier: userInfo.identifier)
         let cookieStorage = transportSession.environment.cookieStorage(for: account)
-        cookieStorage.authenticationCookieData = userInfo.cookieData
-        authenticationStatus.authenticationCookieData = userInfo.cookieData
+        do {
+            try cookieStorage.storeCookies(userInfo.cookies)
+        } catch {
+            WireLogger.authentication.critical("Failed to store cookies: \(error)")
```
netbe: suggestion: if it's critical we could add a assertionFailure here for debug

netbe: consider if the error should be safePublic

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+Authentication.swift
```diff
@@ -60,7 +61,11 @@ extension ZMUserSession {
     /// This will delete user data stored by WireSyncEngine in the keychain.
 
     func deleteUserKeychainItems() {
-        transportSession.cookieStorage.deleteKeychainItems()
+        do {
+            try transportSession.cookieStorage.removeCookies()
+        } catch {
+            WireLogger.authentication.error("Failed to remove cookies: \(error)")
```
netbe: todo: consider if the error should be safePublic

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1550,7 +1549,11 @@ extension ZMUserSession {
             let clientUpdateStatus = applicationStatusDirectory.clientUpdateStatus
 
             clientRegistrationStatus.emailCredentials = nil
-            clientRegistrationStatus.cookieProvider.deleteKeychainItems()
+            do {
+                try clientRegistrationStatus.cookieProvider.removeCookies()
+            } catch {
+                WireLogger.authentication.error("Failed to remove cookies: \(error)")
```
netbe: todo: consider if the error should be safePublic

wire-ios-sync-engine/Source/UserSession/ZMClientRegistrationStatus.swift
```diff
@@ -387,7 +387,11 @@ public class ZMClientRegistrationStatus: NSObject, ClientRegistrationDelegate {
     @objc
     public func invalidateCookieAndNotify() {
         emailCredentials = nil
-        cookieProvider.deleteKeychainItems()
+        do {
+            try cookieProvider.removeCookies()
+        } catch {
+            WireLogger.authentication.error("Failed to remove cookies: \(error)")
```
netbe: todo: consider if the error should be safePublic

wire-ios-transport/Source/Authentication/LegacyCookieStorage.swift
```diff
@@ -0,0 +1,113 @@
...
+    public func setCookieData(from response: HTTPURLResponse, for url: URL) {
+        let headerFields = response.allHeaderFields as? [String: String] ?? [:]
+        let cookies = HTTPCookie.cookies(withResponseHeaderFields: headerFields, for: url)
+
+        guard cookies.first?.name == Self.cookieName else { return }
+
+        do {
+            try storeCookies(cookies)
+        } catch {
+            WireLogger.authentication.error("Failed to store cookies: \(error)")
```
netbe: todo: consider if the error should be safePublic

wire-ios/WireUITests/MultiBackendSupportTests.swift
```diff
@@ -141,6 +142,67 @@ final class MultiBackendSupportTests: WireUITestCase {
         XCTAssert(try ConversationsPage().conversationCell(named: conversationC).waitForExistence(timeout: 2.0))
     }
 
+    @MainActor
+    func testReLoginWhenMultipleBackends_TC_10550() async throws {
```
netbe, line 146: praise: nice to provide an automatic test

WireDomain/Sources/WireDomain/Account/AccountManager.swift
```diff
@@ -179,7 +179,7 @@ public final class AccountManager: NSObject, Sendable {
 
     public static func delete(at root: URL) {
         AccountStore.delete(directory: AccountURLs(root: root).accounts)
-        UserDefaults.shared().selectedAccountIdentifier = nil
+        UserDefaults.shared()!.selectedAccountIdentifier = nil
```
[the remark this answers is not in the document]
netbe: suggestion: makes sense to me cc @samwyndham

netbe (approved):
Looks good, left some suggestions, also check sonarQube comments some are relevant can be fixed by any AI tool

wire-ios-mocktransport/Tests/Source/Registration/MockTransportSessionRegistrationTests.m
samwyndham: Fixed in https://github.com/wireapp/wire-ios/pull/4596/commits/2ced36f37e7bd51496aeacba55d6fd6ecb5f5b6d

wire-ios-transport/Source/Authentication/LegacyCookieStorage.swift
samwyndham: Implemented in https://github.com/wireapp/wire-ios/pull/4596/commits/799f528ffe7e154b28eb86468c85925e4e368a7b

wire-ios-sync-engine/Source/UserSession/ZMClientRegistrationStatus.swift
samwyndham: Implemented in https://github.com/wireapp/wire-ios/pull/4596/commits/799f528ffe7e154b28eb86468c85925e4e368a7b

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
samwyndham: Implemented in https://github.com/wireapp/wire-ios/pull/4596/commits/799f528ffe7e154b28eb86468c85925e4e368a7b

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+Authentication.swift
samwyndham: Implemented in https://github.com/wireapp/wire-ios/pull/4596/commits/799f528ffe7e154b28eb86468c85925e4e368a7b

wire-ios-sync-engine/Source/UnauthenticatedSession/UnauthenticatedSession.swift
samwyndham: Implemented  as safe public in https://github.com/wireapp/wire-ios/pull/4596/commits/799f528ffe7e154b28eb86468c85925e4e368a7b.

I haven't bothered adding the assertionFailure because I don't want to add too much code. I feel that it should be done via `requireInternal(_:_:file:line:)` but I think that method needs to work to ensure it works in safe public which is outside the scope of this PR.

wire-ios-sync-engine/Source/UnauthenticatedSession/UnauthenticatedSession.swift
samwyndham: I haven't bothered adding the assertionFailure because I don't want to add too much code. I feel that it should be done via requireInternal(_:_:file:line:) but I think that method needs to work to ensure it works in safe public which is outside the scope of this PR.

samwyndham:
"DO NOT MERGE" label is applied as there is a crash when sending gifs (due to a nil response)

PR #4601, opened 2026-04-21
fix: prevent deadlock when using withExpiringActivity - WPB-24585

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24585" title="WPB-24585" target="_blank">BugWPB-24585</a>  [iOS] Crash on ExpiringActivityManager
  </td></table>
  <br />
 

### Issue

This PR fix an issue which could lead to all thread pool threads being blocked when `withExpiringActivity` is called more than once in parallel.

---

### Checklist

---


WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/withExpiringActivity.swift
```diff
@@ -16,26 +16,34 @@
 // along with this program. If not, see http://www.gnu.org/licenses/.
 //
 
+private let sharedManager = ExpiringActivityManager(
+    performer: ExpiringActivityProcessInfoWrapper()
+)
+
+private let sharedManagerV2 = ExpiringActivityManagerV2(
+    performer: ExpiringActivityProcessInfoWrapper()
+)
```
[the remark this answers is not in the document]
caldrian: This is WIP, I'll delete one implementation.

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityManagerV2.swift
```diff
@@ -0,0 +1,99 @@
...
+            onAllTasksFinished = { semaphore.signal() }
+
+            logger.debug("Registering expiring activity [reason: \(reason)]")
+
+            performer.performExpiringActivity(reason: "ExpiringActivityManager") { [self] isExpiring in
+                if isExpiring {
+                    logger.debug("System is revoking background time, cancelling all tasks [reason: \(reason)]")
+                    Task { await self.cancelAll() }
+                    semaphore.signal()
+                } else {
```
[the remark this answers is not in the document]
caldrian: added a generation counter

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/withExpiringActivity.swift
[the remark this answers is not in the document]
caldrian: I replaced ExpiringActivityManager with ExpiringActivityManagerV2

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityManager.swift
```diff
@@ -0,0 +1,122 @@
...
+
+    /// Track a task under the shared expiring activity. If this is the first
+    /// active task, an expiring activity is registered with the system.
+    ///
+    /// - Parameters:
+    ///   - reason: A human-readable reason used for logging. The system-level
+    ///     expiring activity always uses a static reason since multiple tasks
+    ///     share a single registration.
+    ///   - task: The task to protect. It will be cancelled if the system reclaims background time.
+    func track(reason: String, task: Task<some Sendable, some Error>) {
```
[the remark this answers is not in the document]
caldrian: The old implementation has been replaced.

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityManagerV2.swift
```diff
@@ -0,0 +1,99 @@
...
+        self.performer = performer
+    }
+
+    /// Track a task under the shared expiring activity. If this is the first
+    /// active task, an expiring activity is registered with the system.
+    ///
+    /// - Parameters:
+    ///   - reason: A human-readable reason used for logging.
+    ///   - task: The task to protect. It will be cancelled if the system reclaims background time.
+    func track(reason: String, task: Task<some Sendable, some Error>) {
```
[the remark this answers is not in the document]
caldrian: Removed `async` from the protocol.

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityManager.swift
```diff
@@ -0,0 +1,122 @@
...
+                    $0.isActive = false
+                    $0.cancellations.removeAll()
+                }
+            }
+
+            performer.performExpiringActivity(reason: "ExpiringActivityManager") { [self] isExpiring in
+                if isExpiring {
+                    logger.debug("System is revoking background time, cancelling all tasks [reason: \(reason)]")
+                    cancelAll()
+                    semaphore.signal()
```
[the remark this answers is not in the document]
caldrian: The implementation has been replaced by `ExpiringActivityManagerV2`

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityManager.swift
```diff
@@ -0,0 +1,109 @@
...
+    func track(reason: String, task: Task<some Sendable, some Error>) {
+        activeCount += 1
+        cancellations.append { task.cancel() }
+        let trackGeneration = generation
+
+        if activeCount == 1 {
+            let semaphore = DispatchSemaphore(value: 0)
+            onAllTasksFinished = { semaphore.signal() }
+
+            logger.debug("Registering expiring activity [reason: \(reason)]")
```
samwyndham, line 83: question: Does it make sense to log the reason as that is only for one task and now this expiring activity is for many tasks. In fact maybe we don't need a reason parameter at all because we just pass "ExpiringActivityManager" to the system

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityManager.swift
```diff
@@ -0,0 +1,109 @@
...
+            cancel()
+        }
+        generation += 1
+        activeCount = 0
+        cancellations.removeAll()
+        onAllTasksFinished = nil
+    }
+
+    private func taskDidFinish(generation: Int) {
+        guard generation == self.generation else { return }
```
samwyndham, line 116: question: Can you outline a situation where this guard would be false. My brain is broken.

samwyndham (commented):
So the general idea is to block until the last task completes right? I like the approach. I just left one comment where maybe you can outline how generations work. I'm trying to think through the logic of taskDidFinish

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityManager.swift
caldrian: Don't you think it could be useful when reading the logs?

Image

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityManager.swift
caldrian: In `track(reason:task:)` we copy the value of the member variable `generation` into a local `trackGeneration` variable and schedule a cleanup (`taskDidFinish`). 
If in the meantime the system triggers the closure with `isExpiring` being `true` and we started to cancel all, we want to prevent `taskDidFinish` from executing.

By increasing the generation value we reset the state for future calls.

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityManager.swift
```diff
@@ -0,0 +1,109 @@
...
+                    semaphore.signal()
+                } else {
+                    logger.debug("System granted background time, blocking until all tasks finish [reason: \(reason)]")
+                    semaphore.wait()
+                    logger.debug("All tasks finished, releasing expiring activity [reason: \(reason)]")
+                }
+            }
+        }
+
+        Task.detached { [self] in
```
netbe, line 98: question: in case of the second activity this doesn't increase the time of activity right?

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityManager.swift
```diff
@@ -0,0 +1,109 @@
...
+            _ = try? await task.value
+            await taskDidFinish(generation: trackGeneration)
+        }
+    }
+
+    private func cancelAll() {
+        for cancel in cancellations {
+            cancel()
+        }
+        generation += 1
```
netbe, line 109: question: I am not sure to understand the generation idea

samwyndham (approved):
LGTM

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityManager.swift
samwyndham: I'm not sure... I think it could be confusing because with your changes the `reason` is always for the first activity but this might not be the activity that the activity manager is waiting on

netbe (commented):
Leaving some comments to fully understand

johnxnguyen (commented):
Looks good to me but would like to see replies to @netbe 's questions

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityManager.swift
caldrian: True, I removed the reason from the logs where only the first reason wins.

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityManager.swift
caldrian: We ask the system for a **single** window of letting the app continue some work, before being suspended. 
Within this window we can call `track` as often as we want. 

If all our tasks finish within the window, the semaphore will be released early and the expiring activity torn down.

If we have unfinished tasks and the system tells us to stop, we will end the expiring activity immediately and schedule the cancellation of all tracked tasks.

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityManager.swift
caldrian: Under normal conditions (all tasks finish on their own), the generation never changes. It only matters when the system revokes background time. In that case, we cancel all tasks and immediately reset the manager so it's ready for new work. But the old cancelled tasks may still be winding down — when they eventually finish, the generation check tells them "you belong to a previous cycle, ignore your cleanup." Without it, their cleanup would interfere with the new cycle's state.

johnxnguyen: Good comment, I suggest to add it to the source code

WireFoundation/Sources/WireUtilitiesPackage/ExpiringActivityRegistration/ExpiringActivityManager.swift
johnxnguyen: This is a good explanation and i suggest to add it to the doc comment for the actor

johnxnguyen (approved):

PR #4606, opened 2026-04-21
feat: remove preview error when url is nil - WPB-24272

OlivellaO:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24272" title="WPB-24272" target="_blank">TaskWPB-24272</a>  [iOS] Non-preview Enabled files are displaying as big card with an error
  </td></table>
  <br />
 

### Issue

For some file types there is no preview to be shown. In those cases we were showing an error view to the user.
This PR updates the code so it will not try to load the image from url if url is nil.

Would be nice to consider for the future to add a "generic" type for file types that have no preview, right now we are just assuming that when url is nil there is no preview.

### Testing

Upload a Zip file for example and no preview or error when loading the preview should be shown.

---

### Checklist

---


jullianm (approved):
LGTM

WilhelmOks (approved):

PR #4613, opened 2026-04-22
feat: multiple media selection on Drive conversations - WPB-24046

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24046" title="WPB-24046" target="_blank">TaskWPB-24046</a>  [iOS] Improve sending multiple images on iOS
  </td></table>
  <br />
 

### Issue

Currently user has to select each image from the gallery (custom and native one) one by one and approve it in full screen view. 

This PR introduces changes to allow the user to frictionlessly upload multiple media at once. 

**Overview**

- On selection of assets from the _custom_ gallery (at the bottom), assets are now directly attached  - no full screen confirmation.
- Multiple assets can be selected and they are easily identifiable from the custom gallery (checkmarked). They can be deselected which automatically removes the related drafts from the preview.
- Full screen confirmation now only opens on tap of the attached asset (image or video)
- When the asset (image) is edited on that confirmation screen, it updates the attached asset.
- The native `PHPickerViewController` is introduced to allow multiple media selection (previously `UIImagePickerController` was used) 

### Testing

**Wire drive conversations**
- Hit camera icon button
- Select multiple media image / video, observe that full screen confirmation screen no longer opens but selected media are directly attached and uploaded.
- Hit “files” icon button, select multiple files and validate.
- In the input bar, tap on image attachment, it now opens full screen confirmation, edit the image (draw something), tap “OK”, attachment image will be upated.  Tapping “OK” or “Cancel” on an un-edited image will simply dimiss the screen
- Tap again on that updated attachment to check that asset has been correctly updated.
- Tap on a video attachment, observe that it opens the same full screen view, “OK” or “Cancel” simply dismisses the screen.
- Tap on file attachment, observe that it doesn’t open anything..
- In the media scrollable view, hit the “Gallery” icon button, it should open the native gallery screen allowing multiple media selection
- Select multiple media (videos/images), validate and observe selected media are correctly attached.
- Take a picture with the camera, observe that the draft is correctly attached
- Remove some attachments and attach new media based on the previous use cases to ensure drafts  behaviour is preserved.

**Non-Wire drive conversations (non regression test)**
- Behaviour should be as before as nothing was changed for these conversations.

### Checklist

---


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+ImagePickerDriveConversation.swift
```diff
@@ -0,0 +1,83 @@
...
+
+            do {
+                try FileManager.default.removeTmpIfNeededAndCopy(fileURL: url, tmpURL: videoTempURL)
+            } catch {
+                return WireLogger.wireDrive.error("Cannot copy video from \(url) to \(videoTempURL): \(error)")
+            }
+
+            AVURLAsset.convertVideoToUploadFormat(at: videoTempURL, fileLengthLimit: fileLengthLimit) { url, _, error in
+                guard error == nil, let url else { return }
+                self.uploadFiles(at: [url])
```
johnxnguyen: question: does the temp directory need to be cleanup here too?

WireMessaging/Sources/WireMessagingUI/Utilities/Array+SafeIndex.swift
```diff
@@ -0,0 +1,29 @@
...
+
+public extension Array {
+    subscript(safeIndex index: Int) -> Element? {
+        guard index >= 0, index < endIndex else {
+            return nil
+        }
+
+        return self[index]
+    }
+}
```
johnxnguyen: question: could we move this to WireFoundation?

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/AttachmentsCarousel/AttachmentsCarouselViewModel.swift
```diff
@@ -63,6 +63,11 @@ public final class AttachmentsCarouselViewModel: ObservableObject {
         }
     }
 
+    public func draft(for item: AttachmentsCarouselItem) -> WireDriveDraft? {
+        guard let index = items.firstIndex(of: item) else { return nil }
+        return drafts[safeIndex: index]
```
johnxnguyen: question: is this the only place we use this subscript? If it is, it doesn't seem necessary.

johnxnguyen (commented):
Looks good, left a couple questions.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+ImagePicker.swift
```diff
@@ -44,28 +45,40 @@ extension ConversationInputBarViewController {
...
-            pickerController.mediaTypes = mediaTypes
-            pickerController.videoMaximumDuration = userSession.maxVideoLength
-            pickerController.videoExportPreset = AVURLAsset.defaultVideoQuality
-            if sourceType == .camera {
-                let settingsCamera: SettingsCamera? = Settings.shared[.preferredCamera]
-                pickerController.cameraDevice = settingsCamera == .back ? .rear : .front
-            }
+            if useWireDrive(), sourceType != .camera {
+                // Allows multiple media selection on Wire drive conversations.
+                // Non-drive conversations use the (legacy) `UIImagePickerController` API.
```
netbe: question: would it be possible to use PHPicker on other conversations? I know it might be out of scope, but just wonder how big would be the change? maybe on a separate PR?

WireMessaging/Sources/WireMessagingDomain/WireDrive/UseCases/UploadDraftUseCase.swift
```diff
@@ -136,7 +141,7 @@ package struct UploadDraftUseCase: WireDriveUploadDraftUseCaseProtocol, WireDriv
         }
 
         let draft = WireDriveDraft(
-            nodeID: UUID(),
+            nodeID: nodeID ?? UUID(),
```
OlivellaO: If we need an UUID to create a WireDriveDraft, why is it not mandatory on the invoke function?
Maybe there is a reason for it but I would prefer to avoid nullability and avoid assigning this UUID in different places (default value on SendableImage & ConversationInputBarViewController if provided by draft object).
We could remove the default value on SendableImage and if draft.nodeId is null just use UUID()

OlivellaO (approved):
Approved with question

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/AttachmentsCarousel/AttachmentsCarousel.swift
```diff
@@ -79,6 +79,7 @@ private struct AttachmentsCarouselItemView: View {
         ZStack {
             ZStack {
                 content
+                    .contentShape(Rectangle())
```
WilhelmOks: suggestion: add a comment describing what this line is for. I think in this case it's to make it tappable?

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/AttachmentsCarousel/AttachmentsCarouselViewModel.swift
WilhelmOks: I agree with @johnxnguyen

WilhelmOks (approved):
LGTM. Added a few comments.

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/AttachmentsCarousel/AttachmentsCarouselViewModel.swift
jullianm: removed in https://github.com/wireapp/wire-ios/pull/4613/changes/700d3937b67771b683328653cbb50527011fe5ec

WireMessaging/Sources/WireMessagingUI/Utilities/Array+SafeIndex.swift
jullianm: removed in https://github.com/wireapp/wire-ios/pull/4613/changes/700d3937b67771b683328653cbb50527011fe5ec

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+ImagePicker.swift
jullianm: yes it would be possible but only for photos: `PHPickerViewController` doesn't handle camera.
The `UIImagePickerController` handles both.

We'll always end up with two different APIs to use  however it might make more sense to use the `PHPickerViewController` for photos (all conversations) and continue to use `UIImagePickerController` (or another API) only for camera. 


I don't think this would be too difficult to change (separated task / PR).

WireMessaging/Sources/WireMessagingUI/WireDrive/Components/AttachmentsCarousel/AttachmentsCarousel.swift
jullianm: done

WireMessaging/Sources/WireMessagingDomain/WireDrive/UseCases/UploadDraftUseCase.swift
jullianm: this is optional because sometimes the draft doesn't exist and we create it from scratch, and sometimes it does and we want to update it. I agree the naming is confusing here, I will find something better.

jullianm: done, the naming is a bit more explicit and gives more context on why it can be optional.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+ImagePickerDriveConversation.swift
jullianm: I extracted the legacy code into a separated reusable function here however you might be right.. I'll check it

jullianm:
@netbe @johnxnguyen @WilhelmOks @OlivellaO FYI, I made some improvements / fixes following the PT, these are visible in this single commit: https://github.com/wireapp/wire-ios/pull/4613/changes/700d3937b67771b683328653cbb50527011fe5ec

- Handle drafts selection/deselection on the custom picker: making it visible to the user (checkmark icon) and automatically synchronizing it with the draft preview (when asset is deselected, the draft preview is automatically removed)
- Prevent duplicated drafts uploads from the native picker 
- Adjust send button in edition screen

Feel free to review it if you want.

WireMessaging/Sources/WireMessagingDomain/WireDrive/UseCases/UploadDraftUseCase.swift
OlivellaO: This doesn't fix the problem but it's introducing and extra parameter and more complexity TBH. If we gonna leave it nullable i prefer the way it was before.

If the draft doesn't exist and we have to created from scratch we should assign an UUID when doing it. You're still assigning that UUID on 2 places and i don't think we should make things nullable for convenience. If a WireDriveDraft or SendableImage need and ID (and it makes sense), you should not be able to create one without ID.

jullianm: @OlivellaO this nullability gives us the indication whether we're creating a new draft or updating an existing one with different business logic inside the `invoke` method. So if we're removing the optionally on this parameter, we would need to add another parameter to know this information, I don't think this is optimal.

However, we could introduce an enum that would be make it more clear at callsite on the `invoke` method:

```
enum DraftType {
    case image(editedNodeID: UUID?)
    case video
    case file

    var id: UUID {
        switch self {
        case image(let editedNodeID):
            return editedNodeID ?? UUID()
         default:
            return UUID()
        }
    }
}
```

What do you think ? 
Please suggest something else if I'm missing your point.

Also for the context, `SendableImage` is only used for draft images and when an image is edited, the ID parameter will be non nil so we know we're updating a draft, for all other use cases it will be nil.

OlivellaO: As i commented on the second comment, i will avoid adding some extra logic and adding complexity to this.
**_If we need to keep it null then the first implementation you did was good, no need for changes._**

I will not add an extra Enum since it will increase complexity for no reason (and we add another code where if UUID is null we just create a new one)

I not a big fun of using null to know if we are updating a draft or not. We are assuming null means it's not being update and if there is no other use case were this could happen then it's fine. But if that can be null when is not being updated we could run into some issues.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+ImagePickerDriveConversation.swift
jullianm: @johnxnguyen I looked at the implementation details and the temp directory is always cleaned up once the video has been processed.

PR #4636, opened 2026-04-24
chore: support for macOS Designed for iPad mode - WPB-25053

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25053" title="WPB-25053" target="_blank">TaskWPB-25053</a>  [iOS] Running iOS app instead of Web macOS client
  </td></table>
  <br />
 

### Issue

In order to use more the Beta builds, this PR improve support for the app running on macOS via iPad design.

- add resizable window
- select part of messages
- record video
- disable callkit so calling works

### Testing

* Build as iPad
---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS.xcodeproj/project.pbxproj
+++ b/wire-ios/Wire-iOS.xcodeproj/project.pbxproj
@@ -2442,8 +2442,9 @@
 					"@executable_path/Frameworks",
 				);
 				LIBRARY_SEARCH_PATHS = "$(inherited)";
+				SUPPORTED_PLATFORMS = "iphoneos iphonesimulator";
 				SUPPORTS_MACCATALYST = NO;
-				SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD = NO;
+				SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD = YES;
 				SUPPORTS_XR_DESIGNED_FOR_IPHONE_IPAD = NO;
 				SWIFT_COMPILATION_MODE = singlefile;
 				TARGETED_DEVICE_FAMILY = "1,2";
@@ -2459,8 +2460,9 @@
 					"@executable_path/Frameworks",
 				);
 				LIBRARY_SEARCH_PATHS = "$(inherited)";
+				SUPPORTED_PLATFORMS = "iphoneos iphonesimulator";
 				SUPPORTS_MACCATALYST = NO;
-				SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD = NO;
+				SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD = YES;
 				SUPPORTS_XR_DESIGNED_FOR_IPHONE_IPAD = NO;
 				TARGETED_DEVICE_FAMILY = "1,2";
 			};
@@ -2470,6 +2472,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 01B71A962D6D2DAB002C5A38 /* Tests-Debug.xcconfig */;
 			buildSettings = {
+				SUPPORTED_PLATFORMS = "iphoneos iphonesimulator";
 				SUPPORTS_MACCATALYST = NO;
 				SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD = NO;
 				SUPPORTS_XR_DESIGNED_FOR_IPHONE_IPAD = NO;
@@ -2482,6 +2485,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 01B71A962D6D2DAB002C5A38 /* Tests-Debug.xcconfig */;
 			buildSettings = {
+				SUPPORTED_PLATFORMS = "iphoneos iphonesimulator";
 				SUPPORTS_MACCATALYST = NO;
 				SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD = NO;
 				SUPPORTS_XR_DESIGNED_FOR_IPHONE_IPAD = NO;
--- a/wire-ios/Wire-iOS/Sources/Helpers/UIApplication/UIApplication+Permissions.swift
+++ b/wire-ios/Wire-iOS/Sources/Helpers/UIApplication/UIApplication+Permissions.swift
@@ -55,7 +55,7 @@ extension UIApplication: ApplicationProtocol {
     }
 
     static func wr_requestOrWarnAboutPhotoLibraryAccess(_ grantedHandler: @escaping (Bool) -> Void) {
-        PHPhotoLibrary.requestAuthorization { status in
+        PHPhotoLibrary.requestAuthorization(for: .readWrite) { status in
             DispatchQueue.main.async {
                 switch status {
                 case .restricted:
--- a/wire-ios/Wire-iOS/Sources/Helpers/UIApplication/UIApplication+Permissions.swift
+++ b/wire-ios/Wire-iOS/Sources/Helpers/UIApplication/UIApplication+Permissions.swift
@@ -68,7 +68,7 @@ extension UIApplication: ApplicationProtocol {
                 case .authorized:
                     grantedHandler(true)
                 case .limited:
-                    fallthrough
+                    grantedHandler(true)
                 @unknown default:
                     break
                 }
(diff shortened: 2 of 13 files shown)


wire-ios/Wire-iOS/Sources/Helpers/UIApplication/UIApplication+Permissions.swift
caldrian: Btw, `requestAuthorization(_:)` is [deprecated](https://developer.apple.com/documentation/photos/phphotolibrary/requestauthorization(_:)).
```suggestion
        PHPhotoLibrary.requestAuthorization(for: .readWrite) { status in
```

wire-ios/Wire-iOS/Wire-Info.plist
```diff
@@ -154,7 +154,7 @@
 	<key>UIRequiredDeviceCapabilities</key>
 	<array/>
 	<key>UIRequiresFullScreen</key>
-	<true/>
+	<false/>
```
caldrian, line 157: This is a big change. We need to test resizing thoroughly, especially with `horizontalSizeClass` changes before disabling this.

wire-ios/Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -2442,8 +2442,9 @@
 					"@executable_path/Frameworks",
 				);
 				LIBRARY_SEARCH_PATHS = "$(inherited)";
+				SUPPORTED_PLATFORMS = "iphoneos iphonesimulator";
 				SUPPORTS_MACCATALYST = NO;
-				SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD = NO;
+				SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD = YES;
```
samwyndham, line 2447: question: Will this change apply to the production apps too?

wire-ios/Wire-iOS/Sources/Helpers/UIApplication/UIApplication+Permissions.swift
```diff
@@ -68,7 +68,7 @@ extension UIApplication: ApplicationProtocol {
                 case .authorized:
                     grantedHandler(true)
                 case .limited:
-                    fallthrough
+                    grantedHandler(true)
```
samwyndham, line 71: question: This is a change in behavior on iPhone/iPad too right? Do you know why we previously fell through?

wire-ios/Wire-iOS/Wire-Info.plist
```diff
@@ -154,7 +154,7 @@
 	<key>UIRequiredDeviceCapabilities</key>
 	<array/>
 	<key>UIRequiresFullScreen</key>
-	<true/>
```
samwyndham, line 157: question: This is a pretty big change but probably the right direction to go. But have you thought about possible issues here? I guess there must be a bunch of things to test.

wire-ios/Wire-iOS/Wire-Info.plist
samwyndham: Also any time we are doing `UIDevice.current.userInterfaceIdiom` we might be making assumptions about screen sizes.

wire-ios/Wire-iOS/Wire-Info.plist
samwyndham: Can we make this change only in non production builds?

wire-ios/Wire-iOS/Sources/UserInterface/Camera/MacVideoRecorderViewController.swift
```diff
@@ -0,0 +1,217 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import AVFoundation
+import UIKit
+
+/// A minimal video recorder for Mac (Designed for iPad).
+/// UIImagePickerController with .camera source crashes on Mac because Portrait Effects
+/// tries to create Metal textures with unsupported IOSurface formats. This replaces it.
+final class MacVideoRecorderViewController: UIViewController {
```
samwyndham, line 25: question: Did you check if this is possible and simpler in SwiftUI?

samwyndham (commented):
Nice work. I really like the idea of getting this working and it's nice that the changes aren't too great. 

My main concern is that we are changing a few fundamentals in production. Some questions:

1. Will Designed for iPad also appear in production?
2. Can we make the change to full screen mode internal/beta only?

wire-ios/Wire-iOS/Sources/UserInterface/Camera/MacVideoRecorderViewController.swift
netbe: no i didn't 👀

wire-ios/Wire-iOS/Wire-Info.plist
netbe: maybe not, or using a value from xcconfig

wire-ios/Wire-iOS.xcodeproj/project.pbxproj
netbe: yes

wire-ios/Wire-iOS/Sources/Helpers/UIApplication/UIApplication+Permissions.swift
netbe: Good catch on the question — the previous code was actually a latent bug: `.limited` → `fallthrough` → `@unknown default` → `break`, so `grantedHandler` was never called.

**Repro (pre-fix):**
1. Reset photo permission for Wire (Settings → Wire → Photos → none).
2. In a conversation, save an image to the Camera Roll (`SavableImage.saveToLibrary`).
3. On the iOS prompt, tap "Select Photos…" and pick a subset → status returns `.limited`.
4. `grantedHandler` never fires → silent failure: no save, no error, and `writeInProgess` stays `true`, blocking subsequent saves.

Treating `.limited` like `.authorized` (access granted, just to a subset) is the intended behavior. Same on iPhone/iPad — this fixes the bug rather than introducing a behavior change.

wire-ios/Wire-iOS/Wire-Info.plist
netbe: I did run it quickly and did not see any issue, do you have something in mind ?

wire-ios/Wire-iOS/Sources/Helpers/UIApplication/UIApplication+Permissions.swift
netbe: done

wire-ios/Wire-iOS/Sources/UserInterface/Camera/MacVideoRecorderViewController.swift
netbe: not simpler in SwiftUI

wire-ios/Wire-iOS/Wire-Info.plist
samwyndham: I'm not really sure but off the top of my head I guess the tab bar needs checking when in narrow split screen and possible edge insets for the bubbles in conversations.

wire-ios/Wire-iOS.xcodeproj/project.pbxproj
samwyndham: In my opinion we shouldn't do that. Users already have access to a mac app. At least this seems like a product decision.

wire-ios/Wire-iOS/Sources/UserInterface/Camera/MacVideoRecorderViewController.swift
samwyndham: Shame. Thanks for checking!

samwyndham (commented):
@netbe This is cool and I'm close to approving but I don't think it should be enabled for production apps or at least the ramifications of this should be considered and discussed with product. The ramifications are:

- In the mac app store there would now be multiple versions of the app. 
- We would have to support this new version and any new bugs

wire-ios/Wire-iOS/Wire-Info.plist
caldrian: I think we could remove UIRequiresFullScreen from Info.plist and set the value as build setting:
https://developer.apple.com/documentation/Xcode/build-settings-reference#Requires-Full-Screen
Than we're able to set different values (for AppStore and for internal TestFlight builds).

PR #4643, opened 2026-04-27
fix: disable file tracker when restoring file - WPB-25021

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25021" title="WPB-25021" target="_blank">BugWPB-25021</a>  [iOS] Restoring a File Version causes multiple openings
  </td></table>
  <br />
 

### Issue

Fixes an issue where restoring a file would trigger multiple file openings. The root cause is that when fetching the URL of a file, we also send updates to a Combine's subject which is observed from the main Files view, which in turn triggers opening the asset using the file tracker.
Solution here was to temporarily disable the file tracker when restoring a file as file tracking is not necessary when restoring a file.

### Testing

- go to version history on a file (context menu)
- restore a version, observe that the asset is opened once (file tracker disabled)
- go back to files view, tap on multiple files, observe that assets are downloaded and UI states correctly shown (file tracking enabled)

### Checklist

---


WilhelmOks (approved):
I don't particularly like this solution but I have no better idea how to solve this.

OlivellaO (approved):

PR #4657, opened 2026-04-29
chore: add UI tests for switching accounts - WPB-24968

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24968" title="WPB-24968" target="_blank">TaskWPB-24968</a>  [iOS] Automation - TC-8941 - As a user, I am able to switch between accounts.
  </td></table>
  <br />
 

### Issue

This PR:

- Introduces a UI test that covers switching accounts - https://app.testiny.io/IOS/testcases/tc/8941. The success of account switching is done by looking at the contents of the conversations list as this is something I have seen breaking in practice.

Additionally it does some refactoring to fix issues I experienced working with multiple backends above:

1. Deleted `BackendContext.current` as I found this shared mutable state lead to issues keeping `UserHelpers` in sync with  the current value of `BackendContext.current`. Instead this value is now passed through as needed.
2. Introduced some static methods for creating, tracking, and cleaning up `UserHelper`. This means that deletion of created accounts only needs to happen in once place and it will clean up accounts created for all backends. `UserHelper()` is now `UserHelper.default` and `UserHelper(backend: .anta)` is now `UserHelper.instance(backend: .anta)`.
3. Deleted userHelper instance variable from `WireUITestCase`. Use `UserHelper.default` instead.

https://github.com/user-attachments/assets/2e74958a-10f1-490c-8124-ddcdb17b3e25

### Testing

Run UI tests

---

### Checklist


wire-ios/WireUITests/Helper/WireUITestCase.swift
```diff
@@ -61,9 +58,9 @@ class WireUITestCase: XCTestCase {
     }
 
     override func tearDown() async throws {
+        await app.terminate()
```
samwyndham: This ensures the app is killed before deleting the created users remotely which could lead to a crash in the app.

wire-ios/WireUITests/AccountManagementTests.swift
```diff
@@ -73,4 +73,73 @@ final class AccountManagementTests: WireUITestCase {
         XCTAssertTrue(webViewPage.webViewOpened(), "WebView didn't open")
 
     }
+
+    @MainActor
+    func testSwitchingAccounts_TC_8941() async throws {
```
findms: Does it make sense to move this test in MultiBackendSupportTests where multi backend account getting added?

findms (approved):
lgtm. Left a comment to move this test to different Test file..

David-Henner (approved):
nice work! 🚀

wire-ios/WireUITests/AccountManagementTests.swift
samwyndham: Moved it in https://github.com/wireapp/wire-ios/pull/4657/commits/22bb835cde0909dc9a098bf020acbf2165d04b0a

PR #3373, opened 2025-07-22
refactor: load sessions with async/await - WPB-17719

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17719" title="WPB-17719" target="_blank">TaskWPB-17719</a>  [iOS] Deterministically load the user session
  </td></table>
  <br />
 

### Issue
In order to support multibackend we need to be able to load user sessions deterministically. One step to help achieve this is to make use of async/await when loading sessions. This PR removes the completion handler approach of loading sessions and replaces them with async/await.

Specifically:
- Load the CoreDataStack using async/await
- Load ZMUserSession via SessionManager with async/await
- Load SharingSession and NotificationSession with async/await

### Testing
Use the app as normal, switch accounts, receive push notifications, tap on push notifications to open the app, share messages from the share extension.

---

### Checklist

- [x] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [x] Description is filled and free of optional paragraphs.
- [x] Adds/updates automated tests.

---

### UI accessibility checklist

_If your PR includes UI changes, please utilize this checklist:_
- [ ] Make sure you use the API for UI elements that support large fonts.
- [ ] All colors are taken from WireDesign.ColorTheme or constructed using WireDesign.BaseColorPalette.
- [ ] New UI elements have Accessibility strings for VoiceOver.

PR #4663, opened 2026-04-30
fix: links aren't shared via share extension - WPB-24816

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24816" title="WPB-24816" target="_blank">BugWPB-24816</a>  [iOS] [Share ext] Sharing website from Safari doesn't include link
  </td></table>
  <br />
 

### Issue
When sharing a webpage via the share extension (such as from Safari) only the webpage title is included in the message and not the link.

This is because the webpage is the default text content and we must extract the url and append it to the text content, but a change introduced in https://github.com/wireapp/wire-ios/pull/3373 causes the attachments to be loaded later than usual due to being put inside a `Task`. As a result, when the view appears and we check to append any possible urls, there are no attachments loaded yet and nothing is appended.

The solution is to move the loading of attachments outside of the Task (it's not necessary to have it in the task) so that it is done synchronously before the view appears.

### Testing
1. Be logged in.
2. Open safari and try to share a webpage, such as www.apple.com
3. Before sending the message, assert that the text content includes the link.
4. Send the message, and check in the main app (sender or receiver) that the link is included in the sent message.

---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS Share Extension/Sources/View Controllers/ShareExtensionViewController.swift
+++ b/wire-ios/Wire-iOS Share Extension/Sources/View Controllers/ShareExtensionViewController.swift
@@ -162,12 +162,12 @@ final class ShareExtensionViewController: SLComposeServiceViewController {
         currentAccount = accountManager?.selectedAccount
         ExtensionBackupExcluder.exclude()
 
+        if let sortedAttachments = extensionContext?.attachments.sorted {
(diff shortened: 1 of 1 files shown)


samwyndham (approved):

netbe (approved):
LGTM:)

PR #4664, opened 2026-04-30
fix: mentions rendered incorrectly - WPB-21687

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21687" title="WPB-21687" target="_blank">BugWPB-21687</a>  [iOS] Mentioned replaces first letter of name with '@'
  </td></table>
  <br />
 

### Issue
It has been observed that in some cases mentions are rendered incorrectly, as shown here:
testThatItHandlesMalformedMentionRange_MissingAtSymbol 320-0 old

It's suspected that this is due to the sender of the text message incorrectly calculating the range of the mention to only include the name and not the the @ prefix. For instance, in the message `Hello @Bruno! How are you?` the mention range that is sent in the protobuf message should be location 6 with length 6 (encompassing `@Bruno`). But if instead the mention range is location 7 length 5 (encompassing `Bruno`) then the iOS app will render the mention incorrectly as shown in the previous attachment.

This PR accounts for this hypothetical case by checking if the mention range includes the prefix @. If it does not, it will check that the string has a @ prefix available and if it does will adjust the mention range to encompass also this @ prefix. Then it proceeds as normal with the mention rendering. The result is:

testThatItHandlesMalformedMentionRange_MissingAtSymbol 320-0

### Testing
Since it is not know how to reproduce this hypothetical case, there are no steps. However it seems to have something to do with copying and pasting a message containing mentions on the web app and sending the message, then editing the message to apply the mentions.

---

### Checklist

---


samwyndham (approved):
It looks good but I'm torn to whether this fix should be done on iOS if it really is a bug coming from web. Perhaps we could consider this a temporary fix and add some additional logging to help identify the underlying issue. Is Android not affected by the bug also?

johnxnguyen:
> It looks good but I'm torn to whether this fix should be done on iOS if it really is a bug coming from web. Perhaps we could consider this a temporary fix and add some additional logging to help identify the underlying issue. Is Android not affected by the bug also?

Yes I also agree with you, but I can try to raise it with the web team but I don't know how quickly they would be able to address it. I'll keep it open for now and see if we can at least verify that the bug may be coming from their side.

caldrian (approved):

PR #4689, opened 2026-05-11
fix: prevent main app from blocking notification extension (take 2) - WPB-23511

johnxnguyen:
### Issue
This PR is a second attempt at fixing the issue where the main app can block the notification service extension from running. The first attempt to fixing this (https://github.com/wireapp/wire-ios/pull/4572) was reverted due to code that it depended on being reverted too.

For more information about the issue, see the first PR.

This PR attempts to solve the issue in a similar way but uses a slightly different approach: Instead of using expiring activities via `ProcessInfo`, use the background task api from `UIApplication` as it is simpler and only needs to be run for the main app.

There is a new extension on `CoreCrypto` that mirrors the `transaction` method but adds an argument for a `BackgroundTaskManager`. This manager is an abstraction around the `UIApplication` background task api, which is conformed to by `UIApplication.shared` in the main app and `NoOpBackgroundTaskManager` for the extensions. The next transaction method invokes the existing transaction method after starting a background task and cleans up (as well as handle task expiration) appropriately.

### Testing

To artificially verify this fix you can use the included test helpers:
1. Open the developer tools, go to actions, tap "simulate 15s CC transaction"
3. Send a message to the app from another device.
4. Assert you don't see the notification
5. After 15 seconds you should see the notification
6. Any further messages should result in notifications without delay.

To test expiration of the background task, repeat the process but select "simulate 60s CC transaction"
- While in the background and after sending messages to the backgrounded app, assert you see no notification
- After about 30 seconds, the background task should expire and notifications should appear

---

### Checklist

- [ ] Adds/updates automated tests. (TODO)

---


WireDomain/Sources/WireDomain/Synchronization/PullPendingUpdateEventsV2Sync.swift
```diff
@@ -134,7 +134,10 @@ public struct PullPendingUpdateEventsSyncV2: PullPendingUpdateEventsSyncV2Protoc
         var storedEnvelopes: [(UpdateEventEnvelope, Int64)] = []
 
         // decrypt
-        try await coreCryptoProvider.coreCrypto().transaction { coreCryptoContext in
+        let backgroundTaskManager = coreCryptoProvider.backgroundTaskManager
+        try await coreCryptoProvider.coreCrypto().transaction(
+            backgroundTaskManager: backgroundTaskManager
+        ) { coreCryptoContext in
```
David-Henner: if the `backgroundTaskManager` and `coreCrypto` are  both held by `coreCryptoProvider`, is it necessary to pass the manager for every transaction ?

wire-ios-data-model/Source/Core Crypto/CoreCryptoProvider.swift
```diff
@@ -409,3 +427,44 @@ public actor CoreCryptoProvider: CoreCryptoProviderProtocol {
...
+        }
+
+        let taskID = backgroundTaskManager.beginBackgroundTask(
+            withName: "core crypto transaction",
+            expirationHandler: {
+                WireLogger.coreCrypto.warn(
+                    "Background task expiring, cancelling core crypto transaction",
+                    attributes: .safePublic
+                )
+                transactionTask.cancel()
```
David-Henner: so, cancelling the task ensures that core crypto releases the lock?

WireDomain/Sources/WireDomain/Synchronization/PullPendingUpdateEventsV2Sync.swift
johnxnguyen: The CoreCrypto provider doesn't perform the actual transaction so it doesn't know to pass the background task manager to core crypto. I know that it's not a very elegant solution and perhaps we can find a way to improve it. However I did discuss with @typfel that this perhaps could be moved inside core crypto if we can validate the solution on the iOS side first.

wire-ios-data-model/Source/Core Crypto/CoreCryptoProvider.swift
johnxnguyen: This would depend on cancellation checks in the given `block`, so no it does not guarantee the transaction is cancelled. I did not check extensively each transaction so that is something to do, but I would also be surprised if there is a transaction that lasts long enough for the task to expire.

WireFoundation/Sources/WireUtilitiesPackage/FileArchiving/BackgroundTaskManager.swift
```diff
@@ -0,0 +1,74 @@
...
+}
+
+extension UIApplication: BackgroundTaskManager {}
+
+/// A background task manager that does nothing.
+///
+/// Since the BackgroundTaskManager is only intended to be
+/// used from a main app target, this implementation can be used
+/// when running in an app extension.
+public struct NoOpBackgroundTaskManager: BackgroundTaskManager {
```
netbe: question: so we expect that NSE and ShareExtension never hold a file lock when suspended ?

netbe (commented):
left a question

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
```diff
@@ -1653,4 +1653,19 @@ extension ZMUserSession {
 
 }
 
+public extension ZMUserSession {
+
+    // TODO: [WPB-25551] Clean up testing code
+    func _simulateLongCCTransaction(seconds: Int) async throws {
```
netbe, line 1659: praise: good to have a way to test this

question: did you try this with the previous implementation? with the ProcessInfo imp

WireFoundation/Sources/WireUtilitiesPackage/FileArchiving/BackgroundTaskManager.swift
johnxnguyen: In the case of the NSE, either it completes or it expires. In the case that it expires we cancel the ongoing task and it should clean up gracefully. For the share extension I didn't look yet so I'm not sure, do you think it could be a problem?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession.swift
johnxnguyen: Yes I did and it also was useful

WireFoundation/Sources/WireUtilitiesPackage/FileArchiving/BackgroundTaskManager.swift
netbe: the usage is on foreground, I haven't seen any crash on share ext, so ok for now

netbe (approved):
Looks good, however it's not completely clear to me if the approach with ProcessInfo was not working @johnxnguyen

WireDomain/Sources/WireDomain/Synchronization/PullPendingUpdateEventsV2Sync.swift
David-Henner: Understood, thanks :) Perhaps you could add a TODO with a ticket to reflect this ?

David-Henner (approved):

johnxnguyen:
> Looks good, however it's not completely clear to me if the approach with ProcessInfo was not working @johnxnguyen

I believe it was working but I was concerned about adding extra load on the expiring activity manager as I wasn't sure it risky. Perhaps we can discuss it more and refine this solution.

wire-ios-data-model/Source/Core Crypto/CoreCryptoProvider.swift
```diff
@@ -428,7 +418,37 @@ public actor CoreCryptoProvider: CoreCryptoProviderProtocol {
 
 }
 
-public extension CoreCryptoProtocol {
+/// A wrapper object for CoreCrypto that ensures transactions are executed
+/// within a background task (if a task manager is provided).
+public final class SafeCoreCrypto {
```
netbe: suggestion: move to own file

netbe (approved):
lgtm:) SafeCoreCrypto is back 🐦‍🔥

David-Henner (approved):
good idea using `SafeCoreCrypto` ♻️

wire-ios-data-model/Source/Core Crypto/CoreCryptoProvider.swift
johnxnguyen: done

PR #4695, opened 2026-05-12
fix: crash when removing user on conversation header - WPB-25593

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25593" title="WPB-25593" target="_blank">BugWPB-25593</a>  [iOS][XCUITest] Exception while deleting user for UITest
  </td></table>
  <br />
 

### Issue

Fixes a crash `(EXC_BAD_ACCESS in swift_getObjectType)` that occurs when entering a 1:1 conversation whose other participant has been deleted (e.g., team member removed). Root cause: ZMConversation.creator was annotated nonnull in the ObjC header, but the CoreData model has always declared the relationship as Optional with a Nullify delete rule — so when the destination User row is removed, creator is set to nil by the store. The annotation was a lie the model never honored, and the resulting nil leaked into Swift code typed as non-optional. The protocol upcast in `ConversationStartedSystemMessageCellDescription.makeConfiguration(from:) `then crashed reading the type metadata at address 0x0.

**Changes**

- Annotation fix at the source. ZMConversation.creator is now nullable in ZMConversation.h and ZMConversation+Internal.h. This aligns the
  Swift/ObjC surface with what the CoreData model has always promised.
- Propagate the optional through system-message helpers instead of fabricating a fallback sender, since the creator is not always self:
   - `ZMConversation.appendSystemMessage(sender:)` and `appendFailedToAddUsersSystemMessage(sender:)` now accept ZMUser? (underlying
  ZMSystemMessage.sender was already optional).
   - `MessageLocalStore.createSystemMessage(sender:)` now accepts ZMUser? and forwards the optional creator unchanged for `.newConversationCreated` and `.receiptModeIsOn`.
- Crash site: `ConversationStartedSystemMessageCellDescription.makeConfiguration(from:)` now safely reads conversation.creator, falling back to the existing "someone" string when nil.
- Channel history depth row in `GroupConversationHeaderView` is skipped when the creator is gone (the row text is meaningless without a sender).
- Test: `SnapshotCenterTests.testThatItSnapshotsSetValues` updated to try XCTUnwrap the now-optional creator?.objectID.

### Testing

- Run `RemoveUserTests.testUserDeletedForPersonalUser_TC_9490` and `testRemoveTeamMemberAndConversationListUpdated_TC_9491` — they should no longer crash while inside the 1:1 conversation when the other user is deleted.
- Sanity-check group conversations: opening a normal group still renders the "started the conversation" header.

**Note:** because the other user gets deleted, the conversation.creator is then nil and the app will render a Group header instead of 1:1

---

### Checklist

---

### DIFF
--- a/WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
+++ b/WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
@@ -980,7 +980,7 @@ public final class MessageLocalStore: MessageLocalStoreProtocol {
 
     private func createSystemMessage(
         messageType: ZMSystemMessageType,
-        sender: ZMUser,
+        sender: ZMUser?,
         users: Set<ZMUser>? = nil,
         addedUsers: Set<ZMUser> = Set(),
         clients: Set<UserClient>? = nil,
--- a/wire-ios-data-model/Source/Model/Conversation/ZMConversation+Internal.h
+++ b/wire-ios-data-model/Source/Model/Conversation/ZMConversation+Internal.h
@@ -109,7 +109,7 @@ NS_ASSUME_NONNULL_END
 @property (readonly, nonatomic, nonnull) NSMutableSet<ZMMessage *> *mutableMessages;
 @property (readonly, nonatomic, nonnull) NSSet<ZMMessage *> *hiddenMessages;
 @property (readonly, nonatomic) enum ZMConnectionStatus relatedConnectionState; // This is a computed property, needed for snapshoting
-@property (nonatomic, nonnull) ZMUser *creator;
+@property (nonatomic, nullable) ZMUser *creator;
 @property (nonatomic, nullable) NSDate *lastModifiedDate;
 @property (nonatomic) ZMConversationType conversationType;
 @property (nonatomic, readonly) BOOL isSelfConversation;
--- a/wire-ios-data-model/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
+++ b/wire-ios-data-model/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
@@ -677,7 +677,7 @@ extension ZMConversation {
     @discardableResult
     func appendSystemMessage(
         type: ZMSystemMessageType,
-        sender: ZMUser,
+        sender: ZMUser?,
         users: Set<ZMUser>?,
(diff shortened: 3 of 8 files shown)


WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
```diff
@@ -980,7 +980,7 @@ public final class MessageLocalStore: MessageLocalStoreProtocol {
 
     private func createSystemMessage(
         messageType: ZMSystemMessageType,
-        sender: ZMUser,
+        sender: ZMUser?,
```
KaterinaWire, line 983: **question:** what will system messages look like if the sender is nil?

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Internal.h
```diff
@@ -109,7 +109,7 @@ NS_ASSUME_NONNULL_END
 @property (readonly, nonatomic, nonnull) NSMutableSet<ZMMessage *> *mutableMessages;
 @property (readonly, nonatomic, nonnull) NSSet<ZMMessage *> *hiddenMessages;
 @property (readonly, nonatomic) enum ZMConnectionStatus relatedConnectionState; // This is a computed property, needed for snapshoting
-@property (nonatomic, nonnull) ZMUser *creator;
+@property (nonatomic, nullable) ZMUser *creator;
```
KaterinaWire, line 112: wow, so conversation can be without a creator 😳

WireDomain/Sources/WireDomain/Repositories/Message/MessageLocalStore.swift
KaterinaWire: I believe "Someone ...." 🤔

KaterinaWire (approved):

wire-ios-data-model/Source/Model/Conversation/ZMConversation+Internal.h
netbe: for instance if the other user is deleted and he initiated the 1-1, then yes it will be removed (nullified)

findms (approved):
lgtm. tests also passed

Image

PR #4703, opened 2026-05-12
fix: certificate enrollment view not displayed by developer tools - WPB-24270

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24270" title="WPB-24270" target="_blank">BugWPB-24270</a>  [iOS] Certificate Expiration debug tool not working anymore on bund-next environments
  </td></table>
  <br />
 

### Issue

The certificate enrollment view is not being displayed by the developer tools menu when tapping on the `enroll` button.

### Cause

The developer tools sheet has `modalPresentationStyle != .fullScreen`, so the while loop in `topmostViewController` stops one level too early and returns the parent VC — not the developer tools VC. However, the parent VC already has a `presentedViewController` (the developer tools sheet), so `present()` fails silently.

This only affects builds that have `SecurityFlags.useEmbeddedIDPUserAgent` evaluate to `true`, because in this case `WebViewUserAgent` is used. It relies on a `targetViewController` to be presented. That target VC is the one captured by `topmostViewController()`

Builds with  `SecurityFlags.useEmbeddedIDPUserAgent`  evaluating to `false` rely on `OIDExternalUserAgentIOS`, which uses the window directly, instead of a target VC.

### Solution

use `topmostViewController(onlyFullScreen: false)` to find and use the developer tools sheet as target VC

---

### Checklist

### DIFF
--- a/wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DeveloperE2ei/DeveloperE2eiViewModel.swift
+++ b/wire-ios/Wire-iOS/Sources/Developer/DeveloperTools/DeveloperE2ei/DeveloperE2eiViewModel.swift
@@ -50,7 +50,7 @@ final class DeveloperE2eiViewModel: ObservableObject {
     func enrollCertificate() {
         guard
             let session = userSession,
-            let topmostViewController = UIApplication.shared.topmostViewController()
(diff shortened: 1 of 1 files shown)


netbe (approved):
simple fix

KaterinaWire (approved):

PR #4721, opened 2026-05-18
fix: ignore ExpiringActivityNotAllowedToRun error - WPB-25714

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25714" title="WPB-25714" target="_blank">BugWPB-25714</a>  [iOS] Beta reports ExpiringActivityNotAllowedToRun errors
  </td></table>
  <br />
 

### Issue

the stopWork() method of ExpiringActivityManager is called at two different points:
1. the task is cancelled via the cancel closure
2. in the expiring condition case when the activity is about to expire

This causes a race condition whichever runs first will cancel and nil the task
The second will raise and error ExpiringActivityNotAllowedToRun

**Solution:**

Differentiate the two cases and ignore the ExpiringActivityNotAllowedToRun error on sync 

### Testing

As it's a race condition difficult to know, it seems looking at the logs, it involves a call.

---

### Checklist

---

### DIFF
--- a/wire-ios-system/Source/ExpiringActivity.swift
+++ b/wire-ios-system/Source/ExpiringActivity.swift
@@ -46,10 +46,14 @@ public func withExpiringActivity(reason: String, block: @escaping () async throw
     try await manager.withExpiringActivity(reason: reason, block: block)
 }
 
+/// Single-use: one instance per activity. The only entry point is the
+/// `withExpiringActivity(reason:block:)` free function above, which creates a
+/// fresh manager every call. Members below are `private` to enforce this — the
+/// actor's idempotency tracking (`didStartWork`) is not designed to reset.
 actor ExpiringActivityManager {
 
     let api: any ExpiringActivityInterface
-    var task: Task<Void, any Error>?
+    private var task: Task<Void, any Error>?
     // Stored as actor state so resumeOnce() is serialised through the actor
     // executor — no lock required, no threads blocked.
     private var continuation: CheckedContinuation<Void, any Error>?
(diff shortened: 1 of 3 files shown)


KaterinaWire (approved):

wire-ios-system/Source/ExpiringActivity.swift
```diff
@@ -112,6 +116,7 @@ actor ExpiringActivityManager {
     }
 
     func startWork(block: @escaping () async throws -> Void, semaphore: DispatchSemaphore) -> Task<Void, any Error> {
+        didStartWork = true
```
David-Henner, line 126: I assume this is never set back to `false` because we create a new manager for every expiring activity? How do we protect against misuse if someone decides to use the same manager instance for multiple activities ?

wire-ios-system/Tests/ExpiringActivityTests.swift
```diff
@@ -112,13 +112,18 @@ class ExpiringActivityTests: XCTestCase {
 
     // MARK: - Double-resume regression tests (crash: EXC_BREAKPOINT on CheckedContinuation.resume)
 
-    /// Regression test for the crash where outer-task cancellation and the expiry callback
-    /// both tried to resume the continuation:
+    /// Regression test for the bug where outer-task cancellation and the expiry callback
+    /// both invoked `stopWork()`, and the second call threw `ExpiringActivityNotAllowedToRun`
+    /// because `self.task` had already been nilled by the first call:
```
David-Henner: I find this comment confusing. If I was looking at the tests after this PR was merged, and looking at the code, I'd have no idea where this comes from because the code responsible for the bug has been changed

wire-ios-system/Source/ExpiringActivity.swift
netbe: Pushed a new commit that makes `task`, `startWork`, and `stopWork` `private` to enforce this and added a comment. That would at least document how to use it

wire-ios-system/Tests/ExpiringActivityTests.swift
netbe: Agreed — the original wording described code that no longer exists. Rewrote both regression-test doc comments to describe the race each test guards against in terms of current behavior, without re-narrating the deleted implementation.

David-Henner (approved):

PR #4734, opened 2026-05-19
fix: invalid cookie data in NSE - WPB-24202

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24202" title="WPB-24202" target="_blank">BugWPB-24202</a>  [iOS] Push Issue - Message only appears when opening the app
  </td></table>
  <br />
 

### Issue

Some users are experiencing an issue with notifications not being delivered. Looking through their logs we see repeated issues with `invalidCookieData`.

> [notifications] - [ERROR] Failed to fetch cookies: invalidCookieData(reason: "Error Domain=NSCocoaErrorDomain Code=4864 \"*** -[NSKeyedUnarchiver _initForReadingFromData:error:throwLegacyExceptions:]: incomprehensible archive..."

In our code base this means that we have fetched the cookie data from the keychain but we fail to convert this to cookie data using the NSKeyedUnarchiver. The data should be in the `bplist` format but it is not (see my comment in the ticket for more info).

From testing this is the exact behavior we would see if we tried to decrypt the cookie with an incorrect cookie key. However strangely this issue only affects the NSE not the main app which both use the same cookie key stored in shared user defaults. The cookie key is created lazily if user defaults doesn't already have one set.

Searching online I found cases where UserDefaults would return nil for an already set value and the best guess at the cause was related to permissions - UserDefaults being accessed before the device was first unlocked after a restart.

In the logs for a user I looked at this happens over and over again. Additionally of note we also have this line for each time a notification is received with the **exact same process id**:

> [INFO] loading new notification service[process_id:319][process_name:Wire Notification Service Extension]

So it is possible that these repeated failures are all happening in the same process.

#### A hypothesis

This PR attempts to do a possible fix of the following hypothesis that I have not managed to reproduce.

1. Device is restarted
2. A notification is received before the device pin has been entered.
3. The cookies key is requested to instantiate CookiesStorage.
4. Although there is already a cookie key in UserDefaults (a new one is created on first run after original install), the cookie key is unavailable because the device has not yet been unlocked - therefore a new incorrect key is returned.
5. As the process is for some reason kept alive each new notification also fails as we continue using the incorrect cookie key.

The fix here is to fail early if we don't have an existing cookie key. 

### Testing

Ensure that push notifications are still working.

---

### Checklist


wire-ios/Wire Notification Service Extension/NotificationService.swift
```diff
@@ -105,6 +105,14 @@ final class NotificationService: UNNotificationServiceExtension {
             return nil
         }
 
+        guard let cookiesKey = UserDefaults.existingCookiesKey else {
+            WireLogger.notifications.critical(
+                "no cookie encryption key, not loading service",
+                attributes: .safePublic
+            )
```
johnxnguyen: suggestion: maybe it's worth stating in the log that this may be expected if the phone restarted and hasn't been unlocked yet.

johnxnguyen (approved):

wire-ios/Wire Notification Service Extension/NotificationService.swift
samwyndham: I changed it to `warn` level as it may not exactly be an error but might surprise users.

johnxnguyen (approved):

netbe (approved):
LGTM:)

PR #4738, opened 2026-05-20
fix: prevent main app from blocking notification extension (take 3) - WPB-23511

johnxnguyen:
<table>
<td>
  </td></table>
  <br />
 

### Issue
This is another attempt to fixing an issue where Core Crypto transactions can sometimes be suspended before completing, preventing CCs internal process locking mechanism from releasing the lock, resulting in other processes being unable to perform transactions.

The previous attempt https://github.com/wireapp/wire-ios/pull/4689 used background tasks when performing Core Crypto transactions, but it was observed that it didn't always work. On top of that, the background task API was only available in the main app target and not the extensions.

This PR builds on the previous attempt by using the expiring activity API via `ExpiringActivityManager`, which is available in all app processes.

For more information about the issue and testing, see #4689 .

---

### Checklist

---


wire-ios-system/Source/ExpiringActivity.swift
johnxnguyen, line 1: Changes in this file: adjust the function signature to allow the block to return a value.

wire-ios-data-model/Source/Core Crypto/SafeCoreCrypto.swift
```diff
@@ -62,30 +44,13 @@ public final class SafeCoreCrypto {
...
-    private func transaction<Result>(
-        backgroundTaskManager: any BackgroundTaskManager,
-        block: @escaping (any CoreCryptoContextProtocol) async throws -> Result
+    ///
+    public func transaction<Result>(
+        block: @escaping @Sendable (any CoreCryptoContextProtocol) async throws -> Result
     ) async throws -> Result {
-        let transactionTask = Task {
+        try await withExpiringActivity(reason: "core crypto transaction") { [coreCrypto] in
             try await coreCrypto.transaction(block)
```
samwyndham, line 52: question: Do we know that cancellation of this task will actually cancel the underlying transaction?

samwyndham (approved):
🙏

wire-ios-data-model/Source/Core Crypto/SafeCoreCrypto.swift
johnxnguyen: so it depends on the block, whether it has cancellation checks. I looked and found that the vast majority of blocks only wrap a single invocation to core crypto, so we can't add any additional cancellation checks. There are a couple related to E2EI that do a bit more which may need to have some cancellation checks but I would look at that separately.

netbe (approved):
LGTM;)

PR #4742, opened 2026-05-20
fix: allow guests access to shared drive - WPB-25247

jullianm:
WPB-25247

### Issue

This PR allows guests to access conversations shared drives even if they're on a team with drive disabled.
Also added UI test.

### Testing

- Have userA on drive enabled team
- Have userB from another team (drive disabled)
- Connect both users
- UserA creates a drive enabled group conv with userB
- UserB should have access to shared drive on this group conv (drive tab should not show though)

### Checklist

---


WireUI/Sources/WireLocators/Locators.swift
```diff
@@ -60,6 +60,7 @@ public enum Locators {
 
         case bottomBarRecentListButton
         case bottomBarSettingsButton
+        case bottomBarDriveButton = "Drive"
```
WilhelmOks: why not leave it with the implicit string value?

WireUI/Sources/WireLocators/Locators.swift
```diff
@@ -174,6 +175,8 @@ public enum Locators {
         case recordingTime
         case showOtherRowButton
         case pingButton
+        case guestsArePresent = "Guests are present"
```
WilhelmOks, line 178: why not leave it with the implicit string value?

WilhelmOks (approved):
lgtm, just left some questions

WireUI/Sources/WireLocators/Locators.swift
jullianm: fixed

WireUI/Sources/WireLocators/Locators.swift
jullianm: accessibility IDs are not set on the related component and adjusting it would require a bit of refactoring of that component itself. I think it is out of scope of this PR so I'm simply using the banner text here.

OlivellaO (approved):
LGTM

PR #4745, opened 2026-05-20
fix: e2ei certificate update flow - WPB-24855

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24855" title="WPB-24855" target="_blank">BugWPB-24855</a>  [iOS] [Security] Certificate renewal delay occurs when renewing expired E2EI certificate by user
  </td></table>
  <br />
 

### Issue

- The e2ei certificate enrollment flow has no activity indicators between the different steps. Under bad network conditions, it appears as if nothing is happening for a while. 
- The "Update Certificate" popup can be displayed more than once.

### Solution

- Add an activity indicator between each e2ei certificate enrollment flow step
- Prevent the popup from being shown during the certificate update flow

### Changes

- Introduced `LifecycleAwareUserAgent` to wrap the different agents conforming to `OIDExternalUserAgent` and provide callbacks on the agent's presentation and dismissal
- Introduced `E2EIEnrollmentFlow` to manage the presentation of activity indicators 
- Updated all the certificate enrollment flows to rely on `E2EIEnrollmentFlow` for a consistent UX
- Added a flag in the certificate update flow to prevent showing the "Update Certificate" popup when the flow is ongoing
- Added a debug action to trigger the "Update Certificate" popup
- Added tests

### Testing

- Log in with a user in a team with e2ei enabled
- Use the developer tools to enroll an e2ei certificate with short expiration time
- Background & foreground the app, the "Update Certificate" popup should appear
- Verify that the popup doesn't appear more than once when reopening the app
- Update the certificate and verify that the activity indicators appear before & after the web view 

---

### Checklist


wire-ios/Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
```diff
@@ -865,7 +865,8 @@ extension AuthenticationCoordinator {
 
         Task { @MainActor in
             do {
-                let certificateChain = try await e2eiCertificateUseCase.invoke(authenticate: oauthUseCase.invoke)
+                let certificateChain = try await e2eiCertificateUseCase
```
netbe, line 868: thought: I wonder if the oauthUseCase should be injected inside the first one instead of plain block

wire-ios/Wire-iOS/Sources/UserInterface/Blacklist/BlockerViewController.swift
```diff
@@ -380,10 +380,19 @@ extension BlockerViewController {
         }
 
         let oauthUseCase = OAuthUseCase(targetViewController: { rootViewController })
+        let enrollmentFlow = E2EIEnrollmentFlow(
+            oauthUseCase: oauthUseCase,
+            targetVC: { rootViewController }
+        )
+
+        enrollmentFlow.showActivityIndicator()
+        defer { enrollmentFlow.dismissActivityIndicator() }
```
netbe: question: is it intended that it dismisses at the end of the function ?
Maybe using a do catch instead just dismiss once

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/LifecycleAwareUserAgent.swift
```diff
@@ -0,0 +1,63 @@
...
+        self.onDismissed = onDismissed
+    }
+
+    func present(
+        _ request: any OIDExternalUserAgentRequest,
+        session: any OIDExternalUserAgentSession
+    ) -> Bool {
+        let didPresent = wrapped.present(request, session: session)
+        if didPresent, let onPresented {
+            MainActor.assumeIsolated { onPresented() }
```
netbe, line 55: question: can you explain why we need assumeIsolated ?

netbe (approved):
Looks good, left a couple questions

wire-ios/Wire-iOS/Sources/UserInterface/Blacklist/BlockerViewController.swift
```diff
@@ -380,10 +380,19 @@ extension BlockerViewController {
...
 
         let certificateChain = try await activeUserSession
             .enrollE2EICertificate
-            .invoke(authenticate: oauthUseCase.invoke)
+            .invoke(authenticate: enrollmentFlow.authenticate)
+
+        enrollmentFlow.dismissActivityIndicator()
```
KaterinaWire: **question:** why do we call `enrollmentFlow.dismissActivityIndicator()` here if we do the same in the defer block?

wire-ios/Wire-iOS/Sources/UserInterface/DeviceView/DeviceInfoViewModel.swift
```diff
@@ -141,14 +141,12 @@ final class DeviceInfoViewModel: ObservableObject {
 
     @MainActor
     func enrollClient() async {
-        isActionInProgress = true
```
KaterinaWire, line 144: **question:** why did you remove `isActionInProgress` ? We still use `isActionInProgress`  in the `OtherUserDeviceDetailsView` and `DeviceInfoViewModel`.

wire-ios/Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
David-Henner: maybe? I'm not interested in changing this since it seems to work well so far

wire-ios/Wire-iOS/Sources/UserInterface/E2EIdentity/LifecycleAwareUserAgent.swift
David-Henner: `OIDExternalUserAgent` is an Obj-C protocol with no actor annotation, so Swift treats `present` and `dismiss` as callable from any context. 

We can't annotate it `@MainActor`, but AppAuth always calls it on the main thread since it deals with UI. 

`assumeIsolated` is tells the compiler we're already on the main actor, so we can call the `@MainActor`-bound `onPresented()` and `onDismiss()` closures synchronously.

wire-ios/Wire-iOS/Sources/UserInterface/Blacklist/BlockerViewController.swift
David-Henner: done [c1896a3](https://github.com/wireapp/wire-ios/pull/4745/commits/c1896a31d83edecc4e17bcaac204ddd3b3630fbe)

wire-ios/Wire-iOS/Sources/UserInterface/Blacklist/BlockerViewController.swift
David-Henner: updated [c1896a3](https://github.com/wireapp/wire-ios/pull/4745/commits/c1896a31d83edecc4e17bcaac204ddd3b3630fbe)

wire-ios/Wire-iOS/Sources/UserInterface/DeviceView/DeviceInfoViewModel.swift
David-Henner: The purpose of `isActionInProgress` is to show an activity indicator in the views you mentioned, but since the certificate enrollment flow now manages its own activity indicator we need to avoid using it in the enrollment flow, otherwise we'd end up with two simultaneous spinners overlaying on each other.

KaterinaWire (approved):

PR #4749, opened 2026-05-21
fix: save SPM cache - WPB-25840

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25840" title="WPB-25840" target="_blank">BugWPB-25840</a>  [iOS] CI SPM cache not used properly
  </td></table>
  <br />
 

### Issue

Right now, if a run is cancelled the cache is not saved. the SPM cache can take up to 20 min to build mostly because of aws repo checkout size.

Instead of save cache if job is successful, save it right after cache is built.

* It now uses a single cache key to save time in checkout of packages but always run the resolution so the correct dependencies are used (datadog on|off countly on|off)
* Save cache only on develop (default branch) so all branches derived from it use it

### Testing

I did some tests running this PR and playground build. But only when merged to develop we can see the benefits

---

### Checklist

---


.github/actions/setup-test-environment/action.yaml
```diff
@@ -61,30 +55,8 @@ runs:
...
-    - name: Resolve Swift Package Dependencies
-      if: steps.cache-swift-packages.outputs.cache-hit != 'true'
-      shell: bash
-      run: |
-        cd "$REPO_ROOT" && xcodebuild \
-          -resolvePackageDependencies \
-          -disableAutomaticPackageResolution \
-          -clonedSourcePackagesDirPath "$PACKAGES_DIR"
-
-    - name: Restore DerivedData Cache
```
netbe, line 80: also removes the derived cache for now, as it took a lot of space

.github/workflows/_reusable_run_tests_parallel.yml
```diff
@@ -105,10 +104,9 @@ jobs:
 
       - name: Setup test environment
         uses: ./.github/actions/setup-test-environment
-        with:
-          scheme: ${{ matrix.scheme }}
 
       - name: "Test ${{ matrix.scheme }}"
+        timeout-minutes: 20
```
netbe:
```suggestion
        timeout-minutes: 15
```

johnxnguyen (approved):

caldrian (approved):
Did we consider importing AWS as pre-built binary?

PR #4754, opened 2026-05-22
fix: re-login on 4.16 branch - WPB-25092

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25092" title="WPB-25092" target="_blank">BugWPB-25092</a>  [iOS] Re-login with E2EI throws something went wrong on verification code and shows infinite loader
  </td></table>
  <br />
 

### Issue

This PR fixes re-login on 4.16 branch. This was also fixed in https://github.com/wireapp/wire-ios/pull/4596 but rather that cherry picking that significant change, I have done a minimal fix here.

While investigating I realized the underlying cause for this bug was different than I originally thought. This PR fixes the symptoms but doesn't attempt to fix the underlying cause (see https://wearezeta.atlassian.net/browse/WPB-25846) which still requires investigation and is not safe for the scope of 4.16.

The reason for this bug is that a cached cookie is not being deleted. Currently `ZMPersistentCookieStorage` caches cookies using a key that includes the backend server but when instantiating a `ZMPersistentCookieStorage` instance during the logout process we pass in an incorrect backend server URL so the cookie is not deleted.

The workaround is to stop using the server as part of the cache key. Within keychain we don't use the server for the key so it really makes no sense using it for the cached cookie as the account is already very very very likely to be unique.

In this PR I kept changes to a minimum so didn't change the API of `ZMPersistentCookieStorage` even though backend server is no longer used.

### Testing

See ticket.

---

### Checklist

---

### DIFF
--- a/wire-ios-transport/Source/Authentication/ZMPersistentCookieStorage.m
+++ b/wire-ios-transport/Source/Authentication/ZMPersistentCookieStorage.m
@@ -47,7 +47,6 @@ static dispatch_queue_t isolationQueue(void)
 
 @interface ZMPersistentCookieStorage ()
 
-@property (nonatomic, readonly) NSString *serverName;
 @property (nonatomic, readonly) NSArray<NSHTTPCookie *> *authenticationCookies;
 @property (nonatomic, readonly) BOOL useCache;
 
@@ -84,7 +83,9 @@ static dispatch_queue_t isolationQueue(void)
 {
     self = [super init];
     if (self) {
-        _serverName = [serverName copy];
+        // As a specific fix for https://wearezeta.atlassian.net/browse/WPB-25092 on 4.16 we just ignore `serverName`
+        // and don't change the API of `ZMPersistentCookieStorage` to have minimum impact on the codebase.
+        // In recent code `ZMPersistentCookieStorage` no longer exists.
         _userIdentifier = [userIdentifier copy];
         _useCache = useCache;
     }
@@ -136,11 +137,7 @@ static dispatch_queue_t isolationQueue(void)
 
 - (NSString *)cookieKey
 {
-    if (nil != self.accountName) {
-        return [[self.accountName stringByAppendingString:@"_"] stringByAppendingString:self.serverName];
-    } else {
-        return self.serverName; // Legacy and migration support
-    }
+    return self.accountName;
 }
 
 - (NSString *)accountName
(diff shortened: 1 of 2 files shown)


wire-ios-transport/Source/Authentication/ZMPersistentCookieStorage.m
```diff
@@ -136,11 +137,7 @@ - (NSDate *)authenticationCookieExpirationDate
 
 - (NSString *)cookieKey
```
samwyndham, line 138: The cookie key is only used for in-memory cache so is safe to chaneg.

wire-ios-transport/Source/Authentication/ZMPersistentCookieStorage.m
```diff
@@ -420,47 +417,4 @@ - (void)setRequestHeaderFieldsOnRequest:(NSMutableURLRequest *)request;
...
-
-#pragma mark – Legacy Storage Migration
-
-
-@interface ZMPersistentCookieStorageMigrator ()
-@property (nonatomic, readonly) NSUUID *userIdentifier;
-@property (nonatomic, readonly) NSString *serverName;
-@end
-
-@implementation ZMPersistentCookieStorageMigrator
```
samwyndham, line 433: This is for a very old migration and any existing cookie will have expired long ago as affected versions of the app have been long depricated

wire-ios-transport/Source/Authentication/ZMPersistentCookieStorage.m
netbe: question: so what would happen when user upgrade? Does this impact multiple accounts on different servers?

netbe (commented):

wire-ios-transport/Source/Authentication/ZMPersistentCookieStorage.m
samwyndham: @netbe 

> so what would happen when user upgrade?

It has no effect because this key is only used for the in memory cache, not by the Keychain (the keychain uses the account ID only) An upgrade would cause an app restart so the cache would be empty.

> Does this impact multiple accounts on different servers?

It would only cause issues if the same user ID existed on multiple servers, which is VERY unlikely as these are UUIDs. But if that was the case it would already cause issues as we don't include the server in the persisted keychain key, only the account ID.

David-Henner (approved):

netbe (approved):
LGTM:)

PR #4762, opened 2026-05-26
fix: Accepting an incoming call on a non‑active account - WPB-20468

KaterinaWire:
<table>
<td>
  </td></table>
  <br />
 

### Issue

Accepting an incoming call on a non‑active account sometimes shows only the green top banner instead of the full‑screen calling UI. Reproducible mainly with group calls.
When a call is accepted on a non‑active  account, the app switches accounts and the root view controller is replaced. The call UI is briefly presented during this transition, then lost as the root view controller swap completes. However, the router's internal state in `ActiveCallRouter` is left in an inconsistent state: `isActiveCallShown` is **true**, but `mainWindow.rootViewController.presentedViewController` is **nil**, the calling UI is no longer in the view hierarchy.
  
On the follow‑up `updateActiveCallPresentationState()` triggered from `AppRootRouter.applicationDidTransition(to: .authenticated)`, the guard !isActiveCallShown blocks a re‑presentation, so the user is left with only the call‑top‑overlay banner.

### Solution

In `presentActiveCall(for:animated:)`, before evaluating the existing guard, reconcile the router's bookkeeping against UIKit's actual state. If we believe the call UI is shown but it isn't actually attached to the window` (isActiveCallShown && !isCallUIActuallyPresented())`, reset the flags so the presentation can proceed cleanly. `isCallUIActuallyPresented()` checks the live `presentedViewController` chain, it's the ground truth, while the bool flags are optimistic bookkeeping that can drift during a rootViewController. swap.

### Testing

_
### Checklist

---


caldrian (approved):
👍

wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallController/ActiveCallRouter.swift
```diff
@@ -127,7 +127,20 @@ extension ActiveCallRouter: ActiveCallRouterProtocol {
...
+        }
+        return presentedActiveCallViewController.presentingViewController != nil
+    }
+
     func presentActiveCall(for voiceChannel: VoiceChannel, animated: Bool) {
+        if isActiveCallShown, !isCallUIActuallyPresented() {
+            isActiveCallShown = false
+            isPresentingActiveCall = false
+            presentedActiveCallViewController = nil
+        }
```
David-Henner, line 154: suggestion: add a comment here to give some context on why this is needed, explaining the bug we're fixing and maybe adding the ticket reference.

David-Henner (approved):
approving with a comment 👍

wire-ios/Wire-iOS/Sources/UserInterface/Calling/CallController/ActiveCallRouter.swift
KaterinaWire: Done

PR #4771, opened 2026-05-27
fix: calling UI hangs - WPB-24483

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24483" title="WPB-24483" target="_blank">BugWPB-24483</a>  [iOS] Signaling failure & App hangs on "Hanging up"
  </td></table>
  <br />
 

### Issue

When a user joins an MLS group call, `setUpMLSConference` awaits `mlsService.createOrJoinSubgroup(...)` to join (or create) the conference subgroup. This is a network-bound operation, and on slow or unresponsive backends it can hang indefinitely. While that `await` is in flight:
 - the call UI is stuck on "Connecting…".
- tapping "end call" has no effect, the in-flight `Task` continues, `avsWrapper.startCall` / `answerCall` was never invoked, and the call snapshot remains in `.outgoing` / `.answered` so the UI has no terminating state to react to.

### Solution

Race `createOrJoinSubgroup` against a 10 second timeout using `withThrowingTaskGroup`. If the timeout wins, `Failure.mlsConferenceSetupTimeout` is thrown and caught by the existing error handler, which routes through `onMLSConferenceFailure` → posts `.terminating(.unknown)` → clears the snapshot → leaves the subconversation. The UI returns to a closable state and the user can retry the call.

  **Notes:**
  - The timeout is enforced before AVS is told to start/answer the call, so we never end up with a half-set-up conference.
 
### Testing

Use Network Link Conditioner to join the call.

### Checklist

---


jullianm (approved):
LGTM

wire-ios-sync-engine/Source/Calling/WireCallCenterV3.swift
```diff
@@ -740,7 +743,8 @@ public extension WireCallCenterV3 {
             Task {
                 do {
                     // Join the subgroup or create it if it doesn't exist
-                    let subgroupID = try await mlsService.createOrJoinSubgroup(
+                    let subgroupID = try await self.createOrJoinSubgroupWithTimeout(
```
netbe, line 746: suggestion: retry up to 3 times with a backoff could also help

netbe (approved):
left a suggestion

PR #4417, opened 2026-03-10
chore: delete unused code - WPB-20362

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20362" title="WPB-20362" target="_blank">TaskWPB-20362</a>  [iOS/Integration] - Searching for Apps specifically
  </td></table>
  <br />
 

### Issue

This PR deletes some unused code and contains some cosmetic changes.

---

### Checklist

- [ ] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [ ] Description is filled and free of optional paragraphs.
- [ ] Adds/updates automated tests.

---

### UI accessibility checklist

_If your PR includes UI changes, please utilize this checklist:_
- [ ] Make sure you use the API for UI elements that support large fonts.
- [ ] All colors are taken from WireDesign.ColorTheme or constructed using WireDesign.BaseColorPalette.
- [ ] New UI elements have Accessibility strings for VoiceOver.

PR #4781, opened 2026-05-29
fix: show profile pictures in search - WPB-25967

marcoconti83:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25967" title="WPB-25967" target="_blank">BugWPB-25967</a>  [iOS] No profile pictures in search results
  </td></table>
  <br />
 

### Issue
Steps To Reproduce
- Log in to iOS
- Go to the search (for people)
- Search for not-connected, not-in-the-same team users

Actual Behaviour
- I see no profile picture in the search result (for non-connected, not in the same team users)
- If I click on the search result, the send connection request screen opens, and I see no profile picture there either

Expected Behaviour
- I see the profile picture in search result and in the send connection request screen

### Testing

Added unit tests, and I tested it manually on a local build on staging

### Notes

It seems that this functionality was previously there, but was removed. Commit 16faeb5ed5 ("chore: delete unused code - WPB-20362", #4417, merged 2026-03-11) deleted `wire-ios-sync-engine/Source/Synchronization/Strategies/SearchUserImageStrategy.swift` and removed its registration from
`wire-ios-sync-engine/Source/Synchronization/StrategyDirectory.swift`


netbe (approved):
Looks good, I wonder if we could add an assertion in one of the UITest to prevent this regression. @findms what do you think?

findms:
> Looks good, I wonder if we could add an assertion in one of the UITest to prevent this regression. @findms what do you think?

yep.make sense. will add a task to add this assertion in a UItest or if needed will add a testcase.

caldrian (approved):

PR #4807, opened 2026-06-03
fix: silence in-app ringtone on inactive accounts - WPB-20468

caldrian:
<table>
<td>
  </td></table>
  <br />
 

### Issue

- Gate `SoundEventListener` ringtone playback on the active session. Inactive accounts no longer play the in-app ringtone for incoming calls — the system notification's sound is the audible signal for them.
- When this session becomes the active one (e.g. after the user switches accounts), re-evaluate call state and start the ringtone if a call is still ringing. When the session loses active status while ringing, stop the ringtone.
- Track `didStartRingtone` per listener so that one session's call-end doesn't silence another session's active ringtone (`AVSMediaManager` is a singleton).
- Adds `Notification.Name.sessionManagerActiveUserSessionDidChange`, posted from `SessionManager.setActiveUserSession`, consumed by `SoundEventListener`.

### Testing

- Two accounts, CallKit disabled. Active on A. Trigger an incoming call to B. Confirm: system notification sound plays once, no in-app ringtone overlap, no persistent in-app ringtone while still on A.
- Tap the notification (default tap). Switch to B. Confirm: in-app ringtone starts on B as the call UI appears.
- Accept/decline from the call UI. Confirm: ringtone stops.
- Active on A, call ringing on A (in-app ringtone playing). Switch to B (no call on B). Confirm: A's ringtone stops on switch.
- Active on A with a call established, second incoming call on B. Confirm: A's call audio is unaffected by B's call lifecycle.

---

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
+++ b/wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
@@ -167,6 +167,13 @@ public protocol SessionManagerSwitchingDelegate: AnyObject {
     func confirmSwitchingAccount(completion: @escaping (Bool) -> Void)
 }
 
+public extension Notification.Name {
+    /// Posted by `SessionManager` after the active user session changes. The notification's
+    /// `object` is the new active `ZMUserSession?` (may be `nil`).
+    static let sessionManagerActiveUserSessionDidChange = Notification
+        .Name("SessionManagerActiveUserSessionDidChange")
+}
+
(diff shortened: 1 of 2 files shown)


caldrian:
Closing this PR since fixing the issues around ringing
- with/without notification permissions,
- using/not using CallKit,
- being called on the active account vs on a background account,
- having the app in foreground or background
is a much bigger task.

PR #4810, opened 2026-06-04
fix: present incoming call UI after account switch - WPB-20468

caldrian:
<table>
<td>
  </td></table>
  <br />
 

### Issue

- Default tap on an incoming-call notification used to fall through to `showContent`, which navigated to the conversation. That competed with the post-activation call-UI presentation, producing a brief call UI flicker before landing
on the conversation with a visible "Join call" button.
- For default-tap responses on the `incomingCall` category, switch to the target account but skip the conversation navigation. `AppRootRouter`'s `.authenticated` transition then drives `updateActiveCallPresentationState()` to present
the incoming-call UI — the same path used when the call arrives while the account is already active.
- Adds `SessionManagerType.activateAccount(of:)` — a navigation-free account switch — and its `MockSessionManager` counterpart.
- Detects the call category via both the legacy `PushNotificationCategory.incomingCall` ("incomingCallCategory") and the new `WireDomain.NotificationCategory.incomingCall` ("incomingCall") raw values, since both paths produce call
notifications.

### Testing

- Two accounts, CallKit disabled. Active on A; receive an incoming call on B.
- Tap the incoming-call notification (default tap, not the "Ignore" action).
- Confirm: switches to B, presents the incoming-call UI immediately, no flicker, no navigation to the conversation.
- Accept / decline from the call UI — same behavior as if B had been the active account when the call arrived.
- Tap "Ignore" on the notification: call rejected, no UI navigation (existing behavior unaffected).
- Default tap on a regular message notification: still navigates to the conversation (no regression).

---

### Checklist

---

### DIFF
--- a/wire-ios-sync-engine/Source/SessionManager/SessionManager+Push.swift
+++ b/wire-ios-sync-engine/Source/SessionManager/SessionManager+Push.swift
@@ -163,6 +163,10 @@ public extension SessionManager {
         }
     }
 
+    func activateAccount(of session: ZMUserSession) {
+        activateAccount(for: session) {}
+    }
+
     func showUserProfile(user: UserType) {
         presentationDelegate?.showUserProfile(user: user)
     }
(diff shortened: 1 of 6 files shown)


jullianm (approved):
Nice work

samwyndham (approved):
Nice work!

PR #4823, opened 2026-06-08
chore: separating critical tests with all tests to reduce run time on PR - WPB-25757

findms:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25757" title="WPB-25757" target="_blank">TaskWPB-25757</a>  [iOS] reduce UITests execution time
  </td></table>
  <br />
 

### Issue
Currently all tests runs on PR which takes a huge amount of time and occasionally cancels due to timeout as reached 90 min threshold. 

**Changes**

Added fastlane support for UI test tags using comments like:
```
/// [critical]
@MainActor
func testExample_TC_1234() async throws
```
By this change, all the tests marked as critical will run on PR but all the tests on nightly.

### Testing

Ran on CI https://github.com/wireapp/wire-ios/actions/runs/27137194608
Total time took ~ 48 mins
image

image

---

### Checklist

---


wire-ios/WireUITests/PersonalUserTests.swift
```diff
@@ -18,8 +18,10 @@
 
 import XCTest
 
+/// core
```
samwyndham: question: What does the `/// core` comment do?

.github/workflows/_reusable_run_tests.yml
```diff
@@ -151,8 +155,14 @@ jobs:
...
       - name: "Test new critical flows"
         if: ${{ inputs.run_critical_flows }}
+        env:
+          UI_TEST_TAGS: ${{ inputs.ui_test_tags }}
         run: |
-          bundle exec fastlane run_uitests
+          if [ -n "$UI_TEST_TAGS" ]; then
+            bundle exec fastlane run_uitests tags:"$UI_TEST_TAGS"
+          else
+            bundle exec fastlane run_uitests
```
netbe, line 168: question: so this runs all UITests right?

.github/workflows/critical_flows.yaml
```diff
@@ -25,7 +30,8 @@ jobs:
       all: false
       notify_secret: WIRE_IOS_QA_WEBHOOK
       test_dependencies: false
-      run_critical_flows: true
+      run_critical_flows: ${{ inputs.run_critical_uitests }}
```
netbe: suggestion: change it to run_uitests

fastlane/Fastfile
```diff
@@ -599,7 +614,114 @@ platform :ios do
             skip_package_dependencies_resolution: true,
             number_of_retries: build.number_of_retries,
             xcargs: ci_args
-        )
+        }
+        run_tests_options[:only_testing] = options[:only_testing] if options[:only_testing]&.any?
```
netbe, line 623: question: what does options[:only_testing]&.any? the `&any?` do?

wire-ios/WireUITests/BackupRestoreHistoryTests.swift
```diff
@@ -19,8 +19,10 @@
 import WireFoundation
 import XCTest
 
+/// core
```
samwyndham: thought: I wonder if it is worth having a more defined format for the tags to prevent false positives when identifying tags in cases where we may give tests other doc comments. Something simple like:

[critical]

fastlane/Fastfile
```diff
@@ -599,7 +614,114 @@ platform :ios do
...
+
+        ui_test_metadata.select do |test|
+            test_tags = test[:tags].to_set
+            tag_groups.all? { |group| group.any? { |tag| test_tags.include?(tag) } } &&
+                (blocked_tags & test_tags).empty?
+        end
+    end
+
+    def ui_test_metadata
+        test_root = File.expand_path("../wire-ios/WireUITests", __dir__)
```
netbe, line 668: thought: if we extend other targets to UITests we'll have to change this. nothing to change for now, just want to remember this

wire-ios/WireUITests/README
```diff
@@ -44,3 +44,39 @@ Look for the following line in the logs:
 ```
 
 Once the service is up, run the relevant UI tests locally.
+
+### Mark and run tagged UI tests
```
netbe, line 48: praise: updated documentation

netbe (approved):
Looks good, just left some comments for understanding.

suggestion: as only critical tests will be run on each PRs, should we modify the nightly job to run the rest OR all UI tests?

question: What happen to untagged tests they are not run for each PR, right ? I guess it's ok

fastlane/Fastfile
findms: Yes for now it scans only WireUiTests path to pick tests for only-testing.

.github/workflows/_reusable_run_tests.yml
findms: Yes for nightly with no tag it runs all as `bundle exec fastlane run_uitests`

wire-ios/WireUITests/BackupRestoreHistoryTests.swift
findms: Just t understand this @samwyndham ..you mean in place of critical we should like this [critical] and same for other tags?

fastlane/Fastfile
findms: So options[:only_testing] contains the tests to execute, and any? checks that the list is not empty. &. safely handles the case where no tests are passed (nil), by avoiding any error thrown.

wire-ios/WireUITests/PersonalUserTests.swift
findms: this core == core messenger scope ..maybe we should put as whole core-messenger rather than just core.

findms:
> Looks good, just left some comments for understanding.
> 
> suggestion: as only critical tests will be run on each PRs, should we modify the nightly job to run the rest OR all UI tests?
> 
> question: What happen to untagged tests they are not run for each PR, right ? I guess it's ok

@netbe 
yes we need to enable nightly again and with increased time as well like more than ~90 mins for nightly.
Yes. For any new test/or untagged test - they will be not be picked by PR but nightly as default.

**There is just 1 issue still which came while addressing the feedback: When we run UITests with Testiny run name, we want to run all tests but due to ~90mins limit it may be cancelled and testiny tests are not executed correctly.**

wire-ios/WireUITests/PersonalUserTests.swift
samwyndham: Sorry. I added this comment early before I read the updated documentation and forgot to delete it. Closing.

wire-ios/WireUITests/BackupRestoreHistoryTests.swift
samwyndham: Yes or some other format so that:

- We can still use regular comments / doc comments in our code and the regex parser can pick out the tags.
- For humans looking at the code it is clear that it isn't just a comment but has behavior.

findms: updated similar to this 
/// [critical]
/// [collaboration]
/// [critical, collaboration]

samwyndham (approved):

PR #4826, opened 2026-06-08
chore: adopt UIScene lifecycle - WPB-25097

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25097" title="WPB-25097" target="_blank">TaskWPB-25097</a>  [iOS] Adopt UISceneDelegate
  </td></table>
  <br />
 

### Issue

This PR migrates the app for the UIScene lifecycle. This is necessary because apple will stope supporting the legacy lifecycle with iOS 27 SDK and the app will crash on launch:

Considering we have big plans for our apps architecture, this PR attempts to do a kind of minimum to solve the immediate problem without resorting to big hacks. It only supports a single app instance so we don't gain anything from these changes accept compatibility with the future SDK. To achieve this the following changes were necessary:

- Introduced SceneDelegate that takes over some responsibilities from AppDelegate
- Launch options are now bundled in UIScene.ConnectionOptions - an app can be launched with an array of URLs. In our case only the 1st URL is supported.
- Deleted the `ApplicationLaunchType`. This abstraction is hard to make work with the UIScene lifecycle and it was only being used in two places:
    - `CallQualityController` where a simple alternative was found
    - `SoundEventListener` from testing this code never hit the `appDelegate.launchType == .push` path (the behavior doesn't work) so I just removed the related code.

#### Snapshot updates

There were 4 screens whose snapshots needed to be update. I looked into this and I don't believe there was any changes in the app - rather they relate to how snapshots are rendered in the library - specifically the view.safeAreaInsets that are applied to the view controllers being tested by the library.

I decided just to update the snapshots as the safe area insets applied in the snapshots are not related to those applied in the actual app. 

### Testing

General testing of the app.

---

### Checklist


wire-ios/Wire-iOS/Wire-Info.plist
```diff
@@ -130,6 +130,25 @@
 	<array>
 		<string>RedactedScript-Regular.ttf</string>
 	</array>
+	<key>UIApplicationSceneManifest</key>
+	<dict>
+		<key>UIApplicationSupportsMultipleScenes</key>
+		<false/>
```
samwyndham, line 138: We don't support multiple scenes. Changing this would be a big task

WireLogging/Sources/WireLogging/WireLogger+Instances.swift
```diff
@@ -37,35 +39,34 @@ public extension WireLogger {
...
     static let mls = WireLogger(tag: "mls")
+    static let network = WireLogger(tag: "network")
     static let notifications = WireLogger(tag: "notifications")
     static let performance = WireLogger(tag: "performance")
+    static let proteus = WireLogger(tag: "proteus")
     static let push = WireLogger(tag: "push")
     static let pushChannel = WireLogger(tag: "push-channel")
-    static let webSocket = WireLogger(tag: "websocket")
-    static let proteus = WireLogger(tag: "proteus")
+    static let sceneDelegate = WireLogger(tag: "SceneDelegate")
```
samwyndham, line 58: This is the only content change in this file. Other than that I sorted the options.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+LifeCycle.swift
```diff
@@ -21,13 +21,6 @@ import WireLogging
 
 public extension ZMUserSession {
 
-    func application(
-        _ application: ZMApplication,
-        didFinishLaunching launchOptions: [UIApplication.LaunchOptionsKey: Any?]
-    ) {
-        startEphemeralTimers()
```
samwyndham, line 28: This was moved to the caller of this method as this method no longer makes sense.

wire-ios/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
```diff
@@ -249,8 +249,6 @@ final class ZClientViewController: UIViewController {
                 .appendingPathComponent(remoteIdentifier.uuidString, isDirectory: true)
         }
 
-        NotificationCenter.default.post(name: NSNotification.Name.ZMUserSessionDidBecomeAvailable, object: nil)
```
samwyndham, line 266: This is no longer observed.

wire-ios/Wire-iOS/Sources/AppDelegate.swift
```diff
@@ -66,28 +46,23 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        UIApplication.shared.connectedScenes.compactMap { $0.delegate as? SceneDelegate }.first
+    }
 
-    private(set) var mainWindow: UIWindow!
+    // MARK: - Public propertie
 
-    // Singletons
-    var unauthenticatedSession: UnauthenticatedSession? {
-        SessionManager.shared?.unauthenticatedSession
+    var mainWindow: UIWindow? {
```
samwyndham, line 59: This is now optional as its existence is a little later in the lifecycle.

wire-ios/Wire-iOS/Sources/AppDelegate.swift
```diff
@@ -29,32 +30,11 @@ import WireLogging
...
-
-    private let cookieStorage = CookieStorage(cookieEncryptionKey: UserDefaults.cookiesKey())
+    // MARK: - Private properties
 
-    private lazy var voIPPushManager: VoIPPushManager = .init(
-        application: UIApplication.shared,
-        pushTokenService: pushTokenService
-    )
-
-    private let pushTokenService = PushTokenService()
```
samwyndham, line 57: These have moved to `AppDependencies` so they are accessible here and from the SceneDelegate.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -639,9 +632,9 @@ public final class SessionManager: NSObject, SessionManagerType {
     }
 
     @MainActor
-    public func start(launchOptions: LaunchOptions) async {
+    public func start(connectionOptions: UIScene.ConnectionOptions) async {
         if
-            let url = launchOptions[UIApplication.LaunchOptionsKey.url] as? URL,
+            let url = connectionOptions.urlContexts.first?.url, // Currently we only support one URL
```
netbe, line 656: question: what's the url? same as launch url so deeplink?

wire-ios/Wire-iOS Tests/ReferenceImages/CallGridViewControllerSnapshotTests/testActiveSpeakersIndicators_Conference.1.png
netbe, line 1: question: why are snapshots recorded here?

wire-ios/Wire-iOS/Sources/SceneDelegate.swift
```diff
@@ -0,0 +1,296 @@
...
+    private let appStateCalculator = AppStateCalculator()
+    private var connectionOptions: UIScene.ConnectionOptions?
+
+    // MARK: - Public properties
+
+    private(set) var appRootRouter: AppRootRouter?
+
+    // MARK: - UISceneDelegate
+
+    var window: UIWindow?
```
netbe, line 40: suggestion: private set

wire-ios/Wire-iOS/Sources/SceneDelegate.swift
```diff
@@ -0,0 +1,296 @@
...
+            sessionManager: sessionManager,
+            appStateCalculator: appStateCalculator,
+            trackingManager: TrackingManager(
+                sessionManager: sessionManager,
+                availabilityChecker: .default,
+            ),
+            sceneConnectionOptions: connectionOptions
+        )
+
+        (UIApplication.shared.delegate as? AppDelegate)?.queueInitializationOperations()
```
netbe, line 198: question: should this be called inside the scenedelegate? maybe the launchOperations need some review. I wonder if some should be done even if a scene is not connected (background)

netbe (commented):
left some questions before approving

wire-ios/Wire-iOS Tests/ReferenceImages/CallGridViewControllerSnapshotTests/testActiveSpeakersIndicators_Conference.1.png
samwyndham: Oh right, I forgot to mention this in the description. I've updated the PR description. Please see there.

wire-ios/Wire-iOS/Sources/SceneDelegate.swift
samwyndham: I can't unfortunately because this is the UISceneDelegate API:

```swift
optional var window: UIWindow? { get set }
```

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
samwyndham: Yes exactly. I was not able to find the reason that this is now a collection of URLs. The best I could work out with the help of a LLM (e.g. maybe nonsense) is that:

1. An app could quickly try and launch our app with multiple calls in succession. The OS might batch these into a single instantiation.
2. Future proofing for future Apple plans.

I couldn't find a better explanation.

wire-ios/Wire-iOS/Sources/AppDelegate.swift
```diff
@@ -225,125 +124,39 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
-        )
-
-        switch launchType {
-        case .url,
-             .push:
-            break
-        default:
-            launchType = .direct
-        }
-    }
```
David-Henner, line 274: question: what replaces the app lifecycle events since these are removed ?

David-Henner (approved):

wire-ios/Wire-iOS/Sources/AppDelegate.swift
samwyndham: These app lifecycle are no longer called on the AppDelegate. However, the equivalent notifications still exist and can be observed.

wire-ios/Wire-iOS/Sources/SceneDelegate.swift
samwyndham: Conceptually launch options should really be handled by the AppDelegate as they shouldn't really be specific to a scene. I've taken some time to understand the launch operations better. As far as I understand these should really be done before we setup the appRouteRouter but there is some complexity here as access to things like group user defaults can be blocked by file protection. Making worthwhile changes here is somewhat risky and I don't want to really add additional risk to this current PR. Therefore I've created a Jira ticket to look into to this further: https://wearezeta.atlassian.net/browse/WPB-26555 and don't plan to address your comment in this PR.

netbe (approved):
LGTM:)

PR #4821, opened 2026-06-08
chore: pin GitHub actions to specific hashes - WPB-24577

samwyndham:
### Issue

This PR pins GitHub actions to specific hashes based on the existing version. If a version was something like `v4` then I have looked for the latest action starting v4 - (eg. v4.3.2) and pinned it to that hash.

### Testing

Run GitHub actions

---

### Checklist

- [x] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [x] Description is filled and free of optional paragraphs.
- [ ] Adds/updates automated tests.

PR #4842, opened 2026-06-10
chore: pin GitHub actions to specific hashes - WPB-24577

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24577" title="WPB-24577" target="_blank">TaskWPB-24577</a>  [iOS] set specific hash for github actions
  </td></table>
  <br />
 

### Issue

This PR pins GitHub actions to specific hashes based on the existing version. If a version was something like v4 then I have looked for the latest action starting v4 - (eg. v4.3.2) and pinned it to that hash as was done in https://github.com/wireapp/wire-ios/pull/4821 but to the 4.16 branch

### Testing

Run GitHub actions

---

### Checklist

### DIFF
--- a/.github/actions/allure-reporting/action.yaml
+++ b/.github/actions/allure-reporting/action.yaml
@@ -13,7 +13,7 @@ runs:
   steps:
     - name: Setup Node.js
       if: always()
-      uses: actions/setup-node@v4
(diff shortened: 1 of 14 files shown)


caldrian (approved):

netbe (approved):
LGTM;)

PR #4850, opened 2026-06-11
fix: audio message effect preview playback - WPB-24864

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24864" title="WPB-24864" target="_blank">BugWPB-24864</a>  [iOS] [Audio Messaging] Audio message cannot be played anymore when changing the effect during playback by user
  </td></table>
  <br />
 

### Issue

When recording an audio message and previewing it with effects, playback would break: after the first effect, subsequent effects were cut off or didn't play at all, often accompanied by ~0.6s UI hangs on each tap, until no effect could be played.

Root causes in `AudioEffectsPickerViewController`:

1. **Shared audio session churn (main bug).** The effect-preview player routed through the global `MediaPlaybackManager`/AVS. AVS reacts to every `.playing`/`.completed` by synchronously deactivating and reactivating the shared `AVAudioSession` on the main thread. Cycling effects hammered this deactivate→reactivate path, which blocked the main thread (the hangs) and intermittently failed with `AVAudioSession` error `560030580` ("Deactivating an audio session that has running I/O"), leaving the session wedged so later previews were silent.

2. **Stale apply completions.** Every effect is written to a single shared `effect.wav` and applied asynchronously on a background queue. Completions were applied unconditionally, so a completion superseded by a newer effect change could build an `AVAudioPlayer` from an already-overwritten file and notify the delegate with the wrong result.

Fix:

- The preview player no longer talks to `MediaPlaybackManager`/AVS. It manages a plain `AVAudioSession` itself: activates `.playback` once on play (idempotent across effect changes, no per-effect deactivation), and the picker releases the session on teardown after stopping the player.
- Added an `effectApplyGeneration` token so only the latest effect-apply completion plays and notifies the delegate; superseded completions are ignored.

### Testing

1. Open a conversation and record an audio message.
2. Tap an effect to preview it — it plays.
3. Change the effect repeatedly (including rapidly).

Expected: every effect preview plays in full, there are no per-tap hangs, and playback never wedges. Confirmed against the prior device logs where each switch produced `set_active=0 → could not set active: NO err=560030580 Session deactivation failed` plus a hang — those entries are gone.

---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioEffectsPickerViewController.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioEffectsPickerViewController.swift
@@ -16,6 +16,7 @@
 // along with this program. If not, see http://www.gnu.org/licenses/.
 //
 
+import AVFoundation
 import avs
 import UIKit
 import WireCommonComponents
@@ -59,6 +60,11 @@ final class AudioEffectsPickerViewController: UIViewController {
     var normalizedLoudness: [Float] = []
     private var lastLayoutSize = CGSize.zero
 
+    /// Identifies the most recent effect-apply request. All applied effects are written to the same
+    /// `effect.wav` file on a background queue, so a completion that is no longer the latest must not
+    /// play or hand its (about-to-be-overwritten) file to the delegate.
+    private var effectApplyGeneration = 0
+
     var selectedAudioEffect: AVSAudioEffectType = .none {
         didSet {
             if selectedAudioEffect == .reverse {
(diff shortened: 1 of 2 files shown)


wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioEffectsPickerViewController.swift
```diff
@@ -59,6 +60,26 @@ final class AudioEffectsPickerViewController: UIViewController {
...
+        _ effect: AVSAudioEffectType,
+        _ inputPath: String,
+        _ outputPath: String,
+        _ completion: @escaping () -> Void
+    ) -> Void
+
+    /// Applies an audio effect, writing the result to `outputPath` and calling `completion` on the main
+    /// thread once done. Injectable so tests can drive the completion deterministically: the real
+    /// `AVSAudioEffectType.apply` runs on a background queue and no-ops under tests, so the completion
+    /// (and the stale-completion guard around it) can't otherwise be exercised.
```
samwyndham: suggestion: Make these comments for terse. To me they don't feel like doc comments but rather comments that an LLM adds to explain how things are working.

samwyndham (approved):
LGTM although I feel like there is too much code comments. For example:

```
/// Injectable so tests can drive the completion deterministically: the real
/// `AVSAudioEffectType.apply` runs on a background queue and no-ops under tests, so the completion
/// (and the stale-completion guard around it) can't otherwise be exercised.
```

Can't this just be `Injectable for testing`.

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioEffectsPickerViewController.swift
samwyndham: Also is it true that the apply runs on a background queue?

johnxnguyen (approved):

wire-ios/Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioEffectsPickerViewController.swift
netbe:
```suggestion
    /// thread once done.
```

jullianm (approved):
LGTM

PR #4875, opened 2026-06-18
fix: show skin tone reactions in message details [WPB-26545]

MohamadJaara:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26545" title="WPB-26545" target="_blank">BugWPB-26545</a>  [iOS] emojis with skin color do not show in the message details screen 
  </td></table>
  <br />
 

## Summary

Fixes message details not showing reactions when the reaction emoji includes a skin tone modifier.

## Why this happened

The conversation cell renders reactions directly from the stored reaction string, so emoji such as `🙌🏻` are visible in the message bubble.

The message details screen used a different path: it tried to resolve every reaction through `EmojiRepository` before creating the reactions section. Skin tone variants are valid reaction strings, but they are not necessarily present as exact entries in the emoji repository. When lookup failed, the data source dropped the reaction entirely, so the Reactions tab showed the empty state even though the message had reactions.

## How it was fixed

The message details data source now keeps valid reaction rows even when the emoji repository cannot resolve the exact reaction string.

For skin tone emoji, it falls back to looking up the base emoji after removing Fitzpatrick skin tone modifiers, so the UI can still show a localized emoji name while preserving the original reaction emoji in the header.

For unknown or newer emoji that are not in the repository at all, the data source now displays the raw emoji with the user count instead of hiding the reaction.

## Tests

Added regression coverage for:
- a reaction with a skin tone modifier, e.g. `🙌🏻`
- a reaction missing from `EmojiRepository`, e.g. `🫡`

### DIFF
--- a/wire-ios/Wire-iOS Tests/MessageDetails/MessageDetailsViewControllerTests.swift
+++ b/wire-ios/Wire-iOS Tests/MessageDetails/MessageDetailsViewControllerTests.swift
@@ -328,6 +328,53 @@ final class MessageDetailsViewControllerTests: XCTestCase {
         verify(detailsViewController)
     }
 
+    func testThatItShowsReactionWithSkinToneModifier() throws {
+        // GIVEN
+        conversation = createGroupConversation()
+
+        let message = MockMessageFactory.textMessage(withText: "Message")
+        message.senderUser = SelfUser.provider?.providedSelfUser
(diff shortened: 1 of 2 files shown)


netbe (approved):
LGTM:) please add a jira ticket though

jullianm (approved):

PR #4880, opened 2026-06-22
fix: prevent main app from blocking notification extension (take 4) - WPB-23511

samwyndham:
<table>
<td>
  </td></table>
  <br />
 

### Issue

This is yet another attempt to fix issues with expiring activity. The key decisions this time are:

- From the main app (where UIApplication is available) always use the [beginBackgroundTask(expirationHandler:)](https://developer.apple.com/documentation/uikit/uiapplication/beginbackgroundtask(expirationhandler:)) API instead of the [performExpiringActivity(withReason:using:)](https://developer.apple.com/documentation/Foundation/ProcessInfo/performExpiringActivity(withReason:using:)) API. This requires no semaphore and in dummy app found that this worked perfectly fine with 1000 simultaneous tasks vs the 64 tasks of `performExpiringActivity(withReason:using:)`.
- Deleted the current implementation of `withExpiringActivity`. I replaced it with a similar method, `withBackgroundTask` that takes a `BackgroundTaskExecuter` which is a protocol. In the main app we pass in a `AppBackgroundTaskExecuter` which uses `the beginBackgroundTask(expirationHandler:)` API. From the app extensions we currently pass in a `PassthroughTaskExecuter` which does nothing but perform the task (but could be changed.) The idea is to allow different implementations per target.
- From tests I found that we must show the notification within a very short time after `UNNotificationServiceExtension.serviceExtensionTimeWillExpire` is called. Extending background time doesn't help here. However using `performExpiringActivity(withReason:using:` does appear to give us more time to allow for clean up. So in this implementation once `UNNotificationServiceExtension.serviceExtensionTimeWillExpire` is called I cancel the ongoing task using `performExpiringActivity(withReason:using:` to give us extra time for the cancellation to succeed.
- From the share extension I do something similar which is triggered when the the extension enters the background or the user hits cancel - basically give extra time for the cancellation to succeed.
- I intentionally didn't abstract cancellation using `performExpiringActivity` because I feel the necessary Semaphore usage is a hack so we shouldn't encourage it.

### Testing

@johnxnguyen Any ideas here? The only thing I can think of is to monitor logging while the app is being used.

---

### Checklist


wire-ios-data-model/Source/Core Crypto/SafeCoreCrypto.swift
```diff
@@ -41,51 +41,12 @@ public final class SafeCoreCrypto {
     public func transaction<Result>(
         block: @escaping (any CoreCryptoContextProtocol) async throws -> Result
     ) async throws -> Result {
-        if let backgroundTaskManager {
-            try await transaction(
-                backgroundTaskManager: backgroundTaskManager,
-                block: block
-            )
-        } else {
+        try await withBackgroundTask(
```
samwyndham, line 44: This is now always performed but will use a PassthroughTaskExecuter when called from an app extension.

wire-ios-system/Source/BackgroundTasks.swift
```diff
@@ -0,0 +1,40 @@
...
+) async throws -> T {
+    try await executer.execute(name: name, operation: operation)
+}
+
+/// A `BackgroundTaskExecuter` that simply executes the operation without any background task management.
+public struct PassthroughTaskExecuter: BackgroundTaskExecuter {
+
+    public init() {}
+
+    public func execute<T>(name: String?, operation: sending @escaping () async throws -> T) async throws -> T {
```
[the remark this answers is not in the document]
samwyndham: Fixed

wire-ios/Wire-iOS/Sources/AppBackgroundTaskExecuter.swift
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        let result = try await withTaskCancellationHandler {
+            try await task.value
+        } onCancel: {
+            task.cancel()
+        }
+        application.endBackgroundTask(taskID.value)
+
+        return result
```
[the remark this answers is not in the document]
samwyndham: Fixed in https://github.com/wireapp/wire-ios/pull/4880/commits/02d155e8545b069b090394a7081092978ad38db9

wire-ios-system/Source/BackgroundTasks.swift
```diff
@@ -0,0 +1,40 @@
...
+    func execute<T: Sendable>(name: String?, operation: sending @escaping () async throws -> T) async throws -> T
+}
+
+public func withBackgroundTask<T: Sendable>(
+    name: String? = nil,
+    executer: any BackgroundTaskExecuter,
+    operation: sending @escaping @isolated(any) () async throws -> T
+) async throws -> T {
+    try await executer.execute(name: name, operation: operation)
+}
```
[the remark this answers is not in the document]
samwyndham: Added tests for `AppBackgroundTaskExecuter` in 79e495e

wire-ios-system/Source/BackgroundTasks.swift
```diff
@@ -0,0 +1,43 @@
...
+/// A `BackgroundTaskExecuter` that simply executes the operation without any background task management.
+public struct PassthroughTaskExecuter: BackgroundTaskExecuter {
+
+    public init() {}
+
+    public func execute<T: Sendable>(
+        name: String?,
+        operation: sending @escaping () async throws -> T
+    ) async throws -> T {
+        try await operation()
```
netbe, line 40: question: should we check for cancellation here?

wire-ios/Wire-iOS Share Extension/Sources/Content/PostContent.swift
```diff
@@ -100,7 +101,25 @@ final class PostContent {
...
+
+        // Keep the share extension alive until posting content is cancelled.
+        ProcessInfo.processInfo.performExpiringActivity(withReason: "cancelling ongoing task") { isExpired in
+            if isExpired {
+                semaphore.signal()
+            } else {
+                _ = semaphore.wait(wallTimeout: .now() + .seconds(30))
+            }
+        }
     }
```
netbe, line 123: question: is there a way to factor out this logic with the one in NSE?

WireDomain/Sources/WireDomain/Notifications/Protocols/NotificationServiceProtocol.swift
```diff
@@ -1,28 +0,0 @@
...
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
-// GNU General Public License for more details.
-//
-// You should have received a copy of the GNU General Public License
-// along with this program. If not, see http://www.gnu.org/licenses/.
-//
-
-import UserNotifications
-
-public protocol NotificationServiceProtocol {
```
netbe, line 21: praise: cleanup

netbe (commented):
left some comments. I wonder the reasoning around leaving the CC transactions not wrapped. Does this count in backgroundActivity be long enough so filelocks (CC side) is released?

Looking forward to run this to see how it behaves

wire-ios/Wire-iOS Share Extension/Sources/Content/PostContent.swift
samwyndham: Yes for sure and I actually did that then reverted it. From the PR description:

> I intentionally didn't abstract cancellation using performExpiringActivity because I feel the necessary Semaphore usage is a hack so we shouldn't encourage it.

However I can easily bring it back if you think it would be better with a suitable scary name.

wire-ios-system/Source/BackgroundTasks.swift
samwyndham: I'm not sure it is necessary because it will be called immediately after `withBackgroundTask` is called. If it was necessary I think it would make more sense that the caller of `withBackgroundTask` checks before calling `withBackgroundTask` or they check within the `operation`.

wire-ios/Wire-iOS/Sources/AppBackgroundTaskExecuter.swift
```diff
@@ -0,0 +1,96 @@
...
+
+        let task = Task {
+            WireLogger.backgroundActivity.warn("will start background task: \(name)")
+            let result = try await operation()
+            WireLogger.backgroundActivity.warn("did end background task: \(name)")
+            return result
+        }
+
+        let taskID = TaskID()
+        taskID.value = application.beginBackgroundTask(withName: name) {
```
johnxnguyen, line 73: question: do we need to check or handle the case where the task id returned is `.invalid`?

johnxnguyen (commented):
Looks good, just one question

wire-ios/Wire-iOS/Sources/AppBackgroundTaskExecuter.swift
samwyndham: Discussed this with @johnxnguyen offline. We are not really sure in what cases `beginBackgroundTask` might return .invalid. Therefore for the time being we will log this as an error but won't prevent the task from running. Added this logging in https://github.com/wireapp/wire-ios/pull/4880/commits/743083c88ceb7a7d221c4087c6e269245505cd00

johnxnguyen (approved):

wire-ios/Wire-iOS Share Extension/Sources/Content/PostContent.swift
netbe: not necessary then, i guess if it's used a 3rd time that would be something to do

netbe (approved):
Looks good to me for now, let's see it in action

PR #4892, opened 2026-06-23
fix: not failing when missing snapshots - WPB-26403

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26403" title="WPB-26403" target="_blank">BugWPB-26403</a>  [iOS] CI won't fail for non-existent snapshot reference images
  </td></table>
  <br />
 

### Issue

Currently missing snapshot tests don't necessarily result in a test failure when running on CI. This PR fixes the issue by:

1. Updates `SnapshotHelper.defaultRecordMode` to first check if we are running in the CI
2. Change how the CI env is propagated. Now it is set as a env variable via Fastlane - `TEST_RUNNER_CI`. This is supported by `xcodebuild` as a way of passing env variables to test runners. The `TEST_RUNNER_` prefix is stripped. This means that we no longer need to add the `${CI}` env variable to test plans which is brittle and the underlying cause for this failure that led to this bug ticket being created.
3. Test CI env for a specific value "true" which is what is set by GitHub actions. We are in control of the CI used so it makes sense that we can test for specific values.
4. Remove CI env variables from test plans as they are no longer needed.

### Testing

Tested that it works in this job were I had deleted one snapshot test: https://github.com/wireapp/wire-ios/actions/runs/28037475222/job/82994846904?pr=4892#step:5:878

---

### Checklist

### DIFF
--- a/WireFoundation/Sources/WireTesting/Utilities/SnapshotHelper.swift
+++ b/WireFoundation/Sources/WireTesting/Utilities/SnapshotHelper.swift
@@ -33,10 +33,10 @@ public struct SnapshotHelper {
     private var snapshotReferenceDirectory = ""
 
     private var defaultRecordMode: SnapshotTestingConfiguration.Record? {
-        let CIEnv = ProcessInfo.processInfo.environment["CI"]
+        let ciEnv = ProcessInfo.processInfo.environment["CI"]
         let recordEnv = ProcessInfo.processInfo.environment["SNAPSHOT_TESTING_RECORD"]
 
-        if CIEnv == "true" {
+        if ciEnv == "true" {
             return .never
         } else if let recordEnv, let record = SnapshotTestingConfiguration.Record(rawValue: recordEnv) {
             return record
--- a/WireFoundation/Sources/WireTesting/Utilities/SnapshotHelper.swift
+++ b/WireFoundation/Sources/WireTesting/Utilities/SnapshotHelper.swift
@@ -36,7 +36,7 @@ public struct SnapshotHelper {
         let CIEnv = ProcessInfo.processInfo.environment["CI"]
         let recordEnv = ProcessInfo.processInfo.environment["SNAPSHOT_TESTING_RECORD"]
 
-        if CIEnv == "true" || CIEnv == "1" {
+        if CIEnv == "true" {
             return .never
         } else if let recordEnv, let record = SnapshotTestingConfiguration.Record(rawValue: recordEnv) {
             return record
--- a/WireFoundation/Sources/WireTesting/Utilities/SnapshotHelper.swift
+++ b/WireFoundation/Sources/WireTesting/Utilities/SnapshotHelper.swift
@@ -33,14 +33,15 @@ public struct SnapshotHelper {
     private var snapshotReferenceDirectory = ""
 
     private var defaultRecordMode: SnapshotTestingConfiguration.Record? {
-        let ci = ProcessInfo.processInfo.environment["CI"]
-        if let value = ProcessInfo.processInfo.environment["SNAPSHOT_TESTING_RECORD"],
-           let record = SnapshotTestingConfiguration.Record(rawValue: value) {
-            return record
-        } else if ci == nil || ci?.isEmpty == true {
-            return .missing
-        } else {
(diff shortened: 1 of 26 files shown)


WireNetwork/Tests/WireNetworkTests/Helpers/HTTPRequestSnapshotHelper.swift
```diff
@@ -25,8 +25,7 @@ import struct WireNetwork.HTTPRequest
 struct HTTPRequestSnapshotHelper {
 
     private var defaultRecordMode: SnapshotTestingConfiguration.Record? {
-        let ci = ProcessInfo.processInfo.environment["CI"]
-        return (ci == nil || ci?.isEmpty == true) ? .missing : .never
+        ProcessInfo.processInfo.environment["CI"] == "true" ? .never : .missing
```
netbe, line 28: suggestion: we should allow overriding the mode when running locally.
the .failing or .failed mode is interesting as it will re-record all failing snapshots which imo is a time saver

WireNetwork/Tests/WireNetworkTests/Helpers/RequestSnapshotter.swift
```diff
@@ -139,8 +139,7 @@ final class RequestSnapshotter {
     }
 
     private var defaultRecordMode: SnapshotTestingConfiguration.Record? {
-        let ci = ProcessInfo.processInfo.environment["CI"]
-        return (ci == nil || ci?.isEmpty == true) ? .missing : .never
+        ProcessInfo.processInfo.environment["CI"] == "true" ? .never : .missing
```
netbe, line 142: same here

netbe (approved):
looks good, I didn't know about TEST_RUNNER_.

left one suggestion, it might be good to document the behaviour (one line) in README or Agents.md

caldrian (approved):
Very nice. It's a shame I also didn't know about passing variables using the `TEST_RUNNER_` prefix

WireFoundation/Sources/WireTesting/Utilities/SnapshotHelper.swift
```diff
@@ -33,14 +33,15 @@ public struct SnapshotHelper {
     private var snapshotReferenceDirectory = ""
 
     private var defaultRecordMode: SnapshotTestingConfiguration.Record? {
-        let ci = ProcessInfo.processInfo.environment["CI"]
-        if let value = ProcessInfo.processInfo.environment["SNAPSHOT_TESTING_RECORD"],
-           let record = SnapshotTestingConfiguration.Record(rawValue: value) {
+        let ciEnv = ProcessInfo.processInfo.environment["CI"]
+        let recordEnv = ProcessInfo.processInfo.environment["SNAPSHOT_TESTING_RECORD"]
+
+        if ciEnv == "true" {
```
caldrian, line 39: suggestion
Actually, after having another thought I think we might want to adjust this check to not only compare against the string `"true"`.

Could we use for example [NSString.boolValue](https://developer.apple.com/documentation/foundation/nsstring/boolvalue)? 
I saw that you removed some code which set `"CI=1"`, that wouldn't work anymore, if I'm not mistaken.

samwyndham: Now Fastlane just passes through what is set by the CI for that value. For GitHub actions it is always "true". As we will not change CI that often to me it makes sense to just tayler it to the CI we are using. That way if someone searches for "CI" environment in GitHub documentation they will see that it is "true". NSString.boolValue looks interesting but it doesn't support "true" so we would need some mapping.

WireNetwork/Tests/WireNetworkTests/Helpers/HTTPRequestSnapshotHelper.swift
samwyndham: Good point. I will improve this in a follow up PR.

WireNetwork/Tests/WireNetworkTests/Helpers/RequestSnapshotter.swift
samwyndham: Good point. I will improve this in a follow up PR.

PR #4897, opened 2026-06-23
feat: upgrade to CoreCrypto v.10.0.0 - WPB-27418

typfel:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27418" title="WPB-27418" target="_blank">TaskWPB-27418</a>  [iOS] Adopt Core Crypto 10
  </td></table>
  <br />
 

### Issue
This PR integrates CoreCrypto 10.0.0, moving from the currently integrated version of 9.2.1. For details about what has changed, see the [Core Crypto Book](https://wireapp.github.io/core-crypto/v10.0.0/cc-book/cc10/whats-new.html).

### Testing
There are several changes introduced in Core Crypto 10, so a general test suite should be run. However, the largest changes relate to End-to-end-identity, which don't have much automated test coverage. Therefore it's recommended to manually test the E2EI flows.

---

### Checklist

---


wire-ios-data-model/Source/Proteus/ProteusService.swift
```diff
@@ -312,10 +312,7 @@ public final class ProteusService: ProteusServiceInterface {
 
     public var lastPrekeyID: UInt16 {
         get async {
-            let lastPrekeyID = try? await coreCrypto.transaction {
-                try $0.proteusLastResortPrekeyId()
-            }
-            return lastPrekeyID ?? UInt16.max
```
johnxnguyen, line 318: question: the [CC book](https://wireapp.github.io/core-crypto/v10.0.0/cc-book/cc10/migration-guide.html) says that `proteusLastResortPrekeyId` is now a static method on `CoreCrypto`. Why aren't we using it here?

wire-ios-data-model/Tests/MLS/MLSActionExecutorTests.swift
```diff
@@ -435,69 +376,41 @@ class MLSActionExecutorTests: ZMBaseManagedObjectTest {
...
-
-        // When
-        _ = try await sut.joinGroup(.random(), groupInfo: .random())
-
-        // Then
-        await fulfillment(of: [expectation], timeout: 1)
+//        XCTAssertEqual(
+//            mockJoinByExternalCommitArguments.first,
+//            mockGroupInfo // TODO jacob allow comparing group info
+//        )
```
johnxnguyen: This needs to be addressed

johnxnguyen (commented):
This is just a first pass, I'll take a second look again soon.

wire-ios-data-model/Tests/MLS/MLSActionExecutorTests.swift
johnxnguyen: removed because there's no way to make this assertion.

wire-ios-data-model/Source/Proteus/ProteusService.swift
johnxnguyen: fixed

valerio-bettini:
@johnxnguyen @typfel How can we test it to make sure it all works?

wire-ios-request-strategy/Sources/E2EIdentity/EnrollE2EICertificateUseCase.swift
```diff
@@ -58,26 +61,48 @@ public final class EnrollE2EICertificateUseCase: EnrollE2EICertificateUseCasePro
...
+    private let featureRepository: LegacyFeatureRepositoryInterface
+    private let keyRotator: E2EIKeyPackageRotating
+    private let coreCryptoProvider: CoreCryptoProviderProtocol
     private let context: NSManagedObjectContext
 
     // MARK: - Life cycle
 
     public init(
         e2eiRepository: E2EIRepositoryInterface,
+        apiVersion: APIVersion?,
```
samwyndham, line 89: thought: feels a bit weird that this can be nil. If it is nil I guess we are in a broken state? I think it is worth double checking if this can happen.

wire-ios-data-model/Source/Proteus/ProteusService.swift
```diff
@@ -312,10 +312,7 @@ public final class ProteusService: ProteusServiceInterface {
 
     public var lastPrekeyID: UInt16 {
```
samwyndham, line 313: nitpick suggestion: unless the async is needed:

```swift
public let lastPrekeyID =  UInt16.max
```

samwyndham (approved):
A lot of changes which I've looked at commit by commit but looks good as far as I can see but obviously we need some manual testing.

wire-ios-data-model/Source/Proteus/ProteusService.swift
samwyndham: @johnxnguyen I made this comment for a particular commit but then I see you changed the implementation a little. Any idea why @typfel made it a constant?

johnxnguyen: No I don't

wire-ios-request-strategy/Sources/E2EIdentity/EnrollE2EICertificateUseCase.swift
johnxnguyen: So I looked into this and this is what I've found:
- `EnrollE2EICertificateUseCase` is injected with an optional API version
- The use case is now initialized in `ZMUserSession`, the api version being injected is from `resolvedBackendMetadata.apiVersion`
- Previously at the exact same place where the use case is created, the `E2EIRepository` was being created with the exact same optional api version.
- The `E2EIRepository` no longer has the optional api version (and other fields) because they moved to the use case.

My conclusion is that nothing has really changed here. Having said that, it is not a nice state to have a nil api version but I believe this is a harder architectural issue to solve.

I'm inclined to think that this change is ok, given that it has essentially preserved existing behavior. What do you think?

samwyndham: Yes lets keep it as is.

johnxnguyen (approved):

PR #4913, opened 2026-06-24
chore: inline xcresult analysis, remove per-framework macOS runners - WPB-26561

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26561" title="WPB-26561" target="_blank">TaskWPB-26561</a>  [iOS] reduce CI bottleneck
  </td></table>
  <br />
 

### Issue

The `analyze-xcresults` job in `_reusable_run_tests.yml` spawned one dedicated macOS Cirrus CI runner (`ghcr.io/cirruslabs/macos-runner:sequoia`) per xcresult file (one per framework under test) via a matrix strategy. On a PR touching many frameworks this created N extra `sequoia` runner jobs purely for annotation — adding to the runner queue and wasting capacity.

The analysis is now an inline step inside the existing `run-tests` job. It parses all JUnit XML files produced during the test run and emits `::error` annotations via GitHub's workflow commands, covering every framework at zero additional runner cost.

### Testing

- Open a PR against `release/cycle-4.16` that touches at least one framework.
- Confirm the `analyze-xcresults` job no longer appears in the Actions run.
- Confirm test failure annotations still appear on the PR diff.

---

### Checklist

### DIFF
--- a/.github/workflows/_reusable_run_tests.yml
+++ b/.github/workflows/_reusable_run_tests.yml
@@ -180,6 +180,27 @@ jobs:
         run: |
           echo "xcresult files: ${{ steps.find-xcresults.outputs.xcresult_files }}"
 
+      - name: Annotate test failures
+        if: ${{ always() && github.event_name == 'pull_request' }}
+        continue-on-error: true
+        run: |
+          python3 -c "
+import xml.etree.ElementTree as ET, glob, sys
+files = glob.glob('build/reports/*.junit') + glob.glob('artifacts/**/*.junit', recursive=True)
+for f in files:
+    try:
+        for tc in ET.parse(f).iter('testcase'):
+            failure = tc.find('failure')
+            if failure is not None:
+                classname = tc.get('classname', '')
+                name = tc.get('name', 'Unknown')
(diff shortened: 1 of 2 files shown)


.github/workflows/_reusable_run_tests.yml
```diff
@@ -180,6 +180,27 @@ jobs:
         run: |
           echo "xcresult files: ${{ steps.find-xcresults.outputs.xcresult_files }}"
 
+      - name: Annotate test failures
+        if: ${{ always() && github.event_name == 'pull_request' }}
+        continue-on-error: true
+        run: |
+          python3 -c "
+import xml.etree.ElementTree as ET, glob, sys
```
caldrian: question:
I usually prefer to not have code in the workflow files. Do you think in this case we should extract it into a script or keep it as is?

netbe: yes to a script sounds better I agree

caldrian (approved):

samwyndham (approved):

PR #4916, opened 2026-06-24
fix: conversation wrongly marked as read - WPB-24856

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24856" title="WPB-24856" target="_blank">BugWPB-24856</a>  [iOS] [Messaging] Messages are not displayed correctly when receiving new chats by user
  </td></table>
  <br />
 

### Issue

New group conversations created while the app is backgrounded show no unread indicator after the user opens the app, even though the message is visible when entering the conversation.

### Causes

When a new group conversation is first fetched from the backend via `UpdateConversationItem`, `updateOrCreateGroupConversation` runs with `isInitialFetch = true` and unconditionally sets `lastReadServerTimeStamp = lastModifiedDate`. Because `lastModifiedDate` is derived from the conversation's `lastEventTime` — which the backend sets to the timestamp of the most recent event — it ends up equal to the incoming message's `serverTimestamp`.

When the stored `conversation.proteusMessageAdd` event is then processed, `calculateLastUnreadMessages` queries for messages where `serverTimestamp > lastReadServerTimeStamp`. Since both timestamps are equal, the message is excluded and `internalEstimatedUnreadCount` stays 0.

The `isInitialFetch` flag was originally intended to mark conversations as read during slow sync (fresh install / re-login), where no messages exist in Core Data yet and marking as read prevents badge flooding. However, `isInitialFetch` is true for any conversation first seen by the device, including newly created ones arriving via the NSE while the app is backgrounded.

### Solution

Added a `markAsRead: Bool` parameter to `storeConversation` and `updateOrCreateGroupConversation`. The `lastReadServerTimeStamp` assignment is now gated behind this flag. Only `PullAllConversationsSync` — the genuine slow sync path — passes `markAsRead: true`. All other callers, including `UpdateConversationItem` via `ConversationRepository`, pass `markAsRead: false`.

### Testing

**Requirements:** two Proteus accounts on the same backend.

**Steps:**
1. Log in as User A, background the app and lock the phone.
2. Log in as User B, create a new group conversation with User A and send a message.
3. Unlock the phone and open the Wire app as User A.

**Expected result:** the new conversation appears at the top of the list with an unread indicator.

---

### Checklist

---

### DIFF
--- a/WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift
+++ b/WireDomain/Sources/WireDomain/Repositories/Conversations/LocalStore/ConversationLocalStore.swift
@@ -469,7 +469,8 @@ public final class ConversationLocalStore: ConversationLocalStoreProtocol {
         _ conversation: Conversation,
         timestamp: Date,
         isFederationEnabled: Bool,
-        isMLSEnabled: Bool
+        isMLSEnabled: Bool,
+        markAsRead: Bool
     ) async {
         guard let conversationType = conversation.type else {
             return
@@ -500,7 +501,8 @@ public final class ConversationLocalStore: ConversationLocalStoreProtocol {
                 withID: id,
                 serverTimestamp: timestamp,
                 isFederationEnabled: isFederationEnabled,
-                isMLSEnabled: isMLSEnabled
+                isMLSEnabled: isMLSEnabled,
+                markAsRead: markAsRead
             )
 
         case .`self`:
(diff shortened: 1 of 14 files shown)


netbe (approved):
LGTM:)
I wonder if this impacts also the bugs on restore backup and unread messages

valerio-bettini (approved):

johnxnguyen (approved):

PR #4918, opened 2026-06-24
chore: reduce CI runner queue depth - WPB-26561

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26561">WPB-26561</a> [iOS] reduce CI bottleneck
</td>
</table>

### Issue

With 4 macOS runners and 53+ queued jobs, the CI pipeline was severely starved. Multiple root causes:

1. **No `max-parallel` cap** on the parallel test matrix — one runner per framework, so N frameworks × M concurrent PRs = N×M runner slots requested simultaneously.
2. **`analyze-xcresults` job** spawned one dedicated Cirrus CI `sequoia` macOS runner per xcresult file just for annotations.
3. **Superseded build runs** (Edge, Beta, C1/C2/C3) were not cancelled when a newer commit pushed to the same branch.

### Changes

**Bin-pack framework tests into batches (`_reusable_run_tests_parallel.yml`)**
- Added `max_parallel` input (default 3, leaving 1 runner free for build jobs)
- Fastlane `list_schemes_json` lane now bin-packs schemes into ≤`max_parallel` batches and writes `/tmp/batches.json`
- Matrix switches from per-scheme to per-batch; each batch runner executes its schemes sequentially
- Fixed step ID reference (`steps.bin-pack` → `steps.list-schemes`) that would have silently skipped all batched tests

**Replace `analyze-xcresults` matrix job with inline annotation step**
- Removed the separate job that used a matrix of Cirrus CI `sequoia` runners (one per xcresult)
- Added `ruby scripts/annotate_test_failures.rb` as an inline step in the existing test job
- Ruby script uses only `rexml/document` from stdlib — no gem dependencies
- Workflow command properties are correctly percent-encoded (`%`, `\r`, `\n`, `:`, `,`)

**Cancel superseded build runs**
- Added `concurrency: cancel-in-progress: true` keyed on `${{ github.workflow }}-${{ github.ref }}` to Edge, Beta, C1/C2/C3 Production, and C1/C2/C3 Restricted workflows
- A newer push/dispatch on the same branch cancels the older run; different release branches (`cycle-4.16` vs `cycle-4.22`) remain independent

### Testing

- Open a PR touching multiple frameworks and confirm at most 3 parallel macOS test jobs are active at once
- Confirm the `analyze-xcresults` job no longer appears in Actions runs
- Push two commits in quick succession to `develop` and confirm the first Edge build is cancelled

---

### Checklist

### DIFF
--- a/.github/workflows/_reusable_run_tests.yml
+++ b/.github/workflows/_reusable_run_tests.yml
@@ -225,6 +225,27 @@ jobs:
         run: |
           echo "xcresult files: ${{ steps.find-xcresults.outputs.xcresult_files }}"
 
+      - name: Annotate test failures
+        if: ${{ always() && github.event_name == 'pull_request' }}
+        continue-on-error: true
+        run: |
+          python3 -c "
+import xml.etree.ElementTree as ET, glob, sys
+files = glob.glob('build/reports/*.junit') + glob.glob('artifacts/**/*.junit', recursive=True)
+for f in files:
+    try:
+        for tc in ET.parse(f).iter('testcase'):
+            failure = tc.find('failure')
+            if failure is not None:
+                classname = tc.get('classname', '')
+                name = tc.get('name', 'Unknown')
+                msg = (failure.get('message') or failure.text or '').replace('%', '%25').replace('\r', '%0D').replace('\n', '%0A')
+                title = f'{classname}.{name}'.replace(',', '')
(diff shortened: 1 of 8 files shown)


netbe:
Superseded — splitting into focused PRs.

PR #4934, opened 2026-06-26
fix: sharing link over share extension - WPB-24816

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24816" title="WPB-24816" target="_blank">BugWPB-24816</a>  [iOS] [Share ext] Sharing website from Safari doesn't include link
  </td></table>
  <br />
 

### Issue
When sharing a link from Safari via the Wire share extension, the link isn't included in the text. The issue appears to be a timing issue:
1. Sharing attachments are stored locally, but this doesn't happen immediately because it was moved inside a task that loads the current account.
2. Meanwhile the when the view appears, we try to read the stored attachment to display the link in the text field, but since none are found no link is added.

The solution is to store the attachment earlier. But this created other problems, as it seems that you can only load the attachment once which happens when fetching the URL to append to the text field. When sending the message containing the link, another code path would try to load the attachment again and get stuck, preventing the message from being sent.

But I found that this second load of the attachment wasn't even necessary, so I deleted it.

### Testing
1. Go to www.apple.com in Safari
2. Open the share sheet, tap Wire, see that the link is in the text field.
3. Choose a conversation and tap send.

---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS Share Extension/Sources/Content/SendController.swift
+++ b/wire-ios/Wire-iOS Share Extension/Sources/Content/SendController.swift
@@ -67,19 +67,19 @@ final class SendController {
     ) {
         var sendables: [UnsentSendable] = attachments.compactMap {
             if $0.hasGifImage {
-                return UnsentGifImageSendable(
+                UnsentGifImageSendable(
                     conversation: conversation,
                     sharingSession: sharingSession,
                     attachment: $0
                 )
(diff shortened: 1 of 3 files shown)


netbe (approved):
LGTM:)

samwyndham (approved):
LGTM

PR #4944, opened 2026-07-01
chore:  uitest for app language changed to german - WPB-26347

findms:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26347" title="WPB-26347" target="_blank">TaskWPB-26347</a>  [iOS] Automate - 8944 - change app language to german from english
  </td></table>
  <br />
 

### Issue

new test: App language changed to german

### Testing

https://github.com/user-attachments/assets/92149c46-ea3a-452f-8a12-3b541bfab0ff

---

### Checklist

---


wire-ios/WireUITests/SettingsTests.swift
```diff
@@ -150,4 +150,28 @@ final class SettingsTests: WireUITestCase {
...
+            )
+
+        let conversationsPage = try await loginToBackend(user: user)
+
+        // Verify the conversation list empty state is localized in German
+        conversationsPage
+            .verifyStringInGerman(
+                "Willkommen 👋",
+                "Verbinden Sie sich mit anderen oder erstellen Sie eine neue Gruppe, um zu kommunizieren!"
+            )
```
jullianm, line 175: @findms So we only verify one string here ? Maybe I misunderstood but if the purpose of this test is to ensure that the app is correctly switched to DE by the system, I think it's safe to assume that the OS will handle that correctly then I don't think we need a test for that.

jullianm (commented):
Left a comment before approving to clarify the purpose of this test.

wire-ios/WireUITests/SettingsTests.swift
findms: @jullianm I also agree that it's hard to automate it and maybe not worth to have a test that the app is changed to fully german with 1 or 2 string.
Ideally for manual test also for this testcase, I switch the app and De and just spend 5 mins or so with the app if there is any english string exist. That's same I tried to do here with 1-2 string.
I am okay to close it and keep this test manual. What do you think?

jullianm: Yes I think it makes sense to keep it manual. Also in debug mode we have this option to show non-localized strings (Run > Options), maybe we could somehow leverage that option in an automated test that would identify untranslated texts in the app and fail if any is found.

PR #4946, opened 2026-07-01
fix: user status not shown - WPB-26710

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26710" title="WPB-26710" target="_blank">BugWPB-26710</a>  [iOS] iOS does not show status from other clients
  </td></table>
  <br />
 

### Issue

Availability status is never shown for other users, even though the display code is wired up correctly.

### Cause

Availability updates are broadcast via Proteus (`/broadcast/proteus/messages`) and arrive as `conversation.proteusMessageAdd` events in the recipient's self conversation. In `ConversationProteusMessageAddEventProcessor`, the event is processed through `canAddMessage`, which returns `false` when the conversation is the self conversation and the sender is not the self user. This silently drops the availability message before it ever reaches `ConversationProtobufMessageProcessor`.

### Solution

Decode the protobuf message before the conversation fetch and `canAddMessage` gate. If the content is `.availability`, call `userLocalStore.updateUser(with:availability:)` directly and return — no conversation is needed for this case.

### Testing

**Requirements:** two accounts in the same team on a Proteus-enabled backend.

**Steps:**
1. Log in as User A.
2. On another device or simulator, log in as User B and change their availability status (e.g. set to Busy).
3. On User A's device, open a conversation with User B or view their profile.

**Expected result:** User B's availability dot reflects the status they set in step 2.

---

### Checklist

---

### DIFF
--- a/WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationMessageAddEventProcessor/ConversationProteusMessageAddEventProcessor.swift
+++ b/WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationMessageAddEventProcessor/ConversationProteusMessageAddEventProcessor.swift
@@ -88,7 +88,7 @@ struct ConversationProteusMessageAddEventProcessor: ConversationProteusMessageAd
             )
         }
 
-        // Ensure is not self conversation, sender is self user and conversation is not read-only
+// Ensure the conversation isn't forced read-only, and reject messages sent by non-self users into the self conversation.
         guard await messageLocalStore.canAddMessage(
             conversation: conversation,
             senderID: senderID.id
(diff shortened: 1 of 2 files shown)


caldrian (approved):

WireDomain/Sources/WireDomain/Event Processing/ConversationEventProcessor/ConversationMessageAddEventProcessor/ConversationProteusMessageAddEventProcessor.swift
```diff
@@ -87,6 +79,28 @@ struct ConversationProteusMessageAddEventProcessor: ConversationProteusMessageAd
             )
         }
 
+        // Availability broadcasts don't belong to a specific conversation and must bypass
+        // the conversation fetch and `canAddMessage` gate entirely.
```
David-Henner:
```suggestion
        // the `canAddMessage` gate entirely.
```

netbe (commented):
question: should we also remove this check for mls?

netbe (approved):
I'll approve it but it would be nice to figure out why the check was there in the first place

PR #4947, opened 2026-07-01
fix: crash on implicit unwrapped optional - WPB-26725

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26725" title="WPB-26725" target="_blank">BugWPB-26725</a>  [iOS] App crashing when tapping on notification from group for inactive account
  </td></table>
  <br />
 

### Issue

- AppLockModule.Interactor.handleAuthenticationResult dispatches its result to the main queue guarding only [weak self], while its presenter property is a weak, implicitly-unwrapped optional (AppLockPresenterInteractorInterface!) that isn't guarded.
- The app-lock module can be torn down and rebuilt mid-flight: AppRootRouter.showAppLock rebuilds a fresh AppLockModule.View/Presenter/Interactor every time the app state re-enters .locked (e.g. when the Face ID/Touch ID/passcode system prompt briefly resigns the app). The old Interactor is kept alive by the pending LocalAuthentication callback closure, but its Presenter (owned only by the now-discarded View) gets deallocated, zeroing the weak presenter reference.
- When the stale callback later fires with .denied, .needCustomPasscode, or .unavailable, the code force-unwraps the nil presenter → EXC_BREAKPOINT / "Unexpectedly found nil while implicitly unwrapping an Optional value" crash (confirmed via crash.log, matches Thread 0 crash + Thread 14's LAContext completion path).

Fix: guard let self, let presenter else { return } in the dispatched closure — mirrors an existing pattern already used elsewhere in the codebase (ImagePickerManager.swift, AudioTrackPlayer.swift, ConversationListViewController+NavigationBar.swift). Stale/orphaned callbacks now no-op instead of crashing.

File changed: Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Interactor.swift (1 line)

### Testing

account A and B are logged in, A is the active account
app is int he background
B receives a notifcation from a group
B tapps on the notification -> app crashes

---

### Checklist

---

### DIFF
--- a/wire-ios/Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Interactor.swift
+++ b/wire-ios/Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Interactor.swift
@@ -123,7 +123,7 @@ extension AppLockModule.Interactor: AppLockInteractorPresenterInterface {
 
     private func handleAuthenticationResult(_ result: AppLockModule.AuthenticationResult) {
         DispatchQueue.main.async(group: dispatchGroup) { [weak self] in
-            guard let self else { return }
+            guard let self, let presenter else { return }
 
             switch result {
             case .granted:


valerio-bettini:
I'd rather change the implicitly unwrapped optional declaration into an optional
`        weak var presenter: AppLockPresenterInteractorInterface?`
and unwrap it when we need it

PR #4950, opened 2026-07-01
fix: make conversation context menu heading accessible - WPB-14753

acv-w:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14753" title="WPB-14753" target="_blank">TaskWPB-14753</a>  [iOS] Headings not developed/ not meaningful: Context menus #72
  </td></table>
  <br />
 

### Issue

The conversation-list options menu used the conversation name alone as its heading, for example `Guest area`. That label is not meaningful and was not exposed as an accessibility heading.

This PR updates the conversation options heading to use the meaningful accessibility label `Conversation options for <conversation name>`.

It covers both conversation-list entry points:
- long-press context menu
- swipe-right action sheet / bottom sheet

### Testing

- Verified the long-press context menu uses `Conversation options for <conversation name>`.
- Verified the swipe-right action sheet title uses `Conversation options for <conversation name>`.
- Verified the relevant element is marked as an accessibility header.

---

### Checklist

---

- [X] Make sure you use the API for UI elements that support large fonts.
- [X] All colors are taken from WireDesign.ColorTheme or constructed using WireDesign.BaseColorPalette.
- [X] New UI elements have Accessibility strings for VoiceOver.


wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationActionController.swift
```diff
@@ -145,13 +156,32 @@ final class ConversationActionController {
...
+        if let titleLabel = self as? UILabel, titleLabel.text == text {
+            titleLabel.isAccessibilityElement = true
+            titleLabel.accessibilityLabel = label
+            titleLabel.accessibilityTraits.insert(.header)
+            return true
+        }
+
+        for subview in subviews where subview.markTitleAsHeader(matching: text, accessibilityLabel: label) {
+            return true
+        }
```
netbe: question: why do we need to look at all subview here? is there not a more direct way to set the title?

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationActionController.swift
```diff
@@ -55,22 +55,33 @@ final class ConversationActionController {
...
         let controller = UIAlertController(title: title, message: nil, preferredStyle: .actionSheet)
         actions.map(alertAction).forEach(controller.addAction)
         controller.addAction(.cancel())
 
+        let markTitleAsHeader = {
+            guard let title, let accessibilityTitle else { return }
+            controller.view.markTitleAsHeader(matching: title, accessibilityLabel: accessibilityTitle)
+        }
+        markTitleAsHeader()
+
```
netbe: question: could you explain why this is necessary?

netbe (commented):
left some questions to better understand the changes

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationActionController.swift
acv-w: because on UIAlertControllers we have a title string property but the actual title label is not exposed. it's not elegant but these were the requirements. thoughts?

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationActionController.swift
acv-w: same explanation as above, basically looking for the title label manually and settings the accessibility property. I'm going to simplify the block (semantics).

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationActionController.swift
netbe: weird that Apple does not expose proper API, but not the first time. I wonder then if we should push back the requirement. for the time being let's go with this solution. 

suggestion: add a comment there to explain this hack.
thought: i wonder if it's worth moving this to a custom UIAlertController if it's needed else where

netbe (approved):
LGTM

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationActionController.swift
David-Henner: I would also suggest to push back against the requirement. Perhaps we can compromise on changing the title directly, which would then be read by VoiceOver and satisfy the "meaningfulness" requirement at least

David-Henner (approved):

PR #4964, opened 2026-07-06
chore: test video call - WPB-26642

findms:
### Issue

new test - Switching video call and other participants also showing video.

### Testing

Ran on local 
image

https://github.com/user-attachments/assets/84adbda0-2ed6-4a77-bde6-51f9daa50efe

---

### Checklist

---


KaterinaWire:
@findms But it doesn't show the self video or any indication that the self video works fine, right?

wire-ios/WireUITests/Helper/CallingManager.swift
```diff
@@ -0,0 +1,195 @@
...
+            if let flowAfter,
+               !checkAudioSent || flowAfter.audioPacketsSent > flowBefore.audioPacketsSent,
+               !checkAudioReceived || flowAfter.audioPacketsReceived > flowBefore.audioPacketsReceived,
+               !checkVideoSent || flowAfter.videoPacketsSent > flowBefore.videoPacketsSent,
+               !checkVideoReceived || flowAfter.videoPacketsReceived > flowBefore.videoPacketsReceived {
+                return
+            }
+
+            try await Task.sleep(for: .seconds(1))
+        } while Date() < deadline
```
KaterinaWire: **question:** why do we need a `deadline`?

wire-ios/WireUITests/Helper/CallingManager.swift
```diff
@@ -0,0 +1,195 @@
...
+
+        throw RuntimeError(
+            "CallingService no positive flow change for \(instanceId). Before: \(flowBefore). After: \(String(describing: flowAfter))"
+        )
+    }
+}
+
+private extension Set<String> {
+    func containsStatus(_ status: String?) -> Bool {
```
KaterinaWire: **question:** I don't quite understand why we need to compare statuses

wire-ios/WireUITests/Helper/CallingManager.swift
findms: We poll CallingService because call state and /flows counters are async. To avoid waiting forever when media never arrives, we keep a bounded wait. I removed the deadline var and use timeout directly now.

wire-ios/WireUITests/Helper/CallingManager.swift
findms: make sense to status comparison was not needed here. We only need currentCall.id before switching video on, and CallingService can still show _WAITING_. Updated to wait for current call instead of active.

findms:
> @findms But it doesn't show the self video or any indication that the self video works fine, right?

@KaterinaWire This is limitation on Simulator for self video as it doesn't show self preview. So we try to assert this by buttons and packets received.

johnxnguyen (approved):

KaterinaWire (approved):

PR #4981, opened 2026-07-08
fix: keep participant focus on selection - WPB-14670

acv-w:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14670" title="WPB-14670" target="_blank">EpicWPB-14670</a>  [iOS] - Create Group - Accessibility findings
  </td></table>
  <br />
 

### Issue

Keyboard focus jumped to the first contact after selecting or deselecting participants with the keyboard.

This PR updates participant selection handling so contact selection changes no longer reload the contacts collection view. Visible cells are updated in place, preserving focus on the interacted contact.

---

### Checklist

---


KaterinaWire:
@acv-w Honestly, I don't understand what the issue is? The description mentions a keyboard, but we're talking about an iPad? Could you provide a video demonstrating the issue? Just so I can better understand and see how the changes might affect the app's behavior.

acv-w:
> @acv-w Honestly, I don't understand what the issue is? The description mentions a keyboard, but we're talking about an iPad? Could you provide a video demonstrating the issue? Just so I can better understand and see how the changes might affect the app's behavior.

@KaterinaWire Thanks. To clarify: the issue is that toggling a participant checkbox updates UserSelection, and ContactsSectionController was reacting by calling reloadData() on the contacts collection view. That rebuilds the visible cells, so UIKit loses the currently focused contact and focus falls back to the first contact in the list.

This PR removes those selection-triggered reloads and updates visible cells selection state in place instead. The intended behavior change is narrow: when a contact is selected/deselected, focus stays on that same contact.

I don’t think a video adds much for reviewing the code path to be honest.

KaterinaWire (approved):

wire-ios/Wire-iOS/Sources/UserInterface/StartUI/Common/SectionControllers/Contacts/ContactsSectionController.swift
```diff
@@ -105,30 +105,52 @@ class ContactsSectionController: SearchSectionController {
...
+    private func updateVisibleSelectionStates() {
+        guard allowsSelection, let collectionView else { return }
+        collectionView.visibleCells.compactMap { $0 as? UserCell }.forEach { cell in
+            guard let user = cell.avatarImageView.user else { return }
+            setSelected(selection?.users.contains(user) ?? false, for: cell, in: collectionView)
+        }
+    }
+
+    private func setSelected(_ selected: Bool, for cell: UserCell?, in collectionView: UICollectionView) {
+        guard let cell else { return }
```
samwyndham, line 129: nitpick (opinionated): to me this is a bit of an any pattern having the API accept a nil but then returning immediately. I guess as this is private maybe this is the better choice to avoid the alternative boilerplate but I think it should be avoided for internal/public APIs.

wire-ios/Wire-iOS Tests/Search/SearchResultsViewControllerTests.swift
```diff
@@ -45,4 +46,52 @@ final class SearchResultsViewControllerTests: XCTestCase {
         // THEN
         XCTAssertNil(sut)
     }
+
+    func testThatSelectionChangesDoNotReloadContacts() {
```
samwyndham, line 50: praise: testing this behaviour.

wire-ios/Wire-iOS/Sources/UserInterface/StartUI/Common/SectionControllers/Contacts/ContactsSectionController.swift
```diff
@@ -105,30 +105,52 @@ class ContactsSectionController: SearchSectionController {
...
+        if let indexPath = collectionView.indexPath(for: cell) {
+            if selected {
+                collectionView.selectItem(at: indexPath, animated: false, scrollPosition: [])
```
samwyndham, line 132: question: I assume this doesn't result in some kind of loop with the the didSelectItemAt being called which calls this again?

samwyndham (commented):
Nice work. Just left one question before approving.  Please re-assign me for review once answered.

wire-ios/Wire-iOS/Sources/UserInterface/StartUI/Common/SectionControllers/Contacts/ContactsSectionController.swift
acv-w: Here the optional comes directly from cellForItem(at:), and nil intentionally means there is no cell to update, while the underlying selection should still continue. Since this helper is private, I’d prefer keeping that no-op centralized rather than duplicating the unwrap at both call sites. That said I'll admit I'm also opinionated so I'll refactor if you think I should.

wire-ios/Wire-iOS/Sources/UserInterface/StartUI/Common/SectionControllers/Contacts/ContactsSectionController.swift
acv-w: collectionView(_:didSelectItemAt:) is not called when the selection is changed programmatically with selectItem(at:). The same applies to deselectItem(at:) and didDeselectItemAt, so this cannot create a delegate loop.

https://developer.apple.com/documentation/uikit/uicollectionviewdelegate/collectionview%28_%3Adidselectitemat%3A%29

samwyndham (approved):

PR #5003, opened 2026-07-10
chore: split WireNetwork into three packages - WPB-15347

caldrian:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15347" title="WPB-15347" target="_blank">BugWPB-15347</a>  [iOS] Issues with linking dependencies
  </td></table>

### Issue

This PR attempts to prevent duplicate symbols of types in WireNetwork by splitting the framework into three packages each having only one dynamic product.

After merging the GitHub Actions caches should be cleared!

---

### Checklist

---


WireNetwork/WireNetworkTests/Package.swift
```diff
@@ -3,35 +3,15 @@
 import PackageDescription
 
 let package = Package(
-    name: "WireNetwork",
+    name: "WireNetworkTests",
```
johnxnguyen, line 6: question: why do we need a separate package for tests? A separate package for WireNetworkSupport makes sense, but I would expect that in WireNetwork there is simply a test target.

samwyndham (commented):
I think in practice 1 or 2 packages is enough. Some options include:

1. Have a single `WireNetwork` package with a single product containing multiple targets - `WireNetwork` & `WireNetworkSupport`. **Also**, update the automockable stencil so that mocks are generated inside an `if DEBUG`. I think this could be the easiest general approach to our project. This also means potentially using our mocks within SwiftUI previews which could be helpful.
2. Have 2 packages - `WireNetwork` & `WireNetworkSupport`. `WireNetworkTests` still live inside the `WireNetwork` package but **do not** depend on the `WireNetworkSupport` package. Instead as the mocks are generated just them generated in both `WireNetworkTests` and `WireNetworkSupport`.
3. Have 2 packages - `WireNetwork` & `WireNetworkSupport`. `WireNetworkTests` lives inside `WireNetworkSupport` - I don't like this option.

@caldrian what do you think?

netbe:
removing myself from review since 2 reviewers are there already. Ping me if needed

WireNetwork/WireNetworkTests/Package.swift
caldrian: If `WireNetworkSupport` is in a separate package, `WireNetworkTests` must be in a separate as well, otherwise we'd have a cyclic dependency, which is not allowed.

caldrian:
This PR was just an experiment. The actual solution might involve getting rid of the -Support modules, and optionally changing the way we generate mocks.

PR #5006, opened 2026-07-10
fix: video export failing when the source file is already an .mp4 - WPB-26457

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26457" title="WPB-26457" target="_blank">BugWPB-26457</a>  [iOS] Not able to attach video file on opening photos in drive conversation
  </td></table>
  <br />
 

### Issue

Exporting/converting a video to upload format was failing with `AVFoundationErrorDomain Code=-11800` (wrapping `NSOSStatusErrorDomain Code=-16979`) whenever the source video's temp filename already ended in `.mp4`.

Root cause: `convertVideoToUploadFormat`/`convert(filename:)` derive the converted file's output path from the source's basename with the extension swapped to `.mp4`. When the source was already `.mp4`, the computed output path was identical to the source path. The code then deleted any "leftover" file at that path before starting the export — which deleted the source file the `AVAssetExportSession` was still reading from, causing the export to fail.

Fixing that first collision exposed a second, related bug: once the export was made to succeed by writing to a distinct intermediate file and moving it into place, the converted file ended up at the same path as the source. The `deleteSourceFile` cleanup step in `convertVideoToUploadFormat` then deleted that path, wiping out the just-converted file before it could be uploaded (`NSCocoaErrorDomain Code=260` / `ENOENT` downstream, e.g. in `processWireDriveVideo`).

Both are fixed in `AVAsset+VideoConvert.swift`:
- `convert(filename:)` now exports to a uniquely named (UUID) intermediate file, then moves it to the final output path only after the export session has finished reading from the source.
- `convertVideoToUploadFormat` now skips deleting the source file if its path is the same as the converted output's path (i.e. when they collided).

### Testing

- Reproduce by sharing/uploading a video via WireDrive (or camera roll) whose temp copy already has a `.mp4` extension (e.g. `processWireDriveVideo` in `ConversationInputBarViewController+ImagePickerDriveConversation.swift`) — before the fix this failed export with `-11800`/`-16979`, and after a partial fix failed upload with `ENOENT`. After both fixes, conversion and upload succeed.
- Existing unit test `AVURLAsset_conversionTests.testThatVideoIsConvertedToUploadFormat` (`Wire-iOS Tests/AVAsset/AVURLAsset+conversionTests.swift`) still covers the non-colliding case.

---

### Checklist

---


jullianm:
@findms could you please test it and share info here whether it fixes your issue or not. 
Feel free to merge this PR when QA validated and approved while I'm away.

findms:
> @findms could you please test it and share info here whether it fixes your issue or not. Feel free to merge this PR when QA validated and approved while I'm away.

@jullianm I have tried building with this branch fix/video-export-fails but it’s same for me..Still after selecting a video file from picker it doesn't show as attached(or compressing ).

netbe (approved):
Looks good, it would be good to add tests for this

findms (approved):

PR #5023, opened 2026-07-16
fix: show MLS-specific learn more link for empty key packages - WPB-24214

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24214" title="WPB-24214" target="_blank">StoryWPB-24214</a>  [iOS] Learn more link shows wrong info on support article
  </td></table>
  <br />
 

### Issue

When adding a member to an MLS conversation fails because the backend returns no MLS key packages for them (no MLS client, or an MLS client with no key packages published), we currently show the same "Learn more" link as a genuine Federation/unreachable-backend failure. That's misleading, since the two causes are unrelated and the federation article doesn't explain the actual problem.

Since the client can't yet distinguish "no MLS client" from "backend unreachable" purely from the empty key-packages response, we instead stop mapping this case to the generic Federation system message and route it to a new, MLS-specific system message type with a link to the MLS support article — mirroring the approach Android shipped in wireapp/wire-android#4659.

Technical approach:
- Added `ZMSystemMessageTypeFailedToAddParticipantsMLS`, a new case on the existing `ZMSystemMessageType` enum (no Core Data migration needed, since `systemMessageType` is already a persisted attribute).
- Threaded this reason from `MLSService.MLSAddMembersError.failedToClaimKeyPackages` through both the legacy request-strategy path (`ConversationParticipantsService`) and the WireDomain live-sync path (`CreateGroupConversationUseCase`/`CreateChannelUseCase`) up to the system message append call.
- `ConversationFailedToAddParticipantsSystemMessageCellDescription` now shows `WireURLs.shared.mlsInfo` instead of `WireURLs.shared.unreachableBackendInfo` when the reason is the new MLS type. Body text is unchanged (already generic, not federation-specific wording).

### Testing

1. Create a user without an MLS client using WireAPIDebugger, and accept the connection request.
2. Try to add that user to an existing group, or try to create a new MLS conversation with them.
3. Confirm the "X could not be added to the group" message now shows a "Learn more" link pointing to the MLS support article, not the federation/unreachable-backends one.

---

### Checklist


johnxnguyen (approved):

jullianm (approved):

jullianm:
Tested and validated by QA. Merging.

PR #5040, opened 2026-07-21
fix: Recover unestablished MLS groups - WPB-21123

acv-w:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21123" title="WPB-21123" target="_blank">BugWPB-21123</a>  [iOS] Recover from having a local unestablished MLS group
  </td></table>
  <br />
 

### Issue

An MLS conversation can remain in Core Crypto at epoch 0. Because `conversationExists` still returns `true`, the client incorrectly treats it as established when sending messages.

Before sending an MLS message, detect locally groups that have epoch 0, wipe them, and re-establish before encrypting and sending the message.

This PR also adds a developer flag, `simulateUnestablishedMLSGroup`, which leaves the next locally created MLS group at epoch 0 for manual reproduction and testing.

### Testing

1. Enable `simulateUnestablishedMLSGroup`. 
2. Create an MLS group.
3. Send the first message.
4. Verify the (epoch-0) group is wiped, re-established, and the message is sent successfully.

---

### Checklist

---


jullianm (approved):
LGTM

wire-ios-request-strategy/Sources/Message Sending/MessageSenderTests.swift
```diff
@@ -460,7 +460,7 @@ final class MessageSenderTests: MessagingTestBase {
 
     // MARK: - Send MLS Message
 
-    func testThatWhenSendingMlsMessageSucceeds_thenCompleteWithoutErrors() async throws {
```
johnxnguyen, line 463: question: do we no longer have a test for the happy path?

johnxnguyen (commented):
Looks good, just had one question

wire-ios-request-strategy/Sources/Message Sending/MessageSenderTests.swift
acv-w: @johnxnguyen we still have a happy path test: testThatWhenSendingMlsMessageSucceeds_thenCommitPendingProposalsInGroup. It tests sending from an established epoch 1 group and verifies that recovery is not triggered. The previous ...CompleteWithoutErrors test covered the same path, so I used that one to cover the new epoch 0 recovery case. Let me know your thoughts.

johnxnguyen (approved):
great work

PR #5047, opened 2026-07-22
feat(mls): parallelize key package claims - WPB-27420

MohamadJaara:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27420" title="WPB-27420" target="_blank">TaskWPB-27420</a>  [iOS] parallelize kypackage caliming when creating MLS conv
  </td></table>
  <br />
 

## What changed

- claim MLS key packages concurrently with a maximum of four active requests
- preserve deterministic key-package and failed-user ordering
- stop launching new requests when the current claim batch cannot continue
- add coverage for concurrency limits, ordering, cancellation, connectivity, and aggregate results

## Impact

Adding multiple members to an MLS conversation completes faster while keeping the existing result and error behavior predictable.

## Validation

The focused MLS service test suite passes with both commits applied.


wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -530,30 +550,130 @@ public final class MLSService: MLSServiceInterface {
...
+                if !fatalFailureDetected, nextUserIndex < users.count {
+                    addKeyPackageClaimTask(
+                        for: users[nextUserIndex],
+                        at: nextUserIndex,
+                        ciphersuite: ciphersuite,
+                        notificationContext: notificationContext,
+                        to: &group
+                    )
+                    nextUserIndex += 1
+                }
```
johnxnguyen: question: is it intended that if there is a fatal failure detected then we ignore it and wait for the group to complete all it's tasks? Since a `CancellationError` is considered a fatal failure, I would expect that we should clean up immediately.

MohamadJaara: With a limit of N, the worst case is N-1 calls already running and no new calls are started after the fatal error however i can use `group.cancelAll()` to short circuit all running claims. Imho, it's not worth it 😅 but happy to add it if you think otherwise

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -530,30 +550,130 @@ public final class MLSService: MLSServiceInterface {
             return []
         }
 
-        var result = [KeyPackage]()
+        try Task.checkCancellation()
+
+        let notificationContext = context.notificationContext
+        let claims = await withTaskGroup(of: KeyPackageClaim.self) { group in
```
samwyndham: thought (no change necessary): This is now at least the second time where we are using a task group with a concurrency limit. I wonder if this can be abstracted to make things easier to reason about in context

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -530,30 +550,130 @@ public final class MLSService: MLSServiceInterface {
...
+                break
+            }
+        }
+
+        if let networkStackError = error as? NetworkStackError,
+           case .proxyCredentialsRequired = networkStackError {
+            return true
+        }
+
+        return false
```
samwyndham: suggestion (opinionated): Consider using a switch + pattern matching instead of ifs

wire-ios-data-model/Source/MLS/MLSService.swift
samwyndham: I think it is worth cancelling ongoing tasks. The reason is that why it doesn't matter much whether these tasks finish or not, they use network resources which are a bottleneck and there could be other tasks waiting on those resource.

samwyndham (commented):
Nice work! Before approving I think it is worth adding group cancellation as John mentioned

wire-ios-data-model/Source/MLS/MLSService.swift
samwyndham: suggestion: I like the extraction of the task to it's struct but maybe this method is no longer needed now and this can be done directly in `addMembersToConversation`

johnxnguyen (approved):
Nice work!

samwyndham (approved):
Nice work @MohamadJaara. Added a small optional change.

PR #5071, opened 2026-07-27
fix: failure collecting simulator diagnostics - WPB-27452

samwyndham:
### Issue

It's been discovered that often during CI/CD runs the jobs stop for 600 seconds collecting diagnostics information form simulator.

> DETestOperationsObserverDebug: Failure collecting diagnostics from simulator: Timed out after 600.0 seconds while waiting for a response from the invoked process.

I have found this to affect the following test schemes - it may well affect others:

- `WireDataModel`
- `WireTransport`
- `WireUtilities`

Downloading an inspecting the xcresult files I have found that the resulting xcresult files for the above are much bigger then others:

Screenshot 2026-07-27 at 16 46 26

Inspecting the xcresults we see they contain a lot of diagnostics data:

> xcrun xcresulttool export diagnostics --path WireDataModel.xcresult    --output-path Diagnostics

Screenshot 2026-07-27 at 16 53 52

In the failure cases the `xcresults` contain a folder `simctl_diagnostics`. This folder contains all kinds of diagnostics information about the system. In practice I don't think we ever look at this data.

Screenshot 2026-07-27 at 17 04 57

This information should only be generated when there are failures. However for the above 3 schemes it seems that it is always generated.

As we don't actually use use this data and collecting it is leading to timeouts the easy fix here is to disable it:

> -collect-test-diagnostics on-failure|never   Whether or not testing collects verbose diagnostics (like a sysdiagnose) when encountering a failure

This PR does that. 

### Testing

Tested that it works here: It can be seen to work here: https://github.com/wireapp/wire-ios/actions/runs/30270114149

---

### Checklist

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -639,7 +639,8 @@ platform :ios do
             cloned_source_packages_path: ENV['PACKAGES_DIR'],
             disable_package_automatic_updates: true,
             skip_package_dependencies_resolution: true,
-            number_of_retries: build.number_of_retries
+            number_of_retries: build.number_of_retries,
+            xcargs: "-collect-test-diagnostics never" # Disable verbose diagnostics as these are generally not useful and collecting them can cause 600 second delays per test scheme.
         }
         run_tests_options[:only_testing] = options[:only_testing] if options[:only_testing]&.any?
         run_tests_options[:skip_testing] = options[:skip_testing] if options[:skip_testing]&.any?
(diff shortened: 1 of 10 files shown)


fastlane/Fastfile
```diff
@@ -639,7 +639,8 @@ platform :ios do
             cloned_source_packages_path: ENV['PACKAGES_DIR'],
             disable_package_automatic_updates: true,
             skip_package_dependencies_resolution: true,
-            number_of_retries: build.number_of_retries
+            number_of_retries: build.number_of_retries,
+            xcargs: "-collect-test-diagnostics never" # Disable verbose diagnostics as these are generally not useful and collecting them can cause 600 second delays per test scheme.
```
johnxnguyen, line 643: praise: for leaving an explanation

johnxnguyen (approved):
Nice work 🕵🏼

caldrian (approved):

PR #5073, opened 2026-07-28
feat: handle 403 error from backend with new eligible admin candidates - WPB-25300

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25300" title="WPB-25300" target="_blank">TaskWPB-25300</a>  [iOS] Handle new API error when promoting user as admin
  </td></table>
  <br />
 

### Issue

Handles the (very edge) case where removing/leaving a group conversation fails because it would leave the conversation without an admin, and the backend returns a fresh list of eligible admin candidates.

- Added `ConversationRemoveParticipantError.requiresAdmin([EligibleMember])` case, with a nested `EligibleMember` struct parsed from the backend's `eligible_members` payload (`id`/`domain`).
- `RemoveParticipantActionHandler` now parses a 403 response's `eligible_members` payload and reports `.requiresAdmin` with the decoded members, falling back to the default error handling if the payload can't be parsed.
- `AdminSelectionViewModel` gained a `showError` init parameter that immediately sets `promotionState` to `.failed` so the admin selection screen can be re-presented already showing an error.
- `ConversationActionController+Leave.presentAdminSelection` now accepts a `showAlert` flag and, when promoting an admin fails with `.requiresAdmin`, filters the conversation's participants down to the backend-provided eligible candidates and re-presents admin selection with that list and the error shown.
- `performAdminPromotion` now takes an optional `removeParticipantResultHandler` so the caller can react to the result of removing the leaving/removed participant (including the new `.requiresAdmin` case) instead of only the promotion step.
- Added unit tests for `RemoveParticipantActionHandler` covering the new 403/`eligible_members` handling.

https://github.com/user-attachments/assets/b0010520-a526-453b-86be-8a62519fa3a4

### Testing

Note: Testable in debug mode only by commenting out code that promotes the user and keeps only the code that removes the participant so API fails with 403.
- unlock and enable adminless group feature on staging
- enable developer flag
- leave a conversation with other participants
- promote an admin
- screen opens, hit promote
- API fails with 403 returning the list of eligible members 
- screen re-opens with the new list of participants  (in that case the same list of participants) and an alert error.
see attached video above for demo

- Run the new `RemoveParticipantActionHandlerTests` covering the 403 response handling.

---

### Checklist

---


wire-ios-data-model/Source/Model/Conversation/Actions/RemoveParticipantAction.swift
```diff
@@ -18,11 +18,28 @@
...
+        case (.conversationNotFound, .conversationNotFound):
+            true
+        case (.failedToRemoveMLSMembers, .failedToRemoveMLSMembers):
+            true
+        case (.invalidOperation, .invalidOperation):
+            true
+        default:
+            false
+        }
+    }
```
caldrian: question: Can't this code be synthesized?

```suggestion
    }
```

caldrian (approved):

wire-ios-data-model/Source/Model/Conversation/Actions/RemoveParticipantAction.swift
```diff
@@ -18,11 +18,28 @@
...
 }
```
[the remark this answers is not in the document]
jullianm: done

wire-ios-data-model/Source/Model/Conversation/Actions/RemoveParticipantAction.swift
jullianm: It was not necessary, removed it thanks.

wire-ios-request-strategy/Sources/Request Strategies/Conversation/Actions/RemoveParticipantActionHandler.swift
```diff
@@ -130,6 +130,18 @@ class RemoveParticipantActionHandler: ActionHandler<RemoveParticipantAction> {
...
+            let payload = response.payload
+            let dictionary = payload?.asDictionary()
+            let eligibleMembers = (dictionary?["eligible_members"] as? [[AnyHashable: Any]])?
+                .compactMap(ConversationRemoveParticipantError.EligibleMember.init(payload:))
+
+            guard let eligibleMembers, !eligibleMembers.isEmpty else {
+                fallthrough
+            }
+
+            action.notifyResult(.failure(ConversationRemoveParticipantError.requiresAdmin(eligibleMembers)))
```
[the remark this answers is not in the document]
jullianm: added the `!eligibleMembers.isEmpty` as a safety measure, in case `eligible_members` would be empty in payload, the backend would not return an error, but falls back to the failed auto-promotion scenario directly.

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationActionController+Leave.swift
```diff
@@ -197,8 +229,14 @@ extension ConversationActionController {
...
     }
 
+    private typealias RemoveParticipantResultHandler = (Result<Void, Error>) -> Void
+
     @MainActor
-    private func performAdminPromotion(user: UserType, in conversation: ZMConversation) async throws {
+    private func performAdminPromotion(
+        user: UserType,
+        in conversation: ZMConversation,
+        removeParticipantResultHandler: RemoveParticipantResultHandler? = nil
```
samwyndham, line 238: suggestion: For me it is a bit funny that this method is both throws and has a removeParticipantResultHandler which I see is a little like a `Result`. I wonder if it would make more sense to parse the handler in this method and throw the contained error.

samwyndham (approved):

wire-ios/Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationActionController+Leave.swift
jullianm: @samwyndham This is not possible because the completion handler is executed in a Task. 

In details, if we want to parse the handler and throws the error at this level, we'll need to make the completion handler throws and the `removeOrShowError` rethrows however as mentioned, at callsite, the error will never reach the catch blocks since we can't bridge an unstructured Task back to a synchronous caller.

PR #5074, opened 2026-07-28
fix: recover MLS conversations after key-package exhaustion - WPB-27395

acv-w:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27395" title="WPB-27395" target="_blank">BugWPB-27395</a>  [iOS] Recover MLS conversations after key-package exhaustion
  </td></table>
  <br />
 

### Issue

When an MLS client runs out of key packages while offline, it can miss Welcome messages and leave new or migrated conversations pending.

After quick sync, the client now checks the backend key-package count when recovery is pending, refills packages if needed, and recovers pending MLS or mixed-protocol conversations using the existing external-commit recovery paths. 

### Testing

Automated coverage was updated for key-package checks and active-session recovery.

Manual QA:
1. Exhaust Client A’s key packages while it is offline (use WireAPIDebugger).
2. Add the user to a new MLS conversation from client B and send a new message.
3. Bring Client A online in the foreground and wait for sync.
4. Verify that the conversation is recovered and messages can be received.

---

### Checklist

---

### DIFF
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -1030,6 +1030,7 @@ public final class MLSService: MLSServiceInterface {
                 logger.info("pending conversation recovery batch: recovered \(recovered) of \(batch.count)")
             }
 
+            // Return whether another batch should be processed immediately.
             return recovered == batch.count && pendingConversations.count > batch.count
         } catch {
             logger.error("failed pending conversation recovery batch: \(String(reflecting: error))")
--- a/wire-ios-data-model/Source/MLS/MLSService.swift
+++ b/wire-ios-data-model/Source/MLS/MLSService.swift
@@ -63,6 +63,8 @@ public final class MLSService: MLSServiceInterface {
     private let userDefaults: PrivateUserDefaults<Keys>
     private let logger = WireLogger.mls
     private let groupsBeingRepaired = GroupsBeingRepaired()
+    private let pendingConversationRecoveryLock = NSLock()
+    private var isRecoveringPendingConversations = false
     private let featureRepository: LegacyFeatureRepositoryInterface
     private weak var resetBrokenMLSConversationDelegate: (any ResetBrokenMLSConversationDelegate)?
     private let onEpochChangedSubject = PassthroughSubject<MLSGroupID, Never>()
@@ -92,6 +94,13 @@ public final class MLSService: MLSServiceInterface {
     // The number of days to wait until refreshing the key material for a group.
 
     private static let epochChangeBufferSize: Int = 1000
+    private static let pendingConversationRecoveryBatchSize = 10
+
+    private typealias PendingConversationRecovery = (
+        groupID: MLSGroupID,
+        qualifiedID: QualifiedID,
+        shouldResync: Bool
+    )
 
     private let maxRetryAttempts = 3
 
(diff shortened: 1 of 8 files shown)


wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -967,6 +982,93 @@ public final class MLSService: MLSServiceInterface {
...
+
+            if recovered > 0 {
+                await save(context)
+            }
+
+            if !batch.isEmpty {
+                logger.info("pending conversation recovery batch: recovered \(recovered) of \(batch.count)")
+            }
+
+            return recovered == batch.count && pendingConversations.count > batch.count
```
johnxnguyen, line 1034: suggestion: it's not clear what the meaning of the return type is, perhaps you could add a comment?

johnxnguyen: I saw the doc comment on the protocol, but a small comment here would be helpful, such as:

`whether there are still more conversation to process`

wire-ios-data-model/Tests/MLS/MLSServiceTests.swift
```diff
@@ -1816,9 +1844,10 @@ final class MLSServiceTests: ZMConversationTestsBase, MLSServiceDelegate {
         XCTAssertNotNil(privateUserDefaults.date(forKey: .keyPackageQueriedTime))
     }
 
-    func test_UploadKeyPackages_DoesNotSetKeyPackageQueriedTime_WhenGenerationFails() async {
+    func test_UploadKeyPackages_ClearsKeyPackageQueriedTime_WhenGenerationFails() async {
```
johnxnguyen, line 1847: question: why did we need to change this?

johnxnguyen (commented):
Looks good, just a couple comments.

wire-ios-data-model/Tests/MLS/MLSServiceTests.swift
acv-w: The test now starts with an existing keyPackageQueriedTime and verifies that the failure removes it. The previous name described the old setup, where no timestamp existed and the test only checked that one wasn’t created. The rename reflects this new behvior.

wire-ios-data-model/Source/MLS/MLSService.swift
acv-w: Makes sense. Added a quick clarification. Thank you!

johnxnguyen (approved):

wire-ios-data-model/Source/MLS/MLSService.swift
```diff
@@ -63,6 +63,8 @@ public final class MLSService: MLSServiceInterface {
     private let userDefaults: PrivateUserDefaults<Keys>
     private let logger = WireLogger.mls
     private let groupsBeingRepaired = GroupsBeingRepaired()
+    private let pendingConversationRecoveryLock = NSLock()
```
samwyndham, line 66: praise: using a lock!

samwyndham: optional suggestion: Use https://developer.apple.com/documentation/os/osallocatedunfairlock so that `isRecoveringPendingConversations` is only accessible via a lock

samwyndham (approved):
Nice work!

PR #5079, opened 2026-07-29
chore: UItest - Admin(not the last one) able to leave the group without seeing any admin‑selection modal - WPB-27599

findms:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27599" title="WPB-27599" target="_blank">TaskWPB-27599</a>  [iOS] Automate - able to leave the group without seeing any admin‑selection modal
  </td></table>
  <br />
 

### Issue

new Test: Being a 2nd last admin, leaving group does't show admin promotion popup.

### Testing

_Describe how to test._

Ran locally

https://github.com/user-attachments/assets/9b3e5e44-2455-45e6-85b9-a424bf3cf740

---

### Checklist

---


wire-ios/WireUITests/AdminPromotionTests.swift
```diff
@@ -217,4 +193,91 @@ class AdminPromotionTests: WireUITestCase {
...
+            .openUserDetailsPage(byName: secondAdmin.name)
+            .enableGroupAdmin()
+            .goBackToConversationDetailsPage()
+
+        XCTAssertTrue(
+            conversationDetailsWithSecondAdmin.adminCell(named: secondAdmin.name).waitForExistence(timeout: 5),
+            "Second admin should appear in the admin section"
+        )
+
+        // Once another admin exists, owner can leave directly without seeing the promote admin flow.
```
jullianm: question: Shouldn't the UI test actually start here ? Based on the test name, it seems to me we just want, as a pre-requisite, to have a second admin in the conversation (which we can easily setup using the dedicated method from the UserHelper) so all the code before is unnecessary and make the UITest longer to execute. 
What do you think?

jullianm (commented):
Left one question before approving

wire-ios/WireUITests/AdminPromotionTests.swift
findms: @jullianm I thought the same but in this test we are asserting that single admin can not leave group without promoting another admin(which seems duplicate as this is catered in another test) so yes we can make this as pre-requisite with 2 admin and just leave the conversation and no admin promotion dialog. Will change it.

findms: changes pushed.

KaterinaWire (approved):

jullianm (approved):

PR #5086, opened 2026-07-30
fix: clear retained account data after SSO identity change - WPB-27416

acv-w:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27416" title="WPB-27416" target="_blank">TaskWPB-27416</a>  [iOS] Clear retained account data when logging in with a different SSO identity
  </td></table>
  <br />
 

### Issue

- Pass the IdP ID returned by `/sso/get-by-email` through the SSO authentication flow.
- Store the successfully used IdP ID with the local account.
- Before reopening retained account data, compare the stored and current IdP IDs.
- Show a confirmation dialog when the IdP changed or the stored IdP is unknown.
- Keep the existing “Account already in use” behavior for active accounts.
- On confirmation, reuse the existing full account-removal flow before completing a clean login.
- Cancelling leaves retained data unchanged.

---

### Checklist

---


wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1186,15 +1215,25 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+        )
 
-        do {
-            try FileManager.default.removeItem(at: CoreDataStack.accountDataFolder(
-                accountIdentifier: accountID,
-                applicationContainer: sharedContainerURL
-            ))
-        } catch {
-            WireLogger.sessionManager.critical("Impossible to delete the account \(account): \(error)")
+        if FileManager.default.fileExists(atPath: accountDataFolder.path) {
```
samwyndham: suggestion: I believe apple discourages using this API (https://developer.apple.com/documentation/foundation/filemanager/fileexists(atpath:)) in this case and instead suggests attempting the retrieval then handling the error which I think is CocoaError.fileNoSuchFile

wire-ios/Wire-iOS/Sources/Authentication/AuthenticationInterfaceBuilder.swift
```diff
@@ -348,8 +348,10 @@ final class AuthenticationInterfaceBuilder {
             appStoreURL: WireURLs.shared.appOnItunes,
             accountsPublisher: CurrentValuePublisher(subject: CurrentValueSubject(accounts)),
             registrationAnalyticsTracker: registrationAnalyticsTracker,
-            isAccountAlreadyLoggedIn: { userID in
-                SessionManager.shared?.accountManager.accounts.contains { $0.userIdentifier == userID } ?? false
+            isAccountAlreadyLoggedIn: { result in
+                guard let sessionManager = SessionManager.shared,
+                      let account = sessionManager.accountManager.account(with: result.userID) else { return false }
+                return result.multiIngressIdentityProviderID == nil || sessionManager.isAccountActive(account)
```
samwyndham, line 354: question: Just to double check my understanding, am I correct that the logic here is:

1. In cases where `result.multiIngressIdentityProviderID == nil` the logic is the same as before.
2. In cases where `result.multiIngressIdentityProviderID != nil` we now check whether we have a cookie for that account.

samwyndham (commented):
Nice work. I left a few comments before approving. Also please check Copilots comments to see if anything needs addressing.

wire-ios/Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
```diff
@@ -565,6 +580,54 @@ extension AuthenticationCoordinator {
...
+        do {
+            try sessionManager.purgeRetainedAccountData(for: context.0.userID)
+            executeActions([.completeWireAuthenticationLogin(context)])
+        } catch {
+            logger.error("Failed to purge retained account data: \(error)")
+            restartWireAuthentication(error: NSError(userSessionErrorCode: .unknownError, userInfo: nil))
+        }
+    }
+
+    private func restartWireAuthentication(error: Error? = nil) {
```
KaterinaWire, line 619: **question:** I only see one use of the `restartWireAuthentication` and the error is hardcoded, am I wrong? My question is actually why is the error here optional?

WireAuthentication/Sources/WireAuthenticationLogic/DetermineAuthMethodUseCase.swift
```diff
@@ -46,7 +46,7 @@ package struct DetermineAuthMethodUseCase: DetermineAuthMethodUseCaseProtocol {
 
             do {
                 let ssoCode = try await authenticationAPI.getSSOCode(forEmail: email)
-                return .loginViaSSO(code: ssoCode)
+                return .loginViaSSO(code: ssoCode, multiIngressIdentityProviderID: ssoCode)
             } catch AuthenticationAPIError.unsupportedEndpointForAPIVersion, AuthenticationAPIError.ssoCodeNotFound {
```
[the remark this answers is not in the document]
acv-w: sso_code is the historical field name; the backend type is the IdP ID and the same UUID is passed to /sso/initiate-login, so no change is needed.

WireAuthentication/Sources/WireAuthenticationAPI/Models/AuthenticationResult.swift
```diff
@@ -51,6 +51,10 @@ public struct AuthenticationResult: Equatable, Hashable, Sendable {
 
     public let proxyCredentials: ProxyCredentials?
 
+    /// The identity provider selected by multi-ingress email discovery, if applicable.
+
+    public package(set) var multiIngressIdentityProviderID: UUID?
+
```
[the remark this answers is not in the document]
acv-w: AuthenticationResult is a value type. Its only mutation is on a fresh local value before navigation, so I kept the minimal implementation.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1186,15 +1215,25 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+            do {
+                try FileManager.default.removeItem(at: accountDataFolder)
+            } catch {
+                if keepAccountOnFailure {
+                    throw error
+                }
```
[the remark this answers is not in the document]
acv-w: I moved account-directory removal before credentials and defaults

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
acv-w: Good point. I removed the existence pre-check and attempt deletion directly.

wire-ios/Wire-iOS/Sources/Authentication/AuthenticationInterfaceBuilder.swift
acv-w: Yes. Without an ID, the behavior stays the same. With an ID, we only block login if the account is still active.

wire-ios/Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
acv-w: There are two call paths. Cancelling the dialog sends .restartWireAuthentication, which calls restartWireAuthentication() without an error and resets silently. If deleting the retained data fails, we call it with the generic error so it resets and shows an alert. The error is optional to support both cases.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1159,11 +1168,48 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+        let accountID = account.userIdentifier
+        let accountDataFolder = CoreDataStack.accountDataFolder(
+            accountIdentifier: accountID,
+            applicationContainer: sharedContainerURL
+        )
+
+        do {
+            try FileManager.default.removeItem(at: accountDataFolder)
+        } catch let error as NSError where error.domain == NSCocoaErrorDomain && error.code == NSFileNoSuchFileError {
+            // The desired state has already been reached.
```
KaterinaWire, line 1204: `suggestion:` maybe we can add some logs (I don't mean error logs, just debug or info)

wire-ios/Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
```diff
@@ -565,6 +580,54 @@ extension AuthenticationCoordinator {
...
+        let error = error as NSError?
+        stopActivityIndicator()
+        Task { @MainActor [weak self] in
+            guard let self else { return }
+            stateDidChange(stateController.currentStep, mode: .reset)
+            if let error {
```
KaterinaWire, line 625: **suggestion:** `let error = error as NSError? {`

KaterinaWire (approved):

KaterinaWire:
@acv-w I left some suggestions and there are failed tests.

acv-w:
> @acv-w I left some suggestions and there are failed tests.

The previous run failed because of an unrelated CoreCrypto build error

wire-ios/Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
acv-w: Thank you, I prefer the shorter syntax for brevity.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
acv-w: Thanks! Might be a bit too verbose? An already-missing folder is expected here, so I kept the comment instead of adding a log.

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
```diff
@@ -1164,11 +1173,48 @@ public final class SessionManager: NSObject, SessionManagerType {
...
+    ) throws {
+        let accountID = account.userIdentifier
+        let accountDataFolder = CoreDataStack.accountDataFolder(
+            accountIdentifier: accountID,
+            applicationContainer: sharedContainerURL
+        )
+
+        do {
+            try FileManager.default.removeItem(at: accountDataFolder)
+        } catch let error as NSError where error.domain == NSCocoaErrorDomain && error.code == NSFileNoSuchFileError {
```
samwyndham, line 1203: suggestion: I believe the following is also possible.

```suggestion
        } catch CocoaError.fileNoSuchFile {
```

samwyndham (approved):
Nice work!

wire-ios-sync-engine/Source/SessionManager/SessionManager.swift
acv-w: This compiles, but it doesn’t catch the NSError thrown by removeItem, so I’ll keep the current check.

PR #2719, opened 2025-03-20
feat: create channel use case - WPB-16398

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16398" title="WPB-16398" target="_blank">TaskWPB-16398</a>  [iOS] Use CreateChannelUseCase instead of ConversationService 
  </td></table>
  <br />
 

### Issue

- Add `createGroupConversation` API endpoint in WireAPI
- Create and implement CreateChannelUseCase in WireDomain (supported in API >= v8)
- Create and implement CreateGroupConversationUseCase in WireDomain
- Integrate these two use cases into legacy path (ConversationCreationController) and remove usage of legacy ConversationService

### Testing

- Create MLS group conversation (< API v8) 
- Create MLS channel (> API v8)
- Create Proteus group conversation

### Checklist

- [ ] Title contains a reference JIRA issue number like `[WPB-XXX]`.
- [ ] Description is filled and free of optional paragraphs.
- [ ] Adds/updates automated tests.

---

PR #5110, opened 2026-08-06
fix: personal user cannot add users to existing group - WPB-27822

jullianm:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27822" title="WPB-27822" target="_blank">BugWPB-27822</a>  [iOS] Personal users can not add users to existing group
  </td></table>
  <br />
 

### Issue

A bug was introduced a while ago in `ConversationService` where passing an empty `access_roles_v2` array to the API  puts the conversation in a broken state (although the API returns successfully), this happens only for personal users. When legacy code was later ported in https://github.com/wireapp/wire-ios/pull/2719, same issue was reproducible.

After that conversation is created,  trying to add participants becomes impossible with API raising a 403 "conversation access denied".

This is actually illegal API side to pass an empty array, we should pass nil.
Fix was implemented on both legacy / new code.

### Testing

First use case:
- Create a personal user (use `testRegisterAsPersonalUser_TC_8971` and put a breakpoint to get the email / password)
- Login with that personal user
- Connect with a contact which is part of a team
- Create a group conv with that contact
- Make that contact a group admin
- Login with that contact
- Try to add other participants
- **Before fix :** "Could not add participant" pop up with API request failing ({"code":403,"label":"access-denied","message":"Conversation access denied"},"status":403}
- **After fix:** Able to add participants 
- Now switch to personal user account and open the group conv
- Try to connect with a participant from that conv
- You should be able to connect, and that contact should receive the connection request

### Checklist

---


wire-ios-request-strategy/Sources/Payloads/Conversation/Payload+NewConversation.swift
```diff
@@ -102,7 +102,11 @@ extension Payload {
             self.name = action.name
             self.access = action.accessMode?.stringValue
             self.legacyAccessRole = action.legacyAccessRole?.rawValue
-            self.accessRoles = action.accessRoles.map(\.rawValue)
+            self.accessRoles = if action.accessRoles.isEmpty {
+                nil
+            } else {
+                action.accessRoles.map(\.rawValue)
+            }
```
samwyndham, line 109: question: I see from their docs it says:

> Please note that an empty list is not allowed when creating a new conversation.

Do you know why this is the case? I'm wondering if this fix is happening to work but also whether it is in fact correct to send nil to the API?

jullianm: yes it's illegal to pass an empty array, conversation is successfully created but in a broken state, if you try to add new participants to that conversation, API fails with ` ({"code":403,"label":"access-denied","message":"Conversation access denied"},"status":403}`. They expect a nil value instead of an empty array (issue was raised to them), I've tested with and without that fix and this appears to be correct.

WireNetwork/Tests/WireNetworkTests/APIs/Rest/ConversationsAPI/ConversationsAPITests.swift
```diff
@@ -2186,6 +2186,40 @@ final class ConversationsAPITests: XCTestCase {
...
+
+        let validateRequest: (URLRequest) -> Bool = { request in
+            guard let body = request.httpBody,
+                  let json = try? JSONSerialization.jsonObject(with: body) as? [String: Any] else {
+                return false
+            }
+
+            // then
+            // when empty access roles are provided, the field should be omitted from the request body.
+            return json["access_role_v2"] == nil
```
johnxnguyen: suggestion: rather than validating the request, we can simply update the snapshots.

johnxnguyen: fixed: 18011ae0cde546b89bb5a4197843505ed1ea8a63

johnxnguyen (approved):

wire-ios-request-strategy/Sources/Payloads/Conversation/Payload+NewConversation.swift
samwyndham: Strange API but OK

samwyndham (approved):

PR #5115, opened 2026-08-10
feat: inform participants of canceled meetings - WPB-25517

acv-w:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25517" title="WPB-25517" target="_blank">TaskWPB-25517</a>  [iOS] Inform participant of a canceled meeting
  </td></table>
  <br />
 

### Issue

Participants are not notified when a host deletes a scheduled meeting.

This change:

- Generates a native cancellation notification before the meeting is removed locally.
- Shows the meeting title and host display name.

### Testing

Using separate host and participant accounts:

1. Enable `wireMeetings` and the team `meetings` flag.
2. Create a scheduled meeting and invite the participant.
3. Wait for it to appear in the participant’s Meetings list.
4. Delete and confirm deletion from the host client.
5. Verify the meeting disappears without interacting with the notification.
6. Verify:
   - Title is the meeting title.
   - Body is `By <Host Display Name> canceled this meeting`.
   - Tapping does not navigate.

---

### Checklist

---


WireDomain/Sources/WireDomain/Notifications/Builders/Meeting/MeetingDeleteEventNotificationBuilder.swift
```diff
@@ -0,0 +1,62 @@
...
+        guard meeting.creatorID.id != accountID else { return nil }
+        guard let host = try? await userLocalStore.fetchUser(
+            id: meeting.creatorID.id,
+            domain: meeting.creatorID.domain
+        ) else { return nil }
+        guard let hostName = await userLocalStore.name(for: host), !hostName.isEmpty else { return nil }
+
+        let content = UNMutableNotificationContent()
+        content.title = meeting.title
+        content.body = "By \(hostName) canceled this meeting"
```
johnxnguyen: issue: this needs to be localized

WireDomain/Sources/WireDomain/Notifications/Components/NSEClientScope.swift
```diff
@@ -381,11 +391,24 @@ final class NSEClientScope: Component<NSEClientScopeDependency> {
     private func generateNotificationsUseCase(eventID: UUID) -> GenerateNotificationUseCase {
         GenerateNotificationUseCase(
             conversationEventBuilder: conversationEventBuilder,
+            meetingDeleteEventBuilder: meetingDeleteEventNotificationBuilder,
```
johnxnguyen: question: do you know if there will be more meeting related notifications? if yes, then it may be nice to group them into a "meeting event notification builder".

WireDomain/Sources/WireDomain/Notifications/Models/NotificationCategory.swift
```diff
@@ -26,11 +26,12 @@ public enum NotificationCategory: String, CaseIterable {
     case incomingCall
     case missedCall
     case incomingConnectionRequest
+    case meetingCancellation
 
     /// Available actions for each category
     private var actions: [NotificationAction] {
         switch self {
-        case .nonActionable:
+        case .nonActionable, .meetingCancellation:
```
johnxnguyen, line 34: question: instead of creating a new category, could we use `nonActionable`?

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
```diff
@@ -313,6 +316,7 @@ public struct IncrementalSync: IncrementalSyncProtocol {
                     "processing live event: \(event.name)",
                     attributes: .incrementalSyncV2 + [.eventEnvelopeID: envelope.id, .eventType: event.name]
                 )
+                await beforeProcessingLiveEvent(event)
```
johnxnguyen, line 319: question: why does the notification need to be scheduled before the event is processed?

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+Push.swift
```diff
@@ -109,6 +109,15 @@ public extension ZMUserSession {
 // The `SessionManager` forwards `UNUserNotificationCenterDelegate` calls to a suitable `ZMUserSession` instance.
 extension ZMUserSession {
 
+    func handleMeetingCancellationNotification(_ content: UNNotificationContent) async {
+        let isActive = await MainActor.run { [application] in
+            application.applicationState == .active
+        }
+        guard isActive || journal[.isConsumableNotificationsEnabled] else { return }
```
johnxnguyen: question: why do we check for consumable notifications here?

WireDomain/Sources/WireDomain/Notifications/Builders/Meeting/MeetingDeleteEventNotificationBuilder.swift
acv-w: Fixing, thanks!

WireDomain/Sources/WireDomain/Notifications/Components/NSEClientScope.swift
acv-w: I completely agree. In the current scope, cancellation is the only one, so a meeting event builder would wrap a single builder. I suggest adding the grouping when another meeting notification is introduced.

WireDomain/Sources/WireDomain/Notifications/Builders/Meeting/MeetingDeleteEventNotificationBuilder.swift
```diff
@@ -0,0 +1,62 @@
...
+        guard meeting.creatorID.id != accountID else { return nil }
+        guard let host = try? await userLocalStore.fetchUser(
+            id: meeting.creatorID.id,
+            domain: meeting.creatorID.domain
+        ) else { return nil }
+        guard let hostName = await userLocalStore.name(for: host), !hostName.isEmpty else { return nil }
+
+        let content = UNMutableNotificationContent()
+        content.title = meeting.title
+        content.body = String.formated(key: "push.notification.body.senderCanceledMeeting", bundle: .module, hostName)
```
KaterinaWire, line 53: **suggestion:** I'm not sure if this is critical, but there is a warning "'module' is deprecated: Use generated Bundle.module instead (package currently doesn't automatically generate .module for the Localizable.strings)"

WireDomain/Sources/WireDomain/Notifications/Models/NotificationCategory.swift
acv-w: Both categories have no custom actions. However, with nonActionable, this notification’s tap would fall through to showContent and open the conversation list because it has no conversation ID. nonActionable opens the conversation list, while a cancellation tap should not navigate anywhere.

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
acv-w: meeting.delete contains only the ID, and processing deletes the cached meeting that provides the title and creator ID. This hook builds and schedules together, so scheduling also happens first.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+Push.swift
acv-w: The check only affects background delivery. In legacy mode the NSE handles it, so scheduling here could duplicate the notification. With consumable notifications, the main app may own the push channel in the background, preventing the NSE from handling it, so it must schedule locally.

KaterinaWire: @johnxnguyen I can imagine that Alex doesn't know the history with consumable notifications and that this code behind` journal[.isConsumableNotificationsEnabled]` is not active now

WireDomain/Sources/WireDomain/Notifications/Builders/Meeting/MeetingDeleteEventNotificationBuilder.swift
acv-w: Good catch. This is an existing WireDomain warning with no generated replacement yet. Fixing it requires a broader migration, so we should address it separately.

WireDomain/Sources/WireDomain/Notifications/Builders/Meeting/MeetingDeleteEventNotificationBuilder.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    func buildContent(event: MeetingDeleteEvent) async -> UserNotification?
+
+}
+
+struct MeetingDeleteEventNotificationBuilder: MeetingDeleteEventNotificationBuilderProtocol {
+
+    let meetingLocalStore: any MeetingLocalStoreProtocol
+    let userLocalStore: any UserLocalStoreProtocol
+    let developerFlagStorage: UserDefaults
```
KaterinaWire, line 35: **suggestion:** We're temporarily hiding the `Wire meeting` feature behind two flags (featureConfig and developerFlag) during development. The `developer flag` should be removed before release. Perhaps you could add a TODO comment next to `let developerFlagStorage: UserDefaults`.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+Push.swift
acv-w: @johnxnguyen @KaterinaWire please let me know if there's something I'm missing here.

WireDomain/Sources/WireDomain/Notifications/Builders/Meeting/MeetingDeleteEventNotificationBuilder.swift
acv-w: Will do, thanks!

WireDomain/Sources/WireDomain/Notifications/Builders/Meeting/MeetingDeleteEventNotificationBuilder.swift
```diff
@@ -0,0 +1,62 @@
...
+    let featureConfigLocalStore: any FeatureConfigLocalStoreProtocol
+    let accountID: UUID
+
+    func buildContent(event: MeetingDeleteEvent) async -> UserNotification? {
+        guard developerFlagStorage.bool(forKey: DeveloperFlag.wireMeetings.rawValue) else { return nil }
```
KaterinaWire, line 40: **issue:** this could be an issue when we remove DeveloperFlag.wireMeetings flag

WireDomain/Sources/WireDomain/Notifications/Components/NSEClientScope.swift
```diff
@@ -154,15 +166,13 @@ final class NSEClientScope: Component<NSEClientScopeDependency> {
...
         } else {
             eventStream = try await pullEventsUseCase.invoke(publicKeys: publicKeys)
+            notifications = try await generateNotificationsUseCase(
+                eventID: eventID
+            ).invoke(
+                updateEvents: eventStream
+            )
         }
 
-        // Generate notifications from events.
```
KaterinaWire, line 159: **question:** I didn't understand the need of the changes in this file. could you please explain?

WireDomain/Sources/WireDomain/Notifications/Builders/Meeting/MeetingDeleteEventNotificationBuilder.swift
acv-w: Yes. When wireMeetings is removed, we’ll remove this guard and developerFlagStorage together. The TODO above tracks that cleanup.

WireDomain/Sources/WireDomain/Notifications/Components/NSEClientScope.swift
acv-w: This flow already existed, but it didn’t handle meeting events. The change adds the meeting notification builder to the existing NSE flow and creates the notification before the cached meeting is deleted.

KaterinaWire: @acv-w  I was asking about the need to remove 
```
let notifications = try await generateNotificationsUseCase(
            eventID: eventID
        ).invoke(
            updateEvents: eventStream
        )
```
and add it separately to the `if dependency.journal[.isConsumableNotificationsEnabled] `and "else" blocks.

KaterinaWire: @johnxnguyen I don't think we need to do something special for `isConsumableNotificationsEnabled`, but correct me if I'm wrong.

acv-w: @KaterinaWire the split was protecting a very specific race when the app foregrounds during consumable notification processing. This is a real possibility but I'm rolling it back as it's probably out of scope for this ticket.

To give more context: in the consumable path we must build it before releasing the push-channel lock. Otherwise, the app could process the delete event and remove the cached meeting first. The else block keeps the existing legacy behavior.

KaterinaWire (approved):

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+Push.swift
johnxnguyen: Consumable notifications isn't active, it was a project started a while ago then ran into issues and got paused.

It's still not clear to me what the check allows. 

> The check only affects background delivery

Do I understand correctly that we want to show this notification from the MA even when the MA is in the background? In which case would this be? During a CallKit call?

johnxnguyen (commented):
Left another question

WireDomain/Sources/WireDomain/Synchronization/IncrementalSync.swift
johnxnguyen: Makes sense. I don't really like it but I know this is a limitation of the code structure 😅

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+Push.swift
acv-w: Since consumable notifications aren’t active, I removed that check.

acv-w:
> Left another question

I responded in the thread.

wire-ios-sync-engine/Source/UserSession/ZMUserSession/ZMUserSession+Push.swift
KaterinaWire: > Do I understand correctly that we want to show this notification from the MA even when the MA is in the background? In which case would this be? During a CallKit call?

@johnxnguyen here's a design https://www.figma.com/design/RsDpGg0Ah7RqlyogqBDzYJ/Meetings?node-id=420-87553&m=dev
it's not about CallKit call, just a regular notification about the WM

johnxnguyen (approved):

PR #5164, opened 2026-08-25
fix: Fix landscape rotation bubble margins - WPB-21023

valerio-bettini:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21023" title="WPB-21023" target="_blank">BugWPB-21023</a>  [iOS] Chat view renders incorrectly when device is rotated to landscape mode
  </td></table>
  <br />
 

### Description 
Fixes incorrect bubble margins after rotation by changing when the conversation resets its section controllers: instead of only resetting on iPad, we now reset only when the contentWidth actually changes. This is a single-line change in ConversationTableViewDataSource.swift.

### What changed

`ConversationTableViewDataSource.swift`: replace
`guard UIDevice.current.userInterfaceIdiom == .pad else { return }` with
`guard contentWidth != oldValue else { return }`

### Result: 
`resetSectionControllers()` is triggered when the measured content width changes (any device).

### How to test
Open a conversation with short/long messages and attachments.
Rotate on iPhone and iPad (portrait ⇄ landscape): verify bubbles don’t clip and margins/alignment are correct.
Confirm scroll position and composer/keyboard state are preserved.
Run UI/snapshot tests if available.

### Risk 
Low — change is localized to the conversation data source; main risk is reload-related side effects (scroll/keyboard). Run the checks above before merging.

### DIFF
--- a/WireUI/Sources/WireMainNavigationUI/Containers/MainSplitViewController.swift
+++ b/WireUI/Sources/WireMainNavigationUI/Containers/MainSplitViewController.swift
@@ -143,6 +143,11 @@ public final class MainSplitViewController<Sidebar, TabController>: UISplitViewC
         with coordinator: any UIViewControllerTransitionCoordinator
     ) {
         super.viewWillTransition(to: size, with: coordinator)
+
+        // On iPhone the split view must never tile into two columns, even during a
+        // transient size change from a rotation attempt that gets reverted (the phone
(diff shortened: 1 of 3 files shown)


samwyndham (approved):
🥳

jullianm (approved):
LGTM

PR #5167, opened 2026-08-26
chore: avoid debug logs marked as safe public - WPB-28310

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28310" title="WPB-28310" target="_blank">TaskWPB-28310</a>  [iOS] Avoid public but debug logs
  </td></table>
  <br />
 

### Issue

We have a number of logs whose attributes mark them as `(safe)public`. These logs appear in logs files that can be shared from production builds. However, when logging with the `debug` log level, these are stripped - i.e. there is a contradiction here between the developers intent and what is actually logged.

In this PR I promote the log level of some `debug` logs to `info` where:

1. It is safe to do so. e.g. we are not exposing personal information
2. It should not affect performance. e.g. I avoid this change in tight loops.

The main reason for this change is due to missing logs in a bug ticket I'm currently looking into - especially related to sync starting / stopping within `IncrementalSync.internalPerform(backgroundAccessibleOnly:)`

### Testing

Check that logs appear in log files.

---

### Checklist


wire-ios-data-model/Source/Core Crypto/CoreCryptoProvider.swift
```diff
@@ -257,7 +257,7 @@ public actor CoreCryptoProvider: CoreCryptoProviderProtocol {
     // https://forums.swift.org/t/structured-caching-in-an-actor/65501/13
     private func getCoreCrypto() async throws -> CoreCrypto {
         guard !loadingCoreCrypto else {
-            WireLogger.coreCrypto.debug(
+            WireLogger.coreCrypto.info(
```
jullianm, line 260: question: does it make sense to enforce that at the `WireLogger.debug` method level directly i.e adding a precondition that a debug log cannot have `.safePublic` attributes ?

jullianm (approved):
LGTM, left one question

johnxnguyen (approved):

wire-ios-data-model/Source/Core Crypto/CoreCryptoProvider.swift
samwyndham: I think it could be a good idea but will avoid doing it now due to time constraints - I would have to also fix a few other areas that I'm not sure about. But with soon planned changes there will no longer be the concept of safe public because logs will already be safe at compile time and log level will be the single determining factor for whether we add a log line to the logs in public.

PR #5195, opened 2026-09-08
fix: prevent ringing for unnamed two-person meetings - WPB-28207

acv-w:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28207" title="WPB-28207" target="_blank">BugWPB-28207</a>  [iOS] Invitee participant receives ringing ringtone for Meet Now/scheduled Meeting
  </td></table>
  <br />
 

### Issue

Unnamed team meetings with just the host and one invitee could be treated as one-to-one conversations, causing invitees to ring.

Keep meetings classified as groups, while supporting migrations from older databases that lack `groupType`.

---

### Checklist

---


jullianm (approved):

wire-ios-data-model/Source/Model/Conversation/ZMConversation.m
```diff
@@ -463,6 +463,8 @@ - (ZMConversationType)conversationType
     // local participant roles, so check its count first to avoid unncessary iterations.
 
     if (conversationType == ZMConversationTypeGroup &&
+        (self.entity.attributesByName[ZMConversationGroupTypeKey] == nil ||
```
netbe, line 466: question: i don't really understand this condition, what defines a meeting conversation?

also the comments should be updated if needed

netbe (commented):
need more details before approving

wire-ios-data-model/Source/Model/Conversation/ZMConversation.m
acv-w: Meetings are groups marked with `groupType == .meeting`. They should stay groups even when unnamed and containing only two people.

The first check handles older databases without `groupType`. We can’t use `isMeeting` here because it calls this same getter. I’ll update the comments to explain both points.

netbe (approved):
all good

PR #5209, opened 2026-09-09
fix: Suppress meeting call notifications sound - WPB-28207

acv-w:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28207" title="WPB-28207" target="_blank">BugWPB-28207</a>  [iOS] Invitee participant receives ringing ringtone for Meet Now/scheduled Meeting
  </td></table>
  <br />
 

### Issue

Meeting call events can arrive before the app knows the conversation is a meeting, causing incoming or missed call notifications with sound.

Fetch missing conversation details before allowing a call notification. Also prevent suppressed call events from falling back to a generic message notification.

---

### Checklist

---


WireDomain/Sources/WireDomain/Notifications/Builders/Conversation/ConversationCallingEventNotificationBuilder.swift
```diff
@@ -474,11 +476,28 @@ extension ConversationCallingEventNotificationBuilder {
...
+                return false
+            }
+
+            guard needsBackendUpdate else { return true }
+
+            // Calls can decrypt before the NSE has the conversation's meeting metadata.
+            guard let qualifiedID = await conversationLocalStore.qualifiedID(for: conversation) else { return false }
+
+            do {
+                let response = try await conversationsAPI.getConversations(for: [WireNetwork.QualifiedID(qualifiedID)])
```
netbe, line 489: question: So we are fetching the whole conversation to know if it's a meeting conversation, did i understand that right ? If so, I wonder if the info should be in the encrypted message then

netbe (commented):

WireDomain/Sources/WireDomain/Notifications/Builders/Conversation/ConversationCallingEventNotificationBuilder.swift
acv-w: Yes, carrying the flag in the encrypted message would avoid the fetch. But we’d still need the lookup for older clients, so it adds complexity for now. I’d keep this PR focused on the fix and consider the flag as a separate improvement.

netbe: ok, let's go with it for now. but this might be the responsibility of avs. The iOS app goes into some hacks and looking at the encrypted message because avs is not in the NSE (for now) to decide to ring (wake the app) or not.

netbe (approved):
thanks for the info. approved

caldrian (approved):

PR #5213, opened 2026-09-10
fix: session expired alert not shown when removed from team - WPB-24285

netbe:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24285" title="WPB-24285" target="_blank">BugWPB-24285</a>  [iOS] No alert shown to user when they are logged out after being removed from the team
  </td></table>
  <br />
 

## Summary
- Team removal (and any other `.accessTokenExpired` session invalidation) now surfaces the existing "Your session expired" alert instead of silently logging the user out. Root cause: the backend revokes the access token before the account-deleted event is processed, so the logout went through a code path that `AppRootRouter.presentAlertForDeletedAccountIfNeeded` didn't check.
- Alert presentation (for this and the two sibling cases in the same method) now presents on `UIApplication.topmostViewController(onlyFullScreen: false)` instead of `rootViewController` directly, so it isn't silently dropped when a SwiftUI sheet is already presented.
- `AccountStore.deleteAccount` no longer logs a spurious error when the account file was already removed.
- Added `WireUITests.TeamManageTests.test_TeamMemberRemovedFromTeam_SeesSessionExpiredAlert` (TC-6091, TC-6258) to reproduce the scenario, backed by a new `#if DEBUG`-only `TeamsAPI.removeMemberFromTeam` endpoint (available from API v5 onward).

## Test plan
- [x] Run `TeamManageTests.test_TeamMemberRemovedFromTeam_SeesSessionExpiredAlert` against a real backend
- [x] Manually verify: log in as team member B, have team owner A remove B from the team, confirm B sees "Your session expired" and lands on the welcome screen
- [ ] Confirm no regression for existing `.accountDeleted` alert flows (missing passcode, database wiped)

### DIFF
--- a/WireDomain/Sources/WireDomain/Account/AccountStore.swift
+++ b/WireDomain/Sources/WireDomain/Account/AccountStore.swift
@@ -117,19 +117,16 @@ struct AccountStore {
 
     @discardableResult
     func deleteAccount(_ account: Account) -> Bool {
-        let url = url(for: account.userIdentifier)
-        guard fileManager.fileExists(atPath: url.path(percentEncoded: false)) else {
-            // Already deleted, nothing to do.
-            return true
-        }
-
         do {
-            try fileManager.removeItem(at: url)
+            try fileManager.removeItem(at: url(for: account.userIdentifier))
             return true
-        } catch {
-            let accountDescription = account.safeForLoggingDescription
-            let errorDescription = error.safeForLoggingDescription
-            log.error("Unable to delete account \(accountDescription), error: \(errorDescription)")
+        } catch let error as NSError {
+            // Already deleted, nothing to do, and no need to log a spurious error.
+            if error.domain != NSCocoaErrorDomain || error.code != NSFileNoSuchFileError {
+                let accountDescription = account.safeForLoggingDescription
+                let errorDescription = error.safeForLoggingDescription
+                log.error("Unable to delete account \(accountDescription), error: \(errorDescription)")
+            }
             return false
         }
     }
--- a/WireDomain/Sources/WireDomain/Account/AccountStore.swift
+++ b/WireDomain/Sources/WireDomain/Account/AccountStore.swift
@@ -117,8 +117,14 @@ struct AccountStore {
 
     @discardableResult
     func deleteAccount(_ account: Account) -> Bool {
+        let url = url(for: account.userIdentifier)
+        guard fileManager.fileExists(atPath: url.path(percentEncoded: false)) else {
+            // Already deleted, nothing to do.
+            return true
+        }
+
         do {
-            try fileManager.removeItem(at: url(for: account.userIdentifier))
+            try fileManager.removeItem(at: url)
             return true
         } catch {
             let accountDescription = account.safeForLoggingDescription
--- a/WireNetwork/Sources/WireNetwork/APIs/Rest/TeamsAPI/TeamsAPI.swift
+++ b/WireNetwork/Sources/WireNetwork/APIs/Rest/TeamsAPI/TeamsAPI.swift
@@ -95,6 +95,21 @@ public protocol TeamsAPI {
         ) async throws -> UUID
     #endif
 
+    /// Remove a member from a team.
+    /// - Parameters:
+    ///   - access_token: access token of the user performing the removal (e.g. the team owner).
+    ///   - teamID: the id of the team.
+    ///   - userID: the id of the member to remove.
+    ///   - password: password of the user performing the removal.
+    #if DEBUG
+        func removeMemberFromTeam(
+            access_token: String,
+            teamID: UUID,
+            userID: UUID,
+            password: String
+        ) async throws
+    #endif
+
     func getWhitelistedBots(
         for teamID: Team.ID,
         with prefix: String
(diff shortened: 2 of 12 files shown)


wire-ios/Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -23,8 +23,6 @@
 /* Begin PBXBuildFile section */
 		0143895E2BC81DC200EC6920 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 0143895D2BC81DC200EC6920 /* PrivacyInfo.xcprivacy */; };
 		0145AE992B1156FC0097E3B8 /* WireSyncEngineSupport.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 0145AE982B1156FC0097E3B8 /* WireSyncEngineSupport.framework */; };
-		FADE000000000000000000A2 /* WireDomainSupport.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = FADE000000000000000000A1 /* WireDomainSupport.framework */; };
```
netbe, line 26: todo: remove changes here

wire-ios/WireUITests/TeamManageTests.swift
```diff
@@ -164,6 +164,38 @@ final class TeamManageTests: WireUITestCase {
         )
     }
 
+    /// [WPB-24825] Bug: no alert shown when the self user is removed from the team
```
samwyndham: nitpick: I'm not sure it is necessary to a comment referencing this bug ... long term our code might fill with these kinds of comments.

wire-ios/WireUITests/TeamManageTests.swift
```diff
@@ -164,6 +164,38 @@ final class TeamManageTests: WireUITestCase {
...
+    /// TC-6091, TC-6258
```
samwyndham: question: Is this enough for the testiny test cases to be marked as passed automatically?

WireNetwork/Sources/WireNetwork/APIs/Rest/TeamsAPI/TeamsAPI.swift
```diff
@@ -95,6 +95,21 @@ public protocol TeamsAPI {
...
+    ///   - password: password of the user performing the removal.
+    #if DEBUG
+        func removeMemberFromTeam(
+            access_token: String,
+            teamID: UUID,
+            userID: UUID,
+            password: String
+        ) async throws
+    #endif
+
```
samwyndham, line 112: question: Should the implementations also be `if #DEBUG`?

WireDomain/Sources/WireDomain/Account/AccountStore.swift
```diff
@@ -120,10 +120,13 @@ struct AccountStore {
         do {
             try fileManager.removeItem(at: url(for: account.userIdentifier))
             return true
-        } catch {
-            let accountDescription = account.safeForLoggingDescription
-            let errorDescription = error.safeForLoggingDescription
-            log.error("Unable to delete account \(accountDescription), error: \(errorDescription)")
+        } catch let error as NSError {
+            // Already deleted, nothing to do, and no need to log a spurious error.
+            if error.domain != NSCocoaErrorDomain || error.code != NSFileNoSuchFileError {
```
samwyndham: suggestion pattern match against `CocoaError.fileNoSuchFile`

samwyndham (approved):
Nice work

WireDomain/Sources/WireDomain/Account/AccountStore.swift
netbe: could you elaborate ? I am not sure to follow

WireNetwork/Sources/WireNetwork/APIs/Rest/TeamsAPI/TeamsAPI.swift
netbe: not necessarly, do you think it's needed?  The goal is that only the UITest code is using this

wire-ios/WireUITests/TeamManageTests.swift
netbe: no, but I'll fix this with Manish or Antje

WireNetwork/Sources/WireNetwork/APIs/Rest/TeamsAPI/TeamsAPI.swift
samwyndham: I guess adding the #if DEBUG is nice because:

1. Less code in the finally binary (a tiny bit)
2. We could accidentally depend on the implementations if accessed not via the protocol. 

But it's minor so up to you if you want to change that.

WireDomain/Sources/WireDomain/Account/AccountStore.swift
samwyndham: I mean you could do something like:

```swift
do {
  // ..
} catch CocoaError.fileNoSuchFile {
  // Already deleted, nothing to do, and no need to log a spurious error.
  return false
} catch {
  // old implementation
}
```

WireNetwork/Sources/WireNetwork/APIs/Rest/TeamsAPI/TeamsAPI.swift
netbe: i'll keep it as it is now so we match existing pattern

jullianm (approved):
LGTM

PR #5246, opened 2026-09-16
chore: in-house runners builds - WPB-28349

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28349" title="WPB-28349" target="_blank">TaskWPB-28349</a>  [iOS] Migrate Playground builds workflows to run on in-house runners
  </td></table>
  <br />
 

### Issue

This PR updates our deployment workflows (playground, edge, beta, etc) to allow choosing which mac runner they use for the build & release step from:

- cirrus (the current solution)
- in-house (our mac minis)
- github

Currently we default to `cirrus` so everything should keep working without issue. `in-house` may work but I'm still setting up the runners so best avoided for now. This will become the default soon though. `github` uses `githubs` runners... however it looks like they've recently stopped supported iOS 26.1 simulator so it won't work until we update the simulator we use.

### Testing

Start a github action build

---

### Checklist


.github/actions/cache-swift-packages/action.yaml
```diff
@@ -1,10 +1,16 @@
 name: "Cache Swift Packages"
 description: "Restore the SPM checkouts cache and always reconcile workspace state. Sharing one cache across flag combinations is safe because the cloned packages (checkouts/) don't depend on ENABLE_COUNTLY/ENABLE_DATADOG — WireAnalytics/Package.swift's WarningPrevention target keeps Countly and Datadog in the resolved graph unconditionally. The env-conditional bits (manifest.db, workspace-state.json) are regenerated by the always-on resolve step against the caller's current env. Requires PACKAGES_DIR and REPO_ROOT to be set in env."
 
+inputs:
+  cache_enabled:
+    description: "Whether to restore and save the Swift package cache"
+    default: "true"
+
```
samwyndham, line 8: For in-house builds we don't want to cache packages as this is super slow and the unbuilt pacaagkes are already cached on the mac minis.

.github/workflows/_reusable_app_release.yml
```diff
@@ -142,13 +150,16 @@ jobs:
...
+        with:
+          cache_enabled: ${{ inputs.mac_runner != 'in-house' }}
 
       - name: Setup Ruby
         uses: ruby/setup-ruby@afeafc3d1ab54a631816aba4c914a0081c12ff2f # v1.310.0
         with:
           bundler-cache: true
           
       - name: Install ImageMagick
+        if: ${{ inputs.mac_runner != 'in-house' }}
```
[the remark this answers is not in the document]
samwyndham: Nothing to fix here. We install imagemagick (and all other global software) when setting up the mac minis (not during running workflows)

.github/workflows/beta_app_release.yml
```diff
@@ -25,6 +33,7 @@ jobs:
       countly_enabled: true
       datadog_enabled: true
       fastlane_action: testflight_beta
+      mac_runner: ${{ inputs.mac_runner || 'cirrus' }}
```
netbe, line 36: question: do we have to do `||` as there is a default value?

netbe (approved):
Looks good.

todo: Don't forget to modify workflows for previous supported branches: LTS_RELEASE and GOV_RELEASE (release/cycle-3.120, release/cycle-4.16)

thought: I wonder if we need this level of flexibility to choose the runner. Or using a env variable from github settings would be enough.

.github/workflows/beta_app_release.yml
findms: Q: I just checked the playground build [logs](https://github.com/wireapp/wire-ios/actions/runs/35112062335/job/104848063603) and I could see  _mac_runner: cirrus_  is this expected? @samwyndham

findms (approved):
lgtm. added a question.

.github/workflows/beta_app_release.yml
samwyndham: @findms Yes those logs are when I tried with cirrus. See these builds for in house. I'm running 4 now simultaneously:

- https://github.com/wireapp/wire-ios/actions/runs/35215082126/job/105181857890
- https://github.com/wireapp/wire-ios/actions/runs/35214881744/job/105181224653
- https://github.com/wireapp/wire-ios/actions/runs/35207530607/job/105176822988
- https://github.com/wireapp/wire-ios/actions/runs/35215082126

samwyndham: > question: do we have to do || as there is a default value?

@netbe Yes because `inputs.mac_runner` is only set when we manually start builds, not for automatic builds when merging to release for example.

PR #5278, opened 2026-09-21
chore: enable in house delivery builds - WPB-28349

samwyndham:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28349" title="WPB-28349" target="_blank">TaskWPB-28349</a>  [iOS] Migrate Playground builds workflows to run on in-house runners
  </td></table>
  <br />
 

### Issue

This PR enables delivery workflows to run on `in-house` runners by default. For non manual runs this is defined by the `DEFAULT_MAC_RUNNER` env variable.

It also configures delivery jobs to happen on a runner with the "delivery" label and "test" jobs to happen on runners with the "test" label. The current usage is to ensure that there are no concurrent delivery builds on the same mac mini as there is currently an issue doing this (which I'll work on next week). 

Here are our list of runners - note we have 2 "delivery" runners, one on each mac mini.

Screenshot 2026-09-21 at 16 43 26

### Testing

Check that this playground build succeeds: https://github.com/wireapp/wire-ios/actions/runs/35613563967/job/106378933951

---

### Checklist


.github/workflows/beta_app_release.yml
```diff
@@ -13,7 +13,7 @@ on:
           - in-house
           - cirrus
           - github
-        default: cirrus
+        default: in-house
```
netbe, line 16: question: can't you use the vars.DEFAULT_MAC_RUNNER here too?

.github/workflows/c1_c2_c3_app_release_restricted.yml
```diff
@@ -10,7 +10,7 @@ on:
           - in-house
           - cirrus
           - github
-        default: cirrus
+        default: in-house
```
netbe, line 13: question: same here vars.DEFAULT_MAC_RUNNER, is it possible?

netbe (approved):
LGTM, left a question

.github/workflows/beta_app_release.yml
samwyndham: No, it's not possible as the workflow_dispatch defaults must be static

.github/workflows/c1_c2_c3_app_release_restricted.yml
samwyndham: See my comment other comment

acv-w (approved):

